Issue #548, opened 2019-04-18
testOrQuery5WithLimits intermittently fails with a NoSuchElementException


alecgrieser:
The `testOrQuery5WithLimits` test has intermittently failed with `NoSuchElementException` trying to pop from the queue:

```
10:36:59 com.apple.foundationdb.record.provider.foundationdb.query.FDBOrQueryToUnionTest > testOrQuery5WithLimits(int)[1] FAILED
10:36:59     java.util.NoSuchElementException
10:36:59         at java.util.ArrayDeque.removeFirst(ArrayDeque.java:285)
10:36:59         at java.util.ArrayDeque.remove(ArrayDeque.java:452)
10:36:59         at com.apple.foundationdb.record.cursors.MapPipelinedCursor.close(MapPipelinedCursor.java:148)
10:36:59         at com.apple.foundationdb.record.cursors.MapCursor.close(MapCursor.java:113)
10:36:59         at com.apple.foundationdb.record.provider.foundationdb.cursors.MergeCursorState.close(MergeCursorState.java:97)
10:36:59         at java.util.ArrayList.forEach(ArrayList.java:1257)
10:36:59         at com.apple.foundationdb.record.provider.foundationdb.cursors.MergeCursor.close(MergeCursor.java:349)
10:36:59         at com.apple.foundationdb.record.provider.foundationdb.cursors.UnorderedUnionCursor.close(UnorderedUnionCursor.java:58)
10:36:59         at com.apple.foundationdb.record.cursors.FilterCursor.close(FilterCursor.java:122)
10:36:59         at com.apple.foundationdb.record.cursors.RowLimitedCursor.close(RowLimitedCursor.java:126)
10:36:59         at com.apple.foundationdb.record.RecordCursorIterator.close(RecordCursorIterator.java:149)
10:36:59         at com.apple.foundationdb.record.provider.foundationdb.query.FDBOrQueryToUnionTest.testOrQuery5WithLimits(FDBOrQueryToUnionTest.java:416)
```

It is referencing this part of the code here:

https://github.com/FoundationDB/fdb-record-layer/blob/cb1fe996d703ab0c661844c6957f5e55e5e0a7ca/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java#L141-L155

[...]

PR #820, opened 2020-02-11
Resolves #817: Index build process should track number of base records processed

### LINKED ISSUE #817: Index build process should track number of base records processed
When building an index, the OnlineIndex build process should keep track of the number of base records that have been processed in building of the index.  This, in conjunction with the number of records in the store (as given by the record count index) can provide an approximation of the % completion of the index build process.

nblintao:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A class that contains the build progress of a given index in a given record store.
+ * <p>
+ * It first contains {@link IndexState}. If the state is {@link IndexState#WRITE_ONLY}, it will also contains the number
+ * of records scanned and the estimated total number of records that need to be scanned.
+ * </p>
+ * <p>
+ * Note that index build state (especially records scanned) serves a informational purpose only in terms that records scanned is not necessarily 100% of records in total when the index build completes in several reasons:
```
alecgrieser:
```suggestion
 * Note that index build state (especially records scanned) serves an informational purpose only as the number of records scanned is not necessarily 100% of records in total when the index build completes for several reasons:
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+    public static CompletableFuture<IndexBuildState> getIndexBuildStateAsync(FDBRecordStore store, Index index) {
+        IndexState indexState = store.getIndexState(index);
+        if (indexState != IndexState.WRITE_ONLY) {
+            return CompletableFuture.completedFuture(new IndexBuildState(indexState));
+        }
+        return getRecordsScannedAsync(store, index).thenCompose(scannedRecords -> {
+            try {
+                return store.getSnapshotRecordCount()
+                        .thenApply(totalRecords -> new IndexBuildState(indexState, scannedRecords, totalRecords));
+            } catch (Exception ex) {
```
alecgrieser: This should probably catch `RecordCoreException`. It's also possible that we should introduce a new exception type, like, `AggregateFunctionNotSupported` or something, and that that function should throw that error, but catching `RecordCoreException` is probably fine for now. (And the exception can be changed safely if `AggregateFunctionNotSupported` is a sub-type of `RecordCoreException`.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+     * @param store the record store containing the index
+     * @param index the index needed to be checked
+     * @return a future that completes to the index build state
+     */
+    public static CompletableFuture<IndexBuildState> getIndexBuildStateAsync(FDBRecordStore store, Index index) {
+        IndexState indexState = store.getIndexState(index);
+        if (indexState != IndexState.WRITE_ONLY) {
+            return CompletableFuture.completedFuture(new IndexBuildState(indexState));
+        }
+        return getRecordsScannedAsync(store, index).thenCompose(scannedRecords -> {
```
alecgrieser: The scanned records and total records queries should be executable in parallel and then combinable into the `IndexBuildState` object, I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+    @Nullable
+    private final Long recordsInTotal;
+
+    /**
+     * Get the build progress ({@link IndexBuildState}) of the given index in the given record store asynchronously.
+     * @param store the record store containing the index
+     * @param index the index needed to be checked
+     * @return a future that completes to the index build state
+     */
+    public static CompletableFuture<IndexBuildState> getIndexBuildStateAsync(FDBRecordStore store, Index index) {
```
alecgrieser: I think `FDBRecordStore` (here and elsewhere) could be `FDBRecordStoreBase<?>` (to avoid us taking an opinion here on which implementation of the interface is used).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * Get the number of records successfully scanned and processed during the online index build process
+     * asynchronously.
+     * <p>
+     * This does not include the records scanned in the {@link OnlineIndexer} whose
+     * {@link OnlineIndexer.Builder#setTrackProgress(boolean)} is set to false.
+     * </p>
+     * @param store the record store containing the index
+     * @param index the index needed to be checked
+     * @return a future that completes to the total records scanned.
```
alecgrieser:
```suggestion
     * @return a future that completes to the total records scanned
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+        return store.getContext().ensureActive()
+                .get(OnlineIndexer.indexBuildScannedRecordsSubspace(store, index).getKey())
+                .thenApply(FDBRecordStore::decodeRecordCount);
+    }
+
+    /**
+     * Get the index state.
+     * @return the index state
+     */
+    public IndexState getIndexState() {
```
alecgrieser: These accessor methods are missing nullability annotations on their return types (especially important for the nullable ones).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+     * </p>
+     * @return the number of scanned records
+     * @see #getRecordsScannedAsync(FDBRecordStore, Index)
+     */
+    public Long getRecordsScanned() {
+        return recordsScanned;
+    }
+
+    /**
+     * The the estimated total number of records that need to be scanned to build the index. Currently, it uses the
```
alecgrieser:
```suggestion
     * Get the estimated total number of records that need to be scanned to build the index. Currently, it uses the
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+    private IndexBuildState(IndexState indexState, long recordsScanned, @Nullable Long recordsInTotal) {
+        this.indexState = indexState;
+        this.recordsScanned = recordsScanned;
+        this.recordsInTotal = recordsInTotal;
+    }
+
+    private IndexBuildState(IndexState indexState) {
+        this.indexState = indexState;
+        this.recordsScanned = null;
+        this.recordsInTotal = null;
```
alecgrieser: Probably `this(indexState, null, null)` instead

alecgrieser:
Whoops, sorry, didn't mean to close. I think I hit that button accidentally.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    public IndexState getIndexState() {
+        return indexState;
+    }
+
+    /**
+     * Get the number of records successfully scanned and processed during the online index build process.
+     * <p>
+     * This does not include the records scanned in the {@link OnlineIndexer} whose
+     * {@link OnlineIndexer.Builder#setTrackProgress(boolean)} is set to false.
```
alecgrieser: Maybe something like, "If the `OnlineIndexer` has been configured to disable tracking progress (by setting `setTrackProgress` to `false`), then the number returned will not reflect the actual number of records scanned." Or something like that?

I wonder if it is worth also clarifying here that this won't include all records added during the build, or if that should be only in one place (at the top, probably) and the other place(s) we want to mention it should reference that one place.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,160 @@
...
+     * asynchronously.
+     * <p>
+     * This does not include the records scanned in the {@link OnlineIndexer} whose
+     * {@link OnlineIndexer.Builder#setTrackProgress(boolean)} is set to false.
+     * </p>
+     * @param store the record store containing the index
+     * @param index the index needed to be checked
+     * @return a future that completes to the total records scanned.
+     */
+    public static CompletableFuture<Long> getRecordsScannedAsync(FDBRecordStore store, Index index) {
```
alecgrieser: Is this intentionally a public method? Do we expect people to call it directly?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildIndexTest.java
```diff
@@ -255,6 +257,18 @@ void singleRebuild(@Nonnull List<TestRecords1Proto.MySimpleRecord> records, @Nul
...
                 additionalScans += (long)recordsWhileBuilding.size();
             }
+
+            if (!wasReadableBeforeBuild) {
+                try (FDBRecordContext context = openContext()) {
+                    IndexBuildState indexBuildState = context.asyncToSync(FDBStoreTimer.Waits.WAIT_GET_INDEX_BUILD_STATES,
+                            IndexBuildState.getIndexBuildStateAsync(recordStore, index));
+                    assertEquals(IndexState.WRITE_ONLY, indexBuildState.getIndexState());
+                    assertEquals(indexBuilder.getTotalRecordsScanned(), indexBuildState.getRecordsScanned());
+                    // Count index is not defined so we cannot lean the records in total from it.
```
alecgrieser:
```suggestion
                    // Count index is not defined so we cannot determine the records in total from it.
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildIndexTest.java
```diff
@@ -255,6 +257,18 @@ void singleRebuild(@Nonnull List<TestRecords1Proto.MySimpleRecord> records, @Nul
             if (recordsWhileBuilding != null && recordsWhileBuilding.size() > 0) {
                 additionalScans += (long)recordsWhileBuilding.size();
             }
+
+            if (!wasReadableBeforeBuild) {
```
alecgrieser: Why does this depend on `wasReadableBeforeBuild` being false?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -2119,6 +2137,19 @@ public Builder setProgressLogIntervalMillis(long millis) {
             return this;
         }
 
+        /**
+         * Set whether or not to track the index build progress by updating the number of records successfully scanned
+         * and processed. The progress is persisted in {@link #indexBuildScannedRecordsSubspace(FDBRecordStore, Index)}
+         * which can be accessed by {@link IndexBuildState#getIndexBuildStateAsync(FDBRecordStore, Index)}, comparing
+         * to the records scanned information in progress log which cannot be accessed programmatically.
```
alecgrieser: I'm not sure what the last sentence here is trying to say. That the object allows one to compare the records scanned with the records in the record store? I'm not sure where the "programmatically" bit comes in.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
nblintao: Fixed wordings for "track progress".

I didn't include clarifications for "all records added during the build" here because strictly speaking they are not "records scanned". But this clarification is useful when it is put under the context of comparing with the records in total. So I guess the top of `IndexBuildState` might be a better place. Added a point to the top.

nblintao:
https://fdb-record-layer-ci.foundationdb.org/job/jobs/job/prb-proto2/1453/testReport/junit/com.apple.foundationdb.record.provider.foundationdb.query/FDBOrQueryToUnionTest/testOrQuery5WithLimits_int__boolean__10_/
Looks ephemeral and unrelated but might be worrisome.
Test this please.

alecgrieser:
> Looks ephemeral and unrelated but might be worrisome.

Is it another instance of #548?

nblintao:
@alecgrieser I think so.

nblintao:
https://fdb-record-layer-ci.foundationdb.org/job/jobs/job/prb-proto3/1427/testReport/junit/com.apple.foundationdb.record.provider.foundationdb.query/FDBOrQueryToUnionTest/testOrQuery5WithLimits_int__boolean__14_/
Another one. Test this please.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/AggregateFunctionNotSupported.java
```diff
@@ -0,0 +1,34 @@
...
+package com.apple.foundationdb.record;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Exception thrown when an aggregate function is not supported.
+ */
+@SuppressWarnings("serial")
+public class AggregateFunctionNotSupported extends RecordCoreException {
```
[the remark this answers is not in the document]
nblintao: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1597,6 +1613,8 @@ public Config build() {
         protected int maxRetries = DEFAULT_MAX_RETRIES;
         protected int recordsPerSecond = DEFAULT_RECORDS_PER_SECOND;
         private long progressLogIntervalMillis = DEFAULT_PROGRESS_LOG_INTERVAL;
+        // TODO: Maybe the performance impact of this is low enough to be always enabled?
```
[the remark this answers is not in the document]
nblintao: OK. I'll keep it as it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/AggregateFunctionNotSupportedException.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * AggregateFunctionNotSupported.java
```
alecgrieser:
```suggestion
 * AggregateFunctionNotSupportedException.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -387,6 +387,8 @@ public String logKey() {
         WAIT_SCAN_RECORDS("wait for scan records"),
         /** Wait for a indexed record scan. */
         WAIT_SCAN_INDEX_RECORDS("wait for scan index records"),
+        /** Wait for getting index build states. */
+        WAIT_GET_INDEX_BUILD_STATES("wait for getting index build states"),
```
scgray: Super pedantic, feel free to ignore, but there is only one build state being waited on, so it should probably be `WAIT_GET_INDEX_BUILD_STATE` and `wait for getting index build state`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
```diff
@@ -0,0 +1,174 @@
...
+                (scannedRecords, recordsInTotal) -> new IndexBuildState(indexState, scannedRecords, recordsInTotal));
+    }
+
+    /**
+     * Load the number of records successfully scanned and processed during the online index build process
+     * asynchronously.
+     * <p>
+     * If the {@link OnlineIndexer} has been configured to disable tracking progress (by setting
+     * {@link OnlineIndexer.Builder#setTrackProgress(boolean)} to {@code false}), then the number returned will not
+     * reflect the actual number of records scanned.
```
scgray: In this and `getRecordsScanned`, when the tracking feature is disabled, the phrase `the number returned will not reflect that actual number of records scanned` is a little ambiguous, particularly given that the `Long` returned can also be `null`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -560,6 +566,11 @@ private void increaseLimit() {
             if (recordsScanned != null) {
                 recordsScanned.incrementAndGet();
             }
+            if (trackProgress) {
+                // Add one to the number of records successfully scanned and processed.
+                store.context.ensureActive().mutate(
+                        MutationType.ADD, scannedRecordsSubspace.getKey(), FDBRecordStore.LITTLE_ENDIAN_INT64_ONE);
```
scgray: @alecgrieser did not raise this as an issue, so I doubt it is one, but would there be any benefit to doing this mutation only once, just before the commit of the scanned range, rather than doing it on each record scanned?  For example, in the proposed RPC interface to FDB (https://github.com/apple/foundationdb/wiki/FoundationDB-RPC-Layer-Requirements) this operation would, presumably, add a round trip each time.

alecgrieser: Hm. With the current client implementation, the only benefit to doing it all at once at the end would be that there would be approximately one fewer JNI hop and a little bit of CPU work on the client, but because the FDB client coalesces mutations, it wouldn't end up issuing one mutation per record here, it would just be one single mutation at the end (similar to if the coalescing were done at the Record Layer level).

I think the plan (or hope) would be that an RPC layer would have some kind of client that also batched updates together in some way (so that, for example, multiple mutations wound up on the same packet to the RPC server), though the overhead would be a little bit higher than if there was just one mutation (for example, you would need to repeat the key multiple times). But that's also somewhat speculative, on my part, given that I think that's still a little up in the air.

So, in conclusion, probably would be a non-zero savings to only have mutation, but also probably not all that expensive the way it is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexBuildState.java
nblintao: Changed to "then the number returned will not include the records scanned by that {@link OnlineIndexer}"

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
nblintao: Collapsed the counts, considering it won't hurt anyways. (I didn't do it initially mainly because I didn't want to write things to encode a `long`)

PR #829, opened 2020-02-20
Resolves #823: Need a way to tell whether a QueryPlan has a RecordQueryLoadByKeysPlan

### LINKED ISSUE #823: Need a way to tell whether a QueryPlan has a RecordQueryLoadByKeysPlan
Along the lines of `hasRecordScan` and `hasIndexScan`.

This could be a new predicate, which would then need to be implemented by all parents, or a new general mechanism to check predicates through the plan tree.

MMcM:
I considered making this have a default of `false` for everyone. But that seemed like it might be dangerous in general. So, as a compromise, I made plans _with children_ default this and other similar methods by passing on the each child. I could be convinced not to even do that, or of some more ambitious scheme for general predicate checking over children.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
@@ -117,6 +117,11 @@ public class RecordQueryLoadByKeysPlan implements RecordQueryPlanWithNoChildren 
         return new HashSet<>();
     }
 
+    @Override
+    public boolean hasLoadBykeys() {
+        return false;
+    }
+
     @Nonnull
     @Override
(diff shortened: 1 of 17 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
```diff
@@ -117,6 +117,11 @@ public KeysSource getKeysSource() {
         return new HashSet<>();
     }
 
+    @Override
+    public boolean hasLoadBykeys() {
+        return false;
```
nschiefer: I think this should be `true`.

MMcM: Yes, indeed. Fixed.

PR #833, opened 2020-02-21
Resolves #828: RankedSet poorly split for non-random keys 

### LINKED ISSUE #828: RankedSet poorly split for non-random keys
The skip-list aims to have 1/16 population at each coarser level. It does this by bit masking with a hash of the key ("score") byte-array. But the hash is just `Arrays.hashCode(byte[])`, which isn't very good.

A dataset whose keys are tuples ending with a couple integer zeros is showing 0.36% split at level 1 instead of 6.25%. This means that level 0 scans are huge.

MMcM:
Introduces an index option for `RANK` and `TIME_WINDOW_LEADERBOARD` indexes that controls what hash function is used.
Note that while it is technically safe to change the hash function for existing data, this is not recommended, as it changes the distribution.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -45,7 +45,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** StoreTimer.getDifference() should not return unchanged metrics [(Issue #832)](https://github.com/FoundationDB/fdb-record-layer/issues/832)
-* **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** Allow setting hash function used by `RankedSet` [(Issue #828)](https://github.com/FoundationDB/fdb-record-layer/issues/828)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
@@ -20,8 +20,6 @@
 
 package com.apple.foundationdb.async;
 
-import com.apple.foundationdb.Transaction;
-import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.KeySelector;
 import com.apple.foundationdb.KeyValue;
 import com.apple.foundationdb.MutationType;
@@ -29,7 +27,9 @@ import com.apple.foundationdb.Range;
 import com.apple.foundationdb.ReadTransaction;
 import com.apple.foundationdb.ReadTransactionContext;
 import com.apple.foundationdb.StreamingMode;
+import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.TransactionContext;
+import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.ByteArrayUtil;
 import com.apple.foundationdb.tuple.ByteArrayUtil2;
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.zip.CRC32;
 
 import static com.apple.foundationdb.async.AsyncUtil.DONE;
 import static com.apple.foundationdb.async.AsyncUtil.READY_FALSE;
@@ -75,6 +76,34 @@ import static com.apple.foundationdb.async.AsyncUtil.READY_FALSE;
  */
 @API(API.Status.MAINTAINED)
 public class RankedSet {
+    /**
+     * Hash using the JDK's default byte array hash.
+     *
+     * This hash does not have great distribution for certain values with low entropy.
+     */
+    public static final HashFunction JDK_ARRAY_HASH = Arrays::hashCode;
+
+    /**
+     * Hash using 32-bit CRC.
+     *
+     * Although not designed for hashing, this does often give better distribution than {@link #JDK_ARRAY_HASH}.
+     */
+    public static final HashFunction CRC_HASH = new HashFunction() {
+        @Override
+        public int hash(byte[] key) {
+            final CRC32 crc = new CRC32();
+            crc.update(key);
+            return (int)crc.getValue();
+        }
+    };
+
+    /**
+     * The default hash function to use.
+     *
+     * For reasons of compatibility with existing data, this defaults to the JDK's array hash.
+     */
+    public static final HashFunction DEFAULT_HASH_FUNCTION = JDK_ARRAY_HASH;
+
     private static final int LEVEL_FAN_POW = 4;
     private static final int[] LEVEL_FAN_VALUES; // 2^(l * FAN) - 1 per level
     public static final int MAX_LEVELS = Integer.SIZE / LEVEL_FAN_POW;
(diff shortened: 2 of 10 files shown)


docs/ReleaseNotes.md
```diff
@@ -45,7 +45,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** Improvement Allow setting hash function used by `RankedSet` [(Issue #828)](https://github.com/FoundationDB/fdb-record-layer/issues/828)
```
alecgrieser:
```suggestion
* **Performance** Allow setting hash function used by `RankedSet` [(Issue #828)](https://github.com/FoundationDB/fdb-record-layer/issues/828)
```

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -143,10 +190,7 @@ public RankedSet(Subspace subspace, Executor executor) {
      */
     public CompletableFuture<Boolean> add(TransactionContext tc, byte[] key) {
         checkKey(key);
-        // Use the hash of the key, instead a p value and randomLevel. The key is likely Tuple-encoded.
```
alecgrieser: While the other lines in this comment are needed any more, I think this first line is still valuable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainerFactory.java
```diff
@@ -64,15 +65,25 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
                                 LogMessageKeys.INDEX_NAME, index.getName());
                     }
                     changedOptions.remove(IndexOptions.RANK_NLEVELS);
                 }
+                if (changedOptions.contains(IndexOptions.RANK_HASH_FUNCTION)) {
+                    RankedSet.HashFunction oldFunction = RankedSetIndexHelper.getHashFunction(oldIndex);
+                    RankedSet.HashFunction newFunction = RankedSetIndexHelper.getHashFunction(index);
+                    if (!oldFunction.equals(newFunction)) {
+                        throw new MetaDataException("rank hash function changed",
+                                LogMessageKeys.INDEX_NAME, index.getName());
```
alecgrieser: Interesting. I suppose we could loosen this to allow for the rank hash function to change (given that it should be compatible). I'm not opposed to being stricter for now, though perhaps that should be an option at some point.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
```diff
@@ -50,6 +54,54 @@
...
+    }
+
+    public static int getNLevels(@Nonnull Index index) {
+        String nlevelsOption = index.getOption(IndexOptions.RANK_NLEVELS);
+        return nlevelsOption == null ? RankedSet.DEFAULT_LEVELS : Integer.parseInt(nlevelsOption);
+    }
+
+    public static RankedSet.HashFunction getHashFunction(@Nonnull Index index) {
+        String hashFunctionOption = index.getOption(IndexOptions.RANK_HASH_FUNCTION);
+        return hashFunctionOption == null ? RankedSet.DEFAULT_HASH_FUNCTION : HashFunctionNames.valueOf(hashFunctionOption).getHashFunction();
```
alecgrieser: I suppose using a registry rather than an enum is more idiomatic (e.g., `FunctionKeyExpression`s) and allows a user to provide their own implementation. I think that switching to a registry in the future is compatible with this current approach, though, so maybe it's fine as is for now. (Perhaps there should be an Issue about it.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
```diff
@@ -50,6 +54,54 @@
 public class RankedSetIndexHelper {
     public static final Tuple COMPARISON_SKIPPED_SCORE = Tuple.from(Comparisons.COMPARISON_SKIPPED_BINDING);
 
+    public static final RankedSet.HashFunction MURMUR3_HASH_FUNCTION = new GuavaHashFunction(Hashing.murmur3_32());
+
+    /**
+     * Use {@code com.google.common.hash.HashFunction}s as {@code RankedSet.HashFunction}s.
+     */
+    public static class GuavaHashFunction implements RankedSet.HashFunction {
```
alecgrieser: This kind of feels like a private class, at least until the user can provide their own hash function implementation (instead of using the enum).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
```diff
@@ -50,6 +54,54 @@
...
+        @Override
+        public int hash(byte[] key) {
+            return guava.hashBytes(key).asInt();
+        }
+    }
+
+    /**
+     * Known hash functions available as index options.
+     */
+    public enum HashFunctionNames {
```
alecgrieser: As this enum is in an `INTERNAL` class, it's technically "unadvised" for users to use this, but the alternative is they need their own constants in their code if they want to use the names here. I suppose we could leave it here if we're not "entirely happy" with this API (yet?), or move it out.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainerFactory.java
MMcM: Yes. I thought being conservative during the transition was best.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
MMcM: Yes. It's also possible that some additional options will require more state.

For example, if we wanted one that seeded Murmur3 from, say, the leaderboard time window id instead of always using 0.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
MMcM: I moved it out into the replacement for the enum and made it private

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
MMcM: I made this top level. Since that meant it should own the other statics and enum init order is fixed, I needed to also make it a regular class.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
MMcM: I didn’t make it extensible, but as a top level regular class it is now maybe a bit more registry like

alecgrieser: Yeah, and even though it's not extensible now, it looks like the kind of thing that could be made a registry if/when needed.

docs/ReleaseNotes.md
alecgrieser: I think the force push erased this change from the PR

MMcM: Fixed

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: Hmm, okay. Restored.

PR #837, opened 2020-02-27
Resolves #806: Rank index could have option for how to handle ties

### LINKED ISSUE #806: Rank index could have option for how to handle ties
Right now, you get ranks of `0, 0, 0, 1`, but one might want `0, 0, 0, 3`.

The underlying `RankedSet` would need the option to store counts greater than one at the leaves, then.

MMcM:
Note that this doesn't actually introduce a `RankedMultiset` class, because that was hard with the current `InstrumentedRankset` class. It would be possible by creating an interface and making that a wrapper instead of a subclass. But I'm not sure it's worth it.

This moves all the `Rankset` config options, including the recently added hash function, into a separate `Config` class so that the constructors don't continue to get out of hand. At some point (but not in this PR) some of them could presumably be deprecated.

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 import java.util.zip.CRC32;
 
@@ -89,14 +90,18 @@ public class RankedSet {
      *
      * Although not designed for hashing, this does often give better distribution than {@link #JDK_ARRAY_HASH}.
      */
-    public static final HashFunction CRC_HASH = new HashFunction() {
-        @Override
-        public int hash(byte[] key) {
-            final CRC32 crc = new CRC32();
-            crc.update(key);
-            return (int)crc.getValue();
-        }
+    public static final HashFunction CRC_HASH = key -> {
+        final CRC32 crc = new CRC32();
+        crc.update(key);
+        return (int)crc.getValue();
     };
+
+    /**
+     * Hash using a random number.
+     *
+     * The result is actually independent of the key.
+     */
+    public static final HashFunction RANDOM_HASH = kignore -> ThreadLocalRandom.current().nextInt();
 
     /**
      * The default hash function to use.
@@ -281,14 +286,38 @@ public class RankedSet {
         this.config = config;
     }
 
+    /**
+     * Initialize a new ranked set.
+     *
+     * Although not (yet) deprecated, this constructor is mainly for compatibility and {@link #RankedSet(Subspace, Executor, Config)}
+     * may be superior in new code.
+     * @param subspace the subspace where the ranked set is stored
+     * @param executor an executor to use when running asynchronous tasks
+     * @param hashFunction hash function for new ranked set
+     * @param nlevels number of skip list levels for new ranked set
+     */
     public RankedSet(Subspace subspace, Executor executor, HashFunction hashFunction, int nlevels) {
         this(subspace, executor, newConfigBuilder().setHashFunction(hashFunction).setNLevels(nlevels).build());
     }
 
+    /**
+     * Initialize a new ranked set.
+     *
+     * Although not (yet) deprecated, this constructor is mainly for compatibility and {@link #RankedSet(Subspace, Executor, Config)}
+     * may be superior in new code.
+     * @param subspace the subspace where the ranked set is stored
+     * @param executor an executor to use when running asynchronous tasks
+     * @param nlevels number of skip list levels for new ranked set
+     */
     public RankedSet(Subspace subspace, Executor executor, int nlevels) {
         this(subspace, executor, newConfigBuilder().setNLevels(nlevels).build());
     }
 
+    /**
+     * Initialize a new ranked set with the default configuration.
+     * @param subspace the subspace where the ranked set is stored
+     * @param executor an executor to use when running asynchronous tasks
+     */
     public RankedSet(Subspace subspace, Executor executor) {
         this(subspace, executor, DEFAULT_CONFIG);
     }
@@ -323,7 +352,7 @@ public class RankedSet {
         return tc.runAsync(tr ->
             countCheckedKey(tr, key)
                 .thenCompose(count -> {
-                    final boolean duplicate = count != null && count > 0;
+                    final boolean duplicate = count != null && count > 0;   // Is this key already present in the set?
                     if (duplicate && !config.isCountDuplicates()) {
                         return READY_FALSE;
                     }
@@ -335,6 +364,9 @@ public class RankedSet {
                         if (level == 0) {
                             future = addLevelZeroKey(tr, key, level, duplicate);
                         } else if (duplicate || (keyHash & LEVEL_FAN_VALUES[level]) != 0) {
+                            // If key is already present (duplicate), then whatever splitting it causes should have
+                            // already been done when first added. So, no more now. It is therefore possible, though,
+                            // that the count to increment matches the key, rather than being one that precedes it.
                             future = addIncrementLevelKey(tr, key, level, duplicate);
                         } else {
                             // Insert into this level by looking at the count of the previous key in the level
@@ -407,8 +439,10 @@ public class RankedSet {
                                 final CompletableFuture<Void> future;
 
                                 if (duplicate) {
-                                    // Always subtract one, never clearing a key. It is possible for this to leave
-                                    // a key with a count of zero, if duplicates were inserted with different hash functions.
+                                    // Always subtract one, never clearing a level count key.
+                                    // Concurrent requests both subtracting one when the count is two will conflict
+                                    // on the level zero key, which was read to detect that this is a duplicate.
+                                    // So it should not be possible for a count to go to zero.
                                     Function<byte[], Void> decrement = k -> {
                                         tr.mutate(MutationType.ADD, subspace.pack(Tuple.from(level, k)), encodeLong(-1));
                                         return null;
@@ -424,6 +458,7 @@ public class RankedSet {
                                     // That would require that the hash function never changes, though.
                                     // This allows for it to change, with the distribution perhaps getting a little uneven
                                     // as a result. It even allows for the hash function to return a random number.
+                                    // It also further guarantees that counts never go to zero.
                                     final byte[] k = subspace.pack(Tuple.from(level, key));
                                     final CompletableFuture<byte[]> cf = tr.get(k);
 
@@ -680,7 +715,8 @@ public class RankedSet {
                         // If the key need not be present and we are on the finest level, then if it wasn't an exact
                         // match, key would have the next rank after the last one. Except in the case where key is less
                         // than the lowest key in the set, in which case it takes rank 0. This is recognizable because
-                        // at level 0, only the leftmost empty array has a count of zero; every other key has a count of one.
+                        // at level 0, only the leftmost empty array has a count of zero; every other key has a count of one
+                        // (or the number of duplicates if those are counted separately).
                         rank++;
                     }
                     return true;
@@ -861,7 +897,6 @@ public class RankedSet {
     }
 
     protected static class Consistency {
-
 
         private final boolean consistent;
         private final int level;
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
@@ -108,11 +108,11 @@ public class RankedSet {
     private static final int[] LEVEL_FAN_VALUES; // 2^(l * FAN) - 1 per level
     public static final int MAX_LEVELS = Integer.SIZE / LEVEL_FAN_POW;
     public static final int DEFAULT_LEVELS = 6;
+    public static final Config DEFAULT_CONFIG = new Config();
 
     protected final Subspace subspace;
     protected final Executor executor;
-    protected final HashFunction hashFunction;
-    protected final int nlevels;
+    protected final Config config;
 
     static {
         LEVEL_FAN_VALUES = new int[MAX_LEVELS];
@@ -144,29 +144,124 @@ public class RankedSet {
     }
 
     /**
+     * Configuration settings for a {@link RankedSet}.
+     */
+    public static class Config {
+        private final HashFunction hashFunction;
+        private final int nlevels;
+
+        protected Config() {
+            this.hashFunction = DEFAULT_HASH_FUNCTION;
+            this.nlevels = DEFAULT_LEVELS;
+        }
+
+        protected Config(HashFunction hashFunction, int nlevels) {
+            this.hashFunction = hashFunction;
+            this.nlevels = nlevels;
+        }
+
+        /**
+         * Get the hash function to use.
+         * @return a {@link HashFunction} used to convert keys to a bit mask used to determine level splits in the skip list
+         */
+        public HashFunction getHashFunction() {
+            return hashFunction;
+        }
+
+        /**
+         * Get the number of levels to use.
+         * @return the number of levels in the skip list
+         */
+        public int getNLevels() {
+            return nlevels;
+        }
+    }
+
+
+    /**
+     * Builder for {@link Config}.
+     *
+     * @see #newConfigBuilder
+     */
+    public static class ConfigBuilder {
+        private HashFunction hashFunction = DEFAULT_HASH_FUNCTION;
+        private int nlevels = DEFAULT_LEVELS;
+
+        protected ConfigBuilder() {
+        }
+
+        public HashFunction getHashFunction() {
+            return hashFunction;
+        }
+
+        /**
+         * Set the hash function to use.
+         *
+         * It is possible to change the hash function of an existing ranked set, although this is not recommended since the distribution in the skip list may
+         * become uneven as a result.
+         * @param hashFunction the hash function to use
+         * @return this builder
+         */
+        public ConfigBuilder setHashFunction(HashFunction hashFunction) {
+            this.hashFunction = hashFunction;
+            return this;
+        }
+
+        public int getNLevels() {
+            return nlevels;
+        }
+
+        /**
+         * Set the hash function to use.
+         *
+         * It is not currently possible to change the number of levels for an existing ranked set.
+         * @param nlevels the number of levels to use
+         * @return this builder
+         */
+        public ConfigBuilder setNLevels(int nlevels) {
+            if (nlevels < 2 || nlevels > MAX_LEVELS) {
+                throw new IllegalArgumentException("levels must be between 2 and " + MAX_LEVELS);
+            }
+            this.nlevels = nlevels;
+            return this;
+        }
+
+        public Config build() {
+            return new Config(hashFunction, nlevels);
+        }
+    }
+
+    /**
+     * Start building a {@link Config}.
+     * @return a new {@code Config} that can be altered and then built for use with a {@link RankedSet}
+     * @see ConfigBuilder#build
+     */
+    public static ConfigBuilder newConfigBuilder() {
+        return new ConfigBuilder();
+    }
+
+    /**
      * Initialize a new ranked set.
      * @param subspace the subspace where the ranked set is stored
      * @param executor an executor to use when running asynchronous tasks
-     * @param hashFunction hash function to use to determine which levels a key splits on
-     * @param nlevels number of skip list levels to maintain
+     * @param config configuration to use
      */
-    public RankedSet(Subspace subspace, Executor executor, HashFunction hashFunction, int nlevels) {
-        if (nlevels < 2 || nlevels > MAX_LEVELS) {
-            throw new IllegalArgumentException("levels must be between 2 and " + MAX_LEVELS);
-        }
-
+    public RankedSet(Subspace subspace, Executor executor, Config config) {
         this.subspace = subspace;
         this.executor = executor;
-        this.hashFunction = hashFunction;
-        this.nlevels = nlevels;
+        this.config = config;
+    }
+
+    public RankedSet(Subspace subspace, Executor executor, HashFunction hashFunction, int nlevels) {
+        this(subspace, executor, newConfigBuilder().setHashFunction(hashFunction).setNLevels(nlevels).build());
     }
 
     public RankedSet(Subspace subspace, Executor executor, int nlevels) {
-        this(subspace, executor, DEFAULT_HASH_FUNCTION, nlevels);
+        this(subspace, executor, newConfigBuilder().setNLevels(nlevels).build());
     }
 
     public RankedSet(Subspace subspace, Executor executor) {
-        this(subspace, executor, DEFAULT_HASH_FUNCTION, DEFAULT_LEVELS);
+        this(subspace, executor, DEFAULT_CONFIG);
     }
 
     public CompletableFuture<Void> init(TransactionContext tc) {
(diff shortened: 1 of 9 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -246,6 +382,80 @@ public RankedSet(Subspace subspace, Executor executor) {
         });
     }
 
+    /**
+     * Removes a key from the set.
+     * @param tc the transaction to use to access the database
+     * @param key the key to remove
+     * @return a future that completes to {@code true} if the set was modified, that is, if the key was present before this operation
+     */
+    public CompletableFuture<Boolean> remove(TransactionContext tc, byte[] key) {
```
MMcM: There is a separate commit to relocate this method, which might make it easier to track how it has changed (other than that).

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -267,17 +477,30 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+
+    /**
+     * Count the number of occurrences of a key in the set.
+     * @param tc the transaction to use to access the database
+     * @param key the key to check for
+     * @return a future that completes to {@code 0} if the key is not present in the ranked set or
+     * {@code 1} if the key is present in the ranked set and duplicates are not counted or
+     * the number of occurrences if duplicated are counted separately
+     */
+    public CompletableFuture<Long> count(ReadTransactionContext tc, byte[] key) {
```
MMcM: If there were a separate `RankedMultiset` class, this would only be on it, though it seems well-defined everywhere.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -643,19 +819,21 @@ private static void checkKey(byte[] key) {
...
-                        .getRange(KeySelector.lastLessThan(k), KeySelector.firstGreaterOrEqual(k), 1)
-                        .asList()
-                        .thenApply(kvs -> {
-                            byte[] prevk = kvs.get(0).getKey();
-                            // If another key were inserted after between this and the target key,
-                            // it wouldn't be the one we should increment any more.
-                            // But do not conflict when key itself is incremented.
-                            byte[] exclusiveBegin = ByteArrayUtil.join(prevk, ZERO_ARRAY);
-                            tr.addReadConflictRange(exclusiveBegin, k);
+                tr.snapshot().getKey(orEqual ? KeySelector.lastLessOrEqual(k) : KeySelector.lastLessThan(k))
```
MMcM: I don't _think_ there was any reason to do this with a `getRange` before. I'm also not sure the `snapshot` is needed any more.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -246,6 +382,80 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+                                                tr.clear(k);
+                                            }
+                                            return null;
+                                        });
+                                    } else {
+                                        final CompletableFuture<byte[]> prevKeyF = getPreviousKey(tr, level, key, false);
+                                        future = CompletableFuture.allOf(cf, prevKeyF)
+                                                .thenApply(vignore -> {
+                                                    final byte[] c = cf.join();
+                                                    long countChange = -1;
```
MMcM: Not much thought went into mixing set and multiset on the same subspace. Other than not violating the consistency invariants. In particular, `remove` always removes (up to) one occurrence. Would it be better if it removed _all_ when not counting duplicates?

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -246,6 +382,80 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+                                        decrement.apply(key);
+                                        future = DONE;
+                                    } else {
+                                        future = getPreviousKey(tr, level, key, true).thenApply(decrement);
+                                    }
+                                } else {
+                                    // This could be optimized to check the hash for which levels should have this key.
+                                    // That would require that the hash function never changes, though.
+                                    // This allows for it to change, with the distribution perhaps getting a little uneven
+                                    // as a result. It even allows for the hash function to return a random number.
```
MMcM: Is it worth having a `RANDOM_HASH_FUNCTION`? Roughly, `threadLocalRandom.getInt()`. And a test to show that it works?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
```diff
@@ -233,38 +244,41 @@ private static Number extractRank(int groupPrefixSize, @Nullable Tuple maybeRank
...
             }
         });
         return state.store.instrument(Events.RANKED_SET_UPDATE, result);
     }
 
+    private static CompletableFuture<Void> removeFromRankedSet(@Nonnull IndexMaintainerState state, @Nonnull RankedSet rankedSet, @Nonnull byte[] score) {
+        return rankedSet.remove(state.transaction, score).thenApply(exists -> {
+            // It is okay if the score isn't in the ranked set yet if the index is
+            // write only because this means that the score just hasn't yet
+            // been added by some record. We still want the conflict ranges, though.
```
alecgrieser: I, um, am not sure this comment is true, or at least doesn't seem right given one of the comments I made in the `remove` function for `RankedSet`. On the other hand, we do test write only indexes as part of index builds, so I'm not entirely sure. (Understanding that this is a pre-existing comment that I probably wrote.)

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: The previous code,
```
.getRange(KeySelector.lastLessThan(k), KeySelector.firstGreaterOrEqual(k), 1)
```
didn't really have that protection, since it used two key selectors from the same input.

Would it be better then (that is, protecting against some screwup with the `('',0)` entry) to do a reverse `getRange` with the left the level limit and the right the target?

alecgrieser: Yeah, I think a range read like that would make sense, and it would protect us from exactly that kind of screw up.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -643,19 +819,21 @@ private static void checkKey(byte[] key) {
...
+                        .thenApply(prevk -> {
+                            if (!orEqual || !Arrays.equals(prevk, k)) {
+                                // If another key were inserted after between this and the target key,
+                                // it wouldn't be the one we should increment any more.
+                                // But do not conflict when key itself is incremented.
+                                byte[] exclusiveBegin = ByteArrayUtil.join(prevk, ZERO_ARRAY);
+                                tr.addReadConflictRange(exclusiveBegin, k);
+                            }
                             // Do conflict if key is removed entirely.
                             tr.addReadConflictKey(subspace.pack(Tuple.from(0, subspace.unpack(prevk).getBytes(1))));
```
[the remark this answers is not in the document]
MMcM: I believe that in the case where this is a duplicate and matches the key itself, this adds another read conflict for the one that was already added by `countCheckedKey` to determine that it was a duplicate. So I'm not sure what the case is where this is a _totally_ superfluous read conflict.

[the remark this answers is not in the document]
alecgrieser: I was thinking something like: you have two entries, `a` and `b` in level 0 and only `a` in level 1. Then if you increment `b`, that will induce a conflict range on `b` in level 0 (from `countCheckedKey`) and the range `(a, b]` (I think) in level 1 and then on just `a` in level 0. But a concurrent increment of `a` would change the value of `a` in level 0, so the increment on `b` could be failed as a result, but if this conflict range is removed, I think the conflict induced by the increment of `a` that fails the increment of `b` wouldn't happen. I think.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: I switched to not using `KeySelector`s and doing a size 1 reverse range scan.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: I fixed the comments. It should not, in fact, be possible to decrement to zero any more.

Which means that if we did decide that `remove` should mean remove all, it would be consistent in clearing any splits induced by the given key.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: I added the basic test for this.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -143,30 +144,153 @@ private static long decodeLong(byte[] v) {
...
-
+    public RankedSet(Subspace subspace, Executor executor, Config config) {
         this.subspace = subspace;
         this.executor = executor;
-        this.hashFunction = hashFunction;
-        this.nlevels = nlevels;
+        this.config = config;
+    }
+
+    public RankedSet(Subspace subspace, Executor executor, HashFunction hashFunction, int nlevels) {
```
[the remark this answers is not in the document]
MMcM: Added such comments.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -179,33 +303,39 @@ public RankedSet(Subspace subspace, Executor executor) {
...
                     for (int li = 0; li < nlevels; ++li) {
-                        int level = li;
+                        final int level = li;
                         CompletableFuture<Void> future;
                         if (level == 0) {
-                            future = addLevelZeroKey(tr, key, level);
-                        } else if ((keyHash & LEVEL_FAN_VALUES[level]) != 0) {
-                            future = addIncrementLevelKey(tr, key, level);
+                            future = addLevelZeroKey(tr, key, level, duplicate);
+                        } else if (duplicate || (keyHash & LEVEL_FAN_VALUES[level]) != 0) {
```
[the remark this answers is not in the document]
MMcM: I amplified these comments.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -143,30 +144,153 @@ private static long decodeLong(byte[] v) {
         int hash(byte[] key);
     }
 
+    /**
+     * Configuration settings for a {@link RankedSet}.
```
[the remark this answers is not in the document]
MMcM: It is enforced at the index level, which has old and new meta-data that can be compared.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainerFactory.java
```diff
@@ -65,24 +65,24 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
             @Override
             public void validateChangedOptions(@Nonnull Index oldIndex, @Nonnull Set<String> changedOptions) {
-                // Allow changing from unspecified to the default (or vice versa), but not otherwise.
-                if (changedOptions.contains(IndexOptions.RANK_NLEVELS)) {
-                    int oldLevels = RankedSetIndexHelper.getNLevels(oldIndex);
-                    int newLevels = RankedSetIndexHelper.getNLevels(index);
-                    if (oldLevels != newLevels) {
-                        throw new MetaDataException("rank levels changed",
-                                LogMessageKeys.INDEX_NAME, index.getName());
+                if (!changedOptions.isEmpty()) {
```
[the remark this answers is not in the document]
MMcM: I disallowed changing duplicates, in keeping with the conservative nature of the existing checks.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/TimeWindowLeaderboardIndexMaintainer.java
```diff
@@ -197,8 +197,9 @@ protected void saveSubDirectory(@Nonnull TimeWindowLeaderboardSubDirectory subdi
                 }
                 final Subspace extraSubspace = getSecondarySubspace();
                 final Subspace leaderboardSubspace = extraSubspace.subspace(leaderboard.getSubspaceKey());
+                final RankedSet.Config config = configBuilder.setNLevels(leaderboard.getNLevels()).build();
```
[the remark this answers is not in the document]
MMcM: I switched from a shared builder to `toBuilder`.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -356,37 +396,59 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+                                    if (level == 0) {
+                                        future = cf.thenApply(c -> {
+                                            if (c != null) {
+                                                tr.clear(k);
+                                            }
+                                            return null;
+                                        });
+                                    } else {
+                                        final CompletableFuture<byte[]> prevKeyF = getPreviousKey(tr, level, key, false);
+                                        future = CompletableFuture.allOf(cf, prevKeyF)
```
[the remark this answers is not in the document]
MMcM: Done

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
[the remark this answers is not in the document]
alecgrieser: Yeah, and that's probably as good as we're gonna get.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainerFactory.java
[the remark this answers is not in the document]
alecgrieser: That sounds right. I'd be tempted to say that even the most liberal of changes shouldn't allow that particular option to change ever (because the data won't be correct if you switch, even if it's understandable).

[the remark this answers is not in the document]
MMcM: Hmm. There are, I think, two points of view:
* The option defines a fundamental characteristic of the skip list, viz., whether the finest level has counts greater than one.
* The option refines the behavior of the `add` method to either maintain counts or return `false` for things already present.

Which of those makes more sense also probably informs whether `remove` always removes one occurrence or is similarly influenced by the config option.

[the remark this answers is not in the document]
alecgrieser: Interesting. I had kind of been coming from the first point of view. It kind of seems like the second point of view makes sense from the point of view of the skip list itself, though it's harder to see how that is a sensible "index" over data. (In that with duplicate tracking off, it is just a data structure for optimizing "rank without counting duplicates" queries; with duplicate tracking on, it is a data structure for optimizing "rank with duplicates counted" queries; but with it changed half-way through, then I don't know if it answers any reasonable queries other than "what is the state of skip list".)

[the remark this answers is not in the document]
MMcM: Yes, agreed. I don't have a great (not implementation based) idea of why a user of the index would want to change the answers to the rank / select operations it optimizes.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
alecgrieser: Maybe I'm missing the change, but it looks like one could still decrement a key to zero (in duplicate counting mode), in that it just decrements the value by 1.

MMcM: In `add`, if a key is already present (based on checking level 0 != 0), then the hash is ignored and each key that was incremented last time is incremented again. In other words, a duplicate value is counted n times at every level reliably.

In `remove`, if a key is not the last occurrence (based on checking level 0 > 1), then it is safe to simply decrement every level because this key must have contributed at least 2 to each of them. If a key is the last occurrence, then if the key decremented at any level is matches exactly (as opposed to being somewhere to the left), it is cleared. This is the only key that might otherwise have been decremented to zero, I believe.

Simultaneous `remove`s of the same key conflict on level 0.

alecgrieser: Oh, I think I see what I was missing: I had though that the `duplicate` boolean was determined by the configuration rather than by the count that actually gets read. I'll have to rethink what this is doing with that in mind.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -356,37 +396,59 @@ public RankedSet(Subspace subspace, Executor executor) {
...
-                                if (level == 0) {
-                                    future = cf.thenApply(c -> {
-                                        if (c != null) {
-                                            tr.clear(k);
-                                        }
+                                if (duplicate) {
+                                    // Always subtract one, never clearing a key. It is possible for this to leave
+                                    // a key with a count of zero, if duplicates were inserted with different hash functions.
+                                    Function<byte[], Void> decrement = k -> {
+                                        tr.mutate(MutationType.ADD, subspace.pack(Tuple.from(level, k)), encodeLong(-1));
```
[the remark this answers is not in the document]
alecgrieser: I see now that as we read the count and, if the count is 1, just remove the key, that this is no longer something to be worried about, though I suppose one could wist hopefully for a data structure with fewer conflicts (e.g., no conflicts when adding or removing the same key). It....might be possible, but it probably requires additional fanciness like conflicts in same "shadow realm", and it can be done later if we decide we really want it.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -220,32 +391,110 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+                                    final byte[] k = subspace.pack(Tuple.from(level, key));
+                                    final CompletableFuture<byte[]> cf = tr.get(k);
+
+                                    if (level == 0) {
+                                        future = cf.thenApply(c -> {
+                                            if (c != null) {
+                                                tr.clear(k);
+                                            }
+                                            return null;
+                                        });
```
alecgrieser: It seems weird (though perhaps) harmless that this re-reads the key for level zero rather than, say, just clearing the key. For one, it almost gives the impression that the key could be null at this point...which seems wrong.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
```diff
@@ -220,32 +391,110 @@ public RankedSet(Subspace subspace, Executor executor) {
...
+                                    // Always subtract one, never clearing a level count key.
+                                    // Concurrent requests both subtracting one when the count is two will conflict
+                                    // on the level zero key, which was read to detect that this is a duplicate.
+                                    // So it should not be possible for a count to go to zero.
+                                    Function<byte[], Void> decrement = k -> {
+                                        tr.mutate(MutationType.ADD, subspace.pack(Tuple.from(level, k)), encodeLong(-1));
+                                        return null;
+                                    };
+                                    if (level == 0) {
+                                        decrement.apply(key);
```
alecgrieser: I suppose rather than decrementing, this could explicitly set the key to `encodeLong(count - 1)`. I think the only thing this would do is underscore that we *aren't* actually making use of the ability of atomic operations to allow for concurrent updates to the level 0 keys.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
alecgrieser: Yeah, this now makes sense to me, and I believe you're right--that essentially because all updates contend on their level 0 key, some of the nasty edge cases that can lead to concurrent decrements resulting in there being a count set to zero. I suppose we may need to re-architect if we decide that that's "too much contention" or something.

alecgrieser: As to the original question, I *think* that having remove always be "remove 1" is fine, and now here my reasoning is: (1) having it that way decreases the number of branches (on the config value) from the code and (2) changing the value of the config parameter is weird enough that we should generally avoid doing it (hence banning changing it at the meta-data level) and shouldn't add complexity to the code to support it.

But my feelings aren't particularly strong.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
alecgrieser: I somehow missed the part of the code in `RankedSet` that checks to make sure that the score isn't zero before decrementing. I am now more confident in this comment in at least the case of not duplicate tracking (so that's good).

I'm trying to think, however, if there are pathologies in the count duplicates case. I think the answer is "yes" because adding to a ranked set index with duplicate counting isn't idempotent. In particular:

1. Imagine having two records with a rank index on field `foo`, both of which have the value `foo = bar`.
1. An index build indexes the first record. Now the count corresponding to `bar` in the `foo` rank index is 1.
1. Someone removes the second record. Now the count corresponding to `bar` in the `foo` rank index is 0 (i.e., it is removed from the ranked set).
1. The index build completes, never having indexed the second record.

So now `bar` has an incorrect rank. This is somewhat of a worst case in that `bar` is missing entirely from the ranked set, but I think one can also be in a position where the rank is wrong, but not removed entirely.

I think this can be fixed by checking if the record is in the built range and only updating the ranked set if it is. It also seems to me like this would be safe for the regular ranked set, though perhaps would induce more conflicts.

MMcM: That would be accomplished by having `isIdempotent` check the option, correct? There is no need for a new instance of such built range checking code.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: Yes, it is simpler to just clear unconditionally.

fdb-extensions/src/main/java/com/apple/foundationdb/async/RankedSet.java
MMcM: Yes, that does seem more obvious and in line with the preceding comment.

fdb-extensions/src/test/java/com/apple/foundationdb/async/RankedSetTest.java
```diff
@@ -118,6 +119,49 @@ private void basicOperations(RankedSet.HashFunction firstHashFunction, RankedSet
                 boolean wasOld = rs.remove(tr, k).join();
                 assertTrue(wasOld);
             }
+            assertFalse(rs.remove(tr, keys[20]).join());
+            return null;
+        });
+    }
+
+    @Test
+    public void duplicates() {
```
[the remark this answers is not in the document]
MMcM: I  don't much care for those tests, either. But I added one with duplicates.

Note that the test already has knowledge of some invariants which are changed in the duplicates case, so some of the checks need to be conditional on the config.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetIndexHelper.java
alecgrieser: Yeah, I believe so--that no new mechanism is necessary because what I sketched out is already done by `isIdempotent`.

PR #842, opened 2020-03-04
Resolves #841: Add a custom index queryability filter for planning.

### LINKED ISSUE #841: RecordQueryPlanner should have a mechanism for changing index queryability on per-query basis
Currently, we have two mechanisms for specifying which indexes the `RecordQueryPlanner` should consider when planning:
- Allowed indexes, which is an exhaustive list of the indexes that the planner should consider regardless of what queryability is specified in the index options.
- The `IndexOptions.ALLOWED_FOR_QUERY_OPTION`. When this is `false` on an index, the planner does not consider that index unless the index is specified as an allowed index.

However, we don't have a way to specify that an index with `IndexOptions.ALLOWED_FOR_QUERY_OPTION` set to `false` can be used, but needn't necessarily be used, except for specifying a lot of indexes as "allowed indexes". This is clunky and abstraction breaking.

Instead, we should have a mechanism to specify that an index should be considered "just for this query" without requiring its usage or listing all indexes.

nschiefer:
Add a new field to `RecordQuery` that allows a user to specify a custom filtering function that determines whether an index should be considered by the planner. This replaces the previous hard-coded check for the value of `IndexOptions.ALLOWED_FOR_QUERY_OPTION`, which remains the default.

This change does not break backwards compatibility.

See #841 for more discussion of why we need this filter.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/IndexQueryabilityFilter.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/IndexQueryabilityFilter.java
@@ -1,0 +1,51 @@
+/*
+ * IndexQueryabilityFilter.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2020 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.metadata.Index;
+import com.apple.foundationdb.record.metadata.IndexOptions;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A filter used to determine whether an index should be considered when planning queries.
+ */
+@API(API.Status.EXPERIMENTAL)
+@FunctionalInterface
+public interface IndexQueryabilityFilter {
+    /**
+     * The default index queryability filter which uses all indexes except those with the
+     * {@link IndexOptions#ALLOWED_FOR_QUERY_OPTION} set to {@code false}.
+     */
+    IndexQueryabilityFilter DEFAULT =
+            index -> index.getBooleanOption(IndexOptions.ALLOWED_FOR_QUERY_OPTION, true);
+
+    /**
+     * Return whether the given index should be considered by the query planner. Note that the planner is not required
+     * to use an index for which {@code isQueryable()} is {@code true}.
+     * @param index an index
+     * @return whether the given index should be considered by the planner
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRestrictedIndexQueryTest.java
```diff
@@ -462,4 +462,53 @@ public void queryAllowedUniversalIndex() throws Exception {
         assertFalse(plan2.hasRecordScan(), "should not use record scan");
         assertEquals(-1692774119, plan2.planHash());
     }
+
+    /**
+     * Verify that the query planner uses the specified {@link com.apple.foundationdb.record.query.IndexQueryabilityFilter}.
+     * If both allowed indexes and a queryability filter are set, verify that the planner uses the allowed indexes.
+     */
+    @Test
```
alecgrieser: I assume that this doesn't work with the new planner (so can't be a `@DualPlannerTest`)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/RecordQuery.java
```diff
@@ -228,6 +242,17 @@ public Builder setAllowedIndex(@Nullable String allowedIndex) {
             return setAllowedIndexes(Collections.singleton(allowedIndex));
         }
 
+        /**
+         * Set a function that defines whether each index should be used by the query planner.
+         * Note that if allowed indexes are used then the queryability filter is ignored.
```
alecgrieser: In theory, the docs for `.setAllowedIndexes` should also explain that they override the queryability filter. This also seems like it should link to the `DEFAULT` filter for completeness.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRestrictedIndexQueryTest.java
nschiefer: I somehow thought so, but a one-line change in the new planner fixes the issue. This is now a `DualPlannerTest`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/RecordQuery.java
nschiefer: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/IndexQueryabilityFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * A filter used to determine whether an index should be considered when planning queries.
+ */
+@API(API.Status.EXPERIMENTAL)
+@FunctionalInterface
+public interface IndexQueryabilityFilter {
+    /**
+     * The default index queryability filter which uses all indexes except those with the {@link IndexOptions#ALLOWED_FOR_QUERY_OPTION}.
```
alecgrieser: Oh, sorry, just realized that this should probably say "except those with the `ALLOWED_FOR_QUERY_OPTION` set to `false`"

alecgrieser:
Sorry about missing that thing in the comments the first time around (whoops)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/IndexQueryabilityFilter.java
nschiefer: Fixed!

PR #847, opened 2020-03-11
Fixes #844: AsyncLoadingCache should cancel work on deadline exceeded

### LINKED ISSUE #844: AsyncLoadingCache should cancel work on deadline exceeded
At the moment, if the `AsyncLoadingCache` starts doing work (including to, say, resolve a key space path element from the directory layer) and it hits a `DeadlineExceededException`, the work isn't cancelled in any way. This means that the work could still be ongoing, even after a caller of the `AsyncLoadingCache` has received an error. This could include doing work touching the database, which can then induce additional load on the system. We should clean up work from that operation.

alecgrieser:
Now, if the right overload is used, the `AsyncLoadingCache` can be configured to cancel work if it encounters an error while attempting to load the value. The `LocatableResolver` code has been modified to use this new code path.

This fixes #844.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -40,7 +40,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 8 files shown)


alecgrieser:
There's some discussion in the issue #844 about whether this is the approach we actually want to take, or if, for example, it would be better to just ditch the `AsyncLoadingCache` (at least for the `LocatableResolver`). It's possible we'll want to close this unmerged if that's the case.

alecgrieser:
I believe our official answer to this is that we don't want this, and that users should be setting the transaction timeout option instead to avoid having work take forever.

PR #836, opened 2020-02-26
Resolve #835: Expose bytes/records scanned through ExecuteState

scgray:
This includes some efforts to still allow for the use of the
`NO_LIMITS` constant in fashion that should have little overhead
on execution performance.

PR #849, opened 2020-03-11
Resolves #848: Allow for setting the FDB transaction timeout on FDBRecordContexts

### LINKED ISSUE #848: Allow for setting the FDB transaction timeout on FDBRecordContexts
At the moment, we can set the timeout option on FDB transactions wrapped in `FDBRecordContext`s by creating a transaction and then calling `.ensureActive().options().setTimeout`, but we probably want to allow users to set that in a more elegant way because:

1. Changing a timeout halfway through a transaction's life is a weird enough thing that we probably want to give users a different API that lets them set it on startup and not worry about it any more.
1. We should give users a way to ensure that the timeout is set when using a runner without needing to set it explicitly.

alecgrieser:
I, um, don't quite know how this blew up into ~700 lines added, but alas. The main goal of this change is to allow for the transaction timeout to be specified through the `FDBRecordContext`. In theory, one could always have called `ensureActive` and then set the option manually, but it seemed better to have a real way of doing this without needing to peek at internals.

This is broken into two commits:

1. The first refactors `FDBRecordContext` to take a `Config` object at creation. This is to ameliorate the fact that there was a growing number of essentially optional parameters into the object, so now those are hidden in an object. But that commit shouldn't make any functional changes *except* (because it simplified the implementation considerably), it is now the case that if you explicitly set a transaction ID to `null`, then it will still look in the MDC context for a `uuid` key. We could keep that behavior the same by, like, distinguishing between a "default" transaction ID and a "set to `null`" transaction ID, but I didn't.
1. The second adds to that config object a "timeout" parameter. The `FDBDatabaseFactory` and `FDBDatabaseRunner` classes also have timeout parameters, and transactions created by a runner will use the runner's timeout value if set, and all transactions will default to the database factory's value if unset. And, if *nothing* in these classes sets a timeout, it will use the value set at the FDB level, so this should be compatible with existing code that has used the various end-runs around the Record Layer to set these values already.

This resolves #848.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
@@ -358,8 +358,7 @@ public class FDBDatabase {
                                         @Nullable FDBStoreTimer timer,
                                         @Nullable WeakReadSemantics weakReadSemantics,
                                         @Nonnull FDBTransactionPriority priority) {
-        final String transactionId = mdcContext == null ? null : mdcContext.get("uuid");
-        return openContext(mdcContext, timer, weakReadSemantics, priority, transactionId);
+        return openContext(mdcContext, timer, weakReadSemantics, priority, null);
     }
 
 
@@ -367,11 +366,11 @@ public class FDBDatabase {
      * Open a new record context with a new transaction begun on the underlying FDB database.
      *
      * <p>
-     * Note that other variants of this method will inspect the MDC context for the transaction ID by looking
-     * for an entry in the map with the key "uuid". This method will ignore whatever is in the MDC context
-     * and use the ID provided as a parameter instead. The transaction ID should typically consist solely of
-     * printable ASCII characters and should not exceed 100 bytes. The ID may be truncated or dropped if the ID will
-     * not fit in 100 bytes. See {@link FDBRecordContext#getTransactionId()} for more details.
+     * If the passed {@code transactionId} is {@code null}, then this method will set the transaction ID
+     * of the given transaction to the value of the "uuid" key in the MDC context if present. The transaction ID
+     * should typically consist solely of printable ASCII characters and should not exceed 100 bytes. The ID may
+     * be truncated or dropped if the ID will not fit in 100 bytes. See {@link FDBRecordContext#getTransactionId()}
+     * for more details.
      * </p>
      *
      * @param mdcContext logger context to set in running threads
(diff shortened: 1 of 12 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -358,20 +358,19 @@ public FDBRecordContext openContext(@Nullable Map<String, String> mdcContext,
...
      * <p>
-     * Note that other variants of this method will inspect the MDC context for the transaction ID by looking
-     * for an entry in the map with the key "uuid". This method will ignore whatever is in the MDC context
-     * and use the ID provided as a parameter instead. The transaction ID should typically consist solely of
-     * printable ASCII characters and should not exceed 100 bytes. The ID may be truncated or dropped if the ID will
-     * not fit in 100 bytes. See {@link FDBRecordContext#getTransactionId()} for more details.
+     * If the passed {@code transactionId} is {@code null}, then this method will set the transaction ID
+     * of the given transaction to the value of the "uuid" key in the MDC context if present. The transaction ID
+     * should typically consist solely of printable ASCII characters and should not exceed 100 bytes. The ID may
+     * be truncated or dropped if the ID will not fit in 100 bytes. See {@link FDBRecordContext#getTransactionId()}
```
alecgrieser: This is alluding to that behavior change regarding the transaction ID that is mentioned in the PR's overview comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunner.java
```diff
@@ -201,6 +201,32 @@
...
+    /**
+     * Get the transaction timeout for all transactions started by this runner. This will return the value configured
+     * for this runner through {@link #setTransactionTimeoutMillis(long)}. Note, however, that if the transaction timeout
+     * is set to {@link FDBDatabaseFactory#DEFAULT_TR_TIMEOUT_MILLIS}, then the actual timeout set for this transaction
+     * will be set to the value in the originating factory.
+     *
+     * @return the configured transacation timeout time in milliseconds
+     * @see #setTransactionTimeoutMillis(long)
+     */
+    long getTransactionTimeoutMillis();
```
alecgrieser: These are "breaking changes" for implementors of the interface, if we care about that use case. They are fine for those that are using the interface.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
```diff
@@ -0,0 +1,308 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
+
+/**
+ * A configuration struct that can be used to set various options on an {@link FDBRecordContext}. Instances
+ * of this configuration object can be passed to {@link FDBDatabase#openContext(FDBRecordContextConfig)}
+ * to create a new transaction with various parameters set according to the values specified here.
+ */
+public class FDBRecordContextConfig {
```
scgray: (just a comment, nothing to do) I kind of have a preference for `FDBRecordContext.Config`, but given the size of this class I get why you did this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunner.java
scgray: Maybe the release notes should call out this minor (unlikely) breaking change (I did something like this on 2.8.103.0:

* **Breaking change** `ByteScanLimiter` and `RecordScanLimiter` are now interfaces. Instances with various concrete behavior are constructed through factory classes. [(PR #836)](https://github.com/FoundationDB/fdb-record-layer/pull/836)

alecgrieser: Okay, added release notes calling it out. For these two methods, I attached the issue (#848) as they seemed "inherent" to that Issue. I also call out the slight change in semantics regarding transaction IDs, but with this PR, as it seemed more "incidental".

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
alecgrieser: Yeah, that was my logic, too. Plus, as this would have a builder, it seemed that that would mean we either needed a `FDBRecordContext.Config.Builder` or a `FDBRecordContext.ConfigBuilder`, neither of which I was particularly excited about.

alecgrieser:
Rebase is just to get on top of master (no changes); extra commit contains more release notes.

PR #851, opened 2020-03-12
Resolves #850: New exception type for scanning non-readable indexes

### LINKED ISSUE #850: New exception type for scanning non-readable indexes
instead of checking strings

nblintao:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreIndexTest.java
```diff
@@ -843,8 +844,8 @@ public void scanWriteOnlyIndex() throws Exception {
             try {
                 recordStore.scanIndexRecords(indexName);
                 fail("Was not stopped from scanning write-only index");
-            } catch (RecordCoreException e) {
-                assertEquals("Cannot scan non-readable index " + indexName, e.getMessage());
+            } catch (ScanNonReadableIndexException e) {
```
shralex: consider catching any exception and asserting that its the right type of exception and then that the index name matches. Also below

nblintao: I think it will also fail if it is an uncaught exception (for say, RecordCoreException but not ScanNonReadableIndexException).

Issue #514, opened 2019-04-04
Index state should be keyed by subspace key rather than by name


alecgrieser:
Index data is stored under a subspace that looks like:

```
common_record_store_prefix + (2L, index.getSubspaceKey())
```

But the index state (i.e., whether the index is disabled or read only or whatever) is stored under a key that looks like:

```
common_record_store_prefix + (5L, index.getName())
```

I believe this is the *only* subspace related to indexes that uses the name instead of the subspace key. (The index secondary space, the index range subspace, and the index uniqueness violations space, the other keyspaces with index-scoped data, all already use the subspace key.) This means that (I think) it is the only place (other than the meta-data) where we persist the name to FDB, which means it's also the only thing that is blocking index renames. Then the subspace key can be *the* canonical on-disk subspace key, and the index name is just used as the index identifier in ephemeral state.

PR #854, opened 2020-03-14
Resolves #852: Enhance KeySpaceCountTree to allow for application-specific resolution

### LINKED ISSUE #852: Enhance KeySpaceCountTree to allow for application-specific resolution
Some directory leaves are record stores. It may be possible to cheaply decode the primary key. And so on.

MMcM:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -83,35 +389,181 @@ protected TupleKeyCountTree newPrefixChild(@Nonnull Object prefix) {
...
+    }
+
+    /**
+     * Determine whether this leaf of the {@link KeySpaceDirectory} tree is the root of a {@link com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore}.
+     * @param context an open transaction to use to read from the database
+     * @param resolvedParent the resolved parent node
+     * @param object the {@link com.apple.foundationdb.tuple.Tuple} element for this node
+     * @return {@code true} if this path stores a record store
+     */
+    protected boolean isRecordStoreLeaf(@Nonnull FDBRecordContext context, @Nonnull Resolved resolvedParent, @Nullable Object object) {
```
scgray: Maybe not for this PR, but it seems that it may be very useful to formalize where record stores live in the keyspace (so that a directory "knows" that is contains record stores)...but, I guess for this to really be useful, then the directory should also have some way to describe how to find the metadata for the stores under it.  Anyway, just thinking out loud.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -171,6 +171,16 @@ public RecordType getRecordTypeForDescriptor(@Nonnull Descriptors.Descriptor des
         return recordType;
     }
 
+    @Nonnull
+    public RecordType getRecordTypeFromRecordTypeKey(@Nonnull Object recordTypeKey) {
```
alecgrieser: This and `getIndexFromSubspaceKey` should probably document that they throw a `MetaDataException` if the object specified is not found in the meta-data.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -23,29 +23,335 @@
...
 import javax.annotation.Nullable;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
 /**
  * Count keys and resolve back to key space paths.
  */
 public class KeySpaceCountTree extends TupleKeyCountTree {
```
alecgrieser: If I understand the implementation correctly, is the idea that users will extend this class and provide their own implementation of `getRecordStoreMetaData` and `isRecordStoreLeaf`? It might be good to clarify that, though this class is maybe a tad experimental for too much documentation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -83,35 +389,181 @@ protected TupleKeyCountTree newPrefixChild(@Nonnull Object prefix) {
...
+     * @param object the {@link com.apple.foundationdb.tuple.Tuple} element for this node
+     * @return a future that completes to the record store's meta-data or {@code null}
+     */
+    protected CompletableFuture<RecordMetaData> getRecordStoreMetaData(@Nonnull FDBRecordContext context, @Nonnull Resolved resolvedParent, @Nullable Object object) {
+        return NO_META_DATA;
+    }
+
+    protected CompletableFuture<Resolved> resolvePrimaryKeyField(@Nonnull FDBRecordContext context, @Nonnull Resolved resolvedParent, @Nullable Object object,
+                                                                 @Nonnull KeyExpression primaryKey, int index) {
+        List<KeyExpression> keys = primaryKey.normalizeKeyForPositions();
```
alecgrieser: Hm, we have a few places that we try and rendezvous the tuple we see with a key expression, but they're all a little scattershot. I think having this is here is probably fine, though perhaps when some of the new planner changes make that easier, it would be good to switch to that, or something.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -83,35 +389,181 @@ protected TupleKeyCountTree newPrefixChild(@Nonnull Object prefix) {
...
+            if (key instanceof FieldKeyExpression) {
+                return CompletableFuture.completedFuture(new ResolvedPrimaryKeyField(resolvedParent, ((FieldKeyExpression)key).getFieldName(), object, object));
+            }
+        }
+        return UNRESOLVED;
+    }
+
+    protected CompletableFuture<Resolved> resolveIndexField(@Nonnull FDBRecordContext context, @Nonnull Resolved resolvedParent, @Nullable Object object,
+                                                            @Nonnull Index index, @Nonnull KeyExpression rootExpression, int fieldIndex) {
+        return UNRESOLVED;
```
alecgrieser: Why is this `UNRESOLVED` rather than following the same logic as in `resolvePrimaryKey`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -83,35 +389,181 @@ protected TupleKeyCountTree newPrefixChild(@Nonnull Object prefix) {
...
+                        final Index index;
+                        try {
+                            index = recordStoreKeyspace.getRecordMetaData().getIndexFromSubspaceKey(object);
+                        } catch (RecordCoreException ex) {
+                            break;
+                        }
+                        return CompletableFuture.completedFuture(new ResolvedIndexKeyspace(resolvedParent, index));
+                    }
+                    break;
+                default:
```
alecgrieser: Would it also make sense to have a `case INDEX_STATE_SPACE` that got the index from the name (instead of the subspace key)? Or do we not want to do that to protect against that changing because of how #514 could be resolved?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
MMcM: Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: Yes. I added Javadoc to that effect on those two methods.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: Yes, agreed. But that feels outside the scope of this branch.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: Absent some new protocol in the factory / maintainer, this class would need to know for each index type how the keys it uses correspond to the root expression.

For example, a time-window leaderboard has a leaderboard index before the group starts.

Is it worth doing for, like, `VALUE`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: Yes. Also, since the current value is the name of the index, resolving it isn't so important. It will display more or less correctly just as an unresolved value.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: I suppose there could be a method on the factory (the maintainer needs a record store and I'm not sure making one from the example path is legit) that returned a list of key expression, defaulting to normalizing the root expression. Aggregate indexes could return just the group and leaderboard could prepend a leaderboard id field.

alecgrieser: Yeah, I think maybe having a method on the factory is the right choice at some point. I believe there's some possibility that something like that will come out of the planner work, in order to accommodate producing some IR based on the index definition. I could also see us wanting to do `VALUE` indexes sooner rather than later, but maybe not yet.

PR #859, opened 2020-03-17
Split out an advanced form of planCoveringAggregateIndex

MMcM:
This should be harmless enough even if it turns out not to be needed.


MMcM:
test this please

PR #861, opened 2020-03-18
Resolves #860: Try planning an IN predicate as an OR of equalities.

### LINKED ISSUE #860: RecordQueryPlanner should try planning IN predicates ORs of equality predicates
Currently, the `RecordQueryPlanner` tries to the plan `IN` predicates using an in-join. However, there are situations where this isn't possible, such as if a sort is specified that does not include the field with the `IN` predicate. Currently, we just 

For example, suppose that we have a compound index `concat(field("a"), field("b"))`. A query like:
```
RecordQuery.newBuilder()
    .setRecordType("Record")
   .setFilter(Query.field("a").in(asList(1, 4, 2))
   .setSort(field("b"))
   .build()
```
can't be planned as an in-join using that index, since doing so would give a sort order of `concat(field("a"), field("b"))`. However, the query can be logically transformed into a query like:
```
RecordQuery.newBuilder()
    .setRecordType("Record")
   .setFilter(Query.or(field("a").equalsValue(1), field("a").equalsValue(4), field("a").equalsValue(2)))
   .setSort(field("b"))
   .build()
```
we can plan it as a union of three scans of the index, each of which is sorted by `field("b")` alone since the first element of the index has an equality predicate.

The planner doesn't even attempt this transformation today, but it should.

nschiefer:
If the planner detects that it can't implement an IN predicate using an in-join (for example, because of an incompatible sort order), it tries to rewrite it as an OR predicate instead.

See #860 for a description of the specific problem that this solves.

The implementation is a bit hacky and ends up relying on pieces of the new planner to make decisions. This might not be great hygiene, but it seemed like the right tool for determining when to use one plan or another.


nschiefer:
It occurred to me that I should probably catch the `DNFTooLargeException` in the `asOr()` branch, in case that gets thrown here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -340,12 +341,27 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
     @Nullable
     private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryComponent filter, boolean needOrdering) {
         final InExtractor inExtractor = new InExtractor(filter);
+        ScoredPlan withInAsOr = null;
         if (planContext.query.getSort() != null) {
-            inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse());
+            if (!inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse())) {
+                // Can't implement as an in join because of the sort order. Try as an OR instead.
```
MMcM: Do we want to have some switch / threshold in the planner config to allow this to be introduced gracefully?

nschiefer: That seems like a reasonable thing to do. I'll add a configuration option to `RecordQueryPlanner`.

nschiefer:
Re: DNF explosion, it looks like this part of the planner doesn't use the full normalization logic, although I'm not sure why. I added a test with a very large CNF (large enough to cause a stack overflow) to make sure this doesn't change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
nschiefer: Done, with some slight gymnastics to keep a backwards-compatible API. I don't know if I got the right balance but I tried. ¯\_(ツ)_/¯

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1195,7 +1236,7 @@ private ScoredPlan planOrderedUnion(@Nonnull PlanContext planContext, @Nonnull L
         if (unionPlan.getComplexity() > complexityThreshold) {
             throw new RecordQueryPlanComplexityException(unionPlan);
         }
-        return new ScoredPlan(1, unionPlan, Collections.emptyList(), anyDuplicates, includedRankComparisons);
+        return new ScoredPlan(0, unionPlan, Collections.emptyList(), anyDuplicates, includedRankComparisons);
```
nschiefer: I'm not sure whether to gate this behind the config parameter, too. That feels hacky, but is probably safer.

MMcM: Yeah, that is probably safer, since reasoning about these constant scores is sometimes tricky.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -341,12 +366,28 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
     @Nullable
     private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryComponent filter, boolean needOrdering) {
         final InExtractor inExtractor = new InExtractor(filter);
+        ScoredPlan withInAsOr = null;
         if (planContext.query.getSort() != null) {
-            inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse());
+            if (!inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse()) && // needs to come first, to clear sort
```
alecgrieser: Oh, hm. So this essentially relies on the side effect from this operation? I don't love having this happen within the if-condition (rather than it happening on the previous line and capturing the success or failure as an explicit `boolean`) as I don't think people expect side effects there. But it's probably acceptable if we want to keep it like this and take the PR prior to changing it (and the comment helps).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+package com.apple.foundationdb.record.query.plan;
+
+import com.apple.foundationdb.annotation.API;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A set of configuration options for the {@link RecordQueryPlanner}.
```
alecgrieser: At some point, we'll probably want to document what these configuration parameters actually do, but perhaps not today.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: Looks like you dropped this: \\

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/InExtractor.java
```diff
@@ -157,10 +193,11 @@ public void setSort(@Nonnull KeyExpression key, boolean reverse) {
             if (!found) {
                 // There is a requested sort ahead of the ones from the IN's, so we can't do it.
                 cancel();
```
alecgrieser: It feels a little weird that this still calls `cancel`, with the new stuff. Like, in theory, it should be able to use the extracted sort components from the `IN` and then use a `UnionPlan` for the rest. Though I think it's still _safe_ to do it as it is here, so perhaps an optimization for another time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1196,7 +1237,12 @@ private ScoredPlan planOrderedUnion(@Nonnull PlanContext planContext, @Nonnull L
         if (unionPlan.getComplexity() > complexityThreshold) {
             throw new RecordQueryPlanComplexityException(unionPlan);
         }
-        return new ScoredPlan(1, unionPlan, Collections.emptyList(), anyDuplicates, includedRankComparisons);
+
+        // If we don't change this when shouldAttemptFailedInJoinAsOr() is true, then we _always_ pick the union plan,
+        // rather than the in join plan.
+        int score = getConfiguration().shouldAttemptFailedInJoinAsOr() ? 0 : 1;
```
alecgrieser: Hmm, interesting. Should the logic in the join filter that chooses between the `Union` and the `InJoin` be changed then (maybe ignoring the score if we can't trust it?).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -261,6 +267,77 @@ public void testInQueryIndexSortedDifferently() throws Exception {
...
+                        concat(field("num_value_3_indexed"), field("str_value_indexed")));
+        complexQuerySetup(hook);
+        final List<Integer> inList = asList(1, 4, 2);
+        RecordQuery query = RecordQuery.newBuilder()
+                .setRecordType("MySimpleRecord")
+                .setFilter(Query.field("num_value_3_indexed").in(inList))
+                .setSort(field("str_value_indexed"))
+                .build();
+
+        if (planner instanceof RecordQueryPlanner) { // will always be true, but it's good to check before casting
```
alecgrieser: I suppose technically, this could be a `@DualPlannerTest`, then (though not saying it should).

Or this could assert the planner type so that if someone were to make it a `DualPlannerTest` at some point, they get an error rather than it succeeding for the wrong reasons.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -261,6 +267,77 @@ public void testInQueryIndexSortedDifferently() throws Exception {
...
+     * gets planned as a normal IN query rather than throwing an exception.
+     */
+    @Test
+    public void cnfAsInQuery() throws Exception {
+        RecordMetaDataHook hook = metaData ->
+                metaData.addIndex("MySimpleRecord", "compoundIndex",
+                        concat(field("num_value_3_indexed"), field("str_value_indexed")));
+        complexQuerySetup(hook);
+
+        // A CNF who's DNF size doesn't fit in an int, expressed with IN predicates.
```
alecgrieser:
```suggestion
        // A CNF whose DNF size doesn't fit in an int, expressed with IN predicates.
```

I'm sorry

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfigurationTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Test of the semantics of setting configuration options on {@link RecordQueryPlanner}.
+ * This is mostly "executable documentation" that makes sure that we never change the contract without meaning to.
+ */
+public class RecordQueryPlannerConfigurationTest {
+    private static RecordQueryPlanner blankPlanner() {
+        return new RecordQueryPlanner(
+                RecordMetaData.newBuilder().setRecords(TestRecordsEnumProto.getDescriptor()).build(),
```
alecgrieser:
```suggestion
                RecordMetaData.build(TestRecordsEnumProto.getDescriptor()),
```
Equivalent, but perhaps more idiomatic

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
nschiefer: I'm happy to extract this as a boolean. I don't really want to do a larger refactor, since this was definitely meant to perturb things as little as possible.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/InExtractor.java
nschiefer: Agreed, but I'd rather not mess with this code if we don't have to.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
nschiefer: Possibly. I'm a bit worried about doing so because I don't really know what the `score` is trying to capture. If we think that's lower risk, I'm for it.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
nschiefer: I suppose it could. 🙃 

I switched this to an assertion instead.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
nschiefer: Whoops, fixed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfigurationTest.java
nschiefer: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: Yeah, don't really need a larger refactor; was more just asking for it to be on its own line, though maybe that's not actually clearer.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
nschiefer: Documented.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
nschiefer: I think it's a little clearer.

PR #863, opened 2020-03-20
Quick and dirty index field resolution.

MMcM:
Needs to be moved into `IndexMaintainerFactory`, but good enough for most types now.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
@@ -527,7 +527,7 @@ public class KeySpaceCountTree extends TupleKeyCountTree {
     }
 
     // TODO: Get this from the IndexMaintainerFactory via some new interface (the IndexMaintainer needs a RecordStore).
-    private List<KeyExpression> indexStoredKeys(@Nonnull Index index) {
+    protected List<KeyExpression> indexStoredKeys(@Nonnull Index index) {
         KeyExpression storedKey = index.getRootExpression();
         if (storedKey instanceof GroupingKeyExpression) {
             if (index.getType().equals(IndexTypes.RANK) ||
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
@@ -25,8 +25,11 @@ import com.apple.foundationdb.clientlog.TupleKeyCountTree;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordMetaData;
 import com.apple.foundationdb.record.metadata.Index;
+import com.apple.foundationdb.record.metadata.IndexTypes;
+import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.RecordType;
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -503,23 +509,42 @@ protected TupleKeyCountTree newPrefixChild(@Nonnull Object prefix) {
...
-                return resolveIndexField(context, resolvedParent, object, index, indexExpression, distance);
+            List<KeyExpression> storedKeys = indexStoredKeys(index);
+            if (distance < storedKeys.size()) {
+                return resolveIndexField(context, resolvedParent, object, index, storedKeys.get(distance), distance);
             }
         }
         return UNRESOLVED;
     }
 
+    // TODO: Get this from the IndexMaintainerFactory via some new interface (the IndexMaintainer needs a RecordStore).
```
MMcM: I'm not opposed to trying, but as #854 points out, this might be more clear as something the new planner develops.

alecgrieser: I actually think we don't want to add this to the index maintainer yet, given (a) the new planner might change things/make things better and (b) this API is still experimental, so adding to the index maintainer factories now seems maybe a little premature.

I could see an argument, though, this method should be `protected` in case someone wants to provide an implementation for a custom index of their own devising by overriding this method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
```diff
@@ -548,24 +573,23 @@ protected boolean isRecordStoreLeaf(@Nonnull FDBRecordContext context, @Nonnull
...
-            KeyExpression key = keys.get(index);
-            while (key instanceof NestingKeyExpression) {
-                key = ((NestingKeyExpression)key).getChild();
-            }
-            if (key instanceof FieldKeyExpression) {
-                return CompletableFuture.completedFuture(new ResolvedPrimaryKeyField(resolvedParent, ((FieldKeyExpression)key).getFieldName(), object, object));
-            }
-        }
-        return UNRESOLVED;
+                                                                 @Nonnull KeyExpression fieldKey, int fieldIndex) {
```
alecgrieser: It looks like `fieldIndex` can be removed as a parameter from both this and `resolveIndexField`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: Yeah, that seems like a reasonable compromise.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/clientlog/KeySpaceCountTree.java
MMcM: I was thinking an overriding method might understand that the zeroth field of the primary key is special, for example. But there are probably other ways it could accomplish such a check.

alecgrieser: Okay, I see. Well, it probably doesn't hurt to leave it like that. I'll approve, and if you want to leave it as is, then I guess feel free to merge.

Issue #853, opened 2020-03-13
KeySpacePath resolution shouldn't use its own transactions


alecgrieser:
At the moment, when a keyspace path that uses the `DirectoryLayerDirectory` (or anything else that uses the `LocatableResolver`) resolves a key through a directory layer, it does so in its own transaction. This means that it must: (1) get its own read version and (2) commit (if any writes are done). The benefit of doing this is that (1) it avoids placing uncommitted data into the cache and (2) transaction conflicts can be retried within the `LocatableResolver`.

I think, on the whole, given the expense of creating new transactions, that this overhead is probably not worth the benefit. I propose that the `LocatableResolver` should be modified so that a user can give it its own transaction, and the key space path resolution can happen in line with everything else. I believe that that requires the following changes:

1. New APIs in the `LocatableResolver` that take `FDBRecordContext`s.
1. The `DirectoryLayerDirectory` should use that API
1. Cache management should only cache things that have actually been committed, which can be done by scheduling an `AfterCommit` hook

The good news is that the implementations of `LocatableResolver` actually already expose methods that only take a context (not a database), so they don't need to be changed (from what I can tell).

PR #865, opened 2020-03-24
Resolves #864: KeySpacePath resolution should re-use the initiating transaction's read version

### LINKED ISSUE #864: KeySpacePath resolution should re-use the initiating transaction's read version
As discussed within #853, each directory layer resolution by a key-space path creates its own transaction to do directory layer things, mainly because otherwise, cache management is difficult.

Issue #853 proposes fixing that by using the same transaction as the main operation, but that still has a somewhat complicated transaction lifestyle story. A simpler (but almost as good) approach would be to have resolving from a `LocatableResolver` use the passed transaction's read version, so the only overhead from these extra transactions is some extra memory (and not a network request).

alecgrieser:
This adds overloads to `LocatableResolver::resolve` that allow a transaction to be passed, and then the resolver uses that transaction's read version when performing reads from the directory layer. In the "usual case" (which should be when a new entry does not need to be created), this transaction won't be committed anyway, so the only loss in efficiency over using the transaction itself is some extra memory (and no extra network requests).

It would probably still be a good idea to do #853 (if we can figure out what we want to do about caching), but this is easier and gets us 99% of the benefit. That being said, there is more cognitive load needed to understand this API, which is perhaps a smell.

This resolves #864.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBReverseDirectoryCache.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBReverseDirectoryCache.java
@@ -376,7 +376,7 @@ public class FDBReverseDirectoryCache {
      * used for tests in order to avoid spurious conflicts.
      */
     @VisibleForTesting
-    public void waitUntilReady() {
+    public void waitUntilReadyForTesting() {
         reverseDirectoryCacheEntry.join();
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBReverseDirectoryCache.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBReverseDirectoryCache.java
@@ -371,6 +371,15 @@ public class FDBReverseDirectoryCache {
         }
     }
 
+    /**
+     * Wait for any asynchronous work started at object creation time to complete. This should only be
+     * used for tests in order to avoid spurious conflicts.
+     */
+    @VisibleForTesting
+    public void waitUntilReady() {
+        reverseDirectoryCacheEntry.join();
+    }
+
     private CompletableFuture<Subspace> getReverseCacheSubspace(LocatableResolver scope) {
         return reverseDirectoryCacheEntry.thenCombine(scope.getBaseSubspaceAsync(), (entry, subspace) ->
                 subspace.subspace(Tuple.from(entry)));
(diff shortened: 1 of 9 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/LocatableResolver.java
```diff
@@ -113,71 +116,194 @@ public FDBDatabase getDatabase() {
         );
     }
 
+    @Nonnull
+    private <T> CompletableFuture<T> runAsyncBorrowingReadVersion(@Nonnull FDBRecordContext parentContext, @Nonnull Function<FDBRecordContext, CompletableFuture<T>> retriable) {
+        final FDBDatabaseRunner runner = parentContext.newRunner();
```
alecgrieser: If we wanted to also set the max attempts, we could do that here

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ExtendedDirectoryLayerTest.java
```diff
@@ -99,12 +99,12 @@ private static void testReadCompatible(LocatableResolver writer, LocatableResolv
         Map<String, Long> allocations = new HashMap<>();
         for (int i = 0; i < 50; i++) {
             String key = "key-" + i;
-            Long value = writer.resolve(null, key).join();
+            Long value = writer.resolve((FDBStoreTimer)null, key).join();
```
MMcM: Perhaps these tests could be improved by yet another overload that _only_ took the key and did the `null` itself.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBReverseDirectoryCache.java
```diff
@@ -371,6 +371,15 @@ public void rebuild(LocatableResolver scope) {
         }
     }
 
+    /**
+     * Wait for any asynchronous work started at object creation time to complete. This should only be
+     * used for tests in order to avoid spurious conflicts.
+     */
+    @VisibleForTesting
+    public void waitUntilReady() {
```
scgray: Since the `@VisibleForTesting` doesn't really help when, say, your favorite IDE is autocompleting potential methods, maybe it would be a good idea to also name this method `waitUntilReadyForTesting` (unless we can restrict the use of methods tagged with this automatically?).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ExtendedDirectoryLayerTest.java
alecgrieser: Yeah, though it's a little weird that the optional parameter is the first (rather than the last).

alecgrieser: I added this as a separate commit with this PR, for size

PR #878, opened 2020-03-26
Resolves #877: Add utility method for casting wrapped type of FDBRecord

### LINKED ISSUE #877: Add utility method for casting wrapped type of FDBRecord
If one has an `FDBRecord<Message>`, it's currently somewhat difficult to turn that into an `FDBRecord<SomeOtherMessageType>`. This is true for various methods throughout the `FDBRecord` hierarchy.

I think we could solve this with a new method on the interface, say, `cast`, that could cast the internal type. It could take a target class (i.e., `Class<M extends Message>`) and (maybe) a builder-supplier for that type. Then we could add a utility for casting a proto message (either just casting the objet if it's already the right type or merging the message if its not) to the right type.

One might want to do this (over just casting the result of `.getRecord`) because it would allow you to keep the other information around on the `FDBRecord` object, e.g., the primary key or version, while also gaining some amount of type safety.

alecgrieser:
This resolves #877, though it's currently just a draft to potentially get eyes on the approach.

This adds a utility class with some static methods. Then the various implementations add ways of recreating themselves with the message type changed. Note that each one (unless I messed up) overrides the return type to be themselves.

Some missing information:

1. Documentation on the methods.
1. A real solution for synthetic records.
1. Tests
1. Release notes

But I wanted this some place people could see.

### DIFF
--- a/examples/src/main/java/com/apple/foundationdb/record/sample/Main.java
+++ b/examples/src/main/java/com/apple/foundationdb/record/sample/Main.java
@@ -88,8 +88,8 @@ public class Main {
             do {
                 result = cursor.getNext();
                 if (result.hasNext()) {
-                    SampleProto.Customer.Builder builder = SampleProto.Customer.newBuilder().mergeFrom(result.get().getRecord());
-                    names.add(builder.getFirstName() + " " + builder.getLastName());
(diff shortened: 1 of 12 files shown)


alecgrieser:

PR #879, opened 2020-03-27
Resolves #866: Replace some aggregate metrics with accurate counters

### LINKED ISSUE #866: Replace some aggregate metrics with accurate counters
I think it should be possible to have `FDBContext()` return federated wrappers around the underlying transaction classes that add store timers to certain operations. Specifically:

```
bytes_read
bytes_written
reads
writes
deletes
commits
```

The corresponding aggregate counters that are currently in place could then be removed.

scgray:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
```diff
@@ -0,0 +1,401 @@
...
+    @Deprecated
+    public void clearRangeStartsWith(byte[] prefix) {
+        underlying.clearRangeStartsWith(prefix);
+        increment(FDBStoreTimer.Counts.DELETES);
+    }
+
+    @Override
+    public void mutate(MutationType opType, byte[] key, byte[] param) {
+        underlying.mutate(opType, key, param);
+        /* Do we want to track each mutation type separately as well? */
```
scgray: As noted, do we want per-mutation type counters as well? Obviously can always come later...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
```diff
@@ -0,0 +1,378 @@
...
+    @Override
+    public CompletableFuture<byte[]> getKey(KeySelector keySelector) {
+        timer.increment(FDBStoreTimer.Counts.READS);
+        return underlying.getKey(keySelector).thenApply(this::recordRead);
+    }
+
+    @Override
+    public AsyncIterable<KeyValue> getRange(KeySelector begin, KeySelector end) {
+        /* Should this could as one read? */
+        timer.increment(FDBStoreTimer.Counts.READS);
```
scgray: As noted, for these iterators, they are being counted a single read. Is this ok?  Can we do better?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
alecgrieser: I think this is probably fine for now. Maybe some day, though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -0,0 +1,249 @@
...
+
+    @Override
+    public CompletableFuture<byte[]> getKey(KeySelector keySelector) {
+        timer.increment(FDBStoreTimer.Counts.READS);
+        return underlying.getKey(keySelector).thenApply(this::recordRead);
+    }
+
+    @Override
+    public AsyncIterable<KeyValue> getRange(KeySelector begin, KeySelector end) {
+        /* Should this could as one read? */
```
alecgrieser: Good question. I think (?) that the FDB metric that goes into machine readable status is 1 range read = 1 read, so if we wanted to be able to correlated, say, reads from this timer to reads from that timer, we'd want them to be the same. This also means that we count work from empty range reads, which we may care about. I suppose we still know how many entries get back from our other, higher level metrics (like `FDBStoreTimer.Counts.LOAD_RECORD`). Hmm.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -634,6 +584,18 @@ public String logKey() {
...
+        BYTES_READ("bytes read", true),
+        /** The number of bytes written, not including deletes. */
+        BYTES_WRITTEN("bytes written", true),
+        /** Total number of delete (clear) operations. */
+        READS("reads", false),
+        /** Total number of write operations. */
+        WRITES("writes", false),
+        /** Total number of delete (clear) operations. */
+        DELETES("deletes", false),
+        /** Total number of mutations operations. */
```
alecgrieser:
```suggestion
        /** Total number of mutation operations. */
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -0,0 +1,249 @@
...
+        @Override
+        public CompletableFuture<List<KeyValue>> asList() {
+            final List<KeyValue> result = new ArrayList<>();
+            final AsyncIterator<KeyValue> iterator = iterator();
+            return AsyncUtil.whileTrue(() -> iterator.onHasNext().thenApply(hasNext -> {
+                if (hasNext) {
+                    result.add(iterator.next());
+                }
+                return hasNext;
+            })).thenApply(vignore -> result);
```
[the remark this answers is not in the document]
scgray: Presumably my loop is just doing what the underlying code would have done, however I agree that the underlying code should be in charge of building the list.  Updated.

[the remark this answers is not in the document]
alecgrieser: Yeah, both the old loop and new look look the same in terms of "what gets recorded", but I was mostly "concerned" with reconstructing the list.

PR #880, opened 2020-04-01
Fixes #876: TextIndex does not instrument BunchedMap reads and writes

### LINKED ISSUE #876: TextIndex does not instrument BunchedMap reads and writes
The `TextIndexMaintainer` currently _kind of_ instruments the reads and writes it does as part of maintenance in that it tries to keep track of the number of keys written (based on token counts), but it doesn't do a great job. The primary reason for this is that the class that does most of the work, the `BunchedMap` is in `fdb-extensions`, which doesn't have a great way of allowing an `FDBStoreTimer` in to add metrics (at the moment).

I think there should be a way to make this more instrumented. One idea would be to add, for example, protected "instrument" methods, that we could then override in `fdb-record-layer` with an instrumented version (or something like that).

There's a similar (though not identical) thing that happens with the rank indexes, though we have more granular instrumentation there.

alecgrieser:
The problem here was that the reading and writing of keys from FDB happened far enough away from the store timer logic, etc., that the previous logic was not even trying to look within the BunchedMap. However, quite a few reads and writes could happen as a result of text indexing, so getting the right numbers is fairly important. Note that once #866 is in, this may matter less (as the total reads and writes for the transaction will be more accurate), but doing it this way allows it as well allows it to (1) include the sizes of deleted values and (2) correctly correlate those operations to index updates. For that reason, I think this has value even after that other issue is addressed.

The architecture of this change is as follows:

1. The `BunchedMap` class has some `instrumentX` methods that instrument specific FDB operations. It provides default implementations that do nothing.
1. The `TextIndexMaintainer` includes its own implementation that includes an `FDBStoreTimer` and updates the appropriate timer events.
1. When index maintenance is performed, if the `FDBRecordContext` has a non-`null` timer, then it uses an instance of the instrumented class and provides its own timer.

This fixes #876.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -50,7 +50,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Performance** The `LocatableResolver::resolve` methods have overloads that allow ancillary transactions started for key space path resolution to avoid starting another read version request [(Issue #864)](https://github.com/FoundationDB/fdb-record-layer/issues/864)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** The metrics for text indexes now contain more accurate numbers for reads and writes [(Issue #876)](https://github.com/FoundationDB/fdb-record-layer/issues/876)
 * **Performance** Index states are now preloaded in fewer round trips [(Issue #881)](https://github.com/FoundationDB/fdb-record-layer/issues/881)
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/map/BunchedMap.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/map/BunchedMap.java
@@ -153,14 +153,16 @@ public class BunchedMap<K,V> {
     }
 
     /**
-     * Instrument a range read. The base implementation does nothing, but extenders are encouraged to
-     * override this method with their own implementations that, for example, records the total numbers
+     * Instrument a range read. The base implementation only returns the original future, but extenders are encouraged
+     * to override this method with their own implementations that, for example, records the total numbers
      * of keys read and their sizes.
      *
      * @param readFuture a future that will complete to a list of keys and values
+     * @return an instrumented future that returns the same values as the original future
      */
-    protected void instrumentRangeRead(@Nonnull CompletableFuture<List<KeyValue>> readFuture) {
-
+    @Nonnull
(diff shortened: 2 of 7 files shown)


alecgrieser:
Adding @scgray due to your work on instrumentation-related things

fdb-extensions/src/main/java/com/apple/foundationdb/map/BunchedMap.java
```diff
@@ -140,6 +152,44 @@ public BunchedMap(@Nonnull BunchedSerializer<K,V> serializer, @Nonnull Comparato
...
     }
 
+    /**
+     * Instrument a range read. The base implementation does nothing, but extenders are encouraged to
+     * override this method with their own implementations that, for example, records the total numbers
+     * of keys read and their sizes.
+     *
+     * @param readFuture a future that will complete to a list of keys and values
+     */
+    protected void instrumentRangeRead(@Nonnull CompletableFuture<List<KeyValue>> readFuture) {
```
scgray, line 162: This should probably return a `CompletableFuture<List<KeyValue>>` as `timer.instrument` returns a future.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexMaintainer.java
```diff
@@ -552,4 +559,54 @@ public boolean canDeleteWhere(@Nonnull QueryToKeyMatcher matcher, @Nonnull Key.E
...
+            timer.increment(FDBStoreTimer.Counts.SAVE_INDEX_VALUE_BYTES, value.length);
+            if (oldValue != null) {
+                // Or should this ignore the value altogether?
+                timer.increment(FDBStoreTimer.Counts.DELETE_INDEX_VALUE_BYTES, oldValue.length);
+            }
+        }
+
+        @Override
+        protected void instrumentRangeRead(@Nonnull CompletableFuture<List<KeyValue>> readFuture) {
+            timer.instrument(FDBStoreTimer.Events.SCAN_INDEX_KEYS, readFuture, executor).whenComplete((list, err) -> {
```
scgray, line 597: `timer.instrument` returns the instrumented future, so I think it probably needs to be returned here.

fdb-extensions/src/main/java/com/apple/foundationdb/map/BunchedMap.java
alecgrieser: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexMaintainer.java
alecgrieser: Sure.

alecgrieser:
I also realized I forgot to update the release notes, so I've addressed that as well.

docs/ReleaseNotes.md
```diff
@@ -50,7 +50,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Performance** The `LocatableResolver::resolve` methods have overloads that allow ancillary transactions started for key space path resolution to avoid starting another read version request [(Issue #864)](https://github.com/FoundationDB/fdb-record-layer/issues/864)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** The metrics for text indexes now contain more accurate numbers for reads and writes [(Issue #876)](https://github.com/FoundationDB/fdb-record-layer/issues/876)
```
scgray, line 53: Nice catch.

PR #893, opened 2020-04-06
Merge 2.8.104.28 to master

nschiefer:
This merges the changes on the 2.8.102.26 and 2.8.104.28 patch releases into `master`.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -370,7 +370,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             boolean canSort = inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse());
             if (!canSort && getConfiguration().shouldAttemptFailedInJoinAsOr()) {
                 // Can't implement as an in join because of the sort order. Try as an OR instead.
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1284,20 +1284,49 @@ private QueryComponent normalizeAndOr(AndComponent and) {
...
             if (child2 instanceof OrComponent && Query.isSingleFieldComparison(child1)) {
-                return OrComponent.from(distributeAnd(child1, ((OrComponent)child2).getChildren()));
+                return OrComponent.from(distributeAnd(Collections.singletonList(child1), ((OrComponent)child2).getChildren()));
             }
         }
         return and;
     }
 
-    private List<QueryComponent> distributeAnd(QueryComponent component, List<QueryComponent> children) {
+    private QueryComponent normalizeAndOrForInAsOr(@Nonnull QueryComponent component) {
```
MMcM, line 1296: For master (versus patch release), do we want this to be the only definition of `normalizeAndOr`? Or just live with this until a more major overhaul?

We could also take this as a clean merge from the release branch and do that separately.

nschiefer: I'd be up for changing this in master, but I think I'd rather do a clean merge. I opened #894 to track this.

PR #895, opened 2020-04-07
Merge 2.8.104.28 to 2.8.110

nschiefer:
This merges the changes on the 2.8.104 patch branch onto the 2.8.110 patch branch.


alecgrieser:
It would probably be good to get a release note on 2.8.110 saying when the merge happened, but perhaps getting a build out first is better.

Issue #482, opened 2019-03-24
Test failure in RankIndexTest::repeatedRankQuery


alecgrieser:
During one pull request build, at least one PR unrelated to rank indexes has encountered an error performing the `RankIndexTest::repeatedRankQuery` test. Here's the test output:

```
com.apple.foundationdb.record.provider.foundationdb.RankIndexTest > repeatedRankQuery() FAILED
    org.opentest4j.AssertionFailedError: expected: <[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus]> but was: <[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus, achilles, hector]>
        at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:52)
        at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:186)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:181)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:486)
        at com.apple.foundationdb.record.provider.foundationdb.RankIndexTest.repeatedRankQuery(RankIndexTest.java:746)
```

Here are the two lists on their own lines:

```
[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus]
[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus, achilles, hector]
```

So it looks like the test is sometimes adding an extra achilles and an extra hector at the end of the query.

One such PR that encountered this error was #160.

alecgrieser:
I suspect that there's something subtle going on in the `RankedSet`. This is classified here as a test failing, but it could also be the sign of some subtle race condition rather than a testing failure.

MMcM:
Happened again with #870

ScottDugas:
I managed to reproduce this locally. I then reopened the store after the test failed, and it fails reliably, so it is whatever is in the index. 
For reference the code that inserts these record looks like:
```java

Code:
rec.setName("patroclus")
        .addScore(-5)
        .addScore(5)
        .addScore(-10)
        .addScore(-11)
        .addScore(-8);
recordStore.saveRecord(rec.build());

rec.clear();
rec.setName("achilles")
[...]

PR #902, opened 2020-04-14
Fixes #901: False positive of uniqueness violations prevents building indexes

### LINKED ISSUE #901: False positive of uniqueness violations prevents building indexes
Assume record A, B have same value in field F. 

1. Add Record A, B
2. Create unique index M on field F.
3. Build unique index M. When the second record hit `updateOneKey`, two entries are added to the uniqueness violation subspace by `addUniquenessCheck`: A->B, B->A. It failed to be marked readable in the end because of the non-empty uniqueness violation subspace. Index M keeps in write-only state (full range has been scanned and is being maintained. )
4. Delete Record B. Uniqueness violation entry B->A is deleted by `updateOneKey`.
5. Build unique index M with BUILD_IF_DISABLED_CONTINUE_BUILD_IF_WRITE_ONLY option (i.e. without clearing the space). All range is built, try to mark it readable but failed because there is one entry in the uniqueness violation subspace (A->B).

I think the fix is that when deleting a uniqueness violation entry, it should delete the last one if it is the second last one of the same index same value field.

nblintao:
See background from https://github.com/FoundationDB/fdb-record-layer/issues/901

The main problem fixed by this PR is that: `FDBRecordStore.resolveUniquenessViolation` used to be relied on to delete the last uniqueness violation entry when delete the second last one. But this is not the only way the uniqueness violation is resolved. It doesn't cover the case that the violation is resolved when record is deleted organically. To resolve this problem, now it will delete the last entry as part of deleting records (See `removeUniquenessViolationsAsync`).

This change also tries to consolidate uniqueness logic at `StandardIndexMaintainer` level:
- `FDBRecordStore .addUniquenessCheck`  -> `StandardIndexMaintainer.checkUniqueness`
- `IndexMaintainer.updateUniquenessViolations` -> `StandardIndexMaintainer .addUniquenessViolations` and `StandardIndexMaintainer.removeUniquenessViolationsAsync`

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
@@ -114,10 +114,27 @@ public abstract class IndexMaintainer {
      * @param valueKey the indexed key that is (apparently) not unique
      * @param primaryKey the primary key of one record that is causing a violation
      * @param existingKey the primary key of another record that is causing a violation (or <code>null</code> if none specified)
-     * @param remove <code>true</code> if removing the violation and <code>false</code> if adding it
+     * @param remove <code>true</code> if removing the violation and <code>false</code> if adding it. When
+     * <code>true</code>, it is favored use {@link #removeUniquenessViolationsAsync(Tuple, Tuple)} instead.
      */
     @Nonnull
     public abstract void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey, boolean remove);
+
+    /**
+     * Remove a uniqueness violation within the database. This is used to keep track of
+     * uniqueness violations that occur when an index is in write-only mode, both during
+     * the built itself and by other writes. This means that the writes will succeed, but
+     * it will cause a later attempt to make the index readable to fail.
+     *
+     * This will remove the last uniqueness violation entry when removing the second last one under the same
+     * <code>valueKey</code>, in contrast to {@link #updateUniquenessViolations(Tuple, Tuple, Tuple, boolean)} which
+     * may not.
+     * @param valueKey the indexed key that is (apparently) not unique
+     * @param primaryKey the primary key of one record that is causing a violation
+     * @return a future that is complete when the uniqueness violation is removed
+     */
+    @Nonnull
+    public abstract CompletableFuture<Void> removeUniquenessViolationsAsync(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey);
 
     /**
      * Scans through the list of uniqueness violations within the database.
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
@@ -106,35 +106,6 @@ public abstract class IndexMaintainer {
     public abstract <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexableRecord<M> oldRecord,
                                                                        @Nullable FDBIndexableRecord<M> newRecord);
 
-    /**
-     * Remove or add a uniqueness violation within the database. This is used to keep track of
-     * uniqueness violations that occur when an index is in write-only mode, both during
-     * the built itself and by other writes. This means that the writes will succeed, but
-     * it will cause a later attempt to make the index readable to fail.
-     * @param valueKey the indexed key that is (apparently) not unique
-     * @param primaryKey the primary key of one record that is causing a violation
-     * @param existingKey the primary key of another record that is causing a violation (or <code>null</code> if none specified)
-     * @param remove <code>true</code> if removing the violation and <code>false</code> if adding it. When
-     * <code>true</code>, it is favored use {@link #removeUniquenessViolationsAsync(Tuple, Tuple)} instead.
-     */
-    @Nonnull
-    public abstract void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey, boolean remove);
-
-    /**
-     * Remove a uniqueness violation within the database. This is used to keep track of
-     * uniqueness violations that occur when an index is in write-only mode, both during
-     * the built itself and by other writes. This means that the writes will succeed, but
-     * it will cause a later attempt to make the index readable to fail.
-     *
-     * This will remove the last uniqueness violation entry when removing the second last one under the same
-     * <code>valueKey</code>, in contrast to {@link #updateUniquenessViolations(Tuple, Tuple, Tuple, boolean)} which
-     * may not.
-     * @param valueKey the indexed key that is (apparently) not unique
-     * @param primaryKey the primary key of one record that is causing a violation
-     * @return a future that is complete when the uniqueness violation is removed
-     */
-    @Nonnull
-    public abstract CompletableFuture<Void> removeUniquenessViolationsAsync(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey);
 
     /**
      * Scans through the list of uniqueness violations within the database.
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainer.java
@@ -20,9 +20,10 @@
 
 package com.apple.foundationdb.record.provider.foundationdb.indexes;
 
-import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.Transaction;
+import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncUtil;
+import com.apple.foundationdb.async.MoreAsyncUtil;
 import com.apple.foundationdb.async.RankedSet;
 import com.apple.foundationdb.record.EvaluationContext;
 import com.apple.foundationdb.record.FunctionNames;
@@ -115,8 +116,12 @@ public class RankIndexMaintainer extends StandardIndexMaintainer {
         final Subspace extraSubspace = getSecondarySubspace();
         final List<CompletableFuture<Void>> futures = new ArrayList<>();
         for (IndexEntry indexEntry : indexEntries) {
-            // First maintain an ordinary B-tree index by score.
+            // Maintain an ordinary B-tree index by score.
             CompletableFuture<Void> updateOrdinaryIndex = updateOneKeyAsync(savedRecord, remove, indexEntry);
+            if (!MoreAsyncUtil.isCompletedNormally(updateOrdinaryIndex)) {
+                futures.add(updateOrdinaryIndex);
+            }
+
             final Subspace rankSubspace;
             final Tuple scoreKey;
             if (groupPrefixSize > 0) {
@@ -127,9 +132,8 @@ public class RankIndexMaintainer extends StandardIndexMaintainer {
                 rankSubspace = extraSubspace;
                 scoreKey = indexEntry.getKey();
             }
-            futures.add(updateOrdinaryIndex.thenCompose(vignore ->
-                    RankedSetIndexHelper.updateRankedSet(state, rankSubspace, config, indexEntry.getKey(), scoreKey,
-                            remove)));
+            futures.add(RankedSetIndexHelper.updateRankedSet(state, rankSubspace, config, indexEntry.getKey(),
+                    scoreKey, remove));
         }
         return AsyncUtil.whenAll(futures);
     }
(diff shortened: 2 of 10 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -114,11 +114,28 @@ public Subspace getSecondarySubspace() {
...
+     *
+     * This will remove the last uniqueness violation entry when removing the second last one under the same
+     * <code>valueKey</code>, in contrast to {@link #updateUniquenessViolations(Tuple, Tuple, Tuple, boolean)} which
+     * may not.
+     * @param valueKey the indexed key that is (apparently) not unique
+     * @param primaryKey the primary key of one record that is causing a violation
+     * @return a future that is complete when the uniqueness violation is removed
+     */
+    @Nonnull
+    public abstract CompletableFuture<Void> removeUniquenessViolationsAsync(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey);
```
nblintao, line 137: Maybe I can just change the signature of `updateUniquenessViolations` rather than adding a new one.

alecgrieser: Yeah, it does seem a little weird that `updateUniquenessViolations(valueKey, primaryKey, existingKey, false)` doesn't call this, other than the fact that it would require async-ness.

I think it's probably better, in the long run, to have the `Async` version of `updateUniquenessViolations` be the thing. So, either we do something like just break the API by making it async (maybe on 2.9?) or do something like add an `Async` version, and by default, the async version calls the synchronous version and then returns `AsyncUtil.DONE`, but then we reverse that in the `StandardIndexMaintainer`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -359,6 +363,26 @@ public void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple p
...
+    public CompletableFuture<Void> removeUniquenessViolationsAsync(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey) {
+        Subspace uniqueValueSubspace = state.store.indexUniquenessViolationsSubspace(state.index).subspace(valueKey);
+        state.transaction.clear(uniqueValueSubspace.pack(primaryKey));
+        // Remove the last entry if it was the second last entry in the unique value subspace.
+        RecordCursor<KeyValue> uniquenessViolationEntries = KeyValueCursor.Builder.withSubspace(uniqueValueSubspace)
+                .setContext(state.context)
+                .setScanProperties(new ScanProperties(ExecuteProperties.newBuilder()
+                        .setReturnedRowLimit(2)
+                        .setIsolationLevel(IsolationLevel.SERIALIZABLE)
+                        .build()))
```
alecgrieser, line 377: I think we probably also want to set the streaming mode on this scan properties to `CursorStreamingMode.WANT_ALL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexMaintainer.java
```diff
@@ -116,7 +116,7 @@ public RankIndexMaintainer(IndexMaintainerState state) {
         final List<CompletableFuture<Void>> futures = new ArrayList<>();
         for (IndexEntry indexEntry : indexEntries) {
             // First maintain an ordinary B-tree index by score.
-            updateOneKey(savedRecord, remove, indexEntry);
+            CompletableFuture<Void> updateOrdinaryIndex = updateOneKeyAsync(savedRecord, remove, indexEntry);
```
alecgrieser, line 119: Is there a reason to not do something like:

```java
if (!MoreAsyncUtil.isCompletedNormally(updateOrdinaryIndex)) {
    futures.add(updateOridnaryIndex);
}
```

Then I think it would look for uniqueness violations in parallel with updating the ranked set. (And the check for if it's completed normally is just to avoid waiting on done futures, but it should be optional.)

I didn't leave a comment about it, but I think the same thing applies to the `TimeWindowLeaderboardIndexMaintainer`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -304,10 +304,11 @@ public boolean skipUpdateForUnchangedKeys() {
      * @param savedRecord the record being indexed
      * @param remove <code>true</code> if removing from index
      * @param indexEntry the entry for the index to be updated
+     * @return a future completed when the key is updated
      */
-    protected <M extends Message> void updateOneKey(@Nonnull final FDBIndexableRecord<M> savedRecord,
-                                                    final boolean remove,
-                                                    @Nonnull final IndexEntry indexEntry) {
+    protected <M extends Message> CompletableFuture<Void> updateOneKeyAsync(@Nonnull final FDBIndexableRecord<M> savedRecord,
```
scgray, line 309: I think I'm ok with this change, but I do know of at least one down-stream project that overrides `updateOneKey`.  If we were really concerned about compatibility, this could be changed to something like:

```
protected <M extends Message> CompletableFuture<Void> updateOneKeyAsync(@Nonnull final FDBIndexableRecord<M> savedRecord,
                                                                        final boolean remove,
                                                                        @Nonnull final IndexEntry indexEntry) {
    updateOneKey(savedRecord, remove, indexEntry);
    return postUpdateOneKeyAsync(savedRecord, remove, indexEntry).
}
```

This is kind of convoluted and gross though, so I'm not strongly advocating for it....just that it is a consideration.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -359,6 +363,26 @@ public void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple p
         }
     }
 
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> removeUniquenessViolationsAsync(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey) {
+        Subspace uniqueValueSubspace = state.store.indexUniquenessViolationsSubspace(state.index).subspace(valueKey);
```
scgray, line 369: Not for this PR, but I think it would make some sense to cache subspace computations that happen at high frequently (e.g. on every record that is accessed or changed), so we don't compute `state.store.indexUniquenessViolationsSubspace(state.index)` on every key.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -114,11 +114,28 @@ public Subspace getSecondarySubspace() {
...
     @Nonnull
     public abstract void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey, boolean remove);
 
+    /**
+     * Remove a uniqueness violation within the database. This is used to keep track of
+     * uniqueness violations that occur when an index is in write-only mode, both during
+     * the built itself and by other writes. This means that the writes will succeed, but
+     * it will cause a later attempt to make the index readable to fail.
+     *
+     * This will remove the last uniqueness violation entry when removing the second last one under the same
```
scgray, line 129: `<p>This will`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -114,11 +114,28 @@ public Subspace getSecondarySubspace() {
...
+     * <code>valueKey</code>, in contrast to {@link #updateUniquenessViolations(Tuple, Tuple, Tuple, boolean)} which
+     * may not.
```
scgray, line 131: It would be really useful too have a description of *why* this has to be done (TBH, I'm not clear on it myself)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
nblintao: Done.

nblintao:
I think I should rebase it to 2.9 after we agree on the API changes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -345,18 +350,74 @@ public boolean skipUpdateForUnchangedKeys() {
...
+                    if (!TupleHelpers.equals(primaryKey, existingKey)) {
+                        if (state.store.isIndexWriteOnly(state.index)) {
+                            addUniquenessViolations(valueKey, primaryKey, existingKey);
+                            addUniquenessViolations(valueKey, existingKey, primaryKey);
+                        } else {
+                            throw new RecordIndexUniquenessViolation(state.index, indexEntry, primaryKey, existingKey);
+                        }
+                    }
+                }, getExecutor()));
+        state.store.getRecordContext().addCommitCheck(checker);
```
nblintao, line 374: Could we just return the future considering `updateOneKey` is now async?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -352,20 +354,54 @@ public boolean skipUpdateForUnchangedKeys() {
...
+    /**
+     * Add a uniqueness violation within the database. This is used to keep track of
+     * uniqueness violations that occur when an index is in write-only mode, both during
+     * the built itself and by other writes. This means that the writes will succeed, but
+     * it will cause a later attempt to make the index readable to fail.
+     * @param valueKey the indexed key that is (apparently) not unique
+     * @param primaryKey the primary key of one record that is causing a violation
+     * @param existingKey the primary key of another record that is causing a violation (or <code>null</code> if none specified)
+     */
+    protected void addUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey) {
```
alecgrieser, line 386:
```suggestion
    protected void addUniquenessViolation(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -106,35 +106,6 @@ public Subspace getSecondarySubspace() {
...
-     * the built itself and by other writes. This means that the writes will succeed, but
-     * it will cause a later attempt to make the index readable to fail.
-     * @param valueKey the indexed key that is (apparently) not unique
-     * @param primaryKey the primary key of one record that is causing a violation
-     * @param existingKey the primary key of another record that is causing a violation (or <code>null</code> if none specified)
-     * @param remove <code>true</code> if removing the violation and <code>false</code> if adding it. When
-     * <code>true</code>, it is favored use {@link #removeUniquenessViolationsAsync(Tuple, Tuple)} instead.
-     */
-    @Nonnull
-    public abstract void updateUniquenessViolations(@Nonnull Tuple valueKey, @Nonnull Tuple primaryKey, @Nullable Tuple existingKey, boolean remove);
```
alecgrieser, line 120: I suppose it's reasonable that `IndexMaintainer` wouldn't have a method for updating uniqueness violations and that that would be something that only lives within the implementations, though it wasn't entirely obvious to me that that's right.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
alecgrieser: I believe the purpose of making it a pre-commit check is that it means that if someone violates a uniqueness constraint, then we don't accidentally commit the transaction.

I could see an argument being made, though, that we'd want to fail fast...if we can. I think we definitely want preserve the behavior where committing is not allowed if there's a uniqueness constraint violation caused by a record insert, though.

alecgrieser: (It's possibly worth memorializing this with a comment--something like, "add a pre-commit check to prevent accidentally committing and getting into an invalid state" or something.)

nblintao: Added the comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
nblintao: Fixed.

nblintao:
And rebased to 2.9

nblintao:
@alecgrieser Trying to commit suggested changes from GitHub UI for the first time. Let's see how it goes.

alecgrieser:
We'll also probably want to squash this when this goes in

nblintao:
Known issue https://github.com/FoundationDB/fdb-record-layer/issues/936
Test this please

alecgrieser:
Hmm, looks like this time it was #482, which should have been fixed. It's possible that change isn't on 2.9 yet or something, in which case it's probably noise

test this please

alecgrieser:
Yeah, this fix is in #914, which appears to only be on master

PR #904, opened 2020-04-14
Resolves #903 Reorganize tests

### LINKED ISSUE #903: FDBRecordStoreTest is huge
FDBRecordStoreTest is almost 4000 lines of test code. It should be broken up into more focused test fixtures for different features or components.

ScottDugas:
This has 4 commits that do cleanup, and one that moves all the tests around


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreUniquenessTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Tests of uniqueness checks.
+ */
+@Tag(Tags.RequiresFDB)
+public class FDBRecordStoreUniquenessTest extends FDBRecordStoreTestBase {
```
alecgrieser, line 43: This kind of feels like it should be, like, an "index" test rather than a uniqueness test, in that it's really testing that behavior rather than the store's. But probably a rename that could happen later.

alecgrieser:
Great, will merge after successful PRB

PR #909, opened 2020-04-17
Merge master to 2.9

ScottDugas:
Conflicts were the ReleaseNotes and FDBDatabaseFactory.
ReleaseNotes hopefully correctly merged...
FDBDatabaseFactory was just new fields near the change
to use FDB api version 620

PR #911, opened 2020-04-17
Resolves #910: Take build Docker Java version to 11

### LINKED ISSUE #325: Update javadoc references to asyncToSync after Java 11 upgrade
There is a bug in Java 8 in which:

    {@link FDBDatabase#asyncToSync(FDBStoreTimer, FDBStoreTimer.Wait, CompletableFuture)}

and

    {@link FDBRecordContext#asyncToSync(FDBStoreTimer.Wait, CompletableFuture)}

are seen as javadoc errors.  As such a number of places in the code use `<code>` tags instead. These should be replaced with `@link` references after we upgrade.

### LINKED ISSUE #910: Take Build Docker Java version to 11
The Record Layer should already be Java 11 compatible, but we should increase our JDK version in our builder to match. We probably don't want to update our target binary version (yet) in case someone is importing us into a Java 8 project, but increasing our build JDK should be fine as long as we still target Java 8 for binary compatibility.

We can probably also start targeting Java 11 for source compatibility, as those Java 8 projects using should be able to take the class files (even if we update our source files).

I propose doing this for 2.9 so that we can warn people.

alecgrieser:
This updates the JDK version we're using in our build docker, and increases the required JDK version for building the project to 11. AFAIK, there was only one thing we wanted in the build from JDK 11 that didn't require source changes, and it was related to Javadoc links. As a result, this resolves #910 (updating the Java version), and it resolves #325 (fixing some links).

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -40,20 +40,6 @@
     </wildcardResourcePatterns>
     <annotationProcessing>
       <profile default="true" name="Default" enabled="true" />
-      <profile name="Gradle Imported" enabled="true">
-        <outputRelativeToContentRoot value="true" />
-        <processorPath useClasspath="false">
-          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.google.guava/guava/23.5-jre/e9ce4989adf6092a3dab6152860e93d989e8cf88/guava-23.5-jre.jar" />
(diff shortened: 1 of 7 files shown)


alecgrieser:
This is against the 2.9 branch as Java version upgrades can be breaking (including this one). We should probably let #909 happen first to avoid merge conflicts, etc.

alecgrieser:
Incidentally, I accidentally created a version of this PR that still used JDK 8, and that PRB failed on the java doc generation, but the real PRB on JDK 11 didn't. So, I suppose that's a good sign this works on Java 11 and Java 11 only.

PR #912, opened 2020-04-17
Resolves #900: Gradle dependency refresh

### LINKED ISSUE #900: gradle.properties refresh
There are newer versions in Maven Central. Nothing looks to have advanced majorly, but it's probably best to update from time to time.

alecgrieser:
This updates most of our dependencies to their most recent versions, just to stay current. We'll probably also want to increase the FDB server version to 6.2.19, but I haven't done that yet because it will conflict with #911 in kind of a weird way.

This resolves #900.


alecgrieser:
I'm not sure if there are some dependencies where it would be worth taking a second look before adopting the newer version. All of the non-gradle/properties code is to fix deprecation warnings.

alecgrieser:
Also, on 2.9 just because transitive dependency increases may be a thing users want to take more seriously or something

PR #913, opened 2020-04-20
Fixes:897 InClause evaluation is N^2

jleach4:
Remove the n^2 evaluation for in clauses.

There are several queries I have been attempting to tune that have 100's of in clause participants coupled with 1K rows returned.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
@@ -39,6 +39,7 @@ import com.apple.foundationdb.tuple.ByteArrayUtil;
 import com.apple.foundationdb.tuple.ByteArrayUtil2;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Descriptors.FieldDescriptor.JavaType;
@@ -246,19 +247,18 @@ public class Comparisons {
         if (value == null || comparand == null) {
             return null;
         }
-        if ((comparand instanceof List<?>)) {
-            boolean hasNull = false;
-            value = toClassWithRealEquals(value);
-            for (Object comparandItem : (List<?>) comparand) {
-                if (value.equals(toClassWithRealEquals(comparandItem))) {
-                    return true;
-                }
-                hasNull |= comparandItem == null;
-            }
-            return hasNull ? null : false;
-        } else {
+        if (!(comparand instanceof List<?>)) {
             throw new RecordCoreException("IN comparison with a non-list type" + comparand.getClass());
         }
+        boolean hasNull = false;
+        value = toClassWithRealEquals(value);
+        for (Object comparandItem : (List<?>) comparand) {
+            if (value.equals(toClassWithRealEquals(comparandItem))) {
+                return true;
+            }
+            hasNull |= comparandItem == null;
+        }
+        return hasNull ? null : false;
     }
 
     @Nullable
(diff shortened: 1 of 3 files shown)


alecgrieser:
@fdb-build okay to test

alecgrieser:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -1011,7 +1028,21 @@ public Type getType() {
...
+            if (value == null) {
+                return null;
+            }
+            switch (type) {
+                case EQUALS:
+                    return value.equals(comparand);
+                case NOT_EQUALS:
+                    return !value.equals(comparand);
+                case STARTS_WITH:
+                    return compareListStartsWith(value, comparand);
```
alecgrieser, line 1040: It feels kind of wrong that this duplicates the logic in `evalListComparison`. It seems like this should instead check if it's an `IN` (and perhaps validate that `inProbe` is not `null`), and then otherwise fall back to `evalListComparison`.

jleach4: I like that, will change.  Thanks Alec.

jleach4:
@alecgrieser Can you take another look?

alecgrieser:
It looks like maybe you didn't push (or the push got lost during the recent GitHub problems)?

jleach4:
@alecgrieser Just Pushed and it seemed to work.  I am not totally happy with this from an organization perspective.  The comparison code might benefit from an actual type system.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -825,6 +837,10 @@ public String toString() {
         private final Type type;
         @Nonnull
         protected final String parameter;
+        @Nullable
+        protected Object comparand;
```
alecgrieser, line 841: I don't think we want to have a `comparand` member on this class. I don't think it's used, even in the new code?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -872,6 +888,15 @@ public Boolean eval(@Nonnull FDBRecordStoreBase<?> store, @Nonnull EvaluationCon
             } else if (comparand == COMPARISON_SKIPPED_BINDING) {
                 return Boolean.TRUE;
             } else {
+                if (type == Type.IN) {
+                    if (listComparison == null) {
```
alecgrieser, line 892: I believe this will do the wrong thing if one evaluates the same comparison with different `EvaluationContext`s. In particular, there's no guarantee that the list comparand hasn't changed.

Perhaps what we really want is a `SetComparison` that can perform `IN` comparisons, and then `IN` would use that? Then I don't think you need to cache the `ListComparison` here (which I assume is being done to avoid recreating the set), and I think the logic in `ListComparison` can remove some `if` blocks.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -693,10 +693,19 @@ public static String toPrintable(@Nullable Object value) {
         private final Type type;
         @Nonnull
         protected final Object comparand;
+        @Nullable
+        protected ListComparison listComparison;
```
alecgrieser, line 697: I suppose I probably should have mentioned this early, but something about this feels wrong. Like, instead of having a `SimpleComparison` have an optional `ListComparison`, it feels like the `IN` comparison should just use a `ListComparison` always (or as much as possible). I'm not actually sure what the right way to fix that is (if the QueryComponent isn't already using a `ListComparison` or something), but I figured I'd mention it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -940,23 +968,42 @@ public ListComparison(@Nonnull Type type, @Nonnull List comparand) {
                 default:
                     throw new RecordCoreException("ListComparison only supports EQUALS, NOT_EQUALS and STARTS_WITH");
             }
-            if (comparand == null || (this.type == Type.IN && comparand.stream().anyMatch(o -> o == null))) {
-                throw new NullPointerException("List comparand is null, or contains null");
+            if (comparand == null) {
+                throw new NullPointerException("List comparand is null");
```
alecgrieser, line 972: I suppose I should have asked about this the first time, but do we actually want this? I don't think a Protobuf message can contain `null`, which I think is why this was written the way it was. Unless there's a reason we're changing this now?

PR #915, opened 2020-04-21
A repro for #482

ScottDugas:
These seem to reliably get the wrong results, although, I definitely
feel like the test could be cleaner


alecgrieser:
These tests have been incorporated into #914, so we could probably close this PR out

PR #916, opened 2020-04-21
Check if #914 fixes the tests in #915

alecgrieser:
This is essentially #914 and #915 merged together. The merge is a little messy and shouldn't be merged because of that. It tweaks the tests in #915 to make them work, but only related to matching, etc.


alecgrieser:
Gonna close this one out in favor of #914, which now incorporates the tests given but doesn't have the bad merge

PR #926, opened 2020-04-29
Resolves #925: Add assertions into record layer

### LINKED ISSUE #925: Validate key/value byte lengths in record layer
FoundationDB has a size limit on keys (10k) and values (100k), but these size limits are enforced at `commit()` time, rather than at the time that the keys and values are inserted.  This has the byproduct that when you get an exception indicating you have exceeded one of these limits, it can be quite difficult to figure out **which** key or value was the problem.

The proposal here is to add a mode into record layer in which operations on keys and values are validated immediately to see if a limit has been exceeded, with an exception being thrown right from that call point rather than deferring to the commit.

scgray:
This change allows `FDBDatabaseFactory` to be configured with a supplier which
can decide whether or not assertions are to be enabled within the record layer
API.  This supplier is polled each time an `FDBRecordContext` is created and,
thus, all operations from the context will have assertions checked (or not)
as it is used.  The initial set of assertions are around validating key and
value byte sizes.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -54,7 +54,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Feature Add assertions into record layer [(Issue #925)](https://github.com/FoundationDB/fdb-record-layer/issues/925)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Breaking change** Change 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -54,7 +54,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature Add assertions into record layer [(Issue #925)](https://github.com/FoundationDB/fdb-record-layer/issues/925)
+* **Feature** Add key/value size enforcement assertions to FDBRecordContext [(Issue #925)](https://github.com/FoundationDB/fdb-record-layer/issues/925)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Breaking change** Change 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
@@ -748,10 +748,13 @@ public class FDBDatabase {
      * @return newly created transaction
      */
     private Transaction createTransaction(Executor executor, @Nullable StoreTimer storeTimer, @Nullable Map<String, String> mdcContext, boolean transactionIsTraced) {
+        final boolean enableAssertions = factory.areAssertionsEnabled();
         Transaction transaction = database.createTransaction(executor);
 
         if (storeTimer != null) {
-            transaction = new InstrumentedTransaction(storeTimer, transaction);
+            transaction = new InstrumentedTransaction(storeTimer, transaction, enableAssertions);
+        } else if (enableAssertions) {
+            transaction = new InstrumentedTransaction(new FDBStoreTimer(), transaction, enableAssertions);
         }
 
         if (transactionIsTraced) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
@@ -31,7 +31,6 @@ import com.apple.foundationdb.record.RecordCoreRetriableTransactionException;
 import com.apple.foundationdb.record.ResolverStateProto;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
-import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.keyspace.LocatableResolver;
 import com.apple.foundationdb.record.provider.foundationdb.keyspace.ResolverResult;
 import com.apple.foundationdb.record.provider.foundationdb.keyspace.ScopedValue;
@@ -410,7 +409,7 @@ public class FDBDatabase {
         openFDB();
         final boolean transactionIsTraced = transactionIsTracedSupplier.get();
         final Executor executor = newContextExecutor(contextConfig.getMdcContext());
-        final Transaction transaction = createTransaction(executor, contextConfig.getTimer(), contextConfig.getMdcContext(), transactionIsTraced);
+        final Transaction transaction = createTransaction(contextConfig, executor, transactionIsTraced);
 
         FDBRecordContext context = new FDBRecordContext(this, transaction, contextConfig, transactionIsTraced);
         final WeakReadSemantics weakReadSemantics = context.getWeakReadSemantics();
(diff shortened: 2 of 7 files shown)


docs/ReleaseNotes.md
```diff
@@ -54,7 +54,7 @@ The `FDBDatabase::getReadVersion()` method has been replaced with the `FDBRecord
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Feature Add assertions into record layer [(Issue #925)](https://github.com/FoundationDB/fdb-record-layer/issues/925)
```
alecgrieser, line 57:
```suggestion
* **Feature** Add assertions into record layer [(Issue #925)](https://github.com/FoundationDB/fdb-record-layer/issues/925)
```

Hmm, this also feels like it should be more descriptive. Like, "added assertions about key and value size" or something. I think keeping it more general ("assertions") is probably fine in the code, though, to future proof it for more kinds of assertions.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -121,6 +122,13 @@
      */
     @Nonnull
     private Supplier<Boolean> transactionIsTracedSupplier = LOGGER::isTraceEnabled;
+    /**
+     * A supplier that, when it returns true, enables assertions/sanity checks on certain operations. This supplier is
+     * polled upon the creation of a new {@code FDBRecordContext} and all subsequent operations on that context will
+     * (or will not) enforce assertions based upon the return value.
+     */
+    @Nonnull
+    private Supplier<Boolean> enableAssertionsSupplier = DEFAULT_ENABLE_ASSERTIONS_SUPPLIER;
```
alecgrieser, line 131: Hmm, I'm not sure this is necessarily the best way to determine whether or not to enable assertions or not. In particular, it wouldn't let you do something like set it for a single transaction (if, for example, if one wanted to turn on "debugging" for a single transaction that both, like, traced it and enabled assertions). Like, the other mechanism I can think of would be to add a new field to the `FDBRecordContextConfig` (and, in theory, one could want that for tracing). I suppose it's possible that we could go with this "for now", see how it goes, and add a new mechanism, if we don't like the way that that works in practice.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -748,10 +748,13 @@ public Transaction createTransaction(Executor executor, @Nullable Map<String, St
...
      */
     private Transaction createTransaction(Executor executor, @Nullable StoreTimer storeTimer, @Nullable Map<String, String> mdcContext, boolean transactionIsTraced) {
+        final boolean enableAssertions = factory.areAssertionsEnabled();
         Transaction transaction = database.createTransaction(executor);
 
         if (storeTimer != null) {
-            transaction = new InstrumentedTransaction(storeTimer, transaction);
+            transaction = new InstrumentedTransaction(storeTimer, transaction, enableAssertions);
+        } else if (enableAssertions) {
+            transaction = new InstrumentedTransaction(new FDBStoreTimer(), transaction, enableAssertions);
```
alecgrieser, line 757: A new `FDBStoreTimer` is a little bit of a weird choice if `StoreTimer` is null, rather than sending `null` down to the `InstrumentedTransaction` (and, um, I guess not instrumenting it). Maybe it's preferable to introducing possible NPEs from that class, though? Not sure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
```diff
@@ -38,70 +39,74 @@
...
     public void clearRangeStartsWith(byte[] prefix) {
-        underlying.clearRangeStartsWith(prefix);
+        underlying.clearRangeStartsWith(checkKey(prefix));
         timer.increment(FDBStoreTimer.Counts.DELETES);
     }
 
     @Override
     public void mutate(MutationType opType, byte[] key, byte[] param) {
-        underlying.mutate(opType, key, param);
+        underlying.mutate(opType, checkKey(key), param);
```
alecgrieser, line 109:
```suggestion
        underlying.mutate(opType, checkKey(key), checkValue(param));
```

alecgrieser: Technically, the versionstamp operations can _shrink_ the key and value sizes, and...I'm not actually sure whether a 100,004 byte value where the last four bytes are a versionstamp offset as part of a `SET_VERSIONSTAMPED_VALUE` operation are actually allowed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
```diff
@@ -38,70 +39,74 @@
...
     @Override
     public void addWriteConflictKey(byte[] key) {
-        underlying.addWriteConflictKey(key);
+        underlying.addWriteConflictKey(checkKey(key));
     }
 
     @Override
     public void set(byte[] key, byte[] value) {
-        underlying.set(key, value);
+        underlying.set(checkKey(key), checkValue(value));
```
alecgrieser, line 74: AFAICT, key and value size limits are only imposed on `set` operations (and `mutate`). I think clear, clear range, the conflict range operations, and the read operations actually allow larger keys. I've validated this by playing around with the Python API in its shell, but maybe it's worth getting better validation.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseTest.java
```diff
@@ -489,4 +489,42 @@ public void performNoOpAgainstFakeCluster() throws IOException {
...
+        try {
+            factory.clear();
+
+            // enabling assertions causes checks to happen in record layer code
+            factory.setEnableAssertionsSupplier(() -> true);
+            try (FDBRecordContext context = database.openContext()) {
+                assertThrows(exception, () -> consumer.accept(context));
+            }
+        } finally {
+            factory.setEnableAssertionsSupplier(() -> false);
```
alecgrieser, line 526: Seems like this should set it to the original asserterions supplier instead of `() -> false`. At the moment, those are the same same thing, but if we ever started running our tests with assertions on for everything (for example), I don't think we'd want this test to mess with that.

docs/ReleaseNotes.md
scgray: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
scgray: Yeah, it just bothers me that we have this log everywhere of `if (storeTimer != null)`.  I took at stab at creating a `NullStoreTimer` implementation so we can enforce that it is never nullable, but it gets into problems because it doesn't really extend into `NullFDBStoreTimer`. Anyway, I updated this to allow for null :(

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scgray: Yeah, you are absolutely right. I had forgotten about that object. I have moved the configuration there.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
scgray: But, just because you can, should you?  Is there a case where this is legitimate and the server can make sense of the trailing bytes?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseTest.java
scgray: Good point, but I removed this manner of configuration anyway.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
alecgrieser: There's one only mediumly contrived example that I can think of right now: if you have a key of exactly 10k bytes, and you want to get the next key in the database, you might perform a scan from `key + \x00`, which would have length 10,001. In theory, if you had the right logic, you can notice that it's hit the limit and only scan from `key[:9999] + str(key[9999] + 1)` (or something like that)

scgray:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -50,6 +50,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(FDBDatabaseFactory.class);
 
     protected static final Function<FDBLatencySource, Long> DEFAULT_LATENCY_INJECTOR = api -> 0L;
+    protected static final Supplier<Boolean> DEFAULT_ENABLE_ASSERTIONS_SUPPLIER = () -> false;
```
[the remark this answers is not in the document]
scgray: Oopsy.  Thanks for catching that.  Removed.

alecgrieser:
```
com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerSimpleTest > testConfigLoader() FAILED
    org.opentest4j.AssertionFailedError: expected: <95> but was: <94>
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
        at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerSimpleTest.testConfigLoader(OnlineIndexerSimpleTest.java:876)
```

Hmm, don't think we've seen this one before

scgray:
opened: https://github.com/FoundationDB/fdb-record-layer/issues/936
test this please

PR #927, opened 2020-05-05
Draft of CentralStore (for discussion)

nblintao:
This is a draft of a centralized record store and how it can be used (see `CentralStoreDemo`).

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
@@ -1,0 +1,114 @@
+/*
+ * CentralStore.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2020 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.foundationdb.centralstore;
+
+import com.apple.foundationdb.record.RecordMetaData;
+import com.apple.foundationdb.record.RecordMetaDataBuilder;
+import com.apple.foundationdb.record.RecordMetaDataProto;
+import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
+import com.apple.foundationdb.record.provider.foundationdb.MetaDataProtoEditor;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.function.Function;
+
+/**
+ * TODO: CentralStore
+ */
+public class CentralStore {
+    @Nonnull
+    private static final CentralStore INSTANCE = new CentralStore();
+
+    @Nullable
+    private CentralStoreConfiguration config;
+    @Nullable
+    private RecordMetaData cachedMetaData;
+
+    @Nonnull
+    public static CentralStore instance() {
+        if (!isConfigured()) {
+            throw new CentralStoreNotSetUpException("CentralStore is not configured");
+        }
+        return INSTANCE;
+    }
+
+    private CentralStore() {}
+
+    public static CentralStoreConfiguration newConfiguration() {
+        return new CentralStoreConfiguration();
+    }
+
+    static void configure(@Nonnull CentralStoreConfiguration config) {
+        INSTANCE.config = config;
+    }
+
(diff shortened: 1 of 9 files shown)


fdb-record-layer-core/src/main/proto/central_store.proto
```diff
@@ -0,0 +1,45 @@
...
+
+message StoreInfo {
+    // TODO: This may be changed to repeated tuple elements. (A tuple element is similar to a "oneof" of the types in TupleFieldsProto.)
+    optional bytes store_subspace = 1;
+    repeated IndexStateEntry index_states = 2;
+    optional uint64 metadata_version = 3;
+
+    message IndexStateEntry {
+        // Subspace tuple key
+        // TODO: The key should be a tuple element.
```
ScottDugas, line 35: the subspaceKey on the index is always a single tuple element right, never a full tuple (which I guess here could be a single tuple element, containing a nested tuple)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStoreTypeStoreInfo.java
```diff
@@ -0,0 +1,126 @@
...
+        return INSTANCE;
+    }
+
+    private CentralStoreTypeStoreInfo() {}
+
+    // Given the subspace of the store, and the information type, return whether or not it should be tracked in
+    // CentralStoreTypeStoreInfo.
+    public static CentralStoreTypeStoreInfo config(@Nullable BiFunction<Subspace, InfoType, Boolean> enable,
+                                                   @Nullable List<Index> indexes) {
+        INSTANCE.enable = enable == null ? (v1, v2) -> true : enable;
```
ScottDugas, line 81: From an API standpoint, it would probably be cleaner to have `enable` be `@Nonnull` and add a static `ALWAYS_ENABLED`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
```diff
@@ -0,0 +1,114 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.function.Function;
+
+/**
+ * TODO: CentralStore
+ */
+public class CentralStore {
+    @Nonnull
+    private static final CentralStore INSTANCE = new CentralStore();
```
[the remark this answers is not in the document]
scgray: Yes, definitely, we've gotten burned by this mistake in the past (we == me, with the reverse directory layer being static a singleton).  It may make the most sense to have a `FDBDatabase` own it's instance of the central store, however Scott D makes a good argument that we may even just want to let the client manage its own central store(s).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Nullable
+    private CentralStoreConfiguration config;
```
scgray, line 43: I will try not to pick any more nits after this one :-)   but, probably `private final`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
```diff
@@ -0,0 +1,114 @@
...
+        RecordMetaDataProto.MetaData.Builder protoBuilder = RecordMetaDataProto.MetaData.newBuilder();
+        config.types.forEach(type ->
+                MetaDataProtoEditor.addRecordType(protoBuilder, type.getTypeDescriptor(),
+                        // Primary key.
+                        Key.Expressions.concat(Key.Expressions.recordType(), type.getPrimaryFields()))
+        );
+        RecordMetaDataBuilder builder = RecordMetaData.newBuilder().setRecords(protoBuilder.build(), true);
+        config.types.forEach(type ->
+                type.getIndexes().forEach(
+                        index -> builder.addIndex(type.getTypeDescriptor().getName(), index)
```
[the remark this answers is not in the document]
scgray: Would it make sense to use the same approach that is used for tests, but allowing the configuration to instead supply a hook that modifies the `RecordMetaDataBuilder`, then after applying the hook, we check for some invariants we expect to be in place (required record types / indexes)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/centralstore/CentralStore.java
```diff
@@ -0,0 +1,114 @@
...
+import com.apple.foundationdb.record.provider.foundationdb.MetaDataProtoEditor;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.function.Function;
+
+/**
+ * TODO: CentralStore
+ */
+public class CentralStore {
```
scgray, line 38: As elucidated below, maybe this should be something more like `CentralStoreFactory`, in which you provide a `KeySpacePath` that identifies where it is along with an `FDBRecordContext` and it returns an instance of a `CentralStore`.  I think, also, the `CentralStore` should internally wrap an instance of the record store itself and present API's to the rest of the record layer to do things like update the metadata about an indexes state, e.g.:

    centralStore.setIndexState(Index index, IndexState state);

(probably an oversimplification here, but you get the idea). 

The main point being that the central store has a clean API for all of the users that the record layer internally needs to maintain metadata about record stores, then if an outside application needs to do more, it can ask for the underlying record store.

alecgrieser:

PR #935, opened 2020-05-07
Resolves #934: Retry OnlineIndexer when it hits RecordStoreStaleMetaDataVersionException

### LINKED ISSUE #934: Retry OnlineIndexer when it hits RecordStoreStaleMetaDataVersionException

nblintao:
This needs to be patched to 2.8.118.0. I manually created a branch called `fdb-record-layer-2.8.118.0` from tag `2.8.118.0`, and then used it as base. Is it the correct way to do it?

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -413,8 +413,10 @@ public class OnlineIndexer implements AutoCloseable {
                 } else {
                     int currTries = tries.getAndIncrement();
                     FDBException fdbE = getFDBException(e);
-                    if (currTries < config.maxRetries && fdbE != null && lessenWorkCodes.contains(fdbE.getCode())) {
-                        if (handleLessenWork != null) {
+                    if (currTries < config.maxRetries && (
+                            (fdbE != null && lessenWorkCodes.contains(fdbE.getCode())) ||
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -413,8 +413,10 @@ public void close() {
                 } else {
                     int currTries = tries.getAndIncrement();
                     FDBException fdbE = getFDBException(e);
-                    if (currTries < config.maxRetries && fdbE != null && lessenWorkCodes.contains(fdbE.getCode())) {
-                        if (handleLessenWork != null) {
+                    if (currTries < config.maxRetries && (
+                            (fdbE != null && lessenWorkCodes.contains(fdbE.getCode())) ||
+                            (e instanceof RecordStoreStaleMetaDataVersionException))) {
```
alecgrieser, line 418: I suppose it's worth noting that this is only the right thing to do if given a `RecordMetaDataProvider` that knows to refresh (somehow) on this error, or this will just spin, hitting the same error each time. I also don't know if the `RecordStoreStaleMetaDataVersionException` will be wrapped in a `CompletionException` at this point. I don't...think so, but I'm not certain.

nblintao: Yeah. It sounds better to retry on the developer side to start a new `OnlineIndexer` instead. Closing this PR.

PR #948, opened 2020-05-19
Fixes #947: TupleKeyCountTree.addPrefixChild only ever adds one child

### LINKED ISSUE #947: TupleKeyCountTree.addPrefixChild only ever adds one child
It's confused about the parent key vs the prefix child's.

MMcM:
Use the prefix's bytes as the unique child key.

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/clientlog/TupleKeyCountTree.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/clientlog/TupleKeyCountTree.java
@@ -157,13 +157,13 @@ public class TupleKeyCountTree {
     }
 
     @Nonnull
-    protected TupleKeyCountTree newPrefixChild(@Nonnull byte[] bytes, @Nonnull Object prefix) {
-        return newChild(bytes, prefix);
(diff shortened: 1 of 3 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/clientlog/TupleKeyCountTree.java
```diff
@@ -152,12 +152,12 @@ private synchronized void addInternal(@Nonnull List<Object> items, int itemPosit
     @Nonnull
     public TupleKeyCountTree addPrefixChild(@Nonnull Object prefix) {
         count++;
-        return children.computeIfAbsent(bytes, b -> newPrefixChild(prefix));
+        final byte[] prefixBytes = Tuple.from(prefix).pack();
+        return children.computeIfAbsent(bytes, b -> newPrefixChild(prefixBytes, prefix));
```
alecgrieser, line 156:
```suggestion
        return children.computeIfAbsent(prefixBytes, b -> newPrefixChild(prefixBytes, prefix));
```
I think? I'm not actually sure, but I thought that's what you were alluding to in the issue text.

MMcM: Yes. I also renamed the formal parameters to not conflict with the field to cut down on such confusion.

PR #955, opened 2020-05-22
Fixes #952: Local version cache is keyed only by Tuple (i.e., primary key) not subspace

### LINKED ISSUE #952: Local version cache is keyed only by Tuple (i.e., primary key) not subspace
To allow for RYW behavior with versionstamps, there is a cache kept from record primary keys to the current versions. These are then keyed by the primary key of the record associated with them. However, this cache is global to the transaction rather than per record store or per subspace, etc., so if two stores are opened in the same transaction, and one writes a record with one primary key to one store and then tries to read a record with the same primary key from a different store, then the version of the first record is returned instead of the version for the second one. (Yikes!)

alecgrieser:
This changes the local version cache to be keyed by `byte[]` rather than by `Tuple`. Then the exact key (in the database) is used to represent the cache entry rather than the record's primary key. This allows multiple stores to be opened in the same transaction and for them to interact with versioned records with the same primary key in each store without having cache entries from one store pollute the cache entries of the other.

This meant changing the API. The previous API was just wrong, in that there was absolutely no way to protect oneself from the bug identified in #952, so it has just been removed (despite the fact that this violates are claims of API stability). These APIs probably should have been marked `INTERNAL` anyway, but I believe they pre-date the API stability annotations, which is why they weren't. I put this against master rather than 2.9 as it is a bug fix that could bite someone who doesn't want to upgrade to 2.9 (despite the API breaking), but perhaps it should be rebased onto that branch if 2.9 is soon.

This fixes #952.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
@@ -122,6 +122,7 @@ public class VersionIndexTest extends FDBTestBase {
     private boolean splitLongRecords;
     private FDBDatabase fdb;
     private Subspace subspace;
+    private Subspace subspace2;
 
     @BeforeEach
     public void setUp() {
@@ -130,9 +131,11 @@ public class VersionIndexTest extends FDBTestBase {
         }
         if (subspace == null) {
             subspace = fdb.run(context -> TestKeySpace.getKeyspacePath("record-test", "unit", "indexTest", "version").toSubspace(context));
+            subspace2 = fdb.run(context -> TestKeySpace.getKeyspacePath("record-test", "unit", "indexTest", "version2").toSubspace(context));
         }
         fdb.run(context -> {
             FDBRecordStore.deleteStore(context, subspace);
+            FDBRecordStore.deleteStore(context, subspace2);
             return null;
         });
         formatVersion = FDBRecordStore.MAX_SUPPORTED_FORMAT_VERSION;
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
```diff
@@ -1059,13 +1295,13 @@ public void maxEverVersionWithinTransaction(int testFormatVersion, boolean testS
...
-            recordStore.saveRecord(record3);
+            FDBRecordVersion version3 = recordStore.saveRecord(record3).getVersion();
+            assertNotNull(version3);
             FDBRecordVersion version4 = getBiggerVersion(expectedMaxVersion);
-            recordStore.saveRecord(record4, version4);
+            FDBStoredRecord<?> storedRecord4 = recordStore.saveRecord(record4, version4);
 
             context.commit();
             assertNotNull(context.getVersionStamp());
-            FDBRecordVersion version3 = FDBRecordVersion.complete(context.getVersionStamp(), context.getLocalVersion(Tuple.from(record3.getRecNo())).get());
```
alecgrieser, line 1068: Wait, this is wrong

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
```diff
@@ -1077,11 +1313,11 @@ public void maxEverVersionWithinTransaction(int testFormatVersion, boolean testS
             MySimpleRecord record6 = MySimpleRecord.newBuilder().setRecNo(1455L).build();
             FDBRecordVersion version5 = getBiggerVersion(expectedMaxVersion);
             recordStore.saveRecord(record5, version5);
-            recordStore.saveRecord(record6);
+            FDBRecordVersion version6 = recordStore.saveRecord(record6).getVersion();
+            assertNotNull(version6);
 
             context.commit();
             assertNotNull(context.getVersionStamp());
-            FDBRecordVersion version6 = FDBRecordVersion.complete(context.getVersionStamp(), context.getLocalVersion(Tuple.from(record6.getRecNo())).get());
```
alecgrieser, line 1084: as is this

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
alecgrieser: A previous version of this had some confusion as to which versions were complete and which were incomplete. The newer version I think doesn't have that any more.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/VersionIndexTest.java
alecgrieser: Same as the other comment.

alecgrieser:
Okay, rebased against 2.9 and changed the base of the PR given that this changes the API. We should target a swift 2.9 release, though, I think.

alecgrieser:
```
com.apple.foundationdb.record.provider.foundationdb.RankIndexTest > repeatedRankQuery() FAILED
    org.opentest4j.AssertionFailedError: expected: <[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus]> but was: <[achilles, patroclus, hector, patroclus, hector, hector, achilles, hector, achilles, hector, patroclus, achilles, hector]>
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
        at com.apple.foundationdb.record.provider.foundationdb.RankIndexTest.repeatedRankQuery(RankIndexTest.java:781)
```
Looks like #482, for which the fix is only on master at the moment.

test this please

PR #957, opened 2020-06-03
Resolves #956: More efficient encryption via Amazon Corretto Crypto Provider

### LINKED ISSUE #956: More efficient encryption via Amazon Corretto Crypto Provider
`TransformedRecordSerializerJCE.decrypt()` tends to use a non-trivial amount of CPU when records are encrypted at rest.  The Amazon Corretto Crypto Provider (https://github.com/corretto/amazon-corretto-crypto-provider) provides more efficient encryption implementations than standard JCA/JCE provides.

scgray:
I realize that this may need discussion or may not be the solution we choose to go with, but I'm opening this PR to at least start the conversation.

### DIFF
--- a/ACKNOWLEDGEMENTS
+++ b/ACKNOWLEDGEMENTS
@@ -4,7 +4,7 @@ material, the use of which is hereby acknowledged.
 
 _____________________
 
-The Apache Software Foundation (Apache Commons Lang)
+The Apache Software Foundation (Apache Commons Lang, Amazon Corretto Crypto Provider)
 
   Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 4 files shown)


scgray:
Thanks for the comment Mike. Yes, I misread https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/README.md on first pass and didn't realize there are more steps to getting it in the configuration, so I'll do as you suggest and try to make some tests that compare between the two.  I think, overall, the main question set out to ask was whether or not there was opposition to adopting the library, which it seems not, but let me know if you think otherwise.

scgray:
Thanks @Scottmitch.  I'll close this out here, and address in upstream projects. Thanks!

PR #961, opened 2020-06-11
Fixes #960 Meta Conflict Range Additions hotspot during large operations.

### LINKED ISSUE #960: Meta Conflict Range Additions hotspot during large operations.
This is a profile from an update that called updateSecondaryIndexes 3K times (i.e. updating 3K records).  The conflict additions hotspot across JNI in this use case. 

Screen Shot 2020-06-11 at 2 17 25 PM

jleach4:
Screen Shot 2020-06-11 at 2 17 25 PM

Hotspots while updating 3K records.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -250,6 +250,13 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     @Nonnull
     private final Cache<Tuple, FDBRawRecord> preloadCache;
 
+    private boolean recordsReadConflict;
+
+    private boolean storeStateReadConflict;
+
+    @Nonnull
+    private final Set<String> indexStateReadConflicts = ConcurrentHashMap.newKeySet(8);
+
     @SuppressWarnings("squid:S00107")
     protected FDBRecordStore(@Nonnull FDBRecordContext context,
                              @Nonnull SubspaceProvider subspaceProvider,
@@ -2771,6 +2778,10 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
      * Add a read conflict key for all records.
      */
     private void addRecordsReadConflict() {
+        if (recordsReadConflict) {
+            return;
+        }
+        recordsReadConflict = true;
         Transaction tr = ensureContextActive();
         byte[] recordKey = getSubspace().pack(Tuple.from(RECORD_KEY));
         tr.addReadConflictRange(recordKey, ByteArrayUtil.strinc(recordKey));
@@ -2784,6 +2795,11 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         if (!getRecordMetaData().hasIndex(indexName)) {
             throw new MetaDataException("Index " + indexName + " does not exist in meta-data.");
         }
+        if (indexStateReadConflicts.contains(indexName)) {
+            return;
+        } else {
+            indexStateReadConflicts.add(indexName);
+        }
         Transaction tr = ensureContextActive();
         byte[] indexStateKey = getSubspace().pack(Tuple.from(INDEX_STATE_SPACE_KEY, indexName));
         tr.addReadConflictKey(indexStateKey);
@@ -2793,6 +2809,10 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
      * Add a read conflict key for the whole record store state.
      */
     private void addStoreStateReadConflict() {
+        if (storeStateReadConflict) {
+            return;
+        }
+        storeStateReadConflict = true;
         Transaction tr = ensureContextActive();
         byte[] indexStateKey = getSubspace().pack(Tuple.from(INDEX_STATE_SPACE_KEY));
         tr.addReadConflictRange(indexStateKey, ByteArrayUtil.strinc(indexStateKey));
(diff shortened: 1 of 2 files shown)


nschiefer:
@alecgrieser this looks good to me, but this is a part of the code base that I don't really know much about, so I requested a review from you.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2771,6 +2781,10 @@ public Range getUnbuiltRange() {
      * Add a read conflict key for all records.
      */
     private void addRecordsReadConflict() {
+        if (recordsReadConflict) {
+            return;
+        }
+        recordsReadConflict = true;
```
alecgrieser, line 2787: This method is only called if one wants to add a conflict range over all records, so it should essentially never be called. I'm surprised it's showing up in your profiling. Maybe it doesn't hurt to also optimize it in this way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2784,6 +2798,14 @@ private void addIndexStateReadConflict(@Nonnull String indexName) {
         if (!getRecordMetaData().hasIndex(indexName)) {
             throw new MetaDataException("Index " + indexName + " does not exist in meta-data.");
         }
+        if (indexStateReadConflicts == null) {
+            indexStateReadConflicts = new HashSet<>(8);
```
alecgrieser, line 2802: Should this be using a `ConcurrentHashMap`? I think multiple threads can access it at once. Because it's safe to set this multiple times for the same index, having this throw away the full hash map in certain race conditions seems fine, but I'm still worried about messing up internal state in the hash map if there are multiple accessors.

jleach4: Yeah, I debated.  I will modify.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2784,6 +2798,14 @@ private void addIndexStateReadConflict(@Nonnull String indexName) {
         if (!getRecordMetaData().hasIndex(indexName)) {
             throw new MetaDataException("Index " + indexName + " does not exist in meta-data.");
         }
+        if (indexStateReadConflicts == null) {
+            indexStateReadConflicts = ConcurrentHashMap.newKeySet(8);
```
alecgrieser, line 2802: Is there a reason this doesn't just set it to a `ConcurrentHashMap` in the constructor? We'll need to access this whenever we access an index, which will be essentially most things that interact with the store, so it seems like just initializing this in the constructor would be simpler and not much more expensive?

jleach4: Sure, let me fix.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -250,6 +251,15 @@
     @Nonnull
     private final Cache<Tuple, FDBRawRecord> preloadCache;
 
+    @Nonnull
+    private boolean recordsReadConflict;
+
+    @Nonnull
+    private boolean storeStateReadConflict;
+
+    @Nullable
```
alecgrieser, line 260:
```suggestion
    @Nonnull
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -250,6 +251,15 @@
     @Nonnull
     private final Cache<Tuple, FDBRawRecord> preloadCache;
 
+    @Nonnull
```
alecgrieser, line 254:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -250,6 +251,15 @@
...
+    private boolean recordsReadConflict;
+
+    @Nonnull
```
alecgrieser, line 257:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -250,6 +251,15 @@
...
+    private boolean storeStateReadConflict;
+
+    @Nullable
+    private Set<String> indexStateReadConflicts = ConcurrentHashMap.newKeySet(8);
```
alecgrieser, line 261:
```suggestion
    private final Set<String> indexStateReadConflicts = ConcurrentHashMap.newKeySet(8);
```

alecgrieser:
Ah, looks like the build complained about an unused import:

```
[ant:checkstyle] [ERROR] /code/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java:118:8: Unused import - java.util.HashSet. [UnusedImports]

> Task :fdb-record-layer-core:checkstyleMain FAILED
> Task :fdb-extensions:printInstructionsOnRunningWithHtmlOutput

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':fdb-record-layer-core:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///code/fdb-record-layer-core/.out/reports/checkstyle/main.html
  Checkstyle files with violations: 1
  Checkstyle violations by severity: [error:1]
```

PR #975, opened 2020-06-17
Fixes #974: OrElseCursor doesn't properly support continuations.

### LINKED ISSUE #974: OrElseCursor doesn't properly support continuations
An `OrElseCursor` is built out of an "inner" cursor and and an "else" cursor. If the inner cursor is non-empty, it returns values from the inner cursor; otherwise, it returns the values from the else cursor.

Currently, the `OrElseCursor` uses pass-through style continuation handling. However, this loses information (specifically: which branch the cursor is using). As a result, if the cursor chooses the else branch, there is no way to continue the cursor's operation.

To resolve this, we probably need to give `OrElseCursor` a structured protobuf continuation, like `FlatMapPipelinedCursor`.

nschiefer:
Use a new structured continuation object and protobuf serialization to record the complete state of the `OrElseCursor`, including which branch was chosen. Rewrite `OrElseCursor`'s logic in a simpler state-machine style, which makes it easy to resume the cursor before it has made a decision about which branch to take.

To avoid problems like we saw with the `FlatMapPipelinedCursor`'s continuations, deprecate the `RecordCursor.orElse()` fluent method and replace it with a static method that interprets the continuation.

Because this changes the continuation format, this is a breaking change.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
@@ -781,10 +781,32 @@ public interface RecordCursor<T> extends AutoCloseable, Iterator<T> {
      * @param func function to be called if the cursor is empty to give another source of records
      * @return a new cursor that returns the same records as this cursor 
      * or the result of {@code func} if this cursor does not produce any records
+     * @deprecated because it does not support continuations and is easy to misuse.
+     *             Use {@link #flatMapPipelined(Function, BiFunction, byte[], int)} instead.
      */
+    @API(API.Status.DEPRECATED)
+    @Deprecated
     @Nonnull
     default RecordCursor<T> orElse(@Nonnull Function<Executor, RecordCursor<T>> func) {
         return new OrElseCursor<>(this, func);
+    }
+
+    /**
+     * Get a new cursor that substitutes another cursor if a given inner cursor is empty.
+     * If a nonnull continuation is provided, it must come from a cursor that is structurally identical to the
+     * cursor being defined.
+     * @param innerFunc a function that takes a continuation and creates an inner cursor
+     * @param elseFunc a function to be called if the cursor is empty to give another source of records
+     * @param continuation a continuation (from a cursor with the same structure) to resume where it left off
+     * @param <T> the type of result returned by the cursor
+     * @return a new cursor that returns the same records as the generated inner cursor
+     * or the result of {@code func} if this cursor does not produce any records
+     */
+    @Nonnull
+    static <T> RecordCursor<T> orElse(@Nonnull Function<byte[], ? extends RecordCursor<T>> innerFunc,
+                                  @Nonnull BiFunction<Executor, byte[], ? extends RecordCursor<T>> elseFunc,
+                                  @Nullable byte[] continuation) {
+        return new OrElseCursor<>(innerFunc, elseFunc, continuation);
     }
 
     /**
@@ -861,6 +883,24 @@ public interface RecordCursor<T> extends AutoCloseable, Iterator<T> {
     @Nonnull
     static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull CompletableFuture<T> future) {
         return new FutureCursor<>(executor, future);
+    }
+
+    /**
+     * Get a new cursor that has the contents of the given future as its only record.
+     * If the continuation is nonnull, return an empty cursor since a {@link FutureCursor} returns only a single record.
+     * @param executor an executor for executing the future
+     * @param future a future that completes to the only element of the cursor
+     * @param continuation a continuation from a future cursor to determine whether the cursor has a single element or no elements
+     * @param <T> the result type of the future
+     * @return a new cursor producing the contents of {@code future} if the continuation is nonnull and an empty cursor otherwise
+     */
+    @Nonnull
+    static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull CompletableFuture<T> future, @Nullable byte[] continuation) {
+        if (continuation == null) {
+            return new FutureCursor<>(executor, future);
+        } else {
+            return RecordCursor.empty();
+        }
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
@@ -21,15 +21,22 @@
 package com.apple.foundationdb.record.cursors;
 
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.RecordCursorContinuation;
+import com.apple.foundationdb.record.RecordCursorProto;
 import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordCursorVisitor;
+import com.apple.foundationdb.tuple.ByteArrayUtil2;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.NoSuchElementException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 
 /**
@@ -42,7 +49,8 @@ public class OrElseCursor<T> implements RecordCursor<T> {
     private final RecordCursor<T> inner;
     @Nonnull
     private final Function<Executor, RecordCursor<T>> func;
-    private boolean first;
+    @Nonnull
+    private RecordCursorProto.OrElseContinuation.State state;
     @Nullable
     private RecordCursor<T> other;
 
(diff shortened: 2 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
```diff
@@ -781,12 +781,23 @@ public boolean isLimitReached() {
...
+     *             Use {@link #flatMapPipelined(Function, BiFunction, byte[], int)} instead.
      */
+    @API(API.Status.DEPRECATED)
+    @Deprecated
     @Nonnull
     default RecordCursor<T> orElse(@Nonnull Function<Executor, RecordCursor<T>> func) {
         return new OrElseCursor<>(this, func);
     }
 
+    @Nonnull
```
alecgrieser, line 794: These two new methods could probably use javadoc comments

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
```diff
@@ -154,4 +214,76 @@ public boolean accept(@Nonnull RecordCursorVisitor visitor) {
         }
         return visitor.visitLeave(this);
     }
+
+    private enum State {
```
alecgrieser, line 218: Given that this is private anyway, is there a reason to not use the proto-generated enum?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
```diff
@@ -57,7 +65,44 @@
     public OrElseCursor(@Nonnull RecordCursor<T> inner, @Nonnull Function<Executor, RecordCursor<T>> func) {
```
alecgrieser, line 65: Should this constructor be deprecated, too?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
```diff
@@ -57,7 +65,44 @@
...
         this.inner = inner;
         this.func = func;
-        this.first = true;
+        this.state = State.UNDECIDED;
+    }
+
+    public OrElseCursor(@Nonnull Function<byte[], ? extends RecordCursor<T>> innerFunc,
```
alecgrieser, line 71: This feels like it should be `INTERNAL`, but I could be wrong

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
```diff
@@ -66,22 +111,37 @@ public OrElseCursor(@Nonnull RecordCursor<T> inner, @Nonnull Function<Executor,
...
+            case USE_OTHER:
+                innerFuture = other.onNext();
+                break;
+            case UNDECIDED:
+                innerFuture = inner.onNext().thenCompose(result -> {
+                    if (result.hasNext()) {
+                        // Inner cursor has produced a value, so we take the inner branch.
+                        state = State.USE_INNER;
+                        return CompletableFuture.completedFuture(result);
+                    } else if (result.getNoNextReason().isOutOfBand()) {
```
alecgrieser, line 128: Okay, trying to think about how this interacts with in-band vs out of band limits. I...think it's right that if the child hits an in-band limit (which would only happen if, say, one had a filter over a row limited cursor, and the inner child of that filter stopped hit its limit prior to it no longer returning results) without hitting a value, then it goes down the else path, but I'm not actually sure if that's right. I guess it's consistent with the prior behavior.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/JoinedRecordPlan.java
```diff
@@ -222,10 +222,14 @@ public JoinedRecordPlan(@Nonnull JoinedRecordType joinedRecordType, @Nonnull Lis
 
     private RecordCursor<EvaluationContext> query(int depth, @Nonnull FDBRecordStore store, @Nonnull EvaluationContext context,
                                                   @Nullable byte[] continuation, @Nonnull ExecuteProperties executeProperties) {
-        RecordCursor<FDBQueriedRecord<Message>> records = queries.get(depth).execute(store, context, continuation, executeProperties);
+        final RecordCursor<FDBQueriedRecord<Message>> records;
         final JoinedType joinedType = joinedTypes.get(depth + 1);
         if (joinedType.constituent.isOuterJoined()) {
-            records = records.orElse(executor -> RecordCursor.fromFuture(executor, CompletableFuture.completedFuture(null)));
+            records = RecordCursor.orElse(innerContinuation -> queries.get(depth).execute(store, context, innerContinuation, executeProperties),
```
alecgrieser, line 228: Given that the continuations for the cursors from the `JoinedRecordPlan` won't be valid any more, it seems like the `planHash` on this class should change (at least--perhaps ideally--if there are any outer joined types in `joinedTypes`).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1012,21 +1014,97 @@ public void testFlatMapReasons() throws Exception {
...
+        // Continue else branch
+        cursor = RecordCursor.orElse(cont -> new FakeOutOfBandCursor<>(RecordCursor.fromList(innerList, cont), 3)
+                .filter(i -> false), orElse, nextResult.getContinuation().toBytes());
+        assertEquals(ImmutableList.of(-4, -5), cursor.asList().join());
+        nextResult = cursor.getNext();
+        assertEquals(RecordCursor.NoNextReason.SOURCE_EXHAUSTED, nextResult.getNoNextReason());
+    }
+
+    @Nonnull
+    private static RecordCursor<Integer> getFilteredFakeOutOfBandCursor(@Nonnull List<Integer> list, int limit, int threshold,
```
alecgrieser, line 1101: The name here should probably reflect the fact that it returns on `OrElse` cursor

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1012,21 +1014,97 @@ public void testFlatMapReasons() throws Exception {
...
+        // Choose the inner branch
+        assertEquals(ImmutableList.of(11, 12), cursor.asList().join());
+        nextResult = cursor.getNext();
+        assertEquals(RecordCursor.NoNextReason.TIME_LIMIT_REACHED, nextResult.getNoNextReason());
+
+        byte[] continuation = nextResult.getContinuation().toBytes();
+        while (continuation != null) {
+            cursor = getFilteredFakeOutOfBandCursor(longList, 3, 10, continuation);
+            nextResult = cursor.getNext();
+            continuation = nextResult.getContinuation().toBytes();
```
alecgrieser, line 1067: It feels like this loop should assert something, even if it's just that it never gets "0" back (so that we know it never hits the else branch), though having it assert about the real contents would be better.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
nschiefer: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
nschiefer: No, except that it was my understanding that we did not idiomatically use protobuf generated classes except when interfacing directly with protobufs. I'd be more than happy to drop this enum.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
nschiefer: Our various analyzers do not understand transitive dependencies, so marking this as "deprecated" creates a warning in the deprecated `RecordCursor.orElse()` method. I can suppress the warning, but I don't actually mind this constructor sticking around.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
nschiefer: Sure, why not.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
nschiefer: I believe that what we have here exactly matches what happened before, and I would like to keep it that way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
alecgrieser: Interesting. I thought using deprecated code in `@Deprecated` methods was generally allowed. If it's not, I think depreciating this constructor and then suppressing deprecation warnings in the deprecated `orElse` method is still the call.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
alecgrieser: Okay, sounds good.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
alecgrieser: I think we probably generally want to keep generated proto out of our API...though we maybe have already broken that rule. But directly interfacing with protos in private methods seems fine to me. This also isn't super high priority

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
nschiefer: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java
nschiefer: Okay, that's fine with me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/JoinedRecordPlan.java
nschiefer: This class is package-private and minimally used so I don't think it's worth the additional code complexity.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
nschiefer: Okay, I changed this to assert about the real contents.

There is an interesting philosophical question here about testing (i.e., should you assert on everything you can think of, or just the bare minimum behavior you are trying to demonstrate). I'm happy to go along, though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/JoinedRecordPlan.java
alecgrieser: Yeah, package-private and we don't seem to be leaking the continuations from that planner, though maybe that's a sign that it shouldn't be `PlanHashable`, though not necessarily for this PR or anything. Okay, that's fine then.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
alecgrieser: I suppose so. It does seem weird to have this loop at all if it's not going to assert on anything (unless the point is just to prove it eventually completes or something)

nschiefer: Right, I was trying to demonstrate that the cursor can indeed be restarted without falling over on continuation parsing exceptions (as happened during testing).

nschiefer:
Got a "request for future version" from `OnlineIndexerBuildValueIndexTest.fiftySplitElements()`, which we already know about.

test this please

PR #978, opened 2020-06-17
Fixes #977: RecordMetaData#recordTypesForIndex is N^2 with Universal index probing...

### LINKED ISSUE #977: RecordMetaData#recordTypesForIndex is N^2 with Universal Index probing.
I have a query with 32 an/or predicates with 8 indexes.  The current planner just spins on CPU on the query hotspotting on RecordMetaData#recordTypesForIndex.

Screen Shot 2020-06-17 at 4 52 36 PM

jleach4:
N^2 shows up with a decent number of indexes and lots of and/or predicate action.

Screen Shot 2020-06-17 at 4 52 36 PM

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
@@ -29,6 +29,7 @@ import com.apple.foundationdb.record.metadata.RecordType;
 import com.apple.foundationdb.record.metadata.SyntheticRecordType;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.LiteralKeyExpression;
+import com.apple.foundationdb.record.util.MapUtils;
 import com.google.protobuf.Descriptors;
 
 import javax.annotation.Nonnull;
@@ -40,6 +41,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Meta-data for Record Layer record stores.
@@ -82,6 +84,7 @@ public class RecordMetaData implements RecordMetaDataProvider {
     @Nullable
     private final KeyExpression recordCountKey;
     private final boolean usesLocalRecordsDescriptor;
+    private final Map<Index, Collection<RecordType>> recordTypesForIndex;
 
     private static final Descriptors.FileDescriptor[] defaultExcludedDependencies = new Descriptors.FileDescriptor[] {
             RecordMetaDataProto.getDescriptor(), RecordMetaDataOptionsProto.getDescriptor(), TupleFieldsProto.getDescriptor()
@@ -118,6 +121,7 @@ public class RecordMetaData implements RecordMetaDataProvider {
         this.usesSubspaceKeyCounter = usesSubspaceKeyCounter;
         this.recordCountKey = recordCountKey;
         this.usesLocalRecordsDescriptor = usesLocalRecordsDescriptor;
+        this.recordTypesForIndex = new ConcurrentHashMap<>();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -372,7 +372,7 @@ public boolean usesSubspaceKeyCounter() {
 
     @Nonnull
     public Collection<RecordType> recordTypesForIndex(@Nonnull Index index) {
-        if (getUniversalIndexes().contains(index)) {
+        if (hasUniversalIndex(index.getName())) {
             return getRecordTypes().values();
         }
         List<RecordType> result = new ArrayList<>();
```
MMcM, line 378: Would it make sense to cache the result of this function back inside the `Index` itself?
For most indexes, it will be the singleton, so that's not much more space and the meta-data has a reasonably long lifetime.

jleach4: I can take a look at @MMcM

jleach4: @MMcM I can cache it back but it seems out of place (universal logic is non-existent on the Index).  Let me know if you need me to do that.

alecgrieser: Alternatively, instead of putting it in the index, it could go in the meta-data, constructed by the builder so that the `Index` doesn't need to have any notion of what types exist in the meta-data

MMcM: Yes, caching in the `RecordMetaData` itself work just fine. There is precedent for caching in the individual meta-data objects with things like `RecordType.recordTypeKey`.

jleach4: @MMcM and @alecgrieser is this still true for an Index?

```
   @Override
    public int hashCode() {
        // Within the context of a single RecordMetaData, this should be sufficient
        return Objects.hash(name, type);
    }
```

alecgrieser: That would still be true, I suppose, though in a single `RecordMetaData`, `name.hashCode()` would also be sufficient, I would think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -118,6 +119,7 @@ protected RecordMetaData(@Nonnull Descriptors.FileDescriptor recordsDescriptor,
         this.usesSubspaceKeyCounter = usesSubspaceKeyCounter;
         this.recordCountKey = recordCountKey;
         this.usesLocalRecordsDescriptor = usesLocalRecordsDescriptor;
+        this.recordTypesForIndex = new HashMap<>();
```
alecgrieser, line 122: Does this need to be a `ConcurrentHashMap`? Alternatively, this could be populated at meta-data build time (for every index), I suppose, and then it's immutable, so concurrency doesn't matter.

jleach4: @alecgrieser Fixing...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -372,25 +375,28 @@ public boolean usesSubspaceKeyCounter() {
...
-            } else if (recordType.getMultiTypeIndexes().contains(index)) {
-                result.add(recordType);
-            }
-        }
-        for (SyntheticRecordType<?> recordType : getSyntheticRecordTypes().values()) {
-            if (recordType.getIndexes().contains(index)) {
-                return Collections.singletonList(recordType);
-            } else if (recordType.getMultiTypeIndexes().contains(index)) {
-                result.add(recordType);
+        return recordTypesForIndex.computeIfAbsent(index, idx -> {
```
alecgrieser, line 378: Should probably be `MapsUtils.computeIfAbsent` to avoid the lockup bug

jleach4: ah yes...

jleach4: fixing...

Issue #936, opened 2020-05-07
OnlineIndexerSimpleTest.testConfigLoader() test failure


scgray:
During a PRB run for https://github.com/FoundationDB/fdb-record-layer/pull/926, this test failed with:

```
org.opentest4j.AssertionFailedError: expected: <95> but was: <94>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerSimpleTest.testConfigLoader(OnlineIndexerSimpleTest.java:876)
```

nblintao:
@jjezra Not same, but it might be related to the issue you found.

MMcM:
The failing condition asserts about the consistency of the state of a running index builder: there is no means of guaranteeing that `getLimit` and `getConfigLoaderInvocationCount` match when two unsynchronized calls are needed to get them.

PR #979, opened 2020-06-18
Resolves #949: Useful system keys should be exposed through a more useful API

### LINKED ISSUE #949: Useful system keys should be exposed through a more useful API
It would be nice to provide some kind of more useful API for exposing system keys through than is currently offered (if the alternative is to have clients roll their own). This probably belongs in the FDB bindings themselves (see https://github.com/apple/foundationdb/issues/3200), but until then, it could be added to `fdb-extensions`.

alecgrieser:
This creates a new set of constants for the system keys that we care about, and then it changes the instances
that were already referencing system keys, and then I added a new class, `FDBSystemOperations`, that allows
a user to use these features.

I'm not...super happy with this API, though. Unlike many operations, it wasn't sufficient to provide a context
for some of these operations, as you might need to set some extra options. But then to provide the user
a way to provide their own retry policies (and things like a timer, etc.), it felt like the least bad way
to pass that information, though we don't use a runner like this in other contexts.

This resolves #949.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
@@ -42,10 +42,6 @@ public class FDBSystemOperations {
         return bytes == null ? null : new String(bytes, StandardCharsets.UTF_8);
     }
 
-    private static <T> T asyncToSync(@Nonnull FDBRecordContext context, @Nonnull CompletableFuture<T> operation) {
-        return context.asyncToSync(FDBStoreTimer.Waits.WAIT_LOAD_SYSTEM_KEY, operation);
-    }
-
     private static <T> T asyncToSync(@Nonnull FDBDatabaseRunner runner, @Nonnull CompletableFuture<T> operation) {
         return runner.asyncToSync(FDBStoreTimer.Waits.WAIT_LOAD_SYSTEM_KEY, operation);
     }
@@ -56,7 +52,7 @@ public class FDBSystemOperations {
      * subsystem will be recruited. This is mostly relevant for
      * <a href="https://apple.github.io/foundationdb/configuration.html#configuring-regions">multi-region configurations</a>,
      * where this value might change if the cluster decides that it needs to fail over to a secondary region. The returned
-     * datacenter ID will be {@code null} if the server's datacenter has not been set.
+     * datacenter ID will be {@code null} if the FDB cluster's datacenter has not been set.
      *
      * <p>
      * Note that this operation must read this information from the database's storage and uses its own transaction.
@@ -76,14 +72,21 @@ public class FDBSystemOperations {
 
     /**
      * Get the primary datacenter of the underlying cluster. This is a blocking version of
-     * {@link #getPrimaryDatacenterAsync(FDBDatabaseRunner)}.
+     * {@link #getPrimaryDatacenterAsync(FDBDatabaseRunner)}. If the FDB cluster's primary datacenter has not
+     * been set, this will return {@code null}.
      *
      * @param runner a runner to use to perform the operation
      * @return the primary datacenter of the database underlying {@code runner}
+     * @see #getPrimaryDatacenterAsync(FDBDatabaseRunner)
      */
     @Nullable
     public static String getPrimaryDatacenter(@Nonnull FDBDatabaseRunner runner) {
         return asyncToSync(runner, getPrimaryDatacenterAsync(runner));
+    }
+
+    @Nonnull
+    private static CompletableFuture<String> getConnectionStringAsyncInternal(@Nonnull FDBRecordContext context) {
+        return context.ensureActive().get(SystemKeyspace.CONNECTION_STR_KEY).thenApply(FDBSystemOperations::nullableUtf8);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
```diff
@@ -0,0 +1,211 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * View of an FoundationDB database used for accessing system and special keys. These are special keys defined within
+ * special system keyspaces that contain information about the underlying FoundationDB cluster, about the client
+ * configuration, or other information available through another API.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class FDBSystemOperations {
```
MMcM, line 38: I agree that these methods have a weird-feeling signature.

How about making `newRunner()` a `FunctionalInterface` implemented by `FDBDatabase` and `FDBRecordContext` and taking one of those?

MMcM (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBTestBase.java
```diff
@@ -67,7 +67,7 @@ public static void setupBlockingInAsyncDetection() {
     public static String createFakeClusterFile(String prefix) throws IOException {
         File clusterFile = File.createTempFile(prefix, ".cluster");
         try (BufferedWriter writer = new BufferedWriter(new FileWriter(clusterFile))) {
-            writer.write("fake:fdbcluster@127.0.0.1:65537\n");
+            writer.write("fake:fdbcluster@127.0.0.1:65535\n");
```
ScottDugas, line 70: Why change the port here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
ScottDugas, line 38: What about the rest of the prefixes in `SystemKeyspace`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperationsTest.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * FDBSystemOperationsTest.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2020 Apple Inc. and the FoundationDB project authors
```
ScottDugas, line 6:
```suggestion
 * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
```diff
@@ -0,0 +1,211 @@
+/*
+ * FDBSystemDatabase.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2020 Apple Inc. and the FoundationDB project authors
```
ScottDugas, line 6:
```suggestion
 * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
```

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
alecgrieser: Hmm, yeah, something like that might make sense, though I'm hesitant to add a whole new interface like that as part of this PR. For now, I went with something simpler, namely I removed all of the public methods that take an `FDBRecordContext` and now the only methods that this class takes are `FDBDatabaseRunner`s, which simplifies this a bit, and for these methods, is actually the same efficiency as the "reads" don't require a real database read (and therefore an FDB read version (I don't think). It also seems like something we could add later if I'm wrong about that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Get the file system path to the cluster file used. This is a synchronous version of
+     * {@link #getClusterFilePathAsync(FDBRecordContext)}.
+     *
+     * @param context a transaction to use to perform the operation
+     * @return the cluster file path
+     * @see #getClusterFilePathAsync(FDBRecordContext)
+     */
+    @Nullable
```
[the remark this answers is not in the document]
alecgrieser: Same response as to https://github.com/FoundationDB/fdb-record-layer/pull/979/files#r574708772

alecgrieser:
One of the PRBs failed for the last commit due to #936

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Get the connection string used to connect to the FDB cluster. This is a synchronous version of
+     * {@link #getConnectionStringAsync(FDBDatabaseRunner)}
+     *
+     * @param runner a runner to use to perform the operation
+     * @return the current cluster connection string
+     * @see #getConnectionStringAsync(FDBDatabaseRunner)
+     * @see #getConnectionStringAsync(FDBDatabaseRunner)
```
ScottDugas, line 120:
```suggestion
```

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
alecgrieser: Whoops, must have added the dupe at some point

alecgrieser:
Probably should "squash and merge" this one

ScottDugas (approved):

MMcM (approved):

PR #982, opened 2020-06-30
Resolves #9: IN-clause to JOIN transformation rework

### LINKED ISSUE #9: IN join in filter can prevent index scan merging
If `IN` is turned into a join over a loop of comparison values, only to be used in to drive a filter operator, and not an actual index scan, it can make the branch of an OR in which it appears no longer ordered by the index scan. Furthermore, the index scan will be repeated unnecessarily.

The `IN` should be pushed further down into the filter. Which is to say, it should be turned back into the runtime `IN` function.

normen662:
- This PR resolves #9 by not destroying the requested order if the inner plan can provide it.
- This PR also fixes a problem where IN to JOIN transformation for a particular IN-clause was performed even if the inner plan was unable to use the correlated filter as an index search argument. This led to repeated scans of an index or even the base data.
- This PR introduces an improved `InExtractor` class that can work in a plan-dependent as well as plan-independent way allowing to make fine-grain decisions whether or not to perform IN-to-JOIN related transformations on a per clause basis.
- This PR also changes plan selection to allow for better comparison of plans that both use IN clauses.


alecgrieser:

PR #984, opened 2020-06-30
Resolves #983: FDBDatabaseRunner should use FDBRecordContextConfig more

### LINKED ISSUE #983: FDBDatabaseRunner should use FDBRecordContextConfig more
By keeping the values for new transactions in a config builder itself, the runner is future-proofed for new capabilities in that config. And transitively the online index builder. These are the places where transactions are created on the client's behalf rather than by the client themselves.

MMcM:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunner.java
```diff
@@ -69,6 +69,18 @@
     @Nonnull
     FDBDatabase getDatabase();
 
+    /**
+     * Get the configuration used in record contexts opened by this runner.
+     * @return configuration to use
+     */
+    FDBRecordContextConfig.Builder getContextConfigBuilder();
```
alecgrieser, line 76:
```suggestion
    @Nonnull
    FDBRecordContextConfig.Builder getContextConfigBuilder();
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
```diff
@@ -344,7 +354,7 @@ public Builder setTransactionTimeoutMillis(long transactionTimeoutMillis) {
             return this;
         }
 
-        private long getTransactionTimeoutMillis() {
+        public long getTransactionTimeoutMillis() {
```
alecgrieser, line 357: I suppose this should have a Javadoc comment along the lines of the other getters in this builder

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -766,14 +766,23 @@ private Transaction createTransaction(@Nonnull FDBRecordContextConfig config,
...
+    }
 
     /**
      * Create an {@link FDBDatabaseRunner} for use against this database.
      * @return a new runner
      */
     @Nonnull
     public FDBDatabaseRunner newRunner() {
-        return new FDBDatabaseRunnerImpl(this);
+        return newRunner(FDBRecordContextConfig.newBuilder().setTransactionTimeoutMillis(factory.getTransactionTimeoutMillis()));
```
alecgrieser, line 785: It is a little weird that this sets the default option here... I suppose another approach would be to make the argument to `FDBDatabaseRunnerImpl` `@Nullable` and then fill it in with a default builder (like this) if `null` is provided, and then filling in the defaults is the job of in `FDBDatabaseRunnerImpl` (or there's a "default config builder" method in `FDBDatabase` and then _that_ does the job of filling in the default values).

I think all of those things could be changed later, though, without affecting the API.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -766,14 +766,23 @@ private Transaction createTransaction(@Nonnull FDBRecordContextConfig config,
         return transaction;
     }
 
+    /**
+     * Create an {@link FDBDatabaseRunner} for use against this database.
+     * @param contextConfigBuilder options for contexts opened by the new runner
```
alecgrieser, line 771: It might be worth clarifying that the builder passed in will be the same object used in the runner, so future changes to the state of the argument will also affect the values as seen in the runner (as it's running).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -784,7 +793,9 @@ public FDBDatabaseRunner newRunner() {
      */
     @Nonnull
     public FDBDatabaseRunner newRunner(@Nullable FDBStoreTimer timer, @Nullable Map<String, String> mdcContext) {
-        return new FDBDatabaseRunnerImpl(this, timer, mdcContext);
+        FDBDatabaseRunner runner = newRunner();
+        runner.getContextConfigBuilder().setTimer(timer).setMdcContext(mdcContext);
+        return runner;
```
alecgrieser, line 798:
```suggestion
        return newRunner(timer, mdcContext, null);
```
I think. I can make arguments that the other two `newRunner`s should remain as is, but this one seems like it should populate the default value of the other one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunner.java
```diff
@@ -69,6 +69,18 @@
     @Nonnull
     FDBDatabase getDatabase();
 
+    /**
+     * Get the configuration used in record contexts opened by this runner.
```
alecgrieser, line 73: It's probably worth clarifying that as the object returned is the same object as (not a copy of) the one used by the runner, if one changes one of the values of the builder, then this will affect the values used when the runner creates new contexts.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/synchronizedsession/SynchronizedSessionRunner.java
```diff
@@ -206,52 +205,18 @@ public FDBDatabase getDatabase() {
...
-
-    @Override
-    public void setWeakReadSemantics(@Nullable FDBDatabase.WeakReadSemantics weakReadSemantics) {
-        underlying.setWeakReadSemantics(weakReadSemantics);
-    }
-
-    @Nonnull
-    @Override
-    public FDBTransactionPriority getPriority() {
-        return underlying.getPriority();
```
alecgrieser, line 249: Looks like get/set transaction timeout millis can be similarly removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -766,14 +766,23 @@ private Transaction createTransaction(@Nonnull FDBRecordContextConfig config,
...
     }
 
+    /**
+     * Create an {@link FDBDatabaseRunner} for use against this database.
+     * @param contextConfigBuilder options for contexts opened by the new runner
+     * @return a new runner
+     */
+    @Nonnull
+    public FDBDatabaseRunner newRunner(@Nonnull FDBRecordContextConfig.Builder contextConfigBuilder) {
+        return new FDBDatabaseRunnerImpl(this, contextConfigBuilder);
```
alecgrieser, line 776: It's a little weird that one might forget to set the default transaction timeout millis here if one isn't careful...I suppose a "default context config builder" method on a database (or, I suppose, a static "default builder" method on a `FDBRecordContextConfig` that took an `FDBDatabase`) as suggested in a different comment might help with that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
MMcM: Added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: This turns out not to be necessary. The old code has some redundancy that was copied into the new. The default is to get from the factory where the default is to leave it alone from the FDB client's setting. There is no need to read from the factory here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: Added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: How so? The default for `weakSemantics` is `null`, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunner.java
MMcM: Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/synchronizedsession/SynchronizedSessionRunner.java
MMcM: Removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: This is no longer the case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
alecgrieser: Right, sorry. I meant just that this method should call `newRunner(FDBStoreTimer, Map, WeakReadSemantics)` with a `null` `WeakReadSemantics` rather than tweaking the context config builder just so that it fits into the Java pattern for "methods with default parameters". But I don't feel strongly about this and could see this staying as is

MMcM: Oh, I see. I would just as soon deprecate those two methods. But either way, I don't think the one with all the arguments is, in fact, the most general one, like in that pattern. The one with one argument is, so maybe they should be expressed in terms of that?

PR #986, opened 2020-07-01
Resolves #985: Get rid of finalizer

### LINKED ISSUE #985: Get rid of finalizer
`TracedTransaction` defines `finalize` to enhance FDB's handling of unclosed transactions.

FDB is planning to remove its own use of `finalize` in the reasonably near future.

Java has already deprecated `finalize` as of Java 9.

There is no need for phantom references or anything so complicated because transactions have an inherently limited lifetime and an explicit `close` already. The only issue is buggy code with leaks.

Track open record contexts on the database object. Periodically check for ones that have been open for too long, close them and log a warning. Optionally (such as for some small percentage of transactions or in test environments) record a stack trace to include in the warning.

MMcM:
Includes #984


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -480,6 +482,22 @@ public void setTransactionIsTracedSupplier(Supplier<Boolean> transactionIsTraced
...
+     * Get whether to warn and close record contexts open for too long.
+     * @return number of seconds after which a context and be closed and a warning logged
+     */
+    public long getWarnAndCloseOpenContextsAfterSeconds() {
+        return warnAndCloseOpenContextsAfterSeconds;
+    }
+
+    /**
+     * Set whether to warn and close record contexts open for too long.
+     * @param warnAndCloseOpenContextsAfterSeconds  number of seconds after which a context and be closed and a warning logged
```
alecgrieser, line 495:
```suggestion
     * @param warnAndCloseOpenContextsAfterSeconds number of seconds after which a context and be closed and a warning logged
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -463,14 +464,15 @@ public long getReverseDirectoryMaxMillisPerTransaction() {
         return reverseDirectoryMaxMillisPerTransaction;
     }
 
+    // TODO: Demote these to UNSTABLE and deprecate at some point.
```
alecgrieser, line 467: I guess I'm not 100% clear on this, but can we deprecate them now (according to our rules)? If this is stable, I suppose we can't remove them until version 3, but I thought deprecating "should" be okay at any time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -1120,6 +1134,55 @@ protected RuntimeException mapAsyncToSyncException(@Nonnull Throwable ex) {
...
+     * @param minAgeSeconds number of seconds above which to warn
+     * @return number of such contexts found
+     */
+    public int warnAndCloseOldTrackedOpenContexts(long minAgeSeconds) {
+        long nanoTime = System.nanoTime() - TimeUnit.SECONDS.toNanos(minAgeSeconds);
+        if (trackedOpenContexts.isEmpty()) {
+            return 0;
+        }
+        try {
+            if (trackedOpenContexts.firstKey() > nanoTime) {
```
alecgrieser, line 1148: I think if this used `pollFirstEntry`, then this wouldn't need to catch an exception (but would have to check for null). I suppose that could then be combined with the check to see if the data structure is empty.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -1120,6 +1134,55 @@ protected RuntimeException mapAsyncToSyncException(@Nonnull Throwable ex) {
         return future.get();
     }
 
+    /**
+     * Log warning and close tracked contexts that have been open for too long.
+     * @param minAgeSeconds number of seconds above which to warn
+     * @return number of such contexts found
+     */
+    public int warnAndCloseOldTrackedOpenContexts(long minAgeSeconds) {
```
alecgrieser, line 1142: Is this public so that a consumer might decide that they want to run this more often (or I suppose in their own tests)? It seems that in general, this should be package private, but I could also see the argument that either of those would be useful features.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -337,6 +340,34 @@ public boolean isLogged() {
...
+    public long getTrackOpenTimeNanos() {
+        return trackOpenTimeNanos;
+    }
+
+    /**
+     * Set the nanosecond time at which this context was opened.
+     * @param trackOpenTimeNanos  time opened
+     */
+    @API(API.Status.INTERNAL)
+    public void setTrackOpenTimeNanos(final long trackOpenTimeNanos) {
```
alecgrieser, line 357: This doesn't feel like it should be `public`. I think I can get behind the others, though they also seem more package-private (unless they're supposed to be visible for testing, or something like that).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -152,22 +154,23 @@
...
-                               @Nonnull FDBRecordContextConfig config,
-                               boolean transactionIsTraced) {
+                               @Nonnull FDBRecordContextConfig config) {
         super(fdb, transaction, config.getTimer());
         this.transactionCreateTime = System.currentTimeMillis();
         this.localVersion = new AtomicInteger(0);
         this.localVersionCache = new ConcurrentSkipListMap<>(ByteArrayUtil::compareUnsigned);
         this.versionMutationCache = new ConcurrentSkipListMap<>(ByteArrayUtil::compareUnsigned);
         this.transactionId = getSanitizedId(config);
+        this.openStackTrace = config.isSaveOpenStackTrace() ? new Throwable() : null;
```
alecgrieser, line 168: Seems like the throwable here should have a message here about how it's not thrown (or equivalently that this is created only for the stack trace).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -1118,6 +1132,55 @@ protected RuntimeException mapAsyncToSyncException(@Nonnull Throwable ex) {
...
+            context.close();
+            count++;
+        }
+        return count;
+    }
+
+    protected void trackOpenContext(FDBRecordContext context) {
+        long key = System.nanoTime();
+        while (key == 0 || trackedOpenContexts.putIfAbsent(key, context) != null) {
+            key++;  // Might not have nanosecond resolution and need something non-zero and unique.
```
scgray, line 1171: Just an observation: I wasn't aware that `nanoTime()` could return 0. That's useful to be aware of.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: You mean `firstEntry`? We don't want to remove it from the map. I assume that the most common case is that the map is not empty but all entries are younger than the threshold. `firstEntry` conses. I suppose it'll probably be on the stack, so maybe this isn't a real issue.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
MMcM: This one does not need to be public; the getter does, for tests, as you say.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
MMcM: Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
MMcM: It needs to be public so that testing frameworks can call it when shutting down and find leaks, for instance. I added an annotation.

I added a KPI for when something is closed because it was open for too long.

PR #996, opened 2020-07-15
Fix typo in QueryableKeyExpression.evalForQuery that was checking the …

MMcM:
… same condition twice.

The error was highlighted by IntelliJ.


MMcM:
Closing in favor more completely addressing #997, which just fixing this would quickly expose.

PR #999, opened 2020-07-17
Resolve #703, OnlineIndexer should limit amount of work by transactio…

### LINKED ISSUE #703: OnlineIndexer should limit amount of work by transaction size
The `OnlineIndexer` should have an additional limit by transaction size. This should be used to avoid writing too much in a single transaction, something especially pertinent for expensive indexes (like full text indexes). Though FDB's large transaction handling is improving, it's still something we probably want to take better control over.

In FDB 6.2, there is now a way to ask a transaction for an estimate of its size. We should use that new API to do this, though that requires #702.

jjezra:
…n size

    While interating, limit range to the one record beyond a predefined max write size.
    Add FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_SIZE/COUNT to count the number of cases
    of terminating a range by size vs. by count limit.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -570,10 +570,10 @@ public class FDBStoreTimer extends StoreTimer {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
-        /** The number of {@link OnlineIndexer} ranges defined by scanned record count limit. */
-        ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
-        /** The number of {@link OnlineIndexer} ranges defined by transaction write size limit. */
-        ONLINE_INDEX_BUILDER_RANGE_BY_SIZE("number of indexer iterations terminated by write limit", false),
+        /** The number of {@link OnlineIndexer} range scans terminated after hitting the scan limit. */
+        ONLINE_INDEX_BUILDER_RANGES_BY_COUNT("number of indexer iterations terminated by scan limit", false),
+        /** The number of {@link OnlineIndexer} range scans terminated after hitting the size limit. */
+        ONLINE_INDEX_BUILDER_RANGES_BY_SIZE("number of indexer iterations terminated by write limit", false),
         /** The number of times that a leaderboard update adds a time window. */
         TIME_WINDOW_LEADERBOARD_ADD_WINDOW("number of leaderboard windows added", false),
         /** The number of times that a leaderboard update deleted a time window. */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -570,6 +570,10 @@ public class FDBStoreTimer extends StoreTimer {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
+        /** The number of indexer range termination by scan limit {@link OnlineIndexer}. */
+        ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
+        /** The number of indexer range termination by write limit {@link OnlineIndexer}. */
+        ONLINE_INDEX_BUILDER_RANGE_BY_SIZE("number of indexer iterations terminated by write limit", false),
         /** The number of times that a leaderboard update adds a time window. */
         TIME_WINDOW_LEADERBOARD_ADD_WINDOW("number of leaderboard windows added", false),
         /** The number of times that a leaderboard update deleted a time window. */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -570,9 +570,9 @@ public class FDBStoreTimer extends StoreTimer {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
-        /** The number of {@link OnlineIndexer} records scanned in a single iteration/transaction. */
+        /** The number of {@link OnlineIndexer} ranges defined by scanned record count limit. */
         ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
-        /** The number of {@link OnlineIndexer} bytes to write in a single iteration/transaction. */
+        /** The number of {@link OnlineIndexer} ranges defined by transaction write size limit. */
         ONLINE_INDEX_BUILDER_RANGE_BY_SIZE("number of indexer iterations terminated by write limit", false),
         /** The number of times that a leaderboard update adds a time window. */
         TIME_WINDOW_LEADERBOARD_ADD_WINDOW("number of leaderboard windows added", false),
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -570,9 +570,9 @@ public class FDBStoreTimer extends StoreTimer {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
-        /** The number of indexer range termination by scan limit {@link OnlineIndexer}. */
+        /** The number of {@link OnlineIndexer} records scanned in a single iteration/transaction. */
         ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
-        /** The number of indexer range termination by write limit {@link OnlineIndexer}. */
+        /** The number of {@link OnlineIndexer} bytes to write in a single iteration/transaction. */
         ONLINE_INDEX_BUILDER_RANGE_BY_SIZE("number of indexer iterations terminated by write limit", false),
         /** The number of times that a leaderboard update adds a time window. */
         TIME_WINDOW_LEADERBOARD_ADD_WINDOW("number of leaderboard windows added", false),
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -561,7 +561,8 @@ public class OnlineIndexer implements AutoCloseable {
 
         AtomicLong recordsScannedCounter = new AtomicLong();
         // Note: This runs all of the updates in serial in order to not invoke a race condition
-        // in the rank code that was causing incorrect results.
+        // in the rank code that was causing incorrect results. If everything were thread safe,
+        // a larger pipeline size would be possible.
 
         final AtomicReference<RecordCursorResult<FDBStoredRecord<Message>>> lastResult = new AtomicReference<>(RecordCursorResult.exhausted());
         final FDBRecordContext context = store.getContext();
@@ -590,14 +591,16 @@ public class OnlineIndexer implements AutoCloseable {
             if (timer != null) {
                 timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
             }
-            return (syntheticPlan == null ?
-                    // case 1: simple
-                    maintainer.update(null, rec) :
-                    // case 2: synatheticPlan - note that Pipeline size is 1, since not all maintainers are thread-safe.
-                    syntheticPlan.execute(store, rec
-                    ).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
-                    // both cases provide a CompletableFuture<Void>
-                    ).thenCompose(vignore ->
+
+            final CompletableFuture<Void> updateMaintainer;
+            if (syntheticPlan == null) {
+                updateMaintainer = maintainer.update(null, rec);
+            } else {
+                // Pipeline size is 1, since not all maintainers are thread-safe.
+                updateMaintainer = syntheticPlan.execute(store, rec).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1);
+            }
+
+            return updateMaintainer.thenCompose(vignore ->
                     context.getApproximateTransactionSize().thenCompose(size -> {
                         if (size >= config.getMaxWriteLimitBytes()) {
                             // the transaction becomes too big - stop iterating
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -130,7 +130,7 @@ public class OnlineIndexer implements AutoCloseable {
     /**
      * Default transaction write size limit. Note that the actual write might be "a little" bigger.
      */
-    public static final int DEFAULT_WRITE_SIZE = 900 * 1024;
+    public static final int DEFAULT_WRITE_LIMIT_BYTES = 900_000;
     /**
      * Default limit to the number of records to attempt in a single second.
      */
@@ -561,10 +561,9 @@ public class OnlineIndexer implements AutoCloseable {
 
         AtomicLong recordsScannedCounter = new AtomicLong();
         // Note: This runs all of the updates in serial in order to not invoke a race condition
-        // in the rank code that was causing incorrect results. If everything were thread safe,
-        // a larger pipeline size would be possible.
+        // in the rank code that was causing incorrect results.
 
-        final AtomicReference<RecordCursorResult<FDBStoredRecord<Message>>> holder = new AtomicReference<>(RecordCursorResult.exhausted());
+        final AtomicReference<RecordCursorResult<FDBStoredRecord<Message>>> lastResult = new AtomicReference<>(RecordCursorResult.exhausted());
         final FDBRecordContext context = store.getContext();
         return AsyncUtil.whileTrue(() -> cursor.onNext().thenCompose(result -> {
 
@@ -573,11 +572,10 @@ public class OnlineIndexer implements AutoCloseable {
                 if (timer != null) {
                     timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_COUNT);
                 }
-                holder.set(result);
+                lastResult.set(result);
                 return AsyncUtil.READY_FALSE;
             }
 
-            // here: implement forEachResultAsync
             final FDBStoredRecord<Message> rec = result.get();
             empty.set(false);
             if (timer != null) {
@@ -588,30 +586,32 @@ public class OnlineIndexer implements AutoCloseable {
                 // This record is not our type, swipe left
                 return AsyncUtil.READY_TRUE;
             }
-            // here: add this index to the transaction
+            // add this index to the transaction
             if (timer != null) {
                 timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
             }
             return (syntheticPlan == null ?
+                    // case 1: simple
                     maintainer.update(null, rec) :
-                    syntheticPlan.execute(store, rec)
-                            .forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
+                    // case 2: synatheticPlan - note that Pipeline size is 1, since not all maintainers are thread-safe.
+                    syntheticPlan.execute(store, rec
+                    ).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
+                    // both cases provide a CompletableFuture<Void>
                     ).thenCompose(vignore ->
                     context.getApproximateTransactionSize().thenCompose(size -> {
-                        if (size >= config.getMaxWriteSize()) {
+                        if (size >= config.getMaxWriteLimitBytes()) {
                             // the transaction becomes too big - stop iterating
                             if (timer != null) {
                                 timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_SIZE);
                             }
-                            holder.set(result);
+                            lastResult.set(result);
                             return AsyncUtil.READY_FALSE;
                         }
                         return AsyncUtil.READY_TRUE;
                     }));
 
-        }), cursor.getExecutor()).thenCompose(vIgnore -> {
-            RecordCursorResult<FDBStoredRecord<Message>> noNextResult = holder.get();
-
+        }), cursor.getExecutor()
+        ).thenCompose(vignore -> {
             long recordsScannedInTransaction = recordsScannedCounter.get();
             if (recordsScanned != null) {
                 recordsScanned.addAndGet(recordsScannedInTransaction);
@@ -620,7 +620,7 @@ public class OnlineIndexer implements AutoCloseable {
                 store.context.ensureActive().mutate(MutationType.ADD, scannedRecordsSubspace.getKey(),
                         FDBRecordStore.encodeRecordCount(recordsScannedInTransaction));
             }
-            byte[] nextCont = empty.get() ? null : noNextResult.getContinuation().toBytes();
+            byte[] nextCont = empty.get() ? null : lastResult.get().getContinuation().toBytes();
             if (nextCont == null) {
                 return CompletableFuture.completedFuture(null);
             } else {
@@ -1450,19 +1450,19 @@ public class OnlineIndexer implements AutoCloseable {
     @API(API.Status.UNSTABLE)
     public static class Config {
         private final int maxLimit;
-        private final int maxWriteSize;
+        private final int maxWriteLimitBytes;
         private final int maxRetries;
         private final int recordsPerSecond;
         private final long progressLogIntervalMillis;
         private final int increaseLimitAfter;
 
-        private Config(int maxLimit, int maxRetries, int recordsPerSecond, long progressLogIntervalMillis, int increaseLimitAfter, int maxWriteSize) {
+        private Config(int maxLimit, int maxRetries, int recordsPerSecond, long progressLogIntervalMillis, int increaseLimitAfter, int maxWriteLimitBytes) {
             this.maxLimit = maxLimit;
             this.maxRetries = maxRetries;
             this.recordsPerSecond = recordsPerSecond;
             this.progressLogIntervalMillis = progressLogIntervalMillis;
             this.increaseLimitAfter = increaseLimitAfter;
-            this.maxWriteSize = maxWriteSize;
+            this.maxWriteLimitBytes = maxWriteLimitBytes;
         }
 
         /**
@@ -1510,13 +1510,11 @@ public class OnlineIndexer implements AutoCloseable {
         }
 
         /**
-         * Get the approximate size limit, represents a desired max transaction message size. Note that the actual write
-         * might be "a little" larger than this limit.
-         * @return a the efficient write size. Typically a transaction will be submitted when it becomes bigger than
-         * this value.
+         * Stop scanning if the write size (bytes) becomes bigger that this value.
+         * @return the write size
          */
-        public long getMaxWriteSize() {
-            return maxWriteSize;
+        public long getMaxWriteLimitBytes() {
+            return maxWriteLimitBytes;
         }
 
         @Nonnull
@@ -1532,7 +1530,7 @@ public class OnlineIndexer implements AutoCloseable {
         public Builder toBuilder() {
             return Config.newBuilder()
                     .setMaxLimit(this.maxLimit)
-                    .setWriteSize(this.maxWriteSize)
+                    .setWriteLimitBytes(this.maxWriteLimitBytes)
                     .setIncreaseLimitAfter(this.increaseLimitAfter)
                     .setProgressLogIntervalMillis(this.progressLogIntervalMillis)
                     .setRecordsPerSecond(this.recordsPerSecond)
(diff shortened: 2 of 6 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
```diff
@@ -697,5 +698,96 @@ private void saveManyRecords(@Nonnull RecordMetaDataHook hook, int count1, int c
...
+        try (FDBRecordContext context = openContext()) {
+            uncheckedOpenSimpleRecordStore(context, hook);
+            recordStore.checkVersion(null, FDBRecordStoreBase.StoreExistenceCheck.ERROR_IF_NOT_EXISTS).join();
+
+            assertTrue(recordStore.isIndexWriteOnly("newIndex"));
+
+            timer.reset();
+
+            // Build in this transaction.
+            try (OnlineIndexer indexBuilder =
```
nblintao, line 734: I think some old code still calls is `indexBuilder` (`OnlineIndexer` used to be called `OnlineIndexBuilder`) but it'd better to be called `indexer` because it's weird to have a "Builder" for a "XXXBuilder".

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
         // Note: This runs all of the updates in serial in order to not invoke a race condition
```
nblintao, line 563: Is this comment still true?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -570,6 +570,10 @@ public String logKey() {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
+        /** The number of indexer range termination by scan limit {@link OnlineIndexer}. */
```
nblintao, line 573: Maybe you can explain more about "range termination" here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
         // in the rank code that was causing incorrect results. If everything were thread safe,
         // a larger pipeline size would be possible.
-        return cursor.forEachResultAsync(result -> {
+
+        final AtomicReference<RecordCursorResult<FDBStoredRecord<Message>>> holder = new AtomicReference<>(RecordCursorResult.exhausted());
+        final FDBRecordContext context = store.getContext();
+        return AsyncUtil.whileTrue(() -> cursor.onNext().thenCompose(result -> {
+
+            if (!result.hasNext()) {
```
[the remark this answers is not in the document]
jjezra: Yes, it had confused me too! 
I asked Scott G. about it - he think that hasNext() should have been something like isValid().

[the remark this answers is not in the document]
alecgrieser: I don't think `isValid` is quite right... The idea is that you want to distinguish getting a result _with_ an item and getting a result that just indicates that it's at the end of iteration. Perhaps `hasElement`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
alecgrieser, line 573: Perhaps something more like:

```suggestion
        /** The number of {@link OnlineIndexer} range scans terminated after hitting the scan limit. */
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1807,6 +1868,28 @@ public Builder setLimit(int limit) {
...
+            return maxWriteSize;
+        }
+
+        /**
+         * Set the approximate maximum transaction write size. Note that the actual size might be "a little" bigger.
+         * he default limit is {@link #DEFAULT_WRITE_SIZE} = {@value #DEFAULT_WRITE_SIZE}.
+         * @param max the desired max write size
+         * @return this builder
+         */
+
```
alecgrieser, line 1886:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1807,6 +1868,28 @@ public Builder setLimit(int limit) {
             return this;
         }
 
+        /**
+         * Get the approximate maximum transaction write size. Note that the actual size might be "a little" bigger.
+         * @return the requested write size. Index iteration stops when exceeding this size.
```
alecgrieser, line 1873: In general, we try and keep `@return` statements as a single "statement" (without a period), so something like `@return the requested write size`, and then additional clarifying comments go in the body.

Also, looking at "Index iteration stops when exceeding this size.", the way I read that is that it stops, like, "iterating over the index and the build stops" or something when it hits the limit, which I don't think is what you're getting at. I think something like "transactions started as part of the index build will be committed after they exceed this size, and a new transaction will be started", or something like that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -570,6 +570,10 @@ public String logKey() {
...
+        ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
+        /** The number of indexer range termination by write limit {@link OnlineIndexer}. */
```
alecgrieser, line 575: In a similar vein (presented mostly for parallelism, but this really should be whatever is decided for the first timer option, but adapted for this one).
```suggestion
        /** The number of {@OnlineIndexer} range scans terminated after hitting the by write limit. */
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -126,6 +127,10 @@
      * Default number of records to attempt to run in a single transaction.
      */
     public static final int DEFAULT_LIMIT = 100;
+    /**
+     * Default transaction write size limit. Note that the actual write might be "a little" bigger.
+     */
+    public static final int DEFAULT_WRITE_SIZE = 900 * 1024;
```
alecgrieser, line 133: I suppose 900 kiB is a fine enough default size as any, though I believe that the FDB limits use base-1000 rather than base-1024 units in general (that is to say, the absolute maximum transaction size is 10 MB = 10 * 10<sup>6</sup> B, not 10 MiB = 10 * 2<sup>20</sup> B).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
+            // here: add this index to the transaction
+            if (timer != null) {
+                timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
+            }
+            return (syntheticPlan == null ?
+                    maintainer.update(null, rec) :
+                    syntheticPlan.execute(store, rec)
+                            .forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
+                    ).thenCompose(vignore ->
+                    context.getApproximateTransactionSize().thenCompose(size -> {
```
alecgrieser, line 600: I think this can be `.thenApply` instead of `.thenCompose`, and then it would return `false` and `true` instead of `READY_FALSE` and `READY_TRUE`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: This appears to still be true, at least the way I'm reading the code

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
```diff
@@ -697,5 +698,96 @@ private void saveManyRecords(@Nonnull RecordMetaDataHook hook, int count1, int c
             context.commit();
         }
     }
+    
+    @Test
+    public void testOnlineIndexBuilderMaxWrite() throws Exception {
+        try (FDBRecordContext context = openContext()) {
+            uncheckedOpenSimpleRecordStore(context, BASIC_HOOK);
+            recordStore.checkVersion(null, FDBRecordStoreBase.StoreExistenceCheck.ERROR_IF_EXISTS).join();
```
alecgrieser, line 706: Hm, I'm not sure this separate transaction is necessary, though the other `OnlineIndexer`-related tests in this file are doing it, so ¯\\\_(ツ)_/¯

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
```diff
@@ -697,5 +698,96 @@ private void saveManyRecords(@Nonnull RecordMetaDataHook hook, int count1, int c
             context.commit();
         }
     }
+    
+    @Test
+    public void testOnlineIndexBuilderMaxWrite() throws Exception {
```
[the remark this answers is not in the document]
alecgrieser: I could also believe that a separate `OnlineIndexerWithRecordTypeKeysTest.java` (or something like that) could be a good idea, and that the other tests of the online indexer in this file could be moved there.

Or, actually, perhaps more holistically, I think we don't (necessarily) want to test this with a test of `RecordTypeKeys`, but instead probably just want a test with some of the more traditional `OnlineIndexer` tests in `OnlineIndexerSimpleTest`.

alecgrieser:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: Sure! (and don't worry about the nitpicking - I'm here to lear). 
I've changed that to DEFAULT_WRITE_LIMIT_BYTES (keeping the unit at the end). Would that be acceptable?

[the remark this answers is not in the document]
nblintao: Sure. A GitHub tip: I'm you are want to send out tons of replies in a batch, you can do it in the reviewer's view https://github.com/FoundationDB/fdb-record-layer/pull/999/files
image

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: Modern C++ usually uses the reverse logic, as in _while(! cursor.isEnd())_. Would that be clearer or even more confusing? 
__
C++ iterator example:
for (ptr = ar.begin(); ptr < ar.end(); ptr++)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
+            if (!result.hasNext()) {
+                // end of the cursor list
+                if (timer != null) {
+                    timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_COUNT);
+                }
+                holder.set(result);
+                return AsyncUtil.READY_FALSE;
+            }
+
+            // here: implement forEachResultAsync
```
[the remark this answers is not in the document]
jjezra: Removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
-                    return syntheticPlan.execute(store, rec).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1);
-                }
-            } else {
-                return AsyncUtil.DONE;
+            if (!recordTypes.contains(rec.getRecordType())) {
+                // This record is not our type, swipe left
+                return AsyncUtil.READY_TRUE;
             }
-        }).thenCompose(noNextResult -> {
+            // here: add this index to the transaction
```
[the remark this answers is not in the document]
jjezra: Removed "here"

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
nblintao: If mean "If everything were thread safe, a larger pipeline size would be possible" part. Will there be problem to have a non-1 pipeline size at the write part?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
jjezra: Yes, this is very unclear. How about: 

```suggestion
        /** The indexer's single transaction range limit - by number of records scanned {@link OnlineIndexer}. */
        ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
        /** The indexer's single transaction range limit - by transaction's write size  {@link OnlineIndexer}. */
        ONLINE_INDEX_BUILDER_RANGE_BY_SIZE("number of indexer iterations terminated by write limit", false),
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
             }
             recordsScannedCounter.incrementAndGet();
-            if (recordTypes.contains(rec.getRecordType())) {
-                if (timer != null) {
-                    timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
-                }
-                if (syntheticPlan == null) {
-                    return maintainer.update(null, rec);
-                } else {
-                    // Pipeline size is 1, since not all maintainers are thread-safe.
```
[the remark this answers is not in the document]
jjezra: Nope. I've added it to the systheticPlan line

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
...
-        }).thenCompose(noNextResult -> {
+            // here: add this index to the transaction
+            if (timer != null) {
+                timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
+            }
+            return (syntheticPlan == null ?
+                    maintainer.update(null, rec) :
+                    syntheticPlan.execute(store, rec)
+                            .forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
+                    ).thenCompose(vignore ->
```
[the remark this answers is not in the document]
jjezra: I've tried to ease the reading with comments...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -558,27 +563,55 @@ private void increaseLimit() {
         // Note: This runs all of the updates in serial in order to not invoke a race condition
         // in the rank code that was causing incorrect results. If everything were thread safe,
         // a larger pipeline size would be possible.
-        return cursor.forEachResultAsync(result -> {
+
+        final AtomicReference<RecordCursorResult<FDBStoredRecord<Message>>> holder = new AtomicReference<>(RecordCursorResult.exhausted());
```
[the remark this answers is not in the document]
jjezra: I like the dual meaning of "last" - as both most recent and lasting...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1474,6 +1509,16 @@ public int getIncreaseLimitAfter() {
             return increaseLimitAfter;
         }
 
+        /**
+         * Get the approximate size limit, represents a desired max transaction message size. Note that the actual write
+         * might be "a little" larger than this limit.
+         * @return a the efficient write size. Typically a transaction will be submitted when it becomes bigger than
```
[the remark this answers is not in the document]
jjezra: Oops.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: I've tried - it became messy, so I'll keep going here (for now) to avoid losing momentum. Sorry for the frequent emails.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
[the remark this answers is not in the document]
jjezra: I agree. However, since my attempt to port it became a little messy, would it be ok to do it in a separate, standalone fix?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
```diff
@@ -697,5 +698,96 @@ private void saveManyRecords(@Nonnull RecordMetaDataHook hook, int count1, int c
...
+        }
+
+        saveManyRecords(BASIC_HOOK, 250, 250);
 
+        try (FDBRecordContext context = openContext()) {
+            uncheckedOpenSimpleRecordStore(context, BASIC_HOOK);
+            recordStore.checkVersion(null, FDBRecordStoreBase.StoreExistenceCheck.ERROR_IF_NOT_EXISTS).join();
+
+            assertEquals(250, recordStore.getSnapshotRecordCountForRecordType("MySimpleRecord").join().intValue());
+            assertEquals(250, recordStore.getSnapshotRecordCountForRecordType("MyOtherRecord").join().intValue());
```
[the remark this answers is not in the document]
jjezra: It is my way to verify the behavior with high write limit, before making sure the "flattened" 1 byte write limit behaves the same. Is that an issue?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I removed the hypothetical part of the comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
jjezra: Oh (already visited this item..). How about combining your/my suggesting to: 
        /** The number of {@link OnlineIndexer} records scanned in a single iteration/transaction. */
        /** The number of {@link OnlineIndexer} bytes to write in a single iteration/transaction. */

jjezra: How about 

```
        /** The number of {@link OnlineIndexer} records scanned in a single iteration/transaction. */
        /** The number of {@link OnlineIndexer} bytes to write in a single iteration/transaction. */
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: 900_000 it is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I've tried that and it didn't work. I assume that the expectations is for a boolean completable future is not happen with the thenApply.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
jjezra: ¯\_(ツ)_/¯

(maybe the second test was a copy&paste, then the rest followed)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: What _exactly_ did you try? It should work if you use `.thenApply` and return vanilla booleans (not CompletableFutures)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
alecgrieser: It's probably worth consolidating the discussion here to a single thread. I'll respond in: https://github.com/FoundationDB/fdb-record-layer/pull/999#discussion_r458965386

alecgrieser: I think that the current comment seems to suggest that the timer's will contain, like, the values of the limits themselves rather than the number of times a transaction is committed because it hits each limit (which I think is what it's actually recording, right?).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
alecgrieser: I'm not sure. I would worry `.isEnd()` being a _little_ confusing, as we also use "end" in the context of "end continuation" to indicate "there is no more data left to read" (as distinct from "we have reached a limit, and that's why _this_ cursor is done"), but I'm not sure I have too many opinions, really

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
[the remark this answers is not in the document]
alecgrieser: I think if a full refactoring is done, then, yeah, that would be a separate PR. If it's adding a new test not using `RecordTypeKey`s, then that feels like it should be in this PR (imo)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: I think we still want to have that hypothetical comment. At the moment, we can't do anything in parallel, but at some point we will... It probably doesn't matter too much, and I think the main upshot from this change is that we'd need to think about how to pipeline the `.whileTrue`, etc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -556,29 +561,57 @@ private void increaseLimit() {
...
             }
-        }).thenCompose(noNextResult -> {
+            // add this index to the transaction
+            if (timer != null) {
+                timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
+            }
+            return (syntheticPlan == null ?
+                    // case 1: simple
+                    maintainer.update(null, rec) :
+                    // case 2: synatheticPlan - note that Pipeline size is 1, since not all maintainers are thread-safe.
```
alecgrieser, line 596:
```suggestion
                    // case 2: syntheticPlan - note that Pipeline size is 1, since not all maintainers are thread-safe.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -556,29 +561,57 @@ private void increaseLimit() {
...
+                    syntheticPlan.execute(store, rec
+                    ).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1)
```
alecgrieser, line 598: I kind of liked @nblintao's suggestion to remove the ternary operator and make it a real `if` statement for readability, but I suppose it doesn't really matter

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RecordTypeKeyTest.java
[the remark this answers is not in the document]
jjezra: Ok. I'll try again.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
jjezra: I've changed that according to your suggestion below. Please review the change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: NP. I'll put it back.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: If the majority claim that A is more readable than B, then more readable it is.

jjezra:
> I think there was one or two outstanding discussions from a previous review, and in particular, I still think it's not quite right to have the new tests only use the `RecordTypeKeyTest` rather than add to include them in one of the files for testing the online indexer

Np. I'll move it in the next commit.

jjezra:
Good morning Tao, Alec, 
I believe that this (third) commit addresses all your comments. Would you find the time today for a re-review? 
Thanks!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerSimpleTest.java
```diff
@@ -883,4 +883,88 @@ public void testConfigLoader() throws Exception {
...
+                                 .build()) {
+
+                Key.Evaluated key = indexer.buildUnbuiltRange(Key.Evaluated.scalar(0L), Key.Evaluated.scalar(25L)).join();
+                assertEquals(1, key.getLong(0));
+                assertEquals(1, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_SIZE));
+                assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGE_BY_COUNT));
+            }
+            recordStore.clearAndMarkIndexWriteOnly("newIndex").join();
+            context.commit();
+        }
```
alecgrieser, line 968: I think this probably could use one more test where it calls `OnlineIndexer.buildIndex` (i.e., building the index across multiple transactions) and makes sure that we get a built index, at the end of it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -570,6 +570,10 @@ public String logKey() {
         ONLINE_INDEX_BUILDER_RECORDS_SCANNED("number of records scanned by online index build", false),
         /** The number of records indexed by {@link OnlineIndexer}. */
         ONLINE_INDEX_BUILDER_RECORDS_INDEXED("number of records indexed by online index build", false),
+        /** The number of {@link OnlineIndexer} records scanned in a single iteration/transaction. */
+        ONLINE_INDEX_BUILDER_RANGE_BY_COUNT("number of indexer iterations terminated by scan limit", false),
+        /** The number of {@link OnlineIndexer} bytes to write in a single iteration/transaction. */
```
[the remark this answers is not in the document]
jjezra: Shall I use: 
```
previous comment for number of records scanned (as it was): 
 /** The number of records indexed by {@link OnlineIndexer}. */
and the new size limitation (as you suggested in earlier comment): 
 /** The number of {@link OnlineIndexer} range scans terminated after hitting the scan limit. */
```
? 
If not, would you suggest another comment(s)?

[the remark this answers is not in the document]
alecgrieser: Yeah, I like `The number of {@link OnlineIndexer} range scans terminated after hitting the scan limit.` (and similar for `ONLINE_INDEX_BUILDER_RANGE_BY_COUNT`.

Although, actually, the names here should probably be `RANGES_BY_COUNT` and `RANGES_BY_SIZE` (plural instead of singular).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerSimpleTest.java
jjezra: Just pushed the thenApply (sorry, forgot about it yesterday)

I would need some guidance writing the additional test. Of all the existing indexer tests, Is there one I can use as an example?

jjezra:
@alecgrieser, @nblintao, 
I hope that this is the final version and everything was addressed. Would you find the time to review it? 
Thank you!

alecgrieser:
Looks like the tests failed due to an error we sometimes see if the FDB cluster (in CI) is too busy:

```
com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerBuildValueIndexTest$Safe > fiftySplitElements() FAILED

        Caused by:
```

We should probably look into how to make our tests more resilient to transient failures like this, but alas

@fdb-build test this please

PR #1001, opened 2020-07-22
Resolves #1000: Some filters can be performed with covering index …

### LINKED ISSUE #1000: Some filters can be performed with covering index scan's partial record
This is complementary with #991, in that the plan element that might introduce that gets a full record from a partial record after a set operation can also be used after a suitable filter. The filter is "pushed-down" in the same way that the set operation was.

MMcM:
… scan's partial record.

* Make `FunctionKeyExpression.Builder`'s `abstract` function public so can be defined anywhere.
* Planner attempts to match the key expressions corresponding to the filter component to the index entry.
* Do a similar match in the `QueryToKeyMatcher`.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FunctionKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FunctionKeyExpression.java
@@ -299,7 +299,7 @@ public abstract class FunctionKeyExpression extends BaseKeyExpression implements
         }
 
         @Nonnull
-        abstract FunctionKeyExpression build(@Nonnull KeyExpression arguments);
+        public abstract FunctionKeyExpression build(@Nonnull KeyExpression arguments);
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -1373,6 +1373,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             if (filterPlan.getChild() instanceof RecordQueryPlanWithIndex) {
                 final RecordQueryPlanWithIndex filteredIndexPlan = (RecordQueryPlanWithIndex) filterPlan.getChild();
                 final Index index = metaData.getIndex(filteredIndexPlan.getIndexName());
+                // A TEXT index, in particular, does not have the actual indexed text in the entry.
                 if (indexTypes.getValueTypes().contains(index.getType())) {
                     final QueryComponent filter = filterPlan.getFilter();
                     final Set<KeyExpression> filterFields = new HashSet<>();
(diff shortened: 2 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FunctionKeyExpression.java
```diff
@@ -299,7 +299,7 @@ public String getName() {
         }
 
         @Nonnull
-        abstract FunctionKeyExpression build(@Nonnull KeyExpression arguments);
+        public abstract FunctionKeyExpression build(@Nonnull KeyExpression arguments);
```
MMcM, line 302: It was previously impossible to build a `FunctionKeyExpression.Builder` outside of this package.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1351,27 +1355,44 @@ private QueryComponent normalizeAndOrForInAsOr(@Nonnull QueryComponent component
...
                 if (newChildPlan != distinctPlan.getChild()) {
                     return new RecordQueryUnorderedPrimaryKeyDistinctPlan(newChildPlan);
                 }
             }
+        } else if (chosenPlan instanceof RecordQueryFilterPlan) {
+            final RecordQueryFilterPlan filterPlan = (RecordQueryFilterPlan)chosenPlan;
+            if (filterPlan.getChild() instanceof RecordQueryPlanWithIndex) {
+                final RecordQueryPlanWithIndex filteredIndexPlan = (RecordQueryPlanWithIndex) filterPlan.getChild();
+                final Index index = metaData.getIndex(filteredIndexPlan.getIndexName());
+                if (indexTypes.getValueTypes().contains(index.getType())) {
```
MMcM, line 1378: A text index, in particular, does not have the actual indexed text in the entry.

MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
nschiefer: This is a great point. I think it's worth adding it as a comment here in case someone forgets this in the future.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1384,6 +1405,11 @@ private RecordQueryPlan tryToConvertToCoveringPlan(@Nonnull PlanContext context,
         for (KeyExpression resultField : context.query.getRequiredResults()) {
             resultFields.addAll(resultField.normalizeKeyForPositions());
         }
+        if (filterFields != null) {
+            for (KeyExpression filterField : filterFields) {
+                resultFields.addAll(filterField.normalizeKeyForPositions());
```
nschiefer, line 1410: I think that this part here only works because `findFilterCoveredFields()` ignored nested repeated fields entirely. That's fine with me, but I think it needs a comment to that effect or someone might mess this up later.

MMcM: I put the comment on `findFilterCoveredFields`; I think I captured the essence of the tricky cases.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: Added there.

PR #1009, opened 2020-08-06
Make some internals related to merging of covering indexes more accessible

MMcM:
* The `MergeCursor` family of classes / methods was package-protected, which made use outside of cursors impossible. Make them `public` / `protected` as appropriate.

* The `RecordQueryCoveringIndexPlan` result of `planCoveringAggregateIndex` can expose the lambda it uses to turn index entries into queried records.


MMcM:
Moved into #1013

PR #1011, opened 2020-08-11
Resolves #1010: Bitmap value indexes

### LINKED ISSUE #1010: Bitmap value indexes
Instead of one key per record containing the primary key, one bit per record in a fixed-size bitmap aligned on even multiples of the position key, which needs to be unique but need not be the primary key.

MMcM:
Also moves a couple of index-type-specific tests into the child package, since I was adding another one there in the main commit.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
@@ -1,0 +1,190 @@
+/*
+ * BitmapValueIndexTest.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2020 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concatenateFields;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+/**
+ * Tests for {@code BITMAP_VALUE} type indexes.
+ */
+@Tag(Tags.RequiresFDB)
+public class BitmapValueIndexTest extends FDBRecordStoreTestBase {
```
alecgrieser, line 59: I suppose there should probably be a test here that checks the behavior on uniqueness violations (making sure there's an error if the `unique` flag is true and possibly a test for if the unique flag is false that it mishandles duplicates the way we expect)

MMcM: Added two new uniqueness tests.

PR #1013, opened 2020-08-13
Resolves #1012: BITMAP_VALUE index cursors can be bit-wise merged

### LINKED ISSUE #1012: BITMAP_VALUE index cursors can be bit-wise merged
For a reasonably compact & efficient evaluation of a Boolean predicate involving their grouping keys.

MMcM:
Overall approach:
* Given a set of `BITMAP_VALUE` indexes differing in the finest grouping key.
* Given a query with:
  - Equality conditions on the coarser grouping keys.
  - A Boolean expression over conditions on the finest keys.
  - Optional inequalities on the position value.
* Produce a plan that:
  - Does covering index entry scans for each of the sub-conditions.
  - Combines the streams of bitmaps from those scans using the bit-wise analogues of the Boolean operators among them.
  - Filters out any bitmaps that end up empty (such as for `AND` with no shared positions within the position range of that bitmap) to cut down result size.

Questions for @normen662: 
* Is it correct for `ComposedBitmapIndexQueryPlan` to not have child plans or should `RecordQueryCoveringIndexPlan` really be `RecordQueryPlanWithChild` too? How do these fit into the `Quantifier` scheme?

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -1530,7 +1530,7 @@ public class RecordQueryPlanner implements QueryPlanner {
     }
     
     @Nullable
-    public RecordQueryPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull String indexName) {
+    public RecordQueryCoveringIndexPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull String indexName) {
         final Index index = metaData.getIndex(indexName);
         KeyExpression indexExpr = index.getRootExpression();
         if (indexExpr instanceof GroupingKeyExpression) {
@@ -1542,7 +1542,7 @@ public class RecordQueryPlanner implements QueryPlanner {
     }
 
     @Nullable
-    public RecordQueryPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull Index index, @Nonnull KeyExpression indexExpr) {
+    public RecordQueryCoveringIndexPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull Index index, @Nonnull KeyExpression indexExpr) {
         final Collection<RecordType> recordTypes = metaData.recordTypesForIndex(index);
         if (recordTypes.size() != 1) {
             // Unfortunately, since we materialize partial records, we need a unique type for them.
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -209,6 +209,24 @@ public class RecordQueryPlanner implements QueryPlanner {
     }
 
     /**
+     * Get the {@link RecordMetaData} for this planner.
+     * @return the meta-data
+     */
+    @Nonnull
+    public RecordMetaData getRecordMetaData() {
+        return metaData;
+    }
+
+    /**
+     * Get the {@link RecordStoreState} for this planner.
+     * @return the record store state
+     */
+    @Nonnull
+    public RecordStoreState getRecordStoreState() {
+        return recordStoreState;
+    }
+
+    /**
      * Create a plan to get the results of the provided query.
      *
      * @param query a query for records on this planner's metadata
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
@@ -20,14 +20,13 @@
 
 package com.apple.foundationdb.record.query.plan.bitmap;
 
-import com.apple.foundationdb.record.RecordMetaData;
+import com.apple.foundationdb.record.RecordStoreState;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.IndexTypes;
 import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
-import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
 import com.apple.foundationdb.record.provider.foundationdb.IndexAggregateGroupKeys;
 import com.apple.foundationdb.record.provider.foundationdb.IndexFunctionHelper;
 import com.apple.foundationdb.record.provider.foundationdb.indexes.BitmapValueIndexMaintainer;
@@ -48,6 +47,7 @@ import com.apple.foundationdb.record.query.plan.plans.RecordQueryCoveringIndexPl
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
@@ -74,6 +74,39 @@ public class ComposedBitmapIndexAggregate {
     }
 
     /**
+     * Try to build a composed bitmap plan for the given query and aggregate function.
+     * @param planner a query planner to use to construct the plans
+     * @param query a query providing target record type, filter, and required fields
+     * @param indexAggregateFunction the function giving the desired position and grouping
+     * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
+     */
+    @Nonnull
+    public static Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQueryPlanner planner,
+                                                                 @Nonnull RecordQuery query,
+                                                                 @Nonnull IndexAggregateFunction indexAggregateFunction) {
+        if (query.getFilter() == null) {
+            return Optional.empty();
+        }
+        return tryBuild(planner, query.getRecordTypes(), indexAggregateFunction, query.getFilter())
+                .flatMap(p -> p.tryPlan(planner, query.toBuilder()));
+    }
+
+    /**
+     * Try to turn this composed bitmap into an executable plan.
+     * @param planner a query planner to use to construct the plans
+     * @param queryBuilder a prototype query providing target record types and required fields
+     * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
+     */
+    @Nonnull
+    public Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQueryPlanner planner,
+                                                          @Nonnull RecordQuery.Builder queryBuilder) {
+        final List<RecordQueryCoveringIndexPlan> indexScans = new ArrayList<>();
+        final Map<IndexNode, ComposedBitmapIndexQueryPlan.IndexComposer> indexComposers = new IdentityHashMap<>();
+        return Optional.ofNullable(plan(root, queryBuilder, planner, indexScans, indexComposers))
+                .map(composer -> new ComposedBitmapIndexQueryPlan(indexScans, composer));
+    }
+
+    /**
      * Try to build a composed bitmap for the given aggregate function and filters.
      * <p>
      * The function should use a supported aggregate function (currently {@value BitmapValueIndexMaintainer#AGGREGATE_FUNCTION_NAME})
@@ -83,47 +116,36 @@ public class ComposedBitmapIndexAggregate {
      * expression that will be transformed into a set the corresponding bit operations on the bitmaps.
      * The filter can also include range conditions on the position field.
      * </p>
-     * @param recordStore the record store containing the indexed data
+     * @param planner a query planner to use to construct the plans
      * @param recordTypeNames the record types on which the indexes are defined
      * @param indexAggregateFunction the function giving the desired position and grouping
      * @param filter conditions on the groups and position
      * @return an {@code Optional} composed bitmap or {@code Optional.empty} if there conditions could not be satisfied
      */
     @Nonnull
-    public static Optional<ComposedBitmapIndexAggregate> tryBuild(@Nonnull FDBRecordStore recordStore,
-                                                                  @Nonnull List<String> recordTypeNames,
+    public static Optional<ComposedBitmapIndexAggregate> tryBuild(@Nonnull RecordQueryPlanner planner,
+                                                                  @Nonnull Collection<String> recordTypeNames,
                                                                   @Nonnull IndexAggregateFunction indexAggregateFunction,
                                                                   @Nonnull QueryComponent filter) {
-        List<QueryComponent> groupFilters = new ArrayList<>();
+        // The filters that are common to all the composed index queries.
+        // They can be equality conditions on the common group prefix (as specified by indexAggregateFunction)
+        // or inequalities on the position.
+        List<QueryComponent> commonFilters = new ArrayList<>();
+        // The filters that are specific to a single index. At present, each comparison must be accomplished by a single
+        // index. It would be possible, though, to have indexes with multiple grouping fields after the common prefix
+        // and to pick any complete covering of indexFilters.
         List<QueryComponent> indexFilters = new ArrayList<>();
-        if (!separateGroupFilters(filter, indexAggregateFunction, groupFilters, indexFilters) || indexFilters.isEmpty()) {
+        if (!separateGroupFilters(filter, indexAggregateFunction, commonFilters, indexFilters) || indexFilters.isEmpty()) {
             return Optional.empty();
         }
-        Builder builder = new Builder(recordStore, recordTypeNames, groupFilters, indexAggregateFunction);
+        Builder builder = new Builder(planner, recordTypeNames, commonFilters, indexAggregateFunction);
         return builder.tryBuild(indexFilters.size() > 1 ? Query.and(indexFilters) : indexFilters.get(0))
             .map(ComposedBitmapIndexAggregate::new);
     }
 
-    /**
-     * Try to turn this composed bitmap into an executable plan.
-     * @param query a base query over the target record types
-     * @param recordMetaData the record meta-data for planning
-     * @param planner a query planner to use to construct the plans
-     * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
-     */
-    @Nonnull
-    public Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQuery.Builder query,
-                                                          @Nonnull RecordMetaData recordMetaData,
-                                                          @Nonnull RecordQueryPlanner planner) {
-        final List<RecordQueryCoveringIndexPlan> indexScans = new ArrayList<>();
-        final Map<IndexNode, ComposedBitmapIndexQueryPlan.IndexComposer> indexComposers = new IdentityHashMap<>();
-        return Optional.ofNullable(plan(root, query, recordMetaData, planner, indexScans, indexComposers))
-                .map(composer -> new ComposedBitmapIndexQueryPlan(indexScans, composer));
-    }
-
     private static boolean separateGroupFilters(@Nonnull QueryComponent filter,
                                                 @Nonnull IndexAggregateFunction indexAggregateFunction,
-                                                @Nonnull List<QueryComponent> groupFilters,
+                                                @Nonnull List<QueryComponent> commonFilters,
                                                 @Nonnull List<QueryComponent> indexFilters) {
         QueryToKeyMatcher matcher = new QueryToKeyMatcher(filter);
         FilterSatisfiedMask filterMask = FilterSatisfiedMask.of(filter);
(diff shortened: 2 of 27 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1530,7 +1530,7 @@ private static boolean addCoveringField(@Nonnull KeyExpression requiredExpr,
     }
     
     @Nullable
-    public RecordQueryPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull String indexName) {
+    public RecordQueryCoveringIndexPlan planCoveringAggregateIndex(@Nonnull RecordQuery query, @Nonnull String indexName) {
```
normen662, line 1533: This should be considered a binary-incompatible change of the stable API.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryCoveringIndexPlan.java
```diff
@@ -73,19 +75,25 @@ public RecordQueryCoveringIndexPlan(@Nonnull RecordQueryPlanWithIndex indexPlan,
...
+        return indexPlan
+                .executeEntries(store, context, continuation, executeProperties)
+                .map(indexEntryToQueriedRecord(store));
+    }
+
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    @API(API.Status.INTERNAL)
+    public <M extends Message> Function<IndexEntry, FDBQueriedRecord<M>> indexEntryToQueriedRecord(final @Nonnull FDBRecordStoreBase<M> store) {
         final RecordMetaData metaData = store.getRecordMetaData();
```
normen662, line 91: all these getters are now done per record and could be done once per execute(). Not sure if in one of these you have anything computationally expensive going on...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: I added a note in the release notes. Do we think stronger action is needed?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+        Builder(@Nonnull final FDBRecordStore recordStore, @Nonnull List<String> recordTypeNames,
+                @Nonnull List<QueryComponent> groupFilters, @Nonnull IndexAggregateFunction indexAggregateFunction) {
+            this.recordStore = recordStore;
+            this.recordTypeNames = recordTypeNames;
+            this.groupFilters = groupFilters;
+            this.indexAggregateFunction = indexAggregateFunction;
+        }
+
+        @Nonnull
+        Optional<Node> build(@Nonnull QueryComponent indexFilter) {
```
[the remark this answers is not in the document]
MMcM: Sure. Changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexQueryPlan.java
```diff
@@ -0,0 +1,503 @@
...
+        return true;
+    }
+
+    @Override
+    public int hashCodeWithoutChildren() {
+        return composer.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
MMcM: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexQueryPlan.java
```diff
@@ -0,0 +1,503 @@
...
+            return false;
+        }
+        ComposedBitmapIndexQueryPlan that = (ComposedBitmapIndexQueryPlan) o;
+        return indexPlans.equals(that.indexPlans) &&
+                composer.equals(that.composer);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(indexPlans, composer);
```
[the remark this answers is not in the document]
MMcM: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryCoveringIndexPlan.java
MMcM: It doesn't map calls to `indexEntryToQueriedRecord` over them; it maps the _result_ of `indexEntryToQueriedRecord`, which is a function, closing over all those variables, over them.

normen662: ah my bad

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: @alecgrieser Do we need a release bump? It's source compatible, but would break on binary level.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+                        .map(childNode -> new OperatorNode(OperatorNode.Operator.NOT, Collections.singletonList(childNode)));
+            }
+            return Optional.empty();
+        }
+
+        @Nonnull
+        Optional<Node> indexScan(@Nonnull QueryComponent indexFilter) {
+            if (bitmapIndexes == null) {
+                bitmapIndexes = findBitmapIndexes(indexAggregateFunction.getName());
+                indexNodes = new HashMap<>();
```
alecgrieser, line 277: I suppose the construction of `indexNodes` should be delayed until after `bitmapIndexes.isEmpty()` is checked... Not that I expect that change will make any, like, measurable difference in performance

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: Good question. (I'm not sure if there's a settled answer on that.) I guess technically we only make guarantees about our API stability (and not our ABI stability). This...feels like it's probably fine to just make a note of, I suppose, but I'm not certain.

It seems a little weird to me that this method would return a concrete implementation of `RecordQueryPlan` rather than an interface (from an API perspective), but I've not yet gone through the code enough to understand the justification for that change yet.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: Sure. Not much of a difference, it's it's only a matter of reordering the statements.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexCursor.java
```diff
@@ -0,0 +1,165 @@
...
+                        final IndexEntry indexEntry = cursorState.getResult().get();
+                        final Tuple indexKey = indexEntry.getKey();
+                        final long position = indexKey.getLong(indexKey.size() - 1);
+                        if (nextPosition > position) {
+                            resultStates.clear();
+                            nextPosition = position;
+                        }
+                        if (nextPosition == position) {
+                            resultStates.add(cursorState);
+                        }
```
alecgrieser, line 102: Hm, this doesn't seem right to me. In particular, I think if you had disjoint ranges of positions for the physical bitmaps from different indexes, then I think it won't merge them together correctly, if I understand how those work. (This doesn't matter for `AND`s, but for `OR`s and `NOT`s, etc.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexCursor.java
```diff
@@ -0,0 +1,165 @@
...
+    protected IndexEntry getNextResult(@Nonnull List<MergeCursorState<IndexEntry>> resultStates) {
+        final List<MergeCursorState<IndexEntry>> cursorStates = getCursorStates();
+        final IndexEntry firstEntry = resultStates.get(0).getResult().get();
+        final int size = firstEntry.getValue().getBytes(0).length;
+        final List<byte[]> bitmaps = new ArrayList<>(cursorStates.size());
+        for (MergeCursorState<IndexEntry> cursorState : cursorStates) {
+            if (resultStates.contains(cursorState)) {
+                byte[] bitmap = cursorState.getResult().get().getValue().getBytes(0);
+                if (bitmap.length != size) {
+                    throw new RecordCoreException("Index bitmaps are not all the same size");
```
alecgrieser, line 123: Should it be part of the planning that the parameter for the size needs to be the same? It seems like the "ideal" version would figure it out then, but maybe it's sufficient for the time being for it to be an error here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexCursor.java
```diff
@@ -0,0 +1,165 @@
...
+        return new ComposedBitmapIndexContinuation(getChildContinuations(), null);
+    }
+
+    @Nonnull
+    public static ComposedBitmapIndexCursor create(@Nonnull List<Function<byte[], RecordCursor<IndexEntry>>> cursorFunctions,
+                                                   @Nonnull Composer composer,
+                                                   @Nullable byte[] byteContinuation,
+                                                   @Nullable FDBStoreTimer timer) {
+        if (cursorFunctions.size() < 2) {
+            throw new RecordCoreArgumentException("not enough child cursors provided to UnionCursor")
```
alecgrieser, line 152:
```suggestion
            throw new RecordCoreArgumentException("not enough child cursors provided to ComposedBitmapIndexCursor")
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
```diff
@@ -242,18 +265,176 @@ public void uniquenessViolationNotChecked() {
         }
     }
 
+    @Test
```
alecgrieser, line 268: I think this could use a few more tests. Namely:

1. Something that uses `not`. If there's a way to get an XOR, then that would be nice too, but I don't think we have that in our API (nor do we simplify "(a &or; b) &and; &not; (a &and; b)" into "a &oplus; b", I don't think).
1. Something that has a union of elements with disjoint physical bitmaps in the index. Something like (using a pseudo-SQL syntax) `SELECT bitmap(rec_no) FROM MySimpleRecord WHERE num_value_2 = 3 OR num_value_2 = 4` where all of the records with `num_value_2 = 3` are in one position group and all of the records where `num_value_2 = 3` are in a different one.

MMcM: Added a test for `NOT`, which can be supported.

Added a test for non-overlapping indexes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+            final KeyExpression additionalKey;
+            if (indexFilter instanceof FieldWithComparison) {
+                additionalKey = Key.Expressions.field(((FieldWithComparison) indexFilter).getFieldName());
+            } else if (indexFilter instanceof QueryKeyExpressionWithComparison) {
+                additionalKey = ((QueryKeyExpressionWithComparison) indexFilter).getKeyExpression();
+            } else {
+                return Optional.empty();
+            }
+            GroupingKeyExpression groupKeyExpression = (GroupingKeyExpression)indexAggregateFunction.getOperand();
+            GroupingKeyExpression fullKey = Key.Expressions.concat(groupKeyExpression.getGroupingSubKey(), additionalKey, groupKeyExpression.getGroupedSubKey())
```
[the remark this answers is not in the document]
MMcM: Renamed to `indexKey`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+                case OR:
+                    return new ComposedBitmapIndexQueryPlan.OrComposer(children);
+                case NOT:
+                    return new ComposedBitmapIndexQueryPlan.NotComposer(children.get(0));
+                default:
+                    throw new IllegalArgumentException("Unknown operator node: " + node);
+            }
+        } else if (node instanceof IndexNode) {
+            return indexComposers.computeIfAbsent((IndexNode)node, indexNode -> {
+                query.setFilter(indexNode.filter);
```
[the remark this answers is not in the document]
MMcM: The builder is now only reused in the advanced case and there is a comment about this.,

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+
+    /**
+     * Try to turn this composed bitmap into an executable plan.
+     * @param query a base query over the target record types
+     * @param recordMetaData the record meta-data for planning
+     * @param planner a query planner to use to construct the plans
+     * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
+     */
+    @Nonnull
+    public Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQuery.Builder query,
```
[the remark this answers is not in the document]
MMcM: Renamed to `queryBuilder` and java doc updated. The "simpler" entry takes a whole `RecordQuery` now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+     * The filter can also include range conditions on the position field.
+     * </p>
+     * @param recordStore the record store containing the indexed data
+     * @param recordTypeNames the record types on which the indexes are defined
+     * @param indexAggregateFunction the function giving the desired position and grouping
+     * @param filter conditions on the groups and position
+     * @return an {@code Optional} composed bitmap or {@code Optional.empty} if there conditions could not be satisfied
+     */
+    @Nonnull
+    public static Optional<ComposedBitmapIndexAggregate> tryBuild(@Nonnull FDBRecordStore recordStore,
```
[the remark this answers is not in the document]
MMcM: I added some more public methods elsewhere so that it could take a planner and a query as the basic entry and made the test fixture use that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexQueryPlan.java
```diff
@@ -0,0 +1,496 @@
...
+                if (operand == null) {
+                    return null;
+                }
+                if (first) {
+                    System.arraycopy(operand, 0, result, 0, result.length);
+                    empty = first = false;
+                } else {
+                    empty = true;
+                    for (int j = 0; j < result.length; j++) {
+                        final byte b = (byte) (result[j] & operand[j]);
```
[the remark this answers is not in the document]
MMcM: The handling of a null bitmap is also different between `AND` and `[X]OR`. I am concerned about JIT doing as well with a virtual in the middle of the inner loop over the bitmap. I added a comment.

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -93,4 +93,13 @@ message OrElseContinuation {
...
     optional bytes continuation = 2;
-}
\ No newline at end of file
+}
+
+message ComposedBitmapIndexContinuation {
+    message CursorState {
+        optional bytes continuation = 1;
+        optional bool exhausted = 2;
+        optional bytes bloom_filter = 3;
```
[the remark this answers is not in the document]
MMcM: Accident.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -0,0 +1,335 @@
...
+        // The position key(s) can also be constrained with inequalities and those go among the group filters.
+        matcher.matchesCoveringKey(((GroupingKeyExpression)indexAggregateFunction.getOperand()).getGroupedSubKey(), filterMask);
+        if (filterMask.allSatisfied()) {
+            return false;   // Not enough conditions left over.
+        }
+        for (FilterSatisfiedMask child : filterMask.getChildren()) {
+            if (child.allSatisfied()) {
+                groupFilters.add(child.getFilter());
+            } else {
+                indexFilters.add(child.getFilter());
```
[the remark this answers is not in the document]
MMcM: A filter can only be satisfied if it matches a key in what was given to the matcher, first the common prefix conditions from the aggregate function and second the position key. I renamed `groupFilters` to `commonFilters` and added some comments to this effect.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexQueryPlan.java
```diff
@@ -0,0 +1,496 @@
...
+                .collect(Collectors.toList());
+        return ComposedBitmapIndexCursor.create(cursorFunctions, composer, continuation, store.getTimer())
+                // Composers can return null bitmaps when empty, which is then left out of the result set.
+                .filter(indexEntry -> indexEntry.getValue().get(0) != null)
+                .map(indexPlans.get(0).indexEntryToQueriedRecord(store));
+    }
+
+    @Override
+    public boolean isReverse() {
+        return false;
```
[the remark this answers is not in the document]
MMcM: Yes, I think it's kind of ill-defined. Like does the whole bitmap reverse or just the enumeration order?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexCursor.java
MMcM: Did you have a particular failure scenario in mind? In `computeNextResultStates` we should get all the states that share a min next position. Then, in `getNextResult`, we collect the bitmap from those and `null` from the rest. `Composer.compose` treats `null` like an empty bitmap, that is one that is all zeros, that is one that has no positions set.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -73,6 +73,39 @@
...
+     * @param indexAggregateFunction the function giving the desired position and grouping
+     * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
+     */
+    @Nonnull
+    public static Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQueryPlanner planner,
+                                                                 @Nonnull RecordQuery query,
+                                                                 @Nonnull IndexAggregateFunction indexAggregateFunction) {
+        if (query.getFilter() == null) {
+            return Optional.empty();
+        }
```
alecgrieser, line 89: Should this also reject the query if `query.getSort() != null`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexQueryPlan.java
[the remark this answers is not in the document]
alecgrieser: Yeah. I...think it would reverse the whole bitmap, though I suppose in some sense, that's a little Endian vs big Endian debate. In any case, leaving it as not allowed (at least for now) rather than trying to settle that debate seems like the right call to me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: Yes, that seems like good added safety.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexCursor.java
MMcM: I added a comment for what it's trying to accomplish.

Somewhere along the line I developed a preference for
```
if (x <op> y) {
  x = y;
}
```

alecgrieser:
Test failure appears to be another one of #936 

test this please

Issue #1002, opened 2020-07-23
Operation logging for debugging purposes


alecgrieser:
For several reasons (including being able to get a full list of operations for each transaction to diagnose transaction conflicts), it would be nice to be able to log the operations from a transaction. Now, these logs are already available through the FDB trace logs, but logging at the record layer could have several benefits:

1. Integration with the same (slf4j) logging framework as the library adopter (rather than using FDB trace logs, which are more persnickety)
1. Ability to translate opaque byte arrays into concept that are more human readable (like tuples or key spaces or translating index space keys)

I believe this could be accomplished with a new `Transaction` implementation that kept track of conflict ranges and operations. Then this information could be logged along with the commit if `.logTransaction` is set. At first, this could just log opaque byte arrays, but it could be turned into something more sophisticated later.

alecgrieser:
Oh, for the latter part (which doesn't need to be done before this issue is done) might benefit from taking on the work done for #816 to produce key spaces and more from the bytes

alecgrieser:
See also: #48

Issue #564, opened 2019-05-02
FDBRecordStoreIndexTest.boundaryPrimaryKeys() is flaky


nschiefer:
On some local machines (at least), the `FDBRecorStoreIndexTest.boundaryPrimaryKeys()` fails (pretty consistently, even) with:

```
the test is meaningless if the records are not across boundaries ==> expected: <true> but was: <false>
org.opentest4j.AssertionFailedError: the test is meaningless if the records are not across boundaries ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:42)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:146)
	at com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreIndexTest.boundaryPrimaryKeys(FDBRecordStoreIndexTest.java:2329)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:436)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$6(TestMethodTestDescriptor.java:170)
	at org.junit.jupiter.engine.execution.ThrowableCollector.execute(ThrowableCollector.java:40)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:166)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:113)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:58)
[...]

PR #1015, opened 2020-08-14
Resolves #1019: Avoid repeating index endpoints conflicts

### LINKED ISSUE #1019: Avoid range conflicts while indexing (if possible)
While setting endpoints for indexer, repeating conflicts with other transaction could prevent the job from running. This can be fixed for idempotent indexes by using per-record conflicts instead of ranges'.

jjezra:
While setting endpoints for indexer, repeating conflicts with other transaction could prevent the job from running.

   resolution: Since the store is marked WRITE_ONLY while determining endpoints and reading the records,
   is seems safe to ignore conflicts caused by insertion of new records - as the inseration should already
   imply valid index entries.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -30,7 +30,7 @@ Additionally, builds for the project now require JDK 11. The project is still ta
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Avoid repeating index endpoints conflicts [(Issue #1019)](https://github.com/FoundationDB/fdb-record-layer/issues/1019)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -538,8 +538,12 @@ public class OnlineIndexer implements AutoCloseable {
             throw new MetaDataException("Store does not have the same metadata");
         }
         final IndexMaintainer maintainer = store.getIndexMaintainer(index);
+        final boolean isIdempotent = maintainer.isIdempotent();
         final ExecuteProperties.Builder executeProperties = ExecuteProperties.newBuilder()
-                .setIsolationLevel(IsolationLevel.SERIALIZABLE);
+                .setIsolationLevel(
+                        isIdempotent ?
+                        IsolationLevel.SNAPSHOT :
+                        IsolationLevel.SERIALIZABLE);
         if (respectLimit) {
             executeProperties.setReturnedRowLimit(limit);
         }
@@ -588,6 +592,9 @@ public class OnlineIndexer implements AutoCloseable {
                 return AsyncUtil.READY_TRUE;
             }
             // add this index to the transaction
+            if (isIdempotent) {
+                store.addRecordReadConflict(rec.getPrimaryKey());
+            }
             if (timer != null) {
                 timer.increment(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
             }
@@ -1015,11 +1022,19 @@ public class OnlineIndexer implements AutoCloseable {
     @Nonnull
     private CompletableFuture<TupleRange> buildEndpoints(@Nonnull FDBRecordStore store, @Nonnull RangeSet rangeSet,
                                                          @Nullable AtomicLong recordsScanned) {
+        boolean isIdempotent = store.getIndexMaintainer(index).isIdempotent();
+        final IsolationLevel isolationLevel =
+                isIdempotent ?
+                // if idempotent: since double indexing is harmless, we can skip the range conflict protection. At worse,
+                // some records will be re-indexed.
+                IsolationLevel.SNAPSHOT :
+                IsolationLevel.SERIALIZABLE;
         final ExecuteProperties limit1 = ExecuteProperties.newBuilder()
                 .setReturnedRowLimit(1)
-                .setIsolationLevel(IsolationLevel.SERIALIZABLE)
+                .setIsolationLevel(isolationLevel)
                 .build();
         final ScanProperties forward = new ScanProperties(limit1);
+
         RecordCursor<FDBStoredRecord<Message>> beginCursor = store.scanRecords(recordsRange, null, forward);
         CompletableFuture<Tuple> begin = beginCursor.onNext().thenCompose(result -> {
             if (result.hasNext()) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -1025,8 +1025,10 @@ public class OnlineIndexer implements AutoCloseable {
         boolean isIdempotent = store.getIndexMaintainer(index).isIdempotent();
         final IsolationLevel isolationLevel =
                 isIdempotent ?
-                // if idempotent: since double indexing is harmless, we can skip the range conflict protection. At worse,
-                // some records will be re-indexed.
+                // If idempotent: since double indexing is harmless, we can use individual records protection instead of
+                // a range conflict one - which means that new records, added to the range while indexing the SNAPSHOT,
+                // will not cause a conflict during the commit. At worse, few records (if added after marking WRITE_ONLY but
+                // before this method's query) will be re-indexed.
                 IsolationLevel.SNAPSHOT :
                 IsolationLevel.SERIALIZABLE;
         final ExecuteProperties limit1 = ExecuteProperties.newBuilder()
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
@@ -23,6 +23,7 @@ package com.apple.foundationdb.record.provider.foundationdb;
 import com.apple.foundationdb.record.TestRecords1Proto;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.IndexTypes;
+import com.apple.foundationdb.record.metadata.Key;
 import com.apple.test.Tags;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -100,11 +101,11 @@ public class OnlineIndexerConflictsTest extends OnlineIndexerTest {
                 ).collect(Collectors.toList());
 
         List<TestRecords1Proto.MyOtherRecord> otherRecords =
-                LongStream.range(0, 7).mapToObj(val -> TestRecords1Proto.MyOtherRecord.newBuilder().setRecNo(val * 2 + 1).setNumValue2((int) val + 1).build()
+                LongStream.range(0, 7).mapToObj(val -> TestRecords1Proto.MyOtherRecord.newBuilder().setRecNo(val * 2 + 1).setNumValue2((int)val + 1).build()
                 ).collect(Collectors.toList());
 
         List<TestRecords1Proto.MyOtherRecord> otherRecordsOverwrite =
-                LongStream.range(0, 7).mapToObj(val -> TestRecords1Proto.MyOtherRecord.newBuilder().setRecNo(val * 2 + 1).setNumValue2((int) val + 101).build()
+                LongStream.range(0, 7).mapToObj(val -> TestRecords1Proto.MyOtherRecord.newBuilder().setRecNo(val * 2 + 1).setNumValue2((int)val + 101).build()
                 ).collect(Collectors.toList());
 
         Index index = new Index("newIndex", field("num_value_2"), IndexTypes.VALUE);
@@ -147,7 +148,7 @@ public class OnlineIndexerConflictsTest extends OnlineIndexerTest {
     }
 
     @Test
-    public void testAddRecordToRangeWhileIndexedIdempotentFailure() {
+    public void testModifyRecordInRangeWhileIndexedIdempotentFailure() {
 
         List<TestRecords1Proto.MySimpleRecord> records =
                 LongStream.range(0, 20).mapToObj(val -> TestRecords1Proto.MySimpleRecord.newBuilder().setRecNo(val).setNumValue2((int)val + 1).build()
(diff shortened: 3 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildSumIndexTest.java
```diff
@@ -103,6 +103,9 @@ public void run() {
                 try (FDBRecordContext context = openContext()) {
                     long sum = recordStore.evaluateAggregateFunction(Collections.singletonList("MySimpleRecord"), aggregateFunction, TupleRange.ALL, IsolationLevel.SNAPSHOT).join().getLong(0);
                     long expected = updatedRecords.stream().mapToInt(msg -> msg.hasNumValue2() ? msg.getNumValue2() : 0).sum();
+                    for (int i = 0; sum != expected && i < 20; i ++) {
+                        Thread.yield();
```
alecgrieser, line 107: Hm, what is this line trying to accomplish?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -539,7 +539,7 @@ private void increaseLimit() {
         }
         final IndexMaintainer maintainer = store.getIndexMaintainer(index);
         final ExecuteProperties.Builder executeProperties = ExecuteProperties.newBuilder()
-                .setIsolationLevel(IsolationLevel.SERIALIZABLE);
+                .setIsolationLevel(IsolationLevel.SNAPSHOT);
```
alecgrieser, line 542: Hm, perhaps this is a sign with our tests if it's not catching a bug here, but I think this needs to validate that the index is idempotent before it sets the isolation level to `SNAPSHOT`. Otherwise, there's not a guarantee that the concurrently written record will be indexed.

We should probably have a comment here explaining why `SNAPSHOT` is used, and then another explaining why we need to add the read conflict range at the place we do.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildSumIndexTest.java
jjezra: This meaningless code made the sum test pass, I do not understand why. 
After distinguishing between idempotent and accumulative indexes, the sum test passes without this ghost hunting.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: The sum index test did fail, do not lose hope :) 
You did mention before that the SNAPSHOT approach must be limited to idempotent indexes only, but I failed to assimilate (sorry). I'll fix it now.

jjezra:
@alecgrieser, please note that the tests you've asked are still missing. I'll add them in another commit.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -0,0 +1,205 @@
...
+            for (int i = 50; i <= 150; i++) {
+                // even numbers 100-300 (that's 101 values)
+                recordStore.saveRecord(records.get(i * 2));
+            }
+            context.commit();
+        }
+
+        openSimpleMetaData(hook);
+        try (FDBRecordContext context = openContext()) {
+            assertTrue(recordStore.uncheckedMarkIndexReadable(index.getName()).join());
```
alecgrieser, line 77: Hmm. IIUC from offline conversations, this was needed to make future commits down the line not fail with a conflict, but this call seems suspicious to me. It essentially marks the index as already built, so the index build won't do much of anything, so it seems more likely to me that this is either masking some other problem, or there is some issue or is a sign that there is some other kind of misconfiguration going on with these tests.

jjezra: I agree. This line was copied from the tests in OnlineIndexerSimpleTest. Would you have any idea why the indexer commit was failing without this line?

alecgrieser: I believe all of the instances of that method being used in `OnlineIndexerSimpleTest`, it does so so that it can then scan the index without either (1) being told that the index isn't readable (which is what happens if you scan an unbuilt index normally) or (2) building the index and therefore disturbing it. It also looks like in all of the instances (that I could find) in `OnlinIndexerSimpleTest`, the transaction that calls `uncheckedMarkIndexReadable` isn't committed (and indeed, there are comments saying `// do NOT commit` instead of calls to `commit()`).

I'm not sure why it's currently failing without that change, though. I had some theories, but most of the theories ended up being falsified by further experiments that you did. It's possible this needs more investigation, which might mean turning on detailed FDB client trace logging, or possibly investing in #1002 (or something temporary along those lines) to see what the conflict is.

jjezra: I've added the test you've suggested above. Here I could mark the index as write only.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
         assertEquals(0, timer.getCount(FDBStoreTimer.Events.COMMIT_FAILURE));
     }
+
+    @Test
+    public void testOnlineIndexerIdempotentNested() throws Exception {
+        // If a conflict happens, this test should succeed by retries
+        Random rand = new Random();
+        final long seed = rand.nextInt();
+        LOGGER.info("Random seed is " + seed);
+        rand = new Random(seed); // make rand repeatable (note that multi threading behaviour is not fully repeatable in this test)
```
alecgrieser, line 211: I don't think this comment applies to this test any more

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
+
+        List<TestRecords1Proto.MySimpleRecord> records =
+                LongStream.range(0, 400).mapToObj(val -> TestRecords1Proto.MySimpleRecord.newBuilder().setRecNo(val).setNumValue2((int)val + 1).build()
+                ).collect(Collectors.toList());
```
alecgrieser, line 215: I don't think this test needs quite as many records now. I think...three would suffice? The main tests have lots of records in order to make sure that during the random operations, we get lots of different combinations. That shouldn't be necessary with this test, though, as it's more about testing specific concurrent operations.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
+            for (int i = 50; i <= 150; i++) {
+                // even numbers 100-300 (that's 101 values)
+                recordStore.saveRecord(records.get(i * 2));
+            }
+            context.commit();
+        }
+
+        openSimpleMetaData(hook);
+        try (FDBRecordContext context = openContext()) {
+            recordStore.markIndexWriteOnly(index.getName()).join();
```
alecgrieser, line 230: Hm, it's weird that you need to mark the index as write only.... It _should_ be marking the index as write only when the record store is opened, I think, unless I'm misunderstanding how this all works.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
+            context.commit();
+        }
+
+        int[] inserts = {99, 101, 133, 201, 350, 299, 301};
+        for (int i = 0; i < inserts.length; i++) {
```
alecgrieser, line 235: I don't think doing this in a loop really adds much. Probably sufficient to just have the one insert. As written, this also builds the full index repeatedly within the loop, which probably isn't quite what we want.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
+            int record_i = inserts[i];
+
+            final FDBStoreTimer timer = new FDBStoreTimer();
+            LOGGER.info("newIndex thread run");
```
alecgrieser, line 239:
```suggestion
```

Now that it's not it's own thread

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
         assertSame(recordStore.getIndexState(index.getName()), IndexState.READABLE);
         assertEquals(0, timer.getCount(FDBStoreTimer.Events.COMMIT_FAILURE));
     }
+
+    @Test
+    public void testOnlineIndexerIdempotentNested() throws Exception {
+        // If a conflict happens, this test should succeed by retries
```
alecgrieser, line 207: But...a conflict shouldn't happen?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
...
+                    indexer.buildRange(recordStore, null, null).join();
+                    try (FDBRecordContext context2 = openContext()) {
+                        recordStore.saveRecord(records.get(record_i));
+                        context2.commit();
+                    }
+                    context1.commit();
+                }
+            }
+
+            assertEquals(0, timer.getCount(FDBStoreTimer.Events.COMMIT_FAILURE));
```
alecgrieser, line 256: This assert isn't hurting anything, I suppose, but it's also already covered by the two contexts committing successfully above. It may be worth adding a comment on `context1.commit()` about how a record is being added in the range added by the indexer, but the transaction still commits because it's value isn't needed by the indexer.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
```diff
@@ -201,5 +201,60 @@ public void run() {
         assertSame(recordStore.getIndexState(index.getName()), IndexState.READABLE);
         assertEquals(0, timer.getCount(FDBStoreTimer.Events.COMMIT_FAILURE));
     }
+
+    @Test
+    public void testOnlineIndexerIdempotentNested() throws Exception {
```
alecgrieser, line 206: Hm, maybe something more like `testAddRecordToRangeWhileIndexedIdempotent`? Or something with those words?

I don't think we want "nested" in the title. We typically use that to refer to nested key expressions (i.e., indexes defined on elements of nested messages), and having it here might cause confusion.

alecgrieser:
Hm, strangely, the test error appears to be a repro of #564. @nblintao You wouldn't happen to know if that test got changed back to use an `assert` instead of an `assume` somehow?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
jjezra: Actually that is intentional, I was trying to add a single record each time outside the range (from both sides), next the the first/last items, and in the middle of it. The result is repeatedly adding a record while repeatedly indexing.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
jjezra: Sure. Shall I keep the other two (random) tests?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
jjezra: Omit the writeOnly but keeping the commit works!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
jjezra: <Oops>

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
alecgrieser: I believe that the way that `OnlineIndexer.buildRange` works, it'll build the whole range, so I think what this test is currently doing is:

1. The first iteration, building the index and inserting a record in separate transactions and asserting that they both succeed
1. All subsequent iterations, inserting a record and then validating that the index doesn't need to be built because it already has

So, I'm not quite sure it's testing what's desired, unless I'm misunderstanding the goal here.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
alecgrieser: I don't think so, really. I think the tests in `OnlineIndexerBuildValueIndexTest` and `OnlineIndexerBuildSumIndexTest` are essentially trying to test the same thing, and so I think we'd want to stick with those (or add more tests there if we think they are insufficient).

I'd also think we'd want to track down the weird conflict if we kept them in.

jjezra: Ok.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerConflictsTest.java
jjezra: Is there an easy way to delete the index?

alecgrieser: If you create a new record store and call `clearAndMarkIndexWriteOnly`, that will delete all of the index's data, if that's what you want to do.

jjezra: Got it: 
        FDBRecordStoreTestBase.RecordMetaDataHook hookDel = metaDataBuilder -> metaDataBuilder.removeIndex(index.getName());

jjezra: Sorry I didn't see your reply before submitting (my browser, as it seems, does not refresh frequently enough). Shall I change the removeIndex to clearAndMarkIndexWriteOnly?

alecgrieser: It depends on what you are trying to accomplish. The `removeIndex` method will...remove the index from the meta-data (so it will require adding it back to then build the index, if that's the idea). Whereas the method I gave simply "clears out" the index. I _think_ you just want to clear it out here, but I could be misunderstanding the desired effect.

jjezra: Either one will work for my desired affect (which is to add a record in different locations while indexing). Would you be willing to review it as is?

docs/ReleaseNotes.md
```diff
@@ -30,7 +30,7 @@ Additionally, builds for the project now require JDK 11. The project is still ta
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Avoid repeating index endpoints conflicts #1015 [(Issue #1015)](https://github.com/FoundationDB/fdb-record-layer/issues/1015)
```
alecgrieser, line 33:
```suggestion
* **Feature** Avoid repeating index endpoints conflicts [(Issue #1015)](https://github.com/FoundationDB/fdb-record-layer/issues/1015)
```

docs/ReleaseNotes.md
```diff
@@ -30,7 +30,7 @@ Additionally, builds for the project now require JDK 11. The project is still ta
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Avoid repeating index endpoints conflicts [(Issue #1015)](https://github.com/FoundationDB/fdb-record-layer/issues/1015)
```
nblintao, line 33: This is supposed to be Issue number not PR number. You could create an Issue from https://github.com/FoundationDB/fdb-record-layer/issues/new/choose

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1015,11 +1022,19 @@ public void rebuildIndex(@Nonnull FDBRecordStore store) {
     @Nonnull
     private CompletableFuture<TupleRange> buildEndpoints(@Nonnull FDBRecordStore store, @Nonnull RangeSet rangeSet,
                                                          @Nullable AtomicLong recordsScanned) {
+        boolean isIdempotent = store.getIndexMaintainer(index).isIdempotent();
+        final IsolationLevel isolationLevel =
+                isIdempotent ?
+                // if idempotent: since double indexing is harmless, we can skip the range conflict protection. At worse,
```
nblintao, line 1028: It's a good place to illustrate on "range conflict protection": allow records being added to the range (so they may be or may not be seen by indexer) while scanning.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1015,11 +1022,19 @@ public void rebuildIndex(@Nonnull FDBRecordStore store) {
...
+                // some records will be re-indexed.
+                IsolationLevel.SNAPSHOT :
+                IsolationLevel.SERIALIZABLE;
         final ExecuteProperties limit1 = ExecuteProperties.newBuilder()
                 .setReturnedRowLimit(1)
-                .setIsolationLevel(IsolationLevel.SERIALIZABLE)
+                .setIsolationLevel(isolationLevel)
```
nblintao, line 1034: Even though only one record is scanned, I think that record should be added to read conflict.

I wouldn't worry about reproducing it from tests though.

alecgrieser (approved):

PR #992, opened 2020-07-14
Resolves #991 Delay Index Fetches to minimize # of base records loaded

jleach4:
Generated plans do not minimize the number of expensive index lookups.

 Union Optimization

 *                                  Starting Plan
 *
 *                                      UnionPlan
 *                                  /              \
 *                                 /                \
 *                 RecordQueryPlan (IndexFetch) RecordQueryPlan (IndexFetch)
 *
 *                                          |
 *                                          |
 *                                          V
 *
 *                                  Transformed Plan
 *
 *                                  UnionPlan (IndexFetch)
 *                                  /              \
 *                                 /                \
 *                         RecordQueryPlan   RecordQueryPlan

Intersection Optimization

 *                                  Starting Plan
 *
 *                                  IntersectionPlan
 *                                  /              \
 *                                 /                \
 *                 RecordQueryPlan (IndexFetch) RecordQueryPlan (IndexFetch)
 *
 *                                          |
 *                                          |
 *                                          V
 *
 *                                  Transformed Plan
 *
 *                                  IntersectionPlan (IndexFetch)
 *                                  /              \
 *                                 /                \
 *                         RecordQueryPlan   RecordQueryPlan

UnorderdPrimaryKeyDistinct

 *                               UnorderedPrimaryKeyDistinct
 *                                          |
 *                                          |
 *                             RecordQueryPlanWithIndex (Index Lookup)
 *
 *                                         ||
 *                                         ||
 *                                         V
 *
 *                               UnorderedPrimaryKeyDistinct (Index Lookup)
 *                                          |
 *                                          |
 *                             RecordQueryPlanWithIndex (Covering Index)
 *
 *
￼

PR #1018, opened 2020-08-18
Resolves #1016: Covering filters on legs of unions/intersections.

### LINKED ISSUE #1016: Filter covering optimizations should work for legs of unions/intersections
In #1001, we added an optimization that allows filters to be executed on partial records produced by covering index scans. In #992, we also added the ability to perform union/intersection/primary key distinct operations before fetching the base records by primary key.

It would be nice to combine these so that we can do covering filter optimizations on legs of unions/intersections with delayed fetches.

### LINKED ISSUE #1017: Covering text index scans return tokens, not requested full text strings
Currently, the planner thinks that full text index entries contain the whole text, when they only contain individual tokens. Therefore, a query that requires the text fields often results in a covering index scan with the "text" being just a single token.

nschiefer:
In #1001, we added an optimization that allows filters to be executed on partial records produced by covering index scans. In #992, we also added the ability to perform union/intersection/primary key distinct operations before fetching the base records by primary key.

This refactors the work from #1001 into a substitution visitor that can work in concert with the union/intersection/distinct visitors. It also corrects some (untriggered) bugs in the existing visitors.

This also fixes #1017; full text fields now require looking at the primary records and are not "satisfied" by the tokens in the full text index entries.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Represents a subset of the fields available in a stream of records, including partial records.
+ *
+ * If a stream includes partial records, all fields are available; they are not represented individually.
```
MMcM, line 47: Does this mean _full_ (not-partial) records, where all fields are available by navigation of said record?

I guess I am unclear whether this comment is speaking generally of scan streams or more specifically of how this class is used.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -0,0 +1,211 @@
...
+
+    @Nonnull
+    public static AvailableFields intersection(@Nonnull List<AvailableFields> toIntersect) {
+        Map<KeyExpression, FieldData> intersection = null;
+        for (AvailableFields fields : toIntersect) {
+            if (!fields.hasAllFields()) {
+                if (intersection == null) {
+                    intersection = new HashMap<>(fields.fields);
+                } else {
+                    Set<KeyExpression> keysToRemove = new HashSet<>(Sets.difference(intersection.keySet(),
```
MMcM, line 180: `Map.keySet` is backed by the map itself, so I think `retainAll` should work here.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
```diff
@@ -1015,10 +1023,12 @@ public void twoRankPredicates() throws Exception {
                         Query.rank(Key.Expressions.field("score").groupBy(Key.Expressions.field("gender"))).equalsValue(1L)))
                 .build();
         RecordQueryPlan plan = planner.plan(query);
-        assertEquals("Index(rank_by_gender [[M, 1],[M, 1]] BY_RANK)" +
-                        " | score LESS_THAN $__rank_0" +
-                        " WHERE __rank_0 = BasicRankedRecord$score.score_for_rank_else_skip(3)",
-                plan.toString());
+        assertThat(plan, scoreForRank(contains(
+                hasToString("__rank_0 = BasicRankedRecord$score.score_for_rank_else_skip(3)")),
+                // need to explicitly construct QueryComponent because we're using an internal parameter
```
MMcM, line 1028: Maybe a helper here in this fixture for that?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
nschiefer: Yes, this should be _full_ records, not partial ones.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
nschiefer: Ah, great, that makes this much tidier.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
nschiefer: Yes, good idea. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -0,0 +1,211 @@
...
+        for (int i = 0; i < valueFields.size(); i++) {
+            KeyExpression valueField = valueFields.get(i);
+            FieldData fieldData = FieldData.of(IndexKeyValueToPartialRecord.TupleSource.VALUE, i);
+            if (!valueField.createsDuplicates() && addCoveringField(valueField, fieldData, builder)) {
+                fields.put(valueField, fieldData);
+            }
+        }
+
+        if (!builder.isValid()) {
+            return NO_FIELDS;
```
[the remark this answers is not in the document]
nschiefer: No, this can happen for fairly pedestrian reasons, such as having a repeated field. `isValid()` is a bit harsh sounding, but I didn't want to change the existing method name.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+            builder.addField(fieldKeyExpression.getFieldName(), fieldData.source, fieldData.index);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    @Nonnull
+    public static AvailableFields intersection(@Nonnull List<AvailableFields> toIntersect) {
```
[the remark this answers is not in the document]
nschiefer: With regard to the first thought, I agree in principle but think this should be handled in a more systematic way. Right now, it's remarkably difficult to say exactly what fields are present, because the way that "present" is defined varies by context. For example, is a repeated field "present"? For the purposes of covering optimizations, it is not, but in general it probably is. I believe that the logic we have here is circumstantially correct and also fairly simple. The join-based framework you're working on will clarify a lot of these issues.

Regarding the second point, that's an oversight on my part. This is only ever called by union and intersection plans, w which are required to have a least two children. I've added a check that throws an exception if this gets called on an empty list.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIntersectionPlan.java
```diff
@@ -202,8 +203,16 @@ public boolean equalsWithoutChildren(@Nonnull RelationalExpression otherExpressi
                comparisonKey.equals(other.comparisonKey);
     }
 
-    @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
+    @Nonnull
+    @Override
+    public AvailableFields getAvailableFields() {
+        return AvailableFields.intersection(quantifiers.stream()
```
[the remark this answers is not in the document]
nschiefer: `RecordQueryIntersectionPlan` requires at least two children (see line 294). However, I've added a check to `AvailableFields` so that it can't be misused.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/FilterVisitor.java
```diff
@@ -0,0 +1,116 @@
...
+    // index entries. Reconstituting that as a singleton in a partial record might work for the simplest case, but
+    // could not for multiple such filter conditions.
+    private static boolean findFilterCoveredFields(@Nonnull QueryComponent filter, @Nonnull Set<KeyExpression> filterFields) {
+        if (filter instanceof FieldWithComparison) {
+            filterFields.add(Key.Expressions.field(((FieldWithComparison)filter).getFieldName()));
+            return true;
+        }
+        if (filter instanceof AndOrComponent) {
+            for (QueryComponent child : ((AndOrComponent)filter).getChildren()) {
+                if (!findFilterCoveredFields(child, filterFields)) {
```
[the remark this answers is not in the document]
nschiefer: It can't quite be a property visitor because the property visitors no longer visit `QueryComponent`s (and probably never will, because the new planner uses `QueryPredicate` instead). It's the same concept, but I think we should leave anything like this for a future refactor.

nschiefer:
The latest changes reverted the refactoring I made in `ReocrdQueryPlanner.planCoveringAggregateIndex()`, since it seems that I couldn't get it working well enough for both the regular covering aggregate tests and the new bitmap index tests. We can leave that for the future, I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        if (intersection == null) {
+            return ALL_FIELDS;
+        } else {
+            return new AvailableFields(intersection);
+        }
+    }
+
+    /**
+     * A pair of a representing a tuple source (key or value) and an index within that tuple source.
```
MMcM, line 196:
```suggestion
     * A pair of a tuple source (key or value) and an index within that tuple source.
```

PR #1021, opened 2020-08-20
 Resolves #1020: Dangling sampled key read operation can log a lot

### LINKED ISSUE #1020: Dangling sampled key read operation can log a lot
The `FDBRecordContext.timedReadSampleKey` spawns a future to do a read purely to time the duration of the operation and relies upon the transaction commit to ensure the completion of the operation. If the transaction is abandoned and closed without commit and the read operation is outstanding, it can cause an error message to be thrown/logged of `error reading sample key` which is annoying and not useful purely for the case that the transaction was closed out from under it.

scgray:


MMcM:
```
com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerSimpleTest > testConfigLoader() FAILED
    org.opentest4j.AssertionFailedError: expected: <92> but was: <91>
```

test this please

MMcM:
```
com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerBuildValueIndexTest$Unsafe > fiftySplitElements() FAILED

        Caused by:
```

test this please

MMcM:
```
Execution failed for task ':fdb-record-layer-spatial-pb3:downloadGeonames'.
> de.undercouch.gradle.tasks.download.org.apache.http.NoHttpResponseException: The target server failed to respond
```

test this please

fdb-extensions/src/main/java/com/apple/foundationdb/ErrorCodes.java
```diff
@@ -0,0 +1,181 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+/**
+ * FDB error codes (from https://apple.github.io/foundationdb/api-error-codes.html) and handy methods to
+ * interpret them.
+ */
+@API(API.Status.MAINTAINED)
+public class ErrorCodes {
```
alecgrieser, line 34: Should this be an enum?

scgray: @alecgrieser Sorry this has been stagnating so long.  I updated this to an `enum`.  It does have the benefit that we can later add other attributes to the error, but it does make other things a little more awkward - for example, there is no good way to represent an error code that we have never seen before.  But, I'm not unhappy with the change.

alecgrieser: I think I like the enum better, though it kind of seems like this should really be an enum that lives in the Java bindings, but alas. The `foundationdb` project already has some automation that can generate C++ files based on these error definitions, so it seems like they should be able to generate this `enum` as well.

There's an alternative version of this change that resolves #1020 but doesn't try and fix the problem of error codes (partially to avoid having two changes in one PR and partially to avoid trying to argue on how the API should look). I'm happy enough with how this is, though.

alecgrieser (changes requested):
Looks like some of the tests are failing, so I'm leaving it on "request changes" (with the requested changes just being to get the tests passing).

It also sounded like you, @scgray, were yourself unsure if you wanted the enum (also the build isn't passing). I also proposed an alternative version that I think would also make sense, which is the "minimal change" to resolve #1020 (i.e., just the "transaction cancelled" checking in `FDBRecordContext`), and then leaving fixing the error code situation to another day.

scgray:
@alecgrieser tests are running now, but should be fixed.  I think the `enum` is ok -- I don't mind splitting it out to another radar if end up having a lot of other aspects to discuss, but I'd like to get it in if we can

.idea/compiler.xml
```diff
@@ -60,8 +81,10 @@
...
       <module name="fdb-record-layer-spatial_main" target="1.8" />
       <module name="fdb-record-layer-spatial_test" target="1.8" />
+      <module name="fdb-record-layer.main" target="11" />
+      <module name="fdb-record-layer.test" target="11" />
       <module name="fdb-record-layer_main" target="1.8" />
       <module name="fdb-record-layer_test" target="1.8" />
     </bytecodeTargetLevel>
   </component>
-</project>
+</project>
```
alecgrieser, line 90: I assume this change wasn't intentionally checked in?

alecgrieser (changes requested):
Okay, code changes look good. Only thing remaining is it looks like some config got checked in that I don't think was intentional

.idea/compiler.xml
scgray: Doh! No, it isn't.  I need to try to understand what is up with this file.  In my environment, it won't compile without it, so I typically have to stash it when I commit, then restore it after.  Anyway, it is removed now

alecgrieser (approved):

PR #1023, opened 2020-08-24
Resolves #1022: Online indexer: increase number of records limit

### LINKED ISSUE #1022: Online indexer: increase number of records limit
After resolving #703 (limit online indexer by transaction size), there is no need to use this limit as a mean to regulate transaction size.

jjezra:
This fix is trivial, but I'm not sure if its effect on the system is trivial too.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -126,7 +126,7 @@ public class OnlineIndexer implements AutoCloseable {
     /**
      * Default number of records to attempt to run in a single transaction.
      */
-    public static final int DEFAULT_LIMIT = 100;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -126,7 +126,7 @@
     /**
      * Default number of records to attempt to run in a single transaction.
      */
-    public static final int DEFAULT_LIMIT = 100;
+    public static final int DEFAULT_REC_NUM_LIMIT = 800;
```
alecgrieser, line 129: I believe, because this is a `public` constant, that this rename is an API-breaking change.

I think I like this name better, but I'm not sure it's worth changing the API for it.

nblintao: Besides, changing a default value sounds like a change that needs adopters' attention, probably it should also be treated as a breaking change.

PR #1027, opened 2020-09-02
Resolves #1026: index matching mechanics

### LINKED ISSUE #1026: planner matching mechanics
This issue tracks progress on the matching mechanics in the Cascades style planner contained in ` com.apple.foundationdb.record.provider.foundationdb.query.temp`. In #987 correlations were introduced to the data model. Index matching is conceptually done by overlaying the data flow graph created by the actual query with the various data flow graphs compiled from index definitions (that are available) in order to find completely subsumed parts of the query that can be answered by an index.

This general matching problem, i.e. find the parts of two graphs that are conceptually doing the same thing and produce in the same resulting records when the eventual plan would be executed, is not straightforward and can be computationally expensive. The specific matching problem establishes for two expressions if expression 1 can be expressed by expression 2 under consideration of a parameter map, and a compensation. So for instance:

`SELECT * FROM T WHERE x = 3 AND y > 5` matches `SELECT * FROM T WHERE y > p1`

with `p1 -> 5` as a parameter mapping and `WHERE x = 3` as additional compensation.

The general matching problem enumerates suitable expression references which can potentially match and attempts to solve the specific matching problem between them. It is somewhat naive to attempt to exhaustively try to match every reference in the query graph to every expression reference in a graph compiled from an index-candidate. In fact, matching can be done bottom up and partial matches can be memoized just like expression graph structures themselves thus eliminating the need  to match useless reference pairs (that cannot actually match at all). For instance two expression reference can only match if we previously established that their children all have matches as well.

This issue should provide the the fundament of the matching framework, that is the necessary data structures, memoization, enumeration of references and expressions considered for matching and the integration of matching into the planner infrastructure.

normen662:
This PR addresses and resolves #1026:

- adds a new package `com.apple.foundationdb.record.query.plan.temp.matching` for matching-related logic. This contains all infrastructural work to match an expression DAG to another expression DAG.
- adds a new package `com.apple.foundationdb.record.query.plan.temp.debug` to support domain-specific debugging from within in the temp planner. The `fdb-record-layer-core` test module contains an implementation of that debugger class which is a Repl that can be used to debug test cases.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/EnumeratingIterator.java
```diff
@@ -0,0 +1,46 @@
...
+ * @param <T> type
+ */
+public interface EnumeratingIterator<T> extends Iterator<List<T>> {
+    /**
+     * Instructs the iterator to advance to the next possible ordering using the given zero-indexed level.
+     *
+     * Example 1: If the last returned ordering of the iterator {@code it} is {@code (e0, e1, e2, e3)} and
+     *            {@code it.skip(2)} is called, the state of the iterator is advanced in a way that either
+     *            reaches the end of iteration or the next item that is returned is {@code (e0', e1', e3', e4')}
+     *            where the prefix {@code (e1', e2', e3')} is not equal to {code (e1, e2, e3)}.
```
MMcM, line 37: I think this example may have changed its mind about how to number the elements midstream.

I believe that `skip(2)` means advance the third level, of which the simplest case is `e0, e1, e2'`, ..., where `e2'` is the successor of `e2`. If there is no such successor, then `e1'` the successor of `e1` and so on.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/EnumeratingIterable.java
```diff
@@ -0,0 +1,73 @@
...
+    EnumeratingIterator<T> iterator();
+
+    static <T> EnumeratingIterable<T> emptyIterable() {
+        return new EmptyIterable<>();
+    }
+
+    /**
+     * An implementation of {@link EnumeratingIterable} that is optimized to work for empty
+     * input sets.
+     * Iterators created by this class, avoid to build complex state objects during their lifecycle.
```
MMcM, line 45:
```suggestion
     * Iterators created by this class, avoid building complex state objects during their lifecycle.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/ExpressionRefTraversal.java
```diff
@@ -21,49 +21,58 @@
...
 /**
- * Utility class to provide a view on a graph given by a root expression reference (to a {@link RelationalExpression}
- * that allows to perform traversal operations that are normally not possible on instances of {@link ExpressionRef}
- * such as {@link #getLeaves()} and {@link FullyTraversableExpressionRef#getParentRefs()}.
+ * Utility class to provide a graph view of a expression reference DAG given by a root expression reference
+ * (to a {@link RelationalExpression} that allows to perform traversal operations that are normally not possible
+ * on instances of {@link ExpressionRef} such as {@link #getLeafRefs()} and {@link #getParentRefs}.
+ *
+ * The implementation of this class assumes that the original DAG is not mutated after the traversal is created.
+ * If the underlying DAG is mutated gainst assumptions, the semantics of this class is defined in a way that calls to
```
MMcM, line 42:
```suggestion
 * If the underlying DAG is mutated against assumptions, the semantics of this class is defined in a way that calls to
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/MatchCandidate.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.temp;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Case class to represent a match candidate. A match candidate on code level is just a name and a data flow graph
+ * that can be matches against a query graph. The match candidate does not keep the root to the graph to be matched but
```
MMcM, line 27:
```suggestion
 * that can be matched against a query graph. The match candidate does not keep the root to the graph to be matched but
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/debug/RestartException.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.temp.debug;
+
+/**
+ * Tag-exception tha is specifically thrown by implementations of {@link Debugger} to signal the planner to restart
```
MMcM, line 24:
```suggestion
 * Tag-exception that is specifically thrown by implementations of {@link Debugger} to signal the planner to restart
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/matching/MatchAccumulator.java
```diff
@@ -0,0 +1,50 @@
...
+
+package com.apple.foundationdb.record.query.plan.temp.matching;
+
+/**
+ * Class to used to accumulate the individual match results (the results of applying the {@link MatchFunction}) into
+ * an object of type {@code R}. This class provides an interface for clients that in its concept is not entirely unlike
+ * {@link java.util.stream.Collector}. Instances of this class, however, are meant to be mutable while collectors
+ * are essentially stateless lambda factories.
+ *
+ * The life cycle of an accumulator (after it's creation) spans multiple calls to {@link #accumulate} which cause the
```
MMcM, line 29:
```suggestion
 * The life cycle of an accumulator (after its creation) spans multiple calls to {@link #accumulate} which cause the
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/matching/BaseMatcher.java
```diff
@@ -0,0 +1,404 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+
+import static com.apple.foundationdb.record.query.plan.temp.TopologicalSort.anyTopologicalOrderPermutation;
+import static com.apple.foundationdb.record.query.plan.temp.TopologicalSort.topologicalOrderPermutations;
+
+/**
+ * Abstract base class for all matchers. Matching is an algorithm that <em>matches</em> two collections of things of type
+ * {@code T} resulting in a bijective mapping (injective and surjective) between the elements of the respective
```
MMcM, line 48: I'm not sure how helpful explaining "bijective" that way is: if you know "injective" and "surjective," you know it. I would spell it out: every element of this collection has a corresponding element of the other collection and vice versa.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/Commands.java
```diff
@@ -0,0 +1,623 @@
...
+                }
+                plannerRepl.println();
+            }
+            return false;
+        }
+
+        @Nonnull
+        @Override
+        public String getCommandToken() {
+            return "QUNS";
```
MMcM, line 589: Could this be `QUANTIFIERS` spelled out or something like that?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/Commands.java
```diff
@@ -0,0 +1,623 @@
...
+     * Restart execution. All entity names remain stable.
+     */
+    @AutoService(Command.class)
+    public static class RestartCommand implements Command<Event> {
+        @Override
+        public boolean executeCommand(@Nonnull final PlannerRepl plannerRepl,
+                                      @Nonnull final Event event,
+                                      @Nonnull final ParsedLine parsedLine) {
+            plannerRepl.restartState();
+            plannerRepl.addInternalBreakPoint(new PlannerRepl.CountingTautologyBreakPoint(1));
```
MMcM, line 416: Does this mean that restart stops after the _first_ thing the new plan does or before it?

MMcM (changes requested):
Some comments from the first pass through.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/AliasMap.java
```diff
@@ -158,14 +163,48 @@
  * This class is immutable, all perceived "mutations" cause a new object to be created.
  *
  */
-@API(API.Status.EXPERIMENTAL)
```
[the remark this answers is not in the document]
normen662: That got lost in a previous merge. Added it back in.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/AliasMap.java
```diff
@@ -158,14 +163,48 @@
...
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final AliasMap aliasMap = (AliasMap)o;
+        return Objects.equals(map, aliasMap.map);
```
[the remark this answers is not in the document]
normen662: It's just coding style really and due to copy/pastes I cannot even claim that it's always like that. Should I always revert to the selfish method call `.equals()` if it's provably non null?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/AliasMap.java
```diff
@@ -176,191 +215,275 @@ public boolean containsTarget(@Nonnull final CorrelationIdentifier alias) {
...
+     * @return the set of entries
+     */
     @Nonnull
-    public Builder derived(int expectedAdditionalElements) {
-        return new Builder(size() + expectedAdditionalElements).putAll(this);
+    public Set<Map.Entry<CorrelationIdentifier, CorrelationIdentifier>> entrySet() {
+        return map.entrySet()
+                .stream()
+                .map(e -> new AbstractMap.SimpleImmutableEntry<>(e.getKey(), e.getValue()))
+                .collect(ImmutableSet.toImmutableSet());
```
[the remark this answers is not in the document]
normen662: Yikes, the backing map is already immutable AND this method is not even called. I made it a non copying getter.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/AliasMap.java
```diff
@@ -377,36 +500,118 @@ private Builder(final int expectedSize) {
...
+     * Find matches between two sets of aliases, given their depends-on sets and a
+     * {@link MatchPredicate}.
+     * This method creates an underlying {@link PredicatedMatcher} to do the work.
+     * @param aliases a set of aliases
+     * @param dependsOnFn a function that returns the set of dependencies for a given alias within {@code aliases}
+     * @param otherAliases a set of other aliases
+     * @param otherDependsOnFn a function that returns the set of dependencies for a given alias within {@code otherAliases}
+     * @param matchPredicate match predicate. see {@link MatchPredicate}
+     *        for more info
+     * @return an iterable of {@link AliasMap}s where each individual {@link AliasMap} is considered one match
```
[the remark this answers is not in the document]
normen662: No, it means that we normally have a conceptual map (just a function really), one for the `this` side (usually the query) and one distinct function for the `other` side (e.g. the materialized view side). They could be merged together as you usually have distinct sets of quantifiers but I think this is cleaner.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CrossProduct.java
```diff
@@ -0,0 +1,266 @@
...
+                    if (state.get(i) != null) {
+                        bound -= 1;
+                    } else {
+                        break;
+                    }
+                }
+            }
+
+            /**
+             * Method that skips advances to the next element on the given zero-indexed level.
```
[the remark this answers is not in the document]
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/EnumeratingIterable.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/ExpressionRefTraversal.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/EnumeratingIterator.java
normen662: Yeah -- the explanation is not good. Changing completely.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/MatchCandidate.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/matching/MatchAccumulator.java
normen662: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/Commands.java
normen662: this is the command you have to type in on the REPL to show all quantifier objects. I'd rather have it very short as I am using this quite often. The other similar commands comparable to this are `REFS` and `EXPS`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/Commands.java
normen662: The first thing the compilation process does it to pop the first task from the task stack and call the debugger with an instance of `ExecutingTaskEvent`. That's when the break point stops execution again and self-removes itself from the list of break points. So to answer your question it will be before the first actual thing happens and identical to what happens when you first run it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/AliasMap.java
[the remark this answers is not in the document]
MMcM: I don't think we have strong guidelines on this. And that we can accommodate some variation.

But it feels a little odd to have variation among methods in the same class defined adjacently. Which I guess argues for making `toString` consistent at least.

MMcM (changes requested):
Just a few small things remaining and then I think we are okay to proceed with this.

MMcM (approved):

PR #1033, opened 2020-09-05
Fixes #1032: In-as-or optimization can infinitely recurse if normalization as "OR" preserves "IN"

### LINKED ISSUE #1032: In-as-or optimization can infinitely recurse if normalization as "OR" preserves "IN"
The planner optimization that can choose to plan in queries as a union can sometimes infinitely recurse. In particular, this line here:

https://github.com/FoundationDB/fdb-record-layer/blob/99de5206c2df22dc98a1d6cf4b795f3bbdf47ace/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java#L407-L410

The `planFilter` call there on line 409 calls another method within the planner. If the transformation of the query filter in `normalizeAndOrForInAsOr` does not remove the "IN" clause (say, because the transformation can't be applied), then that method ends up calling this same method, and then it's stuck in an infinite recursion (or, well, on a machine with finite stack space, a `StackOverflow` error).

I think it would be sufficient in most circumstances to just not recurse if the transformation doesn't change the filter. I believe the transformation is idempotent, so that would be enough to guarantee that it does through the planning logic at most twice.

alecgrieser:
This fixes addresses the problem by having the planner decide not to try and plan the query again if extracting the IN did not actually change the filter. In this way, it will keep recursing until it either removes the IN, plans the sort, or the component stops changing. At present, I can only think of cases where the first "in" extraction results in an "OR" or the component doesn't change (i.e., I think this should only ever recurse zero or one times), but I could be missing something.

There are two commits here. The first refactors some of the tests around this feature to make a new one easier to write, and the second commit introduces a test that exercises this path and then fixes the bug. (I have validated locally that the new test fails (with a `StackOverflow` error) without the planner fix.)

This fixes #1032.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -410,7 +410,11 @@ public class RecordQueryPlanner implements QueryPlanner {
             boolean canSort = inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse());
             if (!canSort && getConfiguration().shouldAttemptFailedInJoinAsOr()) {
                 // Can't implement as an in join because of the sort order. Try as an OR instead.
-                withInAsOr = planFilter(planContext, normalizeAndOrForInAsOr(inExtractor.asOr()));
+                // Only try if extracting the OR actually changes the filter to avoid recursing infinitely.
+                final QueryComponent asOr = normalizeAndOrForInAsOr(inExtractor.asOr());
+                if (!filter.equals(asOr)) {
+                    withInAsOr = planFilter(planContext, asOr);
+                }
             }
         } else if (needOrdering) {
             inExtractor.sortByClauses();
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -18,7 +18,7 @@ In this realase, the various implementations of the `RecordQueryPlan` interface 
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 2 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -406,7 +407,11 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
             boolean canSort = inExtractor.setSort(planContext.query.getSort(), planContext.query.isSortReverse());
             if (!canSort && getConfiguration().shouldAttemptFailedInJoinAsOr()) {
                 // Can't implement as an in join because of the sort order. Try as an OR instead.
-                withInAsOr = planFilter(planContext, normalizeAndOrForInAsOr(inExtractor.asOr()));
+                // Only try if extracting the OR actually changes the filter to avoid recursing infinitely.
+                final QueryComponent asOr = normalizeAndOrForInAsOr(inExtractor.asOr());
+                if (!filter.equals(asOr)) {
```
normen662, line 412: I think this will do what you have in mind, although it's possibly better to make this dependent on the state of the `inExtractor`. We shouldn't attempt it if the `inClauses` in there were non-empty before doing `setSort()`. If that's the case `asOr()` will actually do work and create a new modified filter replacing the IN with and OR construction.

normen662 (changes requested):
Can you take a look at the comment and see if that makes sense?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: I think it's little more subtle than the "in" clauses being non-empty, as we actually already check the emptiness in of the in clauses in `setSort` (and return `true`, so won't go down the path that was infinitely recursing). The somewhat more fundamental issue is that the `in` clauses can't be transformed if they are all `InParameterClause`s, so we could check that (instead of checking to see if things get modified). I have a slight preference for this method, as it seems slightly more resilient to future modifications that introduce other mechanisms for `asOr` improving or changing, but it's not something I'm super committed to either way.

Sorry, it took a while to respond; got distracted.

alecgrieser:
Closing as dupe of #1287 now

PR #1036, opened 2020-09-30
Convert SynchronizedSessionLockedException error message to a non-logging exception 

jjezra:
This exception happens too frequently, and not an error per se.


jjezra:
Question to reviewers: 
1. Should SynchronizedSessionLockedException extend RuntimeException? 
2. Should the second thrower (("Failed to continue the session") be an error instead of an info?

nblintao:
This does not change how the client is going to log this exception. It should be done in the client that using Record Layer.

PR #1037, opened 2020-10-02
Removed items no longer available in IDEA from docs

ohadzeliger:


docs/Building.md
```diff
@@ -29,10 +29,6 @@ There are a few settings that we use for consistency that are saved in the user-
 
 * Preferences > Editor > General > Auto Import > Optimize imports on the fly (for current project)
 
-* Preferences > Build, Execution, and Deployment > Build Tools > Gradle > Runner > Delegate IDE build/run actions to gradle
-
-* Preferences > Build, Execution, and Deployment > Build Tools > Gradle > Runner > Run tests using: > Gradle Test Runner
```
alecgrieser, line 34: I think there are still variants of this in the Gradle "Build and run" section of the menu, but it also looks like the default is now to use the Gradle runner, so I suppose it's fine to remove these.

alecgrieser (approved):

PR #1040, opened 2020-10-12
Resolves #1039: partition residual filters into index fiters and true…

### LINKED ISSUE #1039: differentiate between index filters and residual filters to increase applicability of covering indexes
In the heuristically driven planner, we convert a regular index scan into a covering index scan as a post-processing step after we already picked an index. That index was picked using a heuristic to favor the narrowest index that can support the search arguments while leaving the smallest number of residual filters (filters that cannot be expressed as ranges on an index scan but have to be post-applied). Consider this example:

`index1`: `(a, b)`
`index2`: `(a,b,c)`
query filter: `a = 10 AND b > 3 AND c = 4`

The planner currently favors a plan over `index1` as it is the narrowest index that can support the the query. In this case, however, it would be better to use `index2` since that gives us the same selectivity and allows us to express this index scan using a covering index scan as `c=4` can be evaluated on the partial record coming from the index scan:

`FETCH(FILTER(COVERING(INDEX2, [10, 3], [3, inf]), c = 4))`

This issue should cover proper classification of predicates such as `c = 4` as index filters as opposed to to true residual filters that have to be evaluated after fetching the full record. This new way of classifying filters is then to be used to create a more desirable plan as shown above.

normen662:
… residuals

During planning in the heuristic planner we partition `unsatisfiedFilters` into `unsatisfiedFilters` aka `residuals` and `indexFilters` which can be applied on index records. We use that information to favor an index scan over a wider index if there are index filters that can be applied that would allow us to filter records before we fetch them.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -66,18 +66,22 @@ import com.apple.foundationdb.record.query.plan.plans.RecordQueryFilterPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryIndexPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryIntersectionPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan;
+import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlanWithIndex;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryScanPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryTextIndexPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryTypeFilterPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnionPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedPrimaryKeyDistinctPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedUnionPlan;
-import com.apple.foundationdb.record.query.plan.visitor.FilterVisitor;
-import com.apple.foundationdb.record.query.plan.visitor.RecordQueryPlannerSubstitutionVisitor;
 import com.apple.foundationdb.record.query.plan.temp.explain.PlannerGraphProperty;
 import com.apple.foundationdb.record.query.plan.temp.properties.FieldWithComparisonCountProperty;
+import com.apple.foundationdb.record.query.plan.visitor.FilterVisitor;
+import com.apple.foundationdb.record.query.plan.visitor.RecordQueryPlannerSubstitutionVisitor;
 import com.apple.foundationdb.record.query.plan.visitor.UnorderedPrimaryKeyDistinctVisitor;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -367,7 +371,7 @@ public class RecordQueryPlanner implements QueryPlanner {
         }
     }
 
-    private int indexSizeOverhead(PlanContext planContext, @Nonnull Index index) {
+    private static int indexSizeOverhead(PlanContext planContext, @Nonnull Index index) {
         if (planContext.commonPrimaryKey == null) {
             return index.getColumnSize();
         } else {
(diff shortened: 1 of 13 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBReturnedRecordLimitQueryTest.java
```diff
@@ -65,7 +67,7 @@
      * Verify that a returned record limit works properly against a query with a filter on one field and a sort on another,
      * when the filter field is un-indexed and the sort is in reverse order.
      */
-    @DualPlannerTest
+    @Test
```
normen662, line 70: Had to remove the dual planner annotation as the temp planner cannot plan this anymore.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBReturnedRecordLimitQueryTest.java
```diff
@@ -88,13 +90,13 @@ public void testComplexLimits2() throws Exception {
...
                     myrec.mergeFrom(rec.getRecord());
-                    assertEquals("odd", myrec.getStrValueIndexed());
                     assertEquals(0, myrec.getNumValue2());
+                    assertEquals("odd", myrec.getStrValueIndexed());
                     i += 1;
                 }
             }
             assertEquals(10, i);
-            assertDiscardedAtMost(18, context);
+            assertDiscardedAtMost(34, context);
```
normen662, line 99: While the plan is better it seems that we discard more records. I am assuming that is because of prefetching but it would be nice to understand exactly why that is.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRepeatedFieldQueryTest.java
```diff
@@ -336,17 +338,17 @@ public void sortRepeated() throws Exception {
                     .setFilter(Query.field("name").greaterThan("A"));
             RecordQuery query = builder.setRemoveDuplicates(false).build();
             RecordQueryPlan plan = planner.plan(query);
-            assertThat(plan, filter(query.getFilter(), indexScan(allOf(indexName("customers"), unbounded()))));
-            assertEquals(1833106833, plan.planHash());
-            assertEquals(Arrays.asList(1000L, 1001L, 1000L, 1001L, 1000L, 1000L, 1001L), fetchResultValues(plan, TestRecords4Proto.RestaurantRecord.REST_NO_FIELD_NUMBER,
+            assertThat(plan, fetch(filter(query.getFilter(), coveringIndexScan(indexScan(allOf(indexName("customers-name"), unbounded()))))));
+            assertEquals(-505715770, plan.planHash());
+            assertEquals(Arrays.asList(1000L, 1001L, 1000L, 1001L, 1000L, 1001L, 1000L), fetchResultValues(plan, TestRecords4Proto.RestaurantRecord.REST_NO_FIELD_NUMBER,
```
normen662, line 343: The chose plan uses a different index that uses a second field (`name`) to impose a secondary order which is not specified in the test case.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
```diff
@@ -1116,8 +1116,7 @@ public void headerRankAndIdQuery() throws Exception {
                 .build();
         planner.setIndexScanPreference(QueryPlanner.IndexScanPreference.PREFER_SCAN);
         RecordQueryPlan plan = planner.plan(query);
-        assertEquals("Scan(([buffaloes, 100],[buffaloes]]) | [HeaderRankedRecord]" +
-                        " | score LESS_THAN $__rank_0 WHERE __rank_0 = score_by_nested_id.score_for_rank_else_skip(buffaloes, 2)",
+        assertEquals("Fetch(Covering(Index(score_by_nested_id ([buffaloes, null],[buffaloes, 2]) BY_RANK) -> [score: KEY[1], header: [group: KEY[0], id: KEY[2]]]) | header/{id GREATER_THAN 100})",
```
normen662, line 1119: This plan change is due to different _costing_. Previously `SCAN()` and `INDEX_SCAN()` where intrinsically considered equal and the `PREFER_SCAN` configuration parameter was the tie breaker. Now, the index is considered better as the filter on `header/{id GREATER THAN 100}` can be applied on the index key. Not sure if we want it like this.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBReturnedRecordLimitQueryTest.java
MMcM: Prefetching is not particularly relevant.

With the new plan, the entire index scan must be reversed to accomplish reverse sorting by the first field, which means that we do the odd-2's, then the odd-1's, then odd-0's. So we have to skip over those first two before getting to any matches and then we get 10 in a row. With the old single-field index plan, we go in reverse primary key order and every third matches, so we get the 10 we need faster.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRepeatedFieldQueryTest.java
MMcM: That is, the better (because we can check the filter using the index key) plan _incidentally_ imposes an order by that key rather than by something like the primary key.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
MMcM: `SCAN` gives us something that is capable of evaluating the predicate with no further database traffic. More data is fetched in the scan itself. Whether this matters depends on the selectivity: if most pass, then we end up doing a fetch of the record anyway, using a less-efficient (even in parallel) single point fetch.

I kind of expected that _all_ filters to a `SCAN` would be index filters and not residuals. With that change, the same tie would result, restoring the status quo, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -517,9 +566,38 @@ private ScoredPlan planIndex(@Nonnull PlanContext planContext, @Nonnull QueryCom
                 }
             }
         }
+
         return p;
     }
 
+    private ScoredPlan computeIndexFilters(@Nonnull PlanContext planContext, @Nonnull final ScoredPlan plan) {
+        if (plan.plan instanceof RecordQueryPlanWithIndex) {
```
MMcM, line 574: A `RecordQueryScanPlan` could treat all filters as index filters and not residuals, I think.

MMcM (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
normen662: It does! I have added that functionality.

normen662:
test this please

MMcM (approved):

PR #1047, opened 2020-10-16
Fixes #1046: Setting a time limit can lead to "split record without start" error when reading data

### LINKED ISSUE #1046: Setting a time limit can lead to "split record without start" error when reading data
This may be somewhat difficult to test, but I believe that if the following set of circumstances happen, then it can time out reading in the middle of reading a record:

1. The record store has versionstamp enabled but otherwise does not allow split records
1. A scan of the record store happens with a time limit set
1. The timeout is hit in between reading the version and reading the record itself for the record

In that case, the scan will timeout half way and think it found a record that had a version but no "body".

In reverse, this can be even worse: you can read the body of the record and not the version, so it will effectively lose that version information.

I believe that this bug only affects stores which do not allow split records due to a subtle difference in the way they handle limits. But that's mostly determined from staring at the code and not from experimentation.

alecgrieser:
The issue here is that in the code as it was originally, if the meta-data does not allow split records, then the time limit was enforced on the key value cursor. (If you had split records, then the time limit was removed on the very bottom key value cursor, and then only enforced in the wrapping cursor (the "key value unsplitter").) This meant that if you had versionstamps enabled, it was possible for the limiter to stop midway through, between a record and its version (if the time limit was hit at just the wrong time). Now, this removes the time limit from the key value cursor (always) and instead enforces that limit in the wrapping cursor so that it will always stop after reading a full record. This was already being done for the other kinds of limits, but not the time limit.

This fixes #1046.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -18,7 +18,7 @@ In this realase, the various implementations of the `RecordQueryPlan` interface 
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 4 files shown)


MMcM (approved):

alecgrieser:
Just to protect against merge skew:

test this please

PR #1049, opened 2020-10-21
Resolves #1048: preloadRecordAsync should cache negative results

### LINKED ISSUE #1048: The preload record cache should cache negative hits
The `FDBRecordStore.preloadRecordAsync()` allows for an optimization in which a client can asynchronously ask for records to be fetched on the assumption that it will eventually want to read them at some point in the same transaction.  Currently this cache only contains records that were successfully read, but it could also optimize fetches for records that prove not to exist (saving at minimum the JNI round trip into the FDB driver).

scgray:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBPreloadRecordCache.java
```diff
@@ -0,0 +1,175 @@
...
+    @Nonnull
+    private final Cache<Tuple, EntryImpl>  cache;
+
+    public FDBPreloadRecordCache(int maximumSize) {
+        cache = CacheBuilder.newBuilder().maximumSize(maximumSize).build();
+    }
+
+    /**
+     * Called prior to performing a prefetch for a given {@code Tuple}, returning a future holder that must
+     * be filled in with the record when it has completed.
```
MMcM, line 45: What is the purpose of having incomplete entries in the cache at all? They aren't returned to another load attempt or even shared by it?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBPreloadRecordCache.java
scgray: I should have commented on this.  The current strategy allows for a race condition I believe. 

```
Thread 1                    Thread 2
-----------------      ------------------
begin pre-fetch    
read record
                       replace record
                       invalidate (non-existant) cache entry
put record in cache
```

with the strategy I'm proposing, the holder for the record being pre-fetched is pre-allocated in the cache and completed when the read is finished.  An invalidate kicks the holder out of the cache, so that the results will never be seen.

scgray: There *may* even be a path for the pre-fetch read to pick up a portion of the replaced record as well....

MMcM: Okay. I guess there should be a short comment there for why it does that.

Such race conditions seem more fundamentally problematic. For instance,
```
Thread 1                    Thread 2
-----------------      ------------------
begin pre-fetch    
read record
                       replace record
                       invalidate (incomplete) cache entry
(put record in orphan entry)
take action based on out-of-date record
```
But there's no reason not to try to avoid what can be relatively cheaply.

scgray: I *think* we are saying the same thing...unless I'm interpreting your example wrong.  Since the orphaned entry is ejected from the cache, its value (the record that thread #1 was reading), will never be visible to the outside world to act upon it.   But, yes, I'll add a comment to explain the reason for the design.

MMcM: Yes, but thread 1 continues to make changes elsewhere dependent on stale information. I am not saying we need to try to fix this here. Or that we should give up on concurrency within the same context because of it. Just that it's hard to solve without synchronization outside of what we provide.

scgray:
```
com.apple.foundationdb.record.spatial.geophile.GeophileQueryTest > testJoin() FAILED
        at com.apple.foundationdb.record.provider.foundationdb.FDBExceptions.wrapException(FDBExceptions.java:187)
        at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.lambda$new$0(FDBDatabase.java:167)
        at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.asyncToSync(FDBDatabase.java:1070)
        at com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext.asyncToSync(FDBRecordContext.java:1034)
        at com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase$BaseBuilder.open(FDBRecordStoreBase.java:1939)
        at com.apple.foundationdb.record.spatial.geophile.GeophileQueryTest.loadCountryShapes(GeophileQueryTest.java:200)
        at com.apple.foundationdb.record.spatial.geophile.GeophileQueryTest.testJoin(GeophileQueryTest.java:341)

        Caused by:
```
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBPreloadRecordCache.java
scgray: Got it.

MMcM (approved):

PR #1053, opened 2020-10-24
Fixes #1052: Need grouping / grouped key canonical form

### LINKED ISSUE #1052: Need grouping / grouped key canonical form
The fix for #1050 exposes a potential inconsistency when `Key.Expressions.empty()` is included in `concat`.

There is no compelling need to ever do that, but it does fall out of how `COUNT(*) GROUP BY x` might be indexed.
Which in turn causes problems when checking that this has the same grouping as `SUM(y) GROUP BY x`.

MMcM:


MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java
```diff
@@ -159,15 +160,33 @@ public static KeyExpression getGroupedKey(@Nonnull KeyExpression key) {
...
         if (!(key instanceof GroupingKeyExpression)) {
             return EmptyKeyExpression.EMPTY;
         }
         GroupingKeyExpression grouping = (GroupingKeyExpression) key;
-        return getSubKey(grouping.getWholeKey(), 0, grouping.getGroupingCount());
+        return groupKey(getSubKey(grouping.getWholeKey(), 0, grouping.getGroupingCount()));
+    }
+
+    protected static KeyExpression groupKey(@Nonnull KeyExpression key) {
+        if (key instanceof ThenKeyExpression) {
```
alecgrieser, line 175: It seems like, at least in principle, that this should also recurse down, say, nesting key expressions and remove `Empty` expressions children... On the other hand, that might be better included in some kind of "key canonicalization" function that recurses down all of the various key types.

I could also see the argument that we should fix this to make "incremental progress" and then add full canonicalization later (not letting the perfect be the enemy of the good, etc.)

alecgrieser (commented):
I left a comment about whether this should canonicalize more types than just `Then`. However, I could also be convinced that merging this with just `Then` being canonicalized to remove `empty()` is a good starting point that we can merge and then iterate on later.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java
MMcM: I could see the benefit of a centralized canonical form transform. Though some care is required in plumbing that in, since key expressions are persisted in meta-data and things might rely on equality comparisons there.

But I was thinking of this as local to the these functions and therefore only aiming to take care of the ones for which it is responsible. I do understand that in some of the cases, the affected `Then` actually existed in the input and wasn't constructed here. But per the contract, it could have been, or, looked at another way, the result is now the same as if it fell into the general "splitting" logic.

So I think I would prefer to take this rather than aiming for the more comprehensive one.

alecgrieser: Yeah, I think that makes sense

alecgrieser (approved):

Issue #794, opened 2020-01-02
FDBRecordStore.rebuildIndexWithNoRecord should clear and mark readable and not just assume that the subspace is empty


nblintao:

nblintao:
This was reverted because of https://github.com/FoundationDB/fdb-record-layer/issues/808, so we should reopen it.

PR #1055, opened 2020-10-28
Fixes #1054: "make unbuilt index readable" exception blocks opening a record store

### LINKED ISSUE #1054: "make unbuilt index readable" exception blocks opening a record store
This happens when check version tries to `rebuildIndexWithNoRecord` but the index was not readable (probably related to reusing index name, also see #794 which was reverted)

There are two things need to be fixed:

1. In such situation it should mark the index to readable regardless

2. In any case, marking index as readable should not block the critical path (`checkVersion`), whose failure can people unable to do anything with the record store.

nblintao:
This happens when check version tries to `rebuildIndexWithNoRecord` but the index was not readable (probably related to reusing index name, also see #794 which was reverted)

There are two things need to be fixed:

1. In such situation it should mark the index to readable regardless

2. In any case, marking index as readable should not block the critical path (`checkVersion`), whose failure can people unable to do anything with the record store.

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
@@ -914,11 +914,11 @@ public class MoreAsyncUtil {
     }
 
     /**
-     * Handle when <code>futureSupplier</code> has exception when supplying a future, or the future is completed
+     * Handle when <code>futureSupplier</code> encounters an exception when supplying a future, or the future is completed
      * exceptionally. Unlike the "handle" in CompletableFuture, <code>handlerOnException</code> is not executed if
      * the future is successful.
-     * @param futureSupplier the supplier of future which need to be handled
-     * @param handlerOnException the hanlder when the future has
+     * @param futureSupplier the supplier of future which needs to be handled
+     * @param handlerOnException the handler when the future encounters an exception
      * @param <V> the result type of the future
      * @return future that completes exceptionally if the handler has exception
      */
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
@@ -885,8 +885,8 @@ public class MoreAsyncUtil {
      * @param future future to compose the handler onto
      * @param handler handler bi-function to compose onto the passed future
      * @param exceptionMapper function for mapping the underlying exception to a {@link RuntimeException}
-     * @param <V> return type of original future
-     * @param <T> return type of final future
+     * @param <V> type of original future
+     * @param <T> type of final future
      * @return future with same completion properties as the future returned by the handler
      * @see AsyncUtil#composeHandle(CompletableFuture, BiFunction)
      */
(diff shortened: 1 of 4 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
```diff
@@ -913,6 +913,32 @@ public static void closeIterator(@Nonnull Iterator<?> iterator) {
         });
     }
 
+    /**
+     * Handle when <code>futureSupplier</code> has exception when supplying a future, or the future is completed
```
alecgrieser, line 917:
```suggestion
     * Handle when <code>futureSupplier</code> encounters an exception when supplying a future, or the future is completed
```

fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
```diff
@@ -913,6 +913,32 @@ public static void closeIterator(@Nonnull Iterator<?> iterator) {
...
+     * exceptionally. Unlike the "handle" in CompletableFuture, <code>handlerOnException</code> is not executed if
+     * the future is successful.
+     * @param futureSupplier the supplier of future which need to be handled
```
alecgrieser, line 920:
```suggestion
     * @param futureSupplier the supplier of future which needs to be handled
```

fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
```diff
@@ -913,6 +913,32 @@ public static void closeIterator(@Nonnull Iterator<?> iterator) {
...
+     * @param handlerOnException the hanlder when the future has
```
alecgrieser, line 921: Looks like this sentence drops off half way

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3055,7 +3090,14 @@ public boolean isIndexDisabled(@Nonnull String indexName) {
                     Index index = indexItem.getKey();
                     List<RecordType> recordTypes = indexItem.getValue();
                     IndexState indexState = newStates.getOrDefault(index, IndexState.READABLE);
-                    work.add(rebuildOrMarkIndex(index, indexState, recordTypes, reason, oldMetaDataVersion));
+                    final CompletableFuture<Void> rebuildOrMarkIndexSafely = MoreAsyncUtil.handleOnException(
+                            () -> rebuildOrMarkIndex(index, indexState, recordTypes, reason, oldMetaDataVersion),
+                            exception -> {
+                                // If there is anything issue, simply mark the index as disabled without blocking checkVersion
+                                logExceptionAsWarn("unable build index", exception);
```
alecgrieser, line 3097:
```suggestion
                                logExceptionAsWarn("unable to build index", exception);
```

It also seems like we might want to include things like the index name as a log key. Maybe `logExceptionAsWarn` should take a `KeyValueLogMessage` (and handle adding the log keys from the exception as well)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3117,7 +3159,10 @@ private boolean areAllRecordTypesSince(@Nullable Collection<RecordType> recordTy
             }
         }
 
-        return markIndexReadable(index).thenApply(b -> null);
+        // The index states is supposed to be empty (i.e. readable) because this is a new index. But unfortunately, the
+        // index state is keyed by index name which can be reused. So we need to mark the index readable. In such case,
+        // the built range is unlikely to be full so we should force it.
+        return markIndexReadable(index, true).thenApply(b -> null);
```
alecgrieser, line 3165: There's something about having this catch and then ignore the error that feels a little strange....

Given that we're going to ignore the results of doing the validation, it almost seems like we should call `uncheckedMarkIndexReadable`. I could see alternative arguments that we could, like, check there are no uniqueness violations or data in the index subspace or something and then mark it as `READABLE` (perhaps only in the error case) (and maybe it also should add the `\x00` to `\xff` range to the build space?).

I could also see an argument that if it encounters an error, it should mark the index as `DISABLED` and wait for a rebuild. Then we don't get the index right away (sad) but we also no longer block `checkVersion`.

Something about the extra flag to force marking it as readable doesn't quite seem right.

alecgrieser (changes requested):
In addition to the comments I left in line, I feel like this should probably have a test. Either one that fully simulates the problem case here (adding and then removing and then adding back an index) or one that "cheats" by marking an index that "doesn't exist" as disabled and then adding the index to the meta-data in a subsequent operation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
nblintao: I didn't use `uncheckedMarkIndexReadable` mainly because I'd like it check and log exceptions even if not throw an exception.

I think " mark the index as DISABLED and wait for a **(manual, force)** rebuild" is an option. I just need to remove all my changes in `markIndexReadable` it will work like that. But that depends on how we see such error. If we think consider it as expected (even though not pleasant) as a side effect of reusing index name, I think it should better be working without manual intervention.

nblintao: Force build is not necessary. Just manual. So I think I'm fine with not marking index as readable, especially considering there may be underlying bugs that we need to be aware of.

nblintao:
test this please

alecgrieser (approved):

MMcM (approved):

PR #1058, opened 2020-10-28
Resolves #1057: Composed bitmaps from map-like fields

### LINKED ISSUE #1057: Composed bitmaps from map-like fields
Given a nested repeated field containing a grouping key and a position value (perhaps declared as a Protobuf `map` field) and several indexes including this and some other groups, it should be possible to generated composed bitmaps for Boolean expressions among those groups, just as is possible today with the fixed group and position as top-level fields.

The natural approach given the code today requires a solution to #1056. But something should be possible without that.

Splicing the per-index conditions normally means putting their keys between the last grouping key and the grouped key (the position). But that isn't possible today if those come from a child `concat`. The next best thing is to put the new key before the nested group. This order is slightly less flexible in terms of the kind of per-index conditions it allows, limiting them to equality.

Covering index scans normally reject repeated fields, since the index can only supply a single element. But in this case, that is enough, so a special flag should permit it.

MMcM:
* Don't pick a splice point in the middle of a nested concat.
* Allow for partial record with incomplete repeated field for special case of composed bitmap.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
@@ -192,7 +192,12 @@ public class IndexKeyValueToPartialRecord {
                 default:
                     throw new RecordCoreException("only nested message should be handled by MessageCopier");
             }
-            recordBuilder.setField(fieldDescriptor, nested.toRecord(fieldDescriptor.getMessageType(), kv));
+            final Message message = nested.toRecord(fieldDescriptor.getMessageType(), kv);
+            if (fieldDescriptor.isRepeated()) {
+                recordBuilder.addRepeatedField(fieldDescriptor, message);
+            } else {
+                recordBuilder.setField(fieldDescriptor, message);
+            }
         }
 
         @Override
@@ -298,16 +303,20 @@ public class IndexKeyValueToPartialRecord {
          * @return whether this is a valid use
          */
         public boolean isValid() {
+            return isValid(false);
+        }
+
+        public boolean isValid(boolean allowRepeated) {
             for (Descriptors.FieldDescriptor fieldDescriptor : recordDescriptor.getFields()) {
                 if (fieldDescriptor.isRequired() && !hasField(fieldDescriptor.getName())) {
                     return false;
                 }
-                if (fieldDescriptor.isRepeated() && hasField(fieldDescriptor.getName())) {
+                if (!allowRepeated && fieldDescriptor.isRepeated() && hasField(fieldDescriptor.getName())) {
                     return false;
                 }
                 if (fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
                     Builder builder = nestedBuilders.get(fieldDescriptor.getName());
-                    if (builder != null && !builder.isValid()) {
+                    if (builder != null && !builder.isValid(allowRepeated)) {
                         return false;
                     }
                 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
@@ -27,6 +27,7 @@ import com.apple.foundationdb.record.metadata.IndexTypes;
 import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.ThenKeyExpression;
 import com.apple.foundationdb.record.provider.foundationdb.IndexAggregateGroupKeys;
 import com.apple.foundationdb.record.provider.foundationdb.IndexFunctionHelper;
 import com.apple.foundationdb.record.provider.foundationdb.indexes.BitmapValueIndexMaintainer;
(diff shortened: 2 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -298,16 +303,20 @@ public void addRequiredMessageFields() {
          * @return whether this is a valid use
          */
         public boolean isValid() {
+            return isValid(false);
+        }
+
+        public boolean isValid(boolean allowRepeated) {
```
alecgrieser, line 309: There's something about this where it seems like the right thing is to have a key expression or something like a key expression that means "one entry from a repeated field" to avoid having a Boolean special case, but I suppose that's a much larger discussion. cc @normen662

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -319,9 +321,26 @@ private static boolean separateGroupFilters(@Nonnull QueryComponent filter,
...
+            // Moreover, the caller needs to have arranged for a compatible index to exist, which requires new support.
+            // (https://github.com/FoundationDB/fdb-record-layer/issues/1056)
+            final GroupingKeyExpression groupKey = (GroupingKeyExpression)indexAggregateFunction.getOperand();
+            int lastKeySize = 1;
+            if (groupKey.getWholeKey() instanceof ThenKeyExpression) {
+                ThenKeyExpression thenKey = (ThenKeyExpression)groupKey.getWholeKey();
+                lastKeySize = thenKey.getChildren().get(thenKey.getChildren().size() - 1).getColumnSize();
+            }
+            final ThenKeyExpression splicedKey;
+            if (lastKeySize > groupKey.getGroupedCount()) {
```
alecgrieser, line 335: Can this still end up splicing up a nested concat? I'm thinking of something like if `groupKey.getGroupedCount() == 2` and `groupKey` is something like `concat(nest(a, concat(b, c)), d)`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -319,9 +321,26 @@ private static boolean separateGroupFilters(@Nonnull QueryComponent filter,
...
+                final KeyExpression wholeKey = groupKey.getWholeKey();
+                final int wholeSize = wholeKey.getColumnSize();
+                final int splicePoint = wholeSize - lastKeySize;
+                splicedKey = Key.Expressions.concat(wholeKey.getSubKey(0, splicePoint), indexKey, wholeKey.getSubKey(splicePoint, wholeSize));
```
alecgrieser, line 339: It's a little unclear to me why this is a legal transformation. Why is it okay if some of the grouping columns from the operand end up in the grouped columns?

alecgrieser (changes requested):
I think this could also probably use a test where the new code gets exercised

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: It's supposed to be an invariant that `splicePoint <= groupedCount`, so that the splice is always someplace among the grouping columns. Let me see about implementing this so that is more obviously true.

MMcM: The grouped count of `fullKey` does not change.
It should be that the splice is _always_ in the grouping columns.
Let me try to write this in a way that is more clear.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: The new version will always take whole children of a `Then`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
MMcM: I'm not sure how a new key expression might be used, since as things stand no new types are introduced.

I think the "right thing" is to have dynamic row types so that it isn't necessary to impersonate the record with varying degrees of fidelity. Maybe that's the same thing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -319,9 +321,37 @@ private static boolean separateGroupFilters(@Nonnull QueryComponent filter,
...
+            } else {
+                final KeyExpression wholeKey = groupKey.getWholeKey();
+                final int splicePoint = wholeCount - afterSpliceCount;
+                if (splicePoint == 0) {
+                    splicedKey = Key.Expressions.concat(indexKey, wholeKey);
+                } else {
+                    splicedKey = Key.Expressions.concat(wholeKey.getSubKey(0, splicePoint), indexKey, wholeKey.getSubKey(splicePoint, wholeCount));
+                }
+            }
+            GroupingKeyExpression fullKey = splicedKey.group(groupedCount);
```
alecgrieser, line 354: Okay, it took me a while to refamiliarize myself with this code, but I think I understand now the basic approach here.

This part of the code is essentially trying out different candidate grouping expressions to match each predicate and seeing if a particular index matching that expression exists. So the important invariant is that the grouped key remains the same (and this does that), and then the grouping keys can be changed kind of at will. So, I think I understand how this works, I think.

I guess the remaining question is that it seems like the grouping key order shouldn't really matter (at least for the keys with equality predicates), and that we "should" be able to just place the `indexKey` at the beginning of the expression. But I'd also buy that the planner isn't quite sophisticated enough for that (yet?).

alecgrieser (changes requested):
Okay, I now think I have a much better grasp on what this is doing, which is good (for me at least). I have a few clarifying questions and then one minor change request, I think. Otherwise, good to merge.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
```diff
@@ -483,9 +488,65 @@ public void nonOverlappingOrQuery() {
...
+                    .setRecordType("MyNestedRecord")
+                    .setFilter(Query.and(
+                            Query.field("num_value_1").equalsValue(1),
+                            Query.field("nested").matches(Query.field("entry").oneOfThem().matches(Query.field("str_value").equalsValue("odd"))),
+                            Query.field("num_value_2").equalsValue(3),
+                            Query.field("num_value_3").equalsValue(4)))
+                    .setRequiredResults(Collections.singletonList(field("nested").nest(field("entry", FanOut).nest("num_value"))))
+                    .build();
+            final RecordQueryPlan queryPlan =  ComposedBitmapIndexAggregate.tryPlan((RecordQueryPlanner)planner, recordQuery, bitmap_value_nested_num_by_str)
+                    .orElseThrow(() -> new IllegalStateException("Cannot plan query"));
```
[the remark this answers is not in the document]
MMcM: Yes, that makes sense.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexTest.java
```diff
@@ -483,9 +488,65 @@ public void nonOverlappingOrQuery() {
...
+            createOrOpenRecordStore(context, metaData(nested_rec_no_by_str_nums_hook));
+            setupPlanner(null);
+            final RecordQuery recordQuery = RecordQuery.newBuilder()
+                    .setRecordType("MyNestedRecord")
+                    .setFilter(Query.and(
+                            Query.field("num_value_1").equalsValue(1),
+                            Query.field("nested").matches(Query.field("entry").oneOfThem().matches(Query.field("str_value").equalsValue("odd"))),
+                            Query.field("num_value_2").equalsValue(3),
+                            Query.field("num_value_3").equalsValue(4)))
+                    .setRequiredResults(Collections.singletonList(field("nested").nest(field("entry", FanOut).nest("num_value"))))
```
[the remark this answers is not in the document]
MMcM: 1. Adding `str_value` is fine; the partial record has an entry with both -- which doesn't matter much, since we just get the index entry directly.
2. Removing the condition and expecting a rollup is sensible, but does not work with the code today. Specifically, it fails `separateGroupFilters` as not matching the `bitmap_value` operand.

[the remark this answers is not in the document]
alecgrieser: Okay, sounds good. I suppose at some point, we may care about roll up, but that sounds fine for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: Yes, given some more code here and a somewhat smarter planner, there is no fundamental reason it could not hunt around for an index some some permutation of the equality constraints. But today it just opts for a workable canonical form, which is as far to the right as possible without splitting.

alecgrieser (approved):

PR #1060, opened 2020-10-28
Remove nullability annotations from primitive fields

MMcM:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursor.java
```diff
@@ -53,7 +53,7 @@
  */
 @API(API.Status.MAINTAINED)
 public class KeyValueCursor extends AsyncIteratorCursor<KeyValue> implements BaseCursor<KeyValue> {
-    @Nullable
+    @Nonnull
```
MMcM, line 56: Always initialized non-null (same IDE inspection).

alecgrieser (approved):

alecgrieser (changes requested):
Approved pending spot bugs rule resolution (which is apparently what caused the prb to fail, though it looks like the rule violation itself wasn't saved in the build artifacts)

alecgrieser (approved):

PR #1069, opened 2020-11-11
Resolves #1068: Could allow degenerate composed bitmaps from single index

### LINKED ISSUE #1068: Could allow degenerate composed bitmaps from single index
Although the caller could sometimes determine, there is no reason not to allow all the `ComposedBitmapIndexAggregate` machinery to work with just a single filter, that is, a single covering index scan that doesn't need to be composed with anything else.

MMcM:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -82,9 +83,9 @@
      * @return an {@code Optional} query plan or {@code Optional.empty} if planning is not possible
      */
     @Nonnull
-    public static Optional<ComposedBitmapIndexQueryPlan> tryPlan(@Nonnull RecordQueryPlanner planner,
-                                                                 @Nonnull RecordQuery query,
-                                                                 @Nonnull IndexAggregateFunction indexAggregateFunction) {
+    public static Optional<RecordQueryPlan> tryPlan(@Nonnull RecordQueryPlanner planner,
```
MMcM, line 86: I believe that an earlier review of this class suggested that this was a more appropriate signature. It didn't matter then, but it does now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -99,12 +100,18 @@
...
         final List<RecordQueryCoveringIndexPlan> indexScans = new ArrayList<>();
         final Map<IndexNode, ComposedBitmapIndexQueryPlan.IndexComposer> indexComposers = new IdentityHashMap<>();
-        return Optional.ofNullable(plan(root, queryBuilder, planner, indexScans, indexComposers))
-                .map(composer -> new ComposedBitmapIndexQueryPlan(indexScans, composer));
+        final ComposedBitmapIndexQueryPlan.ComposerBase composer = plan(root, queryBuilder, planner, indexScans, indexComposers);
+        if (composer == null || indexScans.isEmpty()) {
+            return Optional.empty();
+        }
+        if (indexScans.size() == 1) {
+            return Optional.ofNullable(indexScans.get(0));
```
alecgrieser, line 112: Is this correct if there's a `NOT` in the composer tree against a single index scan?

Also, separately, without this change, would it have returned a `ComposedBitmapIndexQueryPlan` with a single `IndexNode` in the tree (which I think is correct, but perhaps inefficient), or would it have returned something less good?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
alecgrieser: I guess this class doesn't have an API stability annotation. It kind of feels like it should have been marked experimental, but alas. Not sure how much of a stickler we want to be here. My inclination here is to just let the signature be changed.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: It would have given an error, insisting that there is more than one child.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: Somehow half the classes in this package are annotated and the other half is not. I made them all experimental, which I believe was the intention.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
MMcM: I added a check to make sure the composer is trivial (pass through the index). It _might_ be possible to get `NOT` to work in this case, but it doesn't look like it would today, even if the composer were given a chance.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
alecgrieser: Yeah, that sounds right

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
alecgrieser: Okay, sounds good. I suppose it would be nice if `NOT` planned correctly, but having it fail to plan is fine (but having it return the wrong plan would have seemed dangerous).

alecgrieser (approved):

PR #1086, opened 2020-12-17
Resolves #1085: Permuted grouped aggregate

### LINKED ISSUE #1085: Permuted grouped aggregate
Consider a query along the lines of 
```
SELECT zip, MAX(value) FROM properties WHERE state = 'MA' GROUP BY zip ORDER BY 2 DESC
```
that aims to return some grouping keys ordered by an aggregate.

An ordinary plan requires unbounded storage of the intermediate aggregates and then a sort of that temporary table.

Some dialects support an interesting twist using the (relationally suspect)
```
SELECT DISTINCT zip FROM properties WHERE state = 'MA' ORDER BY value DESC
```
which can use an index on `state, value`. It produces results right away and needs space for tracking zips already seen.

To do this with precomputed aggregates, two B-trees are needed:

* An ordinary index on all the group keys and the value (so, `state, zip, value`).

* A permuted `MAX` index with `state, MAX(value), zip`.

Then for maintenance:

* An index insert needs to ascertain whether its `value` is the new maximum and update the permuted entry if so.
  * Assume the ordinary index _has not_ yet been updated (we have no entry there).
  * Read ordinary `(state, zip)` reverse, limit 1. This is the maximum from _another_ record for the same group and so the _current_ maximum.
  * If this is absent or less than the value to be inserted: if not absent, remove permuted `(state, that_value, zip)` and insert permuted `(state, value, zip)`.
  
* An index remove needs to determine whether its `value` is the _current_ maximum and update the permuted entry if so.
  * Assume that the ordinary _has_ been updated already (our entry is not there).
  * Read permuted `(state, value, zip)`.
  * If not present, we are not it, stop.
  * Read ordinary `(state, zip)` reverse, limit 1. This is the maximum from _another_ record for the same group and so the _replacement_ maximum.
  * If this is absent or less than `value` (it might be equal): remove permuted `(state, value, zip)` and, if not absent, insert permuted `(state, that_value, zip)`.

MMcM:


docs/ReleaseNotes.md
```diff
@@ -24,7 +24,7 @@ In this realase, the various implementations of the `RecordQueryPlan` interface
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Permuted grouped aggregate [(Issue #1085)](https://github.com/FoundationDB/fdb-record-layer/issues/1085)
```
MMcM, line 27: I am open to suggestions for a better name for this type of index.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexTypes.java
```diff
@@ -147,6 +147,16 @@
...
 
+    /**
+     * An index remmebering minimum values in a way that can enumerate minima ordered by value and not (part of the) group.
+     */
+    public static final String PERMUTED_MIN = "permuted_" + FunctionNames.MIN;
+
+    /**
+     * An index remmebering maximum values in a way that can enumerate maxima ordered by value and not (part of the) group.
+     */
+    public static final String PERMUTED_MAX = "permuted_" + FunctionNames.MAX;
```
MMcM, line 158: Since there is some duplication, it would also be possible to have a `MIN_AND_MAX` type.

MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexMaintainer.java
```diff
@@ -0,0 +1,188 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * An index that maintains an extremum value in a way that can be enumated by that value rather than by group.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class PermutedMinMaxIndexMaintainer extends StandardIndexMaintainer {
```
alecgrieser, line 56: Does this need to override `canDeleteWhere` and/or `deleteWhere` to delete items from the secondary subspace?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+
+/**
+ * Tests for permuted min / max type indexes.
+ */
+@Tag(Tags.RequiresFDB)
+public class PermutedMinMaxIndexTest extends FDBRecordStoreTestBase {
```
alecgrieser, line 50: I suppose in theory, I could see this also having:

1. A test where the grouping expression is zero columns
2. A test where the permuted size is zero
3. A test of delete where

I'm also fine with merging this without the first two tests. A test of delete where would probably be good to have

alecgrieser (changes requested):
It might also be good to get @normen662's opinion on exactly how easy/hard it will be to get the planner to use this kind of index at some point, especially the permuting part.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexTypes.java
MMcM: Yes, you would.

docs/ReleaseNotes.md
MMcM: Hmm. Well, a `VALUE` index can do `MIN` and `MAX`, as they are at the ends of a range. I guess having no permutation could mean skip the secondary keyspace maintenance, but then the index isn't good for other kinds of queries.

I'm inclined to have the index reject things that aren't useful rather than just things that aren't possible. But maybe I'm not thinking about it right.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexTypes.java
```diff
@@ -147,6 +147,16 @@
      */
     public static final String BITMAP_VALUE = "bitmap_value";
 
+    /**
+     * An index remmebering minimum values in a way that can enumerate minima ordered by value and not (part of the) group.
+     */
+    public static final String PERMUTED_MIN = "permuted_" + FunctionNames.MIN;
```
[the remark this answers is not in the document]
MMcM: That is also true of `RANK` indexes, but we didn't do that for just the reason you give. So maybe something to revisit with the new planner.

Another way to do all the would be to have indexes that require that another index (in these cases, a particular `VALUE`) also exist in the schema and then use it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexMaintainer.java
```diff
@@ -0,0 +1,188 @@
...
+    public PermutedMinMaxIndexMaintainer(@Nonnull IndexMaintainerState state) {
+        super(state);
+        type = getType(state.index);
+        permutedSize = getPermutedSize(state.index);
+    }
+
+    protected static int getPermutedSize(@Nonnull Index index) {
+        String permutedSizeOption = index.getOption(IndexOptions.PERMUTED_SIZE_OPTION);
+        if (permutedSizeOption == null) {
+            throw new MetaDataException("permuted size not specified", LogMessageKeys.INDEX_NAME, index.getName());
```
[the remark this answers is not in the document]
MMcM: Yes, but a `VALUE` index can do that and more.

MMcM:
> how easy/hard it will be to get the planner to use this kind of index at some point

We did discuss this in the secret planner chatcave.

The expanded form of an index for matching will have a `group-by` expression for (existing) aggregates and will (sooner than the previous) have a projection made up of qualified fields from the graph below (a "`SELECT`" list) for sort-compatibility. These two together _seem_ to us to handle permutation okay.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexMaintainer.java
```diff
@@ -0,0 +1,188 @@
...
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * An index that maintains an extremum value in a way that can be enumated by that value rather than by group.
+ */
```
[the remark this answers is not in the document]
MMcM: Added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexMaintainer.java
MMcM: Yes, it does. Added.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
MMcM: 1 & 2 fail index validation, again on the principle that an index that doesn't do anything useful should be rejected.

Test for delete where included with that addition.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexMaintainer.java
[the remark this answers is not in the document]
alecgrieser: I think that this index would be better than a `VALUE` index if you wanted to list all of the extrema (by group), not just calculate the extremum for a single group. I suppose you could do that with a skip-scan on a `VALUE` index, though it wouldn't be as performant.

alecgrieser (approved):
I think I could see use cases where `permuted_size == 0` would be useful over a `VALUE` (mainly when scanning `BY_GROUP` but where there is at least one grouping column unspecified), but I also think we could remove the part of the validator that throws an error if `permuted_size == 0` at a later date without any problems, so I'm also happy to take this PR as is.

I'll leave it for you to merge in case you want to respond to the `permuted_size` comments

MMcM:
Yes, that makes sense. If you wanted to scan multiple groups, you would want pre-computed min/max values, which that would give. I still hesitate to call the indexes `MIN` and `MAX`, for fear that people will think they are needed for the single group case.

alecgrieser:
Yeah, I think that's fair

PR #1097, opened 2021-01-11
Resolves #737: Count indexes should be cleared when decremented to zero

### LINKED ISSUE #737: Count indexes should be cleared when decremented to zero
The [`COMPARE_AND_CLEAR`](https://apple.github.io/foundationdb/javadoc/com/apple/foundationdb/MutationType.html#COMPARE_AND_CLEAR) mutation type allows the user specify a value and to conditionally clear a key if the value is equal to that key. The `count` and possibly the `sum` index types should use this to remove keys that have been decremented to zero. In particular, this is good as with grouped aggregates, one can wind up in situations where if the grouping key is something that increases with time, then older grouping keys will never be garbage collected out. (For example, imagine a grouping column of, say, "day" to bucket together the number of records per day. Then if only *k* days of history are kept around, there will only be at most *k* non-zero index keys, but the older days are never cleared out.) Both the `count` and `sum` indexes treat null (i.e., unset keys) identically to keys set to zero.

MMcM:
This takes the approach suggested in the comment of making this an optional behavior defaulting to off, for the reasons stated there: it avoids incompatible changes and makes `SUM` and `COUNT` symmetrical.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
@@ -110,6 +110,11 @@ public class IndexOptions {
      */
     public static final String BITMAP_VALUE_ENTRY_SIZE_OPTION = "bitmapValueEntrySize";
 
+    /**
+     * Whether to remove index entry for {@link IndexTypes#COUNT} type indexes when they decrement to zero.
+     */
+    public static final String CLEAR_WHEN_ZERO = "clearWhenZero";
+
     private IndexOptions() {
     }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
@@ -112,6 +112,13 @@ public class IndexOptions {
 
     /**
      * Whether to remove index entry for {@link IndexTypes#COUNT} type indexes when they decrement to zero.
+     *
+     * This makes the existence of zero-valued entries in the index in the face of updates and deletes
+     * closer to what it would be if the index were rebuilt, but still not always the same.
+     * In particular,<ul>
+     *   <li>A {@code SUM} index will not have entries for groups all of whose indexed values are zero.</li>
+     *   <li>Changing the option for an existing index from {@code false} to {@code true} does not clear any entries.</li>
+     * </ul>
      */
     public static final String CLEAR_WHEN_ZERO = "clearWhenZero";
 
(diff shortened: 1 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
```diff
@@ -110,6 +110,11 @@
      */
     public static final String BITMAP_VALUE_ENTRY_SIZE_OPTION = "bitmapValueEntrySize";
 
+    /**
+     * Whether to remove index entry for {@link IndexTypes#COUNT} type indexes when they decrement to zero.
+     */
+    public static final String CLEAR_WHEN_ZERO = "clearWhenZero";
```
alecgrieser, line 116: It feels like we should have an opinion about whether this index option can be changed or not (without needing to rebuild the index). From what I can tell, it's fine to change it from `false` to `true`, but you might get some zeros back on older data (unless we also add a filter in the index scan to remove zeros), and it's fine to change it from `true` to `false` except you might miss some zeros. Not sure if that's something we should mention/encourage, or if we should tell people to rebuild the index if they want to change this option. It also seems like we should document the decision in the Javadoc here.

Thoughts?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/AtomicMutationIndexMaintainer.java
```diff
@@ -150,6 +155,8 @@ protected static AtomicMutation getAtomicMutation(@Nonnull Index index) {
                 }
             }
 
+            final byte[] compareAndClear = mutation.getCompareAndClearParam();
```
alecgrieser, line 158: minor: this feels like this local variable should be declared closer to where it's used (like right before it's checked for `null`-ity)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
MMcM: I considered whether to override `validateChangedOptions` and concluded that it was not appropriate. Changing the option does not make the index invalid. You just have to understand that the option refers to clearing at the time of decrement to zero. And the caller is free to cause the index to be rebuilt if their requirement is that zero entries go away. There is currently nothing in record layer that makes this kind of decision.

I added a comment about this. Maybe it should go into some documentation someplace, instead?

It's like turning it on with `SUM`; it's possible that some entries will get cleared that would be `0` if rebuilt.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/AtomicMutationIndexMaintainer.java
MMcM: Sure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
alecgrieser: Yeah, I think the contract makes sense. I think that comment gets the idea across. I do feel like we should make it available in the published Javaodoc, and I was thinking that this Javadoc here on the option would be the logical place to document it, but maybe I'm wrong.

alecgrieser (approved):

alecgrieser:
Will squash and merge when the PRB completes

PR #1099, opened 2021-01-13
Resolves #1098:Need a way to tell whether the result of planning …

### LINKED ISSUE #1098: Need a way to tell whether the result of planning a query can only generate at most one record
I believe the new planner will have this as a first-class attribute of its intermediate form. But, for now, it can be a method on the `QueryPlan` itself, since it's kind of an accident and not something the existing planner reasons about explicitly.

MMcM:
… a query can only generate at most one record

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
@@ -25,6 +25,7 @@ import com.apple.foundationdb.record.EvaluationContext;
 import com.apple.foundationdb.record.ExecuteProperties;
 import com.apple.foundationdb.record.PlanHashable;
 import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.RecordMetaData;
 import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
 import com.apple.foundationdb.record.query.plan.temp.RelationalExpression;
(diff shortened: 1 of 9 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
```diff
@@ -128,6 +129,15 @@
      */
     boolean hasLoadBykeys();
 
+    /**
+     * Indicates whether this plan can only return one or zero records.
+     * @param metaData meta-data to use to determine index uniqueness
+     * @return <code>true</code> if this plan returns at most a single record.
+     */
+    default boolean isUnique(@Nonnull RecordMetaData metaData) {
```
normen662, line 137: The name feels like it's misleading as it is not about uniqueness but about maximum cardinality. Other planners use the name maxCardinality which can be `UNKNOWN` and a real number such as `0`, `1`. Where you would mostly expect `1` or `UNKNOWN`. But even things like other actual numbers can be useful, i.e. for understanding the union of three arms of a `UNION` of `maxCardinality == 1` arms would be `3`.

normen662 (commented):
Some general comments:
- this is only on `Plan`s but it would be really useful on `RelationExpression`s as well
- in order to not clutter the interface on `RecordQueryPlan` or `RelationalExpression` can we maybe put this into a `PlannerProperty`? It may not be useful if we have too many different implementations.
- what is the specific semantics if you consider joins to be first class citizens in the plan (in the future). This property should be giving you back the property assuming the _outer_ is bound so it should return the max cardinality per outer. I thought I should clarify that.
- There is a more general version of all of this that computes the max cardinality over a plan given an additional set of bindings (optional), e.g., what's the maxCardinality of a sub plan if I also consider that `x` is bound to some value. That may be useful to determine if a certain transformation should fire or not.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
MMcM: Changed to `maxCardinality`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
```diff
@@ -290,6 +303,11 @@ public ParameterKeySource(String parameter) {
             return (List<Tuple>)context.getBinding(parameter);
         }
 
+        @Override
```
normen662, line 306: That's unfortunate as this number is usually not exactly gigantic.

normen662 (approved):

PR #1105, opened 2021-01-19
Resolves #1103: Configure default nullability annotations in .idea/misc.xml

### LINKED ISSUE #1103: Configure default nullability annotations in .idea/misc.xml
We use `javax.annotation.Nullable` and `javax.annotation.Nonnull` as our default nullability annotations. We should configure that in our `.idea` configuration as the defaults so that developers use the right ones by default.

alecgrieser:
Changes the `.idea` configuration so that the default nullability annotation is the one that the project actually uses (at least currently). This should help developers out in using the right one.

This resolves #1103.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -5,6 +5,49 @@
       <item index="0" class="java.lang.String" itemvalue="com.google.auto.service.AutoService" />
     </list>
   </component>
+  <component name="NullableNotNullManager">
+    <option name="myDefaultNullable" value="javax.annotation.Nullable" />
+    <option name="myDefaultNotNull" value="javax.annotation.Nonnull" />
+    <option name="myNullables">
+      <value>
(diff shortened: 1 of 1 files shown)


.idea/misc.xml
```diff
@@ -5,6 +5,49 @@
       <item index="0" class="java.lang.String" itemvalue="com.google.auto.service.AutoService" />
     </list>
   </component>
+  <component name="NullableNotNullManager">
+    <option name="myDefaultNullable" value="javax.annotation.Nullable" />
+    <option name="myDefaultNotNull" value="javax.annotation.Nonnull" />
+    <option name="myNullables">
+      <value>
+        <list size="14">
```
normen662, line 13: there are no words that allow me to express my dislike of XML

normen662 (approved):

PR #1106, opened 2021-01-19
Resolves #1076: [bug] Continuation unused in scanIndex with EndpointT…

### LINKED ISSUE #1076: [bug] Continuation unused in scanIndex with EndpointType.PREFIX_STRING
There seems to be a bug in the scanIndex API of Record Layer. When using this API with EndpointType PREFIX_STRING, the continuation logic does not work as expected. Instead of continuing from the point specified by the continuation bytes, the scan starts from the beginning.
The code causing the bug seems to be [here](https://github.com/FoundationDB/fdb-record-layer/blob/f310b6803ba56da0ae2296be3966307194468dc6/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursor.java#L235-L240). As the endpoint type is overwritten with EndpointType.CONTINUATION, the null character of the string does not get trimmed and the scan, therefore, starts from the beginning of the index.

dlambrig:
A `PREFIX_STRING` provided by the user is encoded with null termination. When scanning using endpoint type `PREFIX_STRING`, on continuations, this is concatenated with the last key from the previous scan. In so doing, the prefix's null terminator should not be included. 

e.g. suppose there were four keys:

`cata
catb
catc
catd`

.. and `PREFIX_STRING` was `cat`, then the objective is to find all four tuples with key prefixed by `cat`.

if the first scan located two tuples `cata` and `catb`, and then paused (for any reason), then, the second scan would continue where the first left off. In which case, the "last key" in the continuation structure would be `b` (the suffix after `cat`).

the bug was: a null terminator was included between `cat` and `b`, resulting in continuing from `cat\0b`.  That resulted in failed scans; rather than proceeding to keys `catc` and `catd`, the scan would restart at the beginning `cata`.

the solution is to not include the null terminator when concatenating the last key suffix with the prefix.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
@@ -118,6 +118,40 @@ public class KeyValueCursorTest extends FDBTestBase {
             return null;
         });
     }
+    @SuppressWarnings("checkstyle:WhitespaceAround")
+    @Test
+    public void prefixString() {
+
+        fdb.database().run(tr -> {
+            for (char c = 'a'; c < 'k'; c++) {
+                String key = "cat" + c;
+                tr.set(subspace.pack(Tuple.from(key)),Tuple.from(Character.toString(c)).pack());
+            }
+            return null;
+        });
+
+        fdb.run(context -> {
+            KeyValueCursor cursor = KeyValueCursor.Builder.withSubspace(subspace)
+                    .setContext(context)
+                    .setRange(TupleRange.prefixedBy("cat"))
+                    .setContinuation(null)
+                    .setScanProperties(new ScanProperties(ExecuteProperties.newBuilder().setReturnedRowLimit(2).build()))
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
```diff
@@ -174,16 +208,16 @@ public void inclusiveRange() {
                     .setContinuation(null)
                     .setScanProperties(new ScanProperties(ExecuteProperties.newBuilder().setReturnedRowLimit(2).build()))
                     .build();
-            assertEquals(Arrays.asList(Tuple.from(3L, 3L), Tuple.from(3L, 4L)),
+            assertEquals(Arrays.asList(Tuple.from(3L, 0L), Tuple.from(3L, 4L)),
```
dlambrig, line 211: changes in this unit test will be reverted to original on next revision

jleach4:
@dlambrig Will get our build team to look at this.  Sorry for the issue.

alecgrieser:

PR #1117, opened 2021-02-02
Resolve 1107:  Add persistent continuation to 'indexing by index'

jjezra:
In the records store:
        * an indexing type stamp will ensure that a continuation is only allowed for the same indexing type
        * the rangeSet is now used to correlated indexing by index (similarly to indexing by records)

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -26,6 +26,7 @@ import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncUtil;
 import com.apple.foundationdb.async.MoreAsyncUtil;
 import com.apple.foundationdb.async.RangeSet;
+import com.apple.foundationdb.record.IndexBuildProto;
 import com.apple.foundationdb.record.IndexState;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCursor;
@@ -38,6 +39,7 @@ import com.apple.foundationdb.record.provider.foundationdb.synchronizedsession.S
 import com.apple.foundationdb.record.query.plan.synthetic.SyntheticRecordFromStoredRecordPlan;
 import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.Tuple;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -45,7 +47,6 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
@@ -75,7 +76,7 @@ public abstract class IndexingBase {
     private final IndexingThrottle throttle;
 
     private long timeOfLastProgressLogMillis = 0;
-    private boolean fallbackMode = false;
+    private boolean forceStampOverwrite = false;
 
     IndexingBase(IndexingCommon common) {
         this.common = common;
@@ -253,51 +254,64 @@ public abstract class IndexingBase {
     }
 
     @Nonnull
-    public IndexingBase setFallbackMode() {
-        fallbackMode = true;
-        return this;
+    public void setFallbackMode() {
+        forceStampOverwrite = true; // must overwrite a previous indexing method's stamp
     }
 
     @Nonnull
     private CompletableFuture<Void> setIndexingTypeOrThrow() {
-        byte[] stamp = getTypeStamp();
-        final List<Object> additionalLogMessageKeyValues = Arrays.asList(LogMessageKeys.CALLING_METHOD, "setIndexingTypeOrThrow",
-                LogMessageKeys.LOCAL_VERSION, stamp);
-
-        return buildCommitRetryAsync( (store, ignore) -> {
+        return getRunner().runAsync(context -> openRecordStore(context).thenCompose(store -> {
             Transaction transaction = store.getContext().ensureActive();
+            IndexBuildProto.IndexBuildIndexingStamp indexingTypeStamp = getIndexingTypeStamp();
             byte[] stampKey = indexBuildTypeSubspace(store, common.getIndex()).getKey();
             return transaction.get(stampKey)
-                            .thenApply( bytes -> {
-                                if (bytes == null || fallbackMode) {
-                                    transaction.set(stampKey, stamp);
-                                } else if (! matchingTypeStamp(bytes)) {
-                                    if (LOGGER.isWarnEnabled()) {
-                                        LOGGER.warn(KeyValueLogMessage.of("Index was partly built by another method",
-                                                LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
-                                                LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
-                                                LogMessageKeys.INDEXER_ID, common.getUuid(),
-                                                LogMessageKeys.EXPECTED_TYPE, stamp,
-                                                LogMessageKeys.ACTUAL_TYPE, bytes));
-                                    }
-                                    throw new RecordCoreException("This index was partly built by another method");
-                                }
-                                return null;
-                            }); },
-                false,
-                additionalLogMessageKeyValues);
+                    .thenApply( bytes -> {
+                        if (bytes == null || forceStampOverwrite) {
+                            transaction.set(stampKey, indexingTypeStamp.toByteArray());
+                        } else {
+                            IndexBuildProto.IndexBuildIndexingStamp savedStamp ;
+                            try {
+                                savedStamp = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(bytes);
+                            } catch (InvalidProtocolBufferException ex) {
+                                savedStamp = null;
+                            }
+                            if ( savedStamp == null || ! matchingIndexingTypeStamp(savedStamp)) {
+                                throw new RecordCoreException("This index was partly built by another method",
+                                        LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
+                                        LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
+                                        LogMessageKeys.INDEXER_ID, common.getUuid(),
+                                        LogMessageKeys.EXPECTED_TYPE, indexingTypeStamp,
+                                        LogMessageKeys.ACTUAL_TYPE, savedStamp);
+                            }
+                        }
+                        return null;
+                    });
+        } ));
     }
 
     @Nonnull
-    abstract byte[] getTypeStamp();
+    abstract IndexBuildProto.IndexBuildIndexingStamp getIndexingTypeStamp();
 
     @Nonnull
-    abstract boolean matchingTypeStamp(byte[] stamp);
+    abstract boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndexingStamp stamp);
 
     abstract CompletableFuture<Void> buildIndexByEntityAsync();
 
     // Helpers for implementing modules. Some of them are public to support unit-testing.
-    protected boolean shouldLogBuildProgress() {
+    protected void maybeLogBuildProgress(SubspaceProvider subspaceProvider, List<Object> additionalLogMessageKeyValues) {
+        if (LOGGER.isInfoEnabled() && shouldLogBuildProgress()) {
+            final Index index = common.getIndex();
+            LOGGER.info(KeyValueLogMessage.of("Built Range",
+                    LogMessageKeys.INDEX_NAME, index.getName(),
+                    LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
+                    subspaceProvider.logKey(), subspaceProvider,
+                    LogMessageKeys.RECORDS_SCANNED, common.getTotalRecordsScanned().get()),
+                    LogMessageKeys.INDEXER_ID, common.getUuid(),
+                    additionalLogMessageKeyValues);
+        }
+    }
+
+    private boolean shouldLogBuildProgress() {
         long interval = common.config.getProgressLogIntervalMillis();
         long now = System.currentTimeMillis();
         if (interval == 0 || interval < (now - timeOfLastProgressLogMillis)) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.provider.foundationdb;
 
 import com.apple.foundationdb.FDBException;
+import com.apple.foundationdb.MutationType;
 import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncUtil;
@@ -52,11 +53,14 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
+
+import static com.apple.foundationdb.record.RecordCursor.NoNextReason.SOURCE_EXHAUSTED;
 
 /**
  * A base class for different types of online indexing scanners.
@@ -349,11 +353,11 @@ public abstract class IndexingBase {
     }
 
     protected <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore store,
-                                                           @Nonnull RecordCursor<T> cursor,
+                                                           @Nonnull RecordCursor<T> cursor, // at least two items must be requested for this cursor
                                                            @Nonnull Function<RecordCursorResult<T>, FDBStoredRecord<Message>> getRecord,
-                                                           @Nonnull Consumer<RecordCursorResult<T>> lastResultSet,
-                                                           @Nonnull AtomicBoolean isEmpty,
-                                                           @Nonnull AtomicLong recordsScannedCounter) {
+                                                           @Nonnull Consumer<RecordCursorResult<T>> nextResultCont,
+                                                           @Nonnull AtomicBoolean hasMore,
+                                                           @Nullable AtomicLong recordsScanned) {
         final FDBStoreTimer timer = getRunner().getTimer();
         final Index index = common.getIndex();
         final IndexMaintainer maintainer = store.getIndexMaintainer(index);
@@ -363,20 +367,50 @@ public abstract class IndexingBase {
         // Need to do this each transaction because other index enabled state might have changed. Could cache based on that.
         // Copying the state also guards against changes made by other online building from check version.
         // TODO: need some state to avoid generating the same synthetic record via more than one self-join path for non-idempotent indexes.
+        AtomicLong recordsScannedCounter = new AtomicLong();
+
+        final AtomicReference<RecordCursorResult<T>> nextResult = new AtomicReference<>(null);
         return AsyncUtil.whileTrue(() -> cursor.onNext().thenCompose(result -> {
-            if (!result.hasNext()) {
+            RecordCursorResult<T> currResult;
+            boolean isExhausted = false;
+            if (result.hasNext()) {
+                // has next, process one previous item (if exists)
+                currResult = nextResult.get();
+                nextResult.set(result);
+                if (currResult == null) {
+                    // that was the first item, nothing to process
+                    return AsyncUtil.READY_TRUE;
+                }
+            } else {
                 // end of the cursor list
                 timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_COUNT);
-                lastResultSet.accept(result);
-                return AsyncUtil.READY_FALSE;
+                if (result.getNoNextReason() != SOURCE_EXHAUSTED) {
+                    // should we trigger a warning (or an exception) if nextResult is null - i.e. non-exhausted empty list?
+                    nextResultCont.accept(nextResult.get());
+                    hasMore.set(true);
+                    return AsyncUtil.READY_FALSE;
+                }
+                // source is exhausted, fall down to handle the last item and return with hasMore=false
+                currResult = nextResult.get();
+                if (currResult == null) {
+                    // this was an empty list
+                    hasMore.set(false);
+                    return AsyncUtil.READY_FALSE;
+                }
+                // here, process the last item and return
+                nextResult.set(null);
+                isExhausted = true;
             }
-
-            final FDBStoredRecord<Message> rec = getRecord.apply(result);
-            isEmpty.set(false);
+            // here: currResult must have value
+            final FDBStoredRecord<Message> rec = getRecord.apply(currResult);
             timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED);
             recordsScannedCounter.incrementAndGet();
             if (!common.recordTypes.contains(rec.getRecordType())) {
                 // This record is not our type, swipe left
+                if (isExhausted) {
+                    hasMore.set(false);
+                    return AsyncUtil.READY_FALSE;
+                }
                 return AsyncUtil.READY_TRUE;
             }
             // add this index to the transaction
@@ -386,19 +420,36 @@ public abstract class IndexingBase {
             timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
 
             final CompletableFuture<Void> updateMaintainer = updateMaintainerBuilder(syntheticPlan, rec, maintainer, store);
-
+            if (isExhausted) {
+                // we've just processed the last item
+                hasMore.set(false);
+                return updateMaintainer.thenApply(vignore -> false);
+            }
             return updateMaintainer.thenCompose(vignore ->
                     context.getApproximateTransactionSize().thenApply(size -> {
                         if (size >= common.config.getMaxWriteLimitBytes()) {
                             // the transaction becomes too big - stop iterating
                             timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_SIZE);
-                            lastResultSet.accept(result);
+                            nextResultCont.accept(nextResult.get());
+                            hasMore.set(true);
                             return false;
                         }
                         return true;
                     }));
 
-        }), cursor.getExecutor());
+        }), cursor.getExecutor()
+        ).thenApply(vignore -> {
+            long recordsScannedInTransaction = recordsScannedCounter.get();
+            if (recordsScanned != null) {
+                recordsScanned.addAndGet(recordsScannedInTransaction);
+            }
+            if (common.isTrackProgress()) {
+                final Subspace scannedRecordsSubspace = indexBuildScannedRecordsSubspace(store, index);
+                store.context.ensureActive().mutate(MutationType.ADD, scannedRecordsSubspace.getKey(),
+                        FDBRecordStore.encodeRecordCount(recordsScannedInTransaction));
+            }
+            return null;
+        });
     }
 
     private static CompletableFuture<Void> updateMaintainerBuilder(SyntheticRecordFromStoredRecordPlan syntheticPlan,
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -241,7 +241,7 @@ public abstract class IndexingBase {
     private CompletableFuture<Void> doBuildIndexAsync(boolean markReadable) {
         CompletableFuture<Void> buildFuture =
                 setIndexingTypeOrThrow()
-                        .thenCompose(ignore -> buildIndexByEntityAsync());
+                        .thenCompose(ignore -> buildIndexInternalAsync());
 
         if (markReadable) {
             return buildFuture.thenCompose(vignore ->
@@ -269,7 +269,7 @@ public abstract class IndexingBase {
                         if (bytes == null || forceStampOverwrite) {
                             transaction.set(stampKey, indexingTypeStamp.toByteArray());
                         } else {
-                            IndexBuildProto.IndexBuildIndexingStamp savedStamp ;
+                            IndexBuildProto.IndexBuildIndexingStamp savedStamp;
                             try {
                                 savedStamp = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(bytes);
                             } catch (InvalidProtocolBufferException ex) {
@@ -280,8 +280,8 @@ public abstract class IndexingBase {
                                         LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
                                         LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
                                         LogMessageKeys.INDEXER_ID, common.getUuid(),
-                                        LogMessageKeys.EXPECTED_TYPE, indexingTypeStamp,
-                                        LogMessageKeys.ACTUAL_TYPE, savedStamp);
+                                        LogMessageKeys.EXPECTED, indexingTypeStamp,
+                                        LogMessageKeys.ACTUAL, savedStamp);
                             }
                         }
                         return null;
@@ -295,19 +295,20 @@ public abstract class IndexingBase {
     @Nonnull
     abstract boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndexingStamp stamp);
 
-    abstract CompletableFuture<Void> buildIndexByEntityAsync();
+    abstract CompletableFuture<Void> buildIndexInternalAsync();
 
     // Helpers for implementing modules. Some of them are public to support unit-testing.
     protected void maybeLogBuildProgress(SubspaceProvider subspaceProvider, List<Object> additionalLogMessageKeyValues) {
         if (LOGGER.isInfoEnabled() && shouldLogBuildProgress()) {
             final Index index = common.getIndex();
-            LOGGER.info(KeyValueLogMessage.of("Built Range",
+            LOGGER.info(KeyValueLogMessage.build("Built Range",
                     LogMessageKeys.INDEX_NAME, index.getName(),
                     LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
                     subspaceProvider.logKey(), subspaceProvider,
-                    LogMessageKeys.RECORDS_SCANNED, common.getTotalRecordsScanned().get()),
-                    LogMessageKeys.INDEXER_ID, common.getUuid(),
-                    additionalLogMessageKeyValues);
+                    LogMessageKeys.RECORDS_SCANNED, common.getTotalRecordsScanned().get(),
+                    LogMessageKeys.INDEXER_ID, common.getUuid())
+                    .addKeysAndValues(additionalLogMessageKeyValues)
+                    .toString());
         }
     }
 
@@ -362,7 +363,6 @@ public abstract class IndexingBase {
         // Need to do this each transaction because other index enabled state might have changed. Could cache based on that.
         // Copying the state also guards against changes made by other online building from check version.
         // TODO: need some state to avoid generating the same synthetic record via more than one self-join path for non-idempotent indexes.
-
         return AsyncUtil.whileTrue(() -> cursor.onNext().thenCompose(result -> {
             if (!result.hasNext()) {
                 // end of the cursor list
@@ -426,12 +426,12 @@ public abstract class IndexingBase {
         // (2) to allow for write-only indexes to continue to do the right thing.
         RangeSet rangeSet = new RangeSet(store.indexRangeSubspace(index));
         CompletableFuture<Boolean> rangeFuture = rangeSet.insertRange(tr, null, null);
-        CompletableFuture<Void> buildFuture = rebuildIndexByEntityAsync(store);
+        CompletableFuture<Void> buildFuture = rebuildIndexInternalAsync(store);
 
         return CompletableFuture.allOf(rangeFuture, buildFuture);
     }
 
-    abstract CompletableFuture<Void> rebuildIndexByEntityAsync(FDBRecordStore store);
+    abstract CompletableFuture<Void> rebuildIndexInternalAsync(FDBRecordStore store);
 
     // These throttle methods are externalized for testing usage only
     @Nonnull
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -56,11 +56,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
-import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
-
-import static com.apple.foundationdb.record.RecordCursor.NoNextReason.SOURCE_EXHAUSTED;
 
 /**
  * A base class for different types of online indexing scanners.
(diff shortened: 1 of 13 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -282,4 +284,116 @@ public void testIndexFromIndexNoFallbackNonValueSrc() {
...
+        final AtomicLong counter = new AtomicLong(0);
+        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setIndex(tgtIndex).setSubspace(subspace)
+                .setIndexFromIndex(OnlineIndexer.IndexFromIndexPolicy.newBuilder()
+                        .setSourceIndex("src_index")
+                        .forbidRecordScan()
+                        .build())
+                .setLimit(chunkSize)
+                .setTimer(timer)
+                .setConfigLoader(old -> {
```
jjezra, line 316: This was Scott Gray's idea - to add a lambda entity that allows tests interrupting indexing processes. The lambda already exists, and it seems to be more deterministic than short timers.

jjezra:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -182,20 +190,98 @@ private void maybeLogBuildProgress(SubspaceProvider subspaceProvider, byte[] ret
...
+        return ranges.onHasNext().thenCompose(hasNext -> {
+            if (Boolean.FALSE.equals(hasNext)) {
+                return AsyncUtil.READY_FALSE; // no more missing ranges - all done
+            }
+            Range range = ranges.next();
+            Tuple rangeStart = Arrays.equals(range.begin, START_BYTES) ? null : Tuple.fromBytes(range.begin);
+            Tuple rangeEnd = Arrays.equals(range.end, END_BYTES) ? null : Tuple.fromBytes(range.end);
+
+            RecordCursor<FDBIndexedRecord<Message>> cursor =
+                    store.scanIndexRecords(srcIndex, IndexScanType.BY_VALUE, TupleRange.between(rangeStart, rangeEnd), range.begin, scanProperties);
```
alecgrieser, line 217: It feels wrong to me that this is using `range.begin` as the continuation into `scanIndexRecords` here. It's typically not good practice to use a continuation from anything other than `getContinuation` or to treat it as anything other than an opaque byte array. For example, we reserve the right to, say, wrap the continuation in a protobuf at some point in the future, at which point this will break mysteriously. So, I think this should be `null` here, and then I have a suggestion below about how to not grab the continuation any more.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -182,20 +190,98 @@ private void maybeLogBuildProgress(SubspaceProvider subspaceProvider, byte[] ret
...
+                    isEmpty, recordsScannedCounter
+            ).thenCompose(vignore -> {
+                long recordsScannedInTransaction = recordsScannedCounter.get();
+                recordsScanned.addAndGet(recordsScannedInTransaction);
+                if (common.isTrackProgress()) {
+                    store.context.ensureActive().mutate(MutationType.ADD, scannedRecordsSubspace.getKey(),
+                            FDBRecordStore.encodeRecordCount(recordsScannedInTransaction));
+                }
+                final byte[] realEnd =
+                        isEmpty.get() ? null : lastResult.get().getContinuation().toBytes();
```
alecgrieser, line 235: This feels like there's a lot of code dupe between this and the other `buildRangeOnly`.

I think some of that could be removed if the `iterateRangeOnly` took a lambda that would take a record and produced a tuple to use for the range set. For the `byRecords` method, it would return `getPrimaryKey()` but for this version, it would return `.getIndexEntry().getKey()`. And then `iterateRangeOnly` could update the range set (maybe it gets a name change), and then also, you might be able to have unified logic for selecting missing ranges, etc.

There's also the idea from Scott Dugas about packaging some of that up into a "records producer" that knows how to produce records within "some range" and also tell you the proper key to use for a given record.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,8 +252,51 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+                additionalLogMessageKeyValues);
+    }
+
+    @Nonnull
+    abstract byte[] getTypeStamp();
+
+    @Nonnull
+    abstract boolean matchingTypeStamp(byte[] stamp);
+
     abstract CompletableFuture<Void> buildIndexByEntityAsync();
```
alecgrieser, line 297: What is an "entity" in this context? I guess it's not new to this branch, but it seems like a weird word to use.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -182,20 +190,98 @@ private void maybeLogBuildProgress(SubspaceProvider subspaceProvider, byte[] ret
...
+                    if (cont == null) {
+                        return false;
+                    }
+                    nextCont.set(cont);
+                    return true;
+                }), store.getExecutor());
+    }
+
+    @Nonnull
+    private CompletableFuture<byte[]> rebuildRangeOnly(@Nonnull FDBRecordStore store, byte[] cont, @Nonnull AtomicLong recordsScanned) {
```
alecgrieser, line 260: Was the intention to add the range information to this implementation as well? I don't see where that was added.

alecgrieser (changes requested):
I haven't looked over the tests (yet), but here are some comments I found looking over the non-test code.

It's also probably worth it for one of us to look over this "as a whole" to verify if it implements @nblintao's flowchart correctly, as I'm not _entirely_ sure it does

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -355,7 +419,7 @@ private static void timerIncrement(@Nullable FDBStoreTimer timer, FDBStoreTimer.
 
     abstract CompletableFuture<Void> rebuildIndexByEntityAsync(FDBRecordStore store);
 
-    // These two methods are externalized for testing usage only
+    // These throttle methods are externalized for testing usage only
```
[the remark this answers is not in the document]
jjezra: I removed the `@VisibleForTesting` annotation from these functions because it caused compilation warnings (as being called by a non-testing function, although itself marked as `@VisableForTesting`).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,8 +252,51 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+                LogMessageKeys.LOCAL_VERSION, stamp);
+
+        return buildCommitRetryAsync( (store, ignore) -> {
+            Transaction transaction = store.getContext().ensureActive();
+            byte[] stampKey = indexBuildTypeSubspace(store, common.getIndex()).getKey();
+            return transaction.get(stampKey)
+                            .thenApply( bytes -> {
+                                if (bytes == null || fallbackMode) {
+                                    transaction.set(stampKey, stamp);
+                                } else if (! matchingTypeStamp(bytes)) {
```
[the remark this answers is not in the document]
jjezra: `Arrays.equals(bytes, stamp)` was at the first version - I've changed it to the current implementation method to support future changes in the logic. For example, we can add more information - like a timestamp - to the TypeStamp and then let the implementing function decide which field is relevant for matching and which is not. Another example is Multi-indexing permitting continuation when one of the target indexes is deleted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,8 +252,51 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+        final List<Object> additionalLogMessageKeyValues = Arrays.asList(LogMessageKeys.CALLING_METHOD, "setIndexingTypeOrThrow",
+                LogMessageKeys.LOCAL_VERSION, stamp);
+
+        return buildCommitRetryAsync( (store, ignore) -> {
+            Transaction transaction = store.getContext().ensureActive();
+            byte[] stampKey = indexBuildTypeSubspace(store, common.getIndex()).getKey();
+            return transaction.get(stampKey)
+                            .thenApply( bytes -> {
+                                if (bytes == null || fallbackMode) {
+                                    transaction.set(stampKey, stamp);
```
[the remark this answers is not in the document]
nblintao: The name "fallback" might be ambiguous here. I can see there are more than binary options.
A requested indexing method may or may not be usable.
- If the requested method is not usable (for example, the requested source index is not readable or not single-type or not idempotent), we can throw exception for this now. Maybe you don't have to check here since it will be checked in every transaction anyways. But in the future, people may want to have a option for different handlings.
- If the requested method is usable
  - If it is same as the save one, or there is no saved one. Nice and easy.
  - If different, there can be different handlings decided by an enum `OptionWhenIndexingMethodMismatch`:
    - `USE_OLD_INDEXING_METHOD` (aka "fallback")
    - `USE_NEW_INDEXING_METHOD` (force clearing the index and starting from beginning using the new method)
    - `EXCEPTION` (throw a non-retriable exception)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -66,12 +68,36 @@
...
+        return myTypeStamp;
+    }
+
+    @Override
+    boolean matchingTypeStamp(final byte[] stamp) {
+        return Arrays.equals(stamp, myTypeStamp);
+    }
+
+    @Nonnull
+    private byte[] compileTypeStamp() {
```
nblintao, line 96: I think this is `byte[]` because you manually serialize/deserialize it. Since you've now changed to protobuf, it's nicer to return the protobuf object, so that you can get the fields easily.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -96,64 +122,46 @@
...
-        return getRunner().runAsync(context -> {
-            final Index index = common.getIndex();
-            RangeSet rangeSet = new RangeSet(subspace.subspace(Tuple.from(FDBRecordStore.INDEX_RANGE_SPACE_KEY, index.getSubspaceKey())));
-            TransactionContext tc = context.ensureActive();
-            return rangeSet.insertRange(tc, null, null).thenApply(bignore -> null);
-        });
-    }
-
-    private void maybeLogBuildProgress(SubspaceProvider subspaceProvider, byte[] retCont) {
+    private void maybeLogBuildProgress(SubspaceProvider subspaceProvider) {
```
nblintao, line 130: I don't see this is unique to `IndexingByIndex`.  Maybe it's fine because you are going to get rid of is after finishing the full refactoring.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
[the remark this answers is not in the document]
jjezra: Did you mean `private static byte[] compileTypeStamp(String sourceIndex)`? 
Anyway I need to remove it from the constructor to resolve your previous comment.

nblintao (changes requested):
I haven't looked into tests either. @jjezra Let's talk about your decision of
- the optimization you intended for handling ranges
- and separating two `buildRange` code paths because of that
in our next meeting

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: When I used `null`, the cursor included the `range.begin` - which caused duplication. Is there any other way to exclude it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: At this point, "entity" is record or index. How would you rename it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: Great point! 
I did use a simplified logic in the `indexingByIndex`. If it passed all the tests and successfully used in a controlled environment, I was hopping to return and simplify the `indexingByRecords` in a similar way (and handle the rangeSet in this function) - making your code sharing suggestion much easier. Would you agree?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: It was my understanding that in the `rebuildIndex` path we build the whole index without any commit (let alone `rangeSet` subspace). I understood that FDB uses it as optimizations for very small record stores. 
Am I missing something?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,8 +252,51 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
 
+    @Nonnull
+    public IndexingBase setFallbackMode() {
+        fallbackMode = true;
+        return this;
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> setIndexingTypeOrThrow() {
+        byte[] stamp = getTypeStamp();
```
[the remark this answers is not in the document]
jjezra: as in `getIndexingTypeInformation()`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
nblintao: @alecgrieser This method is for building the index in one transaction, which is used by check version.

We should definitely rename this method next time. If it breaks APIs, let's break it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: `getIndexingTypeStamp`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: digging a little more into the code - 
`range.begin`, if not null, is produced by `lastResult.getContinuation()` and should be used as continuation arg.
`range.end`, if not null, is either defined by the caller or as a missingRange - making a top boundary, therefore used as range arg. 
I've set the range to '(null, range.end)` and added a clarifying comment. 
Now the question is whether or not a continuation should be a legal range end at ` rangeSet.insertRange`. Should it?

jjezra: (adding a thought) 
I believe it is safe to assume that `getContinuation` return ordinal keys, and therefore can be used in `rangeSet`. 
When the time comes, though, we should remember to use continuation keys as range for parallel indexing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +252,56 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
     }
 
     @Nonnull
-    public IndexingBase setFallbackMode() {
-        fallbackMode = true;
-        return this;
+    public void setFallbackMode() {
```
nblintao, line 255: The method name should also be changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
nblintao: Did you fix this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,9 +251,57 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+    @Nonnull
+    public void setFallbackMode() {
+        forceStampOverwrite = true; // must overwrite a previous indexing method's stamp
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> setIndexingTypeOrThrow() {
+        return getRunner().runAsync(context -> openRecordStore(context).thenCompose(store -> {
+            Transaction transaction = store.getContext().ensureActive();
+            byte[] stamp = getIndexingTypeStamp();
```
nblintao, line 263: Also, did you fix here?

nblintao (changes requested):
Have you pushed out all your changes? It seems that some comments are not replied or fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: To me, the logic seems to be that `OnlineIndexer` sets the `Indexer` fallback mode because it is called in the fallback path. The fact that the `Indexer` handles it by forcing a stamp overwrite is an implementation decision and shouldn't be the caller's concern. (Note that `OnlineIndexer` knows nothing about the indexing stamp)

If you disagree, though, I'll change it to `setForceStampOverwrite`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: > `private static IndexMetaDataProto.IndexMetaDataIndexingStamp compileIndexingTypeStamp(Index srcIndex)`. 
??

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +252,56 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
     private CompletableFuture<Void> setIndexingTypeOrThrow() {
-        byte[] stamp = getTypeStamp();
-        final List<Object> additionalLogMessageKeyValues = Arrays.asList(LogMessageKeys.CALLING_METHOD, "setIndexingTypeOrThrow",
-                LogMessageKeys.LOCAL_VERSION, stamp);
-
-        return buildCommitRetryAsync( (store, ignore) -> {
+        return getRunner().runAsync(context -> openRecordStore(context).thenCompose(store -> {
             Transaction transaction = store.getContext().ensureActive();
+            byte[] stamp = getIndexingTypeStamp();
             byte[] stampKey = indexBuildTypeSubspace(store, common.getIndex()).getKey();
```
nblintao, line 264: I think getIndexingTypeStamp can also return IndexMetaDataProto.IndexMetaDataIndexingStamp, and in setIndexingTypeOrThrow you can compare two protobufs (which can provide more detailed differences) rather than two byte arrays by converting the stored stamp to protobuf.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: If I remember correctly, you've asked me to change the usage of `buildCommitRetryAsync` to `runAsync` - what am I missing?

fdb-record-layer-core/src/main/proto/index_metadata.proto
```diff
@@ -0,0 +1,40 @@
...
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+syntax = "proto2";
+
+package com.apple.foundationdb.record;
+import "google/protobuf/descriptor.proto";
+
+option java_outer_classname = "IndexMetaDataProto";
```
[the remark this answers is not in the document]
jjezra: Oops, missed this one. I'll fix now

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -231,8 +252,51 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+                                if (bytes == null || fallbackMode) {
+                                    transaction.set(stampKey, stamp);
+                                } else if (! matchingTypeStamp(bytes)) {
+                                    if (LOGGER.isWarnEnabled()) {
+                                        LOGGER.warn(KeyValueLogMessage.of("Index was partly built by another method",
+                                                LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
+                                                LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
+                                                LogMessageKeys.INDEXER_ID, common.getUuid(),
+                                                LogMessageKeys.EXPECTED_TYPE, stamp,
+                                                LogMessageKeys.ACTUAL_TYPE, bytes));
```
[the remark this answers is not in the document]
jjezra: I'll log to Protobuf message (`toString()` is implied, doesn't it?)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
nblintao: Maybe it's fine to call it `setFallbackMode`. But as a public method, you need a JavaDoc explain what it means and what the caller should expect the behavior to be when they set this option.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +254,64 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+        } ));
     }
 
     @Nonnull
-    abstract byte[] getTypeStamp();
+    abstract IndexBuildProto.IndexBuildIndexingStamp getIndexingTypeStamp();
 
     @Nonnull
-    abstract boolean matchingTypeStamp(byte[] stamp);
+    abstract boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndexingStamp stamp);
```
alecgrieser, line 296: Is there a reason that this isn't `getIndexingTypeStamp().equals(stamp)`? I think that has the same semantics you've implemented, but is resilient to adding additional fields later.

It also probably isn't required for version 1, but having a more descriptive diff and therefore a more specific error message probably be good (some day).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
```diff
@@ -75,30 +75,32 @@
...
     @Nonnull
-    byte[] compileTypeStamp() {
-        return IndexMetaDataProto.IndexMetaDataIndexingStamp.newBuilder()
-                .setMethod(IndexMetaDataProto.IndexMetaDataIndexingStamp.Method.BY_RECORDS)
-                .build().toByteArray();
+    private static IndexBuildProto.IndexBuildIndexingStamp compileIndexingTypeStamp() {
+        return
+                IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                        .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS)
+                        .build();
```
alecgrieser, line 97: I think this message can be a `static final` constant for the whole class, as all instances will have the same message.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
```diff
@@ -403,7 +391,10 @@ private TupleRange computeRecordsRange() {
                     rangeDeque.add(new Range(realEnd.pack(), END_BYTES));
                 }
             }
-            maybeLogBuildProgress(subspaceProvider, startTuple, endTuple, realEnd);
+            maybeLogBuildProgress(subspaceProvider,Arrays.asList(
```
alecgrieser, line 394:
```suggestion
            maybeLogBuildProgress(subspaceProvider, Arrays.asList(
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +254,64 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
+    protected void maybeLogBuildProgress(SubspaceProvider subspaceProvider, List<Object> additionalLogMessageKeyValues) {
+        if (LOGGER.isInfoEnabled() && shouldLogBuildProgress()) {
+            final Index index = common.getIndex();
+            LOGGER.info(KeyValueLogMessage.of("Built Range",
+                    LogMessageKeys.INDEX_NAME, index.getName(),
+                    LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
+                    subspaceProvider.logKey(), subspaceProvider,
+                    LogMessageKeys.RECORDS_SCANNED, common.getTotalRecordsScanned().get()),
+                    LogMessageKeys.INDEXER_ID, common.getUuid(),
+                    additionalLogMessageKeyValues);
```
alecgrieser, line 310: If you go down this route, I think you want something like:

```suggestion
            LOGGER.info(KeyValueLogMessage.build("Built Range",
                   LogMessageKeys.INDEX_NAME, index.getName(),
                   LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
                   subspaceProvider.logKey(), subspaceProvider,
                   LogMessageKeys.RECORDS_SCANNED, common.getTotalRecordsScanned().get(),
                   LogMessageKeys.INDEXER_ID, common.getUuid())
                   .addKeysAndValues(additionalLogMessageKeyValues)
                   .toString());
```

(Alternatively, breaking it up into multiple statements might make sense instead of using the fluent syntax might make sense.) The new code adds the common keys to the message and then adds the additional keys to the end.

I also expect the code as written to throw `IllegalArgumentException("keys and values don't match")` but for a second bug where an extra `)` was added after the `RECORDS_SCANNED` pair.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -213,8 +208,10 @@ private void maybeLogBuildProgress(SubspaceProvider subspaceProvider) {
             Tuple rangeStart = Arrays.equals(range.begin, START_BYTES) ? null : Tuple.fromBytes(range.begin);
             Tuple rangeEnd = Arrays.equals(range.end, END_BYTES) ? null : Tuple.fromBytes(range.end);
 
+            // range.begin, if not null, is produced by getContinuation() and should be used at continuation field
+            // range.end, if not null, is either defined by caller or as a missingRange - making a top range, therefore used at range field
             RecordCursor<FDBIndexedRecord<Message>> cursor =
-                    store.scanIndexRecords(srcIndex, IndexScanType.BY_VALUE, TupleRange.between(rangeStart, rangeEnd), range.begin, scanProperties);
+                    store.scanIndexRecords(srcIndex, IndexScanType.BY_VALUE, TupleRange.between(null, rangeEnd), range.begin, scanProperties);
```
alecgrieser, line 214: I guess this is somewhat better than before, but it still seems to me like this isn't really doing what the range set expects one to do. In particular, I think it will complicate parallelized builds in the future and using an index to build a non-idempotent index, where you really do want the range set to be real ranges. I think the flip side is that this might mean that we'll sometimes have to do an additional read, but I think that's preferable to making the range set contain weird things.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +254,64 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
-                                    }
-                                    throw new RecordCoreException("This index was partly built by another method");
-                                }
-                                return null;
-                            }); },
-                false,
-                additionalLogMessageKeyValues);
+                    .thenApply( bytes -> {
+                        if (bytes == null || forceStampOverwrite) {
+                            transaction.set(stampKey, indexingTypeStamp.toByteArray());
```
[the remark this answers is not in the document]
jjezra: Do you mean the use of `indexingTypeStamp.toByteArray()`? 
(If so, please note that in a normal use, it should only happen once)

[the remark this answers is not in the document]
alecgrieser: Yeah, that's right.

That's a good point, so this is probably fine, then.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra: A question - 
Assume that, instead of a second query, we'll ask for N+1 items in cursor and provide N as an argument to the iteration function - that will set as lastResult the N+1's item after processing N items (or null, if cursor has less than N items). 
Will that work? May the cursor have less items than the requested without reaching the records store's end? I've noticed that in `byRecords` the `respectLimit` is optional - would that mean a default limit or the whole the record store?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -363,20 +367,50 @@ private static void timerIncrement(@Nullable FDBStoreTimer timer, FDBStoreTimer.
...
+                    // that was the first item, nothing to process
+                    return AsyncUtil.READY_TRUE;
+                }
+            } else {
                 // end of the cursor list
                 timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_COUNT);
-                lastResultSet.accept(result);
-                return AsyncUtil.READY_FALSE;
+                if (result.getNoNextReason() != SOURCE_EXHAUSTED) {
+                    // should we trigger a warning (or an exception) if nextResult is null - i.e. non-exhausted empty list?
```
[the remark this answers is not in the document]
alecgrieser: Oh, wait, isn't this hit in the current code, too, though? In particular, whenever we hit an _in-band_ limit, e.g., the "returned record limit". So this code is actually probably fine as is. I guess the only thing would be if wanted to pro-actively warn us if we hit an out-of-band limit. Given how the cases (in-band and out-of-band) are actually set up so that they should be handled the same, I'm tempted to say we don't warn and just leave this as is (maybe without this comment).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
alecgrieser: Discussed a little offline, but for posterity:

> Assume that, instead of a second query, we'll ask for N+1 items in cursor and provide N as an argument to the iteration function - that will set as lastResult the N+1's item after processing N items (or null, if cursor has less than N items). Will that work?

In the general case, that can cause problems, because cursors can terminate early if they hit certain "out of bands" limits, such as the time limit or the scanned bytes limit. We don't set those (currently) in the online indexer, so we should be okay to do that "for now". We may want to make sure either the code is resilient to out of band errors or throws an error if the wrong kind of limit is reached.

> I've noticed that in byRecords the respectLimit is optional - would that mean a default limit or the whole the record store?

There's no default limit, so it would scan the whole thing. The code in question is only called with `respectLimit = false` during a call to build an index in a single transaction, so it really does need to read the entire store or it will encounter correctness problems.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -253,51 +254,64 @@ private boolean shouldClearExistingIndexEntries(@Nonnull IndexState indexState,
...
-                additionalLogMessageKeyValues);
+                    .thenApply( bytes -> {
+                        if (bytes == null || forceStampOverwrite) {
+                            transaction.set(stampKey, indexingTypeStamp.toByteArray());
+                        } else {
+                            IndexBuildProto.IndexBuildIndexingStamp savedStamp ;
+                            try {
+                                savedStamp = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(bytes);
+                            } catch (InvalidProtocolBufferException ex) {
+                                savedStamp = null;
```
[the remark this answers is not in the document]
jjezra: Np. Should I add something like `"ProtoException", ex.toString()` to the new thrown exception? Should I add `bytes`? 
```
  } catch (InvalidProtocolBufferException ex) {
                                throw new RecordCoreException("invalid indexing type stamp",
                                        LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
                                        LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
                                        LogMessageKeys.INDEXER_ID, common.getUuid(),
                                        LogMessageKeys.EXPECTED, indexingTypeStamp,
                                        LogMessageKeys.ACTUAL, bytes);
  }
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -220,27 +215,36 @@ boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndex
...
+                        }
+                        final byte[] nextCont =
+                                isEmpty.get() ? null : lastResult.get().getContinuation().toBytes();
+                        if (nextCont == null) {
+                            return CompletableFuture.completedFuture(null);
+                        }
+                        executeProperties.setReturnedRowLimit(1);
+                        final ScanProperties scanProperties1 = new ScanProperties(executeProperties.build());
+                        RecordCursor<FDBIndexedRecord<Message>> nextCursor =
+                                store.scanIndexRecords(srcIndex, IndexScanType.BY_VALUE, TupleRange.between(null, rangeEnd), nextCont, scanProperties1);
```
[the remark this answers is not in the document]
jjezra: Is that the previous commit?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -220,27 +215,36 @@ boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndex
...
+                        return nextCursor.onNext().thenApply(result ->
+                                result.hasNext() ? result.get().getIndexEntry().getKey() : null );
+                    })
+                    .thenCompose(cont -> rangeSet.insertRange(store.ensureContextActive(), packOrNull(rangeStart), packOrNull(cont), true)
+                                .thenApply(ignore -> cont != null));
```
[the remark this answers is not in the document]
jjezra: Mmm.. I'm not sure I understand this one. 
`cont` set `null` signify that 4 lines earlier `hasMore.get()` was false. Shouldn't `hasMore` be the single indication - and determined by the iteration function only - of a source exhaustion?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -207,34 +205,16 @@ boolean matchingIndexingTypeStamp(@Nonnull final IndexBuildProto.IndexBuildIndex
...
-                        final ScanProperties scanProperties1 = new ScanProperties(executeProperties.build());
-                        RecordCursor<FDBIndexedRecord<Message>> nextCursor =
-                                store.scanIndexRecords(srcIndex, IndexScanType.BY_VALUE, TupleRange.between(null, rangeEnd), nextCont, scanProperties1);
-                        return nextCursor.onNext().thenApply(result ->
-                                result.hasNext() ? result.get().getIndexEntry().getKey() : null );
-                    })
+                    hasMore, recordsScanned)
+                    .thenApply(vignore -> hasMore.get() ?
+                                          lastResult.get().get().getIndexEntry().getKey() :
+                                          null)
```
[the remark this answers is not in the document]
jjezra: Ok. Got it now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -349,11 +353,11 @@ private static void timerIncrement(@Nullable FDBStoreTimer timer, FDBStoreTimer.
     }
 
     protected <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore store,
-                                                           @Nonnull RecordCursor<T> cursor,
+                                                           @Nonnull RecordCursor<T> cursor, // at least two items must be requested for this cursor
```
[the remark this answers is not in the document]
jjezra: Oh, it was suppose to mean that the cursor should be called with a minimal limit of 2 items (else it'll be an infinite loop). 
The cursor itself, though, may have any number of items (including 0 and 1). 
I'll replace this comment with one that clarify how this function returns the last cursor's item for continuation. This will be more helpful, I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra:
```
                          } catch (InvalidProtocolBufferException ex) {
                                RecordCoreException protoEx = new RecordCoreException("invalid indexing type stamp",
                                        LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
                                        LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
                                        LogMessageKeys.INDEXER_ID, common.getUuid(),
                                        LogMessageKeys.EXPECTED, indexingTypeStamp,
                                        LogMessageKeys.ACTUAL, bytes);
                                protoEx.initCause(ex);
                                throw protoEx;
 ```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Removed the comment...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
alecgrieser: Yeah, I think something like that latter of the two code snippets (including the `InvalidProtocolBufferException` as the returned exception's cause).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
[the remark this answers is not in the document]
alecgrieser: Oh, I see; this is removed from the latest commit, yeah. I reviewed this commit-by-commit, because it's easier to reason about deltas.

alecgrieser (approved):
Looks good! We should probably plan to squash this on merge into the branch

nblintao (approved):
After this PR is merged, I think the next step is try to merge this branch to main.

PR #1119, opened 2021-02-04
Resolve 1118: Ability to add closable to CursorIterator

jleach4:
Need this functionality for future work coming in follow on PR's.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
@@ -22,8 +22,10 @@ package com.apple.foundationdb.record.cursors;
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.RecordCursorResult;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
+import java.io.Closeable;
 import java.util.Iterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
@@ -37,8 +39,23 @@ import java.util.concurrent.Executor;
  */
 @API(API.Status.MAINTAINED)
 public class IteratorCursor<T> extends IteratorCursorBase<T, Iterator<T>> {
+    private static final Logger LOG = LoggerFactory.getLogger(IteratorCursor.class);
+    private Closeable closeable;
+
     public IteratorCursor(@Nonnull Executor executor, @Nonnull Iterator<T> iterator) {
         super(executor, iterator);
+    }
+
+    /**
+     * Adds an additional closeable that fires prior to the super.close().  Errors are logged but <b>are</b> swallowed.
+     *
+     * @param executor executor
+     * @param iterator iterator
+     * @param closeable closeable
+     */
+    public IteratorCursor(@Nonnull Executor executor, @Nonnull Iterator<T> iterator, Closeable closeable) {
+        super(executor, iterator);
+        this.closeable = closeable;
     }
 
     @Nonnull
@@ -47,4 +64,15 @@ public class IteratorCursor<T> extends IteratorCursorBase<T, Iterator<T>> {
         return CompletableFuture.completedFuture(computeNextResult(iterator.hasNext()));
     }
 
+    @Override
+    public void close() {
+        try {
+            if (closeable != null) {
+                closeable.close();
+            }
+        } catch (Exception ioe) {
+            LOG.warn("close failed",ioe);
+        }
+        super.close();
+    }
 }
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
```diff
@@ -37,14 +39,40 @@
  */
 @API(API.Status.MAINTAINED)
 public class IteratorCursor<T> extends IteratorCursorBase<T, Iterator<T>> {
+    private static final Logger LOG = LoggerFactory.getLogger(IteratorCursor.class);
+    private Closeable closeable;
+
     public IteratorCursor(@Nonnull Executor executor, @Nonnull Iterator<T> iterator) {
         super(executor, iterator);
```
alecgrieser, line 46:
```suggestion
        this(executor, iterator, null);
```

Just so that there's only one "real" constructor

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
```diff
@@ -37,14 +39,40 @@
  */
 @API(API.Status.MAINTAINED)
 public class IteratorCursor<T> extends IteratorCursorBase<T, Iterator<T>> {
+    private static final Logger LOG = LoggerFactory.getLogger(IteratorCursor.class);
+    private Closeable closeable;
```
alecgrieser, line 43:
```suggestion
    @Nullable
    private Closeable closeable;
```

You'll need to import the annotation as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
```diff
@@ -37,14 +39,40 @@
...
     }
 
+    @Override
+    public void close() {
+        try {
+            if (closeable != null) {
+                closeable.close();
+            }
+        } catch (Exception ioe) {
+            LOG.warn("close failed",ioe);
```
alecgrieser, line 74: This should use a `KeyValueLogMessage` to match our other log messages, though it may not make much of a difference if it's just the static message anyway

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/IteratorCursor.java
```diff
@@ -37,14 +39,40 @@
  */
 @API(API.Status.MAINTAINED)
 public class IteratorCursor<T> extends IteratorCursorBase<T, Iterator<T>> {
+    private static final Logger LOG = LoggerFactory.getLogger(IteratorCursor.class);
```
alecgrieser, line 42:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(IteratorCursor.class);
```

To match other usages in this project

alecgrieser (changes requested):
A few minor change requests. I don't see anything too wrong with this change, necessarily, though it's a little awkward that this iterator type in particular would need it. I'm not sure what exactly is going to use this in the "next PR", but I almost wonder if there's a better way to do this and if we should wait for that PR so that it's more motivated (and more people have validated that this is the best way to make that change).

ohadzeliger (approved):
Echo @alecgrieser comment: would be great to see the context for this change.

jleach4:
I am not following a bit @alecgrieser .  This existing class implements AutoClosable and I am asking for an ability to register a closable to be closed prior to the base close() without having to extend it.  So pass in an iterator and the ability to close a resource.

alecgrieser:
My question was more along the lines of "when this is in, what will it be used for and is this the best way to do that"? For example, it's a little strange that _just this cursor_ would allow an `AutoCloseable` to be added to it instead of, say, all cursors.

@sophiabuell sent me the snippet from a585cfacf2 that uses this, which answered the immediate question: https://github.com/sophiabuell/fdb-record-layer/blob/a585cfacf26a6c9134ecf78742baaa6553bb3ed9/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java#L126

Looking at how it's used in that change, I have a more conceptual problem with this approach, which is that the `IteratorCursor` doesn't really implement the contract of a typical `RecordCursor` and could lead to serious performance problems. For example, its continuations are implemented by just encoding the number of entries it needs to "skip over" to get back to where it started, which means that if any query is stopped and restarted, it will re-do all of the work from the first query. In addition, its `onNext` method is blocking (if the underlying iterator is blocking, which I have to imagine the `TextSearcher` would need to be if it's grabbing data from the database), which means that it can eat up threads in thread pools designed to only accommodate short-lived tasks.

I think if we needed to live with those limitations for the proof-of-concept version of Lucene indexes, we could _probably_ make that work, but it also seems like we need a credible plan to fix those limitations. We might be better served with something like a `LuceneIndexSearcherCursor` that wraps an `IndexSearcher` and currently acts like an `IteratorCursor` (perhaps throwing an error if you ask for a continuation) and then add proper continuation and asynchronous I/O support to that cursor.

jleach4:
@alecgrieser That helps a ton.  I was trying to limit the scope for a prototype but I see your points.  Lets close and I will address.  Thanks.

PR #1137, opened 2021-02-18
Resolve #1136: Implement Lucene IndexMaintainer and RecordCursor

### LINKED ISSUE #1136: Implement Lucene IndexMaintainer and RecordCursor with basic cursor support
Put basics in place for index update/delete and basic scans.  Add cursor support (skip, continuations, limit).

jleach4:
Supports updates/deletes with cursor support (limit, skip, continuations).  

The LuceneIndexMaintainer provides core functionality for scanning Lucene Indexes backed by FoundationDB.
The core scan interface is the existing <a href="https://lucene.apache.org/core/7_7_3/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description"> Lucene Query Parser syntax</a>.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
@@ -157,6 +157,8 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
     private boolean dirtyStoreState;
     private boolean dirtyMetaDataVersionStamp;
     private long trackOpenTimeNanos;
+    @Nonnull
+    private final Map<String, Object> session = new LinkedHashMap<>();
 
     protected FDBRecordContext(@Nonnull FDBDatabase fdb,
                                @Nonnull Transaction transaction,
@@ -1299,4 +1301,45 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
             }
         }
     }
+
+    /**
+     * Retrieve a value from the session in the FDBRecordContext.
+     *
+     * @param key key
+     * @param clazz class
+     * @param <T> Class
+     * @return value
+     */
+    @SuppressWarnings("unchecked")
+    @API(API.Status.EXPERIMENTAL)
+    public synchronized <T> T getInSession(@Nonnull String key, @Nonnull Class<T> clazz) {
+        return (T) session.get(key);
+    }
+
+    /**
+     * Put an object into the session of the FDBRecordContext.
+     *
+     * @param key key
+     * @param <T> the type of the value
+     * @param value value
+     */
+    @API(API.Status.EXPERIMENTAL)
+    public synchronized <T extends Object> void putInSessionIfAbsent(@Nonnull String key, @Nonnull T value) {
+        session.put(key, value);
+    }
+
+    /**
+     * Put an object into the session of the FDBRecordContext.
+     *
+     * @param key key
+     * @param clazz class
+     * @param <T> the type of the class
+     * @return value
+     */
+    @SuppressWarnings("unchecked")
+    @API(API.Status.EXPERIMENTAL)
+    public synchronized <T> T removeFromSession(@Nonnull String key, @Nonnull Class<T> clazz) {
+        return (T) session.remove(key);
+    }
+
 }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
@@ -1,0 +1,111 @@
+/*
+ * DirectoryCheck.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.RecordCoreStorageException;
+import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
+import com.apple.foundationdb.subspace.Subspace;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Closes the directory before commit.
+ *
+ */
+@API(API.Status.EXPERIMENTAL)
+public class DirectoryCommitCheckAsync implements FDBRecordContext.CommitCheckAsync {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LuceneIndexMaintainer.class);
+    private final FDBDirectory directory;
+
+    /**
+     * Creates a lucene directory from a subspace (keyspace) and a transaction.
+     *
+     * @param subspace the index subspace that contains the Directory/Files/etc.
+     * @param context context
+     */
+    public DirectoryCommitCheckAsync(@Nonnull Subspace subspace, @Nonnull FDBRecordContext context) {
+        this.directory = new FDBDirectory(subspace, context);
+    }
+
+    /**
+     * Close Lucene Directory.
+     *
+     * @return CompletableFuture
+     */
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> checkAsync() {
+        LOGGER.trace("closing directory check");
+        return CompletableFuture.runAsync(() -> {
+            try {
+                IOUtils.close(directory);
+            } catch (IOException ioe) {
+                LOGGER.error("DirectoryCommitCheckAsync Failed", ioe);
+                throw new RecordCoreStorageException("DirectCommitCheckAsyncFailed for directory=" + directory.getSubspace(), ioe);
+            }
+        },
+        directory.getContext().getExecutor());
+    }
+
+    @Nonnull
+    public Directory getDirectory() {
+        return directory;
+    }
+
+    /**
+     * Attempts to get the commit check from the context and if it cannot find it, creates one and adds it to the context.
+     *
+     * @param state state
+     * @return DirectoryCommitCheckAsync
+     */
+    @Nonnull
+    protected static DirectoryCommitCheckAsync getOrCreateDirectoryCommitCheckAsync(@Nonnull final IndexMaintainerState state) {
+        synchronized (state.context) {
+            DirectoryCommitCheckAsync directoryCheck = state.context.getInSession(getDirectoryName(state), DirectoryCommitCheckAsync.class);
+            if (directoryCheck == null) {
+                directoryCheck = new DirectoryCommitCheckAsync(state.indexSubspace, state.context);
+                state.context.addCommitCheck(directoryCheck);
+                state.context.putInSessionIfAbsent(getDirectoryName(state), directoryCheck);
+            }
+            return directoryCheck;
+        }
+    }
+
+    /**
+     * The directory name in the context.
+     *
+     * @param state state
+     * @return String
+     */
+    private static String getDirectoryName(@Nonnull final IndexMaintainerState state) {
+        return "directory$" + state.index.getName();
+    }
+}
+
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
@@ -1,0 +1,134 @@
+/*
+ * IndexWriterCommitCheckAsync.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.async.AsyncUtil;
+import com.apple.foundationdb.record.RecordCoreStorageException;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
+import org.apache.lucene.codecs.lucene70.Lucene70Codec;
+import org.apache.lucene.index.ConcurrentMergeScheduler;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.MergePolicy;
+import org.apache.lucene.index.TieredMergePolicy;
+import org.apache.lucene.util.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+
+import static com.apple.foundationdb.record.lucene.DirectoryCommitCheckAsync.getOrCreateDirectoryCommitCheckAsync;
+
+/**
+ * This class closes the writer after commit.  The goal here is to only flush the results
+ * at the end of the commit.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class IndexWriterCommitCheckAsync implements FDBRecordContext.CommitCheckAsync {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LuceneIndexMaintainer.class);
+    protected final IndexWriter indexWriter;
+    protected final Executor executor;
+
+
+    /**
+     * Creates an index writer config with merge configurations that limit the amount of data in a segment.
+     *
+     * @param analyzer analyzer
+     * @param directoryCommitCheckAsync directoryCommitCheckAsync
+     * @param executor executor
+     * @throws IOException exception
+     */
+    public IndexWriterCommitCheckAsync(@Nonnull Analyzer analyzer, @Nonnull DirectoryCommitCheckAsync directoryCommitCheckAsync, Executor executor) throws IOException {
+        TieredMergePolicy tieredMergePolicy = new TieredMergePolicy();
+        tieredMergePolicy.setMaxMergedSegmentMB(5.00);
+        tieredMergePolicy.setMaxMergeAtOnceExplicit(2);
+        tieredMergePolicy.setNoCFSRatio(1.00);
+        IndexWriterConfig indexWriterConfig = new IndexWriterConfig(analyzer);
+        indexWriterConfig.setUseCompoundFile(true);
+        indexWriterConfig.setMergePolicy(tieredMergePolicy);
+        indexWriterConfig.setMergeScheduler(new ConcurrentMergeScheduler() {
+            @Override
+            protected void doMerge(final IndexWriter writer, final MergePolicy.OneMerge merge) throws IOException {
+                merge.segments.forEach( (segmentCommitInfo) -> LOGGER.trace("segmentInfo={}", segmentCommitInfo.info.name));
+                super.doMerge(writer, merge);
+            }
+        });
+        indexWriterConfig.setCodec(new Lucene70Codec(Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
+        this.indexWriter = new IndexWriter(directoryCommitCheckAsync.getDirectory(), indexWriterConfig);
+        this.executor = executor;
+    }
+
+    /**
+     * Close IndexWriter.
+     *
+     * @return CompletableFuture
+     */
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> checkAsync() {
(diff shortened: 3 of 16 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexTypes.java
```diff
@@ -142,6 +142,11 @@
      */
     public static final String TEXT = "text";
 
+    /**
+     * An index on the tokens in a text field.
+     */
+    public static final String LUCENE = "lucene";
```
MMcM, line 148: Index types not defined in the core usually don't get named there, but rather in the .jar that adds them.

For instance, `fdb-record-layer-spatial/src/main/java/com/apple/foundationdb/record/spatial/geophile/GeophileIndexTypes`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -696,14 +698,24 @@ default boolean isReady() {
...
+        addCommitCheck(null, check);
+    }
+
+    public synchronized void addCommitCheck(@Nullable String name, @Nonnull CommitCheckAsync check) {
         while (!commitChecks.isEmpty()) {
             if (commitChecks.peek().isReady()) {
                 asyncToSync(FDBStoreTimer.Waits.WAIT_ERROR_CHECK, commitChecks.remove().checkAsync());
+                if (name != null) {
+                    namedCommitChecks.remove(name);
+                }
```
MMcM, line 710: I don't think this is the intended semantics. It look to me like, if any pending check is ready, check it and allow the named check to supersede what was there before, whatever state it was in. Which means if a pending check by the given name is not yet ready, the new check with that name is ignored.

Questions about the desired behavior:
* If a check by the given name is ready, is it checked or is that result ignored? Assuming the check passes, is the new check remembered as well?
* If a check by the given name is pending, does the new check replace it?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Close Lucene Directory while swallowing errors.
+     *
+     * @return CompletableFuture
+     */
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> checkAsync() {
+        LOGGER.trace("closing directory check");
+        return CompletableFuture.runAsync(() -> IOUtils.closeWhileHandlingException(directory));
```
MMcM, line 65: Ideally, this would get an `Executor` from someplace. Could the check be instantiated by an `FDBRecordContext` instead?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Attempts to get the commit check from the context and if it cannot find it, creates one and adds it to the context.
+     *
+     * @param state state
+     * @return DirectoryCommitCheckAsync
+     */
+    @Nonnull
+    protected static DirectoryCommitCheckAsync getOrCreateDirectoryCommitCheckAsync(@Nonnull final IndexMaintainerState state) {
+        synchronized (state) {
```
MMcM, line 81: Probably safer to synchronize on the context, rather than the state, since that's where the named checks are synchronized.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
```diff
@@ -0,0 +1,112 @@
...
+import static com.apple.foundationdb.record.lucene.DirectoryCommitCheckAsync.getOrCreateDirectoryCommitCheckAsync;
+
+/**
+ * This class closes the writer after commit.  The goal here is to only flush the results
+ * at the end of the commit.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class IndexWriterCommitCheckAsync implements FDBRecordContext.CommitCheckAsync {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LuceneIndexMaintainer.class);
+    protected final IndexWriter indexWriter;
```
MMcM, line 53: I kind of wish there were a better place for this to live. Like if a context had a scratchpad, it could do there and then separately a commit hook could close it down. On the other hand, pulling it out of the registered commit hook should work, so I guess it's okay.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Index maintainer for Lucene Indexes backed by FDB.  The insert, update, and delete functionality
+ * coupled with the scan functionality is implemented here.
+ *
+ */
+@API(API.Status.EXPERIMENTAL)
+public class LuceneIndexMaintainer extends StandardIndexMaintainer {
+    private static final Logger LOG = LoggerFactory.getLogger(LuceneIndexMaintainer.class);
+    private final Analyzer analyzer;
+    protected static final String PRIMARY_KEY_FIELD_NAME = "p"; // Need to find reserved names..
```
MMcM, line 80: This is a `TODO`, right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param continuation any continuation from a previous scan invocation
+     * @param scanProperties skip, limit and other properties of the scan
+     * @return RecordCursor
+     */
+    @Nonnull
+    @Override
+    public RecordCursor<IndexEntry> scan(@Nonnull IndexScanType scanType, @Nonnull TupleRange range,
+                                         @Nullable byte[] continuation,
+                                         @Nonnull ScanProperties scanProperties) {
+        LOG.trace("scan scanType={}", scanType);
```
MMcM, line 109: Should check for `BY_LUCENE_QUERY` and throw if not, I think.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,244 @@
...
+        Verify.verify(range.getLow() != null);
+        try {
+            final MultiFieldQueryParser parser = new MultiFieldQueryParser(fieldNames.toArray(new String[0]) , analyzer);
+            Query query = parser.parse(range.getLow().getString(0));
```
MMcM, line 113: Not for now and really a core issue, but seems like we're missing an abstraction in the index API. Like maybe there are index-supported table functions like there are record functions and aggregate functions. And you use one of those with a query argument rather than a scan.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,244 @@
...
+        if (indexEntries.isEmpty()) {
+            return AsyncUtil.DONE;
+        }
+        try {
+            final IndexWriter writer = getOrCreateIndexWriter(state, analyzer);
+            indexEntries.forEach( (entry ) -> {
+                try {
+                    if (remove) {
+                        Query query = SortedDocValuesField.newSlowExactQuery(PRIMARY_KEY_SEARCH_NAME, new BytesRef(savedRecord.getPrimaryKey().pack()));
+                        writer.deleteDocuments(query);
```
MMcM, line 143: Are these potentially blocking calls, too? Or do they just queue work and I/O happens at commit time? Mainly for the record; we don't need to fix absolutely everything all at once.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,244 @@
...
+                                                              @Nonnull TupleRange range,
+                                                              @Nonnull IsolationLevel isolationLevel) {
+        LOG.warn("evaluateAggregateFunction() function={}", function);
+        return unsupportedAggregateFunction(function);
+    }
+
+    @Override
+    public boolean isIdempotent() {
+        LOG.trace("isIdempotent()");
+        return false;
```
MMcM, line 212: Why not? Doesn't saving the same record twice overwrite the document?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -0,0 +1,245 @@
...
+    private final Analyzer analyzer;
+    protected static final String PRIMARY_KEY_FIELD_NAME = "p"; // Need to find reserved names..
+    private static final String PRIMARY_KEY_SEARCH_NAME = "s"; // Need to find reserved names..
+    private final List<String> fieldNames;
+    private final Executor executor;
+
+    public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull Executor executor, @Nonnull Analyzer analyzer) {
+        super(state);
+        this.analyzer = analyzer;
+        this.fieldNames = new ArrayList<>(2);
```
sophiabuell, line 88: nitpick: this should be defined above if its not reliant on arguments.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: This seems to queue work as the change isn't visible until commit is called. Though according to some javadocs deeper in Lucene if the deletes are using too much heap they'll be written to disk. So... both?

scgray:
I'm about 3/4 done reviewing and should be complete later today (so far mostly ignorable nit-picks).

scgray (changes requested):
Mostly my comments were optional.

alecgrieser:
Sorry had to get to a few other things. Will try and take a look tomorrow and/or Wednesday.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
sophiabuell: In the current situation a new check with the same name would 
- if the check is ready it would be checked and returned then the new check is remembered
- If it is not ready the check is not replaced and the old one is remembered. 
I believe the solve here if we WANT to replace the old check would be to create a specific replaceCommitCheck function.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: yes it does overwrite the document.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
jleach4: @MMcM Added a session concept for lookup...  Hope that clarifies.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
MMcM: But mostly happens without I/O, so maybe okay for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
MMcM: The new `session` stuff is what I meant by "scratchpad."

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
MMcM: Not for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
```diff
@@ -0,0 +1,118 @@
...
+        this.indexWriter = new IndexWriter(directoryCommitCheckAsync.getDirectory(), indexWriterConfig);
+        this.executor = executor;
+    }
+
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> checkAsync() {
+        LOGGER.trace("closing writer check");
+        return indexWriter.isOpen() ?
+               CompletableFuture.runAsync(() -> IOUtils.closeWhileHandlingException(indexWriter), executor) :
```
MMcM, line 91: Does `IndexWriter.close` need to happen on transaction rollback (that is, `close` without `commit`)?

That was the idea behind having the session objects be something narrower than `Object` so they could have a lifecycle. Maybe the Java GC does a good enough job?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
jleach4: It should be collected by GC.  

`    // Ensure that only one thread actually gets to do the
    // closing
    if (shouldClose(true)) {
      try {
        if (infoStream.isEnabled("IW")) {
          infoStream.message("IW", "now flush at close");
        }

        flush(true, true);
        waitForMerges();
        commitInternal(config.getMergePolicy());
        rollbackInternal(); // ie close, since we just committed
      } catch (Throwable t) {
        // Be certain to close the index on any exception
        try {
          rollbackInternal();
        } catch (Throwable t1) {
          t.addSuppressed(t1);
        }
        throw t;
      }
    }`

MMcM (approved):

scgray (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @Override
+    public boolean accept(@Nonnull RecordCursorVisitor visitor) {
+        visitor.visitEnter(this);
+        return visitor.visitLeave(this);
+    }
+
+    private synchronized IndexReader getIndexReader() throws IOException {
+        IndexWriterCommitCheckAsync writerCheck = getIndexWriterCommitCheckAsync(state);
+        return writerCheck == null ? DirectoryReader.open(getOrCreateDirectoryCommitCheckAsync(state).getDirectory()) : DirectoryReader.open(writerCheck.indexWriter);
```
ScottDugas, line 229: I'm assuming this is to deal with this note from the [lucene javadoc](https://lucene.apache.org/core/7_7_2/core/org/apache/lucene/store/Directory.html)
> A file open for writing may not be available for read access until the corresponding IndexOutput is closed.

That being said, will this work correctly if the writer is created after starting the cursor. Eg you wanted to do something like an update, where you pipe the result into code that updates records (e.g. I want to update a typo in all the docs)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -0,0 +1,359 @@
...
+import static com.apple.foundationdb.record.lucene.LuceneIndexScanTypes.BY_LUCENE_QUERY;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concatenateFields;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.COMPLEX_DOC;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.SIMPLE_DOC;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for {@code FULL_TEXT} type indexes.
```
ScottDugas, line 57: There is another index for full text queries, I don't think there would actually be any confusion based on the name, but probably still better. 
```suggestion
 * Tests for {@code LUCENE} type indexes.
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -0,0 +1,359 @@
...
+    @Test
+    public void simpleInsertAndSearch() {
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, metaDataBuilder -> {
+                metaDataBuilder.removeIndex(TextIndexTestUtils.SIMPLE_DEFAULT_NAME);
+                metaDataBuilder.addIndex(SIMPLE_DOC, SIMPLE_TEXT_SUFFIXES);
+            });
+            recordStore.saveRecord(createSimpleDocument(1623L, DYLAN, 2));
+            recordStore.saveRecord(createSimpleDocument(1547L, WAYLON, 1));
+            RecordCursor<IndexEntry> indexEntries = recordStore.scanIndex(SIMPLE_TEXT_SUFFIXES, BY_LUCENE_QUERY, TupleRange.allOf(Tuple.from("idiot")), null, ScanProperties.FORWARD_SCAN);
```
ScottDugas, line 119: It looks like `FORWARD_SCAN` is another thing ignored by the actual index maintainer / scan logic.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -0,0 +1,359 @@
...
+                metaDataBuilder.removeIndex(TextIndexTestUtils.SIMPLE_DEFAULT_NAME);
+                metaDataBuilder.addIndex(SIMPLE_DOC, SIMPLE_TEXT_SUFFIXES);
+            });
+            recordStore.saveRecord(createSimpleDocument(1623L, DYLAN, 2));
+            recordStore.saveRecord(createSimpleDocument(1624L, DYLAN, 2));
+            recordStore.saveRecord(createSimpleDocument(1625L, DYLAN, 2));
+            recordStore.saveRecord(createSimpleDocument(1626L, DYLAN, 2));
+            recordStore.saveRecord(createSimpleDocument(1547L, WAYLON, 1));
+            assertEquals(2, recordStore.scanIndex(SIMPLE_TEXT_SUFFIXES, BY_LUCENE_QUERY,
+                    TupleRange.allOf(Tuple.from("idiot")), Ints.toByteArray(2), ScanProperties.FORWARD_SCAN)
```
ScottDugas, line 140: This manually sets the continuation to 2, I think it would be better to have it query look at 2 records, and then query again with the continuation, and make sure that you see all the appropriate entries.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -0,0 +1,359 @@
...
+                .setDocId(docId)
+                .setText(text)
+                .setText2(text2)
+                .setGroup(group)
+                .build();
+    }
+
+
+    @Test
+    public void simpleInsertAndSearch() {
```
ScottDugas, line 111: These seem like a good start to the tests, but there are a few things I can think of to add (either now or later), mostly inspired by the bugs I think I saw in the code above:
1. Multiple indexes on the same store
2. Multiple stores with the same index in the same context
3. Complex queries that match a document on both fields, and use an OR
4. calling `scanIndexRecords` and getting the actual document back
5. saving some records, committing, querying and mapping that into an update
6. Adding to at least some of these tests assertions about the resulting index entries.
7. Updating the metadata to cause the index subpsace to change mid-transaction


(Note: I would not be surprised if that list of tests goes noticeably beyond the testing of many of the existing cursors/indexes, but in that case it's an issue there, not here.

ScottDugas (changes requested):
There's quite a few things in this review, I think almost all of them could be pushed out into a followup issue, since this is experimental, except the session on the RecordContext, although if those are marked EXPERIMENTAL too, I think I'll leave it up to you whether to address the other comments now or put them in an issue. I'm not sure if we currently have a centralized issue tracking these things yet, but it would be good.

fdb-record-layer-lucene/docs/index.md
```diff
@@ -13,3 +13,5 @@ The metadata is keyed by filename and in the value it provides the total size, t
 ### Data Space 
 #### {Directory Bytes}{indexspace}{2}[{key=(1,1), value=byte[]},{key=(1,2), value=byte[]},{key=(2,1), value=byte[]}]
 The data section is keyed via the files id in the metadata directory coupled with a block number.
+
+## Query Functionality
```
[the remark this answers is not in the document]
jleach4: Fixing...

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
jleach4: Fixing

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+@API(API.Status.EXPERIMENTAL)
+public class DirectoryCommitCheckAsync implements FDBRecordContext.CommitCheckAsync {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LuceneIndexMaintainer.class);
+    private final FDBDirectory directory;
+
+    /**
+     * Creates a lucene directory from a subspace (keyspace) and a transaction.
+     *
+     * @param subspace subspace
```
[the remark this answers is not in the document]
jleach4: Fixing

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param context context
+     */
+    public DirectoryCommitCheckAsync(@Nonnull Subspace subspace, @Nonnull FDBRecordContext context) {
+        this.directory = new FDBDirectory(subspace, context);
+    }
+
+    /**
+     * Close Lucene Directory while swallowing errors.
```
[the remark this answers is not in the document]
jleach4: Fixing...

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/DirectoryCommitCheckAsync.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Closes the directory after commit.
```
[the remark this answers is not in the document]
jleach4: Fixing...

ScottDugas (approved):
Seems acceptable for an experimental API, although there are definitely some rough edges, known bugs, and missing test coverage.

sophiabuell (approved):

PR #1079, opened 2020-12-01
Resolve #1078: Allow creating an index from an existing index

jjezra:
When building a single type idempotent index, provide the user with an option
    to request an iteration over an existing, readable, source index instead of a
    full records store scan.

Issue #1007, opened 2020-08-04
OnlineIndexer does not backoff properly when it retries


nblintao:
The delay is converging to 0 instead.

nblintao:
Now I'm looking at `FDBDatabaseRunnerImpl`. It has the same problem as well. I will just put the change in https://github.com/FoundationDB/fdb-record-layer/pull/813.

nblintao:
@alecgrieser I was wondering if it was noticeable that `DatabaseRunner` had never backed off properly? (If I do math correctly) I think the current behavior is retrying every 10 ms, rather than backing off to at maximum 1000 ms.

alecgrieser:
That's definitely possible. I'm not sure we have data confirming or denying that.

nblintao:
@alecgrieser I see. I asked because I was wondering if we'll be lucky to see performance improvement (by reducing conflicts) after adopting the fix, or we would see regressions on at least some types of traffic because it was relying on that behavior?

alecgrieser:
I guess it depends. I think you'd probably see both--some things get better throughput because of more correctly spreading around things to reduce conflicts, and some things will see higher latency because of more of a delay. Which is probably fine--no such thing as free lunch.

PR #1145, opened 2021-02-24
Indexer index from index

jjezra:
* Refactoring OnlineIndexer - splitting it to API and submodules. 
* Adding a new indexing type - Indexing by Index, which allows scanning a source index (as a relevant subset of the records store) for indexing a new target index.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -0,0 +1,519 @@
...
+
+    public int getLimit() {
+        // made public to support tests
+        return throttle.getLimit();
+    }
+
+    protected CompletableFuture<Boolean> throttleDelay() {
+        int limit = getLimit();
+        int recordsPerSecond = common.config.getRecordsPerSecond();
+        int toWait = (recordsPerSecond == IndexingCommon.UNLIMITED) ? 0 : 1000 * limit / recordsPerSecond;
```
jleach4, line 343: Jitter?

alecgrieser: This delay is there to try and control the rate at which records are being indexed to avoid overwhelming the cluster with writes; it is not a delay added as part of an internal retry. The delay added when there's a retry does have jitter (see: https://github.com/FoundationDB/fdb-record-layer/blob/249a83a0b1664234290da3142292c54401803eaf/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java#L222), though it could probably be better.

nblintao: This part is throttling for scanning speed, so I don't think jittering will be necessary. We are doing jitter when we back off on failures, see https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java#L433

nblintao: Actually, if you are wondering about because of the backoff problem I mentioned this morning, it's at https://github.com/FoundationDB/fdb-record-layer/issues/1007

jjezra: Sorry, I'm not sure I understand the question. Can you clarify?

alecgrieser (changes requested):
IIRC, there were some follow up items from the first two PRs, mostly around adding additional tests and then some additional refactoring. I don't think those all need to necessarily be done before we merge this feature branch into `main`, but do we have issues for all of those items so we can keep track of them?

jjezra:
@alecgrieser , 
IIRC, the missing tests are: 
* Multi-type indexes
* There are concurrent builds
* The meta-data change midway through a build (might be somewhat hard to simulate)

And sure, I will create an issue to add them later. Note that we'll also have to add support for ranges & concurrent builds by source index. It's an easy task as the `byIndex` code was designed for it, but I believe it'll be safer to use a new API function (`buiildRangeByIndex`?) to distinguish it from `buildRange` - as each is expecting a very different kind of range.

nblintao:
For the record, this PR is trying to merge two proved PRs: https://github.com/FoundationDB/fdb-record-layer/pull/1079 and https://github.com/FoundationDB/fdb-record-layer/pull/1117

jjezra:
Added two issues: 
https://github.com/FoundationDB/fdb-record-layer/issues/1146 - support concurrent indexing by index.
https://github.com/FoundationDB/fdb-record-layer/issues/1147 - add more tests to indexing by index.

alecgrieser (approved):

nblintao (approved):

PR #1149, opened 2021-02-24
resolve #1148

### LINKED ISSUE #1148: Cascades cost model is unstable
When producing alternatative plans that have same cost, planner choses a plan in an unstable fashion: Subsequent plans may chose a different one out of the set of plans with same cost.
The ask is that plan selection is stable such that we can assert the plan for a given query.

ohadzeliger:
Resolve #1148

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.query.plan.temp;
 
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.PlanHashable;
 import com.apple.foundationdb.record.query.plan.QueryPlanner.IndexScanPreference;
 import com.apple.foundationdb.record.query.plan.RecordQueryPlannerConfiguration;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
```diff
@@ -119,8 +120,22 @@ public int compare(@Nonnull RelationalExpression a, @Nonnull RelationalExpressio
...
+
+        int ufpA = UnmatchedFieldsProperty.evaluate(planContext, a);
+        int ufpB = UnmatchedFieldsProperty.evaluate(planContext, b);
+        if (ufpA != ufpB) {
+            return Integer.compare(ufpA, ufpB);
+        }
+
+        // If plans are indistinguishable from a cost perspective, select one by planHash. This would make the cost model stable
+        // (select the same plan on subsequent plannings).
+        if ((a instanceof PlanHashable) && (b instanceof PlanHashable)) {
```
normen662, line 132: is there any chance this is not a `PlanHashable`

ohadzeliger: In my tests I have not seen it be called with non-PlanHashable items but for defensive coding decided it may be better to check.

normen662 (approved):

PR #1151, opened 2021-02-24
Fixes #1150: @DualPlannerTest does not distinguish tests

### LINKED ISSUE #1150: @DualPlannerTest does not distinguish tests
They are all `[1]` or `[2]`.

MMcM:
This still doesn't work to make navigating to the test in the IDE work. I don't know whether slightly different formatting could accomplish that.


ohadzeliger:
Looks like the test report from `Gradle` contains `()` after the test name (See below) - maybe this could help with the IDE navigation?

Screen Shot 2021-03-10 at 10 01 21 AM

PR #1157, opened 2021-02-27
Resolves #1156: Improve infrastructure for FDBRecordStorePerformanceTest

### LINKED ISSUE #1156: Improve infrastructure for FDBRecordStorePerformanceTest
For instance, JMH.

MMcM:


ohadzeliger (commented):
This looks like a really useful framework to use. My questions are really about the "meta" usage rather than the particulars of the PR:
- Are there plans to calculate a baseline- are there environment/hardware specs that remain relatively constant such that we can compare baseline benchmarks to new results?
- It would be really useful to have IDE integration to these (website says they are "hard").
- Eventually some CI integration would be necessary - What are the considerations we need to take into account?

PR #1162, opened 2021-03-04
Remove deprecated RecordCursor apis

ScottDugas:
I am explicitly not optimizing any implementations now that they
don't need to implement the old api, just removing the api, and
fields that became unused.


ScottDugas:
Closing in favor of #1164

PR #1165, opened 2021-03-06
Lucene plan

sophiabuell:
- Nested fields for future implementation 
- future fix: actually pass through type for sort fields to what is stored in index


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -101,7 +101,7 @@ public static AvailableFields fromIndex(@Nonnull RecordType recordType,
 
         final List<KeyExpression> keyFields = new ArrayList<>();
         final List<KeyExpression> valueFields = new ArrayList<>();
-        if (indexTypes.getTextTypes().contains(index.getType())) {
+        if (indexTypes.getTextTypes().contains(index.getType()) || indexTypes.getLuceneTypes().contains(index.getType())) {
             // Full text index entries have all of their fields except the tokenized one.
```
MMcM, line 105: I'm not sure that's true. I think a Lucene text entry only has the `StoredField`'s, which is only the primary key. All the declared fields turn into indexed, but not stored, fields of the Lucene index. Unlike the existing text index, where the fields other than the tokenized ones are grouping fields.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1219,6 +1229,51 @@ private ScoredPlan planRankWithAnd(@Nonnull CandidateScan candidateScan,
...
+        return null;
+    }
+
+
+    @Nullable
+    private ScoredPlan planLucene(@Nonnull CandidateScan candidateScan,
+                                @Nonnull Index index, @Nonnull QueryComponent filter,
+                                @Nullable KeyExpression sort) {
+        if (sort != null) {
+            // TODO: Full Text: Sorts are not supported with full text queries (https://github.com/FoundationDB/fdb-record-layer/issues/55)
```
MMcM, line 1262: This isn't fundamental, and moreover is unrelated to that issue, right? That is, there is no reason I can think of that a sort expression couldn't be turned into a `Sort` and used in `LuceneRecordCursor.performScan` as an additional argument to `search`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1219,6 +1229,51 @@ private ScoredPlan planRankWithAnd(@Nonnull CandidateScan candidateScan,
...
+            return null;
+        }
+
+        FilterSatisfiedMask filterMask = FilterSatisfiedMask.of(filter);
+        final ScanComparisons scans = getComparisonsForLuceneFilter(filter, filterMask);
+        if (scans == null) {
+            return null;
+        }
+        RecordQueryPlan plan;
+        plan = new RecordQueryIndexPlan(index.getName(), IndexScanType.BY_LUCENE, scans, false);
```
MMcM, line 1272: Is `scans` really in a form that `BY_LUCENE` can digest? Specifically when there is more than one.

Would it be better to assemble various `LuceneQueryComponent`s into a single Lucene query using string operations?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    public void delayFetchOnIntersectionOfLuceneScanWithFilters(boolean shouldDeferFetch) throws Exception {
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context);
+            final QueryComponent filter1 = new LuceneQueryComponent("the continuance");
+            final QueryComponent filter2 = new LuceneQueryComponent("grudge");
```
MMcM, line 262: There is only one sample that has either of these and it has both, right?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -0,0 +1,309 @@
...
+            // Query for full records
+            RecordQuery query = RecordQuery.newBuilder()
+                    .setRecordType(TextIndexTestUtils.SIMPLE_DOC)
+                    .setFilter(Query.and(filter1, filter2))
+                    .build();
+            setDeferFetchAfterUnionAndIntersection(shouldDeferFetch);
+            RecordQueryPlan plan = planner.plan(query);
+            Matcher<RecordQueryPlan> matcher = filter(filter2, indexScan(allOf(indexScanType(IndexScanType.BY_LUCENE),
+                    indexName("Complex$text_index"),
+                    bounds(hasTupleString("[[the continuance],[the continuance]]")))));
```
MMcM, line 272: We've lost `grudge`, haven't we?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -1680,4 +1683,71 @@ public String toString() {
...
+        private final String query;
+
+        public LuceneComparison(@Nonnull String query) {
+            this.query = query;
+        }
+
+        @Nullable
+        @Override
+        public Boolean eval(@Nonnull final FDBRecordStoreBase<?> store, @Nonnull final EvaluationContext context, @Nullable final Object value) {
+            return true;
```
MMcM, line 1703: Is this ever going to be possible? Or should this just throw an exception?

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
sophiabuell: grudge is included in the filter2 LuceneQueryComponent.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
sophiabuell: Yes, I can add one that has one but not the other. Thats a good point thanks!

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: Right. But aside from the temporary state where that returns `true` for everything, is it in fact the future intention to be able to match a query string against a text string at runtime? I was assuming not so that all Lucene queries needed to go into the `search` argument.

sophiabuell: What do you mean by the search argument? I may be missing something about the use case. I assumed we wanted to be able to search distinct strings that appeared separately in the document.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
sophiabuell: So I wasn't entirely sure of this. Both paths worked so I took a guess. Looking back through the code with a bit of a break this makes sense that the other path would be the one to take given that I don't think grouping fields are a part of this index. thanks.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: Yes, I agree that's what we want. Isn't that what `"the continuance" AND grudge` would do? That's why I thought maybe combining separate queries with simple string operations would work.

sophiabuell: So follow up thought. Do we want to support both functionality or should we cause an issue if we have two Lucene queries in the same request?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: I think what this special sub-planner does is take a Boolean tree of:
* Lucene components with a field name (field has to be one of the ones in the Lucene index, either in the stored primary key fields or the unstored index only fields)
* Lucene query fragments without their own field name
* Regular query comparisons, which might or might not involve a field present in the index

From these it produces a _single_ Lucene query string, that drives the scan (Lucene `search`) and an optional residual query component that is checked at runtime against records returned by the Lucene cursor, that is, turned into a regular filter plan with Lucene plan inside.

sophiabuell: So in a previous test where I tested that precisely yes it actually worked. However the name scans is a bit misleading as to how the code actually works here. It works by returning one of the comparisons in the and component, similar to the textPlanner and then the record layer handles the unsatisfied filters in other logic. To your point about combining the queries into a single Lucene query, yeah, that makes more sense as to how to handle this case.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
sophiabuell: If we transform it in the record layer then I guess it doesn't matter. I'll write extra testing for both cases.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: Yes, except that the leftover was a Lucene query component. I am still unsure whether those will work at runtime or it is a requirement to have them all folded into the scan's search string.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -545,6 +546,15 @@ private ScoredPlan planIndex(@Nonnull PlanContext planContext, @Nonnull QueryCom
                     p = computeIndexFilters(planContext, p);
                 }
                 return p;
+            } else if (indexTypes.getLuceneTypes().contains(index.getType())) {
+                p = planLucene(candidateScan, index, filter, sort);
```
normen662, line 550: Is this plan (and the index it uses) correctly communicating `createsDuplicates()`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
sophiabuell: I'm not sure I'm understanding your concern. Both when I put the Scan Comparisons as separate scan comparisons into the same plan and when I leave the second Lucene filter as a unsatisfied filter the plan works and returns the correct documents when executed. The question though that we're trying to answer here I think is which is better and I think that is the third option that you mentioned above of combining the scans into the Lucene search string.

MMcM: Is there something that can take a Lucene query component and a record and tell whether the record matches?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
sophiabuell: Since we're assuming a flat index I'm fairly certain this should be being correctly used from what we discussed before right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
sophiabuell: I think it was decided that since it works both ways implementing the flattening of multiple Lucene components into a complex query string should be postponed. We would also want to allow the planner to compare that scan with a scan that doesn't flatten them.

MMcM: I am okay with postponing, since the caller can do the flattening. But I still don't see how leftovers are evaluated for arbitrary Lucene expressions and not just simple comparisons.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -234,4 +238,8 @@ private void performScan() throws IOException {
         searcher = new IndexSearcher(indexReader);
         topDocs = searcher.search(query, limitRemaining);
```
MMcM, line 239: Doesn't `sort` go in here as the third argument? (Maybe only if non-null?)

MMcM (changes requested):
I think it would be better if everything related to Lucene were pulled into the lucena subsystem. That means replacing `planLucene` with `planOtherIndexType` or something like that and then checking type in the subclass method. Or having a map from index type to special planner extender interface.

I am okay with deferring that to next pass, though. This looks like a reasonable milestone point.

ohadzeliger (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
sophiabuell: wow.....  🤦‍♀️

sophiabuell:
test this please

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        final LuceneIndexQueryPlan that = (LuceneIndexQueryPlan)o;
+        return Objects.equals(sort, that.sort);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), sort);
+    }
+
```
MMcM, line 65: Also needs to override `planHash` to take account of `sort`, I believe.

MMcM (changes requested):

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
normen662: I think the answer to my previous question is _no_. Can we have a bunch of follow-up issues defined for this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -545,6 +545,15 @@ private ScoredPlan planIndex(@Nonnull PlanContext planContext, @Nonnull QueryCom
...
+                if (p != null) {
+                    p = planRemoveDuplicates(planContext, p);
+                }
+                if (p != null) {
+                    p = computeIndexFilters(planContext, p);
```
normen662, line 554: Hmm -- I think this can only ever find an index filter for a predicate that is expressed on a part of the primary key of the type. Can you add a test case to check that it's working as expected, i.e.

filter: `x = 5` and `luceneComponent` and `x` is part of the primary key of the type
you should see that the index filters being computed contain `x=5`

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
sophiabuell: Handled in line 239 of FDBLuceneQueryTest (https://github.com/FoundationDB/fdb-record-layer/pull/1165/files#diff-9c5e48f9f4371e68d15f5f0dd1755b60c23651214984a6a0ea5508ba14cc8352R239)

normen662 (approved):

PR #1168, opened 2021-03-10
Resolves #1138: Planner's key from index should include primary keys

### LINKED ISSUE #5: Queries with order-preserving filter sorting by primary key will not use index for filter
Suppose we have a schema like:
```
message MyRecord {
   optional int64 id = 1 [(field).primary_key = true];
   optional int64 value = 2 [(field).index = {}];
}
```
Then suppose someone has a query like:
```
RecordQuery query = RecordQuery.newBuilder()
    .setRecordType("MyRecord")
    .setFilter(Query.field("value").equalsValue(5L))
    .setSort(Key.Expressions.field("id"))
    .build();
```
Without the sort, the returned plan will be something like `Index(MyRecord$value [[5],[5]])`, which is correct. However, *with* the sort, this will do `Scan(<,>) | [MyRecord] | value EQUALS 5` (i.e., it does a full store scan instead of using the index). But because the sort is by primary key and the filter is an equality comparison, the results of the index scan will be returned in the correct order, so it would be better to return the index scan instead (just like if there was no sort).

Note that if the filter had been something like `Query.field("value").greaterThan(5)`, then this optimization cannot be made.

A slightly more sophisticated example of this can occur if there are multiple filters. For example:
```
message MyRecord {
   optional int64 id = 1 [(field).primary_key = true];
   optional int64 value = 2 [(field).index = {}];
   optional int64 unindexed_value = 3;
}
```
Then the query:
```
RecordQuery query = RecordQuery.newBuilder()
    .setRecordType("MyRecord")
    .setFilter(Query.and(Query.field("value").equalsValue(5L), Query.field("unindexed_value").equalsValue(6L))
    .setSort(Key.Expressions.field("id"))
    .build();
```
The query plan should use the index on value then manually filter out the `unindexed_value` results.

### LINKED ISSUE #1138: Planner's key from index should include primary keys
They are there in the index entry and so available for sorting, etc.

MMcM:
Fixes #5: Queries with order-preserving filter sorting by primary key will not use index for filter

It's possible that we should be more concerned about unanticipated effects of this change and should have some kind of control switch in the planner options to enable it. (It's never incorrect, but might expose previously hidden bugs.)

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -321,7 +321,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             if (sort == null) {
                 p = planNoFilterNoSort(planContext, index);
             } else {
-                p = planSortOnly(new CandidateScan(planContext, index, sortReverse), index.getRootExpression(), sort);
+                p = planSortOnly(new CandidateScan(planContext, index, sortReverse), indexKeyExpressionForPlan(planContext.commonPrimaryKey, index), sort);
             }
             if (p != null) {
                 if (bestPlan == null || p.score > bestPlan.score ||
@@ -452,11 +452,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             bestPlan = planIndex(planContext, filter, null, planContext.commonPrimaryKey, intersectionCandidates);
         }
         for (Index index : planContext.indexes) {
-            KeyExpression indexKeyExpression = index.getRootExpression();
-            if (indexKeyExpression instanceof KeyWithValueExpression) {
-                indexKeyExpression = ((KeyWithValueExpression) indexKeyExpression).getKeyExpression();
-            }
-
+            KeyExpression indexKeyExpression = indexKeyExpressionForPlan(planContext.commonPrimaryKey, index);
             ScoredPlan p = planIndex(planContext, filter, index, indexKeyExpression, intersectionCandidates);
             if (p != null) {
                 // TODO: Consider more organized score / cost:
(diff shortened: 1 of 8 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -470,6 +466,23 @@ private ScoredPlan planFilterWithInJoin(@Nonnull PlanContext planContext, @Nonnu
...
     }
 
+    // Get the key expression for the index entries, which includes primary key fields for normal indexes.
+    private KeyExpression indexKeyExpressionForPlan(@Nullable KeyExpression commonPrimaryKey, @Nonnull Index index) {
+        KeyExpression indexKeyExpression = index.getRootExpression();
+        if (indexKeyExpression instanceof KeyWithValueExpression) {
+            indexKeyExpression = ((KeyWithValueExpression) indexKeyExpression).getKeyExpression();
+        }
+        if (commonPrimaryKey != null && indexTypes.getValueTypes().contains(index.getType())) {
+            final List<KeyExpression> keys = new ArrayList<>(commonPrimaryKey.normalizeKeyForPositions());
```
normen662, line 476: This works as `commonPrimaryKey.createsDuplicates() == false`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -470,6 +466,23 @@ private ScoredPlan planFilterWithInJoin(@Nonnull PlanContext planContext, @Nonnu
         return null;
     }
 
+    // Get the key expression for the index entries, which includes primary key fields for normal indexes.
+    private KeyExpression indexKeyExpressionForPlan(@Nullable KeyExpression commonPrimaryKey, @Nonnull Index index) {
```
normen662, line 470: Is this only ever called for value indexes? if so, let's put a `Verify.verify()` in here!

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: It is called for every index that the planner considers and needs to return something always. The specific handling of value index types is a few lines down (`indexTypes.getValueTypes().contains(index.getType())`), in which case it just returns the as-declared key.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: Right. Duplicates aren't allowed for a primary key. The validator should complain at definition time and save record does `evaluateSingleton`.

In some sense, what you'd like is for the index maintainer to tell the planner what the physical layout of the index entries is. But there isn't actually a vocabulary for that. It might seem that key expressions are, but, in fact, the maintainer is free to interpret those however it chooses. Their evaluation methods are only a convenience for a conventional interpretation. So, really, I suspect, there needs to be a planner-oriented intermediate form for this. That would also be an opportunity to introduce rigorous correlation.

Using normalized key expressions and only doing this for value indexes is kind of a compromise until that comes along.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -470,6 +466,23 @@ private ScoredPlan planFilterWithInJoin(@Nonnull PlanContext planContext, @Nonnu
...
+        KeyExpression indexKeyExpression = index.getRootExpression();
+        if (indexKeyExpression instanceof KeyWithValueExpression) {
+            indexKeyExpression = ((KeyWithValueExpression) indexKeyExpression).getKeyExpression();
+        }
+        if (commonPrimaryKey != null && indexTypes.getValueTypes().contains(index.getType())) {
```
MMcM, line 475: We could put in an additional check of a flag in `RecordQueryPlannerConfiguration` here. That would allow turning this behavior off (giving only the declared fields), if we are concerned about unforeseen consequences.

PR #1170, opened 2021-03-10
Resolves #1155: Need a way to tell whether the result of planning a query is only sorted by the requested keys.

### LINKED ISSUE #1155: Need a way to tell whether the result of planning a query is only sorted by the requested keys
Somewhat in the spirit of #1098, this means that each record has a new value for that sort key.

MMcM:
Includes #1169 and thereby #1168, since if the planner doesn't use the full key it doesn't have a complete picture of how the index entries are ordered, so it may be simpler to only look at the third (last) commit.

It's possible that there is a more general / forward-looking way of exposing the determination than just as another field in the plan.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -1457,9 +1457,31 @@ public class RecordQueryPlanner implements QueryPlanner {
     private KeyExpression getKeyForMerge(@Nullable KeyExpression sort, @Nonnull KeyExpression candidateKey) {
         if (sort == null || sort.isPrefixKey(candidateKey)) {
             return candidateKey;
+        } else if (candidateKey.isPrefixKey(sort)) {
+            return sort;
         } else {
-            return Key.Expressions.concat(sort, candidateKey);
+            return concatWithoutDuplicates(sort, candidateKey);
         }
+    }
+
+    // TODO: Perhaps this should be a public method on Key.Expressions.
+    private ThenKeyExpression concatWithoutDuplicates(@Nullable KeyExpression expr1, @Nonnull KeyExpression expr2) {
+        final List<KeyExpression> children = new ArrayList<>(2);
+        if (expr1 instanceof ThenKeyExpression) {
+            children.addAll(((ThenKeyExpression)expr1).getChildren());
+        } else {
+            children.add(expr1);
+        }
+        if (expr2 instanceof ThenKeyExpression) {
+            for (KeyExpression child : ((ThenKeyExpression)expr2).getChildren()) {
+                if (!children.contains(child)) {
+                    children.add(child);
+                }
+            }
+        } else if (!children.contains(expr2)) {
+            children.add(expr2);
+        }
+        return new ThenKeyExpression(children);
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -321,7 +321,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             if (sort == null) {
                 p = planNoFilterNoSort(planContext, index);
             } else {
-                p = planSortOnly(new CandidateScan(planContext, index, sortReverse), index.getRootExpression(), sort);
+                p = planSortOnly(new CandidateScan(planContext, index, sortReverse), indexKeyExpressionForPlan(planContext.commonPrimaryKey, index), sort);
             }
             if (p != null) {
                 if (bestPlan == null || p.score > bestPlan.score ||
@@ -452,11 +452,7 @@ public class RecordQueryPlanner implements QueryPlanner {
             bestPlan = planIndex(planContext, filter, null, planContext.commonPrimaryKey, intersectionCandidates);
         }
         for (Index index : planContext.indexes) {
-            KeyExpression indexKeyExpression = index.getRootExpression();
-            if (indexKeyExpression instanceof KeyWithValueExpression) {
-                indexKeyExpression = ((KeyWithValueExpression) indexKeyExpression).getKeyExpression();
-            }
-
+            KeyExpression indexKeyExpression = indexKeyExpressionForPlan(planContext.commonPrimaryKey, index);
             ScoredPlan p = planIndex(planContext, filter, index, indexKeyExpression, intersectionCandidates);
             if (p != null) {
                 // TODO: Consider more organized score / cost:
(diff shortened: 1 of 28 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
```diff
@@ -143,6 +143,15 @@ default int maxCardinality(@Nonnull RecordMetaData metaData) {
         return UNKNOWN_MAX_CARDINALITY;
     }
 
+    /**
+     * Indicates whether the ordering of records from this plan is fully accounted for by its requested sort.
+     * That is, each new record will have a different value for that sort key.
+     * @return {@code true} if this plan is fully sorted according to the query definition
+     */
+    default boolean isFullySorted() {
```
MMcM, line 151: Perhaps it would be better to do this with #1177's `QueryPlanInfo`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBNestedFieldQueryTest.java
```diff
@@ -318,12 +317,11 @@ public void nested2() throws Exception {
...
             assertThat(plan,
                     filter(nestedComponent,
-                            intersection(
-                                    indexScan(allOf(indexName("composite"), bounds(hasTupleString("[[something, 1],[something, 1]]")))),
-                                    indexScan(allOf(indexName("duplicates"), bounds(hasTupleString("[[something, something],[something, something]]"))))
-                            )));
+                            indexScan(allOf(indexName("duplicates"), bounds(hasTupleString("[[something, something, 1],[something, something, 1]]"))))
+                            ));
         } else {
             assertThat(plan, primaryKeyDistinct(indexScan(allOf(indexName("complex"), bounds(hasTupleString("[[something, 1, 10, 20],[something, 1, 10, 20]]"))))));
```
normen662, line 326: What should happen here?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/QueryPlanFullySortedTest.java
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Tests for {@link com.apple.foundationdb.record.query.plan.plans.QueryPlan#isFullySorted}.
+ */
+@Tag(Tags.RequiresFDB)
+public class QueryPlanFullySortedTest extends FDBRecordStoreQueryTestBase {
+
+    RecordMetaData metaData;
+    QueryPlanner planner;
+
+    protected void setup(@Nullable RecordMetaDataHook hook) {
```
normen662, line 54: Should this just work in the new planner? Feels like the structure of the plan will compute it. Can we maybe put some `@DualPlannerAnnotation`s here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -71,11 +71,18 @@
     protected final ScanComparisons comparisons;
     protected final boolean reverse;
 
-    public RecordQueryIndexPlan(@Nonnull final String indexName, @Nonnull IndexScanType scanType, @Nonnull final ScanComparisons comparisons, final boolean reverse) {
+    protected final boolean fullySorted;
+
+    public RecordQueryIndexPlan(@Nonnull final String indexName, @Nonnull IndexScanType scanType, @Nonnull final ScanComparisons comparisons, final boolean reverse, final boolean fullySorted) {
```
normen662, line 76: Hmm, so it looks like that if you don't call this constructor the fully sorted property is not set. That makes problematic for the new planner. Can we either compute that property in a common code path or add some logic to `DataAccessRule`?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1033,7 +1046,8 @@ private ScoredPlan planAndWithNesting(@Nonnull CandidateScan candidateScan,
     private ScoredPlan planFieldWithComparison(@Nonnull CandidateScan candidateScan,
                                                @Nonnull KeyExpression index,
```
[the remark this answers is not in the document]
MMcM: Sure.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBAndQueryToIntersectionTest.java
```diff
@@ -448,7 +448,7 @@ public void testComplexLimits4(boolean shouldDeferFetch) throws Exception {
     @ParameterizedTest
     @BooleanSource
     public void testAndQuery7(boolean shouldDeferFetch) throws Exception {
-        RecordMetaDataHook hook = complexPrimaryKeyHook();
+        RecordMetaDataHook hook = complexPrimaryKeyHook(true);
```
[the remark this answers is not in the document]
MMcM: Skips adding a second index index, equivalent for this situation, that just confuses the test.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBNestedFieldQueryTest.java
MMcM: The new planner is more correct. I realizes that the duplicate conditions are both satisfied by the single index key. The old planner insists on finding an index that does each of them separately.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlan.java
```diff
@@ -143,6 +143,15 @@ default int maxCardinality(@Nonnull RecordMetaData metaData) {
         return UNKNOWN_MAX_CARDINALITY;
     }
 
+    /**
+     * Indicates whether the ordering of records from this plan is fully accounted for by its requested sort.
+     * That is, each new record will have a different value for that sort key.
```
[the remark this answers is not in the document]
MMcM: That name is fine, sure.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/QueryPlanFullySortedTest.java
MMcM: Added those annotations and all pass now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
MMcM: Given the current structure, it seemed to make more sense to do this in `RemoveSortRule`, which is currently responsible for lining up the declared sort and the candidate scan's sort. It now additionally sets the new property when appropriate.

MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/rules/RemoveSortRule.java
```diff
@@ -97,6 +99,13 @@ public void onMatch(@Nonnull PlannerRuleCall call) {
             }
         }
 
+        if (!orderingKeysIterator.hasNext()) {
```
normen662, line 102: We cannot mutate a plan in place as there may be other plans ranging over this `RecordQueryPlan`. You need to make a defensive copy here.

normen662 (approved):

normen662:
test this please

PR #1178, opened 2021-03-11
Resolves #1175: represent an aggregation function using a set of a…

### LINKED ISSUE #1175: represent an aggregation function using a set of aggregation keys rather than a KeyExpression
Currently the class `IndexAggregateFunction` (and derivates) use a `KeyExpression` called `operand` to represent both grouping expressions, and grouped expressions (the aggregate function's argument(s)). We also allow for certain attributes of this class, specifically `index` to be nulled out or not. In general it appears as if there are two different versions of objects created from this class:
1. function objects that do not have an index (they are unbound)
2. function objects that do have an index set (they are bound to that said index)

Interestingly, there is no point fixing the order of the grouping expressions when such a function is not yet bound to an index as the index dictates the order of these expressions.

This issue should target the separation of these two use cases into two actual classes, one for the bound and one for the unbound case. In the unbound case we want to represent the grouping keys in set rather than positionally fixed.

normen662:
…ggregation keys

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
@@ -1,0 +1,339 @@
+/*
+ * IndexAggregateFunctionCall.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.metadata;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.metadata.expressions.EmptyKeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.query.expressions.AndComponent;
+import com.apple.foundationdb.record.query.expressions.BaseField;
+import com.apple.foundationdb.record.query.expressions.Comparisons;
+import com.apple.foundationdb.record.query.expressions.FieldWithComparison;
+import com.apple.foundationdb.record.query.expressions.NestedField;
+import com.apple.foundationdb.record.query.expressions.QueryComponent;
+import com.apple.foundationdb.record.query.plan.temp.EnumeratingIterable;
+import com.apple.foundationdb.record.query.plan.temp.TopologicalSort;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Class to represent a construct {@code AGG by Set.of(col1, col2, ...)}. This is the unbound equivalent of
+ * an {@link IndexAggregateFunction} which itself is bound to an index and represents its group-by columns as a list of
+ * {@link KeyExpression} that match the order imposed by bound index rather than a set.
+ *
+ * There are two use cases for this class:
+ * <ol>
+ * <li> regular aggregation functions that need the enumeration of permutation of the grouping expressions in order to
+ *      bind the call to a {@link IndexAggregateFunction}. The logic implementing this use case calls
+ *      {@link #enumerateIndexAggregateFunctionCandidates(String)} at some point.
+ * </li>
+ * <li> aggregation-oriented operators that function more like closures over a group (see
+ *      {@link com.apple.foundationdb.record.query.plan.bitmap.ComposedBitmapIndexAggregate} for such an example)
+ *      in which cases it may be needed to propagate the original {@link GroupingKeyExpression} to the binding phase
+ *      thus eliminating the possibility of an enumeration of permutations of grouping columns altogether.
+ *      In those cases it may be simply impossible to compute an equivalent list of such grouping columns from the original
+ *      grouping key expression without losing some of the nesting/repeating information.
+ * </li>
+ * </ol>
+ *
+ * Note that if a client of this class intends to use an aggregate function call for regular aggregation functions and
+ * intends to call methods such as {@link #enumerateIndexAggregateFunctionCandidates(String)}, the following must hold
+ * in order to yield meaningful results:
+ *
+ * <pre>
+ * {@code groupingKeyExpression = concat(groupingExpressions, groupedExpression).group(groupedExpression.getColumnSize()}
+ * </pre>
+ *
+ * which ideally is true, but may be violated in the presence of repeated structures.
+ */
+@SuppressWarnings("unused")
+@API(API.Status.INTERNAL)
+public class IndexAggregateFunctionCall {
+    /**
+     * Function name used in this function call.
+     */
+    @Nonnull
+    private final String functionName;
+
+    /**
+     * Grouping key expression that holds information over both the grouping and the grouped part of the expression
+     * this function call ranges over.
+     *
+     * In the normal use case this expression is computed by reconstructing a {@link GroupingKeyExpression} using
+     * the iterable of grouping expressions and the grouped expression that is handed in (to certain overloaded
+     * constructors).
+     *
+     * If required, the client can use {@link #IndexAggregateFunctionCall(String, GroupingKeyExpression)} in order
+     * to invert that logic in which case the grouping key expressions and the grouped expression is computed
+     * from the {@link GroupingKeyExpression} passed in to the respective constructor. Additionally, this field
+     * holds the original grouping key expression handed in thus preserving the intricate structure of nested repeated
+     * fields, etc.
+     *
+     * See {@link #toIndexAggregateFunction(String)} which uses this field directly.
+     */
+    @Nonnull
+    private final GroupingKeyExpression groupingKeyExpression;
+
+    /**
+     * A set of grouping expressions. Note that the implementing set is an instance of class {@link ImmutableSet} which
+     * provides iteration in insertion order. That property may be useful for clients that attempt to bind
+     * calls to index aggregate functions using the grouping keys as queried (written).
+     */
+    @Nonnull
+    private final Set<KeyExpression> groupingExpressions;
+
+    /**
+     * The grouped expression, i.e., the expression that captures what is computed per group.
+     */
+    @Nonnull
+    private final KeyExpression groupedExpression;
+
+    /**
+     * Overloaded constructor using a {@link GroupingKeyExpression} that represents both the grouping expressions and
+     * the grouped expression. The constructor is necessary to allow function calls that can use nested repeated
+     * structures to adequately represent function calls over those structures.
+     * @param functionName function name of the call
+     * @param groupingKeyExpression the grouping key expression
+     */
+    public IndexAggregateFunctionCall(@Nonnull String functionName,
+                                      @Nonnull GroupingKeyExpression groupingKeyExpression) {
+        this(functionName,
+                groupingKeyExpression,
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
```diff
@@ -0,0 +1,272 @@
...
+     * function call.
+     * In order to bind the function call, we need to associated the aggregate function call with an index and also
+     * pin the order of grouping key expressions in a way that is compatible with that index. Note that the returned
+     * index aggregate functions are merely candidates for actual index-compatible ones as we don't check here the
+     * conditions that constitute a compatible index. On the contrary, this method is intended to be used by the logic
+     * that builds the actual function call for later consumption.
+     * @param indexName indexName to use
+     * @return a {@link Stream} of {@link IndexAggregateFunction}s
+     */
+    public Stream<IndexAggregateFunction> enumerateIndexAggregateFunctionCandidates(@Nonnull String indexName) {
```
MMcM, line 161: Should use `index` or `indexName` consistently.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
```diff
@@ -0,0 +1,272 @@
...
+@API(API.Status.INTERNAL)
+public class IndexAggregateFunctionCall {
+    @Nonnull
+    private final String name;
+    @Nonnull
+    private final GroupingKeyExpression groupingKeyExpression;
+    @Nonnull
+    private final Set<KeyExpression> groupingExpressions;
+    @Nonnull
+    private final Optional<KeyExpression> groupedExpressionOptional;
```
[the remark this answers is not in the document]
normen662: No (I think), because `EmptyKeyExpression` is still something in terms of other methods using the result of this down the road. The reason I added the `Optional<>` was because I didn't want to presume that there was a no legitimate reason for empty to occur _naturally_. Then again, I don't know whether the original reason for empty in the code base was to actually represent the absence of a key expression or a key expression that returns an empty.
Also, optionals feel more natural to me to indicate the absence of something and nicely compose functionally. Let me know what you want me to do here.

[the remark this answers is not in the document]
MMcM: > the absence of a key expression or a key expression that returns an empty

What is the practical difference? Returning empty means contributing zero columns to the evaluation.

Is the code that would make a distinction using this new class yet to be written? I don't see anything below that wouldn't just work.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
normen662: changed every signature in this class to use  to `indexName`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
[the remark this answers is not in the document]
normen662: For instance, if I want to create a GroupingKeyExpression from it and some grouping keys, I can easily say:

`new GroupingKeyExpression(concat(groupingKey, groupedKey), groupedKey.getColumnSize())`

That's fine if it is not `EmptyKeyExpression`; if it is you probably want to say something like

`new GroupingKeyExpression(groupingKey, 0)`

Now regardless of whether `groupedExpression` is an `EmptyKeyExpression` or an `Optional<KeyExpression>` the code needs to make this distinction (case 1 or 2) with the difference that the caller is forced to think about it from a typing perspective: what happens if the key expression is not there? If it is just the ``EmptyKeyExpression` I can write something like: `new GroupingKeyExpression(concat(groupingKey, groupedKey), groupedKey.getColumnSize())`
even though I may not intend to. And this is exactly what happened downstream where I messed up an index match because further down I got an `EmptyKeyExpression` within a bunch of other keys after some other manipulations.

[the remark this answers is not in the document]
normen662: Resolves off-thread: Will change implementation to use `EmptyKeyExpression` mostly due to convention.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java
```diff
@@ -110,6 +112,25 @@ private IndexFunctionHelper() {
...
+                                                                                                                            @Nonnull IndexAggregateFunctionCall functionCall,
+                                                                                                                            @Nonnull List<String> recordTypeNames) {
+        return indexesForRecordTypes(store, recordTypeNames)
+                .filter(store::isIndexReadable)
+                .flatMap(index ->
+                        functionCall.enumerateIndexAggregateFunctionCandidates(index.getName())
+                                .flatMap(indexAggregateFunction -> {
+                                    final IndexMaintainer indexMaintainer = store.getIndexMaintainer(index);
+                                    return indexMaintainer.canEvaluateAggregateFunction(indexAggregateFunction)
+                                           ? Stream.of(Pair.of(indexAggregateFunction, store.getIndexMaintainer(index)))
```
ohadzeliger, line 125: Isn't `store.getIndexMaintainer(index)` the same as `indexMaintainer`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java
```diff
@@ -110,6 +112,25 @@ private IndexFunctionHelper() {
                 .min(Comparator.comparing(i -> i.state.index.getColumnSize()));
     }
 
+    public static Optional<Pair<IndexAggregateFunction, IndexMaintainer>> enumerateIndexMaintainersForAggregateFunctionCall(@Nonnull FDBRecordStore store,
```
ohadzeliger, line 115: Javadoc (esp. since the min() at line 128 may have some assumptions)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java
```diff
@@ -264,4 +285,11 @@ public static IndexAggregateFunction countUpdates(@Nonnull KeyExpression by) {
         return indexMaintainerForAggregateFunction(store, function, recordTypeNames)
                 .map(i -> function.cloneWithIndex(i.state.index.getName()));
     }
+
+    public static Optional<IndexAggregateFunction> bindAggregateFunctionCall(@Nonnull FDBRecordStore store,
```
ohadzeliger, line 289: Javadoc? Does this return the function that can matched by an index?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.stream.StreamSupport;
+
+/**
+ * Class to represent a construct {@code AGG by Set.of(col1, col2, ...)}. This is the unbound equivalent of
+ * {@link IndexAggregateFunction} which is bound to an index and represents its group-by columns as a list of
+ * {@link KeyExpression} that match the order imposed by bound index rather than a set.
+ */
+@SuppressWarnings("unused")
+@API(API.Status.INTERNAL)
+public class IndexAggregateFunctionCall {
```
ohadzeliger, line 57: Would the name `UnboundIndexAggregateFunction` be appropriate?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java
```diff
@@ -88,7 +88,7 @@
     @Nonnull
```
ohadzeliger, line 88: Is the description in line 85 still accurate given that the new Function does not consider ordering of expressions?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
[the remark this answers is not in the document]
normen662: No, this is not something that can appear in the end as part of something that is executed. This is merely a transitional structure, much like a logical `RelationalExpression`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
```diff
@@ -0,0 +1,269 @@
...
+     * pin the order of grouping key expressions in a way that is compatible with that index. Note that the returned
+     * index aggregate functions are merely candidates for actual index-compatible ones as we don't check here the
+     * conditions that constitute a compatible index. On the contrary, this method is intended to be used by the logic
+     * that builds the actual function call for later consumption.
+     * @param indexName indexName to use
+     * @return a {@link Stream} of {@link IndexAggregateFunction}s
+     */
+    public Stream<IndexAggregateFunction> enumerateIndexAggregateFunctionCandidates(@Nonnull String indexName) {
+        final EnumeratingIterable<KeyExpression> groupingPermutations =
+                TopologicalSort.topologicalOrderPermutations(getGroupingExpressions(), ImmutableSetMultimap.of());
```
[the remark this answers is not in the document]
normen662: From the javadoc of `TopologicalSort`:

```
The main purpose on this class is to provide a specific iterable that can efficiently traverse
all possible permutations of the input set that do not violate the given dependency constraints.

The iterable adheres to the following requirements:

 * it does not violate the given constraints
 * it produces all possible orderings under the given constraints
 ```

If there are no given constraints, it produces all possible orderings, i.e. all permutations of the set.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java (hunk 3 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java:125]: Yup. Fixing.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java (hunk 5 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java:289]: Yes. Will add a javadoc.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java (hunk 6 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java:57]: I guess you can do that. I kind of like function call better due to in my mind being the representation of a call-site of a function. Very much like here -> https://en.wikipedia.org/wiki/Call_site

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java (hunk 7 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/bitmap/ComposedBitmapIndexAggregate.java:88]: Will change the javadoc. This logic here should take a call and bind it. It's one of the TODOs left. If you look at how it's used downstream, the caller binds the call to a function instead of going through this path.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java (hunk 4 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexFunctionHelper.java:115]: Will add javadoc.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java [hunk 9]
```diff
@@ -0,0 +1,269 @@
...
+                groupingKeyExpression.getGroupedSubKey());
+    }
+
+    public IndexAggregateFunctionCall(@Nonnull String name,
+                                      @Nonnull Iterable<KeyExpression> groupingExpressions,
+                                      @Nonnull KeyExpression groupedExpression) {
+        this(name, toGroupingKeyExpression(ImmutableList.copyOf(groupingExpressions), groupedExpression), groupingExpressions, groupedExpression);
+    }
+
+    public IndexAggregateFunctionCall(@Nonnull String name,
```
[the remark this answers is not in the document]
normen662: done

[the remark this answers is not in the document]
normen662: done

ohadzeliger (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexAggregateFunctionCall.java
MMcM: ✅

MMcM (approved):

PR #1185, opened 2021-03-17
Resolves #1175: represent an aggregation function as using a set of a…

### LINKED ISSUE #1175: represent an aggregation function using a set of aggregation keys rather than a KeyExpression

normen662:
…ggregation keys

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
@@ -167,6 +167,12 @@ public class GroupingKeyExpression extends BaseKeyExpression implements KeyExpre
         return getWholeKey().getSubKey(0, getGroupingCount());
     }
 
+    /**
+     * Method that indicates to the caller if the grouping expression defined in this expression is freely
+     * permutable, i.e., would another grouping key expression with a different order of gr
+     * @return
+     */
+    @API(API.Status.INTERNAL)
     public boolean isGroupingPermutable() {
         return !createsDuplicates(); // TODO this can be optimized
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
@@ -166,17 +166,7 @@ public class GroupingKeyExpression extends BaseKeyExpression implements KeyExpre
     public KeyExpression getGroupingSubKey() {
         return getWholeKey().getSubKey(0, getGroupingCount());
     }
-
-    /**
-     * Method that indicates to the caller if the grouping expression defined in this expression is freely
-     * permutable, i.e., would another grouping key expression with a different order of gr
-     * @return
-     */
-    @API(API.Status.INTERNAL)
-    public boolean isGroupingPermutable() {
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
```diff
@@ -167,6 +167,10 @@ public KeyExpression getGroupingSubKey() {
         return getWholeKey().getSubKey(0, getGroupingCount());
     }
 
+    public boolean isGroupingPermutable() {
```
MMcM, line 170: This probably wants `API` `INTERNAL`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
```diff
@@ -167,6 +167,10 @@ public KeyExpression getGroupingSubKey() {
...
+        return !createsDuplicates(); // TODO this can be optimized
```
MMcM, line 171: I'm having trouble formulating the theoretical reason that a single record contributing to more than one aggregate bucket would mean that permutations of the bucket keys aren't allowed. Is the real problem a lack of a solid way to specify conditions on such a key absent correlations? So this method really just means what it does, namely skip trying the particular permutation algorithm?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
normen662: That is absolutely correct. If I cannot take a `GroupingKeyExpression` apart into the grouping expressions (in order to permute them) in a lossless way (i.e. can I put them back together to get the original `GroupingKeyExpression`) I cannot do any permutations. As a consequence this method is used to skip that permuting algorithm (see `IndexFunctionHelper`)

normen662:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/GroupingKeyExpression.java
MMcM: It might be that there is a better name that indicates the discretionary nature. Or maybe just making `INTERNAL` and adding a comment summarizing the reasoning is enough.

MMcM (approved):

PR #1190, opened 2021-03-19
Resolves #1189: Remove additional deprecated RecordCursor methods

### LINKED ISSUE #1189: Remove additional deprecated RecordCursor methods
There are a few deprecated methods in `RecordCursor` that we should clean up as part of 3.0. They are all fairly easy for adopters to replace, and it would clean up that interface.

alecgrieser:
This removes additional deprecated methods from the `RecordCursor` API. These methods should be fairly easy for the user to replace with their non-deprecated variants.

This resolves #1189.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
@@ -425,16 +425,6 @@ public interface RecordCursor<T> extends AutoCloseable {
     }
 
     /**
-     * @deprecated Use {@link #limitRowsTo(int)} instead.
-     * @param limit the maximum number of records to return
-     * @return a new cursor that will return at most {@code limit} records
-     */
-    @Deprecated
-    default RecordCursor<T> limitTo(int limit) {
-        return limitRowsTo(limit);
-    }
-
-    /**
      * Get a new cursor that will only return records up to the given limit.
      * @param limit the maximum number of records to return
      * @return a new cursor that will return at most {@code limit} records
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
```diff
@@ -694,21 +688,6 @@ public boolean isLimitReached() {
...
-     * @param func function to be called if the cursor is empty to give another source of records
-     * @return a new cursor that returns the same records as this cursor 
-     * or the result of {@code func} if this cursor does not produce any records
-     * @deprecated because it does not support continuations and is easy to misuse.
-     *             Use {@link #flatMapPipelined(Function, BiFunction, byte[], int)} instead.
-     */
-    @API(API.Status.DEPRECATED)
-    @Deprecated
-    @Nonnull
-    default RecordCursor<T> orElse(@Nonnull Function<Executor, RecordCursor<T>> func) {
```
MMcM, line 708: I think https://github.com/FoundationDB/fdb-record-layer/blob/7dc225867b944a58314c2ffda42661a6f6e908cd/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/OrElseCursor.java#L65 can go, too.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
alecgrieser: Ah, good call

alecgrieser:
Rebase addresses a merge conflict in the release notes; also added 3ee421d to fix javadoc compilation

MMcM (approved):

PR #1192, opened 2021-03-19
Resolves #1191: Remove non-controversial deprecated methods

### LINKED ISSUE #1191: Remove non-controversial deprecated methods
There are other `@Deprecated` methods beyond those in #1163 and #1189.

Some of these are only deprecated for new usage and still need to remain to support existing, serialized meta-data.

Some are associated with the switchover to resolved key-space paths, which is tricky because of some missing abstractions. They may need to remain until we are certain that there are substitutes for all use cases.

The remainder should be removed in 3.0.

MMcM:
What remains:
* `RecordCursor`, `OrElseCursor` methods (#1189)
* `KeySpace`, `KeySpacePath`, `KeySpacePathImpl`, `KeySpacePathWrapper`, `ScopedDirectoryLayer`, `ScopedInterningLayer`, all around resolved / unresolved paths, which all present challenges in straightforward conversion.
* Override of `@Deprecated` methods from dependencies.
* Some meta-data items, that are deprecated in new code, but may still be present in serialized form and so need to be supported.


MMcM:
test this please

alecgrieser (approved):

PR #1194, opened 2021-03-19
Resolves #1193: Dependency updates for 3.0

### LINKED ISSUE #1193: Dependency updates for 3.0
For version 3.0, we should upgrade our dependency versions to more modern versions just to keep up with the any changes. The big things here are probably guava and protobuf. It looks like the biggest change to guava is the removal of GWT-RPC support, which the Record Layer wasn't using directly, so it should only cause problems for someone who was using that but also not shading guava. The big thing in more recent protobuf versions is that `optional` fields are now back in proto3 by default rather than being an experimental feature, which is cool given how we use those to indicate an unset field.

alecgrieser:
Updates dependency versions to be more modern.


alecgrieser:
In addition to these tests, for the big dependencies (mainly guava and protobuf), I also tried to see if there was anything called out in their release notes, and there didn't look to be anything major, other than `optional`s in proto3 being no longer an experimental feature.

MMcM (approved):

PR #1200, opened 2021-03-23
Update Spotbugs Gradle plugin.

MMcM:
This is less straightforward than the previous branch, due to some incompatible changes. Hence the separate PR.

* The package of the task has changed.
* `excludes` and `sourceSets` are gone.

It is simpler to disable running against the `test` configuration with `spotbugsTest.enabled = false`.

Previously, Protobuf generated files were excluded in three ways:
1. `task.exclude`, which doesn't exist any more.
2. A filter on `task.classes`, which isn't really needed.
3. An `excludeFilter` file, which does everything we need.

The Protobuf exclusions are still needed because:
* `SE_BAD_FIELD`: `Class x defines non-transient non-serializable instance field unknownFields`
* `MS_SHOULD_BE_FINAL`: `PARSER isn't final but should be`
* `UCF_USELESS_CONTROL_FLOW`: `Useless control flow in maybeForceBuilderInitialization()`

There are also two new warnings about nullity from `IdentityBiMap`, which are just ignored.

### DIFF
--- a/gradle/check.gradle
+++ b/gradle/check.gradle
@@ -69,12 +69,13 @@ quickCheck.dependsOn checkstyleMandatory
 apply plugin: 'com.github.spotbugs'
 
 spotbugs {
-    toolVersion = '3.1.7'
+//  toolVersion = '4.2.1'
     ignoreFailures = false
     effort = 'max'
-    sourceSets = [sourceSets.main]
     excludeFilter = rootProject.file('gradle/codequality/spotbugs_exclude.xml')
 }
+
+spotbugsTest.enabled = false
 
 def spotbugsFailed = false
 
@@ -87,7 +88,6 @@ def printInstructionsOnRunningWithHtmlOutput = task('printInstructionsOnRunningW
 }
 
 tasks.withType(rootProject.SpotBugsTask) { task ->
-    task.exclude '**/protogen/**'
     if (project.hasProperty('spotbugsEnableHtmlReport')) {
         // SpotBugs task can only have one report type enabled at a time
         reports {
@@ -100,14 +100,6 @@ tasks.withType(rootProject.SpotBugsTask) { task ->
 
     task.doFirst {
         spotbugsFailed = true
-        def  ft = task.classes
-        if (task.classes == null) {
-            return
-        }
-        task.classes = files(ft.files).filter {
-          !(it.path.contains('Proto$')
-            || it.path.contains('Proto.class'))
-        }
     }
     task.doLast {
         spotbugsFailed = false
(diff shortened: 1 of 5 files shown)


gradle/check.gradle
```diff
@@ -69,13 +69,14 @@ quickCheck.dependsOn checkstyleMandatory
 apply plugin: 'com.github.spotbugs'
 
 spotbugs {
-    toolVersion = '3.1.7'
+//  toolVersion = '4.2.1'
```
alecgrieser, line 72: Is the idea that this comment is documentation that the (default) `toolVersion` is 4.2.1?

gradle/check.gradle
```diff
@@ -69,13 +69,14 @@ quickCheck.dependsOn checkstyleMandatory
...
     ignoreFailures = false
     effort = 'max'
-    sourceSets = [sourceSets.main]
     excludeFilter = rootProject.file('gradle/codequality/spotbugs_exclude.xml')
 }
 
+spotbugsTest.enabled = false
```
alecgrieser, line 78: I think there may be a *fourth* way of excluding files from SpotBugs that may allow us to turn `spotbugsTest` back on, which is by using the `Source` filter in exclude file. At the moment, that's set to exclude all classes with `Proto` in the name:

https://github.com/FoundationDB/fdb-record-layer/blob/8382bf1a5bc172594e1d18a2630c9a603abf99ca/gradle/codequality/spotbugs_exclude.xml#L5-L7

(Which may be why `spotbugsMain` isn't hitting the same problems as `spotbugsTest`? That is, the filter may be correctly excluding all non-test generated protos but missing (some of?) the test generated protos?) In theory, I think we could use the [`<Source>`](https://spotbugs.readthedocs.io/en/stable/filter.html#source) tag to filter out the `protogen` files.

I also think merging this with `spotbugsTest` off and then re-enabling it if/when we figure out the right way to exclude those generated files would make sense.

alecgrieser (changes requested):
I left a possible fourth way to exclude files that we could explore, but I'm also fine with merging this as is and re-enabling `spotbugsTest` later

gradle/check.gradle
alecgrieser: Or maybe that regex should be `"~.*Proto\$?.*"` to match the filtering that's been removed?

gradle/check.gradle
MMcM: Yes, that was the idea. I could just take it out altogether.

gradle/check.gradle
MMcM: Using `Source` is clearer on intent, yes. So I switched to that.

There are actual failures in `spotBugsTest`. I suppose fixing them and keeping it enabled might be worthwhile at some point.

alecgrieser: I see. It would certainly be good to get that re-enabled, but that sounds like a separate thing from this PR. I filed #1204 to track that.

alecgrieser (approved):

PR #1208, opened 2021-03-26
Resolve #1207 - add toString to `QueryPlanResult`

### LINKED ISSUE #1207: Add `toString()` to `QueryPlanResult` to make it pretty in logs

ohadzeliger:
Resolve #1207

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
@@ -24,6 +24,7 @@ import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 /**
  * This class holds some additional information regarding the query plan that can be attached to the plan itself, without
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
```diff
@@ -73,6 +75,24 @@ public QueryPlanInfo() {
...
+    }
+
+    /**
+     * Return the Key Set for info map.
+     * @return the Key Sey for the info map
+     */
+    @SuppressWarnings("java:S1452")
+    @Nonnull
+    public Set<QueryPlanInfoKey<?>> keySet() {
+        return info.keySet();
```
normen662, line 93: Hmm, in general, I don't like mutable structures if they can be avoided. Returning the keyset here per se is allows modifications to happen and is breaking encapsulation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
```diff
@@ -73,6 +75,24 @@ public QueryPlanInfo() {
         return this;
```
normen662, line 75: General comment re: the put. Can we change the contract to create the info field with a map or through a designated builder object and then make it immutable?

ohadzeliger: The usage of this class is really to collect attributes "as we go" in various places of the code and therefore does not lends itself to be immutable (other than copy-on-write or clone-and-add patterns, which are not a good fit for the usage).
I will create an immutable wrapper over the `KeySet` in order to prevent side-channel modification to the contents of the map.

ohadzeliger: This is done.

normen662 (approved):

PR #1210, opened 2021-03-26
Fixes #1209: planNestedFieldChild stop condition isn't quite right

### LINKED ISSUE #1209: planNestedFieldChild stop condition isn't quite right
The old planner optimization in #1143 can get confused about whether it can continue with more keys / conditions after the nested child, generating plans that use only part of the nested key's fields followed by non-nested, even though these follow some unbound nested keys.

MMcM:
See new test case. Prior to this change, it generates an impossible index scan of all three conditions and no residual filter.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -1828,6 +1828,10 @@ public class RecordQueryPlanner implements QueryPlanner {
          */
         private boolean foundComparison;
         /**
+         * True if {@code foundComparison} completely accounted for the child.
+         */
+        private boolean foundCompleteComparison;
+        /**
          * Accumulate matching comparisons here.
          */
         @Nonnull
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1881,8 +1885,8 @@ public ScoredPlan plan() {
             for (KeyExpression child : indexChildren) {
                 if (!doneComparing) {
                     planChild(child);
-                    if (!comparisons.isEquality() || !foundComparison) {
-                        // Didn't add another equality; done matching filters to index.
+                    if (!comparisons.isEquality() || !foundCompleteComparison) {
+                        // Didn't add another equality or only did part of child; done matching filters to index.
```
normen662, line 1889: So if you don't match all filters, don't you have to reapply all filters that is all filters matched and unmatched? E.g. if you have a repeated `a` and you have an index on `a(FanOut).b` and a query `a.oneOfThem(b = 5 and c = 6)` wouldn't you need to do `FILTER(INDEXSCAN([5, 5], a.oneOfThem(b = 5 and c = 6)` since you will have to scope the inner `and` to one `a`?

MMcM: Yes.

In this context, `child` is a `KeyExpression`. This is the case of not matching all of the nested _keys_ to comparisons. In which case you cannot continue up a level with more unnested keys and comparisons because there is an unmatched key in the way.

I can add a test case for the query you describe, since I don't immediately see one.

MMcM: Added.

normen662 (approved):

PR #1216, opened 2021-04-02
Resolves #1213: New indexes should generally begin life as DISABLED instead of WRITE_ONLY if not READABLE

### LINKED ISSUE #1213: New indexes should generally begin life as DISABLED instead of WRITE_ONLY if not READABLE
Right now, when a new index is added to a record store for the first time with existing data, the default behavior is for the store to see if there's enough data in the store for it to build in the index in-line, and if there is, it builds the index. But if there isn't enough data, then the index is created in `WRITE_ONLY` mode and the user has to then build the index. The problem there is that this means that work is now being done to expend the I/O to maintain the index for any records added in the interim, but the index can't actually be used until the index is built anyway, so the record updates use more I/O until the index build is actually performed for essentially no reason.

It should be safe to therefore turn the default state for the index from `WRITE_ONLY` to `DISABLED`. The main consequence of this is that users will have to make sure to use an `IndexStatePrecondition` that is compatible with the new index state when they build the index. The default `IndexStatePrecondition` (`BUILD_IF_DISABLED_CONTINUE_IF_WRITE_ONLY`) actually already supports this, so it's only users of `ERROR_IF_DISABLED_CONTINUE_IF_WRITE_ONLY` who should be affected, and there's already a warning on that API that it's not advisable to use and is there only as a compatibility bridge from before 2.8.90.0.

I think it would make sense to call this out in 3.0 rather than sneak it in at a different time. Also note that this issue describes the _default_ behavior when an index is added. The user also has the option of supplying a `UserVersionChecker` which overrides this decision, so any adopter that does so will not be affected by this change either.

alecgrieser:
This updates the default value in the check version logic and also in the `UserVersionChecker`'s default implementation of `needRebuildIndex`. It leaves the old method around so that users can opt in to the old behavior by providing a `UserVersionChecker` with that old method around. Javadoc has been added in a few places to tell the user what the point of that method is and what the return value means, and also points them to those existing static utility methods that they can use within their implementations. The `OnlineIndexer` already had a default `IndexStatePrecondition` that was compatible with this change, so I mentioned in the release notes how one can change it if one is using the only incompatible `IndexStatePrecondition`. The other changes are to tests, usually to ones that were explicitly testing that an index began as `WRITE_ONLY`, but they usually really only cared that the index wasn't `READABLE`, as that would imply a bug where indexes were being made `READABLE` before they were built. But the tests didn't _actually_ care about the difference between an index being `DISABLED` or `WRITE_ONLY`.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
@@ -174,16 +174,48 @@ public interface FDBRecordStoreBase<M extends Message> extends RecordMetaDataPro
                                                     RecordMetaDataProvider metaData);
 
         /**
-         * Determine what to do about an index needing to be built.
+         * Determine what to do about an index needing to be built. When a {@link FDBRecordStore} is opened,
+         * this method will be called on any index that has been added to the
+         * {@link com.apple.foundationdb.record.RecordMetaData RecordMetaData} since the last time the record store was
+         * opened. The index will then be initialized with the {@link IndexState} returned, which in turn determines
+         * whether the index must be maintained during record inserts and deletes and also whether the index
+         * can be read (for queries, for example). In general, an index is only really useful if it is
+         * {@link IndexState#READABLE}, but if this method returns {@link IndexState#READABLE}, then the
+         * index must be built in the same transaction that opens the record store, which can lead to errors
+         * on large stores if the index cannot be built in the
+         * <a href="https://apple.github.io/foundationdb/known-limitations.html#long-running-transactions">five second
+         * FoundationDB transaction time limit</a>.
+         *
+         * <p>
+         * By default, this will return {@link IndexState#READABLE} for any indexes on new types (which
+         * can be used right away without doing any I/O) or if the number of records in the store is small (below
+         * {@link FDBRecordStore#MAX_RECORDS_FOR_REBUILD}). However, if the record store is large, this will return
+         * {@link IndexState#DISABLED}, which indicates that the index should not be maintained and that it cannot be
+         * used until the index is built by the {@link OnlineIndexer}.
+         * </p>
+         *
+         * <p>
+         * For adopters, two utility methods are provided that can be used to make implementing this method easier.
+         * The first is {@link FDBRecordStore#disabledIfTooManyRecordsForRebuild(long, boolean)}, which replicates
+         * the default behavior. The second is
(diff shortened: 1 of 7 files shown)


alecgrieser:
@MMcM I think you have the most familiarity with the check version logic, but feel free to nominate someone else.
@jjezra I wanted your opinion because I believe this came out of a discussion you and I had.
@scottfines I wanted your eyes on some of the docs, etc., because I wanted to see if the docs I added would make any sense to someone who wasn't as familiar with index states, etc.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -174,16 +174,48 @@ default FDBStoreTimer getTimer() {
...
          * @param indexOnNewRecordTypes <code>true</code> if all record types for the index are new (the number of
          *                              records related to this index is 0), in which case the index is able to be
          *                              "rebuilt" instantly with no cost.
          * @return the desired state of the new index. If this is {@link IndexState#READABLE}, the index will be built right away
+         * @see FDBRecordStore#disabledIfTooManyRecordsForRebuild(long, boolean)
+         * @see FDBRecordStore#writeOnlyIfTooManyRecordsForRebuild(long, boolean)
          */
         default IndexState needRebuildIndex(Index index, long recordCount, boolean indexOnNewRecordTypes) {
-            return FDBRecordStore.writeOnlyIfTooManyRecordsForRebuild(recordCount, indexOnNewRecordTypes);
+            return FDBRecordStore.disabledIfTooManyRecordsForRebuild(recordCount, indexOnNewRecordTypes);
```
jjezra, line 218: Is the `index` arg ever used? (couldn't find a usage in `UserVersionChecker`'s implementations too). 
If not, should it be removed?

jjezra (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
alecgrieser: It's not used in here or in the Record Layer codebase itself, but this is an interface that adopters might themselves implement, and the index argument is there so that they could make different decisions about different indexes. (For example, you could imagine someone wanting to have "experimental" indexes return `DISABLED` all the time, even on new stores or new types. In that case, you could implement a `UserVersionChecker` that somehow identified the index as experimental (perhaps by the `IndexOptions` on that index) and then always returned `IndexState.DISABLED` regardless of the values of the other two parameters on such indexes.)

jjezra (approved):

scottfines (approved):

PR #1230, opened 2021-04-14
Resolves #1229: Expose range size estimate on Record Stores

### LINKED ISSUE #1229: Expose range size estimate on Record Stores
In FDB 6.3, the Transaction API got a new method that allows the user to estimate the size of a range without having to scan all of the data underlying it. In particular, it uses the "byte sample", a sample maintained by FDB so that data distribution can use that as a way to track the amount of data in each shard.

In addition to it potentially being useful for some users who want to know (roughly) how large their database is, we could also use it instead of the record count when creating a record store and want to reason about whether to rebuild an Index or not. See #7.

alecgrieser:
Exposes methods that use the new FDB 6.3 API for estimating the size of a range to estimate the size of a record store and of the records within.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
@@ -180,6 +180,16 @@ abstract class InstrumentedReadTransaction<T extends ReadTransaction> implements
     }
 
     @Override
+    public CompletableFuture<Long> getEstimatedRangeSizeBytes(final byte[] begin, final byte[] end) {
+        return underlying.getEstimatedRangeSizeBytes(begin, end);
+    }
+
+    @Override
+    public CompletableFuture<Long> getEstimatedRangeSizeBytes(final Range range) {
+        return underlying.getEstimatedRangeSizeBytes(range);
+    }
+
+    @Override
     public TransactionOptions options() {
         return underlying.options();
     }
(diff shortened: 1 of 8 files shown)


MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedTransaction.java
```diff
@@ -191,16 +191,6 @@ public void setReadVersion(long l) {
         underlying.setReadVersion(l);
     }
 
-    @Override
-    public CompletableFuture<Long> getEstimatedRangeSizeBytes(final byte[] begin, final byte[] end) {
```
alecgrieser, line 195: This is technically an unrelated change, but there were duplicate identical implementations of this method in `InstrumentedTransaction` and `InstrumentedReadTransaction`, even though `InstrumentedTransaction extends InstrumentedReadTransaction`, so I consolidated to one place. Could remove or make it a separate change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -199,16 +209,6 @@ public Executor getExecutor() {
         return underlying.getExecutor();
     }
 
-    @Override
-    public CompletableFuture<Long> getEstimatedRangeSizeBytes(final byte[] begin, final byte[] end) {
```
alecgrieser, line 203: This is moving these two methods to I think better align them with where these methods are in the interface.

alecgrieser:
test this please

alecgrieser:
Looks like an unrelated test failure to me in the indexing code. Filed #1267 about it

test this please

scottfines (approved):

ScottDugas (approved):

PR #1231, opened 2021-04-15
adding nested fields to lucene indexing functionality

sophiabuell:
- Nested fields
- hasDuplicates
- plan merging
- expansion of fdb components into complex Lucene strings 
- record cursor fix for primary key component positions
- record cursor handling of nested fields.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -32,10 +32,15 @@ import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TupleRange;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.IndexRecordFunction;
 import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.metadata.MetaDataException;
 import com.apple.foundationdb.record.metadata.expressions.FieldKeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.NestingKeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.ThenKeyExpression;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexableRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecord;
 import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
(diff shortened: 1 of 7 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -150,7 +170,11 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
                         document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
                         document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
                         for (int i = 0; i < fieldNames.size(); i++) {
-                            document.add(new TextField(fieldNames.get(i), entry.getKey().getString(i), Field.Store.NO));
+                            String value = entry.getKey().getString(i);
+                            if (value == null) {
+                                value = "";
+                            }
```
MMcM, line 176: Is this better than not adding it to the document? What are the differences in terms of query behavior? Ignoring 3VL, can't I write `NOT x` to exclude documents where `x` is set?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: thats a good point. I think that would be the preferred behavior yes. I'll write a few tests to check it out.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
```diff
@@ -51,45 +54,143 @@ public LucenePlanner(@Nonnull final RecordMetaData metaData, @Nonnull final Reco
...
+                comparisonString = "NOT %s:";
+                break;
+            case TEXT_CONTAINS_ANY:
+            case TEXT_CONTAINS_PREFIX:
+            default:
+                return null;
+        }
+        if (type != Comparisons.Type.IS_NULL) {
+            if (comparison.getComparand() instanceof String) {
+                comparisonString = comparisonString + "\"%s\"";
```
MMcM, line 85: I believe that there are a number of special characters that need escaping. I'm okay with deferring that to a follow-up.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: I accommodated for this in the planner. If we change it then we can change it there too.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
sophiabuell: Well, yes we'll have to escape some of the Lucene query characters like + and ? and (. For now I think this is fine since the parser will probably disallow it in another location.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -136,9 +136,16 @@
                     if (limitRemaining != Integer.MAX_VALUE) {
                         limitRemaining--;
                     }
+                    Tuple setPrimaryKey = Tuple.fromBytes(pk.bytes);
                     Tuple tuple = Tuple.fromList(fieldNames);
+                    int[] keyPos = new int[setPrimaryKey.size()];
+                    for (int i = 0; i < setPrimaryKey.size(); i++) {
+                        keyPos[i] = i+fieldNames.size();
+                    }
+                    state.index.setPrimaryKeyComponentPositions(keyPos);
```
MMcM, line 145: I think I see how this works, by setting up for a later conversion of this index entry for this index to use the updated positions. But I think the intention was that `getPrimaryKeyComponentPositions` was a fixed piece of the meta-data. Which means that the code here needs to rearrange the order of primary key elements in the tuple, rather than changing the meta-data to match how they are arranged now.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
sophiabuell: So... I tried that and got slightly lost in the logic of trying to match fields that were filled vs fields that weren't and setting the placeholder. I think if I'm correct that its possible to have a larger than size 2 primary key? And that its very possible that the fieldNames don't directly match up with those values that are set since we don't have the __recordID in the field names. I considered this might be something I wanted to give more thought to.

MMcM: I haven't tried myself, but I think if you start with a (mutable) list of field names (as dummies for the indexed text values) and then go through the primary key positions and (if not `-1`) replace the field name _and_ remove from the list to be appended later (or, equivalently, collect just the ones that are `-1`), it'll work.

sophiabuell: I think the issue is that the field names do not necessarily contain the primary key fields. So it's possible that we will end up with that approach with an inaccurate tuple length or replacing a name that doesn't match the name we're trying to place the value into.

sophiabuell: Also what is the probability of the field not being in the right place in the list. You know the  overall code better but it seems this relies on both the fieldNames containing the primary key fields that are set being placed in the right position in the field names list. I'm not entirely sure how assured that is.

MMcM: If the primary key field isn't in the declared fields, it isn't in `getPrimaryKeyComponentPositions`, so you don't replace anything and keep it in the list to be appended.

MMcM (approved):

PR #1233, opened 2021-04-15
Resolve #1232: rebuild index when prev by-index continuation isn't po…

### LINKED ISSUE #1232: Indexing: rebuild index when prev by-index continuation isn't possible
When continuing a previous by-index indexing method which is different than the requested one, and the previous source index is not usable (validation exception), attempt rebuilding the index to avoid a dead end. 
The hidden conjecture is that it’s not likely for a source index to return to availability in a way that would allow a continuation of the partly built index, so rebuilding by the requested method is the preferred solution for the default precondition mode (`ERROR_IF_DISABLED_CONTINUE_IF_WRITE_ONLY`).

jjezra:
…ssible

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -179,69 +179,126 @@ public class OnlineIndexer implements AutoCloseable {
     }
 
     @Nonnull
-    private CompletableFuture<Void> handleIndexerReturnOrFallback(Throwable ex, Supplier<CompletableFuture<Void>> fallbackFunction) {
+    private CompletableFuture<Void> indexingLauncher(Supplier<CompletableFuture<Void>> indexingFunc) {
+        return indexingLauncher(indexingFunc, 0);
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> indexingLauncher(Supplier<CompletableFuture<Void>> indexingFunc, int attemptCount) {
+        return indexingLauncher(indexingFunc, attemptCount, null);
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> indexingLauncher(Supplier<CompletableFuture<Void>> indexingFunc, int attemptCount, @Nullable IndexingPolicy requestedPolicy) {
+        // The launcher calls the indexing function, letting the results be handled by the catcher.
+        // The catcher may, on some cases, call the launcher in its retry path. The attemptCount limits the recursion level as a safety net.
+        return AsyncUtil.composeHandle( indexingFunc.get(),
+                (ignore, ex) -> indexingCatcher(ex, indexingFunc, attemptCount + 1, requestedPolicy));
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> indexingCatcher(Throwable ex, Supplier<CompletableFuture<Void>> indexingFunc, int attemptCount, @Nullable IndexingPolicy requestedPolicy) {
+        // (skeleton function, a little long but broken to distinct cases)
         if (ex == null) {
+            // A happy index it is
             return AsyncUtil.DONE;
         }
-        IndexingBase.PartlyBuiltException partlyBuiltException = IndexingBase.getAPartlyBuildExceptionIfApplicable(ex);
+
+        if (attemptCount > 5) {
+            // Safety net, this should never happen
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn(KeyValueLogMessage.build("Too many indexing attempts",
+                        LogMessageKeys.CURR_ATTEMPT, attemptCount)
+                        .addKeysAndValues(common.indexLogMessageKeyValues())
+                        .toString());
+            }
+            throw FDBExceptions.wrapException(ex);
+        }
+
+        indexer = null; // clear cache - avoid reuse of the same indexer
+        final IndexStatePrecondition indexStatePrecondition = common.getIndexStatePrecondition();
+        final IndexingBase.PartlyBuiltException partlyBuiltException = IndexingBase.getAPartlyBuildExceptionIfApplicable(ex);
         if (partlyBuiltException != null) {
+            // An ongoing indexing process with a different method type was found. Some precondition cases should be handled.
             IndexBuildProto.IndexBuildIndexingStamp conflictingIndexingTypeStamp = partlyBuiltException.savedStamp;
-            // Here: an indexing type stamp already exists, with a different type. Few precondition cases are handled
             if (LOGGER.isInfoEnabled()) {
                 LOGGER.info(KeyValueLogMessage.build("conflicting indexing type stamp",
-                        LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
-                        LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
-                        LogMessageKeys.INDEXER_ID, common.getUuid(),
-                        LogMessageKeys.INDEX_STATE_PRECONDITION, common.getIndexStatePrecondition(),
-                        LogMessageKeys.ACTUAL_TYPE, conflictingIndexingTypeStamp
-                ).toString());
+                        LogMessageKeys.CURR_ATTEMPT, attemptCount,
+                        LogMessageKeys.INDEX_STATE_PRECONDITION, indexStatePrecondition,
+                        LogMessageKeys.ACTUAL_TYPE, conflictingIndexingTypeStamp)
+                        .addKeysAndValues(common.indexLogMessageKeyValues())
+                        .toString());
             }
-            if (common.getIndexStatePrecondition() == IndexStatePrecondition.BUILD_IF_DISABLED_CONTINUE_BUILD_IF_WRITE_ONLY) {
+
+            if (indexStatePrecondition == IndexStatePrecondition.BUILD_IF_DISABLED_CONTINUE_BUILD_IF_WRITE_ONLY) {
+                // Make best effort to finish indexing. Attempt continuation of the previous method
                 indexer = null;
-                // now, match the policy to the previous run
+                // Now, match the policy to the previous run
                 IndexBuildProto.IndexBuildIndexingStamp.Method method = conflictingIndexingTypeStamp.getMethod();
-                if ( method == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -179,69 +179,126 @@ private IndexingByIndex getIndexerByIndex() {
...
                 indexer = null;
-                // now, match the policy to the previous run
+                // Now, match the policy to the previous run
                 IndexBuildProto.IndexBuildIndexingStamp.Method method = conflictingIndexingTypeStamp.getMethod();
-                if ( method == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                if (method == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                    // Partly built by records. The fallback indicator should handle the policy
                     fallbackToRecordsScan = true;
-                    return fallbackFunction.get();
+                    return indexingLauncher(indexingFunc, attemptCount);
```
alecgrieser, line 241: I see you're preserving the `indexingFunc` here and in every method call. If I understand the approach, the idea is that the `indexingFunc` method implementations look at state in the `OnlineIndexer` and react to that. So, for example, if the initial index was by-index, but then it had to fallback to a by-records scan, now calling the `indexerFunc` again which reads that the appropriate boolean has been flipped and that this uses the other indexer. Is that the idea?

Not sure if that should be documented somehow, or the right way to document that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -179,69 +179,126 @@ private IndexingByIndex getIndexerByIndex() {
...
-                final KeyValueLogMessage message =
-                        KeyValueLogMessage.build("fallback to records scan",
-                                LogMessageKeys.INDEX_VERSION, common.getIndex().getLastModifiedVersion(),
-                                LogMessageKeys.INDEXER_ID, common.getUuid(),
-                                LogMessageKeys.INDEX_NAME, index.getName());
-                LOGGER.warn(message.toString(), ex);
+
+        if (indexingPolicy.isByIndex() && IndexingByIndex.isValidationException(ex)) {
+            // Validation failed - the source index cannot be used for records scanning
+            if (requestedPolicy != null) {
```
alecgrieser, line 272: There's something about this that makes it seem like this should only be called if the user used
`BUILD_IF_DISABLED_CONTINUE_BUILD_IF_WRITE_ONLY_REBUILD_IF_POLICY_CHANGED`, insofar as the index changing could arguably be seen as similar to a policy changing.... On the other hand, I think it makes sense for the default behavior to rebuild if the index changes in a way that makes the validation now break. So, I'm not certain.

Perhaps if the `IndexingPolicy` was more expressive (as I think we plan on doing) it would be an option on the build. On the other hand, I'm not sure we want to make an already fairly complicated set of options more complicated.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -179,69 +179,126 @@ private IndexingByIndex getIndexerByIndex() {
...
+    @Nonnull
+    private CompletableFuture<Void> indexingCatcher(Throwable ex, Supplier<CompletableFuture<Void>> indexingFunc, int attemptCount, @Nullable IndexingPolicy requestedPolicy) {
+        // (skeleton function, a little long but broken to distinct cases)
         if (ex == null) {
+            // A happy index it is
             return AsyncUtil.DONE;
         }
-        IndexingBase.PartlyBuiltException partlyBuiltException = IndexingBase.getAPartlyBuildExceptionIfApplicable(ex);
+
+        if (attemptCount > 5) {
```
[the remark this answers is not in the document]
jjezra: I'm not sure it has any meaning for the caller. The max attempts, according to the algorithm, is 3 (attempt to continue previous by-index, attempt to rebuild requested by-index, then fallback to by-records). The `attemptCount` is only a recursion safety net in case of a bug. 
(Coming to think about it, it should probably trigger an error, not a warning).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -179,69 +179,126 @@ private IndexingByIndex getIndexerByIndex() {
...
                     indexingPolicy = IndexingPolicy.newBuilder()
                             .setSourceIndexSubspaceKey(sourceIndexSubspaceKey)
-                            .forbidRecordScan() // no third chance
+                            .setForbidRecordScan(origPolicy.forbidRecordScan)
                             .build();
-                    return fallbackFunction.get();
+                    return indexingLauncher(indexingFunc, attemptCount, origPolicy);
                 }
-                // None of the above? let it fall through
+                // No other methods (yet). This line should never be reached.
```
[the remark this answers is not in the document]
jjezra: Added log error + recovery. 
(Not an exception because it's related to the existing type stamp (of a previous indexing or another issue) and will be only repeated when called again.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Yes, you are right (of course:). 
At the beginning of the recovery code there is the line: 
```  indexer = null; // clear cache - avoid reuse of the same indexer ```
Which enforces the next function call building a new indexer according to the modified parameters. I'll add this to the comment to clarify.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java (hunk 2 above)
Comment (jjezra) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java:272]: I had the same dilemma. My rule of thumb was something like "never rebuild an index unless the caller had explicitly mentioned '_REBUILD_' in his request", but after testing the code in an existing project it seems to lead to a dead-end - as the existing project retry indexing with the same default parameters again and again as the indexing process fails again and again because the previous source-index was modified (or became unreadable). 
Since we can't educate every caller to handle this exception properly, and since a rebuild - when a previous source-index is modified - is unavoidable, doing it "under the hood" seems to be practical and reasonable. I hope that the performance gain of the by-index scan would make the penalty of this occasional rebuilding seem marginal.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (jjezra) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java:207]: (Converted to error)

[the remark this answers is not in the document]
Comment (alecgrieser) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java:207]: Hm, I see. I think it's reasonable not to make it a configuration parameter, then, though I still think it should be a constant defined in the class (i.e., a `private static final` class member).

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java [hunk 5]
```diff
@@ -250,7 +253,17 @@ private IndexingByIndex getIndexerByIndex() {
                             .build();
                     return indexingLauncher(indexingFunc, attemptCount, origPolicy);
                 }
+
                 // No other methods (yet). This line should never be reached.
+                if (LOGGER.isErrorEnabled()) {
+                    LOGGER.error(KeyValueLogMessage.build("Invalid previous indexing type stamp",
```
alecgrieser, line 259: This feels like it is should be a warn-level log if no error is thrown

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -250,7 +253,17 @@ private IndexingByIndex getIndexerByIndex() {
...
+                            LogMessageKeys.CURR_ATTEMPT, attemptCount,
+                            LogMessageKeys.ACTUAL_TYPE, conflictingIndexingTypeStamp)
+                            .addKeysAndValues(common.indexLogMessageKeyValues())
+                            .toString());
+                }
+                common.setIndexStatePrecondition(IndexStatePrecondition.BUILD_IF_DISABLED_REBUILD_IF_WRITE_ONLY);
+                return indexingLauncher(indexingFunc, attemptCount);
```
alecgrieser, line 266: Hmm, I'm trying to understand the logic here, but the basic idea is that if we are trying to continue a previous index build and we don't recognize the index build type, then this will try and rebuild from the beginning?

I'm not sure that we want that, as I think this can only happen if someone makes a logic error in the future and introduces a new indexing method but forgets to update this part of the code. Which is to say, that there's a possibility we get into a weird loop where whenever we resume a build with that new method, we end up losing all progress and starting over.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: Changed to: 
```if (attemptCount > INDEXING_ATTEMPTS_RECURSION_LIMIT)```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I see. I was trying to address the possibility of a damaged type stamp. But comment makes sense. (`parseTypeStampOrThrow`, in indexingBase, is likely to throw an error in this case). 
Changed to throwing an exception.

Issue #1248, opened 2021-04-22
Switch from bintray to artifactory for artifact publishing


alecgrieser:
With bintray being sunset on May 1, 2021, we need to put the artifacts somewhere else. We should add support for artifactory as that seems to be the most straightforward replacement technology.

alecgrieser:
Looks like we're publishing the wrong shadow jar after all ☹️

PR #1249, opened 2021-04-22
Adds artifactory publishing logic

alecgrieser:
This adds the artifactory plugin and switches build.py to start using it. This is to try and make progress on #1248, though I don't necessarily think this is going to be the last PR for that, so I'm not marking this as resolving the issue (yet).

### DIFF
--- a/gradle/publishing.gradle
+++ b/gradle/publishing.gradle
@@ -18,6 +18,7 @@
  * limitations under the License.
  */
 
+apply plugin: 'com.jfrog.artifactory'
 apply plugin: 'com.jfrog.bintray'
 
 // Add various details to the pom file to allow for publishing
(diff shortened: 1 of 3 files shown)


gradle/publishing.gradle
```diff
@@ -62,8 +63,8 @@ publishing {
         }
     }
     def publishBuild = Boolean.parseBoolean(findProperty('publishBuild') ?: 'false')
+    def isReleaseBuild = Boolean.parseBoolean(findProperty('releaseBuild') ?: 'false')
```
sophiabuell, line 66: did we need to pull this out?

alecgrieser: No, I guess not. I think I'd done that during part of the refactoring for this, but I don't think it's actually needed (at least not yet). At some point, we may want to so that we could use a different `repoKey` for snapshot and release builds, but not at the moment

sophiabuell: ok well its not a big deal we can keep it here.

sophiabuell (approved):

alecgrieser:
btw, in terms of testing, I haven't been able to run this with working credentials, but with fake credentials, I can get it locally to try and upload and then fail (due to credential errors), which I think is the best I can do

gradle/publishing.gradle
alecgrieser: Reverted in c1822fc

PR #1253, opened 2021-04-23
Resolves #1248: Switch from bintray to artifactory for artifact publishing

### LINKED ISSUE #1248: Switch from bintray to artifactory for artifact publishing

alecgrieser:
This removes the bintray publishing logic in favor of the Artifactory publishing that was added in some previous PRs (namely #1249 and #1252). This is in anticipation of Bintray being ended on May 1, 2021.

This resolves #1248.

### DIFF
--- a/fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
+++ b/fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
@@ -102,7 +102,4 @@ publishing {
             artifact tasks.shadedJavadocJar
         }
     }
-    bintray {
-        publications = ["shadow"]
-    }
 }
--- a/gradle/publishing.gradle
+++ b/gradle/publishing.gradle
@@ -19,7 +19,6 @@
  */
 
 apply plugin: 'com.jfrog.artifactory'
-apply plugin: 'com.jfrog.bintray'
 
 // Add various details to the pom file to allow for publishing
 def addPublishingInfo(publication) {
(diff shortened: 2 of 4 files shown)


fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
```diff
@@ -102,7 +102,4 @@ publishing {
             artifact tasks.shadedJavadocJar
         }
     }
-    bintray {
-        publications = ["shadow"]
```
alecgrieser, line 106: It's possible we'll need to add an `artifactoryPublish` section here that specifies it should use the `shadow` publication. The artifacts for 2.10.178.0 look right to me, so I'm hesitant to change this until we know it's a problem, though: https://ossartifacts.jfrog.io/artifactory/fdb-record-layer/org/foundationdb/fdb-record-layer-core-shaded/2.10.178.0/

gradle/publishing.gradle
```diff
@@ -62,32 +61,9 @@ publishing {
...
-                vcsUrl = "https://github.com/FoundationDB/fdb-record-layer.git"
-            }
-            override = !isReleaseBuild
-            publish = publishBuild
-        }
-    }
 
     artifactoryPublish {
+        def publishBuild = Boolean.parseBoolean(findProperty('publishBuild') ?: 'false')
         skip = !publishBuild
```
sophiabuell, line 67: what are we actually doing with this? It looks like publications is running whether or not we set it.

alecgrieser: The idea is that as a safe-guard, one should need to add `-PpublishBuild=true` to the gradle command to actually start sending things to artifactory. How useful this is as a safe-guard is somewhat questionable, I suppose, though I was mostly trying to maintain compatibility with the old system unless there was a reason to change it (`mutatis mutandis` and all that).

That being said, there also seems to be a bug with skip, I think. It's supposed to not run the publishing (see: https://gist.github.com/DALDEI/40925568cef57e16e8a7), so I added an an additional condition

sophiabuell: thanks!

sophiabuell (approved):

PR #1256, opened 2021-04-26
Resolves #1255: RL uses of FDBDatabaseRunner should include logging d…

### LINKED ISSUE #1255: Record Layer uses of FDBDatabaseRunner.runAsync should include additional logging details
Right now lambdas that are executed through `FDBDatabaseRunner.runAsync` within Record Layer  that encounter retryable exceptions will, largely, just emit stack traces from the runner that include no useful information as to what work was being done.  We should audit Record Layer's use of this call, and ensure that `additionalLoggingProperties` are passed in to give some context.

scgray:
…etails

This was spurred in debugging some conflicts seen from the LocatableResolvers
and reverse directory layer, so there are also some updated changes to errors
and log messages there

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
@@ -22,7 +22,9 @@ package com.apple.foundationdb.record.provider.foundationdb;
 
 import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.system.SystemKeyspace;
+import com.google.common.collect.Lists;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -67,7 +69,7 @@ public class FDBSystemOperations {
             final Transaction tr = context.ensureActive();
             tr.options().setReadSystemKeys();
             return tr.get(SystemKeyspace.PRIMARY_DATACENTER_KEY).thenApply(FDBSystemOperations::nullableUtf8);
-        });
+        }, Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getPrimaryDatacenterAsync"));
     }
 
     /**
@@ -107,7 +109,8 @@ public class FDBSystemOperations {
      */
     @Nonnull
     public static CompletableFuture<String> getConnectionStringAsync(@Nonnull FDBDatabaseRunner runner) {
-        return runner.runAsync(FDBSystemOperations::getConnectionStringAsyncInternal);
+        return runner.runAsync(FDBSystemOperations::getConnectionStringAsyncInternal,
+                Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getConnectionStringAsync"));
     }
 
     /**
@@ -140,7 +143,8 @@ public class FDBSystemOperations {
      */
     @Nonnull
     public static CompletableFuture<String> getClusterFilePathAsync(@Nonnull FDBDatabaseRunner runner) {
-        return runner.runAsync(FDBSystemOperations::getClusterFilePathAsyncInternal);
+        return runner.runAsync(FDBSystemOperations::getClusterFilePathAsyncInternal,
+                Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getClusterFilePathAsync"));
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
@@ -24,11 +24,11 @@ import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.system.SystemKeyspace;
-import com.google.common.collect.Lists;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -69,7 +69,7 @@ public class FDBSystemOperations {
             final Transaction tr = context.ensureActive();
             tr.options().setReadSystemKeys();
             return tr.get(SystemKeyspace.PRIMARY_DATACENTER_KEY).thenApply(FDBSystemOperations::nullableUtf8);
-        }, Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getPrimaryDatacenterAsync"));
+        }, Arrays.asList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getPrimaryDatacenterAsync"));
     }
 
     /**
@@ -110,7 +110,7 @@ public class FDBSystemOperations {
     @Nonnull
     public static CompletableFuture<String> getConnectionStringAsync(@Nonnull FDBDatabaseRunner runner) {
         return runner.runAsync(FDBSystemOperations::getConnectionStringAsyncInternal,
-                Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getConnectionStringAsync"));
+                Arrays.asList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getConnectionStringAsync"));
     }
 
     /**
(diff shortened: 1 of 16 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBSystemOperations.java
```diff
@@ -67,7 +69,7 @@ private static String nullableUtf8(@Nullable byte[] bytes) {
             final Transaction tr = context.ensureActive();
             tr.options().setReadSystemKeys();
             return tr.get(SystemKeyspace.PRIMARY_DATACENTER_KEY).thenApply(FDBSystemOperations::nullableUtf8);
-        });
+        }, Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "FDBSystemOperations::getPrimaryDatacenterAsync"));
```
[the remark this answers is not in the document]
jjezra: (A few Ditto's below..)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -613,6 +617,13 @@ public static PartlyBuiltException getAPartlyBuildExceptionIfApplicable(@Nullabl
         return null;
     }
 
+    protected List<Object> runAsyncDetails(String transactionName, Object ...keysAndValues) {
+        return new ImmutableList.Builder<>()
+                .add(LogMessageKeys.TRANSACTION_NAME, transactionName)
+                .add(common.indexLogMessageKeyValues())
```
jjezra, line 623: Instead of the wrapping function `runAsyncDetails`, wouldn't it make sense to use `common.indexLogMessageKeyValues(String transactionName)`? 
(If `Object ...keysAndValues` is never used, maybe we can wait with the implementation until it's needed)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1990,6 +1992,14 @@ public static OnlineIndexer forRecordStoreAndIndex(@Nonnull FDBRecordStore recor
         return newBuilder().setRecordStore(recordStore).setIndex(index).build();
     }
 
+    protected List<Object> runAsyncDetails(String transactionName, Object ...keysAndValues) {
```
jjezra, line 1995: (having `common.indexLogMessageKeyValues(String transactionName)` would also save the need for this double implementation)

jjezra (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
```diff
@@ -624,7 +625,10 @@ private TupleRange computeRecordsRange() {
             RecordCursor<Tuple> cursor = common.getRecordStoreBuilder().copyBuilder().setContext(context).open()
                     .getPrimaryKeyBoundaries(tupleRange.getLow(), tupleRange.getHigh());
             return context.asyncToSync(FDBStoreTimer.Waits.WAIT_GET_BOUNDARY, cursor.asList());
-        });
+        }, Lists.newArrayList(
+                LogMessageKeys.TRANSACTION_NAME, "IndexingByRecords::getPrimaryKeyBoundaries",
```
alecgrieser, line 629: Looking at how essentially all of these runner creation thingies want to log the transaction name, there's another part of me that wonders if the right approach is to add `name` to the `FDBRecordContextConfig` and then log that in `FDBDatabaseRunnerImpl` (and probably expose it in the `FDBDatabaseRunner` somehow). We'd need to make sure to qualify what makes it different from a transaction's ID (i.e., descriptive name to help the user differentiate different sources of transactions vs unique ID to allow different transaction's to be distinguished in the logs), but it could be more generally useful.

That being said, I think that adding the general purpose "additional key values" is probably also good, so I think we could take this PR and then add names to `FDBRecordContextConfig`s (later).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
```diff
@@ -138,7 +140,10 @@ public static ScopedInterningLayer global(@Nonnull FDBDatabase database) {
 
     @Override
     public CompletableFuture<Void> setWindow(long count) {
-        return database.runAsync(context -> interningLayerFuture.thenCompose(layer -> layer.setWindow(context, count)));
+        return database.runAsync(context -> interningLayerFuture.thenCompose(layer -> layer.setWindow(context, count)),
+                Lists.newArrayList(
+                        LogMessageKeys.TRANSACTION_NAME, "ScopedInterningLayer::setWindow",
+                        LogMessageKeys.RESOLVER));
```
alecgrieser, line 146: This lone `RESOLVER` key is also sus

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/synchronizedsession/SynchronizedSessionRunner.java
```diff
@@ -154,7 +160,9 @@ public UUID getSessionId() {
     public CompletableFuture<Void> endSessionAsync() {
         // Using the underlying runner rather than itself because it should not throw any exception if the session
         // does not hold the lock.
-        return underlying.runAsync(context -> session.releaseLock(context.ensureActive()));
+        return underlying.runAsync(context -> session.releaseLock(context.ensureActive()),
+                Lists.newArrayList(LogMessageKeys.TRANSACTION_NAME, "SynchronizedSessionRunner::endSession",
+                        LogMessageKeys.SESSION_ID, session.getSessionId()));
```
alecgrieser, line 165: Should the `SESSION_ID` get added to the user-supplied list of additional log message keys in this runner's implementation of `runAsync` and friends?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
scgray: Yeah, I was thinking something very similar to this.  The main argument against that approach might be that the runner can be created once, then used many times, and each use might logically want a different name.  Either way, I'm not really pleased with the solution as it is now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
scgray: Yep, agreed. Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ExtendedDirectoryLayer.java
```diff
@@ -216,7 +218,10 @@ public static ExtendedDirectoryLayer global(@Nonnull FDBDatabase database) {
 
     @Override
     public CompletableFuture<Void> setWindow(long count) {
-        return database.runAsync(context -> getHca(context).thenAccept(hca -> hca.setWindow(count)));
+        return database.runAsync(context -> getHca(context).thenAccept(hca -> hca.setWindow(count)),
+                Lists.newArrayList(
+                        LogMessageKeys.TRANSACTION_NAME, "ExtendedDirectoryLayer::setWindow",
+                        LogMessageKeys.RESOLVER));
```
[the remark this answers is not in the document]
scgray: Doh! Nice catch.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
scgray: Gack! Twice, no less! Thanks.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/synchronizedsession/SynchronizedSessionRunner.java
scgray: Good thinking. Done.

jjezra (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
```diff
@@ -143,7 +143,7 @@ public static ScopedInterningLayer global(@Nonnull FDBDatabase database) {
         return database.runAsync(context -> interningLayerFuture.thenCompose(layer -> layer.setWindow(context, count)),
                 Lists.newArrayList(
```
alecgrieser, line 144: minor: In theory, this should also be changed to `Arrays.asList` like the other ones that were just changed.

alecgrieser (approved):
Leaving one comment for one minor thing, but I think we could merge without it if we wanted to, so approving and leaving for you to merge (or squash and merge).

It also sounds like it might be good idea to file a follow up issue about improving the API here based on the comments about how to pass the logging information and also plausibly about adding the transaction name to the runner configuration itself, both of which seem like things we vaguely wish we had, though I think it's not worth waiting for them before taking this PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
scgray: I did the minor fix. I was also wrong about the runner, it does log a consistent error of `Retrying FDB Exception`. Maybe not the clearest message in the world, but it is consistent.

PR #1259, opened 2021-04-27
Fixes #1258: FDBReverseDirectoryCache::putIfExists can throw if using a shared read version

### LINKED ISSUE #1258: FDBReverseDirectoryCache::putIfExists can throw if using a shared read version
If using a `LocatableResolver` with the reverse directory cache, then the `putIfExists` method can lead to permanent allocation failures if one combines that with sharing a read version.

In particular, the chain of events is the following:

1. A value gets allocated by a resolver, let's say the name "foo" gets allocated the value "1"
2. That value gets put into the in-memory reverse directory cache (let's say by doing a reverse lookup)
3. A second name, let's say "bar", goes to the same resolver to get a key allocated. If it's using a new read version, the resolver will always pick a new number, but if it's using an older read version (less than the commit version of the "foo" allocation), it might _also_ allocate the same value as first allocation. This will fail with an error, "Provided value for path key does not match value found in cache", as it compares the (committed) allocation of ("foo" -> 1) with the uncommitted allocation ("bar" -> 1)

If the check were absent in `putIfExists`, then the transaction would have failed to commit anyway, then the resolver would have retried the allocation with a newer transaction that would have seen the effect of the first allocation. So, it should be safe to transform the exception into a retriable one and let existing retry loops handle this case.

alecgrieser:
This adds a test that simulates a problem that was seen with the reverse directory cache. Essentially, if two transactions picked the same value for a name (with the same scoped), then we could fail that operation with a non-retriable error due to some checks that looked at an in-memory cache. The proposed fix here is to make the exception retriable. I added a test that simulates creating a bunch of entries all with the same transaction, and so probabilistically, some of them should allocate to the same value.

This fixes #1258.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -18,7 +18,7 @@ In this realase, the various implementations of the `RecordQueryPlan` interface 
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Uses of FDBDatabaseRunner include logging details for exception logging [(Issue #1255)](https://github.com/FoundationDB/fdb-record-layer/issues/1255)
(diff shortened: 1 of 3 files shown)


alecgrieser:
Rebase addresses merge conflicts (and adds a release note I forgot to add earlier whoops)

scgray (approved):

PR #1202, opened 2021-03-23
Upgrade RecordLayer to use FDB 6.3.12

scottfines:
Addresses Issue #1201

PR #1260, opened 2021-04-27
Add multi-client settings to FDBDatabaseFactory

scottfines:
This just exposes the setting so that we can pass it through the CloudKit.

Issue #642, opened 2019-06-25
FDBDatabaseFactory is ambiguously a singleton


alecgrieser:
The docs for the `FDBDatabaseFactory` calls it a singleton and it has a `instance()` method, but it also has a public constructor, so anyone can call `new FDBDatabaseFactory()`.

We should somehow align these things. It would be a breaking change to make the constructor on that class private, but it could be done in stages. We could also make a `FDBDatabaseFactory` officially not a singleton, which actually has some advantages in that there is some state that one might not want to configure differently on different factories (like retry policy).

PR #1264, opened 2021-04-28
Enable Multi-threaded client for RecordLayer 3

scottfines:
Record Layer 3.0 version of PR #1202

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
@@ -77,6 +77,9 @@ public class FDBDatabaseFactory {
 
     @Nonnull
     private static final FDBDatabaseFactory INSTANCE = new FDBDatabaseFactory();
+    // when set to true, static options have been set on the FDB instance.
+    //made volatile because multiple FDBDatabaseFactory instances can be created technically, and thus can be racy during init
+    private static volatile boolean staticOptionsSet = false;
 
     @Nonnull
     private FDBLocalityProvider localityProvider = FDBLocalityUtil.instance();
@@ -116,6 +119,9 @@ public class FDBDatabaseFactory {
     private long transactionTimeoutMillis = DEFAULT_TR_TIMEOUT_MILLIS;
     private boolean runLoopProfilingEnabled;
 
+    //made volatile because multiple FDBDatabaseFactory instances can be created technically, and thus can be racy during init
+    private static volatile int threadsPerClientVersion = 1; //default is 1, which is basically disabled
+
     /**
      * The default is a log-based predicate, which can also be used to enable tracing on a more granular level
      * (such as by request) using {@link #setTransactionIsTracedSupplier(Supplier)}.
@@ -146,6 +152,7 @@ public class FDBDatabaseFactory {
             }
             fdb = FDB.selectAPIVersion(API_VERSION);
             fdb.setUnclosedWarning(unclosedWarning);
+            setStaticOptions(fdb);
             NetworkOptions options = fdb.options();
             if (!traceFormat.isDefaultValue()) {
                 options.setTraceFormat(traceFormat.getOptionValue());
@@ -167,6 +174,25 @@ public class FDBDatabaseFactory {
             inited = true;
         }
         return fdb;
+    }
+
+    private static synchronized void setStaticOptions(final FDB fdb) {
+        /*
+         * There are a few FDB settings that have to be set statically, but also need to have room
+         * for configuration. For the most part, FDBDatabaseFactory is a singleton and so in _theory_ this shouldn't
+         * matter. However, in practice it is possible to create multiple factories(i.e. in test code and such),
+         * and doing so may cause problems with these settings (errors thrown, that kind of thing). To avoid that,
+         * we have to follow a somewhat goofy pattern of making those settings static, and checking to ensure that
+         * we only set those options once. This block of code does that.
+         *
+         * Note that this method is synchronized on the class; this is so that multiple concurrent attempts to
+         * init an FDBDatabaseFactory won't cause this function to fail halfway through.
+         */
+        if (!staticOptionsSet) {
+            fdb.options().setClientThreadsPerVersion(threadsPerClientVersion);
+
+            staticOptionsSet = true;
+        }
     }
 
     @Nonnull
@@ -364,6 +390,26 @@ public class FDBDatabaseFactory {
             throw new RecordCoreException("run loop profiling can not be enabled as the client has already started");
         }
         this.runLoopProfilingEnabled = runLoopProfilingEnabled;
+    }
+
+    /**
+     * Set the number of threads per FDB client version. The default value is 1.
+     *
+     * @param threadsPerClientV the number of threads per client version. Cannot be less than 1
+     */
+    public static void setThreadsPerClientVersion(int threadsPerClientV) {
+        if (staticOptionsSet) {
+            throw new RecordCoreException("threads per client version cannot be changed as the version has already been initiated");
+        }
+        if (threadsPerClientV < 1) {
+            //if the thread count is too low, disable the setting
+            threadsPerClientV = 1;
+        }
+        threadsPerClientVersion = threadsPerClientV;
+    }
+
+    public static int getThreadsPerClientVersion() {
+        return threadsPerClientVersion;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


scottfines:
test this please

scottfines:
This also appears to hit a spurious Test timeout error(Issue #1267 )

scottfines:
test this please

scottfines:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -115,6 +116,7 @@
     private long stateRefreshTimeMillis = TimeUnit.SECONDS.toMillis(FDBDatabase.DEFAULT_RESOLVER_STATE_CACHE_REFRESH_SECONDS);
     private long transactionTimeoutMillis = DEFAULT_TR_TIMEOUT_MILLIS;
     private boolean runLoopProfilingEnabled;
+    private static volatile int threadsPerClientVersion = 1; //default is 1, which is basically disabled
```
alecgrieser, line 119: I'm somewhat of two minds about making this `static`. The `FDBDatabaseFactory` is supposed to be a singleton, so in theory, this shouldn't need to be `static`. This new parameter also isn't the only configuration parameter that doesn't really make sense to set twice--"run loop profiling", for example, which has a guard on the class being `inited`.

I _think_ we want this to be not-static and for it to be more like the other configuration parameters on the factory, including `runLoopProfilingEnabled`. But it's a bit awkward, especially if someone does create more than one of these `FDBDatabaseFactory` objects, which they're not supposed to do, but also #642.

I also suppose it doesn't hurt, but I think that making it `volatile` is also overkill, as all modifications happen in `synchronized` methods, so the thread cache should never have a stale value (I think).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -366,6 +378,26 @@ public synchronized void setRunLoopProfilingEnabled(boolean runLoopProfilingEnab
         this.runLoopProfilingEnabled = runLoopProfilingEnabled;
     }
 
+    /**
+     * Set the number of threads per FDB client version. The default value is 1.
+     *
+     * @param tPerClientV the number of threads per client version. Cannot be less than 1
+     */
+    public static synchronized void setThreadsPerClientVersion(int tPerClientV) {
```
alecgrieser, line 386: minor: I think it might be a little clearer if the parameter was `threadsPerClientVersion` instead of `tPerClientV`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -366,6 +378,26 @@ public synchronized void setRunLoopProfilingEnabled(boolean runLoopProfilingEnab
...
+            throw new RecordCoreException("threads per client version cannot be changed as the version has already been initiated");
+        }
+        if (tPerClientV < 1) {
+            //if the thread count is too low, disable the setting
+            tPerClientV = 1;
+        }
+        threadsPerClientVersion = tPerClientV;
+    }
+
+    public synchronized int getThreadsPerClientVersion(int threadsPerClientVersion) {
```
alecgrieser, line 397:
```suggestion
    public synchronized int getThreadsPerClientVersion() {
```

I'm also not sure we want/need this to be synchronized, given that it's a read-only operation

alecgrieser (changes requested):
Some minor things, but then a more involved comment about how much we want to protect ourselves from setting this twice, which it sounded like you also had to think over. It's kind of a bit unfortunate, at the moment, because there's a fair number of configuration parameters to FDB that must be set before the network is initialized, and I think the current solution uses state in the `FDBDatabaseFactory` singleton, which should be enough to protect us unless something makes multiple copies of that object while keeping the FDB C client alive (which I think can happen sometimes--usually by frameworks?).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: On the whole, I 1000% agree; static in this world feels pretty craptastic. But from a practical standpoint, making it _non-static_ has some weird problems. The scenario I ran into when I first wrote this up (using a non-static variable, a la PR #1260 ) was that creating two `FDBDatabaseFactory` instances would result in triggering an FDB-layer exception; there did not appear to be a good way of using an instance variable _and_ simultaneously telling the JVM to only use that one instance variable a single time. 

This static approach doesn't feel good to me--in particular, relying on `initDB` to not race between two instances seems iffy; what happens if you want to init two different versions in the same JVM? Is this static approach going to work? 

But in practice, it seems to be the only way around the immediate problem in my mind (unless I'm missing something, which would be great)

scottfines: Regarding the `volatile`, the idea was that it wasn't necessary to synchronize, but then I evolved to synchronize anyway. Looking over that, it looks like it's being synchronized over two different variables, which is bad. I'll fix and remove the `volatile`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: if `threadsPerClientVersion` isn't synchronized, then this has to be (otherwise you might never actually see the changed number)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: changed to `threadsPerClientV`, which is the best I could do with a static field named `threadsPerClientVersion`, sadly. But more clear, at least

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
alecgrieser: Right. In principle, `FDBDatabaseFactory` is supposed to be a singleton, though, so you should never have two of them, unless something (JUnit?) is creating more than one of these behind the scenes, which I think I've seen in the past.

It's definitely a little strange that it's static, but maybe just noting why it's static in a comment so that we know that it's because FDB hits a weird exception if you set this twice would be sufficient enough.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
alecgrieser: Why not shadow the variable and then set it in the method with `FDBDatabaseFactory.threadsPerClientVersion = threadsPerClientVersion`?

alecgrieser (changes requested):
It looks like there are a few things left outstanding from the previous review that could still be addressed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: multiple FDBDatabaseFactory instances are created during the RecordLayer testing code, unfortunately.

I added a comment to the `setStaticOptions` method to explain what it's doing, and why.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: I am not really _worried_, but it's good design to make sure that your concurrency model works. I too prefer `volatile`, but was attempting to be responsive to earlier comments.

I reverted to `volatile` and further clarified the concurrency model.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -366,6 +392,26 @@ public synchronized void setRunLoopProfilingEnabled(boolean runLoopProfilingEnab
...
+            throw new RecordCoreException("threads per client version cannot be changed as the version has already been initiated");
+        }
+        if (threadsPerClientV < 1) {
+            //if the thread count is too low, disable the setting
+            threadsPerClientV = 1;
+        }
+        threadsPerClientVersion = threadsPerClientV;
+    }
+
+    public static int getThreadsPerClientVersion(int threadsPerClientVersion) {
```
alecgrieser, line 411:
```suggestion
    public static int getThreadsPerClientVersion() {
```

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scottfines: whoops--that was pretty dumb, thanks.

Fixed

alecgrieser (approved):

PR #1272, opened 2021-05-04
Fixes #1271: introduce a variant of reduce with a stop condition

### LINKED ISSUE #1271: Allow notifying `RecordCursor::reduce` that a terminal condition has been reached
When reducing over large range scans with continuation logic, it can be convenient to supply a predicate that is checked and returns the reduced value once reached.

The proposed approach would be an additional arity for `RecordCursor::reduce` which accepts a `stopCondition` predicate.
It is not yet clear to me whether `RecordCursor::close` should be called on the cursor when the stop condition has been reached, happy to hear everyone's thoughts on that.

pyr:
Introduce `RecordCursor::reduce` tests at the same time.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
@@ -58,6 +58,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  * An asynchronous iterator that supports continuations.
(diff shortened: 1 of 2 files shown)


jleach4 (approved):

MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
```diff
@@ -952,4 +953,25 @@ public boolean isLimitReached() {
...
+     * @param accumulator function that takes previous reduces value and computes new value by combining with each record
+     * @param stopCondition predicate which runs against the accumulated result before fetching new records
+     * @param <U> the result type of the reduction
+     * @return a future that completes to the result of reduction
+     */
+    @Nullable
+    default <U> CompletableFuture<U> reduce(U identity, BiFunction<U, ? super T, U> accumulator, Predicate<U> stopCondition) {
+        final AtomicReference<U> holder = new AtomicReference<>(identity);
+        return AsyncUtil.whileTrue(() -> onNext().thenApply(result -> {
+            U nextResult = accumulator.apply(holder.get(), result.get());
```
MMcM, line 970: This will get an error (`Tried to call get() on a RecordCoreResult that did not have a next value.`) if the result doesn't have a next.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1180,4 +1180,15 @@ public void hasNextErrorStack() throws Exception {
...
     }
 
+    @Test
+    public void reduceTest() throws Exception {
+        RecordCursor<Integer> cursor = RecordCursor.fromList(Arrays.asList(1, 2, 3, 4, 5, 6, 7));
+        CompletableFuture<Integer> sum = cursor.reduce(0, Integer::sum);
+        assertEquals(28, sum.get());
+
+        cursor = RecordCursor.fromList(Arrays.asList(1, 2, 3, 4, 5, 6, 7));
+        sum = cursor.reduce(0, Integer::sum, x -> x > 10);
```
MMcM, line 1190: Need a variant of this test where the stop condition is never met and it goes to the end. I am pretty sure that would uncover the problem mentioned above.

MMcM (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
pyr: Done, thanks!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
pyr: Indeed, this was a rough translation of our in-house Clojure version and got things mixed up- fixed in latest commit

MMcM:
test this please

MMcM (approved):

Issue #1267, opened 2021-04-29
OnlineIndexerIndexFromIndexTest.testIndexFromIndexPersistentContinueRebuildWhenTypeStampChange timeout failures


alecgrieser:
I've seen at least once during a PRB run that the test `OnlineIndexerIndexFromIndexTest.testIndexFromIndexPersistentContinueRebuildWhenTypeStampChange` has failed. The error message isn't particularly helpful, with it essentially saying that there's some kind of timeout. Here's the stack trace:

```
com.apple.foundationdb.record.RecordCoreException
	at com.apple.foundationdb.record.provider.foundationdb.FDBExceptions.wrapException(FDBExceptions.java:200)
	at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.lambda$new$0(FDBDatabase.java:167)
	at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.asyncToSync(FDBDatabase.java:1037)
	at com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunnerImpl.asyncToSync(FDBDatabaseRunnerImpl.java:340)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexer.asyncToSync(OnlineIndexer.java:790)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexer.buildIndex(OnlineIndexer.java:714)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerIndexFromIndexTest.buildSrcIndex(OnlineIndexerIndexFromIndexTest.java:71)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerIndexFromIndexTest.testIndexFromIndexPersistentContinueRebuildWhenTypeStampChange(OnlineIndexerIndexFromIndexTest.java:595)
Caused by: java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.asyncToSync(FDBDatabase.java:1028)
	... 95 more
```

Unclear if this is really the test or if the CI environment can get overwhelmed, though

alecgrieser:
It's a little concerning that the error message appears not to have a real message associated with it ¯\\\_(ツ)_/¯

[...]

PR #1282, opened 2021-05-06
Fixes #1281: AutoService missing from Record Layer runtime dependencies

### LINKED ISSUE #1281: AutoService missing from Record Layer runtime dependencies
Due to https://github.com/google/auto/pull/797, the `AutoService` annotation that we've been using is now required at runtime in addition to at compile time because the retention level is now "CLASS" instead of "SOURCE". This is to allow for incremental Gradle builds, but it does mean that users of the library will get compiler warnings and/or errors until this is fixed. It also adds a potentially conflicting transitive dependency to our downstream adopters, but I'm not sure there's a good solution to that unless we can get this fixed.

alecgrieser:
This reverts our `AutoService` dependency to 1.0-rc6, which was the previous version. At that version, the `AutoService` annotation is retained at `SOURCE` only, which means that it is not included in the generated `.class` files, so our downstream dependencies don't need to worry about what version of `AutoService` we're using or even that we're using `AutoService` at all (and are free to use their own version or an alternative mechanism for adding to the service registry entirely). However, with 1.0-rc7, the annotation became a `CLASS` retained object, which means that it would need to start being something our downstream dependencies would need to worry about. To avoid putting that onto our dependencies, this just reverts the upgrade. I'm not excited about us being stuck on this older version, but I think it's more expedient, at least until we can get Java 11 modules all worked out.

This fixes #1281.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <resourceExtensions>
-      <entry name=".+\\.(properties|xml|html|dtd|tld)" />
-      <entry name=".+\\.(gif|png|jpeg|jpg)" />
-    </resourceExtensions>
     <wildcardResourcePatterns>
       <entry name="?*.properties" />
       <entry name="?*.xml" />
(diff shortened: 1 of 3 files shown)


MMcM (approved):

alecgrieser:
Looks like PRB failed due to #1267 

test this please

PR #1287, opened 2021-05-20
Resolves #1286: IN join with Union instead of concat

### LINKED ISSUE #1286: IN join with Union instead of concat
At execute time, take the values and make each into a Cursor that gets merged

MMcM:
Not supported by new planner yet.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -75,6 +75,7 @@ import com.apple.foundationdb.record.query.plan.plans.RecordQueryTypeFilterPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnionPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedPrimaryKeyDistinctPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedUnionPlan;
+import com.apple.foundationdb.record.query.plan.temp.PlanContext;
 import com.apple.foundationdb.record.query.plan.temp.explain.PlannerGraphProperty;
 import com.apple.foundationdb.record.query.plan.temp.properties.FieldWithComparisonCountProperty;
 import com.apple.foundationdb.record.query.plan.visitor.FilterVisitor;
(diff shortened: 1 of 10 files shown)


MMcM:
test this please

MMcM:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -432,28 +433,64 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+            if (needOrdering) {
+                scoredPlan.planOrderingKey = inExtractor.adjustOrdering(bestPlan.planOrderingKey, false);
+            }
+            return scoredPlan;
+        }
+        return null;
+    }
+
+    private ScoredPlan planFilterWithInUnion(@Nonnull PlanContext planContext, @Nonnull InExtractor inExtractor) {
+        final ScoredPlan scoredPlan = planFilterForInJoin(planContext, inExtractor.subFilter(), true);
```
normen662, line 478: what exactly does passing true here cause? Does the logic make sure that the query's sort requirement is applied (presumably looking at the passed in planContext?)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/InExtractor.java
```diff
@@ -225,7 +232,7 @@ public RecordQueryPlan wrap(RecordQueryPlan plan) {
     }
 
     @Nullable
-    public PlanOrderingKey adjustOrdering(@Nullable PlanOrderingKey ordering) {
+    public PlanOrderingKey adjustOrdering(@Nullable PlanOrderingKey ordering, boolean asPrefix) {
```
normen662, line 235: Can you please add some Java doc here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionPlan.java
```diff
@@ -0,0 +1,419 @@
...
+                        NodeInfo.IN_UNION_OPERATOR,
+                        valuesSources.stream().map(Object::toString).collect(Collectors.toList()),
+                        ImmutableMap.of()),
+                childGraphs);
+    }
+
+    @Nonnull
+    @Override
+    public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
+        return ImmutableSet.of();
```
normen662, line 178: It's okay to leave as is but this is not correct I think. won't trip up anybody right now since that logic is only used in the new planner.

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: `needOrdering` means that the method should populate the best plan's `planOrderingKey`.

The plan honors the query's sort requirement in both cases.

ohadzeliger (approved):

PR #1293, opened 2021-06-02
Resolve #1266 - Add indexing validation & repair

### LINKED ISSUE #1266: Add indexing validation & repair
We currently have Index Validation which is a read only operation. We should have a mechanism that can scan readable indexes and detect errors - and also fix them. 
This mechanism should also support scrubbing - if called periodically to perform a quota, remember the last checked location (if exists) and continue from there.

jjezra:


jjezra:
This is weird. All the tests passed on my system, and clicking on the failure's 'Details' leads to a bad server (fdb-record-layer-ci.foundationdb.org ?). 
Test this please.

jjezra:
Test this please

jjezra:
Test this please.

jjezra:
Test this please.

jjezra:
org.junit.jupiter.engine.extension.TimeoutExtension.intercept - each time different place. 
Test this please.

jjezra:
Test this please

jjezra:
Rebased.

jjezra:
Test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+    CompletableFuture<Void> buildIndexInternalAsync() {
+        return getRunner().runAsync(context -> openRecordStore(context)
+                .thenCompose( store ->
+                        context.getReadVersionAsync()
+                                .thenCompose(vignore -> {
+                                    SubspaceProvider subspaceProvider = common.getRecordStoreBuilder().getSubspaceProvider();
+                                    return subspaceProvider.getSubspaceAsync(context)
+                                            .thenCompose(subspace ->
+                                                    scrubIndex(subspaceProvider, subspace, null, null)
+                                                    .thenCompose(ignore -> scrubRecords(subspaceProvider, subspace, null, null))
```
alecgrieser, line 123: It's a little weird that the one scrub scrubs both the index and the records. Is that something we're planning on changing in a future change?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+            final byte[] keyBytes = store.indexSubspace(common.getIndex()).pack(valueKey);
+
+            if (LOGGER.isWarnEnabled() && spolicy().shouldReportDangling()) {
+                LOGGER.warn(KeyValueLogMessage.build("Scrubber: dangling index entry",
+                        LogMessageKeys.KEY, valueKey.toString())
+                        .addKeysAndValues(common.indexLogMessageKeyValues())
+                        .toString());
+            }
+            if (spolicy().allowRepair()) {
+                // remove this index
```
alecgrieser, line 225:
```suggestion
                // remove this index entry
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+        // This will always return null (!) - but sometimes will delete a dangling index
+        FDBIndexedRecord<Message> indexResult = cursorResult.get();
+
+        if (! indexResult.hasStoredRecord() ) {
+            // Here: Oh, No! this index is dangling!
+            final FDBStoreTimer timer = getRunner().getTimer();
+            timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_SCRUBBER_INDEXES_DANGLING);
+            final IndexEntry indexEntry = indexResult.getIndexEntry();
+            final Tuple valueKey = indexEntry.getKey();
+            final byte[] keyBytes = store.indexSubspace(common.getIndex()).pack(valueKey);
```
alecgrieser, line 216: It feels a little bit wrong to have this happen here rather than in the index maintainer, though maybe that's fine.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+                                        return false;
+                                    }
+                                }
+                                return !Objects.equals(cont, rangeEnd);
+                            }));
+        });
+    }
+
+    @Nullable
+    private CompletableFuture<FDBStoredRecord<Message>> deleteIndexIfDangling(FDBRecordStore store, final RecordCursorResult<FDBIndexedRecord<Message>> cursorResult) {
```
alecgrieser, line 206: Hm, the API here is a little weird. It kind of feels like something is quite right about the abstractions, because the whole reason this returns a `CompletableFuture<FDBStoredRecord<Message>>` is so that it can be fed into the indexer code, which expects to be given a record (that it then indexes).

It almost feels like the right abstraction would take a method that then processes each record. In the case of the "scrubIndex" case, it would delete any dangling entries, and in the other cases, it would update the index(es) (conditionally in the "scrubRecords" case).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+        Index index = common.getIndex();
+        final RecordMetaData metaData = store.getRecordMetaData();
+        final RecordMetaDataProvider recordMetaDataProvider = common.getRecordStoreBuilder().getMetaDataProvider();
+        if (recordMetaDataProvider == null || !metaData.equals(recordMetaDataProvider.getRecordMetaData())) {
+            throw new MetaDataException("Store does not have the same metadata");
+        }
+        final IndexMaintainer maintainer = store.getIndexMaintainer(index);
+
+        // scrubbing idempotence indexes only (for now)
+        validateOrThrowEx(maintainer.isIdempotent(), "scrubbed index is not idempotent");
```
alecgrieser, line 157: This feels like not quite the right check. Maybe it'll come out in the wash when the index is actually scanned, but I think the thing to check is that the index type supports scanning `BY_VALUE`.

At the moment, there's not a terrific way of knowing if an index type supports scanning `BY_VALUE`. We could theoretically add that as a method on the index maintainer, I suppose.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+
+            if (LOGGER.isWarnEnabled() && spolicy().shouldReportDangling()) {
+                LOGGER.warn(KeyValueLogMessage.build("Scrubber: dangling index entry",
+                        LogMessageKeys.KEY, valueKey.toString())
+                        .addKeysAndValues(common.indexLogMessageKeyValues())
+                        .toString());
+            }
+            if (spolicy().allowRepair()) {
+                // remove this index
+                store.getContext().ensureActive().clear(keyBytes);
```
alecgrieser, line 226: I think this needs to add a conflict range on the (missing) record that it's deleting the index entry of (or not scan the index at `SNAPSHOT`), because I think this is susceptible to the following concurrency bug:

1. A dangling entry exists pointing to a record that doesn't exist
2. Transaction 1 scans the index and sees that it's missing a record, deletes the entry (but not yet committing)
3. Transaction 2 inserts a new record with that primary key and index value, overwriting the old, previously dangling entry and commits
4. Transaction 1 commits (successfully), deleting the new entry, which now means that the record is unindexed

Scanning the index at `SERIALIZABLE` would fix the problem, as operation 4 would mean the read of the index entry step 1 conflicts with step 3 inserting the new index entry. Adding an explicit read conflict range on the (missing) record would also fix the problem as that read conflict range would conflict with the record write in step 3.

It would also be good to get a test of this...somehow, though that test is somewhat tricky to write without a transactional scrubbing variant was created (that scrubs a whole index in one go). Then I think you could create a dangling index entry and start two transactions, one which inserted a replacement record and another that scrubbed the whole index, and make sure that the right operations failed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+                    // should I convert it to a single nested statement?
+                    final IndexEntry indexEntry = new IndexEntry(
+                            index,
+                            FDBRecordStoreBase.indexEntryKey(index, entry.getKey(), rec.getPrimaryKey()),
+                            entry.getValue());
+                    final byte[] keyBytes = maintainer.getIndexSubspace().pack(indexEntry.getKey());
+                    return maintainer.state.transaction.get(keyBytes).thenApply(Objects::isNull);
+                })
+                .collect(Collectors.toList()))
+                .thenCompose(list -> {
```
alecgrieser, line 331:
```suggestion
                .thenApply(list -> {
```

Then the method itself can return `null` or `rec`, as appropriate, not `CompletableFuture.completeFuture`s thereof

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+
+        return AsyncUtil.getAll(indexEntryNoPKs.stream()
+                .map(entry -> {
+                    // should I convert it to a single nested statement?
+                    final IndexEntry indexEntry = new IndexEntry(
+                            index,
+                            FDBRecordStoreBase.indexEntryKey(index, entry.getKey(), rec.getPrimaryKey()),
+                            entry.getValue());
+                    final byte[] keyBytes = maintainer.getIndexSubspace().pack(indexEntry.getKey());
+                    return maintainer.state.transaction.get(keyBytes).thenApply(Objects::isNull);
```
alecgrieser, line 328: I don't think this will work with `RANK` or `TEXT` indexes, which will pass the `isIdemoptent` test.

I almost feel like a method on the index maintainer that is something like "should re-index" or something and returns a `CompletableFuture<Boolean>` might work for this instead.

fdb-record-layer-core/src/main/proto/index_build.proto
```diff
@@ -31,6 +31,7 @@ message IndexBuildIndexingStamp {
     INVALID    = 0;   // let zero be invalid
     BY_RECORDS = 1;   // indexing by records - scanning all records in a record store
     BY_INDEX   = 2;   // indexing by a source index. When source index points to all possible records of the target index.
+    SCRUB_REPAIR =3;  // while readable, verify index validity. Repair if needed.
```
alecgrieser, line 34:
```suggestion
    SCRUB_REPAIR = 3;  // while readable, verify index validity. Repair if needed.
```
Probably worth aligning the other values to the new position of the `=`, too

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -217,6 +212,10 @@ private IndexingByIndex getIndexerByIndex() {
             throw FDBExceptions.wrapException(ex);
         }
 
+        if (isScrubber()) {
```
alecgrieser, line 215: This `if`-statement seems a bit weird to me--like it's a suggestion that something isn't quite right about the abstractions here that one could get to this state.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1953,7 +1989,160 @@ public IndexingPolicy build() {
...
+             * @param reportMissingLimit the max number of missing-indexes errors to report.
+             * @return this builder.
+             */
+            public Builder setReportMissingLimit(final int reportMissingLimit) {
+                this.reportMissingLimit = reportMissingLimit;
+                return this;
+            }
+
+            /**
+             * Set 'scrub but do not repair' mode.
```
alecgrieser, line 2120: I might rephrase this as "Set whether the index is repaired when scrubbing" or something like that. The way it's phrased makes it sound like if it's set to `true`, then it does _not_ repair. (This is clarified in the `param` description, but it should be a little more clear from the overall description, too.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1953,7 +1989,160 @@ public IndexingPolicy build() {
...
+             *            - if true (default), report errors (up to report limits) but always repair.
+             * @return this builder.
+             */
+            public Builder setAllowRepair(boolean val) {
+                this.allowRepair = val;
+                return this;
+            }
+
+            /**
+             * set records/indexes scan quota. The scrubbing will return after scanning this quota, or more, records/indexes
```
alecgrieser, line 2131:
```suggestion
             * Set records/indexes scan quota. The scrubbing will return after scanning this quota, or more, records/indexes
```

I don't love the word "quota" here. Would "scrubbed record limit" maybe be better?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1490,6 +1514,17 @@ public Builder setIndexingPolicy(@Nullable final IndexingPolicy indexingPolicy)
             return this;
         }
 
+        /**
+         * Add a {@link ScrubbingPolicy} policy. If set, this policy will enforce index scrubbing (instead of index
```
alecgrieser, line 1518: Is there anything in the documentation about what index scrubbing actually is?

alecgrieser (changes requested):
Okay, I went through the non-test code. I still need to go through the test code, but I figured you'd want this review sooner rather than later.

I still think that there's something broken about the abstractions. I believe we discussed separating out the top-level API for this from the `OnlineIndexer`, even if it used the indexing machinery under-the-hood. I think that's still a good idea, after looking through this, but it also seems like more generally that it's a sign we need to rethink how the code in this package in structured. I also suspect that you want to merge this first, and then make the change in a follow up PR, but I'm a little concerned about doing that if this could change on disk data (or if it would just be cleaner to merge it in to `main` after that's the way we want it, especially if it introduces a whole new API for users to call).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -341,6 +376,49 @@ public void setFallbackMode() {
...
+                                }
+                                return recordsRanges.onHasNext().thenCompose(hasNextRecord -> {
+                                    if (Boolean.FALSE.equals(hasNextRecord)) {
+                                        tr.clear(recordsRangeSubspace.range());
+                                    }
+                                    return AsyncUtil.DONE;
+                                });
+                            });
+                        }
+                        tr.clear(store.indexRangeSubspace(index).range());
```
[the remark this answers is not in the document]
jjezra: Oh, it's a leftover bug (with your idea of separate scrubbing ranges, there is no reason to clear this data). I was going to remove it when implementing the requested changes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -579,6 +579,10 @@ public String logKey() {
         ONLINE_INDEX_BUILDER_RANGES_BY_COUNT("number of indexer iterations terminated by scan limit", false),
         /** The number of {@link OnlineIndexer} range scans terminated after hitting the size limit. */
         ONLINE_INDEX_BUILDER_RANGES_BY_SIZE("number of indexer iterations terminated by write limit", false),
+        /** The number of missing indexes detected by {@link OnlineIndexer} scrubber. */
+        ONLINE_INDEX_SCRUBBER_INDEXES_MISSING("number of missing indexes detected by scrubber", false),
+        /** The number of dangling indexes detected by {@link OnlineIndexer} scrubber. */
+        ONLINE_INDEX_SCRUBBER_INDEXES_DANGLING("number of dangling indexes detected by scrubber", false),
```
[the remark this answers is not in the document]
jjezra: Changed to: 
`
        /** The number of missing index entries detected by the online scrubber. */
        ONLINE_SCRUBBER_INDEX_ENTRIES_MISSING("number of missing index entries detected by online scrubber", false),
        /** The number of dangling index entries detected by online scrubber. */
        ONLINE_SCRUBBER_INDEX_ENTRIES_DANGLING("number of dangling index entries detected by online scrubber", false),
`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -70,20 +72,39 @@
...
     private final IndexingThrottle throttle;
 
     private long timeOfLastProgressLogMillis = 0;
     private boolean forceStampOverwrite = false;
 
-    IndexingBase(@Nonnull IndexingCommon common) {
+    IndexingBase(@Nonnull IndexingCommon common,
+                 @Nonnull OnlineIndexer.IndexingPolicy policy) {
+        this.common = common;
+        this.policy = policy;
```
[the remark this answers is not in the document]
jjezra: No particular reason. I can move it to common.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -591,6 +700,36 @@ void decreaseLimit(@Nonnull FDBException fdbException,
...
+        if (!isValid) {
+            throw new ValidationException(msg,
+                    LogMessageKeys.INDEX_NAME, common.getIndex().getName(),
+                    LogMessageKeys.SOURCE_INDEX, policy.getSourceIndex(),
+                    LogMessageKeys.INDEXER_ID, common.getUuid());
+        }
+    }
+
+    /**
+     * thrown when indexing validation fails.
```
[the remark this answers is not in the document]
jjezra: Changed to: 
```
    /**
     * thrown when the indexing process fails to meet a precondition.
     */
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -591,6 +700,36 @@ void decreaseLimit(@Nonnull FDBException fdbException,
...
+     */
+    @SuppressWarnings("serial")
+    public static class ValidationException extends RecordCoreException {
+        public ValidationException(@Nonnull String msg, @Nullable Object ... keyValues) {
```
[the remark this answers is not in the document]
jjezra: `public` Removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -70,20 +72,39 @@
     private static final Object INDEX_BUILD_LOCK_KEY = 0L;
     private static final Object INDEX_BUILD_SCANNED_RECORDS = 1L;
     private static final Object INDEX_BUILD_TYPE_VERSION = 2L;
+    private static final Object INDEX_BUILD_SCRUB_INDEX_RANGE = 3L;
+    private static final Object INDEX_BUILD_SCRUB_RECORDS_RANGE = 4L;
```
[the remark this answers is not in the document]
jjezra: Good point! Changed to: 
```
         // This subspace holds the scrubbed ranges of the index itself (when looking for dangling entries)
        return indexBuildSubspace(store, index, INDEX_SCRUBBED_INDEX_RANGES); 
```
And 
```
       // This subspace hods the scrubbed ranges of the records (when looking for missing index entries)
        return indexBuildSubspace(store, index, INDEX_SCRUBBED_RECORDS_RANGES);
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -341,6 +376,49 @@ public void setFallbackMode() {
...
+                        if (indexingTypeStamp.equals(savedStamp)) {
+                            RangeSet indexRangeSet = new RangeSet(indexesRangeSubspace);
+                            RangeSet recordsRangeSet = new RangeSet(recordsRangeSubspace);
+                            AsyncIterator<Range> indexRanges = indexRangeSet.missingRanges(tr).iterator();
+                            AsyncIterator<Range> recordsRanges = recordsRangeSet.missingRanges(tr).iterator();
+                            return indexRanges.onHasNext().thenCompose(hasNextIndex -> {
+                                if (Boolean.FALSE.equals(hasNextIndex)) {
+                                    tr.clear(indexesRangeSubspace.range());
+                                }
+                                return recordsRanges.onHasNext().thenCompose(hasNextRecord -> {
```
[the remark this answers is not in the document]
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -341,6 +376,49 @@ public void setFallbackMode() {
...
+                                    if (Boolean.FALSE.equals(hasNextRecord)) {
+                                        tr.clear(recordsRangeSubspace.range());
+                                    }
+                                    return AsyncUtil.DONE;
+                                });
+                            });
+                        }
+                        tr.clear(store.indexRangeSubspace(index).range());
+                        tr.clear(indexBuildScannedRecordsSubspace(store, index).pack());
```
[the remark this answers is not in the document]
jjezra: I was trying to be consistent with FDBRecordStore.java:3600 (`clearIndexData`): 
```
      tr.clear(Range.startsWith(OnlineIndexer.indexBuildScannedRecordsSubspace(this, index).pack()));
```
Shall I change them both?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -341,6 +376,49 @@ public void setFallbackMode() {
...
+                    if (bytes != null) {
+                        IndexBuildProto.IndexBuildIndexingStamp savedStamp = parseTypeStampOrThrow(bytes);
+                        final Subspace indexesRangeSubspace = indexScrubIndexRangeSubspace(store, index);
+                        final Subspace recordsRangeSubspace = indexScrubRecordsRangeSubspace(store, index);
+                        if (indexingTypeStamp.equals(savedStamp)) {
+                            RangeSet indexRangeSet = new RangeSet(indexesRangeSubspace);
+                            RangeSet recordsRangeSet = new RangeSet(recordsRangeSubspace);
+                            AsyncIterator<Range> indexRanges = indexRangeSet.missingRanges(tr).iterator();
+                            AsyncIterator<Range> recordsRanges = recordsRangeSet.missingRanges(tr).iterator();
+                            return indexRanges.onHasNext().thenCompose(hasNextIndex -> {
```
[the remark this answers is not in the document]
jjezra: Sure! Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -557,6 +640,32 @@ private static void timerIncrement(@Nullable FDBStoreTimer timer, FDBStoreTimer.
         return syntheticPlan.execute(store, rec).forEachAsync(syntheticRecord -> maintainer.update(null, syntheticRecord), 1);
     }
 
+    protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMessageKeyValues,
+                                                       BiFunction<FDBRecordStore, AtomicLong,  CompletableFuture<Boolean>> iterateRange,
+                                                       @Nonnull SubspaceProvider subspaceProvider, @Nonnull Subspace subspace) {
+
+        return AsyncUtil.whileTrue(() ->
+                        buildCommitRetryAsync( iterateRange,
```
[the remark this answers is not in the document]
jjezra: Removed 8 whitespaces (and the one before `iteration`). Cmd-Ctrl-i all the way down left the other lines unchanged.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+    @Nonnull
+    static IndexBuildProto.IndexBuildIndexingStamp compileIndexingTypeStamp() {
+        return
+                IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                        .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.SCRUB_REPAIR)
+                        .build();
+    }
+
+    @Nonnull
+    private OnlineIndexer.ScrubbingPolicy spolicy() {
```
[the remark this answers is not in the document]
jjezra: Changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,359 @@
...
+                        return CompletableFuture.completedFuture(rec);
+                    }
+                    // report only mode
+                    return CompletableFuture.completedFuture(null);
+                });
+    }
+
+    @Override
+    CompletableFuture<Void> rebuildIndexInternalAsync(final FDBRecordStore store) {
+        return null;
```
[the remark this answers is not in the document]
jjezra: Changed to `throw new UnsupportedOperationException()`. 
I'll be happy to implement this function after stabilizing this code a little.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/NoOpIndexMaintainer.java
```diff
@@ -136,4 +137,8 @@ public boolean canDeleteWhere(@Nonnull QueryToKeyMatcher matcher, @Nonnull Key.E
         });
     }
 
+    @Override
+    public <M extends Message> List<IndexEntry> evaluateIndex(@Nonnull FDBRecord<M> record) {
+        return null;
```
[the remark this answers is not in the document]
jjezra: Fixed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/TerribleIndexMaintainer.java
```diff
@@ -161,4 +162,9 @@ public boolean canDeleteWhere(@Nonnull QueryToKeyMatcher matcher, @Nonnull Key.E
         future.completeExceptionally(new UnsupportedOperationException("TerribleIndexMaintainer does not implement performOperation"));
         return future;
     }
+
+    @Override
+    public <M extends Message> List<IndexEntry> evaluateIndex(@Nonnull FDBRecord<M> record) {
+        return null;
```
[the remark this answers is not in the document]
jjezra: Changed to: 
```
        throw new UnsupportedOperationException("TerribleIndexMaintainer does not implement evaluateIndex");        
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Actually it's up to the caller. The `scrubbingPolicy` can be set to scrub Dangling only, Missing only, or both.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Yeh, it felt a little weird to me too. The way I thought about it is that all the other indexers may return a record to be indexed (and, if applicable, added to the conflict list), but the dangling index is pointing to a null record - therefore, even if dangling, this null record is returned (after adjusting the index entry). 
My other concern was that since there is no record to conflict with, there is a (very) slim chance that a newly written in parallel record may - by an amazing coincidence - have just created this very same index. 
Is it likely to happen before the universe ends? If so, can the index entry itself be added to the conflict list?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Should this new maintainer method be part of this PR?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Oh, that answers my previous question. I'll look into adding an explicit read conflict range.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Fixed. 
It is, indeed, cleaner this way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Would `index.getType().equals(IndexTypes.VALUE)' do the trick?

fdb-record-layer-core/src/main/proto/index_build.proto
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I'm not sure about this one - the catcher's job is to catch errors and implement the logic of potential fallback to another indexing method. If `isScrubber`, there is no fallback to fall into. Hence the catcher, as in some other cases, passes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Changed to `Set whether the scrubber, if finding an error, will repair it.`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: "record limit" might be a little misleading. How about `entriesScannedLimit`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I suppose that "Which verifying a readable index's validity" isn't enough:) 
Shall I add more detailed description about scrubbing inside this comment?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: No, just the next commit. This line was a leftover - there is no need to clear the range subspace now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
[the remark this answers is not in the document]
jjezra: Changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: No reason but ease of read. Would you find it more appropriate in common?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Moved `scrubbingPolicy` to common. 
policy is a little tricky, as it can be modified changed during fallback (that's why it was not in common in the first place). I left it as is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: (After consulting with Alec offline)
Added the index entry itself to the conflict list.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Added VALUE index only constraint.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Changed to `entriesScanLimit`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Added more.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
```diff
@@ -0,0 +1,353 @@
...
+    static IndexBuildProto.IndexBuildIndexingStamp compileIndexingTypeStamp() {
+        return
+                IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                        .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.SCRUB_REPAIR)
+                        .build();
+    }
+
+    @Override
+    CompletableFuture<Void> buildIndexInternalAsync() {
+        return getRunner().runAsync(context -> openRecordStore(context)
```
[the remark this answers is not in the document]
jjezra: Oh, the intention to retry each subset (if fails) in `buildCommitRetryAsync` ... `throttledRunAsync`. The subsets are updated for each range. If this logic is broken it affects all the indexers. Is it broken?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
```diff
@@ -187,6 +190,16 @@ public long getLeaseLengthMillis() {
         return leaseLengthMillis;
     }
 
+    @Nonnull
+    public OnlineIndexer.ScrubbingPolicy getScrubbingPolicy() {
+        assert (scrubbingPolicy != null);// this can never happen, just eliminating the compiler warnings
```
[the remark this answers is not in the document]
jjezra: Nope. Only `IndexingScrubber` uses `getScrubbingPolicy`. And it can be created only when `common.isScrubber()' is true.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -579,6 +579,10 @@ public String logKey() {
         ONLINE_INDEX_BUILDER_RANGES_BY_COUNT("number of indexer iterations terminated by scan limit", false),
         /** The number of {@link OnlineIndexer} range scans terminated after hitting the size limit. */
         ONLINE_INDEX_BUILDER_RANGES_BY_SIZE("number of indexer iterations terminated by write limit", false),
+        /** The number of missing index entries detected by the online scrubber. */
+        ONLINE_SCRUBBER_INDEX_ENTRIES_MISSING("number of missing index entries detected by online scrubber", false),
```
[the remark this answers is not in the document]
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1953,7 +1987,160 @@ public IndexingPolicy build() {
...
+             *            - if true (default), report errors (up to report limits) but always repair.
+             * @return this builder.
+             */
+            public Builder setAllowRepair(boolean val) {
+                this.allowRepair = val;
+                return this;
+            }
+
+            /**
+             * Set records/index entries scan limit. The scrubbing task will return after scanning more than this limit.
```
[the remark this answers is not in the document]
jjezra: You are right, this is not a real limit, but more of a quota. 
The indexer/scrubber reads chunks of entries, while the throttle increases/decreases the chunks' size (of to config's max limit). The scrubber stops when the number of scanned entries exceeds the limit/quota. How should it be called?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1953,7 +1987,160 @@ public IndexingPolicy build() {
...
+            int reportDanglingLimit = 1000;
+            int reportMissingLimit = 1000;
+            boolean allowRepair = true;
+            long entriesScanLimit = 0;
+
+            protected Builder() {
+            }
+
+            /**
+             * If never called, the defaulted to allow scanning the index, looking for dangling index entries.
```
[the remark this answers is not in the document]
jjezra: Updated.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerScrubTest.java
```diff
@@ -0,0 +1,321 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Tests for scrubbing readable indexes with {@link OnlineIndexer}.
+ */
+public class OnlineIndexerScrubTest extends OnlineIndexerTest {
```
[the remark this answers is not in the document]
jjezra: :)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -341,6 +376,49 @@ public void setFallbackMode() {
                 });
     }
 
+    @Nonnull
+    private CompletableFuture<Void> setScrubberTypeOrThrow(FDBRecordStore store) {
```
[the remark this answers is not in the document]
jjezra: Yes it should. 
I've eliminated the scrubber's type stamp. Please note that: 
1. The type stamp is still defined - as it is required by `IndexerBase`. 
2. I've also eliminated the `indexBuildScannedRecordsSubspace` reset (which is discussed below) because we can't tell now that a scrubber runs for the first time. I suppose it's consistent in the sense that by scrubbing, the scanned records counter grows beyond the indexing process count. 
One thing that may happen is the scrubber successfully reseting its range-sets while an indexer changes the index state to WRITE_ONLY, but I suppose it's harmless now - as the scrubbing's range-sets are separate from the indexing one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Oh, I see your point. I suppose a different module for each scrubber type would be cleaner from the indexer's prospective, in the expense of adding (not too many) lines of code at the caller's side. 
Would you like me to make this change in this PR, or is it something that can be changed later?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
alecgrieser: I guess the main question I'd like answered is some clarity around how much it can exceed the quota/limit. It's the "more than" language that's tripping me up

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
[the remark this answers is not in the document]
alecgrieser: Hm, interesting. It looks like the `IndexerByRecords` is architected a bit differently, but I think also a bit suspect that the `IndexerByIndex` is doing the right thing here, too. Hmm

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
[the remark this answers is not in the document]
alecgrieser: I guess the claim your making is that everything outside of the `IndexingScrubber` calls `hasScrubbingPolicy` only? I guess that kind of makes sense, but it seems like it would make sense to validate that at run time to protect ourselves from writing bugs in the future. Or we move the `scrubbingPolicy` around so that we can't access it anywhere that it might be `null`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
alecgrieser: I guess the question is how much we care about backwards compatibility on this new code and how hard it would be to extract that (later). I think I'd err on the side of doing the refactoring _now_ while this is fresh rather than trying to untangle it later, but it does probably mean that it takes a little longer to merge this

jjezra:
> I guess there's still somewhat of an open question as to whether we want to change the API first so that the scrubber isn't created by the `OnlineIndexer` before merging this to `main`. That's still kind of my preference, at least partially because I think it would encourage a better organization under-the-hood, but not sure what your preference is

I have no particular preference about the API (as long as it's usable). I suppose I could have another module, similar to `OnlineIndexer`, that is another user of the `IndexingBase` family. `IndexScrubber`? 
Since we don't want to break `OnlineIndexer`'s subclass, it would have to import it, though.

alecgrieser:
Yeah, I think something like that kind of organization would make sense. It might still live in that same Java package, but be a slightly different part of the class hierarchy

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubber.java
jjezra: Ok. I'll push a version with the trivial requested changes fixes, then start working on the refactoring - both this one and the API. If it ends up as a big mess (or a crazy time consumer) we'll be able to reset to the trivial.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: That's hard to tell as the caller can overwrite the default max chunk limit (100 items) with its own.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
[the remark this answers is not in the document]
jjezra: Ok. I've changed `common.getScrubbingPolicy` to Nullable, and created a local Nonnull `getScrubbingPolicy` in the scrubber module.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
alecgrieser: Hm, well, ideally it would let the caller know that the scan limit may be exceeded by up to the chunk limit or something along those lines (or the scrubber would strictly enforce the scan limit to make the API more understandable).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
```diff
@@ -0,0 +1,1046 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+@API(API.Status.UNSTABLE)
+public class OnlineScrubber implements AutoCloseable {
```
alecgrieser, line 46: minor: I think we probably want "index" in the name here. Either "IndexScrubber" or "OnlineIndexScrubber" or something like that. ("IndexScrubber" is a little too close to what some of the internal classes are called, but maybe that means the internal class needs a rename? Unclear.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
```diff
@@ -0,0 +1,1046 @@
...
+    @Nonnull
+    private static final Logger LOGGER = LoggerFactory.getLogger(OnlineScrubber.class);
+
+    @Nonnull private final IndexingCommon common;
+    @Nonnull private final FDBDatabaseRunner runner;
+    @Nonnull private final Index index;
+
+    @Nullable
+    private IndexingBase indexer = null;
```
alecgrieser: Is this used?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
```diff
@@ -61,6 +61,7 @@
     @Nonnull private OnlineIndexer.IndexStatePrecondition indexStatePrecondition;
     @Nonnull public OnlineIndexer.Config config; // this item may be modified on the fly
     @Nullable private final Function<OnlineIndexer.Config, OnlineIndexer.Config> configLoader;
+    @Nullable private final Object scrubbingPolicy;
```
alecgrieser, line 64: Why is this an `Object`? Ideally, we wouldn't need it in the `IndexingCommon` at all as the scrubbers would be separate enough from the indexers that it wouldn't be something that would need to wind up all the way in `IndexingCommon` (maybe in `ScrubbingCommon` or something), but it seems weird that it's essentially untyped here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
```diff
@@ -0,0 +1,226 @@
...
+    static IndexBuildProto.IndexBuildIndexingStamp compileIndexingTypeStamp() {
+        return
+                IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                        .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.SCRUB_REPAIR)
+                        .build();
+    }
+
+    @Override
+    CompletableFuture<Void> buildIndexInternalAsync() {
+        return getRunner().runAsync(context -> openRecordStore(context)
```
alecgrieser, line 108: I guess I'll reiterate a comment from a previous round that I don't think we want this in a runner. It also looks like the store isn't used here, and the only I/O it might do before it calls `scrubIndex` is to get a subspace from the subspace provider. I don't think that should be strictly necessary (the `subspace` should be something that it can create from the provider when needed), but it seems off that it creates the store at least.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -217,6 +211,10 @@ private IndexingByIndex getIndexerByIndex() {
             throw FDBExceptions.wrapException(ex);
         }
 
+        if (common.isScrubber()) {
```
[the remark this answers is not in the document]
alecgrieser: Is the `OnlineIndexer` still involved in running scrubbers? Just checking to make sure we understand what the current state of this code is and if this is still necessary.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -330,16 +335,27 @@ private IndexingByRecords getIndexerByRecordsOrThrow() {
...
             IndexingBase indexingBase = getIndexerByRecords();
-            indexingBase.setFallbackMode();
+            indexingBase.enforceStampOverwrite();
             return indexingBase;
         }
         if (indexingPolicy.isByIndex()) {
             return getIndexerByIndex();
         }
+        if (common.isScrubber()) {
+            return getIndexerScrubber();
```
alecgrieser, line 357: Is this still a possibility?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -330,16 +335,27 @@ private IndexingByRecords getIndexerByRecordsOrThrow() {
         return getIndexerByRecords();
     }
 
+    @Nonnull
+    private IndexingScrubMissing getIndexerScrubber() {
```
alecgrieser, line 339: Does this need to be removed or updated to also possibly return a `IndexingScrubDangling`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
```diff
@@ -0,0 +1,918 @@
...
+     * OnlineScrubber.newBuilder().setRecordStoreBuilder(recordStoreBuilder).setIndex(index).build()
+     * </code></pre>
+     *
+     * <pre><code>
+     * OnlineScrubber.newBuilder().setDatabase(fdb).setMetaData(metaData).setSubspace(subspace).setIndex(index).build()
+     * </code></pre>
+     *
+     */
+    @API(API.Status.UNSTABLE)
+    public static class Builder {
```
alecgrieser, line 248: You mentioned in a conversation not on GitHub that you didn't like the amount of code duplication between this and the `OnlineIndexer.Builder`. I agree, and I _think_ that the most straightforward way to remove code dupe might be to have something like a "abstract index processor builder" or something that contains the configuration parameters relevant to both this and indexing and then handles creating an `OnlineIndexer.Config`, etc. I think it can be done in a way that's at least API (if not ABI) backwards compatible. But also not something that I think needs to be done before the scrubbing logic can be merged in

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
```diff
@@ -0,0 +1,918 @@
...
+             * @return this builder.
+             */
+            public Builder setAllowRepair(boolean val) {
+                this.allowRepair = val;
+                return this;
+            }
+
+            /**
+             * Set records/index entries scan limit.
+             * Note that for efficiency, the scrubber reads and processes chunks of data, and will only check this limit
```
alecgrieser, line 216: What's a "chunk" of data, in this context? Is that defined by the amount of data that can be processed in a single transaction?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerScrubTest.java
```diff
@@ -0,0 +1,309 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Tests for scrubbing readable indexes with {@link OnlineIndexer}.
+ */
+class OnlineIndexerScrubTest extends OnlineIndexerTest {
```
alecgrieser, line 48: Whatever the final name of the `OnlineScrubber` ends up being, this should probably be renamed to `${nameOfClass}Test` just to make it a little more discoverable.

I also think this is still missing some of the tests I'd suggested in: https://github.com/FoundationDB/fdb-record-layer/pull/1293#discussion_r647860963

alecgrieser (changes requested):
Overall, I think this is a much better API, so it looks like this is probably converging, though we should probably think about what ways we want to refactor these related classes to make it a little bit cleaner, though that work can wait until after the merge.

Left a few last mostly minor comments, including an ask for a few more tests

jjezra:
Test this please.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
jjezra: `OnlineIndexScrubber` it is. 
(Initially I called it IndexScrubber, but renamed for the same reason you highlighted)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
jjezra: Mmm... it's not there anymore. 
Were you looking at the pre-cleanup version? Should I have squashed it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
jjezra: The object, when non-null, returns true on `common.isScrubber()`. IndexingBase needs this information for the type-stamp. 
I thought it would be cleaner if the non-scrubbing modules wouldn't have to learn the ScrabbingPolicy's structure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
jjezra: Sure. 
Should I eliminate the `openRecordStore` from `IndexingByIndex` too?

jjezra: I'll leave IndexingByIndex for another time. Changing it will require its own test.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Nope. Removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I've missed it during the cleanup. 
Removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
jjezra: Nope, it's the number of records allowed to be read in a single transaction. This number is dynamic, but in normal operation the scrubber should quickly reach the maximum allowed number (as set by `setLimit`). 
I'll try to clarify...

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerScrubTest.java
jjezra: Done. 
+ Added negative tests within. Refuse scrubbing a non-readable index and refuse scrubbing a non-value index.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineScrubber.java
alecgrieser: Ah, yes, sorry, I thought I'd confirmed that it was in the final clean up version, too, but my mistake.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -0,0 +1,338 @@
...
+                .build()) {
+            indexScrubber.scrubDanglingIndexEntries();
+            indexScrubber.scrubMissingIndexEntries();
+        }
+        assertEquals(3 * 2 , timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_COUNT));
+        assertEquals(chunkSize * 3 * 2, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_MISSING_ENTRIES));
+
+        // refuse to scrub a non-readable index
```
alecgrieser, line 312: minor: it would be nice, if possible, if these two negative tests were their own test functions so that they can be run and inspected independently of the other tests

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -0,0 +1,919 @@
...
+             * @return this builder.
+             */
+            public Builder setAllowRepair(boolean val) {
+                this.allowRepair = val;
+                return this;
+            }
+
+            /**
+             * Set records/index entries scan limit.
+             * Note that for efficiency, the scrubber reads and processes batches of entries (either index' or records),
```
alecgrieser, line 216:
```suggestion
             * Note that for efficiency, the scrubber reads and processes batches of entries (either index entries or records)
```

alecgrieser (changes requested):
A few minor things that are hopefully easy enough to fix up before merging. I also think this is missing a concurrency test that I mentioned at one point, but maybe that's something to add later

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
jjezra: Eliminated from common.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
jjezra: Splited.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Applied.

jjezra:
Test this please

alecgrieser (approved):

PR #1295, opened 2021-06-03
Fixes #1294: Scanning a store with a limit can result in "attempted to return a result with NoNextReason of SOURCE_EXHAUSTED but a non-end continuation"

### LINKED ISSUE #1294: Scanning a store with a limit can result in "attempted to return a result with NoNextReason of SOURCE_EXHAUSTED but a non-end continuation"
When scanning a record store with a limit, if just the right combination of things happen, it's possible to get the error `attempted to return a result with NoNextReason of SOURCE_EXHAUSTED but a non-end continuation`. On d17664ff507cfe, the stack trace looks like:

```
com.apple.foundationdb.record.RecordCoreException: attempted to return a result with NoNextReason of SOURCE_EXHAUSTED but a non-end continuation
            at com.apple.foundationdb.record.RecordCursorResult.withoutNextValue(RecordCursorResult.java:257)
            at com.apple.foundationdb.record.provider.foundationdb.SplitHelper$KeyValueUnsplitter.onNext(SplitHelper.java:658)
            at com.apple.foundationdb.record.cursors.MapPipelinedCursor.tryToFillPipeline(MapPipelinedCursor.java:121)
            at com.apple.foundationdb.async.AsyncUtil$LoopPartial.apply(AsyncUtil.java:355)
            at com.apple.foundationdb.async.AsyncUtil$LoopPartial.run(AsyncUtil.java:378)
            at com.apple.foundationdb.async.AsyncUtil.whileTrue(AsyncUtil.java:432)
            at com.apple.foundationdb.record.cursors.MapPipelinedCursor.onNext(MapPipelinedCursor.java:79)
            at com.apple.foundationdb.record.provider.common.StoreTimer$1.onNext(StoreTimer.java:809)
            at com.apple.foundationdb.record.RecordCursor.getNext(RecordCursor.java:239)
```

Which means that the scanning code isn't quite handling all of the various states correctly.

alecgrieser:
This adds a guard to checkout for the an exhausted inner cursor, and if that happens, the scanning logic will now return an exhausted record cursor instead of trying to create a result with `SOURCE_EXHAUSTED` as the no-next-reason and a non-end continuation. Note that this bug requires that the final record in a store be split record with multiple parts and the out-of-band limit has to line up with hitting that record, so in practice, this should be a fairly rare occurrence, even though I could write a test that hit this 100% of the time.

This fixes #1294.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -18,7 +18,7 @@ Another, smaller change that has been made is that by default, new indexes added
 
 * **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Scanning a record store with split records and a scan limit could sometimes result in errors with the message `attempted to return a result with NoNextReason of SOURCE_EXHAUSTED but a non-end continuation` [(Issue #1294)](https://github.com/FoundationDB/fdb-record-layer/issues/1294)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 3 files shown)


alecgrieser:
Test failure from #1296, which I just filed.

test this please

ScottDugas (approved):

PR #1298, opened 2021-06-09
Resolves #1239: Use either record count or estimated size when determining IndexStatus in checkVersion

### LINKED ISSUE #1239: Use either record count or estimated size when determining IndexStatus in checkVersion
When the estimated size is exposed (see #1229), we should also allow users to use it in `checkVersion` instead of the record count. The size could, in theory, be more useful on record stores that don't have an appropriate index (or for whom executing a query against the index is too costly).

I think there's a way to make this backwards compatible. At some point, we may want to make using the size the default. However, there are also some unknowns about the performance of this on large stores, so until we know that for sure, I think it should be opt-in.

To make this backwards compatible, I propose adding a new method to the `UserVersionChecker` that takes the count and size as parameters, and then by default, it just calls the existing method. The `checkVersion` logic can then call the new method. Because we're not sure of the performance (and because we also don't want to do extra work), I propose that we make both lazy (i.e., a supplier of type `CompletableFuture`).

Hopefully, this could then be used by a user to solve #7.

alecgrieser:
This allows library consumers to create a `UserVersionChecker` that inspects the size of the record store rather than the number of records, which may be useful for stores that don't have a record size index or for whom determing the record count is more work than should be done in a single transaction (see #7). It also makes getting the count or size lazy, and so if the `UserVersionChecker` can determine an index state without needing to read from the database (such as if the index is on new types), it can avoid having to do I/O entirely.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -3434,15 +3434,19 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     }
 
     /**
-     * Get count of records to pass to checker to decide whether to build right away. If all of the new indexes are over
-     * a single type and that type has
+     * Get count of records to pass to a {@link UserVersionChecker} to decide whether to build right away. If all of the
+     * new indexes are over a single type and that type has a record key prefix, then this count will only be over the
+     * record type being indexed. If not, it will be the count of all records of all types, as in that case, the indexer
+     * will need to scan the entire store to build each index. If determining the record count would be too costly (such
+     * as if there is not an appropriate {@linkplain IndexTypes#COUNT count} index defined), this function may return
+     * {@link Long#MAX_VALUE} to indicate that an unknown and unbounded number of records would have to be scanned
+     * to build the index.
      *
      * @param newStore {@code true} if this is a brand new store
      * @param rebuildRecordCounts {@code true} if there is a record count key that needs to be rebuilt
      * @param indexes indexes that need to be built
      * @param singleRecordTypeWithPrefixKey either a single record type prefixed by the record type key or {@code null}
-     * @return a pair of a future that completes to the record count for the version checker
-     * and a flag that is {@code true} if this count is in fact for all record types
+     * @return a future that completes to the record count for the version checker
      */
     @Nonnull
     @SuppressWarnings("PMD.EmptyCatchBlock")
@@ -3468,7 +3472,11 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
             MutableRecordStoreState saveState = recordStoreStateRef.get();
             try {
                 recordStoreStateRef.set(writeOnlyState);
-                // TODO: FDBRecordStoreBase.checkPossiblyRebuild() could take a long time if the record count index is split into many groups (https://github.com/FoundationDB/fdb-record-layer/issues/7)
+                // Note the call below can time out if the count index group cardinality is too high. Users can avoid it by
+                // setting a custom UserVersionChecker that looks at the size estimate rather than the count, but we should
+                // consider checking the size by default or otherwise making the ergonomics around hitting that limitation
+                // better in the future
+                // See: FDBRecordStoreBase.checkPossiblyRebuild() could take a long time if the record count index is split into many groups (https://github.com/FoundationDB/fdb-record-layer/issues/7)
                 return getSnapshotRecordCount();
             } catch (RecordCoreException ex) {
                 // Probably this was from the lack of appropriate index on count; treat like rebuildRecordCounts = true.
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3375,57 +3383,81 @@ private boolean areAllRecordTypesSince(@Nullable Collection<RecordType> recordTy
...
-     * Get count of records to pass to checker to decide whether to build right away.
+     * Get count of records to pass to checker to decide whether to build right away. If all of the new indexes are over
+     * a single type and that type has
+     *
      * @param newStore {@code true} if this is a brand new store
      * @param rebuildRecordCounts {@code true} if there is a record count key that needs to be rebuilt
      * @param indexes indexes that need to be built
+     * @param singleRecordTypeWithPrefixKey either a single record type prefixed by the record type key or {@code null}
      * @return a pair of a future that completes to the record count for the version checker
      * and a flag that is {@code true} if this count is in fact for all record types
```
MMcM, line 3445: Just the count now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3437,7 +3469,7 @@ private boolean areAllRecordTypesSince(@Nullable Collection<RecordType> recordTy
             try {
                 recordStoreStateRef.set(writeOnlyState);
                 // TODO: FDBRecordStoreBase.checkPossiblyRebuild() could take a long time if the record count index is split into many groups (https://github.com/FoundationDB/fdb-record-layer/issues/7)
```
MMcM, line 3471: Will this issue remain open? Will there be another change to address it beyond giving the caller more control like here? If not, maybe the comment doesn't need to reference it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3531,40 +3571,70 @@ public Subspace getLegacyVersionSubspace() {
...
+                    subspaceProvider.logKey(), subspaceProvider.toString(context));
+
+            long recordCount = recordCountRef.get();
+            if (recordCount >= 0L) {
+                msg.addKeyAndValue(LogMessageKeys.RECORD_COUNT, recordCount == Long.MAX_VALUE ? "unknown" : Long.toString(recordCount));
+            }
+            long recordsSize = recordsSizeRef.get();
+            if (recordsSize >= 0L) {
+                msg.addKeyAndValue(LogMessageKeys.RECORDS_SIZE_ESTIMATE, Long.toString(recordsSize));
+            }
```
MMcM, line 3624: I think this deserves a comment: if the checker called the supplier, then log what it yielded.

MMcM (changes requested):

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3375,57 +3383,81 @@ private boolean areAllRecordTypesSince(@Nullable Collection<RecordType> recordTy
...
-                    if (newStore ? LOGGER.isDebugEnabled() : LOGGER.isInfoEnabled()) {
-                        KeyValueLogMessage msg = KeyValueLogMessage.build("upgrading unsplit format on empty store",
-                                                    LogMessageKeys.NEW_FORMAT_VERSION, formatVersion,
-                                                    subspaceProvider.logKey(), subspaceProvider.toString(context));
-                        if (newStore) {
-                            LOGGER.debug(msg.toString());
-                        } else {
-                            LOGGER.info(msg.toString());
+            // If all the new indexes are only for a record type whose primary key has a type prefix, then we can scan less.
+            RecordType singleRecordTypeWithPrefixKey = singleRecordTypeWithPrefixKey(indexes);
```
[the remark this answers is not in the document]
MMcM: This isn't about the cardinality of any of the types, but whether the primary keys are contiguous. It might be possible for multiple record types to be, if you assume that the type keys are dense.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -218,6 +219,15 @@ default FDBStoreTimer getTimer() {
         default IndexState needRebuildIndex(Index index, long recordCount, boolean indexOnNewRecordTypes) {
             return FDBRecordStore.disabledIfTooManyRecordsForRebuild(recordCount, indexOnNewRecordTypes);
         }
+
+        @Nonnull
+        default CompletableFuture<IndexState> needRebuildIndex(Index index,
```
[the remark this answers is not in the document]
alecgrieser: Okay, added some javadoc explaining what this function does. I also specified that the return future should complete after any accessed futures, though I think the only consequence would be some logging, which is probably enough to justify mentioning it, though I don't think it would be the end of the world if not implemented correctly.

alecgrieser:
Rebase is on top of latest changes to `main`. Review comments are responded to in their own commit

MMcM (approved):

ScottDugas (approved):

PR #1300, opened 2021-06-17
Resolves #1299: RecordCursorContinuation should have method to conver…

### LINKED ISSUE #1299: RecordCursorContinuation should have method to convert to ByteString
At the moment, the `RecordCursorContinuation` class has a method to convert to a `byte[]`, but not one to convert to a `ByteString`. The `ByteString` serialization, however, would have a few advantages, including a guarantee of immutability, and it would also be essentially free for many of the cursors (in particular, the ones built on top of protobuf messages). So this might make using cursor continuations slightly more ergonomic.

tian-yizuo:
…t to ByteString

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ByteArrayContinuation.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ByteArrayContinuation.java
@@ -22,6 +22,7 @@ package com.apple.foundationdb.record;
 
 import com.apple.foundationdb.annotation.SpotBugsSuppressWarnings;
 import com.apple.foundationdb.annotation.API;
+import com.google.protobuf.ByteString;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -34,6 +35,8 @@ import java.nio.ByteBuffer;
 public class ByteArrayContinuation implements RecordCursorContinuation {
     @Nonnull
     private final byte[] bytes;
+    @Nullable
+    private ByteString byteString;
 
     @SpotBugsSuppressWarnings("EI2")
     private ByteArrayContinuation(@Nonnull final byte[] bytes) {
@@ -45,6 +48,15 @@ public class ByteArrayContinuation implements RecordCursorContinuation {
     @SpotBugsSuppressWarnings("EI")
     public byte[] toBytes() {
         return bytes;
+    }
+
+    @Override
+    @Nonnull
+    public ByteString toByteString() {
+        if (byteString == null) {
+            byteString = ByteString.copyFrom(bytes);
+        }
+        return byteString;
     }
 
     @Override
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursorContinuation.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursorContinuation.java
@@ -21,7 +21,9 @@
 package com.apple.foundationdb.record;
 
 import com.apple.foundationdb.annotation.API;
+import com.google.protobuf.ByteString;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
@@ -59,6 +61,17 @@ public interface RecordCursorContinuation {
      */
     @Nullable
     byte[] toBytes();
+
+    /**
+     * Serialize this continuation to a ByteString object.
+     * If {@link #toBytes()} returns null, then {@code toByteString()} is supposed to return EMPTY.
+     * @return a (possibly EMPTY) ByteString containing a binary serialization of this continuation
+     */
+    @Nonnull
+    default ByteString toByteString() {
+        final byte[] bytes = toBytes();
+        return bytes == null ? ByteString.EMPTY : ByteString.copyFrom(bytes);
+    }
 
     /**
      * Return whether this continuation is an "end continuation", i.e., represents that the iteration has reached
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/ConcatCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/ConcatCursor.java
@@ -191,17 +191,19 @@ public class ConcatCursor<T> implements RecordCursor<T> {
         @Nonnull
         private final RecordCursorResult<T> nextResult;
         @Nonnull
-        private final Function<byte[], RecordCursorProto.ConcatContinuation> continuationFunction;
+        private final Function<ByteString, RecordCursorProto.ConcatContinuation> continuationFunction;
         private final boolean isEnd;
         @Nullable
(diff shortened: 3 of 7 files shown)


tian-yizuo:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursorContinuation.java
```diff
@@ -60,6 +62,16 @@
...
     byte[] toBytes();
 
+    /**
+     * Serialize this continuation to a ByteString object.
+     * @return a (possibly EMPTY) ByteString containing a binary serialization of this continuation
+     */
+    @Nonnull
+    default ByteString toByteString() {
+        final byte[] bytes = toBytes();
+        return bytes == null ? ByteString.EMPTY : ByteString.copyFrom(bytes);
```
scgray, line 72: It may make the most sense to have @alecgrieser to the actual review, but one comment upon re-reading the original issue is:

> and it would also be essentially free for many of the cursors (in particular, the ones built on top of protobuf messages)

So, I think probably the intent was that this can be a good default implementation, but to look through the subclass of cursors and identify those that already have their continuation encoded as a `ByteString` and just natively return it.

tian-yizuo: I added the overrides for all the implementations that have ```cachedBytes``` encoded from the proto message, by adding a ```cachedByteString``` variable for them, and using the same way to lazily populate the cached ByteString object and return it. Please let me know if it looks good to you. @scgray

alecgrieser, line 72: I'm not sure I'm sold if this is the right behavior if `toBytes()` returns `null`, though I suspect that most of the time, an empty and a null byte string should have the same semantics. The alternative, I think, would be to have this method always return `null` if `toBytes()` is `null`, and then a different method, `toByteStringOrEmpty` could return `ByteString.EMPTY` if this one returns `null`. But maybe that's needlessly complicated.

In any case, we should document the behavior (that if `toBytes` is `null`, then this returns `EMPTY`).

alecgrieser: I think that this approach, caching the the byte string if possible but having the default make a copy, is the only way that doesn't change the interface, so this approach is probably fine.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/ConcatCursor.java
```diff
@@ -218,6 +220,20 @@ private ConcatCursorContinuation(boolean secondCursor, @Nonnull RecordCursorResu
...
 
+        @Override
+        @Nonnull
+        public ByteString toByteString() {
+            if (isEnd()) {
+                return ByteString.EMPTY;
+            } else {
+                //form bytes exactly once
+                if (cachedByteString == null) {
+                    cachedByteString = continuationFunction.apply(nextResult.getContinuation().toBytes()).toByteString();
```
alecgrieser, line 231: I feel like the `toBytes` method on this continuation should be modified to call `.toByteString` so that `cachedByteString` also gets populated and then doesn't need to be recomputed if someone calls `toByteString` followed by `toBytes`. It also means there's only one place where we need to have logic that calls the continuation function.

I think the other continuations that you overrode could probably follow suit.

alecgrieser, line 231: I think `continuationFunction` can now be redefined to be `Function<ByteString, RecordCursorProto.ConcatContinuation>`, and then one can call `nextResult.getContinuation.toByteString()` instead of `.toBytes()`. I _think_ that can actually save on copying byte arrays, at least sometimes.

There are a few other places in the continuations that currently call `ByteString.copyFrom(someContinuation.toBytes())` that can also probably be cleaned up.

alecgrieser (changes requested):
One suggested tweak to the way that the continuations that cache byte string should now compute the `byte[]` version, and also places that can now be cleaned up to use the `.toByteString()` method on continuations rather than copying `.toBytes()` on a continuation into a `ByteString`.

I also think that `ByteArrayContinuation` should probably also be modified to cache the `ByteString` (and possibly also take a byte string as a possible `from` source) because of the number of times that we end up using that continuation type.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursorContinuation.java
tian-yizuo: Added comments for this method. Please let me know if it looks good to you.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/ConcatCursor.java
tian-yizuo: Fixed it. Please take a look. @alecgrieser

tian-yizuo: Fixed the ```continuationFunction``` in this way for both ```ConcatCursorContinuation``` and ```SizeStatisticsCollectorCursorContinuation```. @alecgrieser

tian-yizuo:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ByteArrayContinuation.java
```diff
@@ -34,10 +35,13 @@
 public class ByteArrayContinuation implements RecordCursorContinuation {
     @Nonnull
     private final byte[] bytes;
+    @Nonnull
+    private ByteString byteString;
 
     @SpotBugsSuppressWarnings("EI2")
     private ByteArrayContinuation(@Nonnull final byte[] bytes) {
         this.bytes = bytes;
+        this.byteString = ByteString.copyFrom(bytes);
```
alecgrieser, line 44: Should this be a lazy copy? I believe the consequence of this as written is it will always proactively make a copy and put it in a `ByteString`, even if no one calls `.toByteString()`.

alecgrieser (changes requested):
Okay, looks good, except for one place that should possibly be a lazy copy into a `ByteString`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ByteArrayContinuation.java
tian-yizuo: Good point, fixed it. Thanks! @alecgrieser

alecgrieser (approved):

alecgrieser:
Looks like there were test failures from #1267.

test this please

scgray (approved):

PR #1303, opened 2021-06-25
 Rework Functionality to enable Nesting 

sophiabuell:


sophiabuell:
Jenkins test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/PrefixableExpression.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.metadata.expressions;
+
+public interface PrefixableExpression extends KeyExpression {
```
MMcM, line 23: I do not believe we need this any more, as it is now done without mutating key expressions.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -102,14 +102,18 @@ public static AvailableFields fromIndex(@Nonnull RecordType recordType,
...
             // Full text index entries have all of their fields except the tokenized one.
             keyFields.addAll(TextScanPlanner.getOtherFields(rootExpression));
         } else if (indexTypes.getValueTypes().contains(index.getType()) ||
-                   indexTypes.getRankTypes().contains(index.getType()) ||
-                   indexTypes.getLuceneTypes().contains(index.getType())) {
+                   indexTypes.getRankTypes().contains(index.getType())) {
             keyFields.addAll(KeyExpression.getKeyFields(rootExpression));
             valueFields.addAll(KeyExpression.getValueFields(rootExpression));
+        } else if (indexTypes.getLuceneTypes().contains(index.getType())) {
+            List<KeyExpression> nonPrimaries = rootExpression.normalizeKeyForPositions();
```
MMcM, line 114: There should be a comment here as there is for full text, that Lucene indexes also lack their non-stored fields. In the future, this will need to not assume only primary key fields are stored. And probably not leave holes where they were in the `IndexEntry`. But this is consistent for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -154,42 +141,91 @@ private void parseFieldNames(KeyExpression expression, @Nonnull String parentFie
...
         return AsyncUtil.DONE;
     }
 
+    private void insertDocumentField(final LuceneFieldKeyExpression expression, Object value,
+                                     final Document document, String prefix) {
+        if (!(value == null && expression.getType() != LuceneKeyExpression.FieldType.STRING)) {
+            String fieldName = prefix == null ? expression.getFieldName() : prefix.concat("_").concat(expression.getFieldName());
+            switch (expression.getType()) {
+                case INT:
+                    document.add(new IntPoint(fieldName, (Integer)value));
```
MMcM, line 218: Should be possible to have a stored int field, I think. Maybe just a `TODO` for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -154,42 +141,91 @@ private void parseFieldNames(KeyExpression expression, @Nonnull String parentFie
...
+                    break;
+                case STRING:
+                    document.add(new TextField(fieldName, value == null ? "" : (String)value, expression.isStored() ? Field.Store.YES : Field.Store.NO));
+                    break;
+                default:
+                    break;
```
MMcM, line 224: Shouldn't the default be error not do nothing?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneFieldKeyExpression.java
```diff
@@ -0,0 +1,81 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+
+public class LuceneFieldKeyExpression extends FieldKeyExpression implements LuceneKeyExpression {
+
+
+    private FieldType type;
+    private boolean sorted;
+    private boolean stored;
```
MMcM, line 34: I believe these can be `final`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneThenKeyExpression.java
```diff
@@ -0,0 +1,141 @@
...
+                return IndexTypes.LUCENE.equals(indexType);
+            }
+        };
+
+        visitor.visit(this);
+
+        return children;
+    }
+
+    public boolean fan() {
```
MMcM, line 118: This method needs a comment; it feels to me like the sense is backwards: a primary-key then-field doesn't fan out into multiple document fields.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -135,29 +136,29 @@
                         limitRemaining--;
                     }
                     List<Object> setPrimaryKey = Tuple.fromBytes(pk.bytes).getItems();
-                    List<Object> fields = Lists.newArrayList(fieldNames);
+                    List<Object> fieldValues = Lists.newArrayList(fields);
```
MMcM, line 139: I think it should be possible to do away with these dummy placeholders in the `IndexEntry` by adjusting the `AvailableFields`, and perhaps some other places. But a `TODO` for now is probably okay.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
```diff
@@ -183,11 +184,19 @@ public Index(@Nonnull RecordMetaDataProto.Index proto) throws KeyExpression.Dese
...
+            expr = toKeyWithValueExpression(expr, value);
+        }
+
+        final List<KeyExpressionVisitor> visitors = KeyExpressionVisitor.Registry.instance().visitors();
+        if (!visitors.isEmpty()) {
+            for (KeyExpressionVisitor visitor : visitors) {
+                if (visitor.applies(type)) {
+                    expr = visitor.visit(expr);
+                }
+            }
```
MMcM, line 196: An alternative to changing the serialized representation to recover information is to use `FunctionKeyExpressions`, which while opaque to most uses could be understood to have that additional Lucene information by the Lucene planner. A `TODO` for now is probably okay, as there is other work in this area outstanding.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
```diff
@@ -337,7 +337,7 @@ public boolean equals(Object o) {
             return true;
         }
 
-        if (o == null || getClass() != o.getClass()) {
+        if (o == null || !(o instanceof FieldKeyExpression)) {
```
MMcM, line 340: This impersonation by a subclass shouldn't be necessary. Probably another `TODO`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneTypeConverter.java
```diff
@@ -0,0 +1,82 @@
...
+        fanNext = false;
+        if (fke instanceof LuceneFieldKeyExpression) {
+            return fke;
+        }
+        return new LuceneFieldKeyExpression(fke, LuceneKeyExpression.FieldType.STRING, true, true);
+    }
+
+    @Override
+    public KeyExpression visitThen(final ThenKeyExpression thenKey) {
+        boolean hasPrimaryExpression = fanNext;
```
MMcM, line 52: This seems to be an accident of how the tests are arranged, that the only non-repeated nested message is for the primary key.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -154,42 +141,91 @@ private void parseFieldNames(KeyExpression expression, @Nonnull String parentFie
...
         if (indexEntries.isEmpty()) {
             return AsyncUtil.DONE;
         }
         try {
             final IndexWriter writer = getOrCreateIndexWriter(state, analyzer, executor);
+            Document document = new Document();
+            byte[] primaryKey = savedRecord.getPrimaryKey().pack();
+            BytesRef ref = new BytesRef(primaryKey);
+            document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
+            document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
```
MMcM, line 154: I could imagine wanting to store a multi-field primary-key as _separate_ document fields, so as to be able to query on just some of the components.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
sophiabuell: actually the primary fields are stored but as a byte ref. Thats how the doc_id is added in the tests because its not part of the root key expression, the issue isn't that we can't retrieve them but that they aren't indexed and searchable.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: It should, We need to think how to handle this since IntPoint is inherently not a stored field. Its also possible to have a sorted numeric field, there are a few ways of handling numeric values in Lucene depending on what you want them for.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneThenKeyExpression.java
sophiabuell: If it has a primary key then it does fan out repeatedly and that primary key value is used to prefix it. I think this is consistent with how we use Fan-Out right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: I guess thats a question of design too, should you have to specify the fields in the index definition that you want to query on or should we assume primary field keys are included.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneThenKeyExpression.java
MMcM: Yes, you are right. I think it still needs a comment explaining that this means that one of the children is a key-like field that gets made part of the full field name, if that's what it does.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
MMcM: I am guessing that we will need some level of control over how you intend the index to be used, though with sensible defaults.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneTypeConverter.java
```diff
@@ -0,0 +1,82 @@
...
+        List<KeyExpression> children = thenKey.getChildren().stream()
+                .map(this::visit).collect(Collectors.toList());
+        if (!hasPrimaryExpression) {
+            return new LuceneThenKeyExpression(null, children);
+        }
+        fanNext = false;
+        return new LuceneThenKeyExpression(children.get(0), children);
```
MMcM, line 59: Should be a comment here that (for now) we assume that the first child field is a (unique or nearly so) key among the child records. In the future, we will presumably need a way to control this, perhaps as an alternative to storing wide child records as block child documents.

MMcM (changes requested):

sophiabuell:
Jenkins test this please

MMcM (approved):

PR #1305, opened 2021-06-29
Resolves #1304: Expose transaction option setServerRequestTracing in context config

### LINKED ISSUE #1304: Expose transaction option setServerRequestTracing in context config
When the transaction option `setServerRequestTracing` is enabled, additional tracing is enabled throughout FDB. We should expose it as an option on `FDBRecordContextConfig` so that users can enable tracing on transactions created through the record layer. I believe that this option is new in 6.3?

Issue #1122 is about integrating more generally with distributed tracing, and maybe we should combine these two issues in some way.

alecgrieser:
This adds a configuration option to transactions that allows server request tracing to be enabled for that transaction. The ergonomics at the moment are a little bit weird, as if you set the option and provide a transaction identifier, then there's a single trace log message saying which random trace ID correlates with that transaction ID. However, individual sub-components of the transaction will have their own IDs (partially necessary to support many-to-one relationships between parents and children, which is required to accurately represent the effects of batching within FDB internals), so that only gets you so far. But this is what's required to make what is currently exposed work.

This resolves #1304.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
@@ -473,7 +473,7 @@ public class FDBRecordContextConfig {
 
         /**
          * Set whether FDB server request tracing is enabled for this transaction. If this flag is set to {@code true},
-         * this enables additional logging tracing each FDB server operation associated with this transaction in the
+         * this enables additional logging tracing for each FDB server operation associated with this transaction in the
          * FDB client and server logs. This can be useful for debugging performance problems, but it is also fairly
          * high overhead so it should be enabled sparingly. If a {@linkplain #setTransactionId(String) transaction ID}
          * is set, then the specified ID will be included in one of the client log messages in order to correlate the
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
```diff
@@ -445,6 +461,33 @@ public Builder setLogTransaction(final boolean logTransaction) {
...
+         * @see TransactionOptions#setServerRequestTracing()
+         * @see #setServerRequestTracing(boolean)
+         */
+        public boolean isServerRequestTracing() {
+            return serverRequestTracing;
+        }
+
+        /**
+         * Set whether FDB server request tracing is enabled for this transaction. If this flag is set to {@code true},
+         * this enables additional logging tracing each FDB server operation associated with this transaction in the
```
ScottDugas, line 476:
```suggestion
         * this enables additional logging tracing for each FDB server operation associated with this transaction in the
```

ScottDugas (changes requested):
It would be nice if we could assert that the transaction ended up with the right option, or that the logs contain the right info, but that would be a lot, for what is a fairly straightforward bit of code.

ScottDugas (approved):

PR #1308, opened 2021-07-01
Resolves #1307: RecordCursor.asList() with continuation support

### LINKED ISSUE #1307: RecordCursor.asList() should also return a continuation
A common pattern with see when using Record Layer is to compute a list from a `RecordCursor` along with a continuation. Which means we are doing logic like this a lot:

```
            final List<Long> values = new ArrayList<>();
            final AtomicReference<RecordCursorContinuation> finalContinuation = new AtomicReference<>(RecordCursorEndContinuation.END);
            context.asyncToSync(FDBStoreTimer.Waits.WAIT_THINGY,
                    AsyncUtil.whileTrue(() -> cursor.onNext().thenApply(nextResult -> {
                        if (nextResult.hasNext()) {
                            values.add(nextResult.get());
                            return true;
                        }
                        if (!nextResult.getNoNextReason().isSourceExhausted()) {
                            finalContinuation.set(nextResult.getContinuation());
                        }
                        return false;
                    }), context.getExecutor()));
```

it seems reasonable that we could present this as:

```
public <T> List<T> asList(RecordCursor<T> cursor, AtomicReference<RecordCursorContinuation> continuation);
```

in the `RecordCursor` class.

scgray:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -211,6 +211,29 @@ public void orElseTimingErrorTest() throws Exception {
...
+        int iterations = 0;
+        byte[] continuation = null;
+        do {
+            ++iterations;
+            List<Integer> values = RecordCursor.fromList(ints, continuation).limitRowsTo(10).asList(finalResult).get();
+            if (values.size() > 0) {
+                assertEquals(values.size(), 10);
+                assertEquals(values.get(0), (iterations - 1) * 10);
+                assertTrue(finalResult.get().getNoNextReason().isLimitReached());
+            }
```
ScottDugas, line 229: It would be good to assert about the final NoNextReason too
```suggestion
            } else {
                assertTrue(finalResult.get().getNoNextReason().isSourceExhausted());
            }
```

ScottDugas (changes requested):

scgray:
test this please

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=9af5c8e7e2cd1a3b0f694a4ac262b9f38c75262e74a9e8b5101af302a6beadd7
```
ScottDugas, line 3: What's with the change here?

ScottDugas (commented):
Looks good, but wondering about the gradle-wrapper change, and there's a conflict.

gradle/wrapper/gradle-wrapper.properties
scgray: Oh, that is accidental -- however I was meaning to ask the team if we should keep it.  Recent IntelliJ versions do NOT like having this present and will refuse to do gradle builds until you remove it.  I'll bring it up in a discussion and we can decide if it should stay here, otherwise I'll pull it out.

scgray:
test this please

ScottDugas (approved):

PR #1309, opened 2021-07-02
Adding instrumentation to FDBDirectory

scottfines:
This adds some simple instrumentation to the FDBDirectory so that we have some metric visibility into how Lucene is handling operations w.r.t FDB.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -26,7 +26,11 @@ import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.logging.CompletionExceptionLogHelper;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
+import com.apple.foundationdb.record.metadata.IndexTypes;
+import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer;
 import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.Tuple;
 import com.google.common.base.Verify;
@@ -57,6 +61,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
 
 /**
  * Directory implementation backed by FDB which attempts to
@@ -133,6 +138,7 @@ public class FDBDirectory extends Directory {
      * @return current increment value
      */
     public synchronized long getIncrement() {
+        incrementCallCount(FDBStoreTimer.Counts.LUCENE_GET_INCREMENT_CALLS);
         return context.ensureActive().get(sequenceSubspaceKey).thenApply(
             (value) -> {
                 if (value == null) {
@@ -144,6 +150,13 @@ public class FDBDirectory extends Directory {
                     return sequence;
                 }
             }).join();
+    }
+
+    private void incrementCallCount(FDBStoreTimer.Counts counter) {
+        final FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            timer.increment(counter);
+        }
     }
 
     /**
@@ -161,15 +174,15 @@ public class FDBDirectory extends Directory {
         LOGGER.trace("getFDBLuceneFileReference {}", name);
         FDBLuceneFileReference fileReference = this.fileReferenceCache.getIfPresent(name);
         if (fileReference == null) {
-            return context.ensureActive().get(metaSubspace.pack(name))
+            return context.instrument(FDBStoreTimer.Events.LUCENE_GET_FILE_REFERENCE, context.ensureActive().get(metaSubspace.pack(name))
                     .thenApplyAsync((value) -> {
-                            FDBLuceneFileReference fetchedref = value == null ? null : new FDBLuceneFileReference(Tuple.fromBytes(value));
-                            if (fetchedref != null) {
-                                this.fileReferenceCache.put(name, fetchedref);
+                                FDBLuceneFileReference fetchedref = value == null ? null : new FDBLuceneFileReference(Tuple.fromBytes(value));
+                                if (fetchedref != null) {
+                                    this.fileReferenceCache.put(name, fetchedref);
+                                }
+                                return fetchedref;
                             }
-                            return fetchedref;
-                        }
-                    );
+                    ), System.nanoTime());
         } else {
             return CompletableFuture.completedFuture(fileReference);
         }
@@ -182,6 +195,7 @@ public class FDBDirectory extends Directory {
      */
     public void writeFDBLuceneFileReference(@Nonnull final String name, @Nonnull final FDBLuceneFileReference reference) {
         LOGGER.trace("writeFDBLuceneFileReference {}", reference);
+        incrementCallCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE);
         context.ensureActive().set(metaSubspace.pack(name), reference.getTuple().pack());
         fileReferenceCache.put(name, reference);
     }
@@ -193,6 +207,11 @@ public class FDBDirectory extends Directory {
      * @param value the data to be stored
      */
     public void writeData(long id, int block, @Nonnull byte[] value) {
+        FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            //This may not be correct transactionally
+            timer.increment(FDBStoreTimer.Counts.LUCENE_WRITE_SIZE, value.length);
+        }
         LOGGER.trace("writeData id={}, block={}, valueSize={}", id, block, value.length);
         Verify.verify(value.length <= blockSize);
         context.ensureActive().set(dataSubspace.pack(Tuple.from(id, block)), value);
@@ -217,11 +236,11 @@ public class FDBDirectory extends Directory {
                 throw new RecordCoreArgumentException(String.format("No reference with name %s was found", resourceDescription));
             }
             Long id = reference.getId();
-            return blockCache.get(Pair.of(id, block),
-                    () -> {
-                        return context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block)));
-                    }
-            );
+
+            long start = System.nanoTime();
+            return context.instrument(FDBStoreTimer.Events.LUCENE_READ_BLOCK,blockCache.get(Pair.of(id, block),
+                    () -> context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block)))
+            ), start);
         } catch (ExecutionException e) {
             throw new RecordCoreException(CompletionExceptionLogHelper.asCause(e));
         }
@@ -237,6 +256,26 @@ public class FDBDirectory extends Directory {
     @Override
     @Nonnull
     public String[] listAll() {
+        List<String> outList;
+        long start = System.nanoTime();
+        try {
+            outList = listAllInternal();
+        } finally {
+            record(FDBStoreTimer.Events.LUCENE_LIST_ALL, System.nanoTime() - start);
+        }
+        //noinspection ToArrayCallWithZeroLengthArrayArgument
+        return outList.toArray(new String[outList.size()]);
+    }
+
+    private void record(StoreTimer.Event event, long durationNanos) {
+        final FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            timer.record(event, durationNanos);
+        }
+    }
+
+    private List<String> listAllInternal() {
+        //A private form of the method to allow easier instrumentation
         List<String> outList = new ArrayList<>();
         List<String> displayList = null;
 
@@ -268,8 +307,7 @@ public class FDBDirectory extends Directory {
             LOGGER.debug("listAllFiles -> count={}, totalSize={}", outList.size(), totalSize);
             LOGGER.debug("Files -> {}", displayList);
         }
-
-        return outList.toArray(new String[0]);
+        return outList;
     }
 
     /**
@@ -278,8 +316,9 @@ public class FDBDirectory extends Directory {
      */
     @Override
     public void deleteFile(@Nonnull String name) throws IOException {
+
         LOGGER.trace("deleteFile -> {}", name);
-        boolean deleted = getFDBLuceneFileReference(name).thenApplyAsync(
+        boolean deleted = context.asyncToSync(FDBStoreTimer.Waits.WAIT_LUCENE_DELETE_FILE, getFDBLuceneFileReference(name).thenApplyAsync(
                 (value) -> {
                     if (value == null) {
                         return false;
@@ -289,7 +328,8 @@ public class FDBDirectory extends Directory {
                     this.fileReferenceCache.invalidate(name);
                     return true;
                 }
-        ).join();
+        ));
+
         if (!deleted) {
             throw new NoSuchFileException(name);
         }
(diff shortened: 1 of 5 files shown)


scottfines:
test this please

scottfines:
test this please

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -182,7 +192,14 @@ public synchronized long getIncrement() {
      */
     public void writeFDBLuceneFileReference(@Nonnull final String name, @Nonnull final FDBLuceneFileReference reference) {
         LOGGER.trace("writeFDBLuceneFileReference {}", reference);
-        context.ensureActive().set(metaSubspace.pack(name), reference.getTuple().pack());
+        final FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            long start = System.nanoTime();
+            context.ensureActive().set(metaSubspace.pack(name), reference.getTuple().pack());
+            timer.record(FDBStoreTimer.Events.LUCENE_WRITE_FILE_REFERENCE, System.nanoTime() - start);
```
scgray, line 199: This isn't going to tell you anything useful.  Writes are held purely in the client until the transaction is committed. As such, the time measured for a write will always be almost zero.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -193,6 +210,12 @@ public void writeFDBLuceneFileReference(@Nonnull final String name, @Nonnull fin
      * @param value the data to be stored
      */
     public void writeData(long id, int block, @Nonnull byte[] value) {
+        FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            timer.increment(FDBStoreTimer.Counts.LUCENE_DATA_WRITE);
+            //This may not be correct transactionally
+            timer.increment(FDBStoreTimer.Counts.LUCENE_DATA_SIZE, value.length);
```
scgray, line 217: There is no need for two metrics here, you can just send `LUCENE_DATA_SIZE`.  In Hubble, plotting the `count/sec` of the `LUCENE_DATA_SIZE` will implicitly be the number of writes.  You may, however, want to change the metric to `LUCENE_WRITE_SIZE` to clarify the meaning of the metric.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -217,11 +240,11 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
...
-            return blockCache.get(Pair.of(id, block),
-                    () -> {
-                        return context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block)));
-                    }
-            );
+
+            long start = System.nanoTime();
+            return context.instrument(FDBStoreTimer.Events.LUCENE_READ_BLOCK,blockCache.get(Pair.of(id, block),
+                    () -> context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block)))
+            ),start);
```
scgray, line 247: Super minor, but:  `), start);`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -237,6 +260,21 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
...
+        final FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            long start = System.nanoTime();
+            outList = listAllInternal();
+            timer.record(FDBStoreTimer.Events.LUCENE_LIST_ALL, System.nanoTime() - start);
+        } else {
+            outList = listAllInternal();
+        }
+        //noinspection ToArrayCallWithZeroLengthArrayArgument
+        return outList.toArray(new String[outList.size()]);
```
scgray, line 273: You have a method, above, `incrementCallCount` to centralize the handling of the timer (one of my BIGGEST irritations is that the store timer can be `null` -- we really should have had a no-op implementation of the timer that can stand in for `null`, but ... here we are).  I think it would make sense to similarly have:

```
private void record(StoreTimer.Event event, long durationNanos) {
    final StoreTimer timer = context.getTimer();
    if (timer != null) {
        time.record(event, duration);
    }
}
```

this will greatly simplify this sort of code to just:

```
long start = System.nanoTime()
try {
    return listAllInternal();
} finally {
    record(FDBStoreTimer.Events.LUCENE_LIST_ALL, System.nanoTime() - start);
}
```

I think it's fine to live with the overhead of fetching the time even when a store timer isn't present.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -279,7 +316,8 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
     @Override
     public void deleteFile(@Nonnull String name) throws IOException {
         LOGGER.trace("deleteFile -> {}", name);
-        boolean deleted = getFDBLuceneFileReference(name).thenApplyAsync(
+        long start = System.nanoTime();
+        boolean deleted = context.instrument(FDBStoreTimer.Events.LUCENE_DELETE_FILE,getFDBLuceneFileReference(name).thenApplyAsync(
```
scgray, line 320: Similar to writes, deletes are cached entirely in the client until commit, so I think the only thing you'll really be timing here is `getFDBLuceneFileReference()`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -289,7 +327,8 @@ public void deleteFile(@Nonnull String name) throws IOException {
                     this.fileReferenceCache.invalidate(name);
                     return true;
                 }
-        ).join();
+        ),start).join();
```
scgray, line 330: You shouldn't be using `join()` -- pretty much anywhere.  This is why we have `WAIT` timers.  So the code, should be something like:

    boolean deleted = context.asyncToSync(FDBStoreTimer.Waits.WAIT_LUCENE_DELETE_FILE, 
        getFDBLuceneFileReference(name).thenApply(.... ));

`join()` can block indefinitely, and the `asyncToSync` infrastructure buys you a lot of error handling and other benefits.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scgray: I know this isn't part of this PR, but you don't need (or want) to use `thenApplyAsync()`.  Most `*Async()` calls like this, will create a new future that will execute the code in the lambda -- so the future is created then enqueued for execution in the `Executor` threads.  This is typically only done if the lambda is costly and you don't want this chain of futures to wait for its completion.  In this case (and, by far our most common cases), the cost of the labda execution is minimal (since `clear()` doesn't involve any actual I/O -- and, if it did, it would have returned a future), so it is way more costly do have `thenApplyAsync()` dispatch a new work item, than to just use `thenApply()`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -279,7 +316,8 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
     @Override
     public void deleteFile(@Nonnull String name) throws IOException {
```
scgray, line 317: This isn't part of this PR, but it looks like every call in `FDBDIrectory` is synchronous -- meaning you `join()` or `asyncToSync()` before returning. Hopefully the methods are never called by any asynchronous code?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scgray: I know this is super pendantic, but please try to put spaces after commas, like:

    instrument(FDBStoreTimer.Events.LUCENE_DELETE_FILE, getFDBLuceneFileReference(name)

the main thing (for me) is when reviewing, without the space, it looks like a single expression.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -304,7 +343,7 @@ public void deleteFile(@Nonnull String name) throws IOException {
     @Override
     public long fileLength(@Nonnull String name) throws NoSuchFileException {
         LOGGER.trace("fileLength -> {}", name);
-        FDBLuceneFileReference reference = getFDBLuceneFileReference(name).join();
+        FDBLuceneFileReference reference = context.instrument(FDBStoreTimer.Events.LUCENE_FILE_LENGTH,getFDBLuceneFileReference(name)).join();
```
scgray, line 346: As mentioned above, this should probably be `asyncToSync()` and be a `Wait` instead of an `Event`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -368,10 +407,11 @@ public void rename(@Nonnull final String source, @Nonnull final String dest) {
         LOGGER.trace("rename -> source={}, dest={}", source, dest);
         final byte[] key = metaSubspace.pack(source);
         CompletableFuture<byte[]> completableFuture;
-        completableFuture = context.ensureActive().get(key).exceptionally( (e) -> {
-            throw new RecordCoreArgumentException("Invalid source name in rename function for source: " + source);
-        });
-        completableFuture.thenAcceptAsync( (value) -> {
+        completableFuture = context.instrument(FDBStoreTimer.Events.LUCENE_RENAME,
+                context.ensureActive().get(key).exceptionally((e) -> {
```
scgray, line 411: This should, again, be a `Wait`, also this is only timing the read of the `key`, not the actual rename.

I think you wanted:

```
context.asyncToSync(FDBStoreTimer.Waits.WAIT_LUCENE_RENAME,
    context.ensureActive().get(key).thenApply(value -> {
             this.fileReferenceCache.invalidate(source);
             context.ensureActive().set(metaSubspace.pack(dest), value);
             context.ensureActive().clear(key);
        }));
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -368,10 +407,11 @@ public void rename(@Nonnull final String source, @Nonnull final String dest) {
...
+                    throw new RecordCoreArgumentException("Invalid source name in rename function for source: " + source);
```
scgray, line 412: `get()` doesn't throw an exception unless there is an actual I/O problem or some other problem at the server...it just returns `null` if the value isn't there, so I don't think this exception is actually correct.

scgray (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: this was actually a fun little unwrapping problem--when switching to `asyncToSync` the fact that this error wasn't being thrown (in fact an `IllegalArgumentException` was) became clear :)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: Removed, thanks

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: I _think_ I got all of these, but since they mostly just go completely through my brain without my being able to tell them, I may have missed a couple--my preemptive apologies :)

scottfines:
test this please

scottfines:
test this please

scottfines:
test this please

scottfines:
test this please

scottfines:
test this please

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -193,6 +205,11 @@ public void writeFDBLuceneFileReference(@Nonnull final String name, @Nonnull fin
      * @param value the data to be stored
      */
     public void writeData(long id, int block, @Nonnull byte[] value) {
+        FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            //This may not be correct transactionally
+            timer.increment(FDBStoreTimer.Counts.LUCENE_DATA_SIZE, value.length);
```
scgray, line 211: I still think this should be `LUCENE_WRITE_SIZE` - it is useful to know the type of operation for which the size is being reported.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -367,15 +405,16 @@ public void syncMetaData() throws IOException {
...
         final byte[] key = metaSubspace.pack(source);
-        CompletableFuture<byte[]> completableFuture;
-        completableFuture = context.ensureActive().get(key).exceptionally( (e) -> {
-            throw new RecordCoreArgumentException("Invalid source name in rename function for source: " + source);
-        });
-        completableFuture.thenAcceptAsync( (value) -> {
-            this.fileReferenceCache.invalidate(source);
+        context.asyncToSync(FDBStoreTimer.Waits.LUCENE_RENAME, context.ensureActive().get(key).thenApply((Function<byte[], Void>)value -> {
+            if (value == null) {
+                throw new RecordCoreArgumentException("Invalid source name in rename function for source: " + source);
```
scgray, line 410: Feel free to ignore this for now, but we have a few general rules for our exceptions:

* The text of the exception should be static (no variables in it)
* Logging properties should be added to give context to the error
* The logging properties should provide as much detail as possible to diagnose the source of the error

The idea here is that by having the text of the exception static, that it makes it trivial to search through all logs in the system and find all instances of the error, the extra context information (in this case "source"), can then be used for grouping/filtering/etc. without any regex games when searching. 

Also, when you are testing, it may be trivial to understand the context of this message (maybe it is a single record store with only one Lucene index), but as things scale, and you get millions of databases and maybe even some with multiple lucene indexes, having context on at least which index it is can be helpful.

This is all a long winded way of saying, probably this should be:

```
throw new RecordCoreArgumentException("Invalid source name in rename function for source")
    .addLogInfo(LogMessageKeys.SOURCE,  source);
```

and, ideally, you probably want additional `LogMessageKey`s that tell you which index, which key, and any other information that would be useful for understanding the context of the issue.

I really need add this info into the contributors guide :)

scgray (changes requested):
Only minor comments, however I wanted to point out that I was right about the blocking in async issue...it turns out that the lucene test harnesses don't have our checking enabled, but if you add this to `fdb-record-layer-lucene.gradle `:

```
test {
    useJUnitPlatform {
        ...
        // Configure whether or not tests will validate that asyncToSync isn't being called in async
        // context.  See BlockingInAsyncDetection class for details on values.
        systemProperties = [ "com.apple.foundationdb.record.blockingInAsyncDetection":
            System.getenv('BLOCKING_DETECTION') ?: "IGNORE_COMPLETE_EXCEPTION_BLOCKING" ]
    }
}
```

The tests absolutely explode. I'll open an issue about this.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: I attempted to address this (in a rough-and-ready sort of fashion). Lemme know how well I did :)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: done.

scgray (approved):

PR #1312, opened 2021-07-07
resolves #1311: constrain the amount of enumerations the new planner …

### LINKED ISSUE #1311: constrain the amount of enumerations the new planner can do
Currently, enumerations of matches per rule call are not constraint in any way. A query that causes many variations to be created for a particular group could cause performance issues due to over enumeration. While hypothetical cases like that should be fixed for each individual case, this issue should just address a fail safe to abort planning when over enumeration occurs.

normen662:
…can do

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
@@ -50,6 +50,7 @@ import javax.annotation.Nonnull;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
```diff
@@ -348,6 +349,16 @@ public void setMaxTotalTaskCount(final int maxTotalTaskCount) {
                 .build();
     }
 
+    /**
+     * Set the maximum number of yields that are permitted per rule call within the Cascades planner.
+     * @param maxNumYieldsPerRuleCall the desired maximum number of yields that are permitted per rule call
```
ohadzeliger, line 354: Add statement about '0' being "unrestricted"

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
```diff
@@ -668,18 +683,21 @@ protected boolean shouldExecute() {
...
             final PlannerBindings initialBindings = getInitialBindings();
 
+            final AtomicInteger numMatches = new AtomicInteger(0);
+
             rule.getMatcher()
                     .bindMatches(initialBindings, getBindable())
                     .map(bindings -> new CascadesRuleCall(getContext(), rule, group, aliasResolver, bindings))
                     .forEach(ruleCall -> {
+                        if (isMaxNumMatchesPerRuleCallExceeded(configuration, numMatches.incrementAndGet())) {
+                            throw new RecordQueryPlanComplexityException("Maximum number of matches per rule call (" + configuration.getMaxNumMatchesPerRuleCall() + ") is exceeded");
```
ohadzeliger, line 699: Add the offending rule?

ohadzeliger (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesPlanner.java
normen662: done

ohadzeliger (approved):

Issue #1180, opened 2021-03-13
Allow cursors to stop rather than throw an error on transaction timeouts or retriable errors


alecgrieser:
For some use cases, it would be useful to have an `ExecuteState` option that would allow cursors to stop when they hit a transaction timeout (at least) or perhaps any retriable FDB exception rather than rethrowing the error. For example, in the case where one is doing an inconsistent read, it would make sense to do something where the transaction is allowed to just read until it hits the [five second transaction time limit](https://apple.github.io/foundationdb/known-limitations.html#long-running-transactions), then restart the transaction from there.

I would propose adding a new option to the `ExecuteProperties` à la `failOnScanLimitReached`. If that option is set, then the idea is that cursors should catch any transaction time outs (or perhaps any retriable errors in general) and return a `RecordCursorResult` with a new `NoNextReason`, something like `TRANSACTION_TIMEOUT` or `RETRIABLE_ERROR`, which indicates that the scan terminated early. (This `NoNextReason` would be an out-of-band exception just like, say, the `TIME_LIMIT_REACHED` exception.) The user could then resume execution from where the last cursor left off.

I think adding support for this new termination condition could be done by only editing a few different places, chiefly the `KeyValueCursor` and maybe some of the other cursors that talk directly to FDB rather than going through the `KeyValueCursor`. This is more-or-less the same set of places that might use the `CursorScanLimiter` anyway, I think.

There are some possible draw backs to this, mainly that if there's some kind of problem where FDB is down or has an incompatible version, it can result in retriable errors, which may just mean that the user would want to limit the number of retries that are done on a `RETRIABLE_ERROR`. However, plumbing through the error may be difficult with the current `RecordCursorResult` API (as the `NoNextReason` is just an enum), so the best we may be able to get is logging the error.

alecgrieser:
A few things came up during a discussion about this with some members of the team:

1. There are probably a few other places that we'd need to consider adding error checking. For example, the `MapPipelinedCursor` could hit a `transaction_too_old` error when resolving a record, so it's possible that any cursor that would ever do database work in any of its callbacks could hit this
2. It's possible that much the same thing could be achieved with an an "error translating cursor" that took a base cursor and had an `onNext` method that caught a `transaction_too_old` (or whatever) and translated the cursor element into a `RecordCursorResult` with the proper `NoNextReason`. This could also limit the surface area of this change by limiting the error catching to a single place.
[...]

PR #1317, opened 2021-07-13
Resolves #1316 & #1180:  Integrity validator for LocatableResolvers

### LINKED ISSUE #1316: Validator for LocatableResolvers
We need a way to validate LocatableResolvers, specifically:

* Check to see if the reverse directory is missing entries 
* Check to see that entries in the reverse directory point to the correct forward entry

scgray:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -0,0 +1,301 @@
...
+     * mapping key. For example, if {@code "foo" -&gt; 10} is in the forward mapping and {@code "foo"} is
+     * not found in the reverse mapping, it will be returned as a bad entry. Similarly, if {@code "foo"} is
+     * found in the reverse mapping, but it has a value of {@code 30}, then that is considered an error.
+     *
+     * @param resolver the resolver to scan
+     * @param context the transaction context to use
+     * @param continuation continuation from a previous scan
+     * @param reverseLookupPipelineSize the number of reverse directory lookups to perform in parallel while
+     *   scanning the forward mappings
+     * @param repairMissingEntries if {@code true}, missing reverse directory entries will be restored
```
alecgrieser, line 125: It's probably worth noting somewhere that it's still the caller's responsibility to commit the context to actually repair any missing entries (if I'm reading everything correctly here). I assume that the difficulty of hooking that into the `AutoContinuingCursor` is partially why the synchronous "validate in multiple transactions" method doesn't have a `repairMissingEntries` parameter?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -0,0 +1,301 @@
...
+        return validate(resolver, context, continuation, 10, repairMissingEntries, scanProperties);
+    }
+
+    /**
+     * Validate entries in a {@link LocatableResolver}'s mappings. This method scans the forward mapping
+     * ({@code String} -&gt; {@code long}) and validates that the entry has a corresponding entry in the
+     * reverse mapping, as well as validating that the entry in the reverse mapping points to the forward
+     * mapping key. For example, if {@code "foo" -&gt; 10} is in the forward mapping and {@code "foo"} is
+     * not found in the reverse mapping, it will be returned as a bad entry. Similarly, if {@code "foo"} is
+     * found in the reverse mapping, but it has a value of {@code 30}, then that is considered an error.
```
alecgrieser, line 118: Hm, I think the documentation is a bit misleading here on what situations can lead to a `REVERSE_ENTRY_MISMATCH` result. It seems to imply that if the directory layer has a mapping of "foo" -> 10 and the reverse directory layer has a mapping from 30 -> foo, then this would catch that and throw an error. But I think that the _actual_ error that it catches is if the reverse directory layer has a mapping of 10 -> bar. I think 30 -> foo would go unnoticed, at least with current validation (and would require a second validator to make sure that every entry in the reverse directory layer matched the forward directory layer to be done efficiently).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -0,0 +1,301 @@
...
+     */
+    public static class ValidatedEntry {
+        @Nonnull
+        private final ValidationResult result;
+        @Nonnull
+        private final ResolverKeyValue keyValue;
+        @Nonnull
+        private final String reverseValue;
+
+        public ValidatedEntry(@Nonnull final ValidationResult result,
```
alecgrieser, line 227: These constructors should probably be `private`, or at least `INTERNAL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -0,0 +1,301 @@
...
+        @Override
+        public boolean equals(final Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            final ValidatedEntry validatedEntry = (ValidatedEntry)o;
+            return result == validatedEntry.result && keyValue.equals(validatedEntry.keyValue) && Objects.equals(reverseValue, validatedEntry.reverseValue);
```
alecgrieser, line 284: Was `reverseValue` initially `@Nullable`? It seems like this should either use `Objects::equals` on both the `keyValue` and `reverseValue` fields or `Object::equals` for both

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
```diff
@@ -126,6 +127,25 @@ public static ScopedInterningLayer global(@Nonnull FDBDatabase database) {
...
+                });
+    }
+
+    @Override
+    protected CompletableFuture<Void> putReverse(@Nonnull final FDBRecordContext context, final long value, @Nonnull final String key) {
+        return interningLayerFuture
+                .thenApply(layer -> {
+                    layer.putReverse(context, value, key);
+                    return null;
+                });
```
alecgrieser, line 146:
```suggestion
                .thenAccept(layer -> layer.putReverse(context, value, key));
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
```diff
@@ -126,6 +127,25 @@ public static ScopedInterningLayer global(@Nonnull FDBDatabase database) {
...
     }
 
+    @Override
+    @VisibleForTesting
+    protected CompletableFuture<Void> deleteReverseForTesting(FDBRecordContext context, final long value) {
+        return interningLayerFuture
+                .thenApply(layer -> {
+                    layer.deleteReverseForTesting(context, value);
+                    return null;
+                });
```
alecgrieser, line 137:
```suggestion
                .thenAccept(layer -> layer.deleteReverseForTesting(context, value));
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/AutoContinuingCursor.java
```diff
@@ -72,6 +74,8 @@
 public class AutoContinuingCursor<T> implements RecordCursor<T> {
     private static final Logger LOGGER = LoggerFactory.getLogger(AutoContinuingCursor.class);
 
+    private static final int MAX_RETRY_ON_EXCEPTION_ATTEMPTS = 3;
```
alecgrieser, line 77: Is this used?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/cursors/AutoContinuingCursorTest.java
```diff
@@ -54,28 +64,82 @@ public void getDatabase() {
...
+        CompletionException e = assertThrows(CompletionException.class, () ->
+            testAutoContinuingCursorGivenCursorGenerator((context, continuation) ->
+                    new TestingListCursor<>(list, continuation, () -> {
+                        exceptionCount.incrementAndGet();
+                        CompletableFuture<Void> failedFuture = new CompletableFuture<>();
+                        failedFuture.completeExceptionally(new FDBException("transaction_too_old", FDBError.TRANSACTION_TOO_OLD.code()));
+                        return failedFuture;
+                    }), 3, list));
+        assertEquals(((FDBException) e.getCause()).getCode(), FDBError.TRANSACTION_TOO_OLD.code());
+        assertEquals(exceptionCount.get(), 4);
```
alecgrieser, line 127: If the max attempts is 3 but the exception count is 4, does that indicate there's an off-by-one error somewhere? (Or that the parameter needs to be renamed to "max retries" instead of "max attempts"?)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -0,0 +1,301 @@
...
+                                }
+
+                                if (!reverseKey.equals(keyValue.getKey())) {
+                                    return new ValidatedEntry(ValidationResult.REVERSE_ENTRY_MISMATCH, keyValue, reverseKey);
+                                }
+
+                                return new ValidatedEntry(ValidationResult.OK, keyValue);
+                            }).thenCompose(validatedEntry -> {
+                                if (repairMissingEntries
+                                        && validatedEntry.getValidationResult() == ValidationResult.REVERSE_ENTRY_MISSING) {
```
alecgrieser, line 165: In theory, this could also repair `REVERSE_ENTRY_MISMATCH` errors by replacing the wrong key with the right key in the reverse mapping. Is the idea that that represents something "worse" than just a missing entry (which, for example, could be caused by something like resolving something in a directory layer prior to the reverse directory layer having been added)? (The most plausible cause for a `REVERSE_ENTRY_MISMATCH`, off the top of my head, would be something like a concurrency bug that would accidentally result in two directory layer entries being mapped to the same value, which seems bad.)

alecgrieser (changes requested):

scgray:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
alecgrieser: Hmm, that's strange. `.thenAccept` takes a consumer, so a `void` lambda should be able to be automatically translated into a `Void`. But this is minor, so I guess it's not super important.

alecgrieser (approved):

PR #1320, opened 2021-07-14
Resolves #1319: PlanOrderingKey for Intersection / (ordered) Union

### LINKED ISSUE #1319: PlanOrderingKey for Intersection / (ordered) Union
Since these operators merge according to a `comparisonKey`, their ordering key is that.

This is actually useful for an OR of an AND where the AND turns into an intersection. Without it, `planOr` can't do `planOrderedUnion`.

MMcM:


normen662 (approved):

MMcM:
test this please

PR #1321, opened 2021-07-15
Start documenting coding style

scgray:


docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+```
+LOGGER.info(KeyValueLogMessage.of("Unable to open file",
+    LogMessageKeys.FILENAME, filename));
+```
+
+## Futures
+
+### `join()` or `get()` in production code
+
+Production code (i.e. code that aren't tests), should never call `CompletableFuture.join()`
```
alecgrieser, line 84:
```suggestion
Production code (i.e., code that isn't test code), should never call `CompletableFuture.join()` 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+LOGGER.info(KeyValueLogMessage.of("Unable to open file",
+    LogMessageKeys.FILENAME, filename));
+```
+
+## Futures
+
+### `join()` or `get()` in production code
+
+Production code (i.e. code that aren't tests), should never call `CompletableFuture.join()` 
+or `CompletableFuture.get()`, but instead should call `FDBRecordContext.asyncToSync()` or
```
alecgrieser, line 85:
```suggestion
or `CompletableFuture.get()`, but instead it should call `FDBRecordContext.asyncToSync()` or
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+`FDBDatabase.asyncToSync()`.  The `asyncToSync()` method provides a number of important
+benefits:
+
+* It establishes a client-configurable configurable time limit on how long we are
```
alecgrieser, line 89:
```suggestion
* It establishes a client-configurable time limit on how long we are 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+  to track the time spent waiting on the event
+* It takes care of error handling and wrapping to promote thrown exceptions to the 
+  appropriate `RecordCoreException`
+* It has the ability to detect if `asyncToSync()` is being called from within a 
+  completing future (see **Blocking in asyncronous code**, below)
+
+### Blocking in asynchronous code
+
+The methods `join()`, `get()`, `asyncToSync()` or any method that is known to block 
+(such as syncronous I/O methods), just never be called from within a completable
```
alecgrieser, line 101:
```suggestion
(such as syncronous I/O methods) should never be called from within a completable 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+  completing future (see **Blocking in asyncronous code**, below)
+
+### Blocking in asynchronous code
+
+The methods `join()`, `get()`, `asyncToSync()` or any method that is known to block 
+(such as syncronous I/O methods), just never be called from within a completable 
+future.
+
+The thread pool `Executor` has a limited number of threads that are used to invoke 
+the lambda in the `CompletableFuture` when it completes, so if that completable
```
alecgrieser, line 105:
```suggestion
the lambda in the `CompletableFuture` when it completes, so if that future
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+makes a blocking call, it has permenantly consumed the `Executor` thread until 
+the completion of that operation. In the case where the blocking call is waiting
+for another future to complete (e.g. `asyncToSync()`, `join()` or `get()`) you
```
alecgrieser, line 108:
```suggestion
for another future to complete (e.g., `asyncToSync()`, `join()`, or `get()`), you
```

I suppose the Oxford comma is a bit of a style thing. Maybe we need a comma style document explaining our stance on it 🙃

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+can end up in a situation where the thread pool is entirely consumed with futures
+blocking for completions, and no threads are available to complete those requests
```
alecgrieser, line 110:
```suggestion
blocking for completions, and no threads are available to complete those requests,
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+and the entire system deadlocks. 
+
+In the Record Layer unit tests (any test extending `FDBTestBase`, the `asyncToSync()`
```
alecgrieser, line 113:
```suggestion
In the Record Layer unit tests (any test extending `FDBTestBase`), the `asyncToSync()`
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+call will automatically attempt to detect situations in which it finds itself being
+invoked from within a `CompletableFuture` completion, and throw an exception.
```
alecgrieser, line 115:
```suggestion
invoked from within a `CompletableFuture` completion and throw an exception.
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+
+### `.thenApplyAsync` and `.thenComposeAsync`
+
+In most situations, you do not want to use these methods, but should use `thenApply()` 
+or `thenCompose()`. The `CompletableFuture. *Async()` create and enqueue a new future
```
alecgrieser, line 120:
```suggestion
or `thenCompose()`. The `CompletableFuture.*Async()` methods create and enqueue a new future 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+
+In the Record Layer unit tests (any test extending `FDBTestBase`, the `asyncToSync()`
+call will automatically attempt to detect situations in which it finds itself being
+invoked from within a `CompletableFuture` completion, and throw an exception.
+
+### `.thenApplyAsync` and `.thenComposeAsync`
+
+In most situations, you do not want to use these methods, but should use `thenApply()` 
+or `thenCompose()`. The `CompletableFuture. *Async()` create and enqueue a new future 
+that will execute the code in the lambda, rather than executing the code in the lambda
```
alecgrieser, line 121:
```suggestion
that will execute the code in the lambda in a separate thread, rather than executing the code in the lambda 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+In the Record Layer unit tests (any test extending `FDBTestBase`, the `asyncToSync()`
+call will automatically attempt to detect situations in which it finds itself being
+invoked from within a `CompletableFuture` completion, and throw an exception.
+
+### `.thenApplyAsync` and `.thenComposeAsync`
+
+In most situations, you do not want to use these methods, but should use `thenApply()` 
+or `thenCompose()`. The `CompletableFuture. *Async()` create and enqueue a new future 
+that will execute the code in the lambda, rather than executing the code in the lambda 
+immediately upon completion of this future completes. For example:
```
alecgrieser, line 122:
```suggestion
immediately upon completion of this future. For example:
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+fetchSomeNumber().thenApplyAsync(number -> number * 10);
+```
+
+will create an entirely new future that is enqueued on the `Executor` thread, simply to
```
alecgrieser, line 128:
```suggestion
will create an entirely new future that is enqueued on a separate thread, simply to 
```

The reason I suggest being a little more vague about what thread the task is going to be scheduled on is because this call doesn't actually specify an executor. We could point out that it schedules the task on the common fork join pool, but I think we don't want to here, because I think the point you're trying to make is something like:

1. Prefer `.thenApply` over `.thenApplyAsync` (etc) in general to avoid thread hopping
2. Always use `.thenApplyAsync` with an executor instead of `.thenApplyAsync` without an executor (etc) to avoid scheduling work on the common thread pool

Which are two separate (but important) points.

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+do the multiplication of the number. Whereas:
+
+```
+fetchSomeNumber().thenApply(number -> number * 10);
+```
+
+Will do the calculation immediately, as soon as `number` becomes available.  So the 
+former form is FAR more expensive, for very little gain.  The main use of the `*Async()` 
+methods are cases where the lambda needs to do something relatively expensive (e.g. factor 
+the number) and you don't want to block the current future chain while it happens.
```
alecgrieser, line 138:
```suggestion
the number), and you don't want to block the current future chain while it happens. 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+will create an entirely new future that is enqueued on the `Executor` thread, simply to 
+do the multiplication of the number. Whereas:
+
+```
+fetchSomeNumber().thenApply(number -> number * 10);
+```
+
+Will do the calculation immediately, as soon as `number` becomes available.  So the 
+former form is FAR more expensive, for very little gain.  The main use of the `*Async()` 
+methods are cases where the lambda needs to do something relatively expensive (e.g. factor
```
alecgrieser, line 137:
```suggestion
methods are cases where the lambda needs to do something relatively expensive (e.g., factor 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+the number) and you don't want to block the current future chain while it happens. 
+
+Further, in the cases where you *do* need to use the `*Async()` methods, be sure you 
+pass it an `Executor`, like:
+
+```
+fetchSomeNumber().thenApplyAsync(number -> number * 10, context.getExecutor());
+```
+
+otherwise it is executed in the global thread pool, rather than the Record Layer
```
alecgrieser, line 147:
```suggestion
Otherwise it is executed in the global thread pool, rather than the Record Layer 
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+
+Further, in the cases where you *do* need to use the `*Async()` methods, be sure you 
+pass it an `Executor`, like:
+
+```
+fetchSomeNumber().thenApplyAsync(number -> number * 10, context.getExecutor());
+```
+
+otherwise it is executed in the global thread pool, rather than the Record Layer 
+thread pool
```
alecgrieser, line 148:
```suggestion
thread pool.
```

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
+# FoundationDB Record Layer Coding Style
+
+This document is very much a work in progress, but the intention is to detail rules
+both our code formatting style rules, as well as general paradigms that we follow
```
alecgrieser, line 4: > both our code formatting style rules

Is this true? It seems like our code formatting should (as much as possible) be in the automatic code styling rules, though maybe that's not always possible. It seems like this document is more about practices for general paradigms which are harder to enforce automatically (though maybe not impossible).

docs/Coding_Style.md
```diff
@@ -0,0 +1,148 @@
...
+    LogMessageKeys.PRIMARY_KEY, record.getPrimaryKey(),
+    LogMessageKeys.RECORD_TYPE, record.getRecordType());
+```
+
+By structuring our exceptions this way, we get a number of nice benefits:
+
+* We can easily search our logs for all occurrances of a given exception, 
+  such as finding how often "Found corrupt record" occurs
+* For a given exception, we can filter or group on the logging properties,
+  such as counting the number of correupt records by record type
```
alecgrieser, line 49:
```suggestion
  such as counting the number of corrupt records by record type
```

alecgrieser (changes requested):
Most of these are typos etc that hopefully aren't too hard to apply (if you agree). There's a little bit more that's somewhat more editorial regarding executors and async callbacks, but hopefully that's not too controversial

docs/Coding_Style.md
scgray: You are correct, I switched it to a "Best Practices" slant.

alecgrieser (approved):

PR #1327, opened 2021-07-15
implementing pagination for lucene

sophiabuell:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -99,7 +100,15 @@
         this.limitRemaining = scanProperties.getExecuteProperties().getReturnedRowLimitOrMax();
         this.timer = state.context.getTimer();
         this.query = query;
-        this.currentPosition = continuation == null ? 0 : Ints.fromByteArray(continuation);
+        if (continuation != null) {
+            Tuple continuationTuple = Tuple.fromBytes(continuation);
```
MMcM, line 104: It feels like this is complicated enough to be a Protobuf rather than just a Tuple.

Also this will allow us to evolve cursor handling going forward.

Compare all the `Continuation` messages in `record_cursor.proto`

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
sophiabuell: I can get that going but  we are limited in our continuation complexity by how lucene  handles scanning pagination.

MMcM (approved):

PR #1330, opened 2021-07-16
Resolves #1315: Allow indexes to be marked as replacing existing indexes

### LINKED ISSUE #1315: Allow indexes to be marked as replacing existing indexes
The idea here is to allow for one index to be declared as the replacement for another index. The idea is that once the second index has been built, then the presence of this index configuration should disable (and therefore delete) the first index. The idea here is that a user might want realize that their queries would be better satisfied with the newer index, but the older index might be better than nothing, so they might not want to drop the older index until the first index is built.

In theory, a user can do this now by:

1. Adding the new index to the meta-data
2. Building the new index
3. Dropping the old index

However, if the meta-data is shared by multiple record stores, the user can only do this safely once the index has been built on all stores, which may take a while to complete. This option would allow the user to begin taking advantage of space savings as soon as the second index is built in whichever stores happen to have it.

A few more notes:

* This logic to drop the index should be run during checkVersion time (in case the option gets added for stores that already have built both new indexes) and during index build time
* It would be nice if this option would allow for a many-to-many relationship, i.e., it should handle the case where an index is replaced by multiple newer indexes. One not-so-contrived instance where a user might want this would be something like replacing an index on a single field (let's say `(a)`) with two compound indexes prefixed by that field (let's say `(a, b)` and `(a, c)`) and only dropping the first index when all of the other indexes have been built.
* If this is an index option, then it should go on the old index. That way, if the old index is every dropped entirely from the meta-data, the new index doesn't need to be modified to remove the index option saying it replaces certain indexes

alecgrieser:
work in progress

todo: maybe more tests? also, make Index::replacedByIndexNames return a set

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
@@ -31,6 +31,7 @@ import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpressionVisitor;
 import com.apple.foundationdb.tuple.Tuple;
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Descriptors;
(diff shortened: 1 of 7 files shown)


alecgrieser:
Replaced by #1331 after discovering some concurrency problems

Issue #1296, opened 2021-06-03
OnlineIndexerIndexFromIndexTest.testIndexFromIndexPersistentContinuePreviousByRecords() sometimes fails with "runner has been closed"


alecgrieser:
When running tests in CI, I saw the `testIndexFromIndexPersistentContinuePreviousByRecords` test fail with "runner has been closed" errors, though it only happens intermittently.

The stack trace is:

```
java.util.concurrent.CompletionException: com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner$RunnerClosed: runner has been closed
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1063)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunnerImpl.close(FDBDatabaseRunnerImpl.java:352)
	at com.apple.foundationdb.record.provider.foundationdb.IndexingCommon.close(IndexingCommon.java:200)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexer.close(OnlineIndexer.java:398)
	at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerIndexFromIndexTest.buildSrcIndex(OnlineIndexerIndexFromIndexTest.java:69)
```

The stack trace is hopefully helpful at determining where the runner is being used after closure, though it will require some digging to figure out where it be closed.

alecgrieser:
This caused the tests for #1295 to fail, but only for one of the PRB jobs

nblintao:
`com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerIndexFromIndexTest.testIndexFromIndexPersistentContinuePreviousByIndex()` also have this problem. Looks same
https://fdb-record-layer-ci.foundationdb.org/job/prb-proto3/2550/testReport/junit/com.apple.foundationdb.record.provider.foundationdb/OnlineIndexerIndexFromIndexTest/testIndexFromIndexPersistentContinuePreviousByIndex__/

PR #1331, opened 2021-07-20
Resolves #1315: Allow indexes to be marked as replacing existing indexes

### LINKED ISSUE #1315: Allow indexes to be marked as replacing existing indexes

alecgrieser:
This adds a new index option prefix, `replacedBy`. This then allows the user to use the prefix to specify one or more indexes, and when all of those indexes have been built, the index will be disabled and its data deleted. This allows the user to effectively drop an indexe, though only after when the index has been fully replaced by new indexes. This both allows older unused data to be protected while also protecting the user against removing the index before all of the queries that use it have been redirected to other indexes.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataValidator.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataValidator.java
@@ -24,17 +24,12 @@ import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.RecordMetaData;
 import com.apple.foundationdb.record.RecordMetaDataProvider;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
-import com.google.common.collect.Sets;
 import com.google.protobuf.Descriptors;
 
 import javax.annotation.Nonnull;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
-import java.util.Set;
 
 /**
  * Validator for {@link RecordMetaData}.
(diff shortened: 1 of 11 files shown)


alecgrieser:
proto3 tests appear to have failed due to the same thing as #1296 but with `OnlineIndexerIndexFromIndexTest.testIndexFromIndexOtherSrcIndexBecomesUnusable()`

test this please

alecgrieser:
Another test failure of `OnlineIndexerIndexFromIndexTest.testIndexFromIndexRebuildIfWriteOnlyAndForceBuildAndBuildIfDisabled()` runner closed error. See #1296 

test this please

alecgrieser:
test this please

alecgrieser:
test this please

scgray (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1993,6 +1993,42 @@ private static void checkStoreHeaderInternal(@Nonnull RecordMetaDataProto.DataSt
         }
     }
 
+    /**
+     * Schedule a post-commit hook to look for replaced indexes. This method delays executing the replaced-indexes
```
scgray, line 1997: Isn't this a `pre-commit hook`?

alecgrieser: Ah, yes. A copy-paste bug, I think

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreReplaceIndexTest.java
```diff
@@ -0,0 +1,445 @@
...
+                                IndexOptions.REPLACED_BY_OPTION_PREFIX + "_00", "fakeIndex",
+                                IndexOptions.REPLACED_BY_OPTION_PREFIX + "_01", "MySimpleRecord$str_value_indexed"));
+                metaDataBuilder.addIndex("MySimpleRecord", index);
+            }));
+            assertThat(err.getMessage(), containsString("Index indexWithOneFakeReplacement has replacement indexes [fakeIndex] that are not in the meta-data"));
+        }
+    }
+
+    @Test
+    public void replacementIndexCycleFails() {
```
ScottDugas, line 419: Should there also be a test for `indexA` replacedBy `indexB, indexC` and `indexB` replaced by `indexD`, and `indexC` replaced by `indexD` (resulting in a non-directed cycle)

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataValidator.java
```diff
@@ -131,6 +136,39 @@ protected void validateIndex(@Nonnull Index index) {
...
+                if (!metaData.hasIndex(replacementIndexName)) {
+                    missingReplacements.add(replacementIndexName);
+                }
+            }
+            if (!missingReplacements.isEmpty()) {
+                throw new MetaDataException("Index " + index.getName() + " has replacement indexes "
+                                            + missingReplacements + " that are not in the meta-data");
+            }
+
+            // Check for cycles in the replacement graph
```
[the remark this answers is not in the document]
alecgrieser: Hm, I believe this (directed) cycle detection algorithm is complete for the situations it's designed to protect against, which is to make sure an index never relies on itself (directly or indirectly) from being built. But I think maybe a better check is to just make sure that none of the indexes that you rely on themselves have indexes that they are being replaced by. That way you never get into a situation where `indexA` is to be replaced by `indexB` which in turn is to be replaced by `indexC`, and then if `indexC` gets built, it's ambiguous whether `indexA` should get dropped or not.

Alternatively, the replacement logic could explore further down the graph (and disable an index if either all of its replacement indexes are built or all of its replacement indexes' replacement indexes' (etc.) are built), but it seems better to make the validation stricter and keep the logic simple(r).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1993,6 +1993,100 @@ private static void checkStoreHeaderInternal(@Nonnull RecordMetaDataProto.DataSt
...
+        beginRecordStoreStateRead();
+        final RecordMetaData metaData = getRecordMetaData();
+        final List<Index> indexesToRemove = new ArrayList<>();
+        try {
+            for (Index index : metaData.getAllIndexes()) {
+                final List<String> replacedByNames = index.getReplacedByIndexNames();
+                if (!replacedByNames.isEmpty()) {
+                    // Check if all of the replaced by index names are readable
+                    if (replacedByNames.stream()
+                            .allMatch(replacedByName -> metaData.hasIndex(replacedByName) && isIndexReadable(replacedByName))) {
```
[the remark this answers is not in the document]
alecgrieser: Added a test for that, though that configuration is now banned by the meta-data validator.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreReplaceIndexTest.java
alecgrieser: Added that test, which fails validation now

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2708,7 +2802,7 @@ private void logExceptionAsWarn(KeyValueLogMessage message, Throwable exception)
                 } else {
                     return false;
                 }
-            }).whenComplete((b, t) -> endRecordStoreStateWrite());
+            }).whenComplete((b, t) -> endRecordStoreStateWrite()).thenApply(this::addRemoveReplacedIndexesCommitCheckIfChanged);
```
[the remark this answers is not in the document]
alecgrieser: Yeah, that sounds reasonable enough. I'll work on that method to add the method, but I'll push the other changes I already for another round of review.

alecgrieser (commented):
Attempted to address the comments, at least in spirit. One thing left to do is to add the `getIndexesToBuild` method, which I began, but I also want to add some tests around, so might take a bit more time

ScottDugas (approved):

PR #1332, opened 2021-07-21
adding index scan type for multiFieldParser to enable generic queryin…

sophiabuell:
…g and remove the extra field per document

- remove default store and sort for each field

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -212,7 +212,7 @@ public class FDBStoreTimer extends StoreTimer implements EventKeeper {
         LUCENE_LIST_ALL("lucene list all"),
         /** Number of getFileReference calls in the FDBDirectory.*/
         LUCENE_GET_FILE_REFERENCE("lucene get file references"),
-        /** Number of documents returned from a single Lucene Index Scan */
+        /** Number of documents returned from a single Lucene Index Scan. */
         LUCENE_INDEX_SCAN("lucene search returned documents"),
         ;
 
@@ -649,7 +649,7 @@ public class FDBStoreTimer extends StoreTimer implements EventKeeper {
         LUCENE_BLOCK_READS("lucene block reads", false),
         /** Time to write a file references in Lucene's FDBDirectory.*/
         LUCENE_WRITE_FILE_REFERENCE("lucene write file reference" ,false),
-        /** Matched documents returned from lucene index reader scans **/
+        /** Matched documents returned from lucene index reader scans. **/
         LUCENE_SCAN_MATCHED_DOCUMENTS("lucene scan matched documents", false),
         ;
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -212,6 +212,8 @@ public class FDBStoreTimer extends StoreTimer implements EventKeeper {
         LUCENE_LIST_ALL("lucene list all"),
         /** Number of getFileReference calls in the FDBDirectory.*/
         LUCENE_GET_FILE_REFERENCE("lucene get file references"),
+        /** Number of documents returned from a single Lucene Index Scan */
+        LUCENE_INDEX_SCAN("lucene search returned documents"),
         ;
 
         private final String title;
@@ -647,6 +649,8 @@ public class FDBStoreTimer extends StoreTimer implements EventKeeper {
         LUCENE_BLOCK_READS("lucene block reads", false),
         /** Time to write a file references in Lucene's FDBDirectory.*/
         LUCENE_WRITE_FILE_REFERENCE("lucene write file reference" ,false),
+        /** Matched documents returned from lucene index reader scans **/
+        LUCENE_SCAN_MATCHED_DOCUMENTS("lucene scan matched documents", false),
         ;
 
         private final String title;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/LuceneQueryComponent.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/LuceneQueryComponent.java
@@ -29,7 +29,6 @@ import com.apple.foundationdb.record.query.plan.temp.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.temp.GraphExpansion;
 import com.apple.foundationdb.record.util.HashUtils;
 import com.google.common.base.Verify;
-import com.google.common.collect.Lists;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
 import jdk.jfr.Experimental;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/LuceneQueryComponent.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/LuceneQueryComponent.java
@@ -50,10 +50,9 @@ public class LuceneQueryComponent implements QueryComponent, ComponentWithCompar
     // QueryParserBase expects the query to contain the fields to be run against and takes a default field
     // which in our use case is the primary key field.
     // The MultiFieldQueryParser runs the query against all fields listed and uses AND to join them.
-    // If true and fields not empty we run against all fields listed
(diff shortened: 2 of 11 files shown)


sophiabuell:
Jenkins test this please

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -117,11 +117,22 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
-            Query query = parser.parse(range.getLow().getString(0));
-            return new LuceneRecordCursor(executor, scanProperties, state, query, continuation, state.index.getRootExpression().normalizeKeyForPositions());
+            // This cannot work with nested documents the way that we currently use them. BlockJoin will be essential for this
+            // functionality in this way.
+            if (scanType == IndexScanType.BY_LUCENE_FULL_TEXT) {
+                List<String> fieldNames = listIndexFieldNames(state.index.getRootExpression());
+                MultiFieldQueryParser parser = new MultiFieldQueryParser(fieldNames.toArray(new String[fieldNames.size()]), analyzer);
+                parser.setDefaultOperator(QueryParser.Operator.OR);
+                Query query  = parser.parse(range.getLow().getString(0));
+                return new LuceneRecordCursor(executor, scanProperties, state, query, continuation, state.index.getRootExpression().normalizeKeyForPositions());
```
MMcM, line 129: This seems complicated enough to arrange to put outside the `if`. `MultiFieldQueryParser` extends `QueryParser`, right? So the conditional could just make the right kind of parser.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -99,6 +99,9 @@
         this.executor = executor;
         this.limitManager = new CursorLimitManager(state.context, scanProperties);
         this.limitRemaining = scanProperties.getExecuteProperties().getReturnedRowLimitOrMax();
+        if (this.limitRemaining == Integer.MAX_VALUE) {
+            this.limitRemaining = 200;
```
MMcM, line 103: That does not seem right. We don't want the whole cursor to stop before the supplied limit. We just want to limit the individual searches and repeat within a single cursor lifetime.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/LuceneQueryComponent.java
```diff
@@ -46,23 +45,31 @@
...
-    public static final String FULL_TEXT_KEY_FIELD = "__full_text_key_field__";
 
+    //MultiFieldSearch determines whether MultiFieldQueryParser or QueryParserBase is used.
+    // QueryParserBase expects the query to contain the fields to be run against and takes a default field
+    // which in our use case is the primary key field.
+    // The MultiFieldQueryParser runs the query against all fields listed and uses AND to join them.
+    // If true and fields not empty we run against all fields listed
+    // If false and fields not empty we run with the QueryParserBase not specifying fields
+    // If true and fields empty we run against all fields in the root expression
+    // If false and fields empty we run against the primary key field.
```
MMcM, line 56: Are all these cases actually expected? Or is this just to fill out the contract? That is, why would a caller instantiate multi-fields and not give the fields?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -62,6 +62,7 @@
 
 import static com.apple.foundationdb.record.lucene.DirectoryCommitCheckAsync.getOrCreateDirectoryCommitCheckAsync;
 import static com.apple.foundationdb.record.lucene.IndexWriterCommitCheckAsync.getIndexWriterCommitCheckAsync;
+import static com.google.common.primitives.Ints.min;
```
MMcM, line 65: The only advantage this has over `Math.min` is that it takes more than two arguments, which we don't need here.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -232,14 +241,18 @@ private synchronized IndexReader getIndexReader() throws IOException {
     private void performScan() throws IOException {
         indexReader = getIndexReader();
         searcher = new IndexSearcher(indexReader);
+        int limit = min(limitRemaining, 200);
```
MMcM, line 244: Make `200` a static field. It's fine for now that it's constant, but at least then there's a place for some Javadoc.

MMcM (changes requested):
Need a test where the Lucene cursor needs multiple searches to fulfill the query.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -242,13 +242,30 @@ public void testLimitWithContinuation() {
...
+                metaDataBuilder.addIndex(SIMPLE_DOC, SIMPLE_TEXT_SUFFIXES);
+            });
+            for (int i = 0; i < 800; i++) {
+                recordStore.saveRecord(createSimpleDocument(1623L + i, DYLAN, 2));
+            }
+            assertEquals(251, recordStore.scanIndex(SIMPLE_TEXT_SUFFIXES, IndexScanType.BY_LUCENE_FULL_TEXT,
+                    TupleRange.allOf(Tuple.from("idiot")), null,
+                    ExecuteProperties.newBuilder().setReturnedRowLimit(251).build().asScanProperties(false))
+                    .getCount().join());
+        }
```
MMcM, line 266: There should be some instrumentation in `performSearch` and this should assert on the number of times that was done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -232,14 +243,24 @@ private synchronized IndexReader getIndexReader() throws IOException {
     private void performScan() throws IOException {
```
MMcM, line 243: This needs some instrumentation:
* `Event` for calling `searcher`.
* `Count` for number of docs returned by that.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -124,12 +127,19 @@
...
             try {
                 performScan();
-            } catch (IOException ioe) {
-                throw new RuntimeException(ioe);
+            } catch (IOException ioException) {
+                throw new RuntimeException(ioException);
             }
+            if (timer != null) {
+                timer.recordSinceNanoTime(FDBStoreTimer.Events.LUCENE_INDEX_SCAN, startTime);
+                timer.increment(FDBStoreTimer.Counts.LUCENE_SCAN_COUNT);
```
MMcM, line 139: There is no need for this third metric; `LUCENE_INDEX_SCAN` will have both a count and a total time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -647,6 +649,10 @@ public String logKey() {
         LUCENE_BLOCK_READS("lucene block reads", false),
         /** Time to write a file references in Lucene's FDBDirectory.*/
         LUCENE_WRITE_FILE_REFERENCE("lucene write file reference" ,false),
+        /** Documents returned from lucene index scans **/
+        LUCENE_RETURNED_DOCUMENTS("lucene returned documents", false),
```
MMcM, line 653: I think I might have trouble telling this apart from `LOAD_SCAN_ENTRY`. Maybe `SCANNED` would be clearer than `RETURNED`?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
sophiabuell: I labeled it LUCENE_SCAN_MATCHED_DOCUMENTS since scanned seemed unclear, we are scanning more than just the documents that we are returning from the method.

MMcM (approved):

sophiabuell:
Jenkins test this please

Issue #1322, opened 2021-07-15
The Lucene index blocks in an asynchronous context frequently


scgray:
While reviewing https://github.com/FoundationDB/fdb-record-layer/pull/1309, I noticed that the Lucene index uses `join()` relatively frequently (particularly in `FDBDrectory`).  A first note is that, with the exception of tests, Record Layer code should never use `join()` or `get()` to block on a future, but should instead use `FDBRecordContext.asyncToSync` or `FDBDatabase.asyncToSync()`.  This methods provide some critical functionality to Record Layer:

* They establish a configurable time limit on how long we are willing to wait for the operation
* They automatically define a wait timer that will be included in the `FDBStoreTimer` to track the time spent waiting
* They include automatic error handling and wrapping to promote thrown exceptions to the appropriate `RecordCoreException`, as necessary
* They have the ability to detect if `asyncToSync()` is being called from within a completing future

This last bullet is the key one.

Blocking code **must never** be called within a `CompletableFuture`, which means that `join()`, `get()`, and `asyncToSync()` cannot ever be called within one.  The problem is that there are a limited number of threads that are used to invoke the lambda in the future when it completes, so if you have a future that spawns a bunch of async work, then, within the same future, waits on it to complete, then you have now tied up that thread while it waits.  For the work that it is waiting on to complete, a thread must be available to complete it, so if you get enough threads waiting like this you can starve the thread pool and basically get a deadlock where the whole world freezes. 

In https://github.com/FoundationDB/fdb-record-layer/pull/1309 I asked for the `join()` to be changed to `asyncToSync()`. I then, locally turned on block-in-async detection by changing `fdb-record-layer-lucene.gradle` to:

```
test {
    useJUnitPlatform {
        ...
        // Configure whether or not tests will validate that asyncToSync isn't being called in async
        // context.  See BlockingInAsyncDetection class for details on values.
        systemProperties = [ "com.apple.foundationdb.record.blockingInAsyncDetection":
            System.getenv('BLOCKING_DETECTION') ?: "IGNORE_COMPLETE_EXCEPTION_BLOCKING" ]
    }
}
```

Now, many of the lucene tests file, for example:

```
    com.apple.foundationdb.record.provider.foundationdb.BlockingInAsyncException: Blocking in an asynchronous context
        at com.apple.foundationdb.record.provider.foundationdb.FDBDatabase.logOrThrowBlockingInAsync(FDBDatabase.java:1263)
[...]

PR #1333, opened 2021-07-22
Deal with asynchrony issues in FDBDirectory

scottfines:
see #1322 for more information about the bug. This fix is somewhat temporary, and accomplishes 3 things:

1. Resolves some spotty blocking calls in `FDBDirectory` so that they block less
2. Ensures that the forced synchronous methods in `FDBDirectory` are not called on the context's executor, to avoid deadlocks (still potential performance problems due to waiting for an open FJ thread, but at least it's less likely to actually break)
3. Add the ability to ignore blocking in asynchronous methods when we know that they are safe (for now)

A very good question would be: why do we have to be asynchronous here? and the answer is--because Lucene requires us to, and I'm not in a position to rewrite Lucene's entire API in order to support asynchronous method execution (I'm not even really sure that Lucene's thread model would support it). So here we are.

In the long term, we'll need to move blocking operations to a separate blocking thread pool in order to avoid potential performance bottlenecks, but this should serve in the short term.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
@@ -50,6 +50,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
@@ -1241,18 +1242,53 @@ public class FDBDatabase {
         // we will simply log the fact that this has taken place and where it has taken place here.
         StackTraceElement possiblyAsyncReturningLocation = null;
 
-        for (StackTraceElement stackElement : stack) {
+        for (int pos = 0; pos < stack.length; pos++) {
+            StackTraceElement stackElement = stack[pos];
             if (stackElement.getClassName().startsWith(CompletableFuture.class.getName())) {
-                logOrThrowBlockingInAsync(behavior, isComplete, stackElement, BLOCKING_IN_ASYNC_CONTEXT_MESSAGE);
+                //look to see if we can ignore this because of careful annotations
+                boolean canIgnore = false;
+                for (int annotationPos = pos - 1; annotationPos >= 0; annotationPos--) {
+                    StackTraceElement possibleIgnore = stack[annotationPos];
+                    if (canIgnoreBlockingInAsync(possibleIgnore.getClassName(), possibleIgnore.getMethodName())) {
+                        canIgnore = true;
+                        break;
+                    }
+                }
+                if (!canIgnore) {
+                    logOrThrowBlockingInAsync(behavior, isComplete, stackElement, BLOCKING_IN_ASYNC_CONTEXT_MESSAGE);
+                }
             } else if (stackElement.getMethodName().endsWith("Async")) {
                 possiblyAsyncReturningLocation = stackElement;
             }
         }
-
         if (possiblyAsyncReturningLocation != null && !isComplete) {
             // Maybe one day this will be configurable, but for now we will only allow this situation to log
             logOrThrowBlockingInAsync(BlockingInAsyncDetection.IGNORE_COMPLETE_WARN_BLOCKING, isComplete,
                     possiblyAsyncReturningLocation, BLOCKING_RETURNING_ASYNC_MESSAGE);
+        }
+    }
+
+    private boolean canIgnoreBlockingInAsync(final String className, final String methodName) {
+        try {
+            Class<?> clazz = Class.forName(className);
+            final Method[] methods = clazz.getMethods();
+            boolean found = false;
+            for (Method method : methods) {
+                /*
+                 * Since the same method name can be used with different argument sets, we just
+                 * check all of them--if one of them lacks the annotation, then it's not safe.
+                 */
+                if (method.getName().equals(methodName)) {
+                    found = true;
+                    if (method.getAnnotation(IgnoreBlockingInAsync.class) == null) {
+                        return false;
+                    }
+                }
+            }
+            return found;
+        } catch (ClassNotFoundException e) {
+            //this shouldn't happen, but when in doubt, let the caller think this isn't annotated for safety
+            return false;
         }
     }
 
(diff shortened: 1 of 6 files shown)


scottfines:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -1241,21 +1242,56 @@ private void checkIfBlockingInFuture(CompletableFuture<?> future) {
...
         }
-
         if (possiblyAsyncReturningLocation != null && !isComplete) {
             // Maybe one day this will be configurable, but for now we will only allow this situation to log
             logOrThrowBlockingInAsync(BlockingInAsyncDetection.IGNORE_COMPLETE_WARN_BLOCKING, isComplete,
                     possiblyAsyncReturningLocation, BLOCKING_RETURNING_ASYNC_MESSAGE);
         }
     }
 
+    private boolean canIgnoreBlockingInAsync(final String className, final String methodName) {
```
scgray, line 1271: This should probably have a test added in `FDBDatabaseTest` (there is a suite of tests around this named `testBlockingInAsync*`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -255,70 +256,90 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
...
-    }
-
-    private void record(StoreTimer.Event event, long durationNanos) {
-        final FDBStoreTimer timer = context.getTimer();
-        if (timer != null) {
-            timer.record(event, durationNanos);
+            List<String> outList = instrument.get();
+            //noinspection ToArrayCallWithZeroLengthArrayArgument
+            return outList.toArray(new String[outList.size()]);
+        } catch (ExecutionException ee) {
```
scgray, line 266: I think this can all be replaced with:

```
catch (Exception e) {
   throw FDBExceptions.wrap(e);
}
```

scgray (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -255,70 +256,90 @@ public void writeData(long id, int block, @Nonnull byte[] value) {
...
-            outList.add(name);
-            FDBLuceneFileReference fileReference = new FDBLuceneFileReference(Tuple.fromBytes(kv.getValue()));
-            // Only composite files and segments are prefetched.
-            if (name.endsWith(".cfs") || name.endsWith(".si") || name.endsWith(".cfe")) {
-                try {
-                    readBlock(name, CompletableFuture.completedFuture(fileReference), 0);
-                } catch (RecordCoreException e) {
-                    LOGGER.warn(KeyValueLogMessage.of("Exception thrown during prefetch", "resource", name, "exception"), e);
+    private CompletableFuture<List<String>> internalListAll() {
+        return context.ensureActive().getRange(metaSubspace.range()).asList().thenApply(keyValues -> {
```
ScottDugas, line 279: What thread pool does this happen on?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -175,7 +175,7 @@ private void incrementCallCount(FDBStoreTimer.Counts counter) {
         FDBLuceneFileReference fileReference = this.fileReferenceCache.getIfPresent(name);
         if (fileReference == null) {
             return context.instrument(FDBStoreTimer.Events.LUCENE_GET_FILE_REFERENCE, context.ensureActive().get(metaSubspace.pack(name))
```
ScottDugas, line 177: What thread pool does this run on? The annotation gets added to the method where `asyncToSync` is called, correct? In that case it seems like this method should document what thread pool its future is running on.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
```diff
@@ -136,7 +143,7 @@ public void testDeleteData() throws Exception {
         directory.deleteFile("test1");
         assertEquals(directory.listAll().length, 0);
 
-        assertCorrectMetricCount(FDBStoreTimer.Waits.WAIT_LUCENE_DELETE_FILE,2);
```
ScottDugas, line 139: Why did this change from 2 to 1?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -404,19 +425,31 @@ public void syncMetaData() throws IOException {
      * @param dest desc
      */
     @Override
+    @IgnoreBlockingInAsync
     public void rename(@Nonnull final String source, @Nonnull final String dest) {
+        /*
+         * Note on concurrency safety:
+         *
+         * DO NOT call this on the context's thread pool--eventually, we'll need a separate thread pool for blocking
+         * operations like this, but in the short term, use the default FJ pool (see deleteFile for a more detailed
```
ScottDugas, line 434:
```suggestion
         * operations like this, but in the short term, use the default fork-join pool (see deleteFile for a more detailed
```

ScottDugas (changes requested):

ngbinh:
Hi, we encountered cases of deadlocks when using record layer lucene indexes. Is there a plan to finish this PR? Thanks

MMcM:
> Hi, we encountered cases of deadlocks when using record layer lucene indexes. Is there a plan to finish this PR? Thanks

Do you attribute these deadlocks to something like fork-join pool exhaustion? I'd like to confirm that the bugs addressed here will fix your cases.

ngbinh:
@MMcM Yes. Under heavy load on records with Lucene indexes enabled, all the workers on the FJ pool are exhausted.

arnaud-lacurie:
Closing old PR. This can be reopened later if need be.

PR #1335, opened 2021-07-22
Resolve 1334: Provide a method to clean readable indexes build data

jjezra:
1. Clear build data when index is marked readable
   2. For existing record-stores, provide an API to clear this data for all readable indexes.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -2634,6 +2634,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                             throw wrapped;
                         } else {
                             updateIndexState(index.getName(), indexKey, IndexState.READABLE);
+                            clearReadableIndexBuildData(tr, index);
                             return true;
                         }
                     });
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3702,6 +3703,20 @@ public void removeFormerIndex(FormerIndex formerIndex) {
         }
     }
 
+    private void clearReadableIndexBuildData(Transaction tr, Index index) {
+        tr.clear(indexRangeSubspace(index).range());
```
alecgrieser, line 3707: In theory, it would be good to have a way of testing this. It seems pretty clear to me it's doing the right thing, but it still would be good to have a test that, say, builds an index and makes sure the range space is empty or something, and that the index can still be used.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2634,6 +2634,7 @@ public Range getUnbuiltRange() {
                             throw wrapped;
                         } else {
                             updateIndexState(index.getName(), indexKey, IndexState.READABLE);
+                            clearReadableIndexBuildData(tr, index);
```
alecgrieser, line 2637: Does this need to be added to `uncheckedMarkIndexReadable`, too, or was this not done on purpose (so that force marking an index as `READABLE` doesn't clear out this information)?

alecgrieser (changes requested):
Mostly looks good. I'm leaving as "request changes" because it would be nice if there were additional tests, but it might be fine to merge without given the nature of the change

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: It seems that the cases of `uncheckedMarkIndexReadable` calls that will leave behind temporary index building information are extreme and rare enough to have this data staying behind. If anything, it'll make a useful debug information (for example, seeing which part of the record store was actually indexed before enforcing a readable state).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: Added `testMarkReadableClearsBuiltRanges`.

jjezra:
test this please

alecgrieser (approved):

PR #1337, opened 2021-07-24
Resolve 1336: Use IndexingPolicy instead of indexStatePrecondition

jjezra:


jjezra:
Comment to reviewer: 
Before marking `setIndexStatePrecondition` as deprecated and adjusting the tests, please provide a feedback for this code approach.

jjezra:
Test this please

build/package-lists/java/package-list
```diff
@@ -170,8 +170,8 @@ javax.xml.transform.stax
 javax.xml.transform.stream
 javax.xml.validation
 javax.xml.ws
-javax.xml.ws.handler
-javax.xml.ws.handler.soap
+javax.xml.ws.desiredAction
+javax.xml.ws.desiredAction.soap
```
alecgrieser, line 174: What's this change about?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1817,28 +1829,51 @@ private static void checkPositive(int value, String desc) {
...
      */
     public static class IndexingPolicy {
-        public static final IndexingPolicy DEFAULT = new IndexingPolicy();
         @Nullable private final String sourceIndex;
         @Nullable private final Object sourceIndexSubspaceKey; // overrides the sourceIndex
         private final boolean forbidRecordScan;
+        private DesiredAction ifDisabled;
+        private DesiredAction ifWriteOnly;
+        private DesiredAction ifMismatchPrev;
+        private DesiredAction ifReadable;
```
alecgrieser, line 1838: It would probably be better if these could be made `final` and thus the configuration object could be immutable. I pointed out all of the places that I think would need to change to make that happen.

Also, how good is our test coverage for the various options here? Ideally, for example, it would be good to have a set of unit tests that were parameterized on previous index state and on `DesiredAction` and confirmed that each combination did the right thing. We may already have all of those tests, though

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1880,6 +1915,60 @@ public static Builder newBuilder() {
...
+                default: throw new RecordCoreException("bad index state: ", state);
+            }
+        }
+
+        /**
+         * Enforce an index rebuild when the index state is WRITE_ONLY. This is useful when other conditions
+         * require a full rebuild.
+         */
+        public void enforceRebuildIfWriteOnly() {
+            ifWriteOnly = DesiredAction.REBUILD;
```
alecgrieser, line 1969: Hm, I would rather that the policy remain immutable, if possible. In particular, I'm concerned about users who may cache a copy of their favorite `IndexingPolicy` and not expect for the policy object itself to sometimes mutate.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1817,28 +1829,51 @@ private static void checkPositive(int value, String desc) {
...
          * Build a non-active object.
          */
         public IndexingPolicy() {
-            this(null, null, false);
+            this(null, null, false,
+                    DesiredAction.REBUILD, DesiredAction.CONTINUE, DesiredAction.CONTINUE, DesiredAction.CONTINUE);
+        }
+
+        public static IndexingPolicy defaultPolicy() {
+            return new IndexingPolicy();
```
alecgrieser, line 1876: Should the default policy be a single object that gets cached (so that `defaultPolicy()` doesn't have to create a new object)? Does the object need to be immutable first before that can be accomplished?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -243,6 +241,10 @@
                     indexingPolicy = IndexingPolicy.newBuilder()
                             .setSourceIndexSubspaceKey(sourceIndexSubspaceKey)
                             .setForbidRecordScan(origPolicy.forbidRecordScan)
+                            .setIfDisabled(origPolicy.ifDisabled)
+                            .setIfWriteOnly(origPolicy.ifWriteOnly)
+                            .setIfMismatchPrev(desiredAction)
+                            .setIfReadable(origPolicy.ifReadable)
```
alecgrieser, line 247: Should `IndexingPolicy` have a `.toBuilder()` method that creates a new `IndexingPolicy.Builder` with all of the previous elements set? I'm a little concerned we might otherwise forget to update the `IndexingPolicy` copying logic here (or somewhere else) and accidentally drop a field we didn't intend to.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1708,6 +1713,7 @@ public Builder setSubspace(@Nonnull Subspace subspace) {
          */
         public Builder setIndexStatePrecondition(@Nonnull IndexStatePrecondition indexStatePrecondition) {
             this.indexStatePrecondition = indexStatePrecondition;
+            this.explicitIndexStatePrecondition = true;
```
alecgrieser, line 1716: Rather than keeping a boolean flag around to indicate whether the index state precondition has been set, I think you could have this update the `IndexingPolicy` directly, either by changing the type of `IndexingPolicy` to `IndexingPolicy.Builder` and mutating the index state `DesiredAction`s in place, or by copying the current indexing policy into a new `Builder`, updating the states on the builder, and then building and setting the `IndexingPolicy`. I think that would also allow you to avoid needing to mutate private data of the `IndexingPolicy` in the `build()` method, which we probably want to avoid.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1949,8 +2042,58 @@ public Builder forbidRecordScan() {
                 return this;
             }
 
+
+            /**
+             * set the initial handling policy, action if the index is in disabled.
```
alecgrieser, line 2047:
```suggestion
             * Set the initial handling policy action if the index is in disabled.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1949,8 +2042,58 @@ public Builder forbidRecordScan() {
...
+             *                   ERROR: reject (with exception).
+             * @return this builder
+             */
+            public Builder setIfDisabled(final DesiredAction ifDisabled) {
+                this.ifDisabled = ifDisabled;
+                return this;
+            }
+
+            /**
+             * set the initial handling policy, action if the index is in write only mode (i.e. partly built).
```
alecgrieser, line 2059:
```suggestion
             * Set the initial handling policy action if the index is in write only mode (i.e. partly built).
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1949,8 +2042,58 @@ public Builder forbidRecordScan() {
...
+             *                    ERROR: reject (with exception).
+             * @return this builder
+             */
+            public Builder setIfWriteOnly(final DesiredAction ifWriteOnly) {
+                this.ifWriteOnly = ifWriteOnly;
+                return this;
+            }
+
+            /**
+             * set the initial handling policy, action if the index is in write-only mode but the previous indexing method mismatches the requested one.
```
alecgrieser, line 2071:
```suggestion
             * Set the initial handling policy action if the index is in write-only mode but the previous indexing method mismatches the requested one.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1949,8 +2042,58 @@ public Builder forbidRecordScan() {
...
+             *                       ERROR: reject (with exception).
+             * @return this builder
+             */
+            public Builder setIfMismatchPrev(final DesiredAction ifMismatchPrev) {
+                this.ifMismatchPrev = ifMismatchPrev;
+                return this;
+            }
+
+            /**
+             * set the initial handling policy, action if the index is in write-only mode but the previous indexing method mismatches the requested one.
```
alecgrieser, line 2083:
```suggestion
             * Set the initial handling policy action if the index is in write-only mode but the previous indexing method mismatches the requested one.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -230,57 +230,38 @@ protected static Tuple convertOrNull(@Nullable Key.Evaluated key) {
...
             }
-
+            boolean shouldClear = desiredAction == OnlineIndexer.IndexingPolicy.DesiredAction.REBUILD;
+            boolean shouldBuild = shouldClear || indexState != IndexState.READABLE;
             message.addKeyAndValue(LogMessageKeys.INITIAL_INDEX_STATE, indexState);
+            message.addKeyAndValue(LogMessageKeys.INDEXING_POLICY_DESIRED_ACTION, desiredAction);
             message.addKeyAndValue(LogMessageKeys.SHOULD_BUILD_INDEX, shouldBuild);
             message.addKeyAndValue(LogMessageKeys.SHOULD_CLEAR_EXISTING_DATA, shouldClear);
+
+
```
alecgrieser, line 256:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -230,57 +230,38 @@ protected static Tuple convertOrNull(@Nullable Key.Evaluated key) {
...
-
-                default:
-                    throw new RecordCoreException("unknown index state precondition " + indexStatePrecondition);
+            OnlineIndexer.IndexingPolicy.DesiredAction desiredAction = policy.getStateDesiredAction(indexState);
+            if (desiredAction == OnlineIndexer.IndexingPolicy.DesiredAction.ERROR) {
+                throw new ValidationException("Index state is not as expected",
+                        LogMessageKeys.INDEX_NAME, index.getName(),
+                        LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
+                        LogMessageKeys.INDEX_STATE, indexState,
+                        LogMessageKeys.INDEXING_POLICY_DESIRED_ACTION, desiredAction
```
alecgrieser, line 246: Won't `desiredAction` always be `ERROR` here?

alecgrieser (changes requested):

build/package-lists/java/package-list
jjezra: This commit fixes a Javadoc issue, but I've taken the opportunity to rename this enum to (what I think is) a more descriptive name.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Making these fields final might be an issue when allowing `IndexStatePrecondition`, which is not part of this object, to override them. 
(another issue could be when setting, on some flows, `ifWriteOnly` to REBUILD - but that's easier to solve)..

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: That's a very good point! 
I'll replace all modifications of this object with a copy/modify

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: If the object is final, then probably yes. 
I'll change it back.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: That'll be a very elegant way to do it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Wouldn't changing `setIndexingPolicy(@Nullable final IndexingPolicy indexingPolicy)` to `setIndexingPolicy(@Nullable final IndexingPolicy.Builder builder)` break backward compatibility? 
If so, are you ok with it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed (and more below)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: Removed this field.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Made final.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Temporary, I've kept them both. 
(This may be changed in the final version of this PR)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1748,32 +1745,63 @@ public Builder setLeaseLengthMillis(long leaseLengthMillis) {
...
             return this;
         }
 
+        /**
+         * Add an {@link IndexingPolicy.Builder}. If set, this policy will determine how the index should be
+         * built.
+         * @param builder an IndexingPolicy builder.
+         * @return this Builder
+         */
+        public Builder setInexingPolicyBuilder(@Nonnull final IndexingPolicy.Builder builder) {
```
alecgrieser, line 1770:
```suggestion
        public Builder setIndexingPolicyBuilder(@Nonnull final IndexingPolicy.Builder builder) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1748,32 +1745,63 @@ public Builder setLeaseLengthMillis(long leaseLengthMillis) {
...
         public OnlineIndexer build() {
             validate();
-            if (explicitIndexStatePrecondition) {
-                indexingPolicy.ifDisabled = indexStatePrecondition.ifDisabled;
-                indexingPolicy.ifWriteOnly = indexStatePrecondition.ifWriteOnly;
-                indexingPolicy.ifMismatchPrev = indexStatePrecondition.ifMismatchPrev;
-                indexingPolicy.ifReadable = indexStatePrecondition.ifReadable;
+            // Here: Supporting dual backward compatibilities:
+            // 1. Allow the use of indexStatePrecondition
+            // 2. Allow the use of an explicit index policy (instead of the index policy builder)
```
alecgrieser, line 1784: Okay, and if I'm understanding the semantics here, they are:

1. If you set both an `IndexingPolicy` and an `IndexingPolicy.Builder` on the builder, it will end up using whichever you set last between the two of them
2. If you set an `IndexStatePrecondition` and an `IndexingPolicy` (or builder), then the values of the `DesiredAction`s will all be set from the `IndexStatePrecondition`, regardless of what's in the policy.

I guess those are reaonsable enough semantics, and probably backwards compatible (i.e., those who don't realize they need to set the `IndexingPolicy`'s `DesiredActions` will see it "do the right thing). But it's probably worth documenting in the corresponding `.setX` method Javadocs

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1915,6 +1933,22 @@ public static Builder newBuilder() {
             return new Builder();
         }
 
+        /**
+         * Create an indexing policy builder from existing object.
+         * @return an {@link IndexingPolicy} builder, defaulted to the existing object's values.
+         */
+        @Nonnull
+        public Builder toBuilder() {
+            return newBuilder()
```
alecgrieser, line 1942: One other approach that I've seen work for this is for the `Builder` to have a private constructor that takes an `IndexingPolicy` and then initializes all of the variables appropriately. If done right, it can be structured so that the compiler will give you an error if you add a field and forget to pass it along. But this is also probably fine if you don't want to make that change.

alecgrieser (changes requested):
Main remaining ask is for a few additional Javadoc comments about what happens in the config object builder if you set multiple things that all set similar information, so that that's a bit more accessible.

I think if you want to deprecate the `IndexStatePrecondition` in this PR, then that might also be a good idea, though not strictly required. Might make sense, even if we don't deprecate it, to note that the new configuration object is preferred to that enum.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: Got it. Yeah, I think it would be "probably fine" to keep it as just the indexing policy and not the builder, if you want to remove one rather than keep two. Or maybe both is fine. It's somewhat up to you.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: 1. Yes. I was trying to imitate the use of last value in other builder function (i.e `builder.setX('a').setX('b')` sets 'b' in X)
2. Yes. As a backward compatibility for today's code.
I'll try to improve the documentation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Oh.

jjezra:
Test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1949,8 +2077,58 @@ public Builder forbidRecordScan() {
...
+            }
+
+            /**
+             * Set the desired action if the index is in a write-only mode but the previous indexing method mismatches the requested one.
+             * @param ifMismatchPrev CONTINUE: (default) attempt to continue the previous indexing process, using the previous policy
+             *                       REBUILD: clear the existing data and rebuild, using the new policy
+             *                       ERROR: reject (with exception).
+             * @return this builder
+             */
+            public Builder setIfMismatchPrev(final DesiredAction ifMismatchPrev) {
```
alecgrieser, line 2112: Sorry, I should have mentioned this earlier, but maybe this should be `setIfMismatchedPrevious` or maybe even `setIfPolicyChanged`. It might also be fine as is; I don't have super strong feelings on this.

alecgrieser (changes requested):
Okay, looks good. Sorry, I have two comments I probably should have mentioned earlier, one that I've leaved in line and the other is that I think this needs a release note

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Sure, NP. Changed it to the your first suggestion.

alecgrieser:
Looks good, except I think this is still missing a release note

alecgrieser (approved):

alecgrieser:
Just needs a rebase, then can merge

jjezra:
Test this please

PR #1346, opened 2021-07-29
Resolve 1345: Index scrubber: log warnings when repair is not allowed

jjezra:
Made some changes in scrubbingPolicy

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
@@ -71,18 +71,22 @@ public class IndexingScrubMissing extends IndexingBase {
 
     @Nonnull private final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy;
     private long scanCounter = 0;
+    private int logWarningCounter;
 
     public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                                 @Nonnull final OnlineIndexer.IndexingPolicy policy,
                                 @Nonnull final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy) {
         super(common, policy, true);
         this.scrubbingPolicy = scrubbingPolicy;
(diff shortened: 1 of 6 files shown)


jjezra:
Please do not review yet..

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -215,7 +219,8 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                     }
                     // Here: Oh, No! the index is missing!!
                     // (Maybe) report an error and (maybe) return this record to be index
-                    if (LOGGER.isWarnEnabled() && scrubbingPolicy.shouldLogWarning()) {
+                    if (LOGGER.isWarnEnabled() && logWarningCounter > 0) {
+                        logWarningCounter --;
```
alecgrieser, line 223: Should it also log an additional warning when `logWarningCounter == 0` so that it's clear that we've truncated the warnings?

alecgrieser (approved):
LGTM. I had one question, but we can merge this without addressing that comment, I think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
jjezra: Mmm... wouldn't a caller who uses `logWarningsLimit=0` expect zero warnings?

alecgrieser: ¯\\\_(ツ)_/¯

It's kind of up to us, I guess, to set expectations. I was thinking specifically a warning that's like "other more specific warnings omitted". That way if someone had this misconfigured and expected to see warnings to inspect some kind of suspected corrupt index, then they'd at least get a message telling them something is up, even if it didn't tell them specifically what was wrong (until they increased the warnings limit).

PR #1359, opened 2021-08-12
Fixes #1358: Support cluster event parsing on FDB 6.3

### LINKED ISSUE #1358: Support cluster event parsing on FDB 6.3
The cluster event parsing tool contains the cluster protocol version within it, and it will only parse events if the protocol version matches. That value is still hard coded to the FDB 6.2 protocol version, so we should update it to be able to parse FDB 6.3 versions.

FDB also added a few extra changes in the transaction sampling data in 6.3 that we should be aware of and try and properly parse. See the changes here (based on the FDB protocol version): https://github.com/apple/foundationdb/blob/c942ee995ebe0718420e24e1dabe1ddb2e0af30c/contrib/transaction_profiling_analyzer/transaction_profiling_analyzer.py

alecgrieser:
This updates the list of supported protocol versions in the cluster log parsing logic, and it updates the logic for 6.3 based on the equivalent set of changes detailed in the Python version of the script: https://github.com/apple/foundationdb/blob/c942ee995ebe0718420e24e1dabe1ddb2e0af30c/contrib/transaction_profiling_analyzer/transaction_profiling_analyzer.py


alecgrieser:
@MMcM I'm especially interested if you have any idea on the right testing strategy for this. I'm also aware that this stuff is used for read-only tools for analyzing cluster analysis, so I guess the worst case if there's a bug is that we incorrectly serialize some events and those tools return bogus information, which isn't amazing, but may be an acceptable cost over more intense tests of the serialization/deserialization logic

MMcM (approved):

PR #1360, opened 2021-08-12
updating lucene updates to not include duplicate and skip indexEntry …

sophiabuell:
…logic

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -71,6 +71,7 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 
@@ -138,14 +139,31 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
     }
 
     protected static class DocumentEntry {
-        String fieldName;
-        Object value;
-        LuceneFieldKeyExpression expression;
+        final String fieldName;
+        final Object value;
+        final LuceneFieldKeyExpression expression;
 
         DocumentEntry(String name, Object value, LuceneFieldKeyExpression expression) {
             this.fieldName = name;
             this.value = value;
             this.expression = expression;
+        }
+
+        @Override
+        public boolean equals(final Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            final DocumentEntry entry = (DocumentEntry)o;
+            return fieldName.equals(entry.fieldName) && Objects.equals(value, entry.value) && expression.equals(entry.expression);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(fieldName, value, expression);
         }
 
     }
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -138,100 +137,134 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
+        if (newRecord != null) {
+            newRecordFields.addAll(getFields(root, newRecord, newRecord.getRecord()));
         }
+        if (oldRecord != null) {
+            oldRecordFields.addAll(getFields(root, oldRecord, oldRecord.getRecord()));
+        }
+
+        // evaluate which fields are changed
+        //Not sure if we want to allow deletion of fields. Or if the updated record will contain all fields not
+        // just the updated ones. Assuming here that the record will be complete if being updated as previously.
```
MMcM, line 237: Both records are complete (or `null`), yes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -138,100 +137,134 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
+        List<DocumentEntry> updatedFields = newRecordFields;
+        updatedFields.removeAll(oldRecordFields);
```
MMcM, line 239: This assumes that `DocumentEntry` has a meaningful (non-reference) `equals`, which it does not right now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -138,100 +137,134 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
-                    break;
-                case LONG:
-                    document.add(new LongPoint(fieldName, (long) value));
-                    break;
-                default:
-                    throw new RecordCoreArgumentException("Invalid type for lucene index field", "type", expression.getType());
+            // if old record has fields in the index and there are changes then delete old record
+            if (!oldRecordFields.isEmpty() && (!updatedFields.isEmpty() || newRecord == null)) {
+                Query query = SortedDocValuesField.newSlowExactQuery(PRIMARY_KEY_SEARCH_NAME, new BytesRef(oldRecord.getPrimaryKey().pack()));
+                writer.deleteDocuments(query);
```
MMcM, line 250: Is this necessary when we call `addDocument` below? That is, without it, would there be two documents with the same fields or would the one below overwrite?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -138,100 +137,134 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
             }
+
+            // if there are changes then update the index with the new document.
+            if (!updatedFields.isEmpty()) {
+                byte[] primaryKey = newRecord.getPrimaryKey().pack();
+                BytesRef ref = new BytesRef(primaryKey);
+                document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
+                document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
+                for (DocumentEntry entry : newRecordFields) {
```
MMcM, line 259: `newRecordFields` and `updatedFields` are the same, so this is only the changed fields. Don't we need them all?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: Without it there would be two documents with the same fields. Same as how there can be multiple fields with the exact same key/value.

sophiabuell: it does not delete the previous one. It passes null to the delNode value in the update method.

MMcM: ✅

MMcM (approved):

PR #1361, opened 2021-08-12
updating to accomidate null values for index

sophiabuell:


jleach4:
Test?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -176,6 +182,21 @@ public void testContinuation() {
         }
     }
 
+    @Test
+    public void testNullValue() {
```
jleach4, line 186: Since an int is not really nullable, it is hard for me to know that a null string will not break the code.

sophiabuell: This doesn't set the string value in one of the documents and therefore its set to null when the message is evaluated in the maintainer getFields method.

jleach4 (approved):
Approving based on it being a plus in the code base but I would still love a null check on a nullable value.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
MMcM: So this tests that the maintainer doesn't crash at save time. Would it work to have a query-based test that included an `isNull` predicate against an unset field and show that this is returned?

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -176,6 +184,23 @@ public void testContinuation() {
...
+    public void testNullValue() throws ExecutionException, InterruptedException {
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, metaDataBuilder -> {
+                metaDataBuilder.removeIndex(TextIndexTestUtils.SIMPLE_DEFAULT_NAME);
+                metaDataBuilder.addIndex(SIMPLE_DOC, SIMPLE_TEXT_SUFFIXES);
+            });
+            recordStore.saveRecord(createSimpleDocument(1623L, 2));
+            recordStore.saveRecord(createSimpleDocument(1632L, DYLAN, 2));
+            RecordCursor<IndexEntry> recordCursor = recordStore.scanIndex(SIMPLE_TEXT_SUFFIXES, IndexScanType.BY_LUCENE_FULL_TEXT,
+                    TupleRange.allOf(Tuple.from("*:* AND NOT text:[* TO *]")), null, ScanProperties.FORWARD_SCAN);
```
MMcM, line 197: Is this what https://github.com/FoundationDB/fdb-record-layer/blob/1470ef42493d8c20dc828139947e012198d7759b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java#L81-L83 should be doing?

MMcM (changes requested):

MMcM (approved):

sophiabuell:
Jenkins test this please

Issue #1363, opened 2021-08-13
OnlineIndexScrubberTest.testScrubberLimits - reduce number of records 


jjezra:
Need to reduce the number of records in the test.

PR #1364, opened 2021-08-13
Resove #1363: OnlineIndexScrubberTest.testScrubberLimits - reduce num…

jjezra:
…ber of records


jjezra:
As far as I can tell, no other test has more than 200 records.

MMcM (approved):

PR #1368, opened 2021-08-18
Resolve #1367: OnlineIndexScrubber: clear ranges when exhausted

### LINKED ISSUE #1367: OnlineIndexScrubber: clear ranges when exhausted
To save (a little - very little) space, clear the scrubbers ranges (`indexScrubIndexRangeSubspace` and `indexScrubRecordsRangeSubspace`) immediately after the scrubber finishes an iteration. 
This change will keep the subspace empty until index scrubbing is called again.

jjezra:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
```diff
@@ -158,6 +159,8 @@ public IndexingScrubDangling(@Nonnull final IndexingCommon common,
 
         return ranges.onHasNext().thenCompose(hasNext -> {
             if (Boolean.FALSE.equals(hasNext)) {
+                Transaction tr = store.getContext().ensureActive();
+                tr.clear(indexScrubIndexRangeSubspace(store, index).range());
                 return AsyncUtil.READY_FALSE; // no more missing ranges - all done
```
alecgrieser, line 164: Probably worth documenting here why the range is being cleared just to avoid confusing future readers

alecgrieser (changes requested):

alecgrieser (approved):

PR #1389, opened 2021-08-27
resolves #1301: integrate boolean normalization into planning process

### LINKED ISSUE #1301: integrate boolean normalization into planning process
The cascades planner currently just calls the boolean normalizer prior to planning in order to convert the filter predicate into its CNF. While that's great most of the time we can also miss out on some plans:

From `FDBOrToUnionTest.testOrQueryNoIndex`:

```
QueryComponent orComponent = Query.or(
                Query.field("num_value_3_indexed").equalsValue(1),
                Query.field("num_value_3_indexed").equalsValue(2),
                Query.field("num_value_3_indexed").equalsValue(4));
        RecordQuery query = RecordQuery.newBuilder()
                .setRecordType("MySimpleRecord")
                .setFilter(Query.and(Query.field("str_value_indexed").equalsValue("even"), orComponent))
                .build();
```

The desired plan is an index scan over `MySimpleRecord` ranged `["even", "even"]` followed by filtering using one predicate applying the three-legged _or_ term.

We have a predicate of the form `a ^ (x v y v z)` which is CNFed into: `(a ^ x) v (a ^ y) v (a ^ z)`. That makes it impossible to find the objectively best plan.

The solution for this problem is to properly integrate boolean normalization into the planning process having a rule to e.g. generate the CNF and a different one to generate the DNF.

normen662:
This PR contains:
- a fix for a bug where during matching a `QueryPredicate` would be reapplied for some cases although that predicate should only be reapplied in its residual form. For instance, a `ValueComparisonRangePredicate` needs to be translated into a conjunct of `ValuePredicate`s if it cannot be used as a search argument in an index scan. This translation is now done as part of the `Compensation` flow.
- a rewritten boolean normalizer called `BooleanPredicateNormalizer` which is based on `BooleanNormalizer` but works on `QueryPredicate`s instead of `QueryComponent`s. In addition to that, this new normalizer can both transform to CNF as well as DNF which comes in handy (CNF and DNF are both different instantiations of a more general algorithm that works identical for both cases except `OR` and `AND`s role are reversed).
- a new rule `NormalizePredicatesRule` which transforms a matched `QueryPredicate` into its DNF as well as into its CNF
- note that it is important that `DNF(pred) == DNF(CNF(pred))` and `CNF(pred) == CNF(DNF(pred))`. If these rules were to get violated it would cause an explosion in search space as the planner would continuously attempt to DNF and CNF derivation chains. These properties are tested and they hold.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanPredicateNormalizer.java
```diff
@@ -0,0 +1,408 @@
...
+
+    private int majorToNormalizedSize(@Nonnull final List<? extends QueryPredicate> children, final boolean negate) {
+        return children.stream().mapToInt(p -> toNormalizedSize(p, negate)).reduce(0, Math::addExact);
+    }
+
+    private int minorToNormalizedSize(@Nonnull final List<? extends QueryPredicate> children, final boolean negate) {
+        return children.stream().mapToInt(child -> toNormalizedSize(child, negate)).reduce(1, Math::multiplyExact);
+    }
+
+    /**
```
MMcM, line 298: The Javadoc on the next few methods and the comment in `trimTerms` assume DNF mode. I don't know how to rewrite them in a way that's still comprehensible. Also, in the trim case, the implication is in the other direction, from subset to superset, and that's what's needed.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanPredicateNormalizer.java
normen662: Will do the java doc changes.

In terms of which way the implication goes, wouldn't it be the same as both

`a v (a ^ b) = a`
and
`a ^ (a v b) = a`

or am I misunderstanding your comment?

normen662: Oh -- do you mean the way the comment in the method is written?

MMcM: Yes, exactly. `(a ∧ b ∧ c) ⇒ (a ∧ b)` and `(a ∨ b) ⇒ (a ∨ b ∨ c)`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanPredicateNormalizer.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    int getNormalizedSize(@Nullable final QueryPredicate predicate) {
+        if (predicate == null) {
+            return 0;
+        }
+        return toNormalizedSize(predicate, false);
+    }
+
+    private int toNormalizedSize(@Nonnull final QueryPredicate predicate, final boolean negate) {
```
ohadzeliger, line 276: In case the numbers can grow quite large, this calculation can take a while. Would it be more efficient to take in the limit and short-circuit the count when that limit is reached instead?

ohadzeliger (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanPredicateNormalizer.java
```diff
@@ -0,0 +1,408 @@
...
+        return normalize(predicate, true);
+    }
+
+    /**
+     * Convert the given predicate to its respective normal form, if necessary. If the size of the normal form would
+     * exceed the size limit, throw a {@link NormalFormTooLargeException}.
+     * @param predicate the query predicate to be normalized
+     * @param failIfTooLarge indicates whether we should throw an exception if the normalization is too big
+     *        (beyond {@link #sizeLimit}).
+     * @return the predicate in disjunctive normal form
```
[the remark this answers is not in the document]
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanPredicateNormalizer.java
```diff
@@ -0,0 +1,408 @@
...
+    /**
+     * Get the limit on the size of the DNF that this normalizer will produce.
+     * @return size limit
+     */
+    public int getSizeLimit() {
+        return sizeLimit;
+    }
+
+    /**
+     * Convert the given predicate to disjunctive normal form, if necessary. If the size of the DNF would exceed the
```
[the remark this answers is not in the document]
normen662: done.

normen662:
test this please

ohadzeliger (dismissed):

MMcM (approved):

PR #1393, opened 2021-08-31
Resolve #1392: Index Scrubber: missing index entry”: Add the (virtual…

### LINKED ISSUE #1392: Index Scrubber: missing index entry”: Add the (virtual) missing index’s key to the log message
This will allow an easier debugging (especially if the scrubber had fixed the problem)

jjezra:
…) missing index’s key to the log message

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
@@ -217,16 +217,10 @@ public class IndexingScrubMissing extends IndexingBase {
                     final byte[] keyBytes = maintainer.getIndexSubspace().pack(valueKey);
                     return maintainer.state.transaction.get(keyBytes).thenApply(indexVal -> indexVal == null ? valueKey : null);
                 })
+                .filter(Objects::nonNull)
                 .collect(Collectors.toList()))
-                .thenApply(list -> {
-                    Tuple missingKey = null;
-                    for (Tuple item: list) {
-                        if (item != null) {
-                            missingKey = item;
-                        }
-                    }
-                    if (missingKey  == null) {
-                        // no null index(s) = no record to index
+                .thenApply(missingIndexesKeys -> {
+                    if (missingIndexesKeys.isEmpty() || (missingIndexesKeys.size() == 1 && missingIndexesKeys.get(0) == null)) {
                         return null;
                     }
                     // Here: Oh, No! the index is missing!!
@@ -235,7 +229,7 @@ public class IndexingScrubMissing extends IndexingBase {
                         logWarningCounter --;
                         LOGGER.warn(KeyValueLogMessage.build("Scrubber: missing index entry",
                                         LogMessageKeys.KEY, rec.getPrimaryKey().toString(),
-                                        LogMessageKeys.INDEX_KEY, missingKey.toString())
+                                        LogMessageKeys.INDEX_KEY, missingIndexesKeys.toString())
                                 .addKeysAndValues(common.indexLogMessageKeyValues())
                                 .toString());
                     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
@@ -217,10 +217,10 @@ public class IndexingScrubMissing extends IndexingBase {
                     final byte[] keyBytes = maintainer.getIndexSubspace().pack(valueKey);
                     return maintainer.state.transaction.get(keyBytes).thenApply(indexVal -> indexVal == null ? valueKey : null);
                 })
-                .filter(Objects::nonNull)
                 .collect(Collectors.toList()))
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -227,7 +234,8 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                     if (LOGGER.isWarnEnabled() && logWarningCounter > 0) {
                         logWarningCounter --;
                         LOGGER.warn(KeyValueLogMessage.build("Scrubber: missing index entry",
-                                LogMessageKeys.KEY, rec.getPrimaryKey().toString())
+                                        LogMessageKeys.KEY, rec.getPrimaryKey().toString(),
```
scgray, line 237: There's a part of me that would actually like to log the record itself, however (a) that could result in an overly large log message and (b) the log could then leak sensitive PII ... so, we'll keep it at just the PK for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -227,7 +234,8 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
...
+                                        LogMessageKeys.INDEX_KEY, missingKey.toString())
```
scgray, line 238: This will only log one of the missing keys, but there could be multiple keys that are missing, so we wouldn't have the full scope of the "damage".  Perhaps you want to do something like:

```
AsyncUtil.getAll(indexEntryNoPKs.stream()
                .map(entry -> { ... })
                .filter(Objects::nonNull)
                .collect(Collectors.toList())
                .thenApply(...)
```

this will give you a list of all the missing keys.

scgray (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
jjezra: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -217,16 +217,10 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
...
-                    Tuple missingKey = null;
-                    for (Tuple item: list) {
-                        if (item != null) {
-                            missingKey = item;
-                        }
-                    }
-                    if (missingKey  == null) {
-                        // no null index(s) = no record to index
+                .thenApply(missingIndexesKeys -> {
+                    if (missingIndexesKeys.isEmpty() || (missingIndexesKeys.size() == 1 && missingIndexesKeys.get(0) == null)) {
```
scgray, line 223: Since the `filter()` wasn't really filtering, this will be a list of `null`'s for entries that are present and non-`null`'s for entries that were missing, so your test needs to be whether or not there is a non-null entry in the list.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -217,16 +217,10 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                     final byte[] keyBytes = maintainer.getIndexSubspace().pack(valueKey);
                     return maintainer.state.transaction.get(keyBytes).thenApply(indexVal -> indexVal == null ? valueKey : null);
                 })
+                .filter(Objects::nonNull)
```
scgray, line 220: This will never filter anything (sorry, this was my fault), as the `.filter()` is on the stream that is producing futures that are looking for the missing keys, so it is filtering on the futures, not the result of the futures.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -235,7 +229,7 @@ public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                         logWarningCounter --;
                         LOGGER.warn(KeyValueLogMessage.build("Scrubber: missing index entry",
                                         LogMessageKeys.KEY, rec.getPrimaryKey().toString(),
-                                        LogMessageKeys.INDEX_KEY, missingKey.toString())
+                                        LogMessageKeys.INDEX_KEY, missingIndexesKeys.toString())
```
scgray, line 232: Since the `filter()` wasn't really filtering, you'll want to strip out the `null` values in the list before logging.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
jjezra: NP. Fixing...

scgray (approved):

PR #1397, opened 2021-09-03
adding grouping functionality to the lucene index

sophiabuell:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -240,6 +244,22 @@ private void insertField(DocumentEntry entry, final Document document) {
         }
     }
 
+    public <M extends Message> DocumentEntry extractGrouping(@Nullable FDBIndexableRecord<M> record){
```
MMcM, line 247: The restriction that the grouping key be a single string value seems unnecessarily strict.

The only code that does not look like it would just work by using whole tuples is `getWriterName` and that could even use hex-encoded bytes, since it is internal only and quite local.

Relatedly, it does not seem that `DocumentEntry` is particularly appropriate for the extracted grouping. I think that things could be simplified by only normalizing the grouped key and using the regular evaluator on the grouping key to determine its value, again as a tuple.

The restriction that there is only one grouping for the entire record may be okay for now, although it will disallow the case of a map-like nested repeated message with the group split in the middle.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -131,7 +133,14 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
                 parser = new QueryParser(PRIMARY_KEY_SEARCH_NAME, analyzer);
             }
             Query query = parser.parse(range.getLow().getString(0));
-            return new LuceneRecordCursor(executor, scanProperties, state, query, continuation, state.index.getRootExpression().normalizeKeyForPositions());
+            String groupingKey = null;
+            for (int i = 1; i < range.getLow().getItems().size(); i++) {
+                Object comparison = range.getLow().get(i);
```
MMcM, line 138: I believe that some of the possible value types do not have equality-preserving `toString`. The one that occurs to me right away is byte array, which will be `[B...`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -240,6 +249,26 @@ private void insertField(DocumentEntry entry, final Document document) {
...
+        KeyExpression expression = state.index.getRootExpression();
+        if (expression instanceof GroupingKeyExpression) {
+            final List<DocumentEntry> groupingExpressions = getFields(((GroupingKeyExpression)expression).getGroupingSubKey(), record, record.getRecord());
+            if (groupingExpressions.isEmpty()) {
+                return null;
+            }
+            String indexPreface = "";
+            List<Key.Evaluated> evaluatedList = ((GroupingKeyExpression)expression).getGroupingSubKey().evaluateMessage(record, record.getRecord());
+            for (Key.Evaluated key : evaluatedList) {
+                Object value = key.values().get(0);
```
MMcM, line 262: This is taking the first column of every row, while I believe it wants to take every column of the first row.

The contract is to only get a single grouping key for each record. Disallowing, for instance, a map-like repeated nested message where every value is _separately_ indexed according to its key. Which is probably okay for right now. So using `evaluateSingleton` might be appropriate.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
```diff
@@ -109,25 +109,28 @@ protected void doMerge(final IndexWriter writer, final MergePolicy.OneMerge merg
...
             }
             return writerCheck.indexWriter;
         }
     }
 
     @Nonnull
-    private static String getWriterName(@Nonnull final IndexMaintainerState state) {
+    private static String getWriterName(@Nonnull final IndexMaintainerState state, @Nullable final String groupingKey) {
+        if (groupingKey != null) {
+            return "writer$" + state.index.getName() + "$" + groupingKey;
```
MMcM, line 132: I suppose it's harmless, but won't this have two dollar signs?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -240,6 +249,24 @@ private void insertField(DocumentEntry entry, final Document document) {
...
+        KeyExpression expression = state.index.getRootExpression();
+        if (expression instanceof GroupingKeyExpression) {
+            final List<DocumentEntry> groupingExpressions = getFields(((GroupingKeyExpression)expression).getGroupingSubKey(), record, record.getRecord());
+            if (groupingExpressions.isEmpty()) {
+                return null;
+            }
+            String indexPreface = "";
+            Key.Evaluated evaluatedKey = ((GroupingKeyExpression)expression).getGroupingSubKey().evaluateSingleton(record);
+            if (!evaluatedKey.containsNonUniqueNull()) {
+                Object value = evaluatedKey.values().get(0);
```
MMcM, line 262: This doesn't appear to match `scan`, which takes all the keys. If I'm right, then a test with a multi-field grouping should be failing. If we don't have one like that, perhaps we should.

MMcM (changes requested):

MMcM (approved):

PR #1403, opened 2021-09-24
Resolves #1402: Add per-transaction metrics and commit failure metrics

### LINKED ISSUE #1402: Publish per-transaction I/O metrics
In order to build realistic simulations of workload, we need the ability to capture I/O rates per-transaction (e.g. reads-per-transaction, writes-per-transaction, etc.).  This should be pretty easy to add to `InstrumentedTransaction`.  In addition, it would be worthwhile to have metrics on commit failures (by reason, such as conflicts)

scgray:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -63,10 +65,24 @@
...
+    // Set of metrics that are accumulated and published once over the entire life of the transaction to allow for
+    // an operations-per-transaction view of things, even if the overall store timer is accumulating stats across
+    // multiple transactions.
+    private boolean transactionMetricsPublished = false;
+    protected final AtomicInteger transactionReads = new AtomicInteger();
+    protected final AtomicInteger transactionWrites = new AtomicInteger();
+    protected final AtomicInteger transactionDeletes = new AtomicInteger();
+    protected final AtomicInteger transactionBytesRead = new AtomicInteger();
+    protected final AtomicInteger transactionBytesWritten = new AtomicInteger();
+    protected final AtomicInteger transactionMutations = new AtomicInteger();
```
MMcM, line 79: Do you want per-transaction `EventKeeper` info, too? If so, I think the basic idea would be to pull this out into a separate class, move the interface implementation onto that and init the various objects in the right order in `FDBDatabase.createTransaction`.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
scgray: I completely re-did the way it works.  Not only are you correct that the EventKeeper data is useful (so I changed how it works), but that my goal of capturing I/O distributions by transaction wasn't reasonably feasible the way I had things implemented.  The new commit has a description of my thinking in it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
```diff
@@ -561,6 +565,28 @@ public static int getThreadsPerClientVersion() {
...
 
+    /**
+     * Installs a listener that will be handed a collection of low-level I/O metrics that were gathered
+     * over the lifetime of a transaction. These metrics will be issued the listener as soon as a transaction
+     * is committed or closed. Because this listener is in the path of the completion of every transaction,
+     * implementations should take care to ensure that they are thread safe and efficiently process the provided
+     * metrics before returning.
+     *
+     * @param transactionMetricListener a consumer that accepts the database for which the transaction was
+     *   committed or closed as well as a {@code StoreTimer} containing the I/O metrics for the transaction
```
MMcM, line 576: This works, but it kind of feels like it's deserving of its own interface instead of just `BiConsumer`.

Might a listener want to know whether the transaction was committed or rolled back?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactory.java
scgray: Yeah, I was just recently thinking about something like that myself. Good suggestion. Done.

MMcM (approved):

PR #1411, opened 2021-09-30
Adding Executor service to Lucene searchers

sophiabuell:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -247,7 +250,11 @@ private synchronized IndexReader getIndexReader() throws IOException {
     private void performScan() throws IOException {
         long startTime = System.nanoTime();
         indexReader = getIndexReader();
-        searcher = new IndexSearcher(indexReader);
+        if (executorService == null) {
+            searcher = new IndexSearcher(indexReader);
+        } else {
+            searcher = new IndexSearcher(indexReader, executorService);
+        }
```
MMcM, line 257: I don't know whether there are concerns about compatibility, but it looks like the one-arg constructor just calls the two-arg with `null` itself.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
sophiabuell: yes that was my thought. I can change it to set the Executor to null?

MMcM: As far as I can tell, doing
```
searcher = new IndexSearcher(indexReader, executorService);
```
would work regardless today. The only question is whether this is an accident and might change at some point.

MMcM (approved):

PR #1413, opened 2021-10-04
Fixes #1412: Store timers to track compression efficacy

### LINKED ISSUE #1412: Add store timers to track compression efficacy
We have metrics around whether or not compression worked (we reduced the size), but not metrics around how effective the compression is.

scgray:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/RecordSerializer.java
```diff
@@ -145,7 +146,12 @@ public String logKey() {
     @API(API.Status.UNSTABLE)
     enum Counts implements StoreTimer.Count {
         /** The number of times that record compression was not effective and the record was kept uncompressed. */
-        ESCHEW_RECORD_COMPRESSION("eschew record compression");
+        ESCHEW_RECORD_COMPRESSION("eschew record compression"),
+        /** Total record bytes for which compression was attemped. */
```
MMcM, line 150:
```suggestion
        /** Total record bytes for which compression was attempted. */
```

MMcM (changes requested):

PR #1423, opened 2021-10-11
Resolves #1424: Add count and countNonNull to streaming aggregate

### LINKED ISSUE #1424: Add support for count and countNonNull in streaming aggregate
Streaming aggregate support should have operators for count (including nulls) and countNonNull.

ohadzeliger:
Add support and tests for count and countNonNull: #1424. 
Specifically, the following are true for aggregations and counts:
1. Zero records aggregate to NULL (e.g. Sum() and Avg() without any records return null). 
2. Exception to the rule above is *Count*, where:
    2.1. count() with *no aggregation criteria and no records* is 0.
    2.2. count() with *some aggregation criteria and no records* returns empty stream. (Same as SQL). 
3. Empty table (no records at all ) returns empty result (no records).
4. All aggregations *ignore* null values: They are considered a no-op for avg(), sum() etc.
5. Exception to the above rule is *Count* where count() with will include null values.
6. In order to exclude null records from the count, use the countNonNull() aggregation.


ohadzeliger:
Closing as this is handled by other changes.

PR #1443, opened 2021-10-27
Shell scripts and docker-compose config for running …

MMcM:
… some version of FDB locally and configuring Record Layer tests to use it.

### DIFF
--- a/docker-local/start.sh
+++ b/docker-local/start.sh
@@ -1,0 +1,91 @@
+#! /bin/bash
+
+#
+# start.sh
+#
+# This source file is part of the FoundationDB open source project
+#
+# Copyright 2021 Apple Inc. and the FoundationDB project authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


docker-local/start.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+
+set -eu
+
+ROOTDIR=$(cd $(dirname $0)/.. && pwd)
+
+if [ $(uname -s) == 'Darwin' ]; then
+    OS=macOS
+    SO_SUFFIX=dylib
+    LIBRARY_PATH=DYLD_LIBRARY_PATH
```
alecgrieser, line 30: Given that `DYLD_LIBRARY_PATH` only works if System Integrity Protection is turned off, we might need to replace this with something that uses the [`install_name_tool`](https://stackoverflow.com/questions/39927235/alternative-for-the-dyld-library-path-trick-since-mac-os-10-11-el-capitan-with-s) instead. However, for those that are willing to turn it off, this could still be useful, so this is probably a good option to have (at least for now)

alecgrieser (approved):

MMcM:
test this please

MMcM:
test this please

PR #1444, opened 2021-11-01
[WIP] lucene multi file reference capabilities

sophiabuell:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -282,32 +346,23 @@ private void record(StoreTimer.Event event, long durationNanos) {
         long totalSize = 0L;
         for (KeyValue kv : context.ensureActive().getRange(metaSubspace.range())) {
             String name = metaSubspace.unpack(kv.getKey()).getString(0);
+            Tuple referenceTuple = Tuple.fromBytes(kv.getValue());
+            if (referenceTuple.get(0) instanceof ArrayList) {
+                FDBLuceneMiltiFileReference multiFileReference = new FDBLuceneMiltiFileReference(Tuple.fromBytes(kv.getValue()));
```
tian-yizuo, line 351: Why not use the ```referenceTuple``` directly here, and also for the other 2 places below?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -196,7 +248,19 @@ private void incrementCallCount(FDBStoreTimer.Counts counter) {
     public void writeFDBLuceneFileReference(@Nonnull final String name, @Nonnull final FDBLuceneFileReference reference) {
         LOGGER.trace("writeFDBLuceneFileReference {}", reference);
         incrementCallCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE);
-        context.ensureActive().set(metaSubspace.pack(name), reference.getTuple().pack());
+        String mergedFileName = getMergedFileName(name);
+        if (mergedFileName != null) {
+            FDBLuceneMiltiFileReference multiFileReference = context.asyncToSync(FDBStoreTimer.Waits.WAIT_LUCENE_FILE_LENGTH, getMultiFDBLuceneFileReference(mergedFileName));
```
tian-yizuo, line 253: My understanding of the motivation for this PR is to reduce the read requests to FoundationDB server for file references for compounded index. Looks one side effect is to have read-and-write instead of only one write when we want to write the reference for the 3 files to FoundationDB. Was there any discussion or consideration on the tradeoff between the less reads and potentially more expensive writes of the references? This comment is not to request any changes, just to understand more context about this optimization.

PR #1448, opened 2021-11-04
Example of Multi-Layer Lucene Optimizations

jleach4:
* Codec with a few optimizations for speeding up compound files
 * sitting on FoundationDB.
 *
 * Optimizations:
 * - .cfe file references are not stored
 * - .cfe file data is serialized as entries on the current compound file reference (.cfs)
 * - .si file references are not stored
 * - .si file data is serialized as segmentInfo on the current compound file reference (.cfs)
 * - .si diagnostic information is not stored
 * - Removed checksum validation on Compound File Reader

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCodec.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCodec.java
@@ -1,0 +1,126 @@
+/*
+ * LuceneOptimizedCodec.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene.codec;
+
+import org.apache.lucene.codecs.Codec;
+import org.apache.lucene.codecs.CompoundFormat;
+import org.apache.lucene.codecs.DocValuesFormat;
+import org.apache.lucene.codecs.FieldInfosFormat;
+import org.apache.lucene.codecs.LiveDocsFormat;
+import org.apache.lucene.codecs.NormsFormat;
+import org.apache.lucene.codecs.PointsFormat;
+import org.apache.lucene.codecs.PostingsFormat;
+import org.apache.lucene.codecs.SegmentInfoFormat;
+import org.apache.lucene.codecs.StoredFieldsFormat;
+import org.apache.lucene.codecs.TermVectorsFormat;
+import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
+import org.apache.lucene.codecs.lucene70.Lucene70Codec;
+
+/**
+ *
+ * Codec with a few optimizations for speeding up compound files
+ * sitting on FoundationDB.
+ *
+ * Optimizations:
+ * - .cfe file references are not stored
+ * - .cfe file data is serialized as entries on the current compound file reference (.cfs)
+ * - .si file references are not stored
+ * - .si file data is serialized as segmentInfo on the current compound file reference (.cfs)
+ * - .si diagnostic information is not stored
+ * - Removed checksum validation on Compound File Reader
+ */
+public class LuceneOptimizedCodec extends Codec {
+
+    private final Lucene70Codec baseCodec;
+    private final CompoundFormat compoundFormat;
+    private final LuceneOptimizedSegmentInfoFormat segmentInfoFormat;
+
+    /**
+     * Instantiates a new codec.
+     */
+    public LuceneOptimizedCodec() {
+        this(Lucene50StoredFieldsFormat.Mode.BEST_SPEED);
+    }
+
+    /**
+     * Instantiates a new codec, specifying the stored fields compression
+     * mode to use.
+     * @param mode stored fields compression mode to use for newly
+     *             flushed/merged segments.
+     */
+    public LuceneOptimizedCodec(Lucene50StoredFieldsFormat.Mode mode) {
+        super("RL");
+        baseCodec = new Lucene70Codec(mode);
+        compoundFormat = new LuceneOptimizedCompoundFormat();
+        segmentInfoFormat = new LuceneOptimizedSegmentInfoFormat();
+    }
+
+
+    @Override
+    public PostingsFormat postingsFormat() {
+        return baseCodec.postingsFormat();
+    }
+
+    @Override
+    public DocValuesFormat docValuesFormat() {
+        return baseCodec.docValuesFormat();
+    }
+
+    @Override
+    public StoredFieldsFormat storedFieldsFormat() {
+        return baseCodec.storedFieldsFormat();
+    }
+
+    @Override
+    public TermVectorsFormat termVectorsFormat() {
+        return baseCodec.termVectorsFormat();
+    }
+
+    @Override
+    public FieldInfosFormat fieldInfosFormat() {
+        return baseCodec.fieldInfosFormat();
+    }
+
+    @Override
+    public SegmentInfoFormat segmentInfoFormat() {
+        return segmentInfoFormat;
+    }
+
+    @Override
+    public NormsFormat normsFormat() {
+        return baseCodec.normsFormat();
+    }
+
+    @Override
+    public LiveDocsFormat liveDocsFormat() {
+        return baseCodec.liveDocsFormat();
+    }
+
+    @Override
+    public CompoundFormat compoundFormat() {
+        return compoundFormat;
+    }
+
+    @Override
+    public PointsFormat pointsFormat() {
+        return baseCodec.pointsFormat();
+    }
+}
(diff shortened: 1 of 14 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -350,7 +373,14 @@ public void deleteFile(@Nonnull String name) throws IOException {
     @Override
     public long fileLength(@Nonnull String name) throws NoSuchFileException {
         LOGGER.trace("fileLength -> {}", name);
+        name = convertToDataFile(name);
         FDBLuceneFileReference reference = context.asyncToSync(FDBStoreTimer.Waits.WAIT_LUCENE_FILE_LENGTH, getFDBLuceneFileReference(name));
+        if (name.endsWith(".cfe")) {
```
sophiabuell, line 378: There seem to be a lot of places where these are hardcoded. If we are going to use term vectors I believe we'll want to add those into the compressed reference as well.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedSegmentInfoFormat.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public void write(final Directory dir, final SegmentInfo info, final IOContext ioContext) throws IOException {
+        SegmentInfo segmentInfoToWrite = new SegmentInfo(new LuceneOptimizedWrappedDirectory(dir), info.getVersion(), info.getMinVersion(), info.name,
+                info.maxDoc(), info.getUseCompoundFile(), info.getCodec(), Collections.emptyMap(),
+                info.getId(),
+                info.getAttributes(), info.getIndexSort());
+        segmentInfoToWrite.setFiles(info.files());
+        segmentInfosFormat.write(new LuceneOptimizedWrappedDirectory(dir), segmentInfoToWrite, ioContext);
```
tian-yizuo, line 58: Could the ```new LuceneOptimizedWrappedDirectory(dir)``` above be declared as a variable and reused here?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedSegmentInfoFormat.java
```diff
@@ -0,0 +1,62 @@
...
+        final String fileName = IndexFileNames.segmentFileName(info.name, "", Lucene70SegmentInfoFormat.SI_EXTENSION);
+        info.setFiles(info.files().stream().filter(file -> !file.equals(fileName)).collect(Collectors.toSet()));
```
tian-yizuo, line 60: Not to request change for this. Just wondering is this line just used to remove the ```si``` from existing segmentInfo in disk before this optimization?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+         */
+    }
+
+    /** Helper method that reads CFS entries from an input stream. */
+    @SuppressWarnings("PMD.AvoidCatchingThrowable")
+    private Map<String, FileEntry> readEntries(byte[] segmentID, Directory dir, String entriesFileName) throws IOException {
+        Map<String, FileEntry> mapping = null;
+        try (ChecksumIndexInput entriesStream = dir.openChecksumInput(entriesFileName, IOContext.READONCE)) {
```
sophiabuell, line 106: Aren't we trying to avoid checksumming the files?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedWrappedDirectory.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.lucene.store.Lock;
+
+import java.io.IOException;
+import java.util.Collection;
+
+import static com.apple.foundationdb.record.lucene.codec.LuceneOptimizedCompoundFormat.DATA_EXTENSION;
+import static com.apple.foundationdb.record.lucene.codec.LuceneOptimizedCompoundFormat.ENTRIES_EXTENSION;
+import static org.apache.lucene.codecs.lucene70.Lucene70SegmentInfoFormat.SI_EXTENSION;
+
+public class LuceneOptimizedWrappedDirectory extends Directory {
```
sophiabuell, line 40: Could a lot of this functionality been incorporated into the FDBDirectory?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedSegmentInfoFormat.java
sophiabuell: This is to prevent the format from saving the segment info because we are instead storing it within the file reference as bytes[].  the Info keeps track of all the files that are written to the segment and is used later at some places to read those files or validate their existence.

sophiabuell: Another related but issue (not with this pr) is that the segmentInfo actually marks the files (non compound) as deleted but they are still on the server when we query files/file sizes. It would be nice to clean these up. I think this happens because of lucene tombstones instead of hard delete calls.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
```diff
@@ -0,0 +1,218 @@
...
+            Throwable priorE = null;
+            try {
+                CodecUtil.checkIndexHeader(entriesStream, LuceneOptimizedCompoundFormat.ENTRY_CODEC,
```
sophiabuell, line 109: It seems like here is where the validation check is happening between the previous write and the current read. This is where we identify a corrupt index. If we're removing the checksum this is probably the line we don't want?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedWrappedDirectory.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+    }
+
+    public LuceneOptimizedWrappedDirectory(Directory directory) {
+        this.wrappedDirectory = directory;
+        this.fdbDirectory = (FDBDirectory)FilterDirectory.unwrap(directory);
+    }
+
+    @Override
```
sophiabuell, line 72: Is there a reason we're specifically calling these out as unsupported?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
sophiabuell: Oh fun times, it seems like this specifically loads the entries stream. Lucene is annoying sometimes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedSegmentInfoFormat.java
jleach4: You are not seeing deletes of the initial non-compound files?  I would file a radar for someone to fix.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
jleach4: We are trying to avoid reading the last block of each data file to retrieve the checksum.  I did not change this code because it is reading a byte[] serialized into the FileReference record.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedWrappedDirectory.java
jleach4: It is a good question.  I am fine if that is the approach we want to take.  I would ask however that we at least keep the removal of fetching the last block in the Reader.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedWrappedDirectory.java
jleach4: The Wrapped Directory was for input and output in a specific case and I did not want to think this is a general directory implementation.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCodec.java
```diff
@@ -0,0 +1,126 @@
...
+ * - .cfe file data is serialized as entries on the current compound file reference (.cfs)
+ * - .si file references are not stored
+ * - .si file data is serialized as segmentInfo on the current compound file reference (.cfs)
+ * - .si diagnostic information is not stored
+ * - Removed checksum validation on Compound File Reader
+ */
+public class LuceneOptimizedCodec extends Codec {
+
+    private final Lucene70Codec baseCodec;
+    private final CompoundFormat compoundFormat;
```
tian-yizuo, line 53: Why the ```compoundFormat``` is declared as the base class but ```segmentInforFormat``` is declared as the inheritance one?

PR #1455, opened 2021-11-09
resolves #1454: Up java source and target compatibility to 11

### LINKED ISSUE #1454: Up java source and target compatibility to 11
It's one better!

normen662:


normen662:
test this please

gradle.properties
```diff
@@ -19,7 +19,7 @@
 #
 
 rootProject.name=fdb-record-layer
-version=3.0
+version=3.1
```
alecgrieser, line 22: A good start, but I believe there are a few more things that need to be done as part of the version bump:

1. ReleaseNotes.md needs to be updated so that the description of what changed in 3.0 is below the "next_release" block. It would also be good to add a 3.1 header and a description of what's new in 3.1 (which is fine to be just the Java 11 changes)
2. One of our pages in the docs has the current version in it just for reference. That one is less important to update, but probably good practice while we're there.

I believe that you should be able to model the changes after the changes in this commit (bumping to 3.0): https://github.com/FoundationDB/fdb-record-layer/commit/9a80491bfe43e870a31dd17a5f46b31d55a12672

alecgrieser (changes requested):
Java 11 changes look good to me. (Hopefully the sonar failure is unrelated. 😬) I left a few notes about upgrading the minor version to 3.1, which hopefully is straightforward to resolve.

gradle.properties
normen662: Done.

docs/ReleaseNotes.md
```diff
@@ -214,7 +220,7 @@ Another, smaller change that has been made is that by default, new indexes added
 
 ### Breaking Changes
 
-In this realase, the various implementations of the `RecordQueryPlan` interface have  moved to API stability level `INTERNAL`. This means that individual implementations may change without notice. Clients that are not creating `RecordQueryPlan` objects directly (but instead using the planner to create plans) should not be affected.
+In this release, the various implementations of the `RecordQueryPlan` interface have  moved to API stability level `INTERNAL`. This means that individual implementations may change without notice. Clients that are not creating `RecordQueryPlan` objects directly (but instead using the planner to create plans) should not be affected.
```
alecgrieser, line 223: Nice

alecgrieser (approved):

Issue #1288, opened 2021-05-20
Publish recent FoundationDB Record layer version to maven repository


sebastienh:
Is there any plan to release an updated version of the record layer jar to a public maven repository? As we speak the latest version released is `2.8.110.0` which is more than one year old. The instructions in [Record Layer Versions](https://foundationdb.github.io/fdb-record-layer/Versioning.html) suggest that the latest version should be available... 

```xml
  <groupId>org.foundationdb</groupId>
  <artifactId>fdb-record-layer-core</artifactId>
  <version>3.0.+</version>
``` 

Thanks for such a nice piece of software :)

mping:
Any updates on this?

sebastienh:
The lastest I have is this: 

> [We are working on getting Maven Central publishing back on line.](https://github.com/FoundationDB/fdb-record-layer/issues/1351#:~:text=We%20are%20working%20on%20getting%20Maven%20Central%20publishing%20back%20on%20line.)

calavera:
is anyone working on this @alecgrieser? is the problem something the community can help with? it's a little bit concerning that there have been so many releases without artifacts being published to Maven Central.

alecgrieser:
I believe @ammolitor has been looking into this along with some other changes to this project (and the FoundationDB core project's) CI system. The main reason this broke has to do with some backend changes to how our publishing pipeline was orchestrated, and we're still working on replacing what the old backend did with the new system.

In terms of ways the community could help, one way I could see might be to submit a patch to our publishing section to add a step to publish directly to a maven central repo (like sonatype) analogous to how we currently publish to artifactory: https://github.com/FoundationDB/fdb-record-layer/blob/233a72a51182a1f35b9bad34ac29b127a437a12b/gradle/publishing.gradle#L65 I think someone on the Apple team would still need to do things like set up the credentials, but if there are places to inject secrets (via gradle or environment variable configuration), I think it could be made to work. 

[...]

PR #1458, opened 2021-11-10
Workaround #1288: Publish recent FoundationDB Record layer version to maven repository

alecgrieser:
While we're working on fixing #1288, this documents a workaround using an alternative maven repo from maven central. Once we start publishing to central again, this can be removed, but it documenting this fix for the interim should hopefully help unblock people.


PierreZ:
Thanks @alecgrieser 🙏

ammolitor (approved):

PR #1461, opened 2021-11-10
resolves #1461: parameter comparisons should declare their correlations correctly

normen662:
- `Comparison` now implements `Correlated<Comparison>`
- `ComparisonRange` now implements `Correlated<ComparisonRange`
- `ValuePredicate` and `ValueCoparisonRange.Sargable` both defer to the comparisons/comparison range's `rebase()` in their own `rebase()`.
- matching logic in `SelectExpression` can now match even if predicates declared multiple correlations
- `Compensation` logic was extended to also allow for complicated join-like `Compensation` trees including the compensation of join predicates (aka multi-correlated comparisons)

Note that all of this is necessary to allow `IN-to-UNION` an `IN-to-JOIN` to reason about order better and to allow us to generalize these rules to multiple in clauses and proper order property housekeeping.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
@@ -38,6 +38,9 @@ import com.apple.foundationdb.record.provider.common.text.TextTokenizerRegistryI
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
 import com.apple.foundationdb.record.provider.foundationdb.cursors.ProbableIntersectionCursor;
 import com.apple.foundationdb.record.query.ParameterRelationshipGraph;
+import com.apple.foundationdb.record.query.plan.temp.AliasMap;
+import com.apple.foundationdb.record.query.plan.temp.Correlated;
+import com.apple.foundationdb.record.query.plan.temp.CorrelationIdentifier;
 import com.apple.foundationdb.record.util.HashUtils;
 import com.apple.foundationdb.tuple.ByteArrayUtil;
 import com.apple.foundationdb.tuple.ByteArrayUtil2;
@@ -639,7 +642,7 @@ public class Comparisons {
      * A comparison between a value associated with someplace in the record (such as a field) and a value associated
      * with the plan (such as a constant or a bound parameter).
      */
-    public interface Comparison extends PlanHashable, QueryHashable {
+    public interface Comparison extends PlanHashable, QueryHashable, Correlated<Comparison> {
         /**
          * Evaluate this comparison for the value taken from the target record.
          * @param store the record store for the query
@@ -701,6 +704,22 @@ public class Comparisons {
         @Nonnull
         default Comparison withParameterRelationshipMap(@Nonnull ParameterRelationshipGraph parameterRelationshipGraph) {
             return this;
+        }
+
+        @Nonnull
+        @Override
+        default Set<CorrelationIdentifier> getCorrelatedTo() {
+            return ImmutableSet.of();
+        }
+
+        @Override
+        default boolean semanticEquals(@Nullable Object other, @Nonnull AliasMap aliasMap) {
+            return this.equals(other);
+        }
+
+        @Override
+        default int semanticHashCode() {
+            return hashCode();
         }
     }
 
@@ -850,6 +869,12 @@ public class Comparisons {
                     throw new UnsupportedOperationException("Hash Kind " + hashKind.name() + " is not supported");
             }
         }
+
+        @Nonnull
+        @Override
+        public Comparison rebase(@Nonnull final AliasMap translationMap) {
+            return this;
+        }
     }
 
     /**
@@ -923,6 +948,17 @@ public class Comparisons {
             return type;
         }
 
+        public boolean isCorrelation() {
+            return internal == Bindings.Internal.CORRELATION;
+        }
+
+        public boolean isCorrelatedTo(@Nonnull final CorrelationIdentifier alias) {
+            if (!isCorrelation()) {
+                return false;
+            }
+            return Bindings.Internal.CORRELATION.identifier(getParameter()).equals(alias.getId());
+        }
+
         @Nullable
         @Override
         public Object getComparand(@Nullable FDBRecordStoreBase<?> store, @Nullable EvaluationContext context) {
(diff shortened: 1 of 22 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -923,6 +948,13 @@ public Type getType() {
             return type;
         }
 
+        public boolean isCorrelatedTo(@Nonnull final CorrelationIdentifier alias) {
+            if (internal != Bindings.Internal.CORRELATION) {
```
MMcM, line 952: Maybe this predicate, already used in `planHash`, should be its own method like `isCorrelation`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -964,18 +1043,26 @@ public String getParameter() {
...
-            }
-            if (o == null || getClass() != o.getClass()) {
-                return false;
+            final AliasMap aliasMap;
+            if (internal == Bindings.Internal.CORRELATION) {
+                final var alias = CorrelationIdentifier.of(Bindings.Internal.CORRELATION.identifier(parameter));
+                aliasMap = AliasMap.identitiesFor(ImmutableSet.of(alias));
+            } else {
+                aliasMap = AliasMap.emptyMap();
             }
```
MMcM, line 1064: Do we not expect `equals` to be done very often on / around these? Because this is now doing allocation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -932,6 +964,53 @@ public Object getComparand(@Nullable FDBRecordStoreBase<?> store, @Nullable Eval
             return context.getBinding(parameter);
         }
 
+        @Nonnull
+        @Override
+        public Comparison rebase(@Nonnull final AliasMap translationMap) {
+            if (internal == Bindings.Internal.CORRELATION) {
+                final var alias = CorrelationIdentifier.of(internal.identifier(parameter));
+                final var translatedAlias = translationMap.getTargetOrDefault(alias, alias);
```
MMcM, line 972: This is introducing `var` in ways that I am not certain there is consensus around yet.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
normen662: Makes sense! Will do that!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
normen662: I am essentially using IntelliJ here. When in Java 11 mode, by default when you introduce a variable it now goes to `var` instead of the static type.  If you don't do that IntelliJ gives you a code-style warning. We can change all that. Is there anything specific about this you don't like?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
normen662: This is already done in a lot of `structuralEquals()` and `semanticEquals()` methods. I can make `emptyMap()` return a static instance so the general (the old-planner code path) is not affected by this at all as that one should never create an actual correlation. I can also minimize the allocations by creating a special subclass of `AliasMap()` that just implements identities thus there won't be any  need for a copy of the aliases.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/Correlated.java
```diff
@@ -151,6 +151,26 @@
...
+     * Static helper to shorten boilerplate for callers that have to deal with nullable objects.
+     * @param one one instance
+     * @param other another instance
+     * @param aliasMap an {@link AliasMap}
+     * @param <S> the type parameter for {@code Correlated}
+     * @return This method returns {@code true} if both objects passed in are {@code null}, {@code false}
+     *         if {@code one} is {@code null} but {@code another} is not and defers to the instance-based semantic
+     *         equality in any other case.
+     */
+    static <S extends Correlated<S>> boolean semanticEquals(@Nullable Correlated<S> one, @Nullable Correlated<S> other, @Nonnull AliasMap aliasMap) {
```
ohadzeliger, line 164: package protested?  Should this be public?

normen662: It's defacto `public` as `Correlated` is an interface.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/ComparisonRange.java
```diff
@@ -158,6 +161,106 @@ public Type getRangeType() {
...
+        if (isInequality()) {
+            if (!that.isInequality()) {
+                return false;
+            }
+            Objects.requireNonNull(this.inequalityComparisons);
+            Objects.requireNonNull(that.inequalityComparisons);
+            if (inequalityComparisons.size() != that.inequalityComparisons.size()) {
+                return false;
+            }
+            return Streams.zip(inequalityComparisons.stream(),
```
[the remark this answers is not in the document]
normen662: Good catch! They are lists for some reason but they actually shouldn't be conceptually. I fixed that to a nested loop comparison which sucks. Maybe we should change that to a `Set` in a follow up and at least get the hashing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/ComparisonRange.java
```diff
@@ -158,6 +161,106 @@ public Type getRangeType() {
...
+    @Override
+    public boolean semanticEquals(@Nullable final Object other, @Nonnull final AliasMap aliasMap) {
+        if (this == other) {
+            return true;
+        }
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        ComparisonRange that = (ComparisonRange)other;
+        if (isEquality()) {
```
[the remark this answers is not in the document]
normen662: `Correlated.semanticEquals()` is null -friendly and should deal with that. Also, in this class it's either/or, meaning it's either an equality or it is a set of inequalities, nothing in between. I added

```
Verify.verify(isEquality() ^ isInequality());
Verify.verify(that.isEquality() ^ that.isInequality());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
normen662: Made `emptyMap()` return a neutral element. In this way the old-planner code path is not punished.

MMcM (approved):

PR #1462, opened 2021-11-11
Support compression for lucene data, with the extensibility to suppor…

tian-yizuo:
…t multiple compression modes and data encryption in future

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -29,7 +29,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Breaking change** Change 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Breaking change** Support compression for Lucene data. Backwards compatibility is not supported, so upgrading requires rebuilding Lucene indexes during which search requests cannot be completed. [(Issue #1466)](https://github.com/FoundationDB/fdb-record-layer/issues/1466)
 * **Breaking change** Change 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Breaking change** Change 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Breaking change** Change 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -188,7 +188,7 @@ public class FDBDirectory extends Directory {
         if (fileReference == null) {
             return context.instrument(FDBStoreTimer.Events.LUCENE_GET_FILE_REFERENCE, context.ensureActive().get(metaSubspace.pack(name))
                     .thenApplyAsync((value) -> {
-                        final FDBLuceneFileReference fetchedref = FDBLuceneFileReference.parseFromBytes(value);
+                        final FDBLuceneFileReference fetchedref = FDBLuceneFileReference.parseFromBytes(LuceneSerializer.decode(value));
                         if (fetchedref != null) {
                             this.fileReferenceCache.put(name, fetchedref);
                         }
@@ -248,11 +248,18 @@ public class FDBDirectory extends Directory {
                 reference.setEntries(storedRef.getEntries());
             }
         }
-        LOGGER.trace("writeFDBLuceneFileReference {}", reference);
         final byte[] fileReferenceBytes = reference.getBytes();
-        increment(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_SIZE, fileReferenceBytes.length);
+        final byte[] encodedBytes = LuceneSerializer.encode(reference.getBytes(), true, false);
+        increment(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_SIZE, encodedBytes.length);
         increment(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_CALL);
-        context.ensureActive().set(metaSubspace.pack(name), fileReferenceBytes);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(KeyValueLogMessage.of("Write lucene file reference",
+                    LogMessageKeys.FILE_NAME, name,
+                    LogMessageKeys.DATA_SIZE, fileReferenceBytes.length,
+                    LogMessageKeys.ENCODED_DATA_SIZE, encodedBytes.length,
+                    LogMessageKeys.FILE_REFERENCE, reference));
+        }
+        context.ensureActive().set(metaSubspace.pack(name), encodedBytes);
         fileReferenceCache.put(name, reference);
     }
 
@@ -263,12 +270,19 @@ public class FDBDirectory extends Directory {
      * @param value the data to be stored
      */
     public void writeData(long id, int block, @Nonnull byte[] value) {
+        final byte[] encodedBytes = LuceneSerializer.encode(value, true, false);
         //This may not be correct transactionally
-        increment(FDBStoreTimer.Counts.LUCENE_WRITE_SIZE, value.length);
+        increment(FDBStoreTimer.Counts.LUCENE_WRITE_SIZE, encodedBytes.length);
         increment(FDBStoreTimer.Counts.LUCENE_WRITE_CALL);
-        LOGGER.trace("writeData id={}, block={}, valueSize={}", id, block, value.length);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(KeyValueLogMessage.of("Write lucene data",
+                    LogMessageKeys.FILE_ID, id,
+                    LogMessageKeys.BLOCK_NUMBER, block,
+                    LogMessageKeys.DATA_SIZE, value.length,
+                    LogMessageKeys.ENCODED_DATA_SIZE, encodedBytes.length));
+        }
         Verify.verify(value.length <= blockSize);
-        context.ensureActive().set(dataSubspace.pack(Tuple.from(id, block)), value);
+        context.ensureActive().set(dataSubspace.pack(Tuple.from(id, block)), encodedBytes);
     }
 
     /**
@@ -293,7 +307,9 @@ public class FDBDirectory extends Directory {
 
             long start = System.nanoTime();
             return context.instrument(FDBStoreTimer.Events.LUCENE_READ_BLOCK,blockCache.get(Pair.of(id, block),
-                    () -> context.instrument(FDBStoreTimer.Events.LUCENE_FDB_READ_BLOCK, context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block))))
+                    () -> context.instrument(FDBStoreTimer.Events.LUCENE_FDB_READ_BLOCK,
+                            context.ensureActive().get(dataSubspace.pack(Tuple.from(id, block)))
+                                    .thenApplyAsync(data -> LuceneSerializer.decode(data)))
             ), start);
         } catch (ExecutionException e) {
             throw new RecordCoreException(CompletionExceptionLogHelper.asCause(e));
@@ -337,7 +353,7 @@ public class FDBDirectory extends Directory {
         for (KeyValue kv : context.ensureActive().getRange(metaSubspace.range())) {
             String name = metaSubspace.unpack(kv.getKey()).getString(0);
             outList.add(name);
-            final FDBLuceneFileReference fileReference = FDBLuceneFileReference.parseFromBytes(kv.getValue());
+            final FDBLuceneFileReference fileReference = FDBLuceneFileReference.parseFromBytes(LuceneSerializer.decode(kv.getValue()));
             // Only composite files are prefetched.
             if (name.endsWith(".cfs")) {
                 try {
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
@@ -1,0 +1,193 @@
+/*
+ * LuceneSerializer.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene.directory;
+
+import com.apple.foundationdb.record.RecordCoreException;
+import com.apple.foundationdb.record.logging.KeyValueLogMessage;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
+import org.apache.lucene.codecs.compressing.CompressionMode;
+import org.apache.lucene.codecs.compressing.Compressor;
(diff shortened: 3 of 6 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,189 @@
...
+    public static byte[] decode(@Nonnull String resourceDescription, @Nullable byte[] data) {
+        if (data == null) {
+            return null;
+        }
+        if (data.length < 2) {
+            throw new RecordCoreException("Invalid data")
+                    .addLogInfo(LogMessageKeys.DATA_VALUE, data);
+        }
+
+        final byte encoding = data[0];
```
jleach4, line 107: Usually when you are compressing or encrypting you act on streams of data vs. byte[].  If you do that, you lose the extra array copy.

jleach4 (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -181,7 +181,7 @@ private void incrementCallCount(FDBStoreTimer.Counts counter) {
         if (fileReference == null) {
             return context.instrument(FDBStoreTimer.Events.LUCENE_GET_FILE_REFERENCE, context.ensureActive().get(metaSubspace.pack(name))
                     .thenApplyAsync((value) -> {
-                        final FDBLuceneFileReference fetchedref = FDBLuceneFileReference.parseFromBytes(value);
+                        final FDBLuceneFileReference fetchedref = FDBLuceneFileReference.parseFromBytes(LuceneSerializer.decode("LuceneFileReference for <" + name + ">", value));
```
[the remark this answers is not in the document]
MMcM: Also, it's doing it always for the sake of trace-level debugging.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,189 @@
...
+            final ByteBuffersDataOutput output = new ByteBuffersDataOutput();
+            output.writeVInt(data.length);
+            compressor.compress(data, 0, data.length, output);
+            final byte[] compressedBytes = output.toArrayCopy();
+
+            // Compress only if it helps to shorten the bytes
+            if (compressedBytes.length + 1 < data.length) {
+                state.setCompressed(true);
+                final byte[] compressedData = new byte[compressedBytes.length + 1];
+                compressedData[0] = (byte) 0; // The version 0 representing HIGH_COMPRESSION, this can be extended to support more compression modes in future
```
MMcM, line 136: Isn't that `COMPRESSION_VERSION_FOR_HIGH_COMPRESSION`? Saving a comment and a cast?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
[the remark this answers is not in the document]
tian-yizuo: Removed the ```resourceDescription``` for trace logging.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -243,7 +243,7 @@ public void writeFDBLuceneFileReference(@Nonnull String name, @Nonnull FDBLucene
         }
         LOGGER.trace("writeFDBLuceneFileReference {}", reference);
         incrementCallCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE);
-        context.ensureActive().set(metaSubspace.pack(name), reference.getBytes());
+        context.ensureActive().set(metaSubspace.pack(name), LuceneSerializer.encode("LuceneFileReference for <" + name + "> with id <" + reference.getId() + ">", reference.getBytes(), true, false));
```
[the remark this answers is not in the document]
tian-yizuo: Fixed it as mentioned above.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Fixed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,184 @@
...
+            return null;
+        }
+        if (data.length < 2) {
+            throw new RecordCoreException("Invalid data")
+                    .addLogInfo(LogMessageKeys.DATA_VALUE, data);
+        }
+
+        final byte encoding = data[0];
+        byte[] decoded = new byte[data.length - 1];
+        System.arraycopy(data, 1, decoded, 0, data.length - 1);
```
MMcM, line 109: This one can be avoided by making `decompress` take a byte array and bounds.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Nullable
+    public static byte[] encode(@Nullable byte[] data, boolean compress, boolean encrypt) {
+        if (data == null) {
+            return null;
+        }
+        final EncodingState state = new EncodingState();
+        byte[] encoded = new byte[data.length + 1];
+        System.arraycopy(data, 0, encoded, 1, data.length);
```
MMcM, line 77: This one can be avoided by making the `ByteBuffersDataOutput` earlier and using it to (optimistically) output the encoding byte and then compress. If the compressed `size` is too large, then it's a little bit more complicated to remake an uncompressed, since there is no rewind support. But the happy path will be better.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
jleach4: Yep, offset makes a lot of sense here.

tian-yizuo: Fixed

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Fixed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,201 @@
...
+            throw new RecordCoreException("Invalid data")
+                    .addLogInfo(LogMessageKeys.DATA_VALUE, data);
+        }
+
+        final byte encoding = data[0];
+        final boolean encrypted = (encoding & ENCODING_ENCRYPTED) == ENCODING_ENCRYPTED;
+        final boolean compressed = (encoding & ENCODING_COMPRESSED) == ENCODING_COMPRESSED;
+
+        byte[] decoded = decryptIfNeeded(encrypted, data, 1);
+        decoded = decompressIfNeeded(compressed, decoded, 1);
```
MMcM, line 113: This is a weird contract, as though `decryptIfNeeded` had to decrypt after offset bytes but still return them in the clear text.

Would it make sense to just remove encryption for now in the interest of getting compression done and revisit later, with whatever fallout that has to the API?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: yeah it is weird that the first byte was still kept after decryption. Removed the encryption/decryption methods.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Nonnull
+    private static byte[] decompressIfNeeded(boolean decompressionNeeded, @Nonnull byte[] data, int offset) {
+        if (!decompressionNeeded) {
+            // Return the array to be the final output for decoding, so the offset bytes need to be removed
+            return Arrays.copyOfRange(data, offset, data.length);
+        }
+
+        if (data.length < 2) {
```
MMcM, line 159: I guess there should be a `- offset` here, even though `offset` is never very big today.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Good catch! Fixed

MMcM (dismissed):

MMcM (dismissed):

jleach4 (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Nonnull
+    private static byte[] compressIfNeeded(boolean compressionNeeded, @Nonnull byte[] data, @Nonnull ByteBuffersDataOutput encodedDataOutput,
+                                           @Nonnull EncodingState state, int offset) {
+        if (!compressionNeeded) {
+            return fallBackToUncompressed(encodedDataOutput.toArrayCopy(), data, state, offset);
+        }
+
+        try (Compressor compressor = CompressionMode.HIGH_COMPRESSION.newCompressor()) {
```
alecgrieser, line 125: This is probably fine for now, but it's probably worth considering how we'd make this configurable in the future to allow for user control over high vs fast compression. Maybe this could become an Issue, so we don't lose track of it, or maybe not. Though for version 1, it being stuck on high compression seems fine

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+            }
+        } catch (Exception e) {
+            throw new RecordCoreException("Lucene data compression failure", e);
+        }
+    }
+
+    private static byte[] fallBackToUncompressed(@Nonnull byte[] compressedData, @Nonnull byte[] originalData,
+                                                 @Nonnull EncodingState state, int offset) {
+        state.setCompressed(false);
+        final byte[] encoded = new byte[originalData.length + 1];
```
alecgrieser, line 146: Does this imply that `offset` is always 1 in this case? Otherwise, it would seem that one might try to copy `originalData` off the end of `encoded`.

It also seems like this is trying to add a byte with value 0 to the beginning of the new byte array, which I think is more easily done without the first arraycopy.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+            throw new RecordCoreException("Invalid data")
+                    .addLogInfo(LogMessageKeys.DATA_VALUE, data);
+        }
+
+        final byte encoding = data[0];
+        final boolean encrypted = (encoding & ENCODING_ENCRYPTED) == ENCODING_ENCRYPTED;
+        final boolean compressed = (encoding & ENCODING_COMPRESSED) == ENCODING_COMPRESSED;
+
+        byte[] decoded = decompressIfNeeded(compressed, data, 1);
+        LOGGER.trace("Decoded data for compressed={}, encrypted={}, encodedDataSize={}, decodedDataSize={}",
```
alecgrieser, line 113: Why is it using this style of logging instead of using a `KeyValueLogMessage` (possibly guarded by an `if` block conditional on `LOGGER.isTraceEnabled()`)?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.lucene.store.ByteBuffersDataOutput;
+import org.apache.lucene.util.BytesRef;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Arrays;
+
+public class LuceneSerializer {
```
alecgrieser, line 38: If I understand this correctly, the idea, in terms of forward and backwards compatibility, is that once this serializer is in use, compression (and later encryption) will be able to be turned off and on for a given index without needing to rebuild, but that rolling this out initially will require rebuilding any existing index (much like the previous Lucene disk format change). If that's correct, it's probably worth calling out in the release notes (again), much like the last one

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
alecgrieser: Or maybe this is fixed by just making `originalData.length + 1` into `originalData.length + offset`

tian-yizuo: Good catch! Yes this should be ```+ offset``` instead of ```+ 1```. Fixed it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Makes sense. Changed to ```KeyValueLogMessage``` for the places touched by this PR. I guess the other loggings in ```FDBDirectory``` may also need to be changed to ```KeyValueLogMessage```, but in a separate PR probably.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Please check if the release note looks good to you now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
tian-yizuo: Opened an issue: https://github.com/FoundationDB/fdb-record-layer/issues/1467

docs/ReleaseNotes.md
```diff
@@ -29,7 +29,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Breaking change** Change 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Breaking change** Support compression for Lucene data. Backwards compatibility is not supported, so rebuilding Lucene indexes with shutting down search requests handling is needed for transition. [(Issue #1466)](https://github.com/FoundationDB/fdb-record-layer/issues/1466)
```
alecgrieser, line 32:
```suggestion
* **Breaking change** Support compression for Lucene data. Backwards compatibility is not supported, so upgrading requires rebuilding Lucene indexes during which search requests cannot be completed. [(Issue #1466)](https://github.com/FoundationDB/fdb-record-layer/issues/1466)
```

The new wording may be slightly more natural, though it may be a matter of taste

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
alecgrieser: Yeah, that sounds right. We do want to have `KeyValueLogMessage` as our standard interface to logs everywhere, but that larger refactoring can definitely be a separate change from this one

alecgrieser (changes requested):
Looks good. Just a wording suggestion for the release note, though it also looks like some of the lucene tests are failing in the most recent PRB

alecgrieser (approved):

PR #1463, opened 2021-11-11
Resolve #1456 : Make QueryPlanResult and QueryPlanInfo immutable.

### LINKED ISSUE #1456: Make QueryPlanResult immutable
In order to prevent concurrency issues with the components of the QueryPlanResult, the class should be made immutable, where from a certain point, no changes could be made to it or to its components.

ohadzeliger:
make Make QueryPlanResult and QueryPlanInfo immutable:
- Create builder for QueryPlanInfo
- Remove setter methods from QueryPlanInfo
- Add constructor with PlanInfo to QueryPlanResult

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
@@ -35,8 +35,20 @@ import java.util.Set;
 public class QueryPlanInfo {
     private final Map<QueryPlanInfoKey<?>, Object> info;
 
-    public QueryPlanInfo() {
-        info = new HashMap<>();
+    public static final QueryPlanInfo EMPTY = new QueryPlanInfo(Collections.emptyMap());
+
+    private QueryPlanInfo(@Nonnull final Map<QueryPlanInfoKey<?>, Object> infoMap) {
+        info = infoMap;
+    }
+
+    @Nonnull
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    @Nonnull
+    public Builder toBuilder() {
+        return new Builder(this);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
```diff
@@ -141,4 +139,43 @@ public String toString() {
...
+         * @return this
+         */
+        @Nonnull
+        public <T> Builder put(@Nonnull QueryPlanInfoKey<T> key, @Nonnull T value) {
+            infoMap.put(key, value);
+            return this;
+        }
+
+        @Nullable
+        public <T> T get(@Nonnull QueryPlanInfoKey<T> key) {
```
normen662, line 173: Normally, builders don't have gets -- do you need this one?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanResult.java
```diff
@@ -39,7 +39,12 @@
 
     public QueryPlanResult(@Nonnull final RecordQueryPlan plan) {
         this.plan = plan;
-        planInfo = new QueryPlanInfo();
+        planInfo = QueryPlanInfo.EMPTY;
```
normen662, line 42: Usually, it's better to have a static method `empty()` on `QueryPlanInfo` to return `EMPTY` when called. This allows you to encapsulate whether `QueryPlanInfo` really should just return the same instance over and over or whether you create a different instance every time. It's a pattern I follow but that's just nitpicking.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfo.java
ohadzeliger: Yes, as it turns out, with the usage flow, we needed to log the contents of the PlanInfo before it is built, hence added a `get()`. It seemed to not interfere with immutability.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanResult.java
ohadzeliger: That makes sense. Done.

normen662 (approved):

PR #1465, opened 2021-11-15
Size metrics for lucene file reference write and count metrics for lu…

tian-yizuo:
…cene index write

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -143,7 +143,7 @@ public class FDBDirectory extends Directory {
      * @return current increment value
      */
     public synchronized long getIncrement() {
-        incrementCallCount(FDBStoreTimer.Counts.LUCENE_GET_INCREMENT_CALLS);
+        increment(FDBStoreTimer.Counts.LUCENE_GET_INCREMENT_CALLS);
         return context.ensureActive().get(sequenceSubspaceKey).thenApply(
             (value) -> {
                 if (value == null) {
@@ -157,10 +157,17 @@ public class FDBDirectory extends Directory {
             }).join();
     }
 
-    private void incrementCallCount(FDBStoreTimer.Counts counter) {
+    private void increment(FDBStoreTimer.Counts counter) {
         final FDBStoreTimer timer = context.getTimer();
         if (timer != null) {
             timer.increment(counter);
+        }
+    }
+
+    private void increment(FDBStoreTimer.Counts counter, int amount) {
+        final FDBStoreTimer timer = context.getTimer();
+        if (timer != null) {
+            timer.increment(counter, amount);
         }
     }
 
@@ -242,8 +249,10 @@ public class FDBDirectory extends Directory {
             }
         }
         LOGGER.trace("writeFDBLuceneFileReference {}", reference);
-        incrementCallCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE);
-        context.ensureActive().set(metaSubspace.pack(name), reference.getBytes());
+        final byte[] fileReferenceBytes = reference.getBytes();
+        increment(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_SIZE, fileReferenceBytes.length);
+        increment(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_CALL);
+        context.ensureActive().set(metaSubspace.pack(name), fileReferenceBytes);
         fileReferenceCache.put(name, reference);
     }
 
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
```diff
@@ -92,7 +92,8 @@ public void testWriteLuceneFileReference() throws Exception {
         CompletableFuture<FDBLuceneFileReference> luceneFileReference = directory.getFDBLuceneFileReference("test1");
         assertNotNull(luceneFileReference.get(5, TimeUnit.SECONDS), "fileReference should exist");
 
-        assertCorrectMetricCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE,2);
+        assertCorrectMetricCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_SIZE, reference1.getBytes().length + reference2.getBytes().length);
+        assertCorrectMetricCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_CALL,2);
```
MMcM, line 96:
```suggestion
        assertCorrectMetricCount(FDBStoreTimer.Counts.LUCENE_WRITE_FILE_REFERENCE_CALL, 2);
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -254,11 +263,9 @@ public void writeFDBLuceneFileReference(@Nonnull String name, @Nonnull FDBLucene
      * @param value the data to be stored
      */
     public void writeData(long id, int block, @Nonnull byte[] value) {
-        FDBStoreTimer timer = context.getTimer();
-        if (timer != null) {
-            //This may not be correct transactionally
-            timer.increment(FDBStoreTimer.Counts.LUCENE_WRITE_SIZE, value.length);
-        }
+        //This may not be correct transactionally
```
MMcM, line 266: I am not sure what potential problem this comment is referring to.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -164,6 +164,13 @@ private void incrementCallCount(FDBStoreTimer.Counts counter) {
         }
     }
 
+    private void incrementSize(FDBStoreTimer.Counts counter, int size) {
```
MMcM, line 167: Why have the more complicated contract that this checks whether counter is a size and does nothing? Why not just another overload of `increment` that takes an `amount`?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: I am also not really sure. I just keep the existing comment. Probably @scottfines may have better idea?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: My understanding is that we use the amount > 1 only for incrementing data size (although the underlying ```StoreTimer.increment(@Nonnull Count event, int amount)``` is not designed to be specific to data size). And we can see the method above has a specific name ```incrementCallCount```, which looks like specific for incrementing of calling count. If so, checker for ```counter.isSize()``` may be meaningful here. But I just looked at the similar code in ```InstrumentedReadTransaction``` and see it just simply have 2 overloads for ```increment```, one with amount argument and the other is to increment only 1. So I just changed the 2 methods here in same way. Please let me know if it looks good.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
tian-yizuo: Fixed.

MMcM (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
scottfines: it's not a big deal, I was referring to the possibility that if you rollback after incrementing this timer, the timer will technically be incorrect. This code doesn't do anything that affects the comment one way or the other.

scottfines (approved):

jleach4 (approved):

PR #1470, opened 2021-11-20
Resolves #1469: Support dynamic Lucene analyzer loading for index mai…

### LINKED ISSUE #1469: Support dynamic Lucene analyzer loading for index maintainer. And support Ngram for lucene indexing

tian-yizuo:
…ntainer. And support Ngram for lucene indexing

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerFactory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerFactory.java
@@ -1,0 +1,55 @@
+/*
+ * LuceneAnalyzerFactory.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import com.apple.foundationdb.record.metadata.Index;
+import org.apache.lucene.analysis.Analyzer;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Each implementation of {@link Analyzer} should have its own implementation
+ * of this factory interface to provide instances of the analyzer to a
+ * {@link LuceneAnalyzerRegistry}. The registry will populate a mapping of names to
+ * analyzers using the methods of this interface.
+ *
+ * @see LuceneIndexMaintainer
+ */
+public interface LuceneAnalyzerFactory {
+    /**
+     * Get the unique name for the text analyzer. This is the name that should
+     * be included in the index meta-data in order to indicate that this analyzer
+     * should be used within a certain index.
+     *
+     * @return the name of the analyzer that this factory creates
+     */
+    @Nonnull
+    String getName();
+
+    /**
+     * Get an instance of the text analyzer given the {@link Index}. For a given factory, each analyzer
+     * should be of the same type, and it should match the result of {@link #getName()}.
+     *
+     * @return an instance of the analyzer that this factory creates
+     */
+    @Nonnull
+    Analyzer getAnalyzer(@Nonnull Index index);
+}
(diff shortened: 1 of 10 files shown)


arnaud-splice (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Each implementation of {@link Analyzer} should have its own implementation
+ * of this factory interface to provide instances of the analyzer to a
+ * {@link LuceneAnalyzerRegistry}. The registry will populate a mapping of names to
+ * analyzers using the methods of this interface.
+ *
+ * @see TextIndexMaintainer
```
MMcM, line 35: Why this and not the `Lucene` index maintainer?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -0,0 +1,130 @@
...
+            LuceneAnalyzerFactory analyzerFactory = registry.get(name);
+            if (analyzerFactory == null) {
+                throw new MetaDataException("unrecognized lucene analyzer for tokenizer", LogMessageKeys.ANALYZER_NAME, name);
+            }
+            return analyzerFactory.getAnalyzer(index);
+        }
+    }
+
+    // Synchronize this method so that we don't need a ConcurrentHashMap but so that
+    // it is still thread safe.
```
MMcM, line 110: What prevents this method growing/rehashing the hash map while someone is trying to access the result of `getRegistry`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    @SuppressWarnings("PMD.CompareObjectsWithEquals")
+    public synchronized void register(@Nonnull LuceneAnalyzerFactory analyzerFactory) {
+        final String name = analyzerFactory.getName();
+        LuceneAnalyzerFactory oldFactory = registry.putIfAbsent(name, analyzerFactory);
+        // If there was a factory already registered and the old factory isn't the same as
+        // the new one, throw an error
+        if (oldFactory != null && oldFactory != analyzerFactory) {
```
MMcM, line 118: Maybe use `.equals`? I can't actually think of a legit reason to register the same factory twice, but if there were one, I could well imagine that it would include registering an "equivalent" one. Of course, you'd have to override `equals` in that factory class for that to work.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerFactory.java
tian-yizuo: Good catch! Because I just copied the Javadoc from ```TextTokenizerFactory``` and edited it, but missed this line. Fixed it now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
tian-yizuo: Fixed

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
tian-yizuo: I just used same abstraction here as ```TextTokenizerRegistry```. In fact I am not really sure in what cases ```getRegistry()``` could be needed. I am guessing probably ```getRegistry()``` and ```register``` are not supposed to be called concurrently? @alecgrieser do you have any ideas about the design of ```TextTokenizerRegistry```? If so, any thoughts about the potential use cases of ```getRegistry()``` and the dynamic ```register```?

alecgrieser: It's been a bit since I've looked at this code, so my understanding is a bit rusty, but I believe the intention here was primarily for `register` to be used during tests (and that's, in fact, the only place I can find it being used). That way test-only tokenizers could be created and swapped in during tests with different properties to test, for example, certain error scenarios. However, it also offers an alternative for people who don't want to set up the `ServiceLoader` correctly, I guess, but that should really only be done in some kind of static initialization block, not while someone is trying to access the registry. 

It does seem like there's a concurrency bug here (i.e., in both registries)  if `register` is used outside of the above contexts. If you don't need to use `register` (and can instead rely on only using the `ServiceLoader`), then that's one way of getting around the problem. Otherwise, we either need to document or fix the concurrency bug, I think

tian-yizuo: @alecgrieser thank you! I just removed the ```register```, ```getRegistry``` and ```reset``` to get rid of the support for registration in fly, considering it is not actively used for now. Please let me know this looks good to you. @MMcM @alecgrieser

MMcM (approved):

PR #1481, opened 2021-12-06
Resolves #1471: Selector and Comparator plans

### LINKED ISSUE #1471: Implement Chooser plans: Selector and Comparator
Implement plans to select one from multiple alternatives and to execute multiple p[lans and compare their results.

ohadzeliger:
This PR introduces two new query plans:
1. Selector plan selects one sub-plan to be executed from its children
2. Comparator plan executes all children plans and compares their results


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -0,0 +1,343 @@
...
+
+        private void logTimer(final long startTime) {
+            if (timer != null) {
+                timer.record(FDBStoreTimer.Events.QUERY_SELECTOR, System.nanoTime() - startTime);
+            }
+        }
+    }
+
+    public interface PlanSelector extends PlanHashable {
+        int selectPlan(List<RecordQueryPlan> plans);
```
alecgrieser, line 278: minor: It's a bit six of one, half dozen of another, but part of me wonders if this should return a `RecordQueryPlan` instead of the position into the list. Just because some implementations (e.g., ones that iterate over the plans before picking the "best" one, based on some criterion) might prefer that interface.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -0,0 +1,343 @@
...
+            if (innerContinuation != null) {
+                this.innerContinuation = ByteString.copyFrom(innerContinuation);
+            } else {
+                this.innerContinuation = null;
+            }
+        }
+
+        @Nullable
+        @Override
+        public byte[] toBytes() {
```
alecgrieser, line 316: Is this the right continuation if `isEnd = true` (I thought that should return `null`)?

We're also not always consistent about this, but in some cases, we use `RecordCursorEndContinuation.END` for end continuations, and then the other continuation types always return `false` for `isEnd`, in which case this wouldn't need to handle the `isEnd = true` case.

alecgrieser (changes requested):
I haven't gone through the test code (yet), but I figured you'd want the feedback on the non-test code sooner rather than later, so I'm submitting that now, separate from the rest of review. Sorry for the delay in getting to this PR

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
```diff
@@ -0,0 +1,314 @@
...
+                .setFilter(Query.field("num_value_2").equalsValue(1))
+                .build();
+        RecordQuery query2 = RecordQuery.newBuilder()
+                .setRecordType("MySimpleRecord")
+                .setFilter(Query.field("num_value_2").equalsValue(2))
+                .build();
+
+        RecordQueryPlan planUnderTest = RecordQueryComparatorPlan.from(plan(query1, query2), primaryKey(), 0);
+
+        int count = querySimpleRecordStore(NO_HOOK, planUnderTest, EvaluationContext::empty,
```
alecgrieser, line 110: I take it that there should be logs of a discrepancy between the two plans, though it's a little hard to see that without either an explicit assertion on the logs (which would require being able to hook into the log system) or some other way of inspecting the cursor for errors detected. I guess something to think about

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
```diff
@@ -0,0 +1,314 @@
...
+
+    @DualPlannerTest
+    void testTwoDifferentInnerPlansOneEndsSooner() throws Exception {
+        complexQuerySetup(NO_HOOK);
+
+        RecordQuery query1 = RecordQuery.newBuilder()
+                .setRecordType("MySimpleRecord")
+                .setFilter(Query.field("num_value_2").equalsValue(1))
+                .build();
+        // This plan has same records as the previous one, but end sooner
```
alecgrieser, line 145:
```suggestion
        // This plan has same records as the previous one, but ends sooner
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
```diff
@@ -0,0 +1,314 @@
...
+                record -> assertThat(record.getNumValue2(), is(1)),
+                context -> assertDiscardedAtMost(10, context));
+        assertEquals(2, count);
+    }
+
+    @DualPlannerTest
+    void testNoInnerPlansFail() throws Exception {
+        complexQuerySetup(NO_HOOK);
+
+        assertThrows(VerifyException.class, () -> RecordQueryComparatorPlan.from(Collections.emptyList(), primaryKey(), 0));
```
alecgrieser, line 190: minor: there's part of me that wonders if this should be a `RecordCoreArgumentException`, partially to make sure that as many errors as possible are in the `RecordCoreException` hierarchy, but I suppose it's a minor thing

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    @DualPlannerTest
+    void testTwoSameInnerPlansWithContinuation() throws Throwable {
```
alecgrieser, line 194: It would probably be good to have some versions of this that failed with out-of-band limits. I think the rows scanned limit should be deterministic enough that if it were used, it wouldn't cause the resulting tests to be flaky

alecgrieser (changes requested):
Okay, and now looked at the tests

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param planHashSupplier function to return the cursor plan's string (for logging purposes)
+     *
+     * @return a cursor containing same records in all child cursors
+     *
+     * @see #create(Function, List, byte[], FDBStoreTimer, int, Supplier, Supplier)
+     */
+    @Nonnull
+    public static <M extends Message, S extends FDBRecord<M>> ComparatorCursor<S> create(
+            @Nonnull FDBRecordStoreBase<M> store,
+            @Nonnull KeyExpression comparisonKey,
```
normen662, line 88: Can you `Verify` that we are not handing comparison keys that use nested repeateds -- look for `createdDuplicates()` on `KeyExpression`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorContinuation.java
```diff
@@ -0,0 +1,145 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+
+class ComparatorCursorContinuation extends MergeCursorContinuation<RecordCursorProto.ComparatorContinuation.Builder, RecordCursorContinuation> {
```
normen662, line 41: What is the semantics for different advances among the plans? Reference plan advances by 5 records, challenger advances by 10: What do we do now? I would assumer to return the 5 records from the reference and _reset_ the other continuation to that record as well? Not really sure how this works.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryChooserPlanBase.java
```diff
@@ -0,0 +1,135 @@
...
+    protected RecordQueryChooserPlanBase(@Nonnull final List<Quantifier.Physical> quantifiers) {
+        Verify.verify(!quantifiers.isEmpty());
+        this.quantifiers = List.copyOf(quantifiers);
+        boolean firstReverse = quantifiers.get(0).getRangesOverPlan().isReverse();
+        if (!getChildStream().allMatch(child -> child.isReverse() == firstReverse)) {
+            throw new RecordCoreArgumentException("children of chooser plan do not all have same value for reverse field");
+        }
+        this.reverse = firstReverse;
+        // Since all child plans are assumed to return the same Values, pick the first one to flow the values back.
+        this.resultValuesSupplier = Suppliers.memoize(quantifiers.get(0)::getFlowedValues);
```
normen662, line 64: This is dangerous. The planner should inject a `PickValuesValue` here. I think I did this for Norse but essentially you create values as results here that look like this
```
PickValuesValue(ordinalQuantifiedColumn, all the zeroeth columns),
PickValuesValue(ordinalQuantifiedColumn, all the first columns),
PickValuesValue(ordinalQuantifiedColumn, all the second columns),
...
PickValuesValue(ordinalQuantifiedColumn, all the nth columns)
```

`PickValuesValue()` has `n+1` inputs, i.e. an integer (a selector) and n options that can be picked from. At runtime you set the ordinal to the option you want and then `eval` the value which gives you the correct datum.

All of this is necessary since we don't ever want to disturb the derivation graph of a value. If you do it the way you do it, transformations would only ever _see_ the `0` variant and are oblivious to others which we want to avoid.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Verify;
+import com.google.common.collect.ImmutableList;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Objects;
+import java.util.Random;
+
+public class RelativePriorityPlanSelector implements RecordQuerySelectorPlan.PlanSelector {
```
normen662, line 34: Not a fan of having top-level classes extend inner classes or implement inner interfaces. Can you pull the nested `PlanSelector` out?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
```diff
@@ -0,0 +1,93 @@
...
+        this(priorities, new Random());
+    }
+
+    @VisibleForTesting
+    RelativePriorityPlanSelector(@Nonnull final List<Double> priorities, @Nonnull final Random random) {
+        Verify.verify(!priorities.isEmpty());
+        this.priorities = ImmutableList.copyOf(priorities);
+        this.random = random;
+        double sumAll = this.priorities.stream().mapToDouble(Double::doubleValue).sum();
+        Verify.verify(sumAll == 1.0D);
```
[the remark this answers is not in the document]
normen662: 1. use `BigDecimal` -- that may be overkill
2. use rational numbers, i.e. use a base of some number, e.g. 100 and then use dividends thereof
3. or we just change this to `sumAll <= 1.0D` since you are summing up floats that may have been truncated already your sum should only err on the small side me thinks. If you really care about a gap between your sum and `1.0D` you can have special logic pick the last plan if you ever fall into that range.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public RelativePriorityPlanSelector(@Nonnull final List<Double> priorities) {
+        this(priorities, new Random());
+    }
+
+    @VisibleForTesting
+    RelativePriorityPlanSelector(@Nonnull final List<Double> priorities, @Nonnull final Random random) {
+        Verify.verify(!priorities.isEmpty());
+        this.priorities = ImmutableList.copyOf(priorities);
+        this.random = random;
```
normen662, line 49: `Random` performance is pretty bad especially if there is contention between threads. Consider using `ThreadLocalRandom` here even though it forsakes stability in multithreaded environments.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
```diff
@@ -0,0 +1,227 @@
...
+     */
+
+    @Nonnull
+    @Override
+    @SuppressWarnings("squid:S2095")
+    public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final FDBRecordStoreBase<M> store,
+                                                                     @Nonnull final EvaluationContext context,
+                                                                     @Nullable final byte[] continuation,
+                                                                     @Nonnull final ExecuteProperties executeProperties) {
+        final ExecuteProperties childExecuteProperties = executeProperties.clearSkipAndLimit();
```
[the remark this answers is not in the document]
ohadzeliger: The intent here is to have the limit on the overarching plan rather than on the children. The child plans should return the same results, but we can't assume that they will all follow the same access patterns (e.g. they may scan different indexes). So the idea was to limit the scan by the parent plan (such that the limit takes all children's scans into account).
Similar pattern was used by `RecordQueryIntersectionPlan` - please let me know if this is not the right way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
```diff
@@ -0,0 +1,243 @@
...
+        // Wait for all cursors to have a valid state
+        return whenAll(cursorStates).thenApply(vignore -> {
+            // if all cursors have a value, compare and continue
+            if (cursorStates.stream().allMatch(cursorState -> hasNext(cursorState))) {
+                compareAllStates(cursorStates);
+                // Regardless of comparison result, return all states (comparison logged results)
+                return cursorStates;
+            }
+            // some cursors have no next result, if all are exhausted, we are done
+            long exhaustedCount = cursorStates.stream().filter(cursorState -> isSourceExhausted(cursorState)).count();
```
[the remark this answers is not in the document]
ohadzeliger: I may be missing something, but the intent was to allow cursors to "independently" progress (using whatever pace they have) and to make sure that when they all return a result, that result gets compared. For that, we should allow cursors to hit their scan limit and continue later (using a continuation), and the only "real" stopage is the cursor reaching exhaustion.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
```diff
@@ -0,0 +1,227 @@
...
+        return ComparatorCursor.create(store, getComparisonKey(),
+                        getChildren().stream()
+                                .map(childPlan -> comparatorCursorFunction(store, context, childExecuteProperties, childPlan))
+                                .collect(Collectors.toList()),
+                        continuation,
+                        referencePlanIndex,
+                        () -> toString(),
+                        () -> planHash(PlanHashKind.STRUCTURAL_WITHOUT_LITERALS))
```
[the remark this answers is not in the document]
alecgrieser: I actually wrote this before I confirmed that the plan and plan hash is actually only logged once, in which case memoizing is probably not necessary

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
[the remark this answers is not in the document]
alecgrieser: I think the difference is that for intersection plans, not clearing the skip and limit is semantically incorrect (i.e., if you had a limit and/or skip you wanted to enforce, if you enforced it on each child individually, because you don't know which values will or won't be in the intersection until you've read them from the children, you might have to read more data from each child before the limit and/or skip is actually correctly satisfied).

For this plan, the idea is that each of the candidates should be identical in what they return (regardless of what data they access), in which case enforcing a limit should be accomplishable by enforcing the limit on each child individually. Maybe we still want to enforce the limit on the comparator instead, though, I'm not sure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
alecgrieser: In theory, if we cared about using a `ThreadLocalRandom` in multiple threads, we could make the `random` field of the class `@Nullable`, and then use `ThreadLocalRandom.current()` if `this.random == null`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: There are ways to do this (e.g. https://stackoverflow.com/questions/3717402/how-can-i-test-with-junit-that-a-warning-was-logged-with-log4j), and there are also OSS that allows you to assert the contents of logs.
I would leave that for some later phase since we don't really do that anywhere else.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: I was following patterns I've seen around (i.e. using `Verify.verify` for parameter verification. I don't mind changing them to `RecordCoreArgumentException` if that's the way to go.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -191,6 +191,10 @@
         QUERY_DISTINCT("compare query records for distinct"),
         /** The amount of time spent in {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedPrimaryKeyDistinctPlan} as part of executing a query. */
         QUERY_PK_DISTINCT("compare record primary key for distinct"),
+        /** The amount of time spent in {@link com.apple.foundationdb.record.query.plan.plans.RecordQuerySelectorPlan} as part of executing a query. */
+        QUERY_SELECTOR("execute one iteration of selected plan"),
```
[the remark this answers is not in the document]
ohadzeliger: `QUERY_SELECTOR` and `QUERY_COMPARATOR` are `Events` that track the amount of time spent in the respective query execution, whereas `PLAN_COMPARATOR` and `PLAN_SELECTOR` are `Counts` that count the number of plans of this type that were created.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -600,6 +604,16 @@ public String logKey() {
         QUERY_AGGREGATE_GROUPS("number of aggregate groups", false),
         /** The max size of aggregate group created by {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryStreamingAggregatePlan}. */
         QUERY_AGGREGATE_GROUP_MAX_SIZE("max size of aggregate group", false),
+        /** The number of query plans that include a {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryComparatorPlan}. */
+        PLAN_COMPARATOR("number of comparator plans", false),
+        /** The number of query plans that include a {@link com.apple.foundationdb.record.query.plan.plans.RecordQuerySelectorPlan}. */
+        PLAN_SELECTOR("number of selector plans", false),
```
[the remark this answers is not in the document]
ohadzeliger: This is incremented by 1 for the `ComparatorPlan` (similar to the way it is done in `RecordQueryIntersectionPlan`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorContinuation.java
ohadzeliger: Since the goal was to log an alert if the plans do not return the same records, I took the approach of allowing the cursors to advance in unison  (meaning that it consumes one record from each when they all have a `next` result). Of course, this means that if a contender "misses" a result, it will never get back in sync with the reference cursor, but from an overall goal perspective, this is OK since the comparison already failed.
It would be more complicated to try and "resync" the cursors after a comparison failure, and I believe it will not add much value, since we would have to resolve the initial comparison failure anyways.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
ohadzeliger: True, but that would also force the `PlanSelector` to store the list of plans, which it does not have to do now (since the selection is pure mathematical operation). I'd rather defer this change until it is proven necessary.

alecgrieser: Couldn't it get it from the `plans` list that are passed in as a parameter?

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -108,3 +108,16 @@ message LuceneIndexContinuation {
     required float score = 2;
     optional uint64 shard = 3;
 }
+
+message SelectorPlanContinuation {
+    required uint64 selectedPlan = 1; // The selected plan executing from the l;ist of optional plans
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -108,3 +108,16 @@ message LuceneIndexContinuation {
...
+    optional bytes innerContinuation = 2; // The selected plan's continuation
+}
+
+message ComparatorContinuation {
+    message CursorState {
+        optional bytes continuation = 1;
+        optional bool started = 2;
+    }
+    repeated CursorState other_child_state = 1;
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,142 @@
...
+ * For each record returned by the sub-cursors, this cursor will ensure that they are the same (by using the comparison
+ * function). The cursor will log and continue if the records are not the same or if one of the sub-cursors ends
+ * prematurely.
+ * The cursor defines a "reference cursor" whose results are assumed to be correct. Values are compared to this
+ * cursor's results, and the reference cursor's results are the ones that are actually returned.
+ *
+ * @param <T> the type of elements returned by the cursor
+ */
+@API(API.Status.EXPERIMENTAL)
+public class ComparatorCursor<T> extends ComparatorCursorBase<T, T> {
```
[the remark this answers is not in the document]
ohadzeliger: You're right. I was using the same structure as the `IntersectionCursor` assuming that we will have more flavors of comparison cursors in the future (e.g. one that maybe fails on mismatch or one that has different comparison strategies). 
For now though, it makes more sense to only have one class with no hierarchy.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Execute children and compare results.
+     * @param store record store from which to fetch records
+     * @param context evaluation context containing parameter bindings
+     * @param continuation continuation from a previous execution of this same plan
+     * @param executeProperties limits on execution
+     * @param <M> the type of records in the store
+     * @return {@link RecordCursor} that iterates through the results of the execution
+     */
+
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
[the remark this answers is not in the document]
ohadzeliger: Agree. Changed the skip and limit to be enforced by the child plan and cleared it from the parent plan - this way we can verify that each child is applying it correctly. Documented the logic.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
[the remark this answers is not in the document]
ohadzeliger: You're right - these will be used to log information when the comparison fails. In reality, though, the `errorLogged` field protects us from issuing log messages for any but the first comparison failure per instance (the idea being that following the first comparison failure we will have many, many, more).
It is good defensive strategy, though, to memoize them (and has no downside). Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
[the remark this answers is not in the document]
ohadzeliger: Hmmm. This is something that we should solve - though I don't know how. Let's discuss offline?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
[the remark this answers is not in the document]
ohadzeliger: As part of the logic change for the out-of-band limits, added more logic about in-band limit as well. Seems like if we have any combination of exhausted plans and in-band-limit hit then we have a "early termination error" (since the reference plan and at least one other plan have to be different). Please take a look

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
```diff
@@ -0,0 +1,243 @@
...
+    protected CompletableFuture<List<KeyedMergeCursorState<T>>> computeNextResultStates() {
+        final List<KeyedMergeCursorState<T>> cursorStates = getCursorStates();
+        // Wait for all cursors to have a valid state
+        return whenAll(cursorStates).thenApply(vignore -> {
+            // if all cursors have a value, compare and continue
+            if (cursorStates.stream().allMatch(cursorState -> hasNext(cursorState))) {
+                compareAllStates(cursorStates);
+                // Regardless of comparison result, return all states (comparison logged results)
+                return cursorStates;
+            }
```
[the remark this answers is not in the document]
ohadzeliger: Added a check for at least one out-of-band limit. Also, changed the no-next-result to the "strongest" which would give preference to the out-of-band status.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
```diff
@@ -0,0 +1,243 @@
...
+     * {@link #computeNextResultStates} method has finished and the cursor has stopped. This can be because all
+     * sub-cursors are exhausted, or otherwise they all have non-exhaustion reason. Calculate the weakest reason
+     * and return it so that continuation can pick up from that point.
+     *
+     * @return the weakest reason for stopping
+     */
+    @Override
+    @Nonnull
+    protected NoNextReason mergeNoNextReasons() {
+        return getWeakestNoNextReason(getCursorStates());
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Override
+    @Nonnull
+    public ComparatorCursorContinuation getContinuationObject() {
+        return ComparatorCursorContinuation.from(this);
```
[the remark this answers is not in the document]
ohadzeliger: Agree - let's discuss offline?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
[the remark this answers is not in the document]
ohadzeliger: For simplicity, used %-based integer solution. This way, the user will specify a list of int values all adding up to 100.
The reason I went with 1-based doubles before was to allow "unlimited" number of sub-plans that offers better flexibility, but this is really not necessary in this case, where we will have a handful (probably no more than 2 or 3) plans that we would have to choose from, and the idea was to just sanity check that their priorities are reasonable. Since this caused some concerns, we should just opt for a simpler solution.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
ohadzeliger: Yes, it could, but then, we also need the index of the plan (so that we can store it in the continuation), and so would have to create a structure that has both plan and index, and the code would get more complex...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
ohadzeliger: It seems clearer logic to delegate continuation state (including `isEnd()`) to the inner continuation, so just added `null` for the `isEnd` case in the selector continuation.
Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
ohadzeliger: Done. random.nextInt() is called once per query (since it will use continuation value after that). Using ThreadLocalRandom.current.nextInt() initializes it once per thread, which is the requests thread and so the same random instance will be used for all plans for this query.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -0,0 +1,343 @@
...
+    @Nonnull
+    @Override
+    public RelationalExpressionWithChildren rebaseWithRebasedQuantifiers(@Nonnull final AliasMap translationMap, @Nonnull final List<Quantifier> rebasedQuantifiers) {
+        return new RecordQuerySelectorPlan(
+                Quantifiers.narrow(Quantifier.Physical.class, rebasedQuantifiers), planSelector);
+    }
+
+    private static void validateParams(@Nonnull final List<? extends RecordQueryPlan> plans, @Nonnull final List<Double> relativePlanPriorities) {
+        Verify.verify(!plans.isEmpty());
+        Verify.verify(!relativePlanPriorities.isEmpty());
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
ohadzeliger: Sure. Just to make sure I understand: Is the problem here that the key itself can evaluate to repeated field?
You also mentioned that in cases sub-plans have predicates that operate on repeated fields then we can have duplicate results (and they can differ in their distinct-ness), and that in this case comparison may fail, but this is a different issue, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorContinuation.java
ohadzeliger: Also: All the plans advance at the same rate - one record at a time - and the iteration stops when one of them reaches a limit. If this limit is "allowed" (e.g. scan limit or time limit), then wee stop the iteration and return a continuation. If the limit is semantically wrong (w/comparison to the reference plan - e.g. record limit or exhaustion) - then we log an error.
In all cases, the reference plan determines the record returned, and the termination status (e.g. exhaustion or record limit).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: Done.

ohadzeliger:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
```diff
@@ -232,6 +232,7 @@
 
     // query plan
     PLAN("plan"),
+    PLAN_HASH("plan_hash"),
```
alecgrieser, line 235: As a heads up, be on the look out for conflicts with #1484

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,346 @@
...
+
+    private ComparatorCursor(@Nonnull List<KeyedMergeCursorState<T>> cursorStates,
+                             @Nullable FDBStoreTimer timer,
+                             final int referencePlanIndex,
+                             @Nonnull final Supplier<String> planStringSupplier,
+                             @Nonnull final Supplier<Integer> planHashSupplier) {
+        super(cursorStates, timer);
+        this.referencePlanIndex = referencePlanIndex;
+        this.planStringSupplier = Suppliers.memoize(planStringSupplier::get);
+        this.planHashSupplier = Suppliers.memoize(planHashSupplier::get);
```
alecgrieser, line 86: I *think* the `::get` here should be extraneous because the arguments are already functions (or `Suppliers`), so they don't need to be lambda-ized

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorBase.java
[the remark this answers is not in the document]
alecgrieser: Agreed in a separate conversation to defer

alecgrieser (changes requested):
One minor change regarding some `Supplier` usage, but looks good to me

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
ohadzeliger: Ack

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
ohadzeliger: Yeah, that was my gut feel too, but then:
`reason: no instance(s) of type variable(s) T exist so that Supplier<String> conforms to Supplier<T>`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryChooserPlanBase.java
ohadzeliger: Done. Created `ValuePickerValue` that stores multiple sub-values and delegates to the `selected` one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
alecgrieser: oof, that's unfortunate. I wonder if this is a bug in the compiler, or something, because it seems like it should conform... Alas

alecgrieser (approved):
We should probably squash and merge (rather than make a merge commit) when this PR is approved, but looks good!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
normen662: I don't think the error message you paste applies to the the code here. What `T` is it talking about?

Out of interest I changed this to

```
this.planStringSupplier = planStringSupplier;
this.planHashSupplier = planHashSupplier;
```

and it builds fine. Please reconsider!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
normen662: If `createsDuplicates()` returns `true` a particular index may actually return multiple key, value pairs for a document (which may flow out of order -- i.e. not next to each other). I just want us to not extend this logic to that sort of complication just yet. The compared correctness would depend on the index chosen which seems to be wrong. In the future we may relax this to an execution-time check, if we really that there is only ever one duplicate. That is the case for maps.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorContinuation.java
normen662: What I meant is that we have seen plans hit the fetch limit. Imagine plan A being an index plan and plan B being a regular primary scan. In that case, one plan (the regular scan) may hit the scan limit by itself while the other one does not. What I think you are writing is that the reference plan decides how we advance and we return the continuations for both sides accordingly.

Related to this: I think the two plans now shared the scan limit, we should be cognizant of that fact.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
ohadzeliger: Agree. My question is what do you propose we do to avoid that. The situation currently is that the query execution will fail since there is a validation inside the `MergeCursor` to check that the number of results of applying the `comparisonKey` has exactly one result.
What I recommend we do is, in a future PR, change the `MergeCursor` to allow multiple records per key and be able to return them correctly. For now, we can either:
1. Fail the creation of the plan
2. Allow the plan to be created but do not create a comparison plan
I don't like option 2 since it will hide the problem and at runtime it will look like the comparison ran without failure.
If you agree, I can implement 1 from above for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursorContinuation.java
ohadzeliger: I believe this is handled correctly now: After the changes to the cursor, the returned result is the "strongest" of the results of the sub-plans, which means that the inefficient plan may spend more of the scan budget, and which will cause an "put-of-band" limit to be hit (either for itself or for the other plan). In that case the wrapping cursor will return that same out-of-band error and cause the plan execution to stop and return a continuation. This means that when this limit is hit, the scanning will continue until there is a result returned from all sub-cursors, at which point the comparison will take place.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
ohadzeliger: Done.

normen662 (dismissed):
From my side ok -- as long as Alec and Mike are ok.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
```diff
@@ -0,0 +1,233 @@
...
+                        () -> planHash(PlanHashKind.STRUCTURAL_WITHOUT_LITERALS))
+                .skipThenLimit(parentExecuteProperties.getSkip(), parentExecuteProperties.getReturnedRowLimit())
+                .map(QueryResult::of);
+    }
+
+    /*
+     * Return a function that creates a cursor for the given child plan using the provided continuation
+     */
+    @Nonnull
+    private <M extends Message> Function<byte[], RecordCursor<FDBQueriedRecord<M>>> comparatorCursorFunction(
```
MMcM, line 142: `comparatorCursorFunction` doesn't seem quite right. I guess `comparandCursorFunction` would work. Or `cursorFunctionForComparator` or something like that.

Or take a different approach like `childCursorFunction`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.concurrent.Executor;
+import java.util.stream.Collectors;
+
+/**
+ * A {@link RecordQueryChooserPlanBase} that selects one of its children to be executed.
+ * This plan can be used to allow the planner to have a few optional plans to be executed where the decision on which
+ * one actually gets run is deferred to execution time.
+ * The selector can be used with a given lambda that selects among the plans, and as a convenience, a relative priority
+ * selector is also provided. The relative priority selector takes a list of priorities (each associated with a plan)
+ * and selects a plan at random with relative priority as given by the priority of the plan.
```
MMcM, line 70: I guess I'm not dead-set against "priority" as used here. But I would go with "probability," because of the (incorrect here) assumption that higher priority _always_ wins.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * RelativePriority.java
```
MMcM, line 2:
```suggestion
 * RelativePriorityPlanSelector.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -0,0 +1,340 @@
...
+        super(quantifiers);
+        this.planSelector = planSelector;
+    }
+
+    /**
+     * Factory method that takes in a list of relative priorities and created a selection policy based on the priorities.
+     *
+     * @param children the list of subplans
+     * @param relativePlanPriorities a list (of the same length as the children) that determines the relative
+     * priority for selecting the plan. Sum of all the priorities must be 1.0
```
MMcM, line 89: This seems more reasonable than the actual contract here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * A utility class to implement an algorithm to make a random selection based off of relative priorities.
+ * The priorities are given in the list of priorities (to the constructor) and the selector chooses a number (between
+ * 0 and the priorities.size - 1), where the chance for a given selection to be picked being relative to its value.
+ * The relative priorities are given as % points: they should all add up to 100.
```
MMcM, line 38: Why not use fractions (doubles), as an earlier comment suggests?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/predicates/ValuePickerValue.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * QueriedValue.java
```
MMcM, line 2:
```suggestion
 * ValuePickerValue.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/predicates/ValuePickerValue.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * A value representing multiple "alternative" values.
+ * This value holds multiple "alternatives" for values and allows the creator to select one of them to be returned as
+ * the "actual" value.
+ * This is useful in cases (like {@link com.apple.foundationdb.record.query.plan.plans.RecordQuerySelectorPlan}) where
+ * there can be one of several plans that can be selected for execution, and this values gives visibility to all
+ * of their {@link Value}s
+ */
+@API(API.Status.EXPERIMENTAL)
+public class ValuePickerValue implements Value.CompileTimeValue {
```
MMcM, line 43: Why? That is, how does deferring the selection of the value to use until such use help? Why not just remember the chosen value at the place that would otherwise call one of these constructors?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBSelectorPlanTest.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * SelectorPlanTest.java
```
MMcM, line 2:
```suggestion
 * FDBSelectorPlanTest.java
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
```diff
@@ -0,0 +1,418 @@
+/*
+ * SelectorPlanTest.java
```
MMcM, line 2:
```suggestion
 * FDBComparatorPlanTest.java
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelectorTest.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * RelativePriorityTest.java
```
MMcM, line 2:
```suggestion
 * RelativePriorityPlanSelectorTest.java
```

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,346 @@
...
+            @Nonnull final Supplier<Integer> planHashSupplier) {
+        return create(
+                (S rec) -> comparisonKey.evaluateSingleton(rec).toTupleAppropriateList(),
+                cursorFunctions, continuation, store.getTimer(),
+                referencePlanIndex, planStringSupplier, planHashSupplier);
+    }
+
+    /**
+     * Create a comparator cursor from several compatibly-ordered cursors.
+     * The returned cursor will the reference results for the reference cursor, preserving order.
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,346 @@
...
+     * @param <T> the type of elements returned by this cursor
+     * @param comparisonKeyFunction the function evaluated to compare elements from different cursors
+     * @param cursorFunctions a list of functions to produce {@link RecordCursor}s from a continuation
+     * @param continuation any continuation from a previous scan
+     * @param timer the timer used to instrument events
+     * @param referencePlanIndex the index of the reference cursor in the given list of sub-cursors
+     * @param planStringSupplier function to return the cursor plan's hash (for logging purposes)
+     * @param planHashSupplier function to return the cursor plan's string (for logging purposes)
+     *
+     * @return a cursor containing all records in all child cursors
```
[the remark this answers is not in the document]
ohadzeliger: Done. Corrected Comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,346 @@
...
+                referencePlanIndex, planStringSupplier, planHashSupplier);
+    }
+
+    /**
+     * Calculate and return the reason the cursor returns a no-next result. At this point, it is assumed that the
+     * {@link #computeNextResultStates} method has finished and the cursor has stopped. This can be because all
+     * sub-cursors are exhausted, or otherwise they all have non-exhaustion reason. Calculate the weakest reason
+     * and return it so that continuation can pick up from that point.
+     *
+     * @return the weakest reason for stopping
```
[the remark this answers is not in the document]
ohadzeliger: Right. Changed comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/ComparatorCursor.java
```diff
@@ -0,0 +1,346 @@
...
+            @Nonnull final Supplier<String> planStringSupplier,
+            @Nonnull final Supplier<Integer> planHashSupplier) {
+        return new ComparatorCursor<>(createCursorStates(cursorFunctions, continuation, comparisonKeyFunction, referencePlanIndex), timer,
+                referencePlanIndex, planStringSupplier, planHashSupplier);
+    }
+
+    /**
+     * Calculate and return the reason the cursor returns a no-next result. At this point, it is assumed that the
+     * {@link #computeNextResultStates} method has finished and the cursor has stopped. This can be because all
+     * sub-cursors are exhausted, or otherwise they all have non-exhaustion reason. Calculate the weakest reason
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryChooserPlanBase.java
```diff
@@ -0,0 +1,168 @@
...
+
+    protected RecordQueryChooserPlanBase(@Nonnull final List<Quantifier.Physical> quantifiers) {
+        Verify.verify(!quantifiers.isEmpty());
+        this.quantifiers = List.copyOf(quantifiers);
+        boolean firstReverse = quantifiers.get(0).getRangesOverPlan().isReverse();
+        if (!getChildStream().allMatch(child -> child.isReverse() == firstReverse)) {
+            throw new RecordCoreArgumentException("children of chooser plan do not all have same value for reverse field");
+        }
+        this.reverse = firstReverse;
+        // Since all child plans are assumed to return the same Values, pick the first one to flow the values back.
```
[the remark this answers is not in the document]
ohadzeliger: Right. This comment was stale - updated to reflect new behavior.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryChooserPlanBase.java
```diff
@@ -0,0 +1,168 @@
+/*
+ * RecordQueryChooserPlan.java
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryComparatorPlan.java
ohadzeliger: Done. childCursorFunction seems to capture the meaning best.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
ohadzeliger: Done. Changed to `probability` across the board.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
ohadzeliger: Assuming the issue here is the semantics of the relative probabilities: sum of 1.0 vs. sum of 100%: This was discussed in previous conversations and modified to be % based (see below comment).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RelativePriorityPlanSelector.java
ohadzeliger: That was the original implementation, but due to PR comments (the main concern was with rounding issues with doubles), this seemed like a more acceptable solution.
The alternative would have been an unbound list of numbers (either fixed or floating point) whose sum would have been the "100%". The issue I had with this approach is that it would have been harder for a user to figure out the exact relative probabilities (esp. after a while, where the numbers may fluctuate and the total become arbitrary).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/predicates/ValuePickerValue.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/predicates/ValuePickerValue.java
ohadzeliger: This is per the design by @normen662 (similar to the type system design) where the entire `Value` population for the plan is kept around for the cases where it is needed. This way, the actual executed plan makes its runtime `values` available, but exploration of the possible values is still available.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBSelectorPlanTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/FDBComparatorPlanTest.java
ohadzeliger: Done.

ohadzeliger:
test this please

ohadzeliger:
Hi @MMcM , I've responded to all of your comments. Can you please take a look?

MMcM (approved):

PR #1485, opened 2021-12-08
Resolves #1486: LuceneIndexMaintainer should have 2 separate analyzers for indexing time and query time

### LINKED ISSUE #1486: LuceneIndexMaintainer should have 2 separate analyzers for indexing time and query time

tian-yizuo:
…query, and use standard analyzer on query time for ngram index

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -14,7 +14,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Use 2 separate analyzers for indexing time and query time in LuceneIndexMaintainer [(Issue #1486)](https://github.com/FoundationDB/fdb-record-layer/issues/1486)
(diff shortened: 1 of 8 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/Pair.java
```diff
@@ -0,0 +1,55 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Utility class for a pair of objects with different types.
+ *
+ * @param <K> The type of left object
+ * @param <V> The type of right object
+ */
+public final class Pair<K, V> {
```
MMcM, line 31: Can't we use `org.apache.commons.lang3.tuple.Pair` as elsewhere in record layer?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/Pair.java
tian-yizuo: Fixed it. And also opened an issue and added release note for this change.

tian-yizuo: Also added some javadoc comment for ```LuceneAnalyzerFactory.getQueryAnalyzer```.

MMcM (approved):

sophiabuell (approved):

PR #1489, opened 2021-12-09
Resolves #1488: Support synonym for lucene indexing.

### LINKED ISSUE #1488: Support synonym for lucene indexing.

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -0,0 +1,127 @@
...
+    @Nullable
+    private SynonymMap cachedSynonymMap = null;
+
+    public SynonymAnalyzer(@Nullable CharArraySet stopwords) {
+        super(stopwords);
+    }
+
+    @Override
+    protected TokenStreamComponents createComponents(String fieldName) {
+        final StandardTokenizer src = new StandardTokenizer();
```
jleach4, line 64: I thought the analyzer comes from CloudKit?  How would you handle non-english languages?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.text.ParseException;
+
+/**
+ * The analyzer for index with synonym enabled.
+ * Only for in fly analysis during query time.
+ * The synonyms are not indexed to disk. So new index rebuilding is needed if the wordnet file gets updated.
```
sophiabuell, line 50: you mean it does not need to be updated correct? That was somewhat the idea behind putting it at query time.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -0,0 +1,127 @@
...
+    @Nonnull
+    private SynonymMap getSynonymMap() {
+        if (cachedSynonymMap == null) {
+            cachedSynonymMap = buildSynonymMap();
+        }
+        return cachedSynonymMap;
+    }
+
+    private static SynonymMap buildSynonymMap() {
+        try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(WORDNET_FILE_NAME)) {
```
MMcM, line 86: How often does this happen? It looks like the file compresses around 4x with something relatively cheap, so we might want to consider that, too.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -0,0 +1,127 @@
...
+import java.io.InputStreamReader;
+import java.text.ParseException;
+
+/**
+ * The analyzer for index with synonym enabled.
+ * Only for in fly analysis during query time.
+ * The synonyms are not indexed to disk. So new index rebuilding is needed if the wordnet file gets updated.
+ */
+public class SynonymAnalyzer extends StopwordAnalyzerBase {
+    private static final String WORDNET_FILE_NAME = "wn_s.pl";
```
sophiabuell, line 53: Is this a issue where we won't be able to adjust the synonym map per the use case? This should be able to be specified by the user which synonyms they want and uploaded somehow I believe

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
tian-yizuo: yeah the "new" is a typo. Good catch!

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
tian-yizuo: I think it might makes sense to support per client synonym map. How about to support it in a followup PR?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
tian-yizuo: Fixed the ```new``` to ```no```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
sophiabuell: I think this ties into john’s question about languages. But yes that’s ok I think. For example it may be good to think about how synonyms can be defined without pushing a new file every time to public record layer repo.

MMcM (approved):

PR #1491, opened 2021-12-10
Resolves #1490: log4j configuration

### LINKED ISSUE #1490: Fix jog4j configuration to patch vulnerability

ohadzeliger:
Change logj4 configuration to patch vulnerability

### DIFF
--- a/fdb-record-layer-core/src/main/resources/log4j.properties
+++ b/fdb-record-layer-core/src/main/resources/log4j.properties
@@ -3,3 +3,5 @@ log4j.rootLogger=DEBUG, console
 log4j.appender.console=org.apache.log4j.ConsoleAppender
 log4j.appender.console.layout=org.apache.log4j.PatternLayout
 log4j.appender.console.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
+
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/resources/log4j.properties
```diff
@@ -3,3 +3,5 @@ log4j.rootLogger=DEBUG, console
 log4j.appender.console=org.apache.log4j.ConsoleAppender
 log4j.appender.console.layout=org.apache.log4j.PatternLayout
 log4j.appender.console.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
+
+log4j.formatMsgNoLookups=true
```
alecgrieser, line 7: We probably want to add this to `gradle/scripts/log4j-test.properties` and `fdb-record-layer-core/src/test/resources/standalone.log4j.properties` as well. I'm...not actually sure what this `log4j.properties` file is for, given that the main record layer dependency only tries to export slf4j as a dependency, not log4j, but ¯\\\_(ツ)_/¯

alecgrieser (changes requested):

PR #1492, opened 2021-12-10
Resolves #1493: Consider replacing single record range deletes with multiple single key deletes

### LINKED ISSUE #1493: Consider replacing single record range deletes with multiple single key deletes
At the moment, when a single record is deleted, we (typically) will end up issuing one range delete that covers the whole record. See this implementation in `KeyValueUnsplitter`:

https://github.com/FoundationDB/fdb-record-layer/blob/ceaaeb538d3328ad6e3b32f03c34c3764a0195b6/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelper.java#L239-L250

Note that the `previousSizeInfo` contains enough information to know how many keys (and which ones) were used to store the previous record, which also means that it knows enough to reconstruct which split points existed for the record in many cases. However, even if that information is known, the delete logic still issues one range delete instead of one delete for each split point.

The current approach has some advantages over issuing multiple single-key deletes including:

1. *Simplicity*. Because we know all of the keys for the record are stored under a single prefix, it's easier to be confident in this logic than in making sure we have correctly considered all subtleties around the way that split records work
2. *Fewer JNI calls*. Each clear would be its own JNI hop, and so a single clear decreases the number of these
3. *Less data in the final commit*. Each delete ends up in the final commit as its own mutation set, and so doing fewer clears means that the final commit message can be smaller
4. *Less work in the resolver*. The one range delete shows up in the FDB transaction resolver as a single mutation range rather than as one range for each key

However, single-key deletes have the following advantage: it is easier for the FDB storage engine to optimize single key deletes. With the current FDB ssd engine (based on sqlite), the difference is usually fairly small: range deletes are inserted into the B-tree, either deleting the relevant keys or inserting a range tombstone into the B-tree page over the relevant range that will be "lazily deleted" (potentially recursively). For single-record clears, the range should typically be over one or more keys on a single leaf page in the B-tree and thus shouldn't need a range tombstone at all, though it may require one in some circumstances. (Similar analysis applies to the Redwood storage engine in development.)

For the in-development RocksDB storage engine (or LSM storage engines more generally), the story is a little different, as all range clears would necessitate either iterating over the range and issuing single key clears for each read key or inserting a range tombstone. At read time, the range tombstones need to be consulted and applied to avoid reading extraneous pages, and the range tombstones also need to be merged in to during compactions. That can be fairly CPU intensive, and so minimizing the number of range tombstones would be preferable. In theory, there are some FDB server optimizations that could be done to support this (if the storage engine could somehow try and perform the delete key-by-key more aggressively at least in some circumstances). There are also some FDB client fixes to this (e.g., for any range delete that intersects with a read range in the read-your-write cache, the FDB client could automatically translate the range delete into a sequence of single key deletes). Finally, the Record Layer could modify some of its delete paths to try and decrease the number of range deletes it issues, the most avoidable of which being single record deletes.

Note: at the FDB protocol level, there actually isn't a "single key clear" and all clears are range clears. The "clear key" operation is translated into a range clear over a single key range (i.e., to clear a single key, the range starting at that key and ending at that key plus `\0` is cleared) at the protocol level. So, for any storage engine to somehow optimize the single key range clear case, it would need to either change the protocol to introduce a separate `ClearKey` mutation type or translate the range back into its original single key.

As a note: one slight optimization to naïvely issuing a series of single-key clears would be to issue: (1) one write conflict range over the whole record and (2) one clear for each key. This would decrease the commit size over the naïve approach (as the final commit would need to have one mutation range for each key and one conflict range for each record instead of one mutation range and one conflict range for each record) though it would still be bigger than the current delete size, and it would mean that identical data would be sent to the resolver. However, it does come at the cost of one additional JNI hop over the naïve approach.

alecgrieser:
This changes the single-record delete path so that in cases where we know which key(s) the record used (including taking split points into account), we issue a series of single-key deletes rather than one clear range over the whole record. As issue #1493 describes in more detail, this approach should be more efficient for the currently-in-development RocksDB storage engine (as well as any other LSM storage engine that might one day be used) while being more-or-less a wash for the other storage engines.

This resolves #1493.


alecgrieser:
Closing this for now. We may want to do something like this for the RocksDB storage engine, so we may need to re-open this at a later date.

MMcM:
I am wondering whether we want a feature flag to allow for more controlled testing of this optimization. The necessary plumbing doesn't look straightforward, though. The best I can come up with is replacing `clearBasedOnPreviousSizeInfo` with an enum and getting that from the `RecordSerializer`, which is regularly extended.

alecgrieser:
Rebase updates branch against latest `main`

alecgrieser:
> I am wondering whether we want a feature flag to allow for more controlled testing of this optimization

I think it would be straightforward enough to add a `RecordLayerPropertyKey` to control this if we wanted to control the testing/rollout of the optimization. It should be safe to do this, I think, given that the old way and the new way should be compatible in various mixed modes, and so there shouldn't be an issue if some versions have this, some instances don't, though I guess we could add some mixed mode testing here in the Record Layer. I'm not super in love with adding more properties here, but it shouldn't be _that_ messy.

I'm also looking a bit closer now at `clearBasedOnPreviousSizeInfo`, and it feels like it should be possible to remove this as an independent parameter here, and it _feels_ like it should be removed entirely. It's usages appear to set it to `true` except in `FDBMetaDataStore`, which is a bit of a special use case of `SplitHelper` anyway, and though I haven't convinced myself of it, seems like it should be able to use the previous split information as well. (But this is probably separate from this PR anyway.)

alecgrieser:
Added a property in bf3e7b1 to allow this to be more selectively rolled out.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/FDBRecordStoreProperties.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Whether single record deletes should unroll range deletes into multiple singe key deletes. This is used
+     * during {@link com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore#saveRecord(Message)} and
+     * {@link com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore#deleteRecord(Tuple)} to control
+     * whether multiple FDB key clears are issued when a record is split across multiple keys, or if instead a single
+     * range clear is issued that covers all keys for that record. Issuing multiple single-key clears can be
+     * more efficient if the underlying FDB storage engine can more efficiently handle single-key clears.
+     */
+    public static final RecordLayerPropertyKey<Boolean> UNROLL_SINGLE_RECORD_DELETES = RecordLayerPropertyKey.booleanPropertyKey(
+            "com.apple.foundationdb.record.recordstore.unroll_single_record_deletes", false);
```
ScottDugas, line 45: Would it make sense to have this be a long, and only unroll if the number of keys is not too many?
i.e. if the record is split into 100 key-value pairs, would it make sense to do a range delete, versus 5 key-value pairs?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
```diff
@@ -121,13 +123,16 @@ public void setSubspace() {
     }
 
     @Nonnull
-    public static Stream<Arguments> splitAndSuffixArgs() {
-        // Arguments: splitLongRecords, omitUnsplitSuffix
-        // Note that "true", "true" is not valid
+    public static Stream<Arguments> splitSuffixAndUnrollArgs() {
```
ScottDugas, line 126: This might be more readable with a `flatMap`, e.g.
```
Stream.of(false, true).flatMap(splitLongRecords -> 
    (splitLongRecords ? Stream.of(false) : Stream.of(false, true)).flatMap(omitUnsplitSuffix -> 
        Stream.of(false, true).map(unrollSingleRecordDeletes -> Arguments.of(splitLongRecords, omitUnsplitSuffix, unrollSingleRecordDeletes))))
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/FDBRecordStoreProperties.java
```diff
@@ -0,0 +1,50 @@
...
+public final class FDBRecordStoreProperties {
+    /**
+     * Whether single record deletes should unroll range deletes into multiple singe key deletes. This is used
+     * during {@link com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore#saveRecord(Message)} and
+     * {@link com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore#deleteRecord(Tuple)} to control
+     * whether multiple FDB key clears are issued when a record is split across multiple keys, or if instead a single
+     * range clear is issued that covers all keys for that record. Issuing multiple single-key clears can be
+     * more efficient if the underlying FDB storage engine can more efficiently handle single-key clears.
+     */
+    public static final RecordLayerPropertyKey<Boolean> UNROLL_SINGLE_RECORD_DELETES = RecordLayerPropertyKey.booleanPropertyKey(
```
jjezra, line 44: Should the property name be more explicit? Like CONVERT_RANGE_DELETES_TO_SINGLE_RECORD_DELETES?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
```diff
@@ -462,7 +472,7 @@ public void deleteWithSplit(boolean splitLongRecords, boolean omitUnsplitSuffix)
 
     @ParameterizedTest(name = "deleteWithSplitAndVersion [splitLongRecords = {0}]")
     @BooleanSource
```
ScottDugas, line 474: Should this be parameterized by `unrollSingleRecordDeletes`? (or omitUnsplitSuffix, but that was already missing)

docs/ReleaseNotes.md
```diff
@@ -23,7 +23,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** Single record deletes now issue multiple single-key clears instead of one range clear, which should be an improvement for LSM-based FDB storage engines [(Issue #1493)](https://github.com/FoundationDB/fdb-record-layer/issues/1493)
```
ScottDugas, line 26: The issue describes: 
> Note: at the FDB protocol level, there actually isn't a "single key clear" and all clears are range clears. The "clear key" operation is translated into a range clear over a single key range (i.e., to clear a single key, the range starting at that key and ending at that key plus \0 is cleared) at the protocol level. So, for any storage engine to somehow optimize the single key range clear case, it would need to either change the protocol to introduce a separate ClearKey mutation type or translate the range back into its original single key.

Has that situation changed, or is this actually going to result in more range deletes on the storage engine, making things worse?

MMcM (dismissed):

ScottDugas (changes requested):

docs/ReleaseNotes.md
MMcM: The RocksDB storage engine recognizes single key range clears: https://github.com/apple/foundationdb/blob/main/fdbserver/KeyValueStoreRocksDB.actor.cpp#L1865-L1866.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/FDBRecordStoreProperties.java
jleach4: The point rangeDeletes are removed when a set operation overwrites that point...

ScottDugas (approved):

MMcM (approved):

PR #1496, opened 2021-12-23
Log message keys should automatically be lowercase when printed

scgray:


alecgrieser:
Closing in favor of #1496

PR #1498, opened 2022-01-04
Resolves #1497and #9: in-to-join and in-to-union planning

### LINKED ISSUE #1497: implement in-to-join alongside with in-to-union in Cascades planner
The Cascades planner currently only supports planning of a subset of cases involving INs by using an IN-UNION plan. That can be dangerous as it is better to use an IN-JOIN plan if possible (memory footprint). Also we should support any number of INs and not just one IN per query block.

normen662:


ohadzeliger (changes requested):
Should there be a configuration parameter for the planner to restrict which in-plan to use?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/SortedInValuesSource.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public boolean isSorted() {
+        return true;
+    }
+
+    @Override
+    public boolean isReverse() {
+        return false;
```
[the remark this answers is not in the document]
normen662: Good catch.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InSource.java
```diff
@@ -0,0 +1,80 @@
...
+        return getValues(null);
+    }
+
+    @Nonnull
+    protected abstract List<Object> getValues(@Nullable EvaluationContext context);
+
+    @Nonnull
+    public abstract RecordQueryInJoinPlan toInJoinPlan(@Nonnull final Quantifier.Physical innerQuantifier);
+
+    public int baseHash(@Nonnull final PlanHashKind hashKind, @Nonnull ObjectPlanHash objectPlanHash) {
```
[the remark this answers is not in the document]
normen662: No -- it's not. This code taken from `RecordQueryInUnionPlan`.

https://github.com/FoundationDB/fdb-record-layer/blob/da38c80d5addfec943f0d3957c5cb3630e957ec0/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionPlan.java#L298

This is really unfortunate but I don't think we can just add that here as this will regress all kinds of plan hashes.

I will add a big todo here. We need to roll the plan hashes in a coordinated manner anyway. This should be done then as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InSource.java
```diff
@@ -0,0 +1,80 @@
...
+
+    protected abstract int size(@Nonnull EvaluationContext context);
+
+    @Nonnull
+    public List<Object> getValues() {
+        return getValues(null);
+    }
+
+    @Nonnull
+    protected abstract List<Object> getValues(@Nullable EvaluationContext context);
```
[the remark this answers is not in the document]
normen662: This is a refactor from

https://github.com/FoundationDB/fdb-record-layer/blob/da38c80d5addfec943f0d3957c5cb3630e957ec0/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionPlan.java#L296

I think per se there is no requirement in `InSource` for the objects to be `Comparable` at all. That only gets imposed in the `Sorted...InSource` subclasses.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InParameterSource.java
```diff
@@ -0,0 +1,119 @@
...
+    @SpotBugsSuppressWarnings("NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE")
+    protected List<Object> getValues(@Nullable final EvaluationContext context) {
+        return getBoundValues(Objects.requireNonNull(context));
+    }
+
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    protected List<Object> getBoundValues(@Nonnull final EvaluationContext context) {
+        final List<Object> binding = (List<Object>)context.getBinding(getParameterName());
+        return Objects.requireNonNullElse(binding, Collections.emptyList());
```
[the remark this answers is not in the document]
normen662: Again, this code retains the original behavior from InJoins/InUnions. I agree with you, but that's what it was coded up like :-(

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/SortedInParameterSource.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * InParameterSource.java
```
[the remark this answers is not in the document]
normen662: Done. BTW, I would have liked to have fewer subclasses of `InSource`, however, I needed to really carefully avoid the old planner creating old plans with new plan hashes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InSource.java
```diff
@@ -0,0 +1,80 @@
...
+    public abstract RecordQueryInJoinPlan toInJoinPlan(@Nonnull final Quantifier.Physical innerQuantifier);
+
+    public int baseHash(@Nonnull final PlanHashKind hashKind, @Nonnull ObjectPlanHash objectPlanHash) {
+        return objectPlanHash.planHash(hashKind);
+    }
+
+    @Nonnull
+    public static List<Object> sortValues(@Nonnull List<Object> values, final boolean isReversed) {
+        if (values.size() < 2 ) {
+            return values;
```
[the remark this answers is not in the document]
normen662: I think this is fine and also just coming from a refactor. The function's contract is to just return the sorted list back without changing the original input list. The copy happens since `.sort()` is sorting in place.

In order to make the contract stronger I will wrap the singleton element into aniqmmutable list factory method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InSource.java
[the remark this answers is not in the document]
ohadzeliger: also, we can *not* use this method for the cascades planner (have it issue a different hash)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InSource.java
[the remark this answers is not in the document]
ohadzeliger: Sounds good.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InParameterSource.java
```diff
@@ -0,0 +1,119 @@
...
+import com.apple.foundationdb.record.PlanHashable;
+import com.apple.foundationdb.record.query.plan.temp.Quantifier;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public class InParameterSource extends InSource {
```
MMcM, line 36: Missing Javadoc. Why didn't this fail the build?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InValuesSource.java
```diff
@@ -0,0 +1,113 @@
...
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.ObjectPlanHash;
+import com.apple.foundationdb.record.PlanHashable;
+import com.apple.foundationdb.record.query.plan.temp.Quantifier;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+
+public class InValuesSource extends InSource {
```
MMcM, line 33: Missing Javadoc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/SortedInParameterSource.java
```diff
@@ -0,0 +1,99 @@
...
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.ObjectPlanHash;
+import com.apple.foundationdb.record.PlanHashable;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Objects;
+
+public class SortedInParameterSource extends InParameterSource {
```
MMcM, line 33: Missing Javadoc. Needs to clarify, in particular, the difference between a source that is _already_ sorted and one that _needs to be_ sorted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/SortedInValuesSource.java
```diff
@@ -0,0 +1,82 @@
...
+package com.apple.foundationdb.record.query.plan.plans;
+
+import com.apple.foundationdb.record.ObjectPlanHash;
+import com.apple.foundationdb.record.PlanHashable;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Objects;
+
+public class SortedInValuesSource extends InValuesSource {
```
MMcM, line 30: Missing Javadoc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
```diff
@@ -236,27 +250,31 @@ private OptionalInt comparePrimaryScanToIndexScan(@Nonnull Map<Class<? extends R
         return OptionalInt.empty();
     }
 
-    private OptionalInt compareInUnion(@Nonnull final RelationalExpression leftExpression,
-                                       @Nonnull final RelationalExpression rightExpression) {
-        if (!(leftExpression instanceof RecordQueryInUnionPlan)) {
+    /**
+     * This comparator compares the left expression which must be of type {@link RecordQueryInUnionPlan} or
+     * {@link RecordQueryInJoinPlan} and only returns and indication that the other plan is considered preferable
```
MMcM, line 255:
```suggestion
     * {@link RecordQueryInJoinPlan} and only returns an indication that the other plan is considered preferable
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
```diff
@@ -236,27 +250,31 @@ private OptionalInt comparePrimaryScanToIndexScan(@Nonnull Map<Class<? extends R
...
+     * or that this plan and the other plan are comparable. It never returns that the in-plan should be preferable.
+     * The reasoning behind this is to avoid plans that were generated out of an IN-transformation that wasn't able
+     * to translate the rewritten equality into an index search argument (SARG).
+     * @param leftExpression this expression
+     * @param rightExpression other expression
+     * @return {@code OptionalInt.empty()} is the comparator is unable to compare the two expressions handed in. That
```
MMcM, line 261:
```suggestion
     * @return {@code OptionalInt.empty()} if the comparator is unable to compare the two expressions handed in. That
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/rules/ImplementInJoinRule.java
```diff
@@ -0,0 +1,354 @@
...
+            if (Sets.intersection(comparisonsCorrelatedTo, explodeAliases).isEmpty()) {
+                //
+                // This case covers comparisons such as  a = <literal> or a = <correlation> where
+                // <correlation> is anchored above and therefore this comparison causes an equality-bound
+                // expression.
+                //
+                // Example: requested ordering: a, b, c
+                //          ins: a explodes over (1, 2)
+                //               c explodes over ('x', 'y')
+                //          plan has equality-bound expressions over a, b, c
```
MMcM, line 227: Why is `b` bound?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InParameterSource.java
normen662: Done. I think this is not an error anymore, as (for as long we are on Java 11) we have to disable warning -> error promotion for javadoc due to an internal JDK 11 bug. That big is fixed in Java 14.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/InValuesSource.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/SortedInParameterSource.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/CascadesCostModel.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/rules/ImplementInJoinRule.java
normen662: Elaborated a little bit:

```
//
// This case covers comparisons such as  a = <literal> or a = <correlation> where
// <correlation> is anchored above and therefore this comparison causes an equality-bound
// expression.
//
// Example: requested ordering: a, b, c
//          INs: a explodes over (1, 2)
//               c explodes over ('x', 'y')
//          inner plan has equality-bound expressions over a, b, c
//
// b is not among the INs, but it is bound via equality somehow. That binding can be a constant or
// another correlation that is anchored above the current SELECT.
//
```

The plan at consideration as equality-bound fields `a, b, c`. That is given. If we look at `b`, `b` just looks like it is a comparison to something but nothing related to the INs in this expression. We should just ignore those fields. (That also is true for things like `b = 5`.

ohadzeliger (approved):

MMcM (approved):

PR #1484, opened 2021-12-08
Log message keys should automatically be lowercase when printed

scgray:

PR #1499, opened 2022-01-04
Resolves #1500: Log message keys should automatically be lowercase when printed

### LINKED ISSUE #1500: Log message keys should automatically be lowercase when printed
Right now, the `LogMessageKeys` enum requires the user to specify a log key in addition to the enum value. However, almost all of them just use the name of the key (in lower case), so it would be more ergonomic if those enums automatically set that value so that developers didn't need to repeat themselves.

alecgrieser:
This is fixing a merge conflict in #1496 that prevented the earlier PR from being merged. Once this is in, we can close that other PR.

This resolves #1500.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
```diff
@@ -32,219 +33,219 @@
...
-    SOURCE_FILE("source_file"),
-    CONTINUED_BUILD("continued_build"),
-    INDEXING_METHOD("indexing_method"),
-    ALLOW_REPAIR("allow_repair"),
-    INDEXING_POLICY("indexing_policy"),
-    INDEXING_POLICY_DESIRED_ACTION("indexing_policy_desired_action"),
+    RESULT,
+    INDEXER_CURR_RETRY,
+    INDEXER_MAX_RETRIES,
+    DIRECTOY,
```
ScottDugas, line 152: It looks like there was a typo in the name
```suggestion
    DIRECTOY("directory"),
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
```diff
@@ -32,219 +33,219 @@
...
+    RESOLVER_PATH,
+    RESOLVER_METADATA,
+    RESOLVER_REVERSE_VALUE,
+    RESOLVER_VALUE,
+    VALIDATION_RESULT,
+    CACHED_KEY,
+    CACHED_STATE,
+    READ_STATE,
+    SHARED_READ_VERSION,
+    MUTATION,
```
ScottDugas, line 232:
```suggestion
    MUTATION("state_mutation"),
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
```diff
@@ -253,38 +254,43 @@
...
+    COMPRESSION_VERSION,
+    DATA_SIZE,
+    DATA_VALUE,
+    ENCODED_DATA_SIZE,
+    ENCRYPTED_EVENTUALLY,
+    ENCRYPTION_SUPPOSED,
+    FILE_ID,
+    FILE_NAME,
+    FILE_REFERENCE,
+    ORIGINAL_DATA_SIZE,
```
ScottDugas, line 281: There's an inconsistency here, where it uses a `-` instead of a `_`
```suggestion
    ORIGINAL_DATA_SIZE("original_data-size"),
```

ScottDugas (changes requested):
Not everything that was updated should be using `toLowerCase`, at least not as-is. It might make sense to change the name of `DIRECTOY` and the log value of `ORIGINAL_DATA_SIZE`, but those would be breaking changes, so maybe not worth it right now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
alecgrieser: I went ahead and changed `DIRECTOY` to `DIRECTORY` because I'm pretty sure the name was a typo

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
alecgrieser: This change was deliberate, as noted in this comment: https://github.com/FoundationDB/fdb-record-layer/pull/1484#discussion_r764766278 The initial hyphen was (presumably) unintentional. I'm not actually sure what keys loggers generally accept, but I think it would generally be good to keep these as a-z and underscore

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
alecgrieser: 👍

ScottDugas (approved):

PR #1505, opened 2022-01-14
Resolves #1504: Support auto complete suggestions for Lucene search

### LINKED ISSUE #1504: Support auto complete suggestions for Lucene search

tian-yizuo:


MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -118,7 +138,20 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
+                      || scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE);
+
+        if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE) {
+            if (suggester == null) {
+                throw new UnsupportedOperationException("Unsupported operation due to no auto complete enabled for the index");
+            }
+            if (continuation != null) {
+                throw new RecordCoreArgumentException("Auto complete does not support scanning with continuation");
+            }
+            return new LuceneAutoCompleteResultCursor(suggester, range.getLow().getString(0), executor, scanProperties, state, null, highlightForAutoCompleteIfEnabled);
```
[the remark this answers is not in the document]
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -152,16 +185,33 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
         Object value = field.getValue();
         Field luceneField;
         switch (field.getType()) {
             case TEXT:
                 luceneField = new TextField(fieldName, (String)value, field.isStored() ? Field.Store.YES : Field.Store.NO);
+                addTermToSuggesterIfNeeded((String) value, fieldName);
                 break;
             case STRING:
                 luceneField = new StringField(fieldName, (String)value, field.isStored() ? Field.Store.YES : Field.Store.NO);
+                addTermToSuggesterIfNeeded((String) value, fieldName);
```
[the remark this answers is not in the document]
tian-yizuo: Removed it for STRING field type.

MMcM (changes requested):

gradle/strict.gradle
```diff
@@ -22,7 +22,7 @@
 
 tasks.withType(JavaCompile) {
     // suppress static due to protoc generating super.addAll calls.
-    options.compilerArgs += ["-Xlint:all", "-Xlint:-processing", "-Xlint:-options", "-Xlint:-static", "-Werror"]
```
MMcM, line 25: Was this an accident. I believe error-on-warn is part of the definition of "strict."

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -0,0 +1,183 @@
...
+    private final AnalyzingInfixSuggester suggester;
+    private final boolean highlight;
+
+    public LuceneAutoCompleteResultCursor(@Nonnull AnalyzingInfixSuggester suggester, @Nonnull String query,
+                                          @Nonnull Executor executor, @Nonnull ScanProperties scanProperties,
+                                          @Nonnull IndexMaintainerState state, boolean highlight) {
+        this.suggester = suggester;
+        this.query = query;
+        if (query.contains(":")) {
+            String[] subStrings = query.split(":", 2);
```
sophiabuell, line 86: should we check here if it is passing a more complex query?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -0,0 +1,183 @@
...
+        return visitor.visitLeave(this);
+    }
+
+    private void performLookup() throws IOException {
+        if (lookupResults != null) {
+            return;
+        }
+        long startTime = System.nanoTime();
+        lookupResults = suggester.lookup(searchTerm,
+                fieldName == null ? Collections.emptySet() : Collections.singleton(new BytesRef(fieldName.getBytes(StandardCharsets.UTF_8))),
```
sophiabuell, line 176: does this default with a null field name to the suggester checking across all fields?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -0,0 +1,183 @@
...
+
+    public LuceneAutoCompleteResultCursor(@Nonnull AnalyzingInfixSuggester suggester, @Nonnull String query,
+                                          @Nonnull Executor executor, @Nonnull ScanProperties scanProperties,
+                                          @Nonnull IndexMaintainerState state, boolean highlight) {
+        this.suggester = suggester;
+        this.query = query;
+        if (query.contains(":")) {
+            String[] subStrings = query.split(":", 2);
+            this.fieldName = subStrings[0];
+            this.searchTerm = subStrings[1];
```
sophiabuell, line 88: We should put checks on this for substance. '*:*' is a "valid" query as is ':' since it doesn't go through the lucene parser yet.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -0,0 +1,183 @@
...
+        } else {
+            this.fieldName = null;
+            this.searchTerm = query;
```
sophiabuell, line 91: may be worth the same check here.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
sophiabuell: additionally its possible that we could pull the field names and pass them through the constructor. That would avoid us mistakenly checking a field that we're not intending to such as the primary key or maybe any other fields we make in the future.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -181,14 +227,15 @@ private void insertField(LuceneDocumentFromRecord.DocumentField field, final Doc
...
+    private void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
+                               byte[] primaryKey, @Nullable AnalyzingInfixSuggester suggester) throws IOException {
         final IndexWriter newWriter = getOrCreateIndexWriter(state, indexAnalyzer, executor, groupingKey);
         BytesRef ref = new BytesRef(primaryKey);
         Document document = new Document();
         document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
         document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
-        for (LuceneDocumentFromRecord.DocumentField field : fields ) {
-            insertField(field, document);
+        for (LuceneDocumentFromRecord.DocumentField field : fields) {
```
sophiabuell, line 237: Would this add the suggester to the primary key fields? is that what we want?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -232,10 +279,22 @@ private void deleteDocument(Tuple groupingKey, byte[] primaryKey) throws IOExcep
...
+        Map<Tuple, AnalyzingInfixSuggester> suggesterMap = new HashMap<>();
+
         // update new
         try {
             for (Map.Entry<Tuple, List<LuceneDocumentFromRecord.DocumentField>> entry : newRecordFields.entrySet()) {
-                writeDocument(entry.getValue(), entry.getKey(), newRecord.getPrimaryKey().pack());
+                if (autoCompleteEnabled) {
+                    final AnalyzingInfixSuggester suggester = getSuggester(entry.getKey());
+                    suggesterMap.put(entry.getKey(), suggester);
+                    writeDocument(entry.getValue(), entry.getKey(), newRecord.getPrimaryKey().pack(), suggester);
```
sophiabuell, line 290: Since the primary key is getting saved possible elsewhere split up we probably don't want to add the suggester to it. It seems like that could expose problems if the user didn't intend it to be searched along with the indexed fields.

sophiabuell: If the fields in the primary key are included in the indexed fields they'll be added to the suggester individually right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: We probably want to ignore the primary key field here since it may contain fields (in my understanding) that the user isn't intending to index/make searchable?

gradle/strict.gradle
tian-yizuo: @MMcM ah sorry this is not supposed to be committed. umm the problem is that it reports warning as error because the code in ```QueryPlanUtils.getIndexEntryFDBQueriedRecordFunction()``` casts the ```toRecord.toRecord(recordDescriptor, indexEntry)``` to ```M```, which was existing code and I just refactored it for reuse.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -232,10 +279,22 @@ private void deleteDocument(Tuple groupingKey, byte[] primaryKey) throws IOExcep
...
 
+        Map<Tuple, AnalyzingInfixSuggester> suggesterMap = new HashMap<>();
+
         // update new
         try {
             for (Map.Entry<Tuple, List<LuceneDocumentFromRecord.DocumentField>> entry : newRecordFields.entrySet()) {
-                writeDocument(entry.getValue(), entry.getKey(), newRecord.getPrimaryKey().pack());
+                if (autoCompleteEnabled) {
+                    final AnalyzingInfixSuggester suggester = getSuggester(entry.getKey());
+                    suggesterMap.put(entry.getKey(), suggester);
```
sophiabuell, line 289: can you put this line in the getSuggester method and make reduce this to an inline if for the argument? I'm not sure if thats easier but it would put these all on one line. something like 
``` writeDocument(entry.getValue(), entry.getKey(), newRecord.getPrimaryKey().pack(), autoCompleteEnabled ? getSuggester(entry.getKey()) : null); 
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -138,6 +149,58 @@ public static LuceneIndexQueryPlan merge(LuceneIndexQueryPlan plan1, LuceneIndex
...
+                                                                      @Nonnull EvaluationContext context,
+                                                                      @Nullable byte[] continuation,
+                                                                      @Nonnull ExecuteProperties executeProperties) {
+        if (scanType != IndexScanType.BY_LUCENE_AUTO_COMPLETE) {
+            return super.executePlan(store, context, continuation, executeProperties);
+        } else {
+            final RecordMetaData metaData = store.getRecordMetaData();
+            final Index index = metaData.getIndex(indexName);
+            final Collection<RecordType> recordTypes = metaData.recordTypesForIndex(index);
+            if (recordTypes.size() != 1) {
```
sophiabuell, line 167: if no lucene index should do this then I think it should be a check for all lucene indexes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -138,6 +149,58 @@ public static LuceneIndexQueryPlan merge(LuceneIndexQueryPlan plan1, LuceneIndex
...
+    /**
+     * Override here to have specific logic to build the {@link QueryResult} for lucene auto complete suggestion result.
+     */
+    @Nonnull
+    @Override
+    public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull FDBRecordStoreBase<M> store,
+                                                                      @Nonnull EvaluationContext context,
+                                                                      @Nullable byte[] continuation,
+                                                                      @Nonnull ExecuteProperties executeProperties) {
+        if (scanType != IndexScanType.BY_LUCENE_AUTO_COMPLETE) {
```
sophiabuell, line 161: This should instead be the else and the other clause should have `if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE)`. This allows for other use cases more easily.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -93,6 +95,67 @@ void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Bui
...
+        @Override
+        public void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Builder recordBuilder,
+                         @Nonnull IndexEntry kv) {
+            Tuple keyTuple = kv.getKey();
+            if (keyTuple.size() < 2) {
+                throw new RecordCoreException("Invalid key tuple for auto-complete suggestion's index entry")
+                        .addLogInfo(LogMessageKeys.KEY_SIZE, keyTuple.size())
+                        .addLogInfo(LogMessageKeys.RECORD_TYPE, recordDescriptor.getName());
+            }
+            String fieldName = (String) keyTuple.get(keyTuple.size() - 2);
```
sophiabuell, line 115: Is there a reason you're not pulling the field from the index entry and instead pulling it from the key tuple? The kv has a field definition in it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -93,6 +95,67 @@ void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Bui
                   @Nonnull IndexEntry kv);
     }
 
+    /**
+     * The copier to populate the lucene auto complete suggestion as a value for the field where it is indexed from.
```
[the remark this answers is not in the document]
tian-yizuo: umm I guess we can let the client to control if they want the suggestion to be with highlighting or not (through index options for now). So I guess there is no need to tell if the suggestion result is with highlight or not in the response. And in the future we probably can support clients to select this in each request rather than configuring it in schema.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
tian-yizuo: The suggester only adds the term in for ```TEXT``` fields. So I assume all fields with ```TEXT``` type are searchable?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -93,6 +95,67 @@ void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Bui
...
+            }
+            String fieldName = (String) keyTuple.get(keyTuple.size() - 2);
+            String value = (String) keyTuple.get(keyTuple.size() - 1);
+            buildFieldForMessage(fieldName, value, recordDescriptor, recordBuilder);
+        }
+    }
+
+    private static boolean buildFieldForMessage(@Nonnull String fieldName, @Nonnull String value,
+                                                @Nonnull Descriptors.Descriptor messageDescriptor, @Nonnull Message.Builder builder) {
+        // If the target field is a direct String type sub-field of current message
```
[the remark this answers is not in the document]
tian-yizuo: Done, please check the PR (https://github.com/FoundationDB/fdb-record-layer/pull/1512), which is also included in this one now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryPlanUtils.java
```diff
@@ -0,0 +1,59 @@
...
+ * Utility class for query planning.
+ */
+public class QueryPlanUtils {
+    private QueryPlanUtils() {
+    }
+
+    /**
+     * The method to get a function from an {@link IndexEntry} to a {@link FDBQueriedRecord} representing a partial record.
+     */
+    public static <M extends Message> Function<IndexEntry, FDBQueriedRecord<M>> getIndexEntryFDBQueriedRecordFunction(final @Nonnull FDBRecordStoreBase<M> store,
```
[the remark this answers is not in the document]
tian-yizuo: Renamed this method, please let me know if it looks better now.

gradle/strict.gradle
tian-yizuo: Reverted this accidental change and suppressed the warning on that method.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
tian-yizuo: Although we follow the same syntax for the lucene function, I guess we don't have to support more complex query than a search term plus an optional field, for the auto-complete case.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
tian-yizuo: Added check for the query to be parsed for search term and field, which is pretty straightforward. please let me know if it looks good to you

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
tian-yizuo: Yeah check is added and right now a query with an empty search term will be rejected.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
tian-yizuo: Yeah put the initialization of suggester into the ```writeDocument``` now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
tian-yizuo: I am not sure I understand where to get the field definition from the ```IndexEntry```. And this passed ```IndexEntry``` is actually a dummy one to include the grouping keys, field name, suggestion and suggestion value. I had a discussion with @MMcM and we agreed to put the keys, field name and suggestion into the key and the suggestion value into the value, although this is arbitrary in some degree.

sophiabuell: That makes sense. :) Its pretty arbitrary

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
sophiabuell: maybe? probably. Thats more one of those use case conditions thats not written. But its ok for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
tian-yizuo: added 3 tests for this check: ```searchForAutoCompleteWithFieldWithoutField```, ```searchForAutoCompleteWithoutFieldWithoutTerm```, and ```searchForAutoCompleteWithFieldWithoutTerm```.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
MMcM: Would it be better to have two different query functions, rather than having anything we need to parse?

Isn't there an issue of completing over all fields multiple times and having a previous completion contain a colon in the middle? For example, suppose with subjects:
* **please** ⟶ Please read: agenda for 1/
* Please read: agenda for 1/**2** ⟶ Please read: agenda for 1/28

sophiabuell: TODO for later: lets make sure this second search term isn't a more complex query. for example a fuzzy search shouldn't be accepted but thats a future issue. For now this is great.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
sophiabuell: does this work in terms of checking when a field isn't provided? I guess this isn't an issue for now because we pass fields via the lucene query function but somethingtothink about for later.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
```diff
@@ -0,0 +1,183 @@
...
+                    if (groupingKeyIndex > - 1) {
+                        if (groupingKeyIndex > groupingKey.size() - 1) {
+                            throw new RecordCoreException("Invalid grouping value tuple given a grouping key")
+                                    .addLogInfo(LogMessageKeys.VALUE, groupingKey.toString());
+                        }
+                        source.buildMessage(groupingKey.get(groupingKeyIndex), (String) value);
+                    } else if (type.equals(LuceneIndexExpressions.DocumentFieldType.TEXT) && fieldNameMatch(fieldName, luceneField, fieldNamePrefix, suffixOverride)) {
+                        source.buildMessage(suggestion, (String) value);
+                    }
+                    return;
```
sophiabuell, line 67: nit: This is an unnecessary return.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
```diff
@@ -0,0 +1,183 @@
...
+                },
+                null, 0, root instanceof GroupingKeyExpression ? ((GroupingKeyExpression) root).getGroupingCount() : 0);
+    }
+
+    private static boolean fieldNameMatch(@Nonnull String traversedFieldName, @Nonnull String givenFieldName,
+                                          @Nullable String fieldNamePrefix, boolean suffixOverride) {
+        // TODO: This is a workaround to deal with the issue that prefix of key path for grouping key could be missing, it will be fixed in a following commit
```
MMcM, line 74: I think it would be better to do an exact match on what is written today, rather than attempting a partial match.

I believe that means separating the grouping and grouped keys like `getGroupedFields` does before calling `LuceneIndexExpressions.getFields`. Or my making `LuceneIndexExpressions.getFields` take the root and have a more complicated contract that gives back the group as well.

This doesn't make it harder to change how the case of the group split in the middle of a nesting is handled later.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
sophiabuell: @MMcM makes a good point. Honestly we don't need to accommodate any field matching here since we have the option to pass the list of fields we want to query through the luceneQueryComponent. I guess it would be easier to just get rid of the parsing altogether.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
tian-yizuo: But the ```LuceneIndexExpression.getFields``` just splits the whole key to grouping keys and grouped keys that the ```entry``` doesn't exist in the field name, for the test case of ```LuceneDocumentFromRecordTest.groupingMap()```. But if I understand correctly, we agree to fix it to be ```entry_value``` rather than ```value``` in the ```TODO``` followup work? @MMcM

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
tian-yizuo: Removed

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
tian-yizuo: A separate PR to fix the existing issue got merge: https://github.com/FoundationDB/fdb-record-layer/pull/1514
Then this if block can be removed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
```diff
@@ -0,0 +1,178 @@
...
+        @Override
+        public Iterable<PartialRecordBuildSource> getChildren(@Nonnull FieldKeyExpression parentExpression) {
+            final String parentField = parentExpression.getFieldName();
+            final Descriptors.FieldDescriptor parentFieldDescriptor = descriptor.findFieldByName(parentField);
+            return Collections.singletonList(new PartialRecordBuildSource(this, parentFieldDescriptor, builder.newBuilderForField(parentFieldDescriptor)));
+        }
+
+        @Override
+        public Iterable<Object> getValues(@Nonnull FieldKeyExpression fieldExpression) {
+            return Collections.singletonList(fieldExpression.getFieldName());
```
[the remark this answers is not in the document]
MMcM: That is why I suggested that `PartialRecordBuildSource` be used to build a regular expression that can match a field name and thereby, in addition to selecting one of several possible source fields, give capturing groups for the dynamically named field name path elements.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
```diff
@@ -0,0 +1,178 @@
...
+                null, 0, root instanceof GroupingKeyExpression ? ((GroupingKeyExpression) root).getGroupingCount() : 0);
+    }
+
+    private static boolean fieldNameMatch(@Nonnull String traversedFieldName, @Nonnull String givenFieldName,
+                                          @Nullable String fieldNamePrefix, boolean suffixOverride) {
+        // If field is not overridden, the names have to match exactly
+        if (!suffixOverride) {
+            return traversedFieldName.equals(givenFieldName);
+        }
+        // If field is overridden by a LUCENE_FIELD_NAME function, and there is a field name prefix, the given name should have same prefix to match
```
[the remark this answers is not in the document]
tian-yizuo: made a new commit to populate the dynamic key field

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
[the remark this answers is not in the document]
tian-yizuo: Fixed the matching logic to compare all the non-dynamic field names on the path.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -589,7 +625,14 @@ void testDataLoad() {
                     .setText(randomWords[1])
                     .setGroup(2)
                     .build();
-            recordStore.saveRecord(dylan);
+            try {
+                recordStore.saveRecord(dylan);
+            } catch (Exception ex) {
+                System.out.println("***********************************Index number****************************************");
+                System.out.println("i=" + String.valueOf(i));
+                System.out.println("***************************************************************************************");
```
MMcM, line 633: Is this supposed to remain in the final version? It's only a test, but it still seems a little loud.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecordTest.java
```diff
@@ -139,6 +211,26 @@ void map() {
...
+
+        // The k1 is supposed to show up in the key field within repeated entry sub-message
+        Descriptors.FieldDescriptor keyField = TestRecordsTextProto.MapDocument.Entry.getDescriptor().findFieldByName("key");
+        assertTrue(entry.hasField(keyField));
+        assertEquals("k1", entry.getField(keyField));
+
+        // The suggestion is supposed to show up in value field within repeated entry sub-message
+        Descriptors.FieldDescriptor valueField = TestRecordsTextProto.MapDocument.Entry.getDescriptor().findFieldByName("value");
+        assertTrue(entry.hasField(valueField));
+        assertEquals("suggestion", entry.getField(valueField));
```
MMcM, line 233:
```suggestion
        Descriptors.Descriptor recordDescriptor = message.getDescriptorForType();
        TestRecordsTextProto.MapDocument.Builder builder = TestRecordsTextProto.MapDocument.newBuilder();
        LuceneIndexKeyValueToPartialRecordUtils.buildPartialRecord(index, recordDescriptor, builder, "k1", "suggestion");
        TestRecordsTextProto.MapDocument partialMsg = builder.build();

        assertEquals(1, partialMsg.getEntryCount());

        TestRecordsTextProto.MapDocument.Entry entry = partialMsg.getEntry(0);

        // The k1 is supposed to show up in the key field within repeated entry sub-message
        assertEquals("k1", entry.getKey());

        // The suggestion is supposed to show up in value field within repeated entry sub-message
        assertEquals("suggestion", entry.getValue());
```

MMcM: (No promise that actually compiles, but I hope the idea is clear.) All this dynamic message stuff seems unnecessary, since already gave it a typed `Builder`.

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
tian-yizuo: sorry this is debugging code kept here accidentally, will be removed

tian-yizuo: Removed it

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecordTest.java
tian-yizuo: Fixed all tests in ```LuceneDocumentFromRecordTest``` in this way.

MMcM (approved):

sophiabuell (approved):

PR #1511, opened 2022-01-27
Spellcheck2

sophiabuell:
This is based off Yizuo's commits and includes those as well. I will continue to rebase what he does on top.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
@@ -162,15 +162,16 @@ public class LuceneIndexQueryPlan extends RecordQueryIndexPlan {
                     .addLogInfo(LogMessageKeys.INDEX_NAME, indexName);
         }
 
-        if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE) {
+        if (scanType == IndexScanType.BY_LUCENE_SPELLCHECK) {
+
+        } if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE){
             final RecordType recordType = recordTypes.iterator().next();
             final RecordCursor<IndexEntry> entryRecordCursor = executeEntries(store, context, continuation, executeProperties);
             return entryRecordCursor
                     .map(QueryPlanUtils.getCoveringIndexEntryToPartialRecordFunction(store, recordType.getName(), indexName, getToPartialRecord(index, recordType), scanType))
                     .map(QueryResult::of);
-        } else {
-            return super.executePlan(store, context, continuation, executeProperties);
         }
+        return super.executePlan(store, context, continuation, executeProperties);
     }
 
     /**
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
@@ -157,10 +157,10 @@ public class LuceneIndexQueryPlan extends RecordQueryIndexPlan {
         final RecordMetaData metaData = store.getRecordMetaData();
         final Index index = metaData.getIndex(indexName);
         final Collection<RecordType> recordTypes = metaData.recordTypesForIndex(index);
+        if (recordTypes.size() != 1) {
+            throw new RecordCoreException("No lucene index should span multiple record types");
+        }
         if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE || scanType == IndexScanType.BY_LUCENE_SPELLCHECK) {
-            if (recordTypes.size() != 1) {
-                throw new RecordCoreException("No lucene index should span multiple record types");
-            }
             final RecordType recordType = recordTypes.iterator().next();
             final RecordCursor<IndexEntry> entryRecordCursor = executeEntries(store, context, continuation, executeProperties);
             return entryRecordCursor
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
@@ -162,13 +162,15 @@ public class LuceneIndexQueryPlan extends RecordQueryIndexPlan {
                     .addLogInfo(LogMessageKeys.INDEX_NAME, indexName);
         }
 
-        if (scanType == IndexScanType.BY_LUCENE_SPELLCHECK) {
-
-        } if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE){
+        if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE || scanType == IndexScanType.BY_LUCENE_SPELLCHECK){
+            if (recordTypes.size() != 1) {
+                throw new RecordCoreException("No lucene index should span multiple record types");
+            }
             final RecordType recordType = recordTypes.iterator().next();
             final RecordCursor<IndexEntry> entryRecordCursor = executeEntries(store, context, continuation, executeProperties);
             return entryRecordCursor
-                    .map(QueryPlanUtils.getCoveringIndexEntryToPartialRecordFunction(store, recordType.getName(), indexName, getToPartialRecord(index, recordType), scanType))
+                    .map(QueryPlanUtils.getCoveringIndexEntryToPartialRecordFunction(store, recordType.getName(), indexName,
+                            getToPartialRecord(index, recordType, scanType), scanType))
                     .map(QueryResult::of);
         }
         return super.executePlan(store, context, continuation, executeProperties);
@@ -179,7 +181,9 @@ public class LuceneIndexQueryPlan extends RecordQueryIndexPlan {
      * The partial record contains the suggestion in the field where it is indexed from, and the grouping keys if there are any.
      */
     @VisibleForTesting
-    public static IndexKeyValueToPartialRecord getToPartialRecord(@Nonnull Index index, @Nonnull RecordType recordType) {
+    public static IndexKeyValueToPartialRecord getToPartialRecord(@Nonnull Index index,
+                                                                  @Nonnull RecordType recordType,
+                                                                  @Nonnull IndexScanType scanType) {
         final IndexKeyValueToPartialRecord.Builder builder = IndexKeyValueToPartialRecord.newBuilder(recordType);
 
         KeyExpression root = index.getRootExpression();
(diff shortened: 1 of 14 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -957,12 +1074,17 @@ private RecordType addIndexAndSaveRecordForAutoComplete(@Nonnull FDBRecordContex
 
     private void assertDocumentPartialRecordFromIndexEntry(@Nonnull RecordType recordType, @Nonnull IndexEntry indexEntry,
                                                            @Nonnull String expectedSuggestion, @Nonnull String fieldName) {
-        Descriptors.Descriptor recordDescriptor = recordType.getDescriptor();
-
-        Message message = LuceneIndexQueryPlan.getToPartialRecord(indexEntry.getIndex(), recordType)
-                .toRecord(recordDescriptor, indexEntry);
+        Descriptors.Descriptor recordDescriptor = TestRecordsTextProto.SimpleDocument.getDescriptor();
+        IndexKeyValueToPartialRecord toPartialRecord = LuceneIndexQueryPlan.getToPartialRecord(SIMPLE_TEXT_WITH_AUTO_COMPLETE, recordType, IndexScanType.BY_LUCENE_AUTO_COMPLETE);
```
MMcM, line 1078: I think this may be a merge skew: it's ignoring the passed in `recordType` and the passed in `indexEntry`'s index and using fixed values for both.

It may be this is why some auto-complete tests are failing in CI.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,234 @@
...
+                .stream()
+                // Sort the suggested words from large to small by score then by frequency then by the field.
+                .sorted(comparing((Suggestion s) -> s.suggestWord.score).reversed()
+                                .thenComparing(comparing((Suggestion s) -> s.suggestWord.freq).reversed())
+                                .thenComparing(s -> s.suggestWord.string))
+                .limit(limit)
+                // Map the words from suggestions to index entries.
+                .map(suggestion -> new IndexEntry(
+                        state.index,
+                        Tuple.from(groupingKey.isEmpty() ? null : groupingKey.get(0).toString(), suggestion.indexField, suggestion.suggestWord.string),
```
MMcM, line 199: `groupingKey` can be anything; I don't see that it has to just be one-long or a string. Can just use `addAll`, I think.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,234 @@
...
+                    && groupingKey.size() == 0
+                    && Arrays.stream(fieldNames).noneMatch(name -> name.equals(fieldAndWord[0]))) {
+                throw new RecordCoreException("Invalid field name in Lucene index query")
+                        .addLogInfo(LogMessageKeys.FIELD_NAME, fieldAndWord[0])
+                        .addLogInfo(LogMessageKeys.INDEX_FIELDS, fieldNames);
+            }
+            fields = List.of(fieldAndWord[0]);
+            wordToSpellCheck = fieldAndWord[1];
+        } else {
+            // TODO in a grouped document this doesn't work because the field names all include the grouping key
```
MMcM, line 103: Do they? Doesn't the grouping key select a whole different Lucene directory?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -161,28 +163,28 @@ public static LuceneIndexQueryPlan merge(LuceneIndexQueryPlan plan1, LuceneIndex
...
-        if (recordTypes.size() != 1) {
-            throw new RecordCoreException("No lucene index should span multiple record types")
-                    .addLogInfo(LogMessageKeys.INDEX_NAME, indexName);
-        }
-
-        if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE) {
+        if (scanType == IndexScanType.BY_LUCENE_AUTO_COMPLETE || scanType == IndexScanType.BY_LUCENE_SPELLCHECK) {
+            if (recordTypes.size() != 1) {
+                throw new RecordCoreException("No lucene index should span multiple record types");
+            }
```
tian-yizuo, line 169: any reason to move this check inside the if block for auto_complete and spell_check. I guess this check can be applied to ordinary lucene search as well?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
tian-yizuo: I am not sure I understand this comment. Does the grouping key necessarily be part of the Lucene field name?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,234 @@
...
+                                        @Nonnull final IndexMaintainerState state,
+                                        @Nullable Tuple groupingKey,
+                                        final String[] fieldNames) {
+        if (value.contains(":")) {
+            String[] fieldAndWord = value.split(":", 2);
+            // only check the validity of the field if the documents are not grouped.
+            if (groupingKey != null
+                    && groupingKey.size() == 0
+                    && Arrays.stream(fieldNames).noneMatch(name -> name.equals(fieldAndWord[0]))) {
+                throw new RecordCoreException("Invalid field name in Lucene index query")
```
tian-yizuo, line 96: As I understand whether the index has a grouping key does not necessarily indicate whether the Lucene field name is a concatenated one. A nested field could also lead to a concatenated field name but it might not have grouping key right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,234 @@
...
+                // Sort the suggested words from large to small by score then by frequency then by the field.
+                .sorted(comparing((Suggestion s) -> s.suggestWord.score).reversed()
+                                .thenComparing(comparing((Suggestion s) -> s.suggestWord.freq).reversed())
+                                .thenComparing(s -> s.suggestWord.string))
+                .limit(limit)
+                // Map the words from suggestions to index entries.
+                .map(suggestion -> new IndexEntry(
+                        state.index,
+                        Tuple.from(groupingKey.isEmpty() ? null : groupingKey.get(0).toString(), suggestion.indexField, suggestion.suggestWord.string),
+                        Tuple.from(suggestion.indexField)))
```
tian-yizuo, line 200: Any reason the ```suggestion.indexField``` needs to occur in both the key tuple and value tuple of ```IndexEntry```?

sophiabuell: This was how we originally set it before incorporating your changes. I can remove it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
sophiabuell: leftover from a commit merge.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,234 @@
...
+
+    public LuceneSpellcheckRecordCursor(@Nonnull final String value,
+                                        @Nonnull final Executor executor,
+                                        final ScanProperties scanProperties,
+                                        @Nonnull final IndexMaintainerState state,
+                                        @Nullable Tuple groupingKey,
+                                        final String[] fieldNames) {
+        if (value.contains(":")) {
+            String[] fieldAndWord = value.split(":", 2);
+            // only check the validity of the field if the documents are not grouped.
```
MMcM, line 92: Why? An old comment about an obsolete reason has been removed. Is there still some other reason?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
MMcM: And that concatenation should be reflected in `fieldNames`, against which it is compared, right?

MMcM (changes requested):

MMcM (approved):

tian-yizuo (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
```diff
@@ -0,0 +1,231 @@
...
+                    throw new RecordCoreException("Spellcheck suggestions lookup failure", e);
+                }
+            }
+            return currentPosition < spellcheckSuggestions.size() ? spellcheckSuggestions.get(currentPosition) : null;
+        }, executor);
+        return spellcheckResult.thenApply(r -> {
+            if (r == null) {
+                return RecordCursorResult.exhausted();
+            } else {
+                return RecordCursorResult.withNextValue(r, continuationHelper(spellcheckSuggestions.get(currentPosition++)));
```
tian-yizuo, line 130: how about adding trace logging for the returned result?

tian-yizuo (changes requested):

tian-yizuo (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneSpellcheckRecordCursor.java
MMcM: This seems to still be the same in the code that got merged, although it moved a few lines away.

sophiabuell: yes. If this is empty then addAll becomes a two item tuple which messes up our expected count for the returned indexEntry. I tried a few different things but this made the most sense.

MMcM: Wouldn't something like
```
groupingKey.addAll(Tuple.from(suggestion.indexField, suggestion.suggestWord.string))
```
be more correct? Or do I misunderstand the requirements?

sophiabuell: If the tuple is empty then it only has those two elements instead of the 3 it has if there’s a grouping key. The grouping key is first and if we have a variance of 2 or 3 elements in the tuple I assumed it would be harder to incorporate. It also maintains consistency with Yizuo’s auto complete.

MMcM: Hmm, okay. This gets cleaned up with the scan bounds changes, which should eliminate all this tuple packing and unpacking.

Issue #1506, opened 2022-01-14
Reconsider IndexMaintainer.scan signature


MMcM:
The exact interpretation of `TupleRange range` varies according to the `IndexScanType scanType`. Yet `TupleRange` is fairly inflexible. Some of the recent full-text addition which navigate data that isn't naturally laid out as ordered ranges require some inelegant hacks.

Since the planner constructs a plan element that contains / constructs the scan type and scan ranges very specifically, there isn't really a need for something so restrictive.

Some ideas that have been proposed so far:
* Give `IndexScanType` a type parameter: `IndexScanType<T>` means `T range`.
* Collapse the two and use type dispatching instead of `IndexScanType` equality. Have only a base class / interface for the new, single argument.

Some care is needed to be as compatible as possible. I believe that compatibility is mandatory for existing calling code, since there is some of this outside just planner-generated plans and even those plans might be in-flight during an upgrade. But only nice to have for index maintainer implementations.

MMcM:
As a motivating case, #1238 proposes more complex Lucene queries made of pieces and a comment notes that `org.apache.lucene.search.Query` might be a more natural way to express this. Rather than trying how to squeeze one of those into a `Tuple`, the scan parameter for such a search could be the query itself.

PR #1516, opened 2022-02-02
Reconsider IndexMaintainer.scan signature

MMcM:
This introduces the following new classes for representing the "bounds" of an index scan:
* `IndexScanParameters`, the interface for what is stored in the plan
* `IndexScanBounds`, the interface for what results from binding parameters from some store to that
* `IndexScanComparisons`, the implementation of the first as `IndexScanType` + `ScanComparisons`
* `IndexScanRange`, the bound form of that with `TupleRange`

This is not exactly the approach suggested in #1506, based on some initial discussions of the proposal. The principle is that given a recognized `IndexScanType` constant, one can downcast the bounds to a known implementation.

For compatibility, all the `MAINTAINED` signatures are retained. Furthermore, an index maintainer that does not need to deal with non-range bound does not need to do any casting; that is done by the compatibility method. A number of tests using the `INTERNAL` APIs are affected; this branch does not introduce a lot of abstraction to make that less the case. Care is also taken to see that plan hashes do not change for the "ordinary" cases.

The first commit does not change any index implementation to use other implementations.

The second commit updates the time window leaderboard index to use a custom scan bounds to pass the time window to give some sense of the change.

This example is still somewhat uninteresting, in that it does not involve interface implementation from outside the normal classes, but only subclasses. That will come when the related issue for Lucene is done, since its currency isn't tuple ranges at all. But as I am thinking that will require a different set of reviewers, I am not including it here. Since that is the main motivation for the change, I could probably prepare a draft PR based on this to give a sense of what it will look like, if that is desired before approving this.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -96,6 +96,20 @@ public Subspace getSecondarySubspace() {
...
+     * @param scanBounds the {@link IndexScanBounds bounds} of the scan to perform
+     * @param continuation any continuation from a previous scan invocation
+     * @param scanProperties skip, limit and other properties of the scan
+     * @return a cursor over index entries in the given range
+     */
+    @Nonnull
+    public RecordCursor<IndexEntry> scan(@Nonnull IndexScanBounds scanBounds,
+                                         @Nullable byte[] continuation,
+                                         @Nonnull ScanProperties scanProperties) {
+        return scan(scanBounds.getScanType(), ((IndexScanRange)scanBounds).getScanRange(), continuation, scanProperties);
```
alecgrieser, line 110: Maybe a `ClassCastException` is fine, but it feels like there should be a more descriptive error if a `scanBounds` is given which is not an `IndexScanRange`. At the very least, getting the index name as one of the exception keys is probably a good idea.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanComparisons.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Nonnull
+    @Override
+    public IndexScanRange bind(@Nonnull FDBRecordStoreBase<?> store, @Nonnull EvaluationContext context) {
+        return new IndexScanRange(scanType, scanComparisons.toTupleRange(store, context));
+    }
+
+    @Override
+    public int planHash(@Nonnull PlanHashKind hashKind) {
+        return scanType.planHash(hashKind) + scanComparisons.planHash(hashKind);
```
alecgrieser, line 83: This feels like the hashes should be combined with something like:

```java
PlanHashable.planHash(hashKind, scanType, scanComparisons);
```

Unless this is needed for plan hash stability purposes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanParameters.java
```diff
@@ -0,0 +1,76 @@
...
+    IndexScanType getScanType();
+
+    /**
+     * Get the bound form of the index scan for use by the maintainer.
+     * @param store store against which the scan will be performed
+     * @param context query parameters for the scan
+     * @return the index scan bounds
+     */
+    @Nonnull
+    IndexScanBounds bind(@Nonnull FDBRecordStoreBase<?> store, @Nonnull EvaluationContext context);
```
alecgrieser, line 54: I think this method's purpose confused me a bit, but I think I understand this correctly:

1. An `IndexScanParameters` can have unbound parameters and is therefore used for planning whereas 
2. An `IndexScanBounds` should bind those parameters to concrete values so that it can be actual executed

And so this `bind` method is intended to translate one to the other by performing that evaluation. Is that the idea?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/TimeWindowLeaderboardIndexMaintainer.java
```diff
@@ -155,12 +157,18 @@ protected void saveSubDirectory(@Nonnull TimeWindowLeaderboardSubDirectory subdi
         state.transaction.set(extraSubspace.pack(SUB_DIRECTORY_PREFIX.addAll(subdirectory.getGroup())), subdirectory.toProto().toByteArray());
     }
 
+    @Override
+    public RecordCursor<IndexEntry> scan(@Nonnull IndexScanType scanType, @Nonnull TupleRange range,
+                                         @Nullable byte[] continuation, @Nonnull ScanProperties scanProperties) {
+        return scan(new IndexScanRange(scanType, range), continuation, scanProperties);
```
alecgrieser, line 163: Should this extract the leaderboard type and timestamp from `range` if this is called with a `BY_TIME_WINDOW` scan type (and then pass a `TimeWindowScanRange` instead of an `IndexScanRange`)? Otherwise, I think this will break callers who call `scan` on time window leaderboard indexes directly (instead of, if there is one, some higher level abstraction, like a query plan).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/TimeWindowScanComparisons.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Nonnull
+    @Override
+    public TimeWindowScanRange bind(@Nonnull final FDBRecordStoreBase<?> store, @Nonnull final EvaluationContext context) {
+        return new TimeWindowScanRange(timeWindow.getLeaderboardType(context), timeWindow.getLeaderboardTimestamp(context), super.bind(store, context).getScanRange());
+    }
+    
+    @Override
+    public int planHash(@Nonnull final PlanHashKind hashKind) {
+        return super.planHash(hashKind) + timeWindow.planHash(hashKind);
```
alecgrieser, line 56: This also seems like it should be `PlanHashable.planHash` in some way instead of simple addition. I guess there's also `PlanHashable.combineHashes`, if the idea is that `PlanHashable.planHash` would cause infinite recursion.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1380,57 +1384,45 @@ private ScoredPlan planText(@Nonnull CandidateScan candidateScan,
...
-                                     @Nullable IndexScanType scanType,
-                                     @Nullable ScanComparisons scanComparisons,
-                                     boolean strictlySorted) {
-        if (scanComparisons == null) {
-            scanComparisons = ScanComparisons.EMPTY;
-        }
         RecordQueryPlan plan;
         Set<String> possibleTypes;
         if (candidateScan.index == null) {
+            final ScanComparisons scanComparisons = ((IndexScanComparisons)scanParameters).getComparisons();
```
alecgrieser, line 1392: Should the `planScan` method take an `IndesScanComparisons` in its signature to prevent this from being called with the wrong type of `IndexScanParameters`? Or will this change to handle more exotic scan parameter types as more are added?

alecgrieser (changes requested):
I think this seems like a good approach overall, effectively abstracting the bounds of a scan. 

One place where it seems like it might be improved might be to somehow link the `IndexScanType` and the type of the `IndexScanBounds`, so that, for example, `BY_VALUE` index scans always got an `IndexScanComparisons` and `BY_TIME_WINDOW` index scans always got a `TimeWindowScanComparisons`. But I don't think the Java type system is quite strong enough to express that idea (or at least I can't think of one).

MMcM:
> One place where it seems like it might be improved might be to somehow link the IndexScanType and the type of the IndexScanBounds, so that, for example, BY_VALUE index scans always got an IndexScanComparisons and BY_TIME_WINDOW index scans always got a TimeWindowScanComparisons. But I don't think the Java type system is quite strong enough to express that idea (or at least I can't think of one).

The suggestion in the original issue called for something like `IndexScanType<P extends IndexScanParameters, B extends IndexScanBounds>`. This class would then have a constructor `(String name, Class<P> parametersClass, Class<B> boundsClass)`. It could implement narrowing methods using the `cast` methods on those classes.

The pattern would then be:
```
if (scanType == IndexScanType.BY_VALUE) {
    IndexScanRange scanRange = IndexScanType.BY_VALUE.narrow(scanParameters.bind());
}
if (scanType == IndexScanType.BY_TIME_WINDOW) {
  TimeWindowScanRange = IndexScanType.BY_TIME_WINDOW.narrow(scanParameters.bind());
}
```

It could give better errors than `ClassCastException`, but otherwise isn't actually much different than what is here. The need to use a concrete instance replaces the need for an explicit cast.

To maintain the same degree of compatibility, the `IndexScanType` class might need to inherit from its parameterized form, `IndexScanTypeP<>`, binding `<IndexScanComparisons, IndexScanRange>`.

This still isn't sufficient to get the better error message in `scan(IndexScanBounds)` where it calls `scan(IndexScanRange)`. The same checks would need to be performed (or perhaps a `static` method from `IndexScanType` used).

All of which seemed like a lot of additional mechanism for very little gain. I am willing to reconsider if it doesn't seem that way to others.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanComparisons.java
MMcM: These are needed for stability.
```
-                return indexName.hashCode() + scanType.planHash(hashKind) + comparisons.planHash(hashKind) + (reverse ? 1 : 0);
+                return indexName.hashCode() + scanParameters.planHash(hashKind) + (reverse ? 1 : 0);
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanBounds.java
```diff
@@ -0,0 +1,40 @@
...
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.IndexScanType;
+
+import javax.annotation.Nonnull;
+
+/**
+ * The bounds for an index scan.
+ * @see IndexMaintainer#scan
+ */
+@API(API.Status.MAINTAINED)
```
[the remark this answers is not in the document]
MMcM: Yes, `UNSTABLE` is probably a better choice for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanBounds.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.IndexScanType;
+
+import javax.annotation.Nonnull;
+
+/**
+ * The bounds for an index scan.
```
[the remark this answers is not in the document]
MMcM: I added another few sentences.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -288,17 +287,24 @@ public void logPlanStructure(StoreTimer timer) {
...
         }
     }
 
+    @Nonnull
+    @Override
+    public ScanComparisons getComparisons() {
+        if (scanParameters instanceof IndexScanComparisons) {
+            return ((IndexScanComparisons)scanParameters).getComparisons();
+        } else {
+            return ScanComparisons.EMPTY;
```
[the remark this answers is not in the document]
MMcM: I can have it throw an error for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScanParameters.java
MMcM: I added another sentence explaining this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/TimeWindowLeaderboardIndexMaintainer.java
MMcM: I guess that is best for a little while, at least, for compatibility. I added back the old code under conditions for that case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
MMcM: Since both times it is called with one I guess there is no reason not to do that for now.

alecgrieser (dismissed):

MMcM:
retest this please

alecgrieser (approved):

PR #1522, opened 2022-02-04
Support custom additional synonyms

arnaud-lacurie:
Additional synonyms were provided in the SOLR format. For ease of use, I converted the previous wn list to the SOLR format as we do not use the additional wordnet fields anyway.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
```diff
@@ -110,7 +110,11 @@
     /**
      * The name of the Lucene analyzer to use with a {@link IndexTypes#TEXT} index.
      */
-    public static final String TEXT_ANALYZER_NAME_OPTION = "testAnalyzerName";
+    public static final String TEXT_ANALYZER_NAME_OPTION = "textAnalyzerName";
```
alecgrieser, line 113: In theory, changing this value could break a meta-data that has already been created and that used the name with a typo. Given that this is only being used on an `EXPERIMENTAL` index type, we could fix this typo and move on, but we could theoretically add a new (deprecated) index option with the old name and check both in the maintainer if we really wanted to. I'm not sure it's worth it in this case.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
```diff
@@ -0,0 +1,96 @@
...
+        registry = initRegistry();
+    }
+
+    @Nonnull
+    public static SynonymAnalyzerRegistry instance() {
+        return INSTANCE;
+    }
+
+    @Nonnull
+    private static Map<String, SynonymAnalyzer> initRegistry() {
```
alecgrieser, line 60: I noticed that this implementation doesn't use the `ServiceLoader` like some of our other registries. Is the idea that we'll use a `ServiceLoader` to initialize the registry at some point in the future (but not now)?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -33,30 +35,61 @@
...
+import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
+import java.util.Objects;
 
 /**
  * The analyzer for index with synonym enabled.
  * Only for in fly analysis during query time.
  * The synonyms are not indexed to disk. So no index rebuilding is needed if the wordnet file gets updated.
  */
 public class SynonymAnalyzer extends StopwordAnalyzerBase {
```
alecgrieser, line 56: Are `SynonymAnalyzer`s thread safe (at least for reading from)? I ask just because if they're going to be cached in the registry, then they might be used by multiple threads.

I also wonder if this might be worth separating into an abstract `SynonymAnalyzer` class with abstract methods for getting the stop words and/or synonym stream, and then putting the "default" stuff (like the file names and stop word list) in a `DefaultSynonymAnalyzer` class. That way, it's clear to other implementors what they need to do to produce their own analyzers.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Map;
+
+/**
+ * Registry for {@link SynonymAnalyzer}s.
+ */
+public class SynonymAnalyzerRegistryImpl implements SynonymAnalyzerRegistry {
+    @Nonnull
+    private static final Logger LOGGER = LoggerFactory.getLogger(SynonymAnalyzerRegistryImpl.class);
+
+    private static final CharArraySet STOP_WORDS = EnglishAnalyzer.ENGLISH_STOP_WORDS_SET;
```
alecgrieser, line 43: It seems a little weird to include the stop word list in the registry rather than putting it in the (default) `SynonymAnalyzer`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("PMD.CompareObjectsWithEquals")
+    public synchronized void register(@Nonnull final SynonymAnalyzer analyzer) {
+        String name = analyzer.getName();
+        SynonymAnalyzer oldAnalyzer = registry.putIfAbsent(name, analyzer);
+        // If there was an already registered and the old analyzer isn't the same as
+        // the new one, throw an error
+        if (oldAnalyzer != null && oldAnalyzer != analyzer) {
```
alecgrieser, line 84: Should this be `!oldAnalyzer.equals(analyzer)`, or are you using pointer-equality deliberately?

I don't know if analyzers implement the `equals` method (or if they should), or if this defaults to pointer equality anyway.

It does seem like if this is using pointer-equality deliberately, we should say why in a comment

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -653,17 +675,47 @@ void scanWithSynonymIndex() {
...
+        // 'synonym', 'nonsynonym'
+        try (InputStream is1 = SynonymAnalyzer.openDefaultFile();
+                InputStream is2 = SynonymAnalyzer.openFile("test.txt");
+                InputStream is = new SequenceInputStream(is1, is2)) {
+            SynonymAnalyzer analyzer = new SynonymAnalyzer(
+                    EnglishAnalyzer.ENGLISH_STOP_WORDS_SET,
+                    COMBINED_SYNONYM_SETS,
+                    new InputStreamReader(is, StandardCharsets.UTF_8)
+            );
+            SynonymAnalyzerRegistryImpl.instance().register(analyzer);
```
alecgrieser, line 705: It might be a good idea to add an `@AfterEach` teardown method that resets the registry to avoid this causing problems in other tests. In theory, if we ever parallelize test execution, we'll also need to worry more about thread safety of the registry itself, but maybe that's a task for another day.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
arnaud-lacurie: Agreed, as this is experimental, I think we can simply fix the typo.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
arnaud-lacurie: I was wondering about that. 
Given that we may have hundreds of SynonymAnalyzer which only differ in their constructor parameters, is `ServiceLoader` a good way to approach that? It feels weird to me to define a class for `USEnglishSynonymAnalyzer`, `UKEnglishSynonymAnalyzer`, `DutchSynonymAnalyzer`, etc.
Is there a better way to do this?

MMcM: You could imagine a `ServiceLoader` loadable class that was responsible for contributing more than one analyzer, by loading a JSON file of their locale+location or searching a known file system directory for specific filename patterns or something like that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
tian-yizuo: @arnaud-lacurie thank you for catching this typo!

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
tian-yizuo: If we want to eventually support more analyzers by using a ```ServiceLoader``` or some similar thing, how about just loading this default analyzer by that loader?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
tian-yizuo: @alecgrieser if I understand this PR correctly, only the ```synonymMap``` could potentially be shared by different threads and it is OK because it is ```final```, and threads using the same analyzer are supposed to use same map. @arnaud-lacurie please correct me if I am wrong.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
tian-yizuo: If we want to eventually support more analyzers by using a ServiceLoader or some similar thing, how about just loading this default analyzer by that loader?

tian-yizuo: If we want to eventually support more analyzers by using a ServiceLoader or some similar thing, how about just loading this default analyzer by that loader?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
arnaud-lacurie: @alecgrieser, I looked more into it and I think you are right. `SynonymMaps` are thread safe, but there is no guarantee that `Analyzers` are. I introduced a `SynonymMap` registry instead.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
arnaud-lacurie: This section of the code was removed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzerRegistryImpl.java
arnaud-lacurie: Done :+1:

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Gets the synonym analyzer of the given set of synonyms. If <code>name</code> is <code>null</code>,
+     * it returns an analyzer over the default set of wordnet english synonyms
+     *
+     * @param name the name of the synonym set
+     * @return the analyzer over the given set of synonyms
+     * @throws MetaDataException if no such tokenizer exists
+     */
+    @Nonnull
+    SynonymMap getSynonymMap(@Nonnull String name);
```
ScottDugas, line 45: The javadoc says that `null` is allowed, but this has `@Nonnull`. I don't have a strong preference between using `@Nullable` here, or requiring they pass in a name.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Registers a new synonym analyzer in this registry. Throws an error if the analyzer over that set is
+     * already present
+     *
+     *
+     * @param name
+     * @param synonymInputStream new map to register
+     * @throws RecordCoreArgumentException if there is that synonym set is already registered.
+     */
+    void register(final String name, @Nonnull InputStream synonymInputStream);
```
ScottDugas, line 56: Would it be better to have this take a `SynonymMap` directly, allowing users to not have it in an input stream in solr format.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    @Nonnull
+    SynonymMap getSynonymMap(@Nonnull String name);
+
+    /**
+     * Registers a new synonym analyzer in this registry. Throws an error if the analyzer over that set is
+     * already present
+     *
+     *
+     * @param name
```
ScottDugas, line 52:
```suggestion
     * @param name the name of synonym map
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+     * @throws RecordCoreArgumentException if there is that synonym set is already registered.
+     */
+    void register(final String name, @Nonnull InputStream synonymInputStream);
+
+    /**
+     * Clears the registry and reloads analyzers from the classpath. This is intended
+     * mainly for testing purposes (to avoid having one test add an analyzer to the
+     * registry that another test cannot override).
+     */
+    void reset();
```
ScottDugas, line 63: It seems like it would be better have the tests register different names, and then use the different names in the tests, rather than exposing a `reset` method.

ScottDugas: That being said, there is `@VisibleForTesting`. And if something is `@VisibleForTesting`, it should probably also be marked as `@API(API.Status.INTERNAL)`, so that if/when this becomes non-experimental, it will already have the right annotation on this method

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistryImpl.java
```diff
@@ -0,0 +1,122 @@
...
+                    final StandardTokenizer src = new StandardTokenizer();
+                    TokenStream tok = new LowerCaseFilter(src);
+                    tok = new FlattenGraphFilter(tok);
+                    return new TokenStreamComponents(src, tok);
+                }
+            });
+            parser.parse(synonymStream);
+            return parser.build();
+        } catch (IOException | ParseException ex) {
+            throw new RecordCoreException("Failed to parse wordnet for synonym analyzer", ex);
```
ScottDugas, line 119: Is this exception message sufficiently general, i.e. should it be about "wordnet"?

ScottDugas: `buildSynonymMap` should probably take the name, and add it as log info, otherwise, users with more than one synonym map, won't know which one failed to parse.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -653,17 +680,42 @@ void scanWithSynonymIndex() {
...
+    }
+
+    @Test
+    void scanWithCombinedSetsSynonymIndex() throws IOException {
+        // The COMBINED_SYNONYM_SETS adds this extra line to our synonym set:
+        // 'synonym', 'nonsynonym'
+        try (InputStream is1 = new USEnglishSynonymMapConfig().getSynonymInputStream();
+             InputStream is2 = SynonymMapConfig.openFile("test.txt");
+             InputStream is = new SequenceInputStream(is1, is2)) {
+            SynonymMapRegistryImpl.instance().register(COMBINED_SYNONYM_SETS, is);
```
ScottDugas, line 705: Rather than call `register` explicitly here, would it make more sense to use the service loader, adding a new `@AutoService(SynonymMapConfig.class)` nested in the test that adds these so they are always there. This would be more in line with the way `IndexMaintainerFactory` works.

This also allows us to make `SynonymMapRegistry` immutable (removing both the `register` and `reset` methods)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapConfig.java
```diff
@@ -0,0 +1,51 @@
...
+     * Get the name of the synonymMap to build
+     * @return the name
+     */
+    String getName();
+
+    /**
+     * Get solr format input stream to build the synonymMap
+     * @return the input stream
+     */
+    InputStream getSynonymInputStream();
```
ScottDugas, line 40: It seems strange to me that this enforces using a specific storage format of the synonyms, rather than having the interface that gets registered return a `SynonymMap` directly.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -121,7 +113,11 @@ public Analyzer getIndexAnalyzer(@Nonnull Index index) {
         @Nonnull
         @Override
         public Analyzer getQueryAnalyzer(@Nonnull Index index, @Nonnull Analyzer indexAnalyzer) {
-            return new SynonymAnalyzer(StandardAnalyzer.STOP_WORDS_SET);
+            String name = index.getOption(IndexOptions.TEXT_SYNONYM_SET_NAME_OPTION);
+            if (name == null) {
+                name = USEnglishSynonymMapConfig.CONFIG_NAME;
+            }
+            return new SynonymAnalyzer(EnglishAnalyzer.ENGLISH_STOP_WORDS_SET, name);
```
ScottDugas, line 120: This seems weird & problematic. Perhaps better as a followup, but it seems like synonyms and stop words should be coupled, and registered together. Or at least, both being specified by index options. For example, this would not be the right behavior, if you are indexing non-english, you might get English stop words but, e.g. French synonyms.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -50,13 +50,26 @@
...
+        return name;
+    }
 
     @Nullable
-    private SynonymMap cachedSynonymMap = null;
+    public CharArraySet getStopwords() {
+        return stopwords;
+    }
 
     public SynonymAnalyzer(@Nullable CharArraySet stopwords) {
```
ScottDugas, line 66: Is this needed anymore? Is there a need, in general, to have a public constructor?

ScottDugas (changes requested):
I also reviewed d64d808 but it was refreshing as I was reviewing

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
arnaud-lacurie: I agree, we'll need to add the stop words to the config once we start supporting other languages.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
arnaud-lacurie: I am not sure why this allows us to remove the reset method?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistryImpl.java
arnaud-lacurie: :+1:

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
arnaud-lacurie: :+1:

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
arnaud-lacurie: 👍

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
arnaud-lacurie: I removed this method as per your recommendation below

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
arnaud-lacurie: Fixed

ScottDugas (dismissed):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
alecgrieser: Cool, cool, sounds good

> if I understand this PR correctly, only the synonymMap could potentially be shared by different threads and it is OK because it is `final`

For the record, I think `final` isn't enough here, because a `final` map can still be mutated and, importantly here, its elements can still have mutable internal state. The only thing `final` guarantees is that you can't change the map pointer to point to something else, which is better than nothing, but not as much as maybe we'd like.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapConfig.java
alecgrieser: I think this is probably fine for now, but I'm also a little hesitant about using an `InputStream` here, as it effectively means that the solr file format becomes a part of our API, rather than using something more neutral. This class isn't _technically_ marked as `EXPERIMENTAL` (though maybe we should fix that), but since it's all part of this experimental lucene index stuff, if we come up with a better way of expressing it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -108,26 +82,8 @@ protected TokenStream normalize(String fieldName, TokenStream in) {
...
-                    tok = new FlattenGraphFilter(tok);
-                    return new TokenStreamComponents(src, tok);
-                }
-            });
-            parser.parse(synonymStream);
-            return parser.build();
-        } catch (IOException | ParseException ex) {
-            throw new RecordCoreException("Failed to parse wordnet for synonym analyzer", ex);
-        }
+    private org.apache.lucene.analysis.synonym.SynonymMap getSynonymMap() {
```
alecgrieser, line 81: Does this class need to be fully specified (rather than `import`ed)?

alecgrieser (dismissed):
I think there are a few minor/touch-up things in this PR that could go in later. But otherwise, LGTM

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapConfig.java
arnaud-lacurie: I marked the class as EXPERIMENTAL for now until we figure this out.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
arnaud-lacurie: Fixed

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistryImpl.java
```diff
@@ -76,33 +68,17 @@ public SynonymMap getSynonymMap(@Nonnull final String name) {
...
-    public void reset() {
-        registry = initRegistry();
-    }
-
     @Nonnull
     private static Map<String, SynonymMap> initRegistry() {
         final Map<String, SynonymMap> registry = new HashMap<>();
         for (SynonymMapConfig config : ServiceLoader.load(SynonymMapConfig.class)) {
-            registry.put(config.getName(), buildSynonymMap(new InputStreamReader(config.getSynonymInputStream(), StandardCharsets.UTF_8)));
+            registry.put(config.getName(), buildSynonymMap(config));
```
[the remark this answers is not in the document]
arnaud-lacurie: I added a warning.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -121,7 +103,11 @@ public Analyzer getIndexAnalyzer(@Nonnull Index index) {
         @Nonnull
         @Override
         public Analyzer getQueryAnalyzer(@Nonnull Index index, @Nonnull Analyzer indexAnalyzer) {
```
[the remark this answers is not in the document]
arnaud-lacurie: As far as I understood, the synonym analyser only kicks in during query time, yes.

alecgrieser (dismissed):

tian-yizuo (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymMapRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+import com.apple.foundationdb.record.metadata.MetaDataException;
+import org.apache.lucene.analysis.synonym.SynonymMap;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Registry for {@link SynonymAnalyzer}s.
+ */
+@API(API.Status.EXPERIMENTAL)
+public interface SynonymMapRegistry {
```
ScottDugas, line 33: Since this introduces a new registry, it probably deserves a release note.

ScottDugas (changes requested):

docs/ReleaseNotes.md
```diff
@@ -24,7 +24,8 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Support custom additional synonyms. This introduces a new SynonymMapRegistry.
```
ScottDugas, line 27: This should have started as an issue, in which case this should be referencing the issue. It's kind of moot at this point though.

ScottDugas (approved):

Issue #1238, opened 2021-04-16
Lucene plans should support simple comparisons on indexed fields and fielded search


MMcM:
Right now, we support `LuceneQueryComponent` with a full-fledged Lucene query string.

In recent PR comments, it was noted that at some point having more than one of those with `AND` or `OR` could do some string manipulation to build the full query (rather than leaving it for afterwards).

In the same place, I believe that many cases of `FieldWithComparison` can also be supported, provided the given field is present in the indexed documents.

For symmetry, I think we could have a search string that comes constrained to a field.

Finally, I think a helper class like `Query` for `LuceneQuery` with static methods would be consistent with what we have elsewhere.

So,
```
Query.and(
  Query.field("year").lessThan(1900),
  Query.field("category").equalsValue("fiction"),
  LuceneQuery.field(Query.field("title"), "city"),
  LuceneQuery.search("\"hope spring\"~5"))
```

MMcM:
Also, since the comparison may be against a parameter, delayed evaluation is needed to construct the actual Lucene search.

PR #1530, opened 2022-02-13
Some clean up of Lucene query planning

### LINKED ISSUE #1517: Adjust Lucene index to use new scan bounds
Once #1516 is complete, update Lucene index support to more naturally represent the various scan cases.

### LINKED ISSUE #1528: Lucene index entry grouping key consistency
I believe that the standard pattern is that the grouping key appears as a left prefix, even though that means that things are not always in the same ordinary position as the group size changes.

The auto-complete index unnecessarily serializes the grouping key in every suggestion. I can just add it back in the scan.

The spell-check index tries to fix the size and assumes something about the data type of the grouping keys.

### LINKED ISSUE #1529: Lucene specific code should be in the lucene module
The hardest one is comparisons, which may require #1517.

MMcM:
Includes:
* Resolves #1517: Adjust Lucene index to use new scan bounds
* Fixes #1528: Lucene index entry grouping key consistency
* Resolves #1529: Lucene specific code should be in the lucene module

The work is split into several commits, each of which is consistent. However, for instance, the last issue isn't complete until the refactoring from the first removes the dependency from core.

The structure of the planner is still substantially the same. There is no particular provision for planning with the Cascades planner here. Individual field predicates expressed using non-Lucene query components are still planned by string manipulation and later parsing, although the framework for changing that is here. In other words, this pull request does not yet directly address #1238. The hope is that such would work now be more localized.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -144,54 +135,24 @@ public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull
...
-        if (scanType.equals(IndexScanType.BY_LUCENE_SPELLCHECK)) {
+        if (scanType.equals(LuceneScanTypes.BY_LUCENE_SPELL_CHECK)) {
             if (continuation != null) {
                 throw new RecordCoreArgumentException("Spellcheck does not currently support continuation scanning");
             }
-            return new LuceneSpellcheckRecordCursor(range.getLow().getString(0), executor,
-                    scanProperties, state, Tuple.fromStream(range.getLow().stream().skip(1)), fieldNames);
+            LuceneScanSpellCheck scanSpellcheck = (LuceneScanSpellCheck)scanBounds;
+            return new LuceneSpellCheckRecordCursor(scanSpellcheck.getFields(), scanSpellcheck.getWord(),
+                    executor, scanProperties, state, scanSpellcheck.getGroupKey());
```
sophiabuell, line 149: I see that we removed the field names from the arguments. Those are different than the default normalized field names provided by the component. I forget exactly how but I remember some issues with Liverpool when using normalized names and the nesting that we do etc.

MMcM: The field names are still there. They just moved to the front, because the various cursors had needlessly drifted apart in their signature. And they come in as a result of parameter binding, not by doing metadata operations here or parsing inside the cursor.

sophiabuell (approved):

tian-yizuo (approved):

tian-yizuo (changes requested):
Looks some of the comments by Sonar make sense, like the unused method parameter "scanType" of ```QueryPlanUtils.getCoveringIndexEntryToPartialRecordFunction```. Could you resolve it?

MMcM:
> Could you resolve it?

Sure. It's green now. I fixed the ones that made the check fail, plus the ones that were easy. There are still some remaining, as you can see, but they are like "finish TODO" or impossible in Java, such as refactoring `enum`s to share code.

tian-yizuo (approved):

PR #1533, opened 2022-02-19
A few bug fix and optimizations for Lucene search

tian-yizuo:
This PR is to resolve these issues:
1, Use 1.0 as NoCFSRatio for Lucene segments merging policy, to make sure it always use compound file.
2, Configure a InforSteam to publish log messages to output teams to loggers.
3, Cache the suggester into ```FDBRecordContext``` similarly as the ```indexWriter``` for ordinary search, so it won't be closed until the commit of context. 
4, Fix the partial record build to skip a field if its value is ```null```.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
@@ -82,6 +82,7 @@ public class IndexWriterCommitCheckAsync implements FDBRecordContext.CommitCheck
             }
         });
         indexWriterConfig.setCodec(new LuceneOptimizedCodec(Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
+        indexWriterConfig.setInfoStream(new LuceneLoggerInfoStream(LOGGER));
         this.indexWriter = new IndexWriter(directoryCommitCheckAsync.getDirectory(), indexWriterConfig);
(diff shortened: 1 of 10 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
```diff
@@ -82,6 +83,7 @@ protected void doMerge(final IndexWriter writer, final MergePolicy.OneMerge merg
             }
         });
         indexWriterConfig.setCodec(new LuceneOptimizedCodec(Lucene50StoredFieldsFormat.Mode.BEST_COMPRESSION));
+        indexWriterConfig.setInfoStream(new PrintStream(new LoggerOutStream(LOGGER), false));
```
MMcM, line 86: Why not just have a `LoggerInfoStream` and use the other overload `IndexWriterConfig.setInfoStream(InfoStream infoStream)`?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/IndexWriterCommitCheckAsync.java
tian-yizuo: Yeah that's a good point. Re-implemented it

MMcM (approved):

sophiabuell (approved):

Issue #1547, opened 2022-03-08
Index scrubber - return the count of bad entries found


jjezra:
Let the functions `scrubDanglingIndexEntries` and `scrubMissingIndexEntries` return the count of bad items found instead of void.

alecgrieser:
This was addressed by #1548, but this issue wasn't associated with it at the time. Closing it out now

PR #1548, opened 2022-03-08
Index scrubber - return the count of bad entries found (#1547)

jjezra:
Let the functions scrubDanglingIndexEntries and scrubMissingIndexEntries return the count of bad items found instead of void

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -24,7 +24,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** Index scrubber - return the count of bad entries found [(Issue #1547)](https://github.com/FoundationDB/fdb-record-layer/issues/1547)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -24,7 +24,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Index scrubber - return the count of bad entries found [(Issue #1547)](https://github.com/FoundationDB/fdb-record-layer/issues/1547)
+* **Feature** Index scrubber - return the count of bad entries found [(Issue #1547)](https://github.com/FoundationDB/fdb-record-layer/issues/1547)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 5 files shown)


docs/ReleaseNotes.md
```diff
@@ -24,7 +24,7 @@ This version of the Record Layer changes the Java source and target compatibilit
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Performance** Index scrubber - return the count of bad entries found [(Issue #1547)](https://github.com/FoundationDB/fdb-record-layer/issues/1547)
```
alecgrieser, line 27: Is this performance? I would think this is more of a `Feature`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -71,15 +71,18 @@
...
     @Nonnull private final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy;
+    @Nonnull private final AtomicLong missingCount;
     private long scanCounter = 0;
     private int logWarningCounter;
 
     public IndexingScrubMissing(@Nonnull final IndexingCommon common,
                                 @Nonnull final OnlineIndexer.IndexingPolicy policy,
-                                @Nonnull final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy) {
+                                @Nonnull final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy,
+                                @Nonnull final AtomicLong missingCOunt) {
```
alecgrieser, line 81:
```suggestion
                                @Nonnull final AtomicLong missingCount) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
```diff
@@ -71,15 +71,18 @@
...
         super(common, policy, true);
         this.scrubbingPolicy = scrubbingPolicy;
         this.logWarningCounter = scrubbingPolicy.getLogWarningsLimit();
+        this.missingCount = missingCOunt;
```
alecgrieser, line 85:
```suggestion
        this.missingCount = missingCount;
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
```diff
@@ -72,15 +72,18 @@
...
     @Nonnull private final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy;
+    @Nonnull private final AtomicLong danglingCount;
     private long scanCounter = 0;
     private int logWarningCounter;
 
     public IndexingScrubDangling(@Nonnull final IndexingCommon common,
                                  @Nonnull final OnlineIndexer.IndexingPolicy policy,
-                                 @Nonnull final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy) {
+                                 @Nonnull final OnlineIndexScrubber.ScrubbingPolicy scrubbingPolicy,
+                                 @Nonnull final AtomicLong danglingCount) {
```
alecgrieser, line 82: What's the use case for supplying the dangling count as a constructor parameter instead of initializing the `danglingCount` to a `new AtomicLong(0)`? It looks like all the methods I could see always passed in 0, unless you're anticipating something like invoking the scrubber multiple times and pre-seeding it with some non-zero count

alecgrieser (changes requested):

docs/ReleaseNotes.md
jjezra: 👍 +1

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
jjezra: 👍

jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubMissing.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
jjezra: The `AtomicLong` object is defined within the `scrubDanglingIndexEntries/scrubMissingIndexEntries` functions and, IIUC, does not accumulate between calls. The reason it is defined in the object's constructor and not in the scrubbing function is that the general purpose indexing/scrubbing function `buildIndexAsync`, which is implemented by `indexingBase`, has no knowledge of this scrubber-specific counter.

alecgrieser: Why not move the instantiation from the `scrubFooIdexEntries` methods to the constructors (and remove the parameters entirely)?

jjezra: How should it transfer the value back to the `scrubFooIndexEntries` methods?

alecgrieser: Oh, I see. You're relying on that value being updated as a side effect of the scrubbing. I'd missed that and assumed it was now part of the scrub return signature.

That feels like something we'll want to clean up in the future (and allow the scrubbing methods to return a more interesting response object--you could imagine getting a list of index violations, for example, perhaps truncated to just the first few), but maybe it's okay for now

alecgrieser (approved):

Issue #1551, opened 2022-03-09
Allow scrubbing of non-"value" indexes


jjezra:

alecgrieser:
Resolved in #1552.

PR #1552, opened 2022-03-09
Allow scrubbing of non-"value" indexes (#1551)

jjezra:
Add declareValueIndex to ScrubbingPolicy's builder.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
@@ -135,23 +135,23 @@ public class OnlineIndexScrubber implements AutoCloseable {
         private final int logWarningsLimit;
         private final boolean allowRepair;
         private final long entriesScanLimit;
-        private final boolean declaredValueIndex;
+        private final boolean ignoreIndexTypeCheck;
 
         public ScrubbingPolicy(int logWarningsLimit, boolean allowRepair, long entriesScanLimit,
-                               boolean declaredValueIndex) {
+                               boolean ignoreIndexTypeCheck) {
 
             this.logWarningsLimit = logWarningsLimit;
             this.allowRepair = allowRepair;
             this.entriesScanLimit = entriesScanLimit;
-            this.declaredValueIndex = declaredValueIndex;
+            this.ignoreIndexTypeCheck = ignoreIndexTypeCheck;
         }
 
         boolean allowRepair() {
             return allowRepair;
         }
 
-        boolean isValueIndex() {
-            return declaredValueIndex;
+        boolean ignoreIndexTypeCheck() {
+            return ignoreIndexTypeCheck;
         }
 
         long getEntriesScanLimit() {
@@ -184,7 +184,7 @@ public class OnlineIndexScrubber implements AutoCloseable {
             int logWarningsLimit = 1000;
             boolean allowRepair = true;
             long entriesScanLimit = 0;
-            boolean declaredValueIndex = false;
+            boolean ignoreIndexTypeCheck = false;
 
             protected Builder() {
             }
@@ -229,20 +229,20 @@ public class OnlineIndexScrubber implements AutoCloseable {
             }
 
             /**
-             * Declare that the scrubbed index is a value index, even if it's type's name is different.
+             * Declare that the index to be scrubbed is valid for scrubbing, regardless of its type's name.
              *
-             * Scrubbing dangling index entries only makes sense for value indexes, where the index entry points back to
-             * the record.
-             * Typically, this function is called to allow scrubbing of an index of a user-defined type. If called, it
-             * is the caller's responsibility to verify that the scrubbed index matches the "value" type criteria.
+             * Scrubbing dangling index entries only makes sense for index entries that point back to the record that had
+             * generated them.
+             * Typically, this function is called to allow scrubbing of an index with a user-defined index type. If called, it
+             * is the caller's responsibility to verify that the scrubbed index matches the one-to-one index-records criteria.
              */
-            public Builder declareValueIndex() {
-                declaredValueIndex = true;
+            public Builder ignoreIndexTypeCheck() {
+                ignoreIndexTypeCheck = true;
                 return this;
             }
 
             public ScrubbingPolicy build() {
-                return new ScrubbingPolicy(logWarningsLimit, allowRepair, entriesScanLimit, declaredValueIndex);
+                return new ScrubbingPolicy(logWarningsLimit, allowRepair, entriesScanLimit, ignoreIndexTypeCheck);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -221,8 +228,21 @@ public Builder setEntriesScanLimit(long entriesScanLimit) {
                 return this;
             }
 
+            /**
+             * Declare that the scrubbed index is a value index, even if it's type's name is different.
```
alecgrieser, line 232:
```suggestion
             * Declare that the scrubbed index is a value index, even if its type's name is different.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -221,8 +228,21 @@ public Builder setEntriesScanLimit(long entriesScanLimit) {
...
+             *
+             * Scrubbing dangling index entries only makes sense for value indexes, where the index entry points back to
+             * the record.
+             * Typically, this function is called to allow scrubbing of an index of a user-defined type. If called, it
+             * is the caller's responsibility to verify that the scrubbed index matches the "value" type criteria.
```
alecgrieser, line 237: It's probably worth clarifying what the "value" criteria are so that callers can know whether it's safe for them to set this flag

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -221,8 +228,21 @@ public Builder setEntriesScanLimit(long entriesScanLimit) {
...
+             */
+            public Builder declareValueIndex() {
```
alecgrieser, line 239: Rather than `declareValueIndex` or `isValueIndex`, it might be better to call this something like `ignoreIndexTypeCheck`. This is a little bit more forward looking, in that it anticipates a future when we might have more than just `VALUE` indexes allowed by default (there's little reason, for example, that we couldn't add `VERSION` indexes today, for example), and I think it also encapsulates a little better what the affect of setting this flag is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -221,8 +228,21 @@ public Builder setEntriesScanLimit(long entriesScanLimit) {
                 return this;
             }
 
+            /**
+             * Declare that the scrubbed index is a value index, even if it's type's name is different.
+             *
+             * Scrubbing dangling index entries only makes sense for value indexes, where the index entry points back to
+             * the record.
+             * Typically, this function is called to allow scrubbing of an index of a user-defined type. If called, it
```
alecgrieser, line 236:
```suggestion
             * Typically, this function is called to allow scrubbing of an index of a user-defined index type. If called, it
```

This may be using "index" one too many times, but I think "index of a user-defined type" is ambiguous in that someone might interpret that to mean "index defined on a user-defined record type" (which is, well, essentially any index)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Oops. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Changed to: 
If called, it is the caller's responsibility to verify that the scrubbed index matches the one-to-one index-records criteria.
Would that do?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Good point! 
I've eliminated the references to "value" index type in this Javadoc. Trying again..

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
alecgrieser: Hm, I think it maybe could use a little more wordsmithing. I think the actual requirement is:

1. For the dangling scrubber job, every index entry needs to contain the primary of the record that generated it so that we can detect if that record is present
2. For the missing entry scrubber, the index key for the record needs to be present in the index

And then there's maybe room to clarify what exactly is meant by the "index key" (e.g., does it get generated by evaluating the index expression against the record or is there some other way for the index maintainer to generate the index keys) or how the primary key gets extracted from each entry. I think this might be what you mean by "one-to-one", but that's also a little vague

jjezra: 1. (did you mean "primary key of the record"?) 
2. I'm afraid that do not understand what you mean.. (shouldn't the index maintainer of any type, if produces any index entry to update, always provide its key?)

alecgrieser: 1. Ah, yes, that should be primary _key_ of the record
2. Not necessarily. Part of the problem here is that there are a few different "keys" that are relevant to this, but I'm going to use "FDB key" to mean "byte array that is the key of a key-value pair in FDB" and "index key" to mean "the result of calling `.getKey()` on an `IndexEntry` object". In a value index, those two things are very closely related: for each index entry returned by `filteredIndexEntries`, there is one FDB key per index entry, and it is derived by combining the index key with the record's primary key in [`FDBRecordStoreBase::indexEntryKey`](https://github.com/foundationdb/fdb-record-layer/blob/2eadf38f86b5f58f806852c40bea821135feeba5/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java#L1046). That's the behavior that the scrubber assumes as part of its validation, but index maintainers are free to do whatever they want, and some have wildly different relationships between FDB and index keys. For example, the atomic mutation indexes have one FDB key written per _group_ (i.e., the first few columns of the index key), and text and lucena indexes explode text data (extracted from a single index key) to many different FDB keys.

jjezra: ✅ updated

alecgrieser (approved):

PR #1553, opened 2022-03-10
Introduce combined NGRAM / SYNONYM index

arnaud-lacurie:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/ngram/NgramAndSynonymAnalyzerFactory.java
```diff
@@ -0,0 +1,72 @@
...
+    @SuppressWarnings("deprecation")
+    @Nonnull
+    @Override
+    public Analyzer getIndexAnalyzer(@Nonnull Index index) {
+        return new NgramAnalyzer.NgramAnalyzerFactory().getIndexAnalyzer(index);
+    }
+
+    @Nonnull
+    @Override
+    public Analyzer getQueryAnalyzer(@Nonnull Index index, @Nonnull Analyzer indexAnalyzer) {
```
tian-yizuo, line 51: I am afraid the main concern with using ngram analyzer for indexing and synonym analyzer for query is, the synonyms of substrings of indexed terms could be considered as matches to return, which is unexpected. This is just the main problem we need to resolve, to support analyzer for combination of ngram and synonym, right?

PR #1555, opened 2022-03-14
Resolves #720 Auto-throttling runner

### LINKED ISSUE #720: Auto-throttling runner
Today we have the `FDBDatabaseRunner` which can be used to iteratively perform work and to repeatedly run a transaction to success, however for certain tasks the success or failure of a given transaction depends on certain factors, such as a rate of conflicts with other transactions or the overall size of the transaction.   The `OnlineIndexer` is an example of such code...in reaction to a high rate of conflicts it will dynamically reduce the amount of records it is attempting to process in a transaction, and as success rates improve, it will dynamically re-increase this rate.  Similarly, it (or at least should be able to) commit the current transaction when it hits a specific size target to avoid saturating the server.

It would be possible to encapsulate the sort of logic in the `OnlineIndexer` into a new kind of generic runner that can help to dynamically adjust the amount of work being done in a transaction in a similar fashion so that anyone needing to do long running tasks without worrying about these details.

ScottDugas:
This introduces a new `LimitedRunner` and `TransactionalLimitedRunner` which replace the old `IndexingThrottle` in a generic way.
There are some properties of `FDBDatabaseRunner` that used to be changeable in the middle of an index run, but maintaining that complexity did not seem worth it.

There are a lot of commits here, so we should do a squash merge, but figure it's worth keeping the original commits around in case anyone needs to look at them later.

Resolves #720


docs/ReleaseNotes.md
```diff
@@ -16,6 +16,9 @@ The FoundationDB Java binding dependency has been updated to 7.1 with this relea
 
 This release also updates downstream dependency versions. Most notably, the proto3 artifacts now require Protobuf version 3.20.1.
 
+`OnlineIndexer` no longer supports changing the initial delay & max delay used when retrying, in the middle of an ongoing
```
ScottDugas, line 19: I'm not sure what the correct way to do this is, but I feel like it should be mentioned, perhaps it warrants a new minor build?

ScottDugas:
There are way too many conflicts for me to pick up from this branch. If I take another attempt at it, it would probably be best to start from scratch.

PR #1557, opened 2022-03-14
Resolves #1556: Upgrade Lucene to 8.11.1 from 7.7.2

### LINKED ISSUE #1556: Upgrade Lucene to 8.11.1 from 7.7.2
There have been many performance fixes over the years...

jleach4:
This is an upgrade of the record layer to Lucene 8.11.1.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
@@ -3,7 +3,7 @@
  *
  * This source file is part of the FoundationDB open source project
  *
- * Copyright 2015-2021 Apple Inc. and the FoundationDB project authors
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,8 +20,8 @@
 
 package com.apple.foundationdb.record.lucene.codec;
 
-
 import org.apache.lucene.codecs.CodecUtil;
+import org.apache.lucene.codecs.CompoundDirectory;
 import org.apache.lucene.index.CorruptIndexException;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.SegmentInfo;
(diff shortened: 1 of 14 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -653,10 +657,24 @@ public void misMatchQueryShouldReturnNoResult(boolean shouldDeferFetch) throws E
...
     public void threadedLuceneScanDoesntBreakPlannerAndSearch() throws Exception {
-        initializeFlat();
         CountingThreadFactory threadFactory = new CountingThreadFactory();
         executorService = Executors.newFixedThreadPool(10, threadFactory);
-        try (FDBRecordContext context = openContext()) {
+        initializeFlat();
+        FDBRecordContext context = null;
+        try {
+            for (int i = 0; i < 200; i++) {
+                context = openContext();
```
alecgrieser, line 666: This is just test code, but this introduces a memory leak. Each time a transaction is created, the previous one isn't closed before the `context` variable is changed to point to the new one.

I think you can get around this by restructuring the loop to be something like:

```java
for (int i = 0; i < 200; i++) {
        // insert a record with random data
        commit(context);
    }
}
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
```diff
@@ -138,61 +136,29 @@ public IndexInput openInput(String name, IOContext context) throws IOException {
...
     /** Returns the length of a file in the directory.
      * @throws IOException if the file does not exist */
     @Override
     public long fileLength(String name) throws IOException {
         ensureOpen();
         FileEntry e = entries.get(IndexFileNames.stripSegmentName(name));
-        if (e == null) {
+        if (e == null)
             throw new FileNotFoundException(name);
-        }
```
alecgrieser, line 169: We generally require braces, even for single line blocks like this. There's a checkstyle check, `NeedBraces`, that's supposed to catch that. It looks like the build failed (during javadoc generation) before it got to `fdb-record-layer-lucene:checkstyleMain`, so it may or may not

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -514,19 +518,19 @@ public void delayFetchOnOrOfLuceneFiltersGivesUnion(boolean shouldDeferFetch) th
             Matcher<RecordQueryPlan> matcher = union(
                     indexScan(allOf(indexScan("Complex$text_index"),
                             indexScanType(IndexScanType.BY_LUCENE_FULL_TEXT),
-                            bounds(hasTupleString("[[(civil blood makes civil hands unclean)],[(civil blood makes civil hands unclean)]]")))),
+                            bounds(hasTupleString("[[(\"civil blood makes civil hands unclean\")],[(\"civil blood makes civil hands unclean\")]]")))),
```
alecgrieser, line 521: Does this imply that the bounds to scans of the `LuceneIndexMaintainer` need to change? I believe that's tested in the `LuceneIndexTest`, so it may not, but if users have to start supplying quotes to their tuple bounds, that's something we should at least note somewhere as something they should react to

alecgrieser (changes requested):
A few small things around Record Layer usage. Someone with more knowledge on Lucene internals should probably take a look/provide feedback on those parts that changed

tian-yizuo (commented):
LGTM basically, will approve after @alecgrieser 's comments get addressed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
jleach4: Let me try again with Aaron's changes to the downloads...

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
jleach4: Added a note saying the bounds are Lucene parsable strings...  I think the problem is that (_text1 text2_) is not the same as ("_text1 text2_") in the syntax specification.

MMcM (approved):

PR #1564, opened 2022-03-23
simplest select

normen662:
This PR contains a set of features which allows the Cascades planner to reason about and to plan actual computation/projection of values instead of merely planning an access path to data that is stored on disk:

- It utilizes the type system introduced in #1546 for `Value`s and for `RelationalExpression`s. Every query graph atom has a type. 
- `RelationalExpression::getResultValues()` is replaced by `RelationalExpression::getResultValue()` motivated by the observation that the result of a classical relational operator is a record. In order to orthogonalize record creation, records on top level should be created in the same way as nested records.
- `QueryResult` is changed to only wrap one result, not a list of results, due to the same reasoning: what used to be flowed as elements in the list held by `QueryResult` now flows as fields in a first class record.
- introduces `RecordQueryMapPlan` as a physical operator that computes a select list of a simple select expression (a non-join)
- plan and matching work to plan (implement) `RecordQueryMapPlan` properly
- integration of streaming aggregation plans
- `@DualPlannerTest` can now be configured to only run with a specified planner, `OLD`, `CASCADES`, or `BOTH`.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/rules/ImplementMapRule.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+import static com.apple.foundationdb.record.query.plan.temp.matchers.ListMatcher.exactly;
+import static com.apple.foundationdb.record.query.plan.temp.matchers.MultiMatcher.some;
+import static com.apple.foundationdb.record.query.plan.temp.matchers.QuantifierMatchers.forEachQuantifierOverPlans;
+import static com.apple.foundationdb.record.query.plan.temp.matchers.RelationalExpressionMatchers.selectExpression;
+
+/**
+ * A rule that implements a logical filter around a {@link RecordQueryPlan} as a {@link RecordQueryFilterPlan}.
```
hatyo, line 47: Documentation should be adapted I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -42,54 +42,56 @@
...
-        this.groupAggregator = groupAggregator;
+        this.streamGrouping = streamGrouping;
     }
 
     @Nonnull
     @Override
-    public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
+    public CompletableFuture<RecordCursorResult<Object>> onNext() {
         if (previousResult != null && !previousResult.hasNext()) {
             // post-done termination condition: Keep returning terminal element after inner is exhausted.
```
MMcM, line 68: Comment looks out to date.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -1725,6 +1726,23 @@ default int getPipelineSize(@Nonnull PipelineOperation pipelineOperation) {
...
+     * Execute a query.
+     * @param plan the plan to execute
+     * @param continuation continuation from a previous execution of this same plan
+     * @param evaluationContext a context for the execution of this plan
+     * @param executeProperties limits on execution
+     * @return a cursor for query results of type {@link QueryResult}
+     * @see RecordQueryPlan#execute
+     */
+    @Nonnull
+    default RecordCursor<QueryResult> executePlan(@Nonnull RecordQueryPlan plan,
```
MMcM, line 1739: The other overloads are called `executeQuery`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/Column.java
```diff
@@ -0,0 +1,78 @@
...
+
+package com.apple.foundationdb.record.query.plan.temp;
+
+import com.apple.foundationdb.record.query.predicates.Value;
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+
+/**
+ * A case class to hold a values computing a column together with information about the field of a record it supplies
```
MMcM, line 29: Java doesn't have case classes. I believe the precedent for things that will be able to be `record` -- the closest thing -- in the future it to have a TODO comment to that effect.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -42,54 +42,56 @@
  * @param <M> the record type that holds the actual data
  */
 @API(API.Status.EXPERIMENTAL)
-public class AggregateCursor<M extends Message> implements RecordCursor<QueryResult> {
+public class AggregateCursor<M extends Message> implements RecordCursor<Object> {
```
MMcM, line 45: This seems odd. Even though multi-column records are now a single value, I would not expect this to have to change.

Looked at another way, this cursor is doing two things: extracting a value from the input cursor and aggregating. I would expect it to either be `RecordCursor<QueryResult>` -> `RecordCursor<QueryResult>` if operating on result sets or `RecordCursor<Object>` -> `RecordCursor<Object>` if operating on aggregatable values, requiring the calling operator to supply the two `map`s. In the latter case, better might be `RecordCursor<M1 extends Message>` -> `RecordCursor<M2 extends Message>` with `StreamingGroup<M1, M2>`.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/rules/ImplementMapRule.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/temp/Column.java
normen662: A `Column` is just a convenience class in order to avoid writing `Pair<Field, Value>` all the time. I removed the `case class` reference in the documentation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
normen662: I changed it back to `RecordCursor<QueryResult>` so it can directly be used by the plan operator `RecordQueryStreamingAggregationPlan`. So in effect `StreamGrouping` aggregates `Value`s and therefore `Object` which is in line with what other `Value`-based operators like filter and also map, `AggregateCursor` does the extraction/repackaging from/to `QueryResult`.

MMcM (approved):

Issue #1567, opened 2022-03-25
Lucene 8.11.1 re-introduced a lot of checksum issues


jleach4:
The PostingsReader performs several checksums requiring larger amounts of data read into memory.

PR #1568, opened 2022-03-25
Lucene 8.11.1 re-introduced a lot of checksum issues #1567

jleach4:
8.11.1 re-introduced checksums that we had attempted to remove in our prior version.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
+++ b/fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
@@ -1,0 +1,2006 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.codecs.lucene84;
+
+import static org.apache.lucene.codecs.lucene84.ForUtil.BLOCK_SIZE;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.DOC_CODEC;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.MAX_SKIP_LEVELS;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.PAY_CODEC;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.POS_CODEC;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.TERMS_CODEC;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.VERSION_COMPRESSED_TERMS_DICT_IDS;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.VERSION_CURRENT;
+import static org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat.VERSION_START;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+import com.apple.foundationdb.annotation.SpotBugsSuppressWarnings;
+import org.apache.lucene.codecs.BlockTermState;
+import org.apache.lucene.codecs.CodecUtil;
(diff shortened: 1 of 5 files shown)


fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
```diff
@@ -0,0 +1,2006 @@
...
+    private static final long BASE_RAM_BYTES_USED = RamUsageEstimator.shallowSizeOfInstance(Lucene84PostingsReader.class);
+
+    private final IndexInput docIn;
+    private final IndexInput posIn;
+    private final IndexInput payIn;
+
+    private final int version;
+
+    /** Sole constructor. */
+    public LuceneOptimizedPostingsReader(SegmentReadState state) throws IOException {
```
tian-yizuo, line 76: Could you add comments to indicate that the difference between this optimized one and the ```Lucene84PostingsReader``` is the removed ```CodecUtil.retrieveChecksum```

tian-yizuo (commented):
LGTM basically, just left one minor comment.

fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
```diff
@@ -0,0 +1,2006 @@
+/*
+ * LuceneOptimizedPostingsReader.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
```
alecgrieser, line 6: If this is taken from code that was originally in Lucene and then modified, I think we need a slightly different attribution header here. I thought there was an example somewhere of a place we've done this, but I'm not actually sure that there is

fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/package-info.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Support for {@code LUCENE} indexes and queries.
```
alecgrieser, line 22: This looks like it's been taken from the `com.apple.foundationdb.record.lucene` package-info. This should probably describe the code in this subpackage specifically, including noting that the codec is adapted from the 8.11 optimized codec

alecgrieser (changes requested):
In addition to the comments left in-line, it might also be helpful to update the release notes. I don't know if the Lucene work has been using them, but I (personally, at least) find those are pretty useful when upgrading the Record Layer dependency so that it's clearer what the differences between the versions are

fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
jleach4: Could not find it..

alecgrieser (dismissed):

alecgrieser:
It looks like the build failed because checkstyle noticed that the new file has a non-standard header:

```
[ant:checkstyle] [ERROR] /codebuild/output/src566016129/src/github.com/FoundationDB/fdb-record-layer/fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java:3: Line does not match expected header line of ' *'. [Header]
```

But that's what we want in this case because we want the Lucene header, not our header.

I think we can suppress this check for this file by adding the following to `gradle/codequality/suppressions.xml`:

```xml
    <suppress checks="Header"
        files=".*[\\/\]org[\\/]apache[\\/]lucene[\\/]codecs[\\/]lucene84[\\/]LuceneOptimizedPostingsReader\.java"/>
```

In theory, we could suppress all checks (that is `checks=".*"`) if we don't want to run our style checks on this code.

tian-yizuo (approved):

tian-yizuo (dismissed):

gradle/codequality/suppressions.xml
```diff
@@ -8,4 +8,6 @@
         files=".*[\\/]protogen[\\/].*"/>
     <suppress checks="JavadocPackage"
         files=".*[\\/]test[\\/].*|.*[\\/]java-proto[23][\\/].*[\\/]metadata[\\/].*"/>
+    <suppress checks="Header"
+              files=".*[\\/\]org[\\/]apache[\\/]lucene[\\/]codecs[\\/]lucene84[\\/]LuceneOptimizedPostingsReader\.java"/>
```
alecgrieser, line 12:
```suggestion
        files=".*[\\/]org[\\/]apache[\\/]lucene[\\/]codecs[\\/]lucene84[\\/]LuceneOptimizedPostingsReader\.java"/>
```

It looks like there's an extra `/` before `org`, which is what caused the gradle check style build to fail

alecgrieser (changes requested):

alecgrieser (approved):

alecgrieser (approved):

MMcM (approved):

PR #1570, opened 2022-03-28
Lucene text field indexing configuration support

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -97,4 +99,45 @@
...
+
+    /**
+     * Determines whether the positions are also stored in the term vectors of the Lucene field.
+     */
+    public static final RecordLayerPropertyKey<Boolean> LUCENE_FULL_TEXT_WITH_TERM_VECTOR_POSITIONS = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.withTermVectorPositions", false);
+
+    /**
+     * Determines whether to omit the normalization values for the Lucene field.
+     */
+    public static final RecordLayerPropertyKey<Boolean> LUCENE_FULL_TEXT_WITH_OMIT_NORMS = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.withOmitNorms", false);
```
alecgrieser, line 121: Should these be dynamic properties? Or put another way, do any of these options affect the format of data on disk in a backwards incompatible way? Or is the idea to make these dynamic properties for now to allow for faster iteration/experimentation, but to migrate these to something else later?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -233,6 +233,19 @@ private boolean insertField(LuceneDocumentFromRecord.DocumentField field, final
...
     }
 
+    private FieldType getTextFieldType(boolean isStored) {
+        FieldType ft = new FieldType();
+        ft.setIndexOptions(LuceneRecordContextProperties.getIndexOptions(this.state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_FULL_TEXT_FIELD_INDEX_OPTIONS), LOG));
+        ft.setTokenized(true);
+        ft.setStored(isStored);
+        ft.setStoreTermVectors(this.state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_FULL_TEXT_WITH_TERM_VECTORS));
+        ft.setStoreTermVectorPositions(this.state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_FULL_TEXT_WITH_TERM_VECTOR_POSITIONS));
+        ft.setOmitNorms(this.state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_FULL_TEXT_WITH_OMIT_NORMS));
```
alecgrieser, line 243: This would seem to suggest that we're saying that all fields get the same value for these values. Is that the idea, or are there use cases for having different values for these options for different fields?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -97,4 +99,45 @@
...
+     * Determines what information is stored in the index of the Lucene field.
+     */
+    public enum LuceneFieldIndexOptions {
+        DOCS,
+        DOCS_AND_FREQS,
+        DOCS_AND_FREQS_AND_POSITIONS,
+        DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS,
+    }
+
+    public static IndexOptions getIndexOptions(LuceneFieldIndexOptions value, Logger logger) {
```
alecgrieser, line 133: I'm not sure this is the right place for this translation of the Record Layer-defined `LuceneFieldIndexOptions` enum to the Lucene-defined `IndexOptions` enum, especially because it means that the property definition class now has a method that exposes something from under Lucene. It might make more sense to put it in the maintainer (or maybe even just to mark this function as `INTERNAL`).

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
jleach4: Different values per field.  For example, we do not want to store positions for specific fields.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
jleach4: These do effect the format of data on disk.  I think the goal would be able to have these configurable...

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
```diff
@@ -238,7 +248,11 @@ public boolean equals(Object o) {
             if (!Objects.equals(value, that.value)) {
                 return false;
             }
-            return type == that.type;
+            if (type != that.type) {
+                return false;
+            }
+            return fieldConfigs == null && that.fieldConfigs == null
+                   || fieldConfigs != null && fieldConfigs.equals(that.fieldConfigs);
```
alecgrieser, line 255:
```suggestion
            return Objects.equals(fieldConfigs, that.fieldConfigs);
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
```diff
@@ -193,12 +193,14 @@ public void addField(@Nonnull T source, @Nonnull final String fieldName, @Nullab
...
         private final boolean stored;
+        private final Map<String, Object> fieldConfigs;
 
-        public DocumentField(@Nonnull String fieldName, @Nullable Object value, LuceneIndexExpressions.DocumentFieldType type, boolean stored) {
+        public DocumentField(@Nonnull String fieldName, @Nullable Object value, LuceneIndexExpressions.DocumentFieldType type, boolean stored, @Nullable Map<String, Object> fieldConfigs) {
             this.fieldName = fieldName;
             this.value = value;
             this.type = type;
             this.stored = stored;
+            this.fieldConfigs = fieldConfigs;
```
alecgrieser, line 203: I think a `null` value of `fieldConfigs` is semantically identical to the empty map (`Collections.emptyMap()`), so it might make sense to initialize the map to empty instead of `null`, and then I think you can save some null checks.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
```diff
@@ -193,12 +193,14 @@ public void addField(@Nonnull T source, @Nonnull final String fieldName, @Nullab
         private final Object value;
         private final LuceneIndexExpressions.DocumentFieldType type;
         private final boolean stored;
+        private final Map<String, Object> fieldConfigs;
```
alecgrieser, line 196: Probably should mark this as `@Nullable` (unless it's made `@Nonnull` via setting it to the empty map if null)

alecgrieser (changes requested):
I like the way this is trending. Just a few smaller comments

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
tian-yizuo: Sounds good. Use ```Collections.emptyMap()``` now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
tian-yizuo: Use Collections.emptyMap() now

alecgrieser (approved):

MMcM (approved):

PR #1573, opened 2022-03-29
Resolve #1572: "rebuilding index failed" does not include primary_key

### LINKED ISSUE #1572: "rebuilding index failed" does not include primary_key
It looks like it fails to do this because  logExceptionAsWarn doesn't look through causes, and this is a `CompletionException`

jjezra:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2846,10 +2846,12 @@ public Range getUnbuiltRange() {
     }
 
     private void logExceptionAsWarn(KeyValueLogMessage message, Throwable exception) {
-        if (exception instanceof LoggableException) {
-            LoggableException loggable = (LoggableException) exception;
-            LOGGER.warn(message.addKeysAndValues(loggable.getLogInfo()).toString(), loggable);
-        } else {
+        if (LOGGER.isWarnEnabled()) {
+            for (Throwable ex = exception;
+                        ex instanceof LoggableException;
```
ScottDugas, line 2851: I don't think this check is right.
If you have:
```
CompletionException
caused by: LoggableException
```

the first iteration of the loop will return false for `ex instanceof LoggableException` and quit.
I think you want to iterate until you find the first `LoggableException` cause (or `null`).

I guess you could also be explicit and just check the cause if it's a `CompletionException` or `ExecutionException`, but that doesn't really seem like it would buy us much.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: How about the current commit? 
Would it cause a duplication of the first (root) exception's fields in the log message? (Should the iteration start at `exception.getCause()`?)

ScottDugas (approved):

PR #1574, opened 2022-03-29
Fixes #1571: Lucene subproject calls CompletableFuture::join

### LINKED ISSUE #1571: Lucene subproject calls CompletableFuture::join
There are a few places in the Lucene subproject implementation that I could find calls to `CompletableFuture::join` or `CompletableFuture::get`. In non-test code, that's generally a bad sign for a few different reasons:

1. It doesn't run the exception wrapping code, so it can expose exceptions from the underlying system instead of, say, `RecordCoreException`s

2. Timeouts aren't controlled by the timeout handler on the database

3. Blocking-in-async detection is always skipped (which may need to be handled a little more delicately for Lucene because of the threading model in play)

We should replace those instances with calls to `FDBRecordContext::asyncToSync`

alecgrieser:
This removes the calls to `CompletableFuture::join` from non-test code within the fdb-record-layer-lucene subproject and replaces them with calls to `FDBRecordContext::asyncToSync` over the future. This also made a few code improvements that I noticed while I was editing things:

1. There were a few remaining calls to `.thenApplyAsync`, which shuffles work onto the common fork join pool, and therefore doesn't work with the Record Layer threading model
1. The `readBlock` method could be pipelined better so that it didn't block on getting the file reference, but instead composed the file-reference future and the future to read the block
1. The parameteried `FDBLuceneQueryTest` tests were hiding their names

This fixes #1571.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -24,7 +24,6 @@ import com.apple.foundationdb.KeyValue;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
-import com.apple.foundationdb.record.logging.CompletionExceptionLogHelper;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.lucene.LuceneEvents;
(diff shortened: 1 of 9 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -152,32 +149,30 @@ public FDBDirectory(@Nonnull Subspace subspace, @Nonnull FDBRecordContext contex
...
+        });
+        return Objects.requireNonNull(context.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_GET_INCREMENT, future));
     }
 
     private void increment(StoreTimer.Count counter) {
-        final FDBStoreTimer timer = context.getTimer();
-        if (timer != null) {
-            timer.increment(counter);
-        }
+        context.increment(counter);
```
alecgrieser, line 167: As a note here: the `FDBRecordContext::increment` method that this is replacing already checks to see if the timer is null (and updates the counter if the timer is there), so this change is to remove the duplication:

https://github.com/FoundationDB/fdb-record-layer/blob/21baa55b2f33f98fb1dfafaf5794f0bf2a583733/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBTransactionContext.java#L137-L141

Same for the other `increment` and `record` methods that are updated here.

tian-yizuo (approved):

alecgrieser:
Rebase addresses merge conflict

PR #1576, opened 2022-03-30
Resolves #1575: Lucene FDBDirectory could cache listAll results

### LINKED ISSUE #1575: Lucene FDBDirectory could cache listAll results
The `FDBDirectory::listAll` method reads the list of file names and file references from the database, and it then returns all file names. These values are then frequently accessed by subsequent calls, and so it would make sense to keep the full set of values cached. There's already an on-demand cache for file reference values, but we should probably just keep the full mapping of file names to file references cached anyway, especially as there are calls in the Lucene code that do things like list the files _twice_ in a row, and the second time should be satisfiable directly from the JVM.

alecgrieser:
This restructures the way the `FDBDirectory` handles caching the file name to `FDBLuceneFileReference` map. Previously, there was a cache where entries were probably filled by a call to `listAll()` that would happen when the directory was opened anyway, but now, this keeps a full cache of file names and file reference information in memory for the directory. In theory, this could get too big if there are too many files, but the calling code relies on being able to list all files anyway, so at least for now, that's a problem either way. Note that the calling code sometimes makes multiple `listAll()` calls in the same request, and so this cache can speed up calls after the first one.

This also tweaks the way that the list is read from the database in that it uses the `WANT_ALL` streaming mode instead of the (default) `ITERATOR` streaming mode. This streaming mode tries to optimize the number of round trips required to complete a range read, which means that the time to first byte can be longer but the time to the last byte can be shorter. Given that the full list is needed in this case before we can do anything useful with it, optimizing for time to last byte should be an improvement in this case.

This resolves #1575.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -21,7 +21,11 @@
 package com.apple.foundationdb.record.lucene.directory;
 
 import com.apple.foundationdb.KeyValue;
+import com.apple.foundationdb.ReadTransaction;
+import com.apple.foundationdb.StreamingMode;
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.async.AsyncIterable;
+import com.apple.foundationdb.async.AsyncUtil;
 import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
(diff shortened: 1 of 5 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -104,7 +108,22 @@
 
     private final LockFactory lockFactory;
     private final int blockSize;
-    private final Cache<String, FDBLuceneFileReference> fileReferenceCache;
+
+    /**
+     * A supplier used to load the {@link #fileReferenceCache}. The supplier allows us to memoize listing all
+     * the files from the database so that we don't have to do that more than once per {@link FDBDirectory}
```
tian-yizuo, line 114: umm is it a correct assumption that the files list is always same within a transaction lifecycle? Like would it get updated when there are some writes in the transaction and you want to read your writes? and the segments merge can also lead to updated files if I understand correctly?

alecgrieser: You're right that the file list managed by a Lucene directory can change over the course of a transaction. To handle that, the cache is updated whenever a file is added, deleted, or renamed. For example, here: https://github.com/FoundationDB/fdb-record-layer/blob/70d252bbf8075c887f1614cb7376f19c57a4d969/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java#L278

This change looks at all of the instances where the previous Guava cache (the `fileReferenceCache`) was updated, and now it makes sure that the full file cache has been loaded first, and then it uses that as the source of truth.

alecgrieser: In principle, it could be wasteful to load the full cache here rather than doing it lazily (which was the caching approach before), but the Lucene runtime was calling `listAll()` anyway (sometimes multiple times), and so in practice, I think it ends up being better to maintain a complete cache

alecgrieser:
Rebase puts the code on top of #1574, but it is otherwise the same

tian-yizuo (approved):

PR #1578, opened 2022-03-30
Resolves #1577: Lucene FDBDirectory should cache file counter

### LINKED ISSUE #1577: Lucene FDBDirectory should cache file counter
The `FDBDirectory` implementation keeps track of a counter that is used to assign files an immutable internal ID, and then file names are linked to that ID via the "meta-data subspace". The current implementation re-reads that counter each time the counter is incremented. This should be served from the FDB-maintained RYW cache, so it shouldn't need to go all the way to a remote FDB cluster, but it could still be lifted up into the JVM and stored as an `AtomicLong` on the directory.

alecgrieser:
This adds a new `AtomicLong` to the `FDBDirectory` where the file sequence counter is cached. The first time a user tries to get an increment, the value is loaded from the database and put into the cache. All subsequent accesses can use that cache, and the `AtomicLong` guarantees thread safety. Note that the FDB read-your-writes cache was already keeping a cached copy, but this removes the extra JNI hop and requirement to go through the FDB network thread in order to get a new increment.

This resolves #1577.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.lucene.directory;
 
 import com.apple.foundationdb.KeyValue;
+import com.apple.foundationdb.MutationType;
 import com.apple.foundationdb.ReadTransaction;
 import com.apple.foundationdb.StreamingMode;
 import com.apple.foundationdb.annotation.API;
@@ -124,6 +125,7 @@ public class FDBDirectory extends Directory {
      */
     private final AtomicReference<ConcurrentSkipListMap<String, FDBLuceneFileReference>> fileReferenceCache;
 
+    private final AtomicLong fileSequenceCounter;
     private final Cache<Pair<Long, Integer>, CompletableFuture<byte[]>> blockCache;
 
     private final boolean compressionEnabled;
@@ -157,9 +159,34 @@ public class FDBDirectory extends Directory {
                 .maximumSize(maximumSize)
                 .recordStats()
                 .build();
+        this.fileSequenceCounter = new AtomicLong(-1);
         this.compressionEnabled = Objects.requireNonNullElse(context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_INDEX_COMPRESSION_ENABLED), false);
         this.encryptionEnabled = Objects.requireNonNullElse(context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_INDEX_ENCRYPTION_ENABLED), false);
         this.fileReferenceMapSupplier = Suppliers.memoize(this::loadFileReferenceCacheForMemoization);
+    }
+
+    private long deserializeFileSequenceCounter(@Nullable byte[] value) {
+        return value == null ? 0L : Tuple.fromBytes(value).getLong(0);
+    }
+
+    @Nonnull
+    private byte[] serializeFileSequenceCounter(long value) {
+        return Tuple.from(value).pack();
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> loadFileSequenceCounter() {
+        long originalValue = fileSequenceCounter.get();
+        if (originalValue >= 0) {
+            // Already loaded
+            return AsyncUtil.DONE;
+        }
+        return context.ensureActive().get(sequenceSubspaceKey).thenAccept(serializedValue -> {
+            // Replace the counter value in the counter unless something else has already
+            // updated it, in which case we want to keep the cached value
+            long loadedValue = deserializeFileSequenceCounter(serializedValue);
+            fileSequenceCounter.compareAndSet(originalValue, loadedValue);
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -157,29 +159,56 @@ public FDBDirectory(@Nonnull Subspace subspace, @Nonnull FDBRecordContext contex
...
+        // Make sure the file counter is loaded from the database into the fileSequenceCounter, then
+        // use the cached value
+        context.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_GET_INCREMENT, loadFileSequenceCounter());
+        long incrementedValue = fileSequenceCounter.incrementAndGet();
+
+        // Use BYTE_MAX here so that if there are concurrent calls (which will get different increments because they
+        // are serialized around the fileSequenceCounter AtomicLong), the largest one always gets
+        // committed into the database
+        byte[] serializedValue = serializeFileSequenceCounter(incrementedValue);
+        context.ensureActive().mutate(MutationType.BYTE_MAX, sequenceSubspaceKey, serializedValue);
```
tian-yizuo, line 209: Could this mutation fail but the counter incrementation above succeed?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -157,29 +159,56 @@ public FDBDirectory(@Nonnull Subspace subspace, @Nonnull FDBRecordContext contex
...
+        return Tuple.from(value).pack();
+    }
+
+    @Nonnull
+    private CompletableFuture<Void> loadFileSequenceCounter() {
+        long originalValue = fileSequenceCounter.get();
+        if (originalValue >= 0) {
+            // Already loaded
+            return AsyncUtil.DONE;
+        }
```
tian-yizuo, line 183: what if after one thread sees a negative counter above this line, but before it reads the value from storage below, another thread updates the counter, but has not mutated the value from storage yet, and this thread loads a stale value from storage?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo, line 209: Could this mutation fail but the counter incrementation above succeed?

tian-yizuo (commented):
Some questions about the consistency between counter and storage, and the potential race condition.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
alecgrieser: I'm having a bit of a hard time following your hypothetical. Are you saying something like this:

1. Thread 1: Gets negative value from fileSequenceCounter
2. Thread 2: Gets negative value from fileSequenceCounter
3. Thread 2: Loads value from storage
4. Thread 2: Puts loaded value into fileSequenceCounter
5. Thread 2: Updates counter value in fileSequenceCounter
6. Thread 1: Loads value from storage
7. Thread 2: Updates value in storage
8. Thread 1: Puts loaded value info fileSequenceCounter

Then, in this case, yes, at step 6, thread 1 will read a value from the database that does not include the update made at step 5 by thread 2. However, note that the `fileSequenceCounter` is updated by calling [`compareAndSet`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html#compareAndSet-int-int-) on line 188, and so it will only be updated if the `fileSequenceCounter` is still negative. So, in the case outlined above, step 8 will actually not update the `fileSequenceCounter`, and so even though thread 1 read a stale value from the database, it will actually get a value from `getIncrement` based on an up-to-date value, so I don't think there's a race

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
alecgrieser: Not in a way that I think would cause a problem. The `mutate` method can fail (if, for example, the transaction has already been committed or closed), but only in ways that would, I think, not result in any data being committed based on the uncommitted data

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: yes that is what I meant. Sorry I just noticed that this stale value is to be CAS updated to the counter instead of being returned directly. Yeah then this looks good to me

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: But does this cause an incorrect value to be updated to the counter? The failed ```mutate``` won't be handled with undoing the counter update above right?

alecgrieser: I think the answer is that, yes, that could theoretically result in a value in the counter that was out of sync with the value in the underlying transaction, but my point is that that doesn't matter. This value is only used internally in the `FDBDirectory`, not exposed to the user, so the only problematic case would be if, somehow, the value in the database's counter wasn't updated, but a value was somehow used as the file ID for some file. But if there is file data committed which uses this counter, then this `mutate` line must also have succeeded. (Because the `mutate` can't fail in a way that still allows subsequent writes to succeed.)

tian-yizuo (approved):

PR #1581, opened 2022-04-01
New ExponentialDelay class, fixes #1565

### LINKED ISSUE #1565: FDBDatabaseRunnerImpl delay trends towards 0
Much like:  #1007
After enough retries FDBDatabaseRunnerImpl will only delay 0 or 1 milliseconds. Instead it should back-off exponentially, and at the end be delaying a random value less than the max delay.

ScottDugas:
This introduces a new class `ExponentialDelay` that manages a delay that backs off exponentially, along with a jitter.
This replaces two implementations of the same idea, one in `FDBDatabaseRunnerImpl` and one in `IndexingThrottle`.
This also, fixes #1565 by the fact that the new class does not have that bug, so using it in `FDBDatabaseRunnerImpl` will clean that up.

I also introduced a new `runners` package, for this, because I felt as though the root was getting too bloated. I did not move `FDBDatabaseRunnerImpl` to the new package, although that could be done (it is INTERNAL). Move `FDBDatabaseRunner` would be a breaking change.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/package-info.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/package-info.java
@@ -1,0 +1,31 @@
+/*
+ * package-info.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Classes for running blocks of asynchronous code with-or-without retries.
+ *
+ * <p>
+ *     Note: for historical/backwards compatibility reasons
+ *     {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner}
+ *     and {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunnerImpl}
+ *     are not included in this package.
+ * </p>
+ */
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/runners/ExponentialDelayTest.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class ExponentialDelayTest {
```
alecgrieser, line 43: I think these tests are good, assuming they pass reliably, but I suppose an alternative would be to include assertions on the `currentDelay` state of the delay class, though that may be a little bit too fragile (that is to say, it relies on internal implementation details instead of the API output)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/package-info.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Classes for running blocks of asynchronous code with-or-without retries.
+ *
+ * <p>
+ *     Note: for historical/backwards compatibility reasons
+ *     {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner}
+ *     and {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunnerImpl}
```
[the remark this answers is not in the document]
ScottDugas: Yes, and moving `FDBDatabaseRunnerImpl` wouldn't require bumping the version, as that is internal. Either way, I would prefer to do that in a separate change.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/runners/ExponentialDelayTest.java
ScottDugas: Yeah, I could have asserted about the trend of `currentDelay`, but that I think I would still want tests like this, that are more functional.
I increased the repetitions for each of the 3 tests by a factor of 1000 (so 5,000; 20,000; 50,000) and they all passed (and it took 4 minutes)

alecgrieser (approved):

Issue #1580, opened 2022-03-31
Lucene index maintainer should support delete where on group key


alecgrieser:
The Lucene index maintainer, if on a group key, should be able to support grouping key delete wheres. For example, if there's a grouping key on `concatenateFields("a", "b")`, it should support delete where for a delete where query bounded on either a predicate on `a` or on `a` and `b`.

This same kind of predicate is supported on the `TextIndexMaintainer` and the `RankIndexMaintainer`, so it should be straightforward to add.

PR #1584, opened 2022-04-04
Fixes #1583: Indexes with groups erroneously accept deleteRecordsWhere queries on grouped columns

### LINKED ISSUE #1583: Indexes with groups erroneously accept deleteRecordsWhere queries on grouped columns
On indexes with a grouping, like the atomic mutation indexes or the rank indexes, the `canDeleteWhere` function should return `true` only if the delete-where predicate can be satisfied by the grouping columns. For example, if you had a `SUM` index on `field("foo").groupBy(field("bar"))`, the index can allow delete where requests with predicates like `Query.field("bar").equalsValue(x)`, but it should reject queries with predicates like `Query.and(Query.field("bar").equalsValue(x), Query.field("foo").equalsValue(y))`.

However, grouping queries currently use the `QueryToKeyMatcher` to determine if the query can be satisfied, and that tries to match the entire key, not just the grouping subkey:

https://github.com/FoundationDB/fdb-record-layer/blob/71e284213866980ff200db09f5fc6612d0007e97/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java#L215

That means that indexes with groupings are accepting grouping predicates that they shouldn't be.

alecgrieser:
This adjusts the methods used by `canDeleteWhere` on indexes defined on a `GroupingKeyExpression` so that it tries to satisfy the query from the grouping sub key rather than the whole key to limit the number of accepted `deleteRecordsWhere` predicates to more closely align with that is actually possible to safely delete. This includes some regression tests that hit the error, as well as a small amount of refactoring that is to set up #1580.

This fixes #1583.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
@@ -514,7 +514,7 @@ public class QueryToKeyMatcher {
         return null;
     }
 
-    @Nullable
+    @Nonnull
     private KeyExpression extractPrefixUntilSplittable(KeyExpression wholeKeyExpression, int originalSplitPoint) {
         int adjustedSplitPoint = originalSplitPoint - 1;
         while (adjustedSplitPoint >= 0) {
@@ -524,7 +524,10 @@ public class QueryToKeyMatcher {
                 adjustedSplitPoint--;
             }
         }
-        return null;
+        // We should always be able to extract 0 columns, so this should be unreachable code
+        throw new RecordCoreException("unable to extract splittable prefix from key expression")
+                .addLogInfo(LogMessageKeys.KEY_EXPRESSION, wholeKeyExpression)
+                .addLogInfo(LogMessageKeys.COLUMN_SIZE, originalSplitPoint);
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
@@ -212,7 +212,8 @@ public class QueryToKeyMatcher {
     private Match matches(@Nonnull QueryComponent query, @Nonnull KeyExpression key,
                           @Nonnull MatchingMode matchingMode, @Nullable FilterSatisfiedMask filterMask) {
         if (key instanceof GroupingKeyExpression) {
-            return matches(query, ((GroupingKeyExpression) key).getWholeKey(), matchingMode, filterMask);
+            KeyExpression group = extractGroupingKey((GroupingKeyExpression) key, matchingMode);
+            return group == null ? Match.none() : matches(query, group, matchingMode, filterMask);
         }
         if (key instanceof KeyWithValueExpression) {
             KeyExpression onlyKey = extractKeyFromKeyWithValue((KeyWithValueExpression)key, matchingMode);
@@ -486,6 +487,17 @@ public class QueryToKeyMatcher {
         throw new KeyExpression.InvalidExpressionException("Unexpected Key Expression type " + key.getClass());
     }
 
+    private KeyExpression extractGroupingKey(GroupingKeyExpression grouping, MatchingMode matchingMode) {
+        try {
+            return grouping.getGroupingSubKey();
+        } catch (BaseKeyExpression.UnsplittableKeyExpressionException err) {
+            if (matchingMode == MatchingMode.SATISFY_QUERY) {
+                return extractPrefixUntilSplittable(grouping.getWholeKey(), grouping.getGroupingCount());
+            }
+        }
+        return null;
+    }
+
     @Nullable
     private KeyExpression extractKeyFromKeyWithValue(KeyWithValueExpression keyWithValue, MatchingMode matchingMode) {
         try {
@@ -496,14 +508,20 @@ public class QueryToKeyMatcher {
             // cover the entire query, so keep trying a more and more selective prefix of the key until
             // we no longer hit the unsplittable exception
             if (matchingMode == MatchingMode.SATISFY_QUERY) {
-                int adjustedSplitPoint = keyWithValue.getSplitPoint() - 1;
-                while (adjustedSplitPoint >= 0) {
-                    try {
-                        return keyWithValue.getInnerKey().getSubKey(0, adjustedSplitPoint);
-                    } catch (BaseKeyExpression.UnsplittableKeyExpressionException innerErr) {
-                        adjustedSplitPoint--;
-                    }
-                }
+                return extractPrefixUntilSplittable(keyWithValue.getInnerKey(), keyWithValue.getSplitPoint());
+            }
+        }
+        return null;
+    }
+
+    @Nullable
+    private KeyExpression extractPrefixUntilSplittable(KeyExpression wholeKeyExpression, int originalSplitPoint) {
+        int adjustedSplitPoint = originalSplitPoint - 1;
+        while (adjustedSplitPoint >= 0) {
+            try {
+                return wholeKeyExpression.getSubKey(0, adjustedSplitPoint);
+            } catch (BaseKeyExpression.UnsplittableKeyExpressionException innerErr) {
+                adjustedSplitPoint--;
             }
         }
         return null;
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreDeleteWhereTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreDeleteWhereTest.java
@@ -20,30 +20,42 @@
 
 package com.apple.foundationdb.record.provider.foundationdb;
 
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.FunctionNames;
 import com.apple.foundationdb.record.IndexScanType;
+import com.apple.foundationdb.record.IsolationLevel;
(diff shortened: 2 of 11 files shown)


alecgrieser:
The sonar cloud issues above are from existing test code in `TextIndexTest`. The change I made was to convert `public void` to `void` to avoid tripping the sonar rule that prefers the latter, and I feel like we should just ignore sonar for this PR on those other methods, but I'm open to other opinions

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
```diff
@@ -496,14 +508,20 @@ private KeyExpression extractKeyFromKeyWithValue(KeyWithValueExpression keyWithV
...
+                return extractPrefixUntilSplittable(keyWithValue.getInnerKey(), keyWithValue.getSplitPoint());
+            }
+        }
+        return null;
+    }
+
+    @Nullable
+    private KeyExpression extractPrefixUntilSplittable(KeyExpression wholeKeyExpression, int originalSplitPoint) {
+        int adjustedSplitPoint = originalSplitPoint - 1;
+        while (adjustedSplitPoint >= 0) {
```
MMcM, line 520: Is it actually useful to have this succeed for `0` by returning `EmptyKeyExpression`?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/QueryToKeyMatcher.java
alecgrieser: I think it never actually needs to return an `EmptyKeyExpression`, and that it could return `null` here. I added a test for the one case I could think where it might actually be needed (which is in the case where it can't pull any columns out of the `wholeKeyExpression`, but the predicate is also empty), and that also worked here (because the emptiness of the predicate is checked before the `QueryToKeyMatcher` is constructed).

I added the test, as that seems like a good regression test to have, but I also left this as is because it seems like it's easier to reason about this code if it returns an `EmptyKeyExpression` if it cannot extract any columns rather than having it return `null`. (Like, in the future, if we added a `Query.any()` component that evaluated to `true` on all inputs, then having this return `EMPTY` would be the correct thing to do here.)

MMcM: I can buy that, but should it then return `EMPTY` instead of `null` if the `try` fails? That is, do we want to distinguish the case where the key expression refuses to extras zero columns?

MMcM (dismissed):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreDeleteWhereTest.java
```diff
@@ -260,6 +562,51 @@ void testDeleteWhereSingleTypeWithFunctionIndexSplit() throws Exception {
         }
     }
 
+    @Test
+    void testDeleteWhereSingleTypeEmptyPredicate() throws Exception {
```
ScottDugas, line 566: Inspired by #1585 should there be a test where the entire key is grouped. eg:
```
new Index("MyRecord$sum_num",
                 new GroupingKeyExpression(field("header").nest(concatenateFields("num")), 1),
                 IndexTypes.SUM)
```
Which should fail if deleting by `num` or `path`, but should on fail if deleting by `null`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
```diff
@@ -1039,7 +1043,167 @@ public void saveComplexWithAggressiveConflictRanges() throws Exception {
...
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, hook);
+            recordStore.saveRecord(doc1);
+            recordStore.saveRecord(doc2);
+            commit(context);
+        }
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, hook);
+            Query.InvalidExpressionException err = assertThrows(Query.InvalidExpressionException.class,
+                    () -> recordStore.deleteRecordsWhere(SIMPLE_DOC, Query.field("text").equalsValue("foo bar")));
```
ScottDugas, line 1073: Why doesn't this work? `SimpleDocument$text` is a value index on `text`, so clearing the range on `"foo bar"` should work, right?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
```diff
@@ -1039,7 +1043,167 @@ public void saveComplexWithAggressiveConflictRanges() throws Exception {
...
+            assertThat(err.getMessage(), containsString("deleteRecordsWhere not supported by index SimpleDocument$text"));
+        }
+    }
+
+    @Test
+    void deleteWhereComplexByGroup() {
+        final ComplexDocument zeroGroupDocument = ComplexDocument.newBuilder()
```
ScottDugas, line 1080: It seems like a weak test to have only record per group, and only one index entry. It seems like it would be better to have multiple records contain the queried text.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreDeleteWhereTest.java
```diff
@@ -170,7 +185,294 @@ void testDeleteWhereGroupedCount() throws Exception {
...
+                    () -> recordStore.deleteRecordsWhere(Query.field("header").matches(Query.and(
+                            Query.field("path").equalsValue(pathToDelete),
+                            Query.field("rec_no").equalsValue(3L)))));
+
+            assertThat(err.getMessage(), containsString(String.format("deleteRecordsWhere not supported by index %s", sumIndex.getName())));
+        }
+    }
+
+    @Test
+    void testDeleteWhereRank() throws Exception {
```
ScottDugas, line 374: I think `rank` can have a groupedCount > 1, (not sure about other aggregates), it would probably be good to have a test of that.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
```diff
@@ -1039,7 +1043,167 @@ public void saveComplexWithAggressiveConflictRanges() throws Exception {
...
+            recordStore.deleteRecordsWhere(COMPLEX_DOC, Query.field("group").equalsValue(0));
+
+            // Group zero should be gone but group 1 should still have an element remaining
+            assertThat(plan.execute(recordStore, EvaluationContext.forBinding("group_value", 0)).asList().join(), empty());
+            assertThat(plan.execute(recordStore, EvaluationContext.forBinding("group_value", 1)).asList().join(), hasSize(1));
+        }
+    }
+
+    @Test
+    void deleteWhereGroupedMap() {
```
ScottDugas, line 1127: I believe text can have a grouped count > 1, it would probably be good to test querying that.

ScottDugas, line 1127: This test also seems like it could use more text, and probably more entries in the map.

ScottDugas (changes requested):
Going through this, I suspect there's a lot of coverage that is missing from `deleteRecordsWhere` that could be added, but I tried to focus my suggestions so as not to turn this PR into "Test all of deleteRecordsWhere"

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
alecgrieser: Sure, added more records here

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
alecgrieser: Sure, added more documents and more map entries here. I also added a test that tries and delete with a predicate on the map key. It can't really do that because of the primary key on the record type, and it would require more work than I think it's worth to get a separate test working that aligned correctly, but it can at least assert that the error comes from the primary key (not the index)

alecgrieser: Text indexes do support a `grouped count > 1` (or a `grouping count > 1`), though that's somewhat separate from the change at hand. This is mainly here because heretofore, I think there were _no_ tests of `deleteRecordsWhere` with the `TEXT` index, which seemed like a mistake. The way that the these two tests are set up, we have coverage now for when there is one grouping key and we issue a delete on that key, when there are no grouping keys, and when there are two grouping keys and we issue a delete on the first one.

In practice, you can't (currently) use the columns after the text column in a `TEXT` index, which is what the columns after the first grouped column are for. I also think, given that the grouped keys are excluded from the key expression in the `QueryToKeyMatcher`, I think the main things we need to make sure tested are:
1. If there are no grouping columns, the records-where-deleter always rejects the delete (which is tested by `deleteWhereSimpleDocuments`)
2. If the `QueryToKeyMatcher` finds a match in the grouping keys, that the index data actually gets deleted (which is tested by the other two tests)

ScottDugas (approved):

Issue #458, opened 2019-03-15
KeyWithValueExpression.getKeyExpression() is incorrect when there is FanOut


nschiefer:
The `getKeyExpression()` method on `KeyWithValueExpression` is supposed to produce the "key" side of the `KeyWithValueExpression`. The current implementation uses `normalizeKeyForPositions()`, which is incorrect in the case of a FanOut.

We should use `getSubKey()` instead.

PR #1585, opened 2022-04-04
Fixing bug with KeyWithValueExpression.getValueExpression()

scottfines:
This fixes a minor issue where `KeyWithValueExpression.getValueExpression()` could break if there were less than 2 elements in the value portion of the expression, and adds a test to verify the fix.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyWithValueExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyWithValueExpression.java
@@ -115,6 +115,8 @@ public class KeyWithValueExpression extends BaseKeyExpression implements KeyExpr
             List<KeyExpression> allKeys = normalizeKeyForPositions();
             if (splitPoint == allKeys.size()) {
                 valueExpression = EmptyKeyExpression.EMPTY;
+            } else if ((allKeys.size() - splitPoint) == 1) {
+                valueExpression = allKeys.get(splitPoint);
             } else {
                 valueExpression = new ThenKeyExpression(allKeys, splitPoint, allKeys.size());
             }
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
@@ -35,6 +35,7 @@ import com.apple.foundationdb.record.metadata.expressions.FunctionKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression.FanType;
+import com.apple.foundationdb.record.metadata.expressions.KeyWithValueExpression;
 import com.apple.foundationdb.record.metadata.expressions.ListKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.NestingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.QueryableKeyExpression;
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
```diff
@@ -766,6 +768,21 @@ void testSplitConcat() {
                 evaluate(splitConcat, numbers));
     }
 
+    @Test
+    void testCanProperlyGetValueExpressionFromKeyWithValue() {
+        /*
+         * Tests for a degenerate case of KeyWithValueExpressions
```
ScottDugas, line 774: Is this really a "degenerate" case of `KeyWithValueExpression`? The test appears to assert that there is one field in the value, which seems like a totally valid usage to me.
If it is, and I'm misunderstanding something about how `KeyWithValueExpression` works, perhaps another reviewer would be better.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyWithValueExpression.java
```diff
@@ -115,6 +115,8 @@ public KeyExpression getValueExpression() {
             List<KeyExpression> allKeys = normalizeKeyForPositions();
             if (splitPoint == allKeys.size()) {
                 valueExpression = EmptyKeyExpression.EMPTY;
+            } else if ((allKeys.size() - splitPoint) == 1) {
+                valueExpression = allKeys.get(splitPoint);
             } else {
                 valueExpression = new ThenKeyExpression(allKeys, splitPoint, allKeys.size());
             }
```
alecgrieser, line 122: Should this call `getInnerKey().getSubKey` like `getKeyExpression` does? I think in addition to single-column value bug added in the test, I think this also mishandles multi-column `KeyExpression`s in its children, which is what #458 was alluding to as well (though it was concerned with the key subexpression instead of the value subexpression). The `normalizeKeyForPositions` function _kind of_ handles it, but it doesn't get all of the nesteds and repeateds correct

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
```diff
@@ -766,6 +768,21 @@ void testSplitConcat() {
...
+         */
+        List<KeyExpression> fields = new ArrayList<>();
+        fields.add(recordType());
+        fields.add(Key.Expressions.field("col0"));
+        fields.add(Key.Expressions.field("col1"));
+        KeyWithValueExpression kwve = new KeyWithValueExpression(Key.Expressions.concat(fields),2);
```
alecgrieser, line 780:
```suggestion
        KeyWithValueExpression kwve = new KeyWithValueExpression(Key.Expressions.concat(fields), 2);
```

It looks like there are missing spaces on some other lines as well

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
```diff
@@ -766,6 +768,21 @@ void testSplitConcat() {
                 evaluate(splitConcat, numbers));
     }
 
+    @Test
+    void testCanProperlyGetValueExpressionFromKeyWithValue() {
```
alecgrieser, line 772: I think "properly" is a little vague here. Perhaps "single value" would be a better term

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/metadata/KeyExpressionTest.java
alecgrieser: I think having one value column is definitely a valid use case. I think you could still use "degenerate" here if you really wanted to, but I don't have too many opinions one way or the other

alecgrieser (changes requested):
I think this change should also add a release note mentioning the bug fix. Ideally, we'd also have a GitHub issue with a bug report

ScottDugas (changes requested):
Also, this should have an associated issue, and probably an update to the release notes.

Issue #1595, opened 2022-04-06
Transition SortedDocValuesField to BinaryPoint to support efficient deletes in Lucene


jleach4:
Binary Points have been optimized for deletes in the new versions of Lucene.

From the javadoc…

  /** 
   * Create a query for matching an exact {@link BytesRef} value.
   * <p><b>NOTE</b>: Such queries cannot efficiently advance to the next match,
   * which makes them slow if they are not ANDed with a selective query. As a
   * consequence, they are best used wrapped in an {@link IndexOrDocValuesQuery},
   * alongside a range query that executes on points, such as
   * {@link BinaryPoint#newExactQuery}.
   */
  public static Query newSlowExactQuery(String field, BytesRef value) {

PR #1600, opened 2022-04-07
Transition SortedDocValuesField to BinaryPoint to support efficient deletes in Lucene #1595

jleach4:
The SortedDocValuesField is known to have issues deleting large number of records and requires the index to read more data than needed during a delete.  The BinaryPoint implementation is known to be fast when deleting the records but has the requirement that the encodings have to be fixed width (yuck).

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
@@ -315,6 +315,29 @@ public class FieldKeyExpression extends BaseKeyExpression implements AtomKeyExpr
         return field.getMessageType();
     }
 
+    /**
+     * Only supporting DOUBLE, INT64, INT32, and BOOL currently.
+     *
+     * @param descriptor Record Descriptor
+     * @return boolean
+     */
+    @Override
+    public boolean isFixedWidth(Descriptors.Descriptor descriptor) {
+        Descriptors.FieldDescriptor fieldDescriptor = descriptor.findFieldByName(fieldName);
+        if (fieldDescriptor.isRepeated() || (nullStandin != null && nullStandin.toProto().equals(RecordMetaDataProto.Field.NullInterpretation.UNIQUE))) {
+            return false;
+        }
+        switch (fieldDescriptor.getJavaType()) {
+            case DOUBLE:
+            case INT:
+            case LONG:
+            case BOOLEAN:
+                return true;
+            default:
+                return false;
+        }
+    }
+
     @Nonnull
     public FanType getFanType() {
         return fanType;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
@@ -208,6 +208,18 @@ public interface KeyExpression extends PlanHashable, QueryHashable {
     }
 
     /**
+     * Method that indicates whether the keys can be stored in a fixed width.  For example, an integer or long can be
+     * stored as a fixed width byte[].
+     *
+     * This is utilized by Lucene to see if we can optimize the key lookup in the case of deletes.
+     *
+     */
+    @API(API.Status.INTERNAL)
+    default boolean isFixedWidth(Descriptors.Descriptor descriptor) {
+        return false;
+    }
+
+    /**
      * Expand this key expression into a data flow graph. The returned graph represents an adequate representation
      * of the key expression as composition of relational expressions and operators
      * ({@link com.apple.foundationdb.record.query.plan.temp.RelationalExpression}s). Note that implementors should
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -50,8 +50,11 @@ import com.apple.foundationdb.record.provider.foundationdb.properties.RecordLaye
 import com.apple.foundationdb.record.query.QueryToKeyMatcher;
 import com.apple.foundationdb.tuple.Tuple;
 import com.apple.foundationdb.tuple.TupleHelpers;
+import com.google.common.primitives.Ints;
+import com.google.common.primitives.Longs;
 import com.google.protobuf.Message;
 import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.document.BinaryPoint;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.DoublePoint;
 import org.apache.lucene.document.Field;
@@ -72,6 +75,8 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.HashSet;
 import java.util.List;
@@ -234,13 +239,18 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         return suggestionAdded;
     }
 
-    private void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
-                               byte[] primaryKey) throws IOException {
+    private <M extends Message> void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
+                               @Nonnull FDBIndexableRecord<M> record) throws IOException {
+        byte[] primaryKey = record.getPrimaryKey().pack();
         final IndexWriter newWriter = getOrCreateIndexWriter(state, indexAnalyzer, executor, groupingKey);
         BytesRef ref = new BytesRef(primaryKey);
         Document document = new Document();
         document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
-        document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
+        if (record.getRecordType().getPrimaryKey().isFixedWidth(record.getRecordType().getDescriptor())) {
+            document.add(new BinaryPoint(PRIMARY_KEY_SEARCH_NAME, retrievePoints(record.getPrimaryKey())));
+        } else {
+            document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
+        }
         final AnalyzingInfixSuggester suggester = autoCompleteEnabled ? getSuggester(groupingKey) : null;
         boolean suggestionAdded = false;
         for (LuceneDocumentFromRecord.DocumentField field : fields) {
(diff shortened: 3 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
```diff
@@ -315,6 +315,29 @@ public SplitKeyExpression split(int splitSize) {
...
+    /**
+     * Only supporting DOUBLE, INT64, INT32, and BOOL currently.
+     *
+     * @param descriptor Record Descriptor
+     * @return boolean
+     */
+    @Override
+    public boolean isFixedWidth(Descriptors.Descriptor descriptor) {
+        Descriptors.FieldDescriptor fieldDescriptor = descriptor.findFieldByName(fieldName);
+        if (fieldDescriptor.isRepeated() || fieldDescriptor.isOptional()) {
```
MMcM, line 327: `isOptional` is probably too strong, as it is often used for things that are, in fact, `NOT NULL`, due to the exact behavior of Protobuf. (Reasonable advice is "always use `optional`.") A better check is probably using the `nullStandin`, which will need to be `NULL_UNIQUE` for nulls to work in a primary key, I think.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -234,13 +239,18 @@ private boolean insertField(LuceneDocumentFromRecord.DocumentField field, final
...
-                               byte[] primaryKey) throws IOException {
+    private <M extends Message> void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
+                               @Nonnull FDBIndexableRecord<M> record) throws IOException {
+        byte[] primaryKey = record.getPrimaryKey().pack();
         final IndexWriter newWriter = getOrCreateIndexWriter(state, indexAnalyzer, executor, groupingKey);
         BytesRef ref = new BytesRef(primaryKey);
         Document document = new Document();
         document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
-        document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
+        if (record.getRecordType().getPrimaryKey().isFixedWidth(record.getRecordType().getDescriptor())) {
```
MMcM, line 249: `RecordType.getPrimaryKey` will include the grouping keys, if they are also part of the primary key. But it's okay for those the be `null`. It's also slightly more efficient to only encode the primary key relative to the index, that is, without any grouping key prefix(es).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
```diff
@@ -207,6 +207,18 @@ default boolean hasLosslessNormalization() {
...
+    /**
+     * Method that indicates whether the keys can be stored in a fixed width.  For example, an integer or long can be
+     * stored as a fixed width byte[].
+     *
+     * This is utilized by Lucene to see if we can optimize the key lookup in the case of deletes.
+     *
+     */
+    @API(API.Status.INTERNAL)
+    default boolean isFixedWidth(Descriptors.Descriptor descriptor) {
+        return false;
```
MMcM, line 219: You will also encounter things like a primary key [component] in a nested "header-like" field.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
```diff
@@ -207,6 +207,18 @@ default boolean hasLosslessNormalization() {
...
 
+    /**
+     * Method that indicates whether the keys can be stored in a fixed width.  For example, an integer or long can be
+     * stored as a fixed width byte[].
+     *
+     * This is utilized by Lucene to see if we can optimize the key lookup in the case of deletes.
+     *
+     */
+    @API(API.Status.INTERNAL)
+    default boolean isFixedWidth(Descriptors.Descriptor descriptor) {
```
alecgrieser, line 218: I don't know if we ever solved this problem, but there's sort of a tension whenever we add a feature that involves walking over the `KeyExpression` tree between adding one function with `instanceof` checks over the various key expression types versus adding a new expression to the `KeyExpression` interface. We definitely used to favor the latter, until we had too many, and now we've kind of gone towards the former. Given that this is only going to be useful for the `Lucene` index and its return value is very much baked in with the Lucene implementation, I think we probably want to go with a function in the Lucene code that knows how to handle the various key expression types

(Fwiw, I think I also experimented with a proper "visitor" class, with "visitX" methods for each of the various `KeyExpression` types, and it worked okay and gave you proper compiler errors if you forgot a type, though it has problems with extending to user-defined `KeyExpression` types, though we also discourage those, now, as `FunctionKeyExpression` gets you most of what you want.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
alecgrieser: Concatenated keys are the other big miss here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
alecgrieser: I think you need `nullStandIn` to be `NOT_NULL` for this to work here, unless there's a representation of `null` that we can think of that can be encoded in the same number of bytes as the non-`null` field values (or can exclude the nullable field somehow)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
```diff
@@ -315,6 +315,29 @@ public SplitKeyExpression split(int splitSize) {
...
+            return false;
+        }
+        switch (fieldDescriptor.getType()) {
```
alecgrieser, line 330: I think you could probably switch on `getJavaType` here and then this would also work for a few more numeric types, like `SFIXED32`, `SINT64`, etc.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -91,6 +91,7 @@
 import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
 import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.COMPLEX_DOC;
 import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.SIMPLE_DOC;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.SIMPLE_DOCUMENT_WITH_TEXT_PRIMARY_KEY;
```
alecgrieser, line 94: I guess it's good to have a regression test with a definitely non-fixed point primary key, but do any of the tests actually use the new format? I think they'd all fail the "is fixed width" check because they all have optional fields as their primary key

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -252,18 +262,45 @@ private void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField>
...
+                throw new UnsupportedEncodingException("Primary Key of Type X Not Supported in Lucene");
+            }
+        }
+        return lookup;
+    }
+
+    private <M extends Message> void deleteDocument(Tuple groupingKey, @Nonnull FDBIndexableRecord<M> oldRecord) throws IOException {
+        Tuple primaryKey = oldRecord.getPrimaryKey();
         final IndexWriter oldWriter = getOrCreateIndexWriter(state, indexAnalyzer, executor, groupingKey);
-        Query query = SortedDocValuesField.newSlowExactQuery(PRIMARY_KEY_SEARCH_NAME, new BytesRef(primaryKey));
```
alecgrieser, line 257: I'm not sure why this is using `newSlowExactQuery`, but should these use `IndexWriter.deleteDocument(Term)` instead with a `Term` containing the primary key field as the field name and the primary key as the value? I don't know how that compares in performance to the binary point approach, but it would seem to be a simpler way of deleting, and it would work for all primary keys

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
jleach4: Ok, I tried Mike's approach but I will let you guys discuss the definition of a null field in the system that can be used in a primary key.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
jleach4: Agreed, changing...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/FieldKeyExpression.java
alecgrieser: I think the salient thing for this code is that the requirement is that the field is not null in order for field to be encoded using the fixed point encoder. I think using `NULL_UNIQUE` will continue to create a primary key with a `null` field in the tuple, which is why I don't think it helps us here. In theory, this could become a runtime check, maybe, though we probably want to try harder to not include the grouping keys in the Lucene-encoded primary key (if they overlap), in that case. (See: `primaryKeyComponentPositions` on the index. I think something like only only including the primary key components where the `primaryKeyComponentPosition < 0 || >= groupingCount` would suffice.)

MMcM: My thought was that only `NULL_UNIQUE` really works as a component in a unique index or primary key. While it is true that you could have a `null` there with `NULL`, for example, there would have to be something else that arranged for ignoring that to _still_ be unique. Such as the other columns by themselves being unique in that case. I don't see any particular downside to requiring that primary key fields that might hold nulls must have that (and adding a runtime check as you suggest). Then we could interpret field not so annotated as `NOT NULL`. Lots of databases don't allow nullable fields in primary keys at all, so having an extension to do that seems okay (to me).

Having said all that, I am happy to take a more conservative approach for now and revisit this when nullability is overhauled as part of a more robust type system.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
tian-yizuo: I guess grouping keys also need override for this probably?

tian-yizuo (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
jleach4: Term search is pretty slow if you are trying to delete 50 records for example.

PR #1602, opened 2022-04-08
Fixes #1601: Make EvaluationContextBuilder API backwards compatible

### LINKED ISSUE #1601: Make EvaluationContextBuilder API backwards compatible
In PR #1564, the maintained `EvaluationContextBuilder` API was broken by introducing a `build()` method that took a parameter. In order to avoid breaking that API for adopters, we should be able to have a default `TypeRepository` set in the builder that can then be overridden

alecgrieser:
This makes the `TypeRepository` handling in the `EvaluationContext` API more-buildery, with a default value of `TypeRepository.EMPTY_SCHEMA`. This also simplifies some logic that was creating a child builder off of a parent context and then had to remember to pass in the parent's `TypeRepository` to the new child, as that's now handled by the `Builder` itself, though it is still possible for the child to override the `TypeRepository` if it wishes. This is a breaking API change of of 3.1.248.0, though it reverts it back to something that is backwards compatible with 3.1.247.0.

This fixes #1601.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/EvaluationContextBuilder.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/EvaluationContextBuilder.java
@@ -37,12 +37,15 @@ import javax.annotation.Nullable;
 public class EvaluationContextBuilder {
     @Nonnull
     protected final Bindings.Builder bindings;
+    @Nonnull
+    private TypeRepository typeRepository;
 
     /**
      * Create an empty builder.
      */
     protected EvaluationContextBuilder() {
         this.bindings = Bindings.newBuilder();
+        this.typeRepository = TypeRepository.EMPTY_SCHEMA;
     }
 
     /**
@@ -54,6 +57,7 @@ public class EvaluationContextBuilder {
      */
     protected EvaluationContextBuilder(@Nonnull EvaluationContext original) {
         this.bindings = original.getBindings().childBuilder();
+        this.typeRepository = original.getTypeRepository();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/EvaluationContextBuilder.java
```diff
@@ -89,16 +93,29 @@ public EvaluationContextBuilder setBinding(@Nonnull String name, @Nullable Objec
...
         return setBinding(Bindings.Internal.CORRELATION.bindingName(alias.getId()), value);
     }
 
+    /**
+     * Set the {@link TypeRepository} for this evaluation context. This is an internal method that
+     * should only be used to supply types during query execution.
+     * @param typeRepository the {@link TypeRepository} to associate with the built evaluation context
+     * @return this {@code EvaluationContextBuilder}
+     */
+    @API(API.Status.INTERNAL)
```
alecgrieser, line 112: I marked the new methods that deal explicitly with a `TypeRepository` as `INTERNAL`, but it also looks like this isn't used (yet?) in the new planner either, so I might be off base with whether that's correct and whether the Javadoc is correct, but at least it is something we could change in the future.

normen662 (changes requested):
The reasoning behind making the `build(...)` method take the `TypeRepository` is to indicate that no caller should ever not set this explicitly. You should be forced to write `build(TypeRepository.empty())` is you really don't want a type repository. By no means should this be considered optional. Normally, there are two standard ways of achieving that: Either you make the builder's constructor take the mandatory arguments or to pass the mandatory arguments to the `build()` method.
I much rather keep this the way it is and then reintroduce a parameterless `build()` that is deprecated which delegates to `build(TypeRepository.empty())`

alecgrieser:
At this point, we've had the new API in a bit, and so regaining backwards compatibility is probably not as big a priority here. I'll go ahead and close this PR.

PR #1603, opened 2022-04-09
Mutiple languages support for Lucene

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -231,29 +237,47 @@ private boolean insertField(LuceneDocumentFromRecord.DocumentField field, final
...
+                return ((Boolean) field.getValue()).toString();
+            default:
+                throw new RecordCoreArgumentException("Invalid type for lucene index field", "type", field.getType());
+        }
     }
 
     private void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
                                byte[] primaryKey) throws IOException {
-        final IndexWriter newWriter = getOrCreateIndexWriter(state, indexAnalyzer, executor, groupingKey);
+        final TextLanguage language = TextLanguage.getLanguageForTexts(fields.stream().map(this::getStringValue).collect(Collectors.toList()));
```
MMcM, line 261: Would it make more sense to only do the full-text fields? It seems odd to count digits from numbers or the letter of "true" and "false" toward the non-CJK counts.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
```diff
@@ -96,19 +99,29 @@ public String getName() {
...
         @Nonnull
         @Override
-        public Analyzer getQueryAnalyzer(@Nonnull Index index, @Nonnull Analyzer indexAnalyzer) {
+        public Map<TextLanguage, Analyzer> getQueryAnalyzerMap(@Nonnull Index index, @Nonnull Map<TextLanguage, Analyzer> indexAnalyzerMap) {
             String name = index.getOption(LuceneIndexOptions.TEXT_SYNONYM_SET_NAME_OPTION);
             if (name == null) {
                 name = EnglishSynonymMapConfig.CONFIG_NAME;
             }
-            return new SynonymAnalyzer(EnglishAnalyzer.ENGLISH_STOP_WORDS_SET, name);
+            final Analyzer analyzer = new SynonymAnalyzer(EnglishAnalyzer.ENGLISH_STOP_WORDS_SET, name);
```
MMcM, line 118: Is this an intermediate step? All the language calculation still leads to the same analyzer?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/TextLanguage.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
```
MMcM, line 28: Needs Javadoc. This will be a failure as soon as the Checkstyle fix merges.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
tian-yizuo: Make sense. Using a filter to only check the ```TEXT``` fields now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/synonym/SynonymAnalyzer.java
tian-yizuo: added comments

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/TextLanguage.java
tian-yizuo: added javadoc.

MMcM (dismissed):

alecgrieser (changes requested):
My main comment here is that overall, I think this is a little limiting for adopters, in that it bakes in a single language detection mechanism and bakes in a limited set of languages. My counter proposal is an "analyzer chooser" abstraction (which is what could go in the registry) with an "analyzer for text" method that could be used by adopters to do language detection (or any other kind of conditional analyzer selection, if there are other use cases)

PR #1612, opened 2022-04-14
Resolve #1611 Support READABLE_UNIQUE_PENDING IndexState

### LINKED ISSUE #1611: Support READABLE_UNIQUE_PENDING IndexState
When finished indexing a unique index, but uniqueness violations are found, the index cannot be declare as READABLE until all the duplications are resolved. However, although not unique, the index still has usability at certain queries. To support this scenario, set this new state.

jjezra:


jjezra:
test this please

jjezra:
test this please

jjezra:
Test this please

jjezra:
Converted to draft to add use cases for the new index state.

jjezra:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -348,17 +353,45 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
...
-                            if (!built) {
-                                allReadable.set(false);
-                            }
-                            return null;
-                        }))
-                        .thenApply(ignore -> allReadable.get())
-                ), common.indexLogMessageKeyValues("IndexingBase::markReadable"));
+
+        AtomicReference<RecordIndexUniquenessViolation> uniquenessViolationExRef = new AtomicReference<>();
+        AtomicBoolean allReadable = new AtomicBoolean();
```
ScottDugas, line 358: This variable might be misleading now. It should be `true` if no states had to change, correct? That would be the implication if `markReadablePlease` is false.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -348,17 +353,45 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
...
+                throw ex;
+            }
+            return allReadable.get() ? AsyncUtil.READY_TRUE : AsyncUtil.READY_FALSE;
+        });
+    }
+
+    private CompletableFuture<Boolean> markIndexReadableSingleTarget(Index index, AtomicBoolean allReadable,
+                                                                     AtomicReference<RecordIndexUniquenessViolation> uniquenessViolationExRef) {
+        // An extension function to reduce markIndexReadable's complexity
+        return common.getRunner().runAsync(context ->
```
ScottDugas, line 375: In addition to potentially changing the state to `UNIQUE_PENDING`, this changes it so that if doing a multi-index build, and one index cannot be marked readable, the others can be right? I don't think that's clear from the issue or release notes, so probably worth calling out in one of those places.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1982,6 +1983,7 @@ public enum DesiredAction {
             ERROR,
             REBUILD,
             CONTINUE,
+            MARK_READABLE,
```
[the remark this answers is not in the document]
ScottDugas: If I'm reading `IndexingBase` correctly, it looks like this action indicates that if it sees an index that is partially built, it should try to mark it as readable (or `READABLE_UNIQUE_PENDING`) without continuing to build. It seems a bit weird to have that as an action on `OnlineIndexer`, as one could just as easily call `FDBRecordStore. markIndexReadableOrUniquePending` or `FDBRecordStore.markIndexReadable` directly (I think)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexState.java
```diff
@@ -55,7 +55,15 @@ public enum IndexState {
...
-    DISABLED(2L, "indexesDisabled");
+    DISABLED(2L, "indexesDisabled"),
+    /**
+     * Indicates that this unique index is fully "indexed", but
+     * some uniqueness violations still exist. This may happen
+     * when the online indexer finds some duplicating records.
+     * In this mode, it is safe to consider an index as {@link #READABLE}
+     * for queries as long as uniqueness is not assumed.
+     */
+    READABLE_UNIQUE_PENDING(3L, "indexesReadableUniquePending");
```
ScottDugas, line 66: Would it make sense to be forward-looking and have this be `READABLE_CONSTRAINT_PENDING` to support potentially other types of constraints on indexes in the future?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -365,7 +365,7 @@ protected <M extends Message> CompletableFuture<Void> updateOneKeyAsync(@Nonnull
                 state.store.countKeyValue(FDBStoreTimer.Counts.DELETE_INDEX_KEY, FDBStoreTimer.Counts.DELETE_INDEX_KEY_BYTES, FDBStoreTimer.Counts.DELETE_INDEX_VALUE_BYTES,
                         keyBytes, valueBytes);
             }
-            if (state.store.isIndexWriteOnly(state.index) && state.index.isUnique()) {
+            if (isWriteOnlyOrUniquePending() && state.index.isUnique()) {
```
ScottDugas, line 368: I know we support unique indexes for rank indexes, am I correct in my understanding that rank indexes will inherit this change? Should other indexes inherit this change?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
             context.commit();
         }
     }
+
+    @Test
+    public void testReadableUniquePendingSingle() {
```
ScottDugas, line 376: There were a bunch of changes to `FDBRecordStore`, It would probably be good to add some tests of those methods directly (which would hopefully expose the fact that `markIndexReadableOrUniquePending` returns `true` even if the index was already `READABLE_UNIQUE_PENDING`.
I think you should be able to avoid using the indexer by:
1. creating a unique index
2. saving some valid records
3. marking the index write only
4. saving an invalid record
5. trying to mark the index readable, or `READABLE_UNIQUE_PENDING`

It would probably be good to test the addition of uniqueness violations on an index in either of those states, and that you can get back the right set of uniqueness violations. 
If there isn't already a test in WriteOnly state, it would probably be good to test adding and removing multiple violations on the same key in both of these states.

ScottDugas, line 376: It would probably be good to add a test that builds an index off an index that has uniqueness violations.
I'm wondering if there could be risk if the destination index made assumptions about the source index being unique, but the only thing I can think of would be if the destination was the same index definition, you could see the indexer detecting that, and just copying the range, but missing the fact that the source had uniqueness violations, resulting in the destination being readable, but with uniqueness violations.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
...
+            context.commit();
+        }
+    }
+
+    private FDBRecordStoreTestBase.RecordMetaDataHook allIndexesHook(List<Index> indexes) {
+        return metaDataBuilder -> {
+            for (Index index: indexes) {
+                metaDataBuilder.addIndex("MySimpleRecord", index);
+            }
+        } ;
```
ScottDugas, line 449:
```suggestion
        };
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
...
+    public void testUniquenessMultiTargetsForbidUniquePending() {
+        testUniquenessMultiTarget(false);
+    }
+
+    @Test
+    public void testUniquenessMultiTargetsAllowUniquePending() {
+        testUniquenessMultiTarget(true);
+    }
+
+    private void testUniquenessMultiTarget(boolean allow) {
```
ScottDugas, line 472: `allow` -> `allowUniquePending`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas: It would probably be good to have tests that add more uniquness violations, both by adding records that have the same violations as existing records, and adding completely new violations. Resolving all should be required before updating to readable.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
...
+        try (FDBRecordContext context = openContext()) {
+            // unique index with uniqueness violation:
+            if (allow) {
+                assertTrue(recordStore.isIndexReadableUniquePending(indexes.get(0)));
+            } else {
+                assertTrue(recordStore.isIndexWriteOnly(indexes.get(0)));
+            }
+            // non-unique index:
+            assertTrue(recordStore.isIndexReadable(indexes.get(1)));
+            // unique index with on unique number (no duplications):
```
ScottDugas, line 514: I'm guessing `on` should be something else here, but I'm not sure what

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
...
+            for (int i = 5; i < records.size(); i++) {
+                assertNull(recordStore.loadRecord(Tuple.from(i)));
+            }
+            context.commit();
+        }
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setIndex(index).setSubspace(subspace)
+                .setIndexingPolicy(OnlineIndexer.IndexingPolicy.newBuilder()
+                        .allowUniquePendingState())
```
ScottDugas, line 545: It would probably be good to test building when an index was in `READABLE_UNIQUE_PENDING`, and still has violations.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas, line 376: Similarly, it would be good to make sure that we test building a unique index from a non-unique index both with and without violations, and erroring or not.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -366,4 +371,184 @@ public void resolveUniquenessViolations() {
...
+                .build()) {
+            indexBuilder.buildIndex(true);
+        }
+        try (FDBRecordContext context = openContext()) {
+            final IndexState indexState = recordStore.getIndexState(indexName);
+            assertEquals(IndexState.READABLE_UNIQUE_PENDING, indexState);
+            assertTrue(recordStore.isIndexReadableUniquePending(indexName));
+            context.commit();
+        }
+        // now try resolving it, and marking readable with another build
```
ScottDugas, line 408: In all of these tests, it would probably be good to scan the index, and assert about the results.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas: Another thing to test on the store directly would be scanning the index when it has uniqueness violations, and seeing those violations.
It would probably also be good to test this for all index types, particularly rank and bitmaps support the unique option, but bitmaps, I don't think support `READABLE_UNIQUE_PENDING` (which is fine, I don't think we need to add support for that here)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas: I'm wondering if there's a way to test if it fails for other reasons, but I can't think of any errors that we could produce, so adding the tests on `FDBRecordStore` that it can change into these states if things are broken might be all we can do, but if you can think of anything that would be great.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas: It would probably also be good to add tests that try to do `markIndexRadableOrUniquePending` when the index is only partially built, but has no uniqueness violations.

docs/ReleaseNotes.md
```diff
@@ -33,7 +33,7 @@ This release also updates downstream dependency versions. Most notably, the prot
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Support READABLE_UNIQUE_PENDING IndexState [(Issue #1611)](https://github.com/FoundationDB/fdb-record-layer/issues/1611)
```
ScottDugas, line 36: It's probably worth noting here or on the issue (I'm not sure what the best practice is), that using this new state is not compatible with older code.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexState.java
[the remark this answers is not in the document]
ScottDugas: I am wondering whether there is a process for transitioning the default, to allow this state. I don't think we currently have a process for this, so coming up with one as part of this change doesn't really make sense.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2908,6 +2908,20 @@ public Range getUnbuiltRange() {
...
+     * Marks an index as readable, but if uniqueness violations is found - mark the index
+     * as unique pending instead of throwing a {@link RecordIndexUniquenessViolation} exception.
+     *
+     * @param index the index to mark readable
+     * @return a future that will either complete exceptionally if the index can not
+     * be made readable/unique-pending or will contain <code>true</code> if the store was modified
+     * and <code>false</code> otherwise
+     */
+    @Nonnull
+    public CompletableFuture<Boolean> markIndexReadableOrUniquePending(@Nonnull Index index) {
```
ScottDugas, line 2921: Would there be reason to mark an index as `READABLE_UNIQUE_PENDING` even if it could be `READABLE` (or is already)? I think to get their now, you would have to add a uniqueness violation.
(Note: I already feel like `FDBRecordStore` has too many methods to make too many things just a method call, so I'm not really pushing for more)

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2933,33 +2952,7 @@ public CompletableFuture<Boolean> markIndexReadable(@Nonnull Index index) {
             byte[] indexKey = indexStateSubspace().pack(index.getName());
             CompletableFuture<Boolean> future = tr.get(indexKey).thenCompose(previous -> {
                 if (previous != null) {
```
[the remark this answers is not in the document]
jjezra: IIUC, yes. `null` state means `READABLE`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexState.java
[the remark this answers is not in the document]
jjezra: Since this state will appear in the DB after the user will actively allow it during indexing, would your second suggestion (depending on users not setting it until after all instances are updated) be acceptable? 
If so, I can add a warning to the proper javadoc (`shouldAllowUniquePendingState`).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1203,7 +1203,7 @@ public CompletableFuture<Integer> countRecords(
     @Nonnull
     public RecordCursor<IndexEntry> scanIndex(@Nonnull Index index, @Nonnull IndexScanBounds scanBounds,
                                               @Nullable byte[] continuation, @Nonnull ScanProperties scanProperties) {
-        if (!isIndexReadable(index)) {
+        if (!isIndexScannable(index)) {
```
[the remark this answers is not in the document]
jjezra: Should it be a breaking change, or can it be part of the `shouldAllowUniquePendingState` disclaimer? 

IIUC, the code is backward compatible, as it wouldn't affect existing users until they'll actively use the new feature. So in a way, `scanIndex` "had always" allowed this state.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2993,6 +2986,43 @@ public CompletableFuture<Boolean> markIndexReadable(@Nonnull String indexName) {
...
+                        "unbuiltRangeEnd", ByteArrayUtil2.loggable(firstUnbuilt.get().end),
+                        subspaceProvider.logKey(), subspaceProvider.toString(context),
+                        LogMessageKeys.SUBSPACE_KEY, index.getSubspaceKey());
+            } else if (uniquenessViolation.isPresent()) {
+                if (allowUniquePending) {
+                    updateIndexState(index.getName(), indexKey, IndexState.READABLE_UNIQUE_PENDING);
+                    // keep the index build data until the uniqueness violations are resolved and the index is marked as readable
+                    return true;
+                }
+                // do to clear build data for sanity check later on - if/when converting to readable
```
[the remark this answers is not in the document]
jjezra: Oops. Fixed and reworded.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3262,6 +3292,68 @@ public boolean isIndexReadable(@Nonnull String indexName) {
...
+     * @return <code>true</code> if the named index is readable-unique-pending and <code>false</code> otherwise
+     * @throws IllegalArgumentException if no index in the metadata has the given name
+     */
+    public boolean isIndexReadableUniquePending(@Nonnull String indexName) {
+        return getIndexState(indexName).equals(IndexState.READABLE_UNIQUE_PENDING);
+    }
+
+    /**
+     * Determine if the index with the given name is scannable - i.e. either readable or
+     * readable-unique-pending for this record store.
```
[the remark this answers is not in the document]
jjezra: Changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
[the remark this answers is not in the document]
jjezra: Rebased. Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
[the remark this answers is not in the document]
jjezra: This new action is used only for `READABLE_UNIQUE_PENDING` state (see `getStateDesiredAction(IndexState state)`).  Note that this function must return a desired action for any state. 

Users can easily mark readable directly, but if they'll call the build index instead (as many automatic tools do) - it's nice to skip the indexing overhead and jump directly to the actual "desired" action. It also adds a sanity check, when `markIndexReadable` verifies that all ranges are set (as expected in this state).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -348,17 +353,45 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
         if (!markReadablePlease) {
```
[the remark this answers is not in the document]
jjezra: 🤷 That's a great question. I've got it as a legacy, can we eliminate this option without breaking backward compatibility?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -348,17 +353,45 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
...
-                        .thenApply(ignore -> allReadable.get())
-                ), common.indexLogMessageKeyValues("IndexingBase::markReadable"));
+
+        AtomicReference<RecordIndexUniquenessViolation> uniquenessViolationExRef = new AtomicReference<>();
+        AtomicBoolean allReadable = new AtomicBoolean();
+
+        // Mark each index readable in its own (retriable, parallel) transaction
+        return forEachTargetIndex(index ->
+                markIndexReadableSingleTarget(index, allReadable, uniquenessViolationExRef)
+        ).thenCompose(ignore -> {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2993,6 +2986,43 @@ public CompletableFuture<Boolean> markIndexReadable(@Nonnull String indexName) {
...
+                        LogMessageKeys.INDEX_NAME, index.getName(),
+                        "unbuiltRangeBegin", ByteArrayUtil2.loggable(firstUnbuilt.get().begin),
+                        "unbuiltRangeEnd", ByteArrayUtil2.loggable(firstUnbuilt.get().end),
+                        subspaceProvider.logKey(), subspaceProvider.toString(context),
+                        LogMessageKeys.SUBSPACE_KEY, index.getSubspaceKey());
+            } else if (uniquenessViolation.isPresent()) {
+                if (allowUniquePending) {
+                    updateIndexState(index.getName(), indexKey, IndexState.READABLE_UNIQUE_PENDING);
+                    // keep the index build data until the uniqueness violations are resolved and the index is marked as readable
+                    return true;
```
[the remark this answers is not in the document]
jjezra: Good point. 
IIUC, currently the function returns false only if the index is already readable. Maybe, when setting the UP state the function should return an unconditional false (and update the javadoc to "return true if changed to READABLE, false otherwise"). 
I'm not sure how valuable the return code will be if it only indicates an arbitrary change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexState.java
jjezra: FMHO, uniqueness pending should be distinguished from other constraints - as it provides the specific information of an index being built but not unique. The planner, for Example, can benefit from having this specific info by the index state.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
jjezra: The standard maintainer seems to be the only entity that adds/clears uniqueness violation items.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: About the index from index - it does not make any uniqueness assumption and will detect uniqueness violations (if applicable) the same way as a by-records scan.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: Niether do I :) 
Changed to "unique index of unique numbers"

docs/ReleaseNotes.md
jjezra: Should it be a breaking change?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: I can't think about a reason to degrade a readable index that doesn't involve disabling the index altogether (or removing the UNIQUE flag - is it possible to do it today without re-indexing?). 
FMVHO, unique indexes in either `WRITE_ONLY` or `READABLE_UNIQUE_PENDING` states should forbid adding a conflicting records (when detected - if write only). But that's should probably be a separate discussion..

docs/ReleaseNotes.md
jjezra: Should the warning be in the release notes or in the `allowUniquePending` javadoc?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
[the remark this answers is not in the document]
jjezra: On a second thought, changed to match the javadoc - `will contain <code>true</code> if the store was modified`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: Most applied - but scanning for uniqueness violations. 
Do you mean index scan (or scrub)? It wouldn't discover UV, wouldn't it?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: The only thing I can think about is the config loader crashes. This is heavily used in indexing tests.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: Added to some.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -367,7 +367,7 @@ protected <M extends Message> CompletableFuture<Void> updateOneKeyAsync(@Nonnull
...
                 return removeUniquenessViolationsAsync(valueKey, savedRecord.getPrimaryKey());
```
jjezra, line 371: If successfully removed uniqueness violation entry (i.e. the counter is reduced to 1 and the entry was deleted), would It be acceptable to attempt converting the index to `READABLE`?

jjezra:
Rebased, resolved conflicts.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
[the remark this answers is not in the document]
alecgrieser: I agree we should note that this is a behavior change in the release notes, in case it does trip any one up. I also allude to this in my comment on the release notes change, but we may even want to add information to the release notes doc that lives outside the one-line note for the sake of legibility.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3023,6 +3017,45 @@ public CompletableFuture<Boolean> markIndexReadable(@Nonnull String indexName) {
...
+
+            if (firstUnbuilt.isPresent()) {
+                throw new IndexNotBuiltException("Attempted to make unbuilt index readable" , firstUnbuilt.get(),
+                        LogMessageKeys.INDEX_NAME, index.getName(),
+                        "unbuiltRangeBegin", ByteArrayUtil2.loggable(firstUnbuilt.get().begin),
+                        "unbuiltRangeEnd", ByteArrayUtil2.loggable(firstUnbuilt.get().end),
+                        subspaceProvider.logKey(), subspaceProvider.toString(context),
+                        LogMessageKeys.SUBSPACE_KEY, index.getSubspaceKey());
+            } else if (uniquenessViolation.isPresent()) {
+                if (allowUniquePending) {
```
alecgrieser, line 3037: I think, as @ScottDugas might have asked at one point, the normal way that we'd handle backwards compatibility here is to guard this on the store's `formatVersion` so that we don't end up writing a new index state that we can't guarantee people have the ability to read. In particular, I'd expect the behavior to be:

1. If the format version >= `UNIQUE_PENDING_FORMAT_VERSION` and `allowUniquePending`, then mark the index as readable/readable unique pending, otherwise
2. Throw a uniqueness violation

Which preserves backwards compatibility (both in behavior and in the data that is written to the index state) until the user chooses to update the format version. If we do the documentation right, that's also an opportunity for the user to see what the new feature means for them, and let them know if they are currently relying on unique indexes always being unique (or not scannable) that they need to adjust their assumptions.

One other advantage of using the format version is that we could (if we wanted to) remove "allow unique pending" entirely from the indexer state, and just always allow unique pending in index builds. Or maybe we want it as part of that configuration in case someone _wants_ index builds with uniqueness violations to fail, but with the format version, the user's decision to allow or not allow that could be de-coupled with their deploy a bit more.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
alecgrieser: I'm trying to remember the full history here, but it may be a result (ultimately) of either:

1. Wanting to support parallelized index builds (where different parts of the keyspace could be farmed out to different workers, in which case you wouldn't want any of them to try and mark the index as readable until they had all completed), or
2. The option existed because we could but not because there was a use case in mind.

I'm not sure if I recall the full history on that off the top of my head, though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -2236,9 +2251,34 @@ public Builder setIfReadable(final DesiredAction ifReadable) {
...
+             * Call {@link #allowUniquePendingState(boolean)} with default true.
+             * @return this builder
+             */
+            public Builder allowUniquePendingState() {
+                return this.allowUniquePendingState(true);
+            }
+
+            /**
+             * After indexing a unique index, if requested to mark the index as readable ({@link #buildIndex() }), but
+             * the index contain duplications, this function will determine the next behaviour:
```
alecgrieser, line 2264:
```suggestion
             * the index contain duplicates, this function will determine the next behavior:
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
alecgrieser: In particular, this doesn't need to go into other index maintainers (at least not the ones defined in the Record Layer) because they all inherit from this maintainer. That includes the `RankIndexMaintainer`, for example.

This might actually be another argument for using the format version, though. In particular, if a user has a custom index maintainer implementation that does not inherit from `StandardIndexMaintainer`, they may have code they need to update regarding how uniqueness constraints and index state interact, and that might be more cognizant of the need to make that change if they also need to update the format version. (For example, they can guard their calls to `.setFormatVersion` to a maximum supported format version, and only update that number once they've updated their index maintainers.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -405,7 +405,7 @@ protected <M extends Message> void checkUniqueness(@Nonnull FDBIndexableRecord<M
                     Tuple existingEntry = unpackKey(getIndexSubspace(), kv);
                     Tuple existingKey = state.index.getEntryPrimaryKey(existingEntry);
                     if (!TupleHelpers.equals(primaryKey, existingKey)) {
-                        if (state.store.isIndexWriteOnly(state.index)) {
+                        if (isWriteOnlyOrUniquePending()) {
```
alecgrieser, line 408: Are these the semantics we want? I think it makes sense for an index in `READABLE_UNIQUE_PENDING` to remove uniqueness violations on record delete, but arguably, it should enforce the uniqueness constraint during record save to avoid making the problem worse. Or, in other words, the user experience would be something like:

1. They build a unique index, but uniqueness violations mean it is now `READABLE_UNIQUE_PENDING`
2. Immediately, any new record saves start failing if they introduce a new uniqueness violation
3. Record deletes allow the user to start cleaning up the uniqueness violations
4. Once that's completed, the index can be marked as `READABLE`
5. Now queries can start relying on the uniqueness constraint

I'm not entirely convinced either way. I think the current behavior (which is effectively that the uniqueness constraint is ignored until the index can be marked as `READABLE`) is also understandable. We should probably document our decision one way or the other.

docs/ReleaseNotes.md
```diff
@@ -33,7 +33,7 @@ This release also updates downstream dependency versions. Most notably, the prot
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Support READABLE_UNIQUE_PENDING IndexState - the use of this index state is not compatible with older code [(Issue #1611)](https://github.com/FoundationDB/fdb-record-layer/issues/1611)
```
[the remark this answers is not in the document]
jjezra: AFAIU, this is correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
jjezra: I strongly agree that new duplications should be forbidden when the index's state is READABLE_UNIQUE_PENDING. I'll implement it in another commit.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: Would making it the default format version (=exceeding HEADER_USER_FIELDS_FORMAT_VERSION) be acceptable?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
ScottDugas: I see a lot of code that calls `scanUniquenessViolations`, but none that actually scan the values of the index and assert that it is scannable and has the right values.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2937,6 +2939,20 @@ public Range getUnbuiltRange() {
         }
     }
 
+    /**
+     * Marks an index as readable, but if uniqueness violations is found - mark the index
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -350,17 +355,46 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
...
+                                store.markIndexReadableOrUniquePending(index) :
+                                store.markIndexReadable(index))
+        ).handle((changed, ex) -> {
+            if (ex == null) {
+                if (Boolean.TRUE.equals(changed)) {
+                    anythingChanged.set(true);
+                }
+                return changed; // ignored
+            }
+            if (ex instanceof RecordIndexUniquenessViolation) {
```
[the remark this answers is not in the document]
jjezra: Changed to `RuntimeException` because couldn't make it `Throwable` - if other exception happens, I assume that the next line will throw a cast exception:
```
 runtimeExceptionAtomicReference.set((RuntimeException)ex);
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -159,7 +159,7 @@ private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store,
         validateOrThrowEx(maintainer.isIdempotent(), "target index is not idempotent");
         // readability - This method shouldn't block if one has already opened the record store (as we did)
         Index srcIndex = getSourceIndex(store.getRecordMetaData());
-        validateOrThrowEx(store.isIndexReadable(srcIndex), "source index is not readable");
+        validateOrThrowEx(store.isIndexScannable(srcIndex), "source index is not readable");
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -234,7 +234,7 @@ private CompletableFuture<Tuple> rebuildRangeOnly(@Nonnull FDBRecordStore store,
 
         // readability - This method shouldn't block if one has already opened the record store (as we did)
         final Index srcIndex = getSourceIndex(store.getRecordMetaData());
-        validateOrThrowEx(store.isIndexReadable(srcIndex), "source index is not readable");
+        validateOrThrowEx(store.isIndexScannable(srcIndex), "source index is not readable");
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -162,7 +162,7 @@ protected RecordCursor<FDBIndexedRawRecord> scanRemoteFetchByValue(@Nonnull fina
                 .setScanProperties(scanProperties)
                 .build();
         // Hard cast - Not ideal but likely needed to avoid complex implementation of specific cursor
-        RecordCursor<MappedKeyValue> mappedResults = keyValues.map(kv -> (MappedKeyValue)kv);
```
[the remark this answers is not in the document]
jjezra: Should these `scan` methods be preventing the scan of non-readable indexes? Is that the expected behavior?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -303,6 +280,41 @@ public void uniquenessViolations() {
...
+        }
+        openSimpleMetaData(hook);
+        try (FDBRecordContext context = openContext()) {
+            recordStore.markIndexWriteOnly(index).join();
+            context.commit();
+        }
+        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setIndex(index).setSubspace(subspace)
+                .build()) {
             indexBuilder.buildIndexAsync().handle((ignore, e) -> {
```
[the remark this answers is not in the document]
jjezra: Changed

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
```diff
@@ -331,18 +367,134 @@ public void resolveUniquenessViolations() {
...
+            if (allowUniquePending) {
+                indexBuilder.buildIndex();
+            } else {
+                RecordCoreException e = assertThrows(RecordCoreException.class, indexBuilder::buildIndex);
                 RuntimeException runE = FDBExceptions.wrapException(e);
                 assertNotNull(runE);
                 assertThat(runE, instanceOf(RecordIndexUniquenessViolation.class));
-                return null;
-            }).join();
+            }
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -350,17 +355,46 @@ public CompletableFuture<Boolean> markIndexReadable(boolean markReadablePlease)
...
-                            }
-                            return null;
-                        }))
-                        .thenApply(ignore -> allReadable.get())
-                ), common.indexLogMessageKeyValues("IndexingBase::markReadable"));
+
+        AtomicReference<RecordIndexUniquenessViolation> uniquenessViolationExRef = new AtomicReference<>();
+        AtomicBoolean anythingChanged = new AtomicBoolean(false);
+
+        // Mark each index readable in its own (retriable, parallel) transaction. If one target fails to become
```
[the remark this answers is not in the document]
ScottDugas: That is quite possible that it would be too complex, and not worth it. I just wanted to bring it up as something to consider, if it didn't add much to the complexity.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -189,13 +189,15 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
...
+
     // The current code can read and write up to the format version below
-    public static final int MAX_SUPPORTED_FORMAT_VERSION = HEADER_USER_FIELDS_FORMAT_VERSION;
+    public static final int MAX_SUPPORTED_FORMAT_VERSION = READABLE_UNIQUE_PENDING_FORMAT_VERSION;
 
     // By default, record stores attempt to upgrade to this version
     // NOTE: Updating this can break certain users during upgrades.
     // See: https://github.com/FoundationDB/fdb-record-layer/issues/709
-    public static final int DEFAULT_FORMAT_VERSION = CACHEABLE_STATE_FORMAT_VERSION;
+    public static final int DEFAULT_FORMAT_VERSION = READABLE_UNIQUE_PENDING_FORMAT_VERSION;
```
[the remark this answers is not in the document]
jjezra: Changed back to `CACHEABLE_STATE_FORMAT_VERSION`.

docs/ReleaseNotes.md
```diff
@@ -10,6 +10,12 @@ As the [versioning guide](Versioning.md) details, it cannot always be determined
 ### Features
 
 This version of the Record Layer allows the FDB API version to be configured through the `FDBDatabaseFactory`. This means that while this version allows the client to be configured to use 7.1 features, it also supports connecting to 6.3 FDB clusters if the API version is set appropriately. Note that setting the API version does restrict the set of potential FDB server versions that can be connected to, so this configuration change should only be made if the FDB server has already been updated.
+ 
+New index state "READABLE_UNIQUE_PENDING" - the proper way to roll this feature out is: 
+1. The adopter should upgrade to the new Record Layer version and deploy the version everywhere. 
+2. Only after all the possible clients are upgraded to support the new state, the adopter may set the allowPendingState on the indexing policy of new index builds. 
+An index may be in this new state if it is fully built, the unique flag is set, and duplications were found during online indexing. From the code point of view, it is defined as scannable but not readable.
```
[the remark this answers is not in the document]
jjezra: Added step
```
2. The format version should be set READABLE_UNIQUE_PENDING_FORMAT_VERSION.
```
is that what you've meant?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerUniqueIndexTest.java
jjezra: True. But can it be done in another PR? It doesn't seem to be directly related to this new state. 
(For the record, we might also want to change the usage of `scanUniquenessViolations` to something like: 
```
try(RecordCursor<RecordIndexUniquenessViolation> cursor = recordStore.scanUniquenessViolations(index) {
...
}
```

jjezra:
Rebased, resolved conflicts, and squashed...

jjezra:
test this please

ScottDugas (approved):

PR #1613, opened 2022-04-15
Choose analyzers based on texts

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -236,12 +239,16 @@ private boolean insertField(LuceneDocumentFromRecord.DocumentField field, final
 
     private void writeDocument(@Nonnull List<LuceneDocumentFromRecord.DocumentField> fields, Tuple groupingKey,
                                byte[] primaryKey) throws IOException {
-        final IndexWriter newWriter = directoryManager.getIndexWriter(groupingKey, indexAnalyzer);
+        final List<String> texts = fields.stream()
+                .filter(f -> f.getType().equals(LuceneIndexExpressions.DocumentFieldType.TEXT))
+                .map(f -> (String) f.getValue()).collect(Collectors.toList());
+        final IndexWriter newWriter = directoryManager.getIndexWriter(groupingKey,
+                indexAnalyzerChooser.chooseAnalyzer(texts));
```
alecgrieser, line 246: I think the `FDBDirectoryWrapper` caches the `IndexWriter` for a given `FDBDirecrtory` (affectively assuming that it will always be opened with the same analyzer each time), so I think this will always return an `IndexWriter` with the first `Analyzer` selected on a given transaction.

The manager might need to do something like close the original writer if the analyzer doesn't match (if `Analyzer`s have an appropriate `.equals` implementation) to avoid having multiple writers open at the same time on the same directory

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -236,12 +239,16 @@ private boolean insertField(LuceneDocumentFromRecord.DocumentField field, final
...
         BytesRef ref = new BytesRef(primaryKey);
         Document document = new Document();
         document.add(new StoredField(PRIMARY_KEY_FIELD_NAME, ref));
         document.add(new SortedDocValuesField(PRIMARY_KEY_SEARCH_NAME, ref));
-        final AnalyzingInfixSuggester suggester = autoCompleteEnabled ? getSuggester(groupingKey) : null;
+        final AnalyzingInfixSuggester suggester = autoCompleteEnabled ? getSuggester(groupingKey, texts) : null;
```
alecgrieser, line 251: I think the `FDBDirectoryWrapper` also caches the `AnalyzingInfixSuggester`, so this also returns an `AnalyzingInfixSuggester` with the first analyzer it receives in a transaction

alecgrieser (changes requested):
It feels like there should be a test somewhere with an `AnalyzerChooser` that sometimes returns an alternative `Analyzer` based on the text contents, but overall, I like this approach to language detection if they require different `Analyzer`s

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
tian-yizuo: Fixed it by having a ```LuceneAnalyzerWrapper``` with a unique identifier for each ```analyzer```. Not the old writer is closed when a new one with a different identifier is needed.

alecgrieser (approved):

PR #1614, opened 2022-04-15
Resolve #1615: Create a transactional runner that does not retry

### LINKED ISSUE #1615: Create a Transactional Runner that does not retry
In order to support situations that have different retry logic from `FDBDatabaseRunner`, it would be useful to extract a class specific to managing the transaction, that doesn't retry at all.

ScottDugas:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/runners/TransactionalRunnerTest.java
```diff
@@ -0,0 +1,414 @@
...
+    @Test
+    void closesContexts() {
+        List<FDBRecordContext> contexts = new ArrayList<>();
+        AtomicInteger completed = new AtomicInteger();
+        try (TransactionalRunner runner = defaultTransactionalRunner()) {
+            for (int i = 0; i < 20; i++) {
+                runner.runAsync(false, context -> {
+                    contexts.add(context);
+                    return MoreAsyncUtil.delayedFuture(60, TimeUnit.SECONDS)
+                            // delayed future isn't guaranteed, so we need to make sure at least some delayed past the close
```
alecgrieser, line 211: Should this test just return a `new CompletableFuture()` (i.e., a future that never completes)?

alecgrieser (changes requested):
Looks good to me. A few minor suggestions

alecgrieser (approved):

PR #1616, opened 2022-04-18
Resolves #1624: Use an optimized BlendedInfixSuggester for auto-complete suggestion, that does not store term vectors

### LINKED ISSUE #1624: Use an optimized BlendedInfixSuggester for auto-complete suggestion, that does not store term vectors
So the scoring based on positions could be done by detecting the positions of search term from matched texts in memory.

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedBlendedInfixSuggesterWithoutTermVectors.java
```diff
@@ -0,0 +1,291 @@
...
+                    continue;
+                }
+            }
+            results.add(result);
+        }
+
+        return new ArrayList<>(results.descendingSet());
+    }
+
+    private double createCoefficient(Set<String> matchedTokens, String prefixToken,
```
MMcM, line 222: I think it would be helpful for this method to have a link to the upstream `BlendedInfixSuggester` version. If I am not mistaken, up to formatting conventions, this class is identical to that one, _except_ for the one method. But due to the way it is implemented, class inheritance is, unfortunately, not possible. So we have a big copy and then this change.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedBlendedInfixSuggesterWithoutTermVectors.java
tian-yizuo: Added javadoc

MMcM (approved):

PR #1629, opened 2022-04-25
plan properties and plan composability

normen662:
- removed the parameters `record` and `message` to `QueryPredicate.eval()` and `Value.eval()` as those classes now entirely depend on the current bindings which are kept in the `EvaluationContext`
- renamed _interesting_ or _required_ properties into wording cased on _constraint_. These are merely renamings and do not constitute logic changes.
- introduced an annotation `@GenerateVisitor` together with an annotation processor to automatically create visitor implementations for `RelationalExpression` and `RecordQueryPlan`
- formally introduced the concept of a plan property. A plan property is memoized and maintained by the planner, you can write rules against the memoized properties, i.e. you can require plans to match a pre to satisfy certain conditions, e.g. you can have an implementation rule conditioned upon the properties of its children plans.
- related to that introduced the concept of a `PlanPartition` which allows to group a set of plans by common plan properties, in various numbers of dimensions, i.e. partition plans by ordering and distinctness would give you plan partitions where a particular partition would contain all plans that share the same ordering and distinctness. Plan partitions are available to each transformation rule and can be matched in the pre of a rule.
- reimplemented a lot of old-style `...PlannerProperty` implementations to become plan properties. Most notably, ordering is one of those that were reimplemented. The property is holds an `Ordering`  instead of an `Optional<Ordering>` used before as we want to enforce that some ordering exists even if that ordering may not be useful. This forces logic to expressly declare the ordering. We now use the following properties:
  - `OrderingProperty` for orderings
  - `DistinctRecordsProperty` to indicate that top-level records are distinct and do not produce duplicates. This property will be modified to be more expressive.
  - `StoredRecordsProperty` to indicate that records flowed out of such a plan do contain additional information such a primary key, etc.
- modified the implementation rules to enforce them to get the correct set of needed plan properties (for their implemented plan children). This yielded to a much cleaner approach and it also allowed us to refuse to implement particular compositions of plans that use inherently incompable properties, for instance `ImplementDistinctRule` can only function on top of an inner plan that does have the `StoredRecordsProperty` (evaluated to `true`).
- refactored runtime operators for union, distinct, and in-union to allow for passing in of comparison functions rather than comparison keys of type `KeyExpression`


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DistinctRecordsProperty.java
```diff
@@ -0,0 +1,310 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * An attribute used to indicate if a plan produces distinct records.
+ */
+public class DistinctRecordsProperty implements PlanProperty<Boolean> {
```
normen662, line 68: This used to be the `CreatesDuplicatesProperty` (well the inverse of that).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveSortRule.java
```diff
@@ -74,12 +73,7 @@ public void onMatch(@Nonnull PlannerRuleCall call) {
             return;
         }
 
-        final Optional<Ordering> orderingOptional = OrderingProperty.evaluate(innerPlan, call.getContext());
-        if (orderingOptional.isEmpty()) {
-            return;
-        }
-
-        final Ordering ordering = orderingOptional.get();
+        final Ordering ordering = OrderingProperty.OrderingVisitor.evaluate(innerPlan);
```
normen662, line 76: Note to reviewer: This call uses the memoized ordering information. The reason we this rule still matches on `anyPlan()` instead of on a plan partition is the nature of this rule to just eliminate expressions.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/TypedMatcherWithPredicate.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * <p>
+ * Extreme care should be taken when implementing <code>ExpressionMatcher</code>, since it can be very delicate.
+ * In particular, expression matchers may (or may not) be reused between successive rule calls and should be stateless.
+ * Additionally, implementors of <code>ExpressionMatcher</code> must use the (default) reference equals.
+ * </p>
+ * @param <T> the bindable type that this matcher binds to
+ */
+@API(API.Status.EXPERIMENTAL)
+public class TypedMatcherWithPredicate<T> extends TypedMatcher<T> {
```
MMcM, line 43: I'm wondering whether `Typed` adds much in the class name.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
```diff
@@ -0,0 +1,295 @@
...
+import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnorderedUnionPlan;
+import com.apple.foundationdb.record.query.plan.sorting.RecordQuerySortPlan;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * An attribute used for an ordering.
```
MMcM, line 65: Needs its own Javadoc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
```diff
@@ -0,0 +1,295 @@
...
+        @Nonnull
+        @Override
+        public Boolean visitInValuesJoinPlan(@Nonnull final RecordQueryInValuesJoinPlan inValuesJoinPlan) {
+            return visitInJoinPlan(inValuesJoinPlan);
+        }
+
+        @Nonnull
+        @Override
+        public Boolean visitCoveringIndexPlan(@Nonnull final RecordQueryCoveringIndexPlan element) {
+            return true;
```
MMcM, line 108: Only for now, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ObjectValue.java
```diff
@@ -0,0 +1,133 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
+import com.google.common.collect.ImmutableSet;
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Set;
+
+/**
+ * A {@link Value} representing any object.
```
MMcM, line 41: Isn't there some better name for this? Would these be literals in the surface syntax?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySetPlan.java
```diff
@@ -152,4 +158,127 @@ default Set<CorrelationIdentifier> tryPushValues(@Nonnull final List<TranslateVa
...
+                };
+            }
+        };
+    }
+
+    /**
+     * A comparison key function that extracts a comparison key for binary comparison by some set operations which
+     * also provides a stable plan hash and hash code.
+     */
+    interface ComparisonKeyFunction extends PlanHashable {
```
MMcM, line 178: Having introduced this abstraction, I think it may be worth pushing it back to the cursor somehow and making this extend that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionOnValuePlan.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public RecordQueryInUnionOnValuePlan(@Nonnull final Quantifier.Physical inner,
+                                         @Nonnull final List<? extends InSource> inSources,
+                                         @Nonnull final Function<CorrelationIdentifier, Value> comparisonKeyValueFunction,
+                                         final boolean reverse,
+                                         final int maxNumberOfValuesAllowed) {
+        this(inner,
+                inSources,
+                CorrelationIdentifier.uniqueID(),
+                comparisonKeyValueFunction,
```
MMcM, line 52: This is a strange contract, that can only work if `comparisonKeyValueFunction` ignores its argument. Why doesn't this just take the `Value` directly?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionOnKeyExpressionPlan.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public class RecordQueryInUnionOnKeyExpressionPlan extends RecordQueryInUnionPlan {
+    public RecordQueryInUnionOnKeyExpressionPlan(@Nonnull final Quantifier.Physical inner,
+                                                 @Nonnull final List<? extends InSource> inSources,
+                                                 @Nonnull final KeyExpression comparisonKeyExpression,
+                                                 final boolean reverse,
+                                                 final int maxNumberOfValuesAllowed) {
+        super(inner,
+                inSources,
+                new ComparisonKeyFunction.OnKeyExpression(comparisonKeyExpression) {
```
MMcM, line 49: This seems too important to be anonymous here. If it can't be an abstract method in the parent, perhaps it could at least be static.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIntersectionOnValuePlan.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.common.collect.ImmutableList;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Union plan that compares using a {@link Value}.
```
MMcM, line 41: Intersection.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIntersectionOnKeyExpressionPlan.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Union plan that compares using a {@link KeyExpression}.
```
MMcM, line 42: Intersection.

MMcM (changes requested):
I am surprised that there aren't any new tests. Or non-trivial changes to existing ones to add dual testing. Because that suggests that this is entirely cleanup, with no new planner capabilities exposed. Which I am not sure is true.

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/GenerateVisitorAnnotationProcessor.java
```diff
@@ -0,0 +1,343 @@
...
+                        rootTypeElement, "The class %s cannot be annotated with this annotation.",
+                        rootTypeElement.getQualifiedName().toString());
+                return true;
+            }
+
+            final var generateVisitor = annotatedElement.getAnnotation(GenerateVisitor.class);
+            final var rootTypeMirror = rootTypeElement.asType();
+
+            final var packageOfRoot = elementUtils.getPackageOf(rootTypeElement);
+            final var subClassTypeMirrors = moduleElement
```
MMcM, line 162: How will it ever be possible to implement subclasses in other modules?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PropertiesMap.java
```diff
@@ -0,0 +1,117 @@
...
+public class PropertiesMap {
+    private static final Set<PlanProperty<?>> plannerAttributes =
+            ImmutableSet.<PlanProperty<?>>builder()
+                    .add(OrderingProperty.ORDERING)
+                    .add(DistinctRecordsProperty.DISTINCT_RECORDS)
+                    .add(StoredRecordProperty.STORED_RECORD)
+                    .build();
+
+    @Nonnull
+    private final Map<RecordQueryPlan, Map<PlanProperty<?>, ?>> planAttributesMap;
```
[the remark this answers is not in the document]
normen662: The `PropertiesMap` is supposed to be the map you describe. What is missing from `PropertiesMap` are better type safe getters and similar things. Will do that.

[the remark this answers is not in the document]
MMcM: I think it might also benefit from some type-safe `private` methods when dealing with a `PlanProperty<T>` and a `T` at the same time, which is otherwise not there.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/BindingMatcher.java
```diff
@@ -197,6 +197,33 @@ default Class<?> unboundRootClass() {
         }
     }
 
+    /**
+     * Create a dummy instance that can be used to manually set a binding in a {@link PlannerBindings}.
```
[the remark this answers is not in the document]
normen662: How about: `An instance of a matcher that can be used to directly set and retrieve a value for a matcher in a PlannerBindings instance.`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
normen662: yes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ObjectValue.java
normen662: Yes, they are something like folded constants. Do you have a good suggestion for naming?

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/GenerateVisitorAnnotationProcessor.java
normen662: Yes, you can do it by overriding `visitDefault()` and putting a switch in there.

hatyo (commented):
I read through this PR over few sessions, and I think I need to spend more time on it to fully understand it, but just to make sure that I don't unnecessarily block anything, I am fine if you want to bring this in @normen662 as long as the comments from @MMcM are addressed. I will continue reading it later on.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -141,16 +141,16 @@ void testInQueryNoIndex() throws Exception {
                             .where(queryComponents(exactly(equalsObject(filter)))));
 
             assertEquals(-1139367278, plan.planHash(PlanHashable.PlanHashKind.LEGACY));
-            assertEquals(-1907300063, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
-            assertEquals(-1694772440, plan.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS));
+            assertEquals(97941305, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
```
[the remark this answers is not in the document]
normen662: Note that `LEGACY` is unaffected by this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PropertiesMap.java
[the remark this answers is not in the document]
normen662: I added a private `computePropertyValue()` which is now generic by some `P`. None of the other info is leaked out in an unsafe fashion, i.e. all callers either get a typed value via

```
public <P> Map<RecordQueryPlan, P> getPlannerAttributeForAllPlans(@Nonnull final PlanProperty<P> planProperty)
```

or they get data that does not enable them to get to property values (i.e. only `PlanPartition`s and `PlannerProperty` sets (which do not hold their computed values).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ContainsExpressionInReferenceMatcher.java
```diff
@@ -0,0 +1,76 @@
...
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.query.plan.cascades.GroupExpressionRef;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
+import com.google.common.base.Verify;
+
+import javax.annotation.Nonnull;
+import java.util.stream.Stream;
+
+/**
+ * A binding matcher that matches the expression if the same object as another {@link BindingMatcher}.
```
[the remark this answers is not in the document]
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/BindingMatcher.java
```diff
@@ -197,6 +197,33 @@ default Class<?> unboundRootClass() {
...
+     * @param <T> type parameter
+     * @return a new instance of a matcher that can be used to manually set a binding in a {@link PlannerBindings}
+     *         object
+     */
+    static <T> BindingMatcher<T> instance() {
+        return new BindingMatcher<T>() {
```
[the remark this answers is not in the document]
normen662: I made it more obvious that a matcher constructed in this way cannot be used for regular structural matching purposes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/TypedMatcherWithPredicate.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
+
+import javax.annotation.Nonnull;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+/**
+ * A <code>BindingMatcher</code> is an expression that can be matched against a
```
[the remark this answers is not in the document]
normen662: I removed a bunch of identical commenting on the overrides of `bindMatchesSafely()` which were largely not useful as they just copy/pasted the java doc of the base method.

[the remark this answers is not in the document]
normen662: Also changed some other java doc around in the matcher hierarchy to make it more consistent with itself.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIntersectionOnKeyExpressionPlan.java
normen662: nice.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
normen662: `Value`s do not have a current message or current object. Everything is explicit. That sometimes causes a chicken and egg problem:

If you want to express
`RecordConstructorValue(current.x, current.y)`

The contract that the reference for current is actually named `current` is not baked in to the system. So the `Function<CorrelationIdentifier, Value>` constructs a value when it is passed the `current` reference.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionOnKeyExpressionPlan.java
normen662: These have been moved in the super class.

PR #1638, opened 2022-05-03
resolves #1671: match index hints in a query

### LINKED ISSUE #1671: allow for per use index hints
Currently we can only limit the set of usable indexes per query. That approach is too heavy-handed and should be per use of a record type. This issue will add the capability to hand in an `IndexHint` to the `FullUnorderedScanExpression` which contains a set of access hints which then from constraints on the matching engine to only match the query against those match candidates that represent these access hints.

pengpeng-lu:
* This PR is to support index hint in a query
* Add index info to FullUnorderedScanExpression, then when match candidates, the query can be subsumed by a candidate only when the index is included in the hint indexes.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
@@ -31,11 +31,9 @@ import java.util.Set;
 public class AccessHints {
     @Nonnull
     private final Set<AccessHint> accessHintSet = new HashSet<>();
-    private final boolean noAccessHint;
 
     public AccessHints(AccessHint... accessHints) {
         this.accessHintSet.addAll(Arrays.asList(accessHints));
-        this.noAccessHint = this.accessHintSet.isEmpty();
     }
 
     @Nonnull
@@ -43,10 +41,18 @@ public class AccessHints {
         return accessHintSet;
     }
 
+    public int size() {
+        return accessHintSet.size();
+    }
+
     public boolean containsAll(AccessHints other) {
-        // if accessHint is not set, it's considered to include all possible hints
-        if (noAccessHint) {
+        // if no hint is set, it's considered to include all possible hints
+        if (size() == 0) {
             return true;
+        }
+        // if other is empty, this does not include other
+        if (other.size() == 0) {
+            return false;
         }
         // check that all hints in other exist in this
         for (AccessHint hint: other.accessHintSet) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
@@ -45,7 +45,7 @@ public class AccessHints {
         return accessHintSet.size();
     }
 
-    public boolean containsAll(AccessHints other) {
+    public boolean containsAll(@Nonnull AccessHints other) {
         // if no hint is set, it's considered to include all possible hints
         if (size() == 0) {
             return true;
@@ -54,21 +54,6 @@ public class AccessHints {
         if (other.size() == 0) {
             return false;
         }
-        // check that all hints in other exist in this
-        for (AccessHint hint: other.accessHintSet) {
-            if (!contains(hint)) {
-                return false;
-            }
-        }
-        return true;
-    }
-
-    private boolean contains(AccessHint hint) {
-        for (AccessHint accessHint: accessHintSet) {
-            if (accessHint.equals(hint)) {
-                return true;
-            }
-        }
-        return false;
+        return accessHintSet.containsAll(other.getAccessHintSet());
     }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
@@ -43,21 +43,21 @@ public class AccessHints {
         return accessHintSet;
     }
 
-    public boolean contains(AccessHints other) {
+    public boolean containsAll(AccessHints other) {
         // if accessHint is not set, it's considered to include all possible hints
         if (noAccessHint) {
             return true;
         }
         // check that all hints in other exist in this
         for (AccessHint hint: other.accessHintSet) {
-            if (!containsOneHint(hint)) {
+            if (!contains(hint)) {
                 return false;
             }
         }
         return true;
     }
 
-    private boolean containsOneHint(AccessHint hint) {
+    private boolean contains(AccessHint hint) {
         for (AccessHint accessHint: accessHintSet) {
             if (accessHint.equals(hint)) {
                 return true;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
@@ -1,0 +1,68 @@
+/*
+ * AccessHints.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Represents a set of AccessHint a query or a match candidate has.
+ */
+public class AccessHints {
+    @Nonnull
+    private final Set<AccessHint> accessHintSet = new HashSet<>();
+    private final boolean noAccessHint;
+
+    public AccessHints(AccessHint... accessHints) {
+        this.accessHintSet.addAll(Arrays.asList(accessHints));
+        this.noAccessHint = this.accessHintSet.isEmpty();
+    }
+
+    @Nonnull
+    public Set<AccessHint> getAccessHintSet() {
+        return accessHintSet;
+    }
+
+    public boolean contains(AccessHints other) {
+        // if accessHint is not set, it's considered to include all possible hints
+        if (noAccessHint) {
+            return true;
+        }
+        // check that all hints in other exist in this
+        for (AccessHint hint: other.accessHintSet) {
+            if (!containsOneHint(hint)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean containsOneHint(AccessHint hint) {
+        for (AccessHint accessHint: accessHintSet) {
+            if (accessHint.equals(hint)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
@@ -45,7 +45,7 @@ public class AccessHints {
         return accessHintSet.size();
     }
 
-    public boolean containsAll(@Nonnull AccessHints other) {
+    public boolean satisfies(@Nonnull AccessHints other) {
         // if no hint is set, it's considered to include all possible hints
         if (size() == 0) {
             return true;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
@@ -26,12 +26,18 @@ import java.util.Objects;
 /**
  * Represents an index hint.
  */
-public class IndexAccessHint extends AccessHint {
+public class IndexAccessHint implements AccessHint {
     @Nonnull
     private final String indexName;
 
     public IndexAccessHint(@Nonnull final String indexName) {
         this.indexName = indexName;
+    }
+
+    @Override
+    @Nonnull
+    public String getAccessHintType() {
+        return "INDEX";
     }
 
     @Override
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
@@ -30,7 +30,7 @@ public class IndexAccessHint implements AccessHint {
     @Nonnull
     private final String indexName;
 
-    public IndexAccessHint(@Nonnull String indexName) {
+    public IndexAccessHint(@Nonnull final String indexName) {
         this.indexName = indexName;
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
@@ -1,0 +1,64 @@
+/*
+ * IndexAccessHint.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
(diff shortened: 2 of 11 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
```diff
@@ -325,7 +326,9 @@ static Optional<MatchCandidate> fromPrimaryDefinition(@Nonnull final RecordMetaD
     @Nonnull
     static GroupExpressionRef<RelationalExpression> createBaseRef(@Nonnull RecordMetaData metaData, @Nonnull final Set<String> allAvailableRecordTypes, @Nonnull final Set<String> recordTypesForIndex) {
         final var quantifier =
-                Quantifier.forEach(GroupExpressionRef.of(new FullUnorderedScanExpression(allAvailableRecordTypes)));
+                Quantifier.forEach(GroupExpressionRef.of(new FullUnorderedScanExpression(allAvailableRecordTypes, metaData.getAllIndexes().stream()
```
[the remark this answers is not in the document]
normen662: You will need to pass the name of the index this expansion is for to the expansion visitors that create the `MatchCandidate`. That name should be passed in here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RelationalExpression.java
```diff
@@ -107,10 +114,14 @@ static RelationalExpression fromRecordQuery(@Nonnull PlanContext context,
         final GroupExpressionRef<? extends RelationalExpression> baseRef;
         Quantifier.ForEach quantifier;
         if (recordTypes.isEmpty()) {
-            baseRef = GroupExpressionRef.of(new FullUnorderedScanExpression(context.getMetaData().getRecordTypes().keySet()));
+            baseRef = GroupExpressionRef.of(new FullUnorderedScanExpression(context.getMetaData().getRecordTypes().keySet(), context.getMetaData().getAllIndexes().stream()
```
normen662, line 117: As both branches have a common subexpression, maybe extract that and set a local variable beforehand. Also, you could add a new method to `MetaDataPlanContext` which gets you the index names quickly, as the index names are the key set of the bi-map `indexesByName`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/FullUnorderedScanExpression.java
```diff
@@ -134,7 +143,15 @@ public String toString() {
     @Nonnull
     @Override
     public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candidateExpression, @Nonnull final AliasMap aliasMap, @Nonnull final IdentityBiMap<Quantifier, PartialMatch> partialMatchMap) {
-        return exactlySubsumedBy(candidateExpression, aliasMap, partialMatchMap);
+        if (getClass() != candidateExpression.getClass()) {
+            return ImmutableList.of();
+        }
+        // if query doesn't contain any index, or contains candidate's index, the query should be exactly subsumed by the candidate
+        if (getIndexSet().size() == 0 || getIndexSet().containsAll(((FullUnorderedScanExpression)candidateExpression).getIndexSet())) {
```
normen662, line 150: Please use `.isEmpty()` instead (IntelliJ gives you a warning actually). Also, I think instead of defaulting `isEmpty()` ==> no hinting, why not just setting all available index names when the fuse is created and no hints are specified.

normen662 (changes requested):
Hi,

the major code path is exactly as discussed. Nice! Please do work on these two bullets:
- setting the name of the index into the fuse of the `MatchCandidate`s when they get created. That name is available but needs to be threaded into the right place(s).
- test cases! Please add (at least) a query test case that demonstrates the hinting mechanism, i.e. even though a better index is available the worse one wins because it is hinted. Look into `FDBSimpleQueryGraphTest` for inspiration on how to do that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/FullUnorderedScanExpression.java
pengpeng-lu: Good idea! will set all available index names, it is clearer.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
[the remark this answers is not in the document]
pengpeng-lu: I'm a little confused here: we can `getRecordTypes` from both `MetaDataPlanContext` and `RecordMetaData`, and the result seems a bit different for these two. Which one should we use in `new FullUnorderedScanExpression`? Also we can `getIndexNames` for both, which one should we use?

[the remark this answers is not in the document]
MMcM: The ones in `MetaDataPlanContext` are derived from the query being planned. So, the record types being queried ("`FROM` clause") and the allowed indexes ("`INDEX` hints"). The ones in `RecordMetaData` are the ones in the schema.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
```diff
@@ -314,7 +331,7 @@ static Optional<MatchCandidate> fromPrimaryDefinition(@Nonnull final RecordMetaD
                                                           final boolean isReverse) {
         if (commonPrimaryKey != null) {
             final var availableRecordTypes = metaData.getRecordTypes().keySet();
-            final var baseRef = createBaseRef(metaData, availableRecordTypes, recordTypes);
+            final var baseRef = createBaseRef(metaData, availableRecordTypes, recordTypes, commonPrimaryKey.toString());
```
pengpeng-lu, line 334: I think this is wrong, but don't know how to represent name of the primary key?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java
```diff
@@ -188,5 +191,80 @@ public void testSimplePlanGraph() throws Exception {
...
+                IndexQueryabilityFilter.TRUE,
+                false,
+                ParameterRelationshipGraph.empty());
+
+        assertMatchesExactly(plan2,
+                mapPlan(
+                        typeFilterPlan(
+                                scanPlan()
+                                        .where(scanComparisons(range("([1],>")))))
+                        .where(result(recordConstructorValue(exactly(fieldValue("name"), fieldValue("rest_no"))))));
```
pengpeng-lu, line 267: working on changing the expected plan now.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java
```diff
@@ -188,5 +191,78 @@ public void testSimplePlanGraph() throws Exception {
...
+                Optional.empty(),
+                IndexQueryabilityFilter.TRUE,
+                false,
+                ParameterRelationshipGraph.empty());
+
+        Assertions.assertFalse(mapPlan(
+                typeFilterPlan(
+                        scanPlan()
+                                .where(scanComparisons(range("([1],>")))))
+                .where(result(recordConstructorValue(exactly(fieldValue("name"), fieldValue("rest_no"))))).matchesExactly(plan2));
```
pengpeng-lu, line 266: This tests that a different plan is generated, how to construct the plan that can assertMatchExactly this new plan?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
pengpeng-lu: use empty string for primary key. it implies that when a user enforces an index hint, we will NOT use primary storage in the plan.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
```diff
@@ -314,7 +316,7 @@ static Optional<MatchCandidate> fromPrimaryDefinition(@Nonnull final RecordMetaD
                                                           final boolean isReverse) {
         if (commonPrimaryKey != null) {
             final var availableRecordTypes = metaData.getRecordTypes().keySet();
-            final var baseRef = createBaseRef(metaData, availableRecordTypes, recordTypes);
+            final var baseRef = createBaseRef(metaData, availableRecordTypes, recordTypes, "");
```
normen662, line 319: This feels hacky. How about defining a new class that can encapsulate these cases better:
1. match candidate side
    1. this is an index (there may even be other structures later as some indexes can be used in different ways, etc.)
    2. this is a primary access for the record type
2. query side
    1. no hinting for this access
    2. hinting to a set of indexes/and or primary

So maybe a new class called `AccessHints` which is handed in here that contains a set of `AccessHint`. `AccessHint` can say what it's hinting, i.e. a primary access or an index AND it can give the name if asked and even throw an exception if asked for the name of a primary access. `AccessHints` can encapsulate the check you currently do in `subsumedBy()` where you assign special meaning to the empty set.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/FullUnorderedScanExpression.java
```diff
@@ -65,15 +65,24 @@ public class FullUnorderedScanExpression implements RelationalExpression, Planne
     @Nonnull
     private final Set<String> recordTypes;
 
-    public FullUnorderedScanExpression(final Set<String> recordTypes) {
+    @Nonnull
+    private final Set<String> indexSet;
+
+    public FullUnorderedScanExpression(final Set<String> recordTypes, final Set<String> indexSet) {
```
normen662, line 71: Some java doc would be nice.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/FullUnorderedScanExpression.java
```diff
@@ -134,7 +143,15 @@ public String toString() {
     @Nonnull
     @Override
     public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candidateExpression, @Nonnull final AliasMap aliasMap, @Nonnull final IdentityBiMap<Quantifier, PartialMatch> partialMatchMap) {
-        return exactlySubsumedBy(candidateExpression, aliasMap, partialMatchMap);
+        if (getClass() != candidateExpression.getClass()) {
+            return ImmutableList.of();
+        }
+        // if 1) query contains candidate's indexes; AND 2) the query can be exactly subsumed by the candidate, the query can be subsumed by the candidate
+        if (getIndexSet().isEmpty() || getIndexSet().containsAll(((FullUnorderedScanExpression)candidateExpression).getIndexSet())) {
```
normen662, line 150: I understand what this is doing but this goes back to my comment earlier. The fact we assign special meaning to the empty set here is not great.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Represents a set of AccessHint a query or a match candidate has.
+ */
+public class AccessHints {
```
hatyo, line 31: This class looks a bit superfluous to me since it appears to only wrap a `Set` of `AccessHint`s, can we get rid of it and just use a `Set<AccessHint>` instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
```diff
@@ -0,0 +1,68 @@
...
+        if (noAccessHint) {
+            return true;
+        }
+        // check that all hints in other exist in this
+        for (AccessHint hint: other.accessHintSet) {
+            if (!contains(hint)) {
+                return false;
+            }
+        }
+        return true;
```
hatyo, line 57:
```suggestion
        return accessHintSet.empty() || accessHintSet.containsAll(other);
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
```diff
@@ -0,0 +1,68 @@
...
+        this.accessHintSet.addAll(Arrays.asList(accessHints));
+        this.noAccessHint = this.accessHintSet.isEmpty();
+    }
+
+    @Nonnull
+    public Set<AccessHint> getAccessHintSet() {
+        return accessHintSet;
+    }
+
+    public boolean containsAll(AccessHints other) {
```
hatyo, line 46:
```suggestion
    public boolean containsAll(@Nonnull final AccessHints other) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/IndexAccessHint.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Objects;
+
+/**
+ * Represents an index hint.
+ */
+public class IndexAccessHint implements AccessHint {
+    @Nonnull
+    private final String indexName;
+
+    public IndexAccessHint(@Nonnull String indexName) {
```
hatyo, line 33:
```suggestion
    public IndexAccessHint(@Nonnull final String indexName) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PrimaryAccessHint.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Objects;
+
+/**
+ * Represents a primary key hint.
+ */
+public class PrimaryAccessHint implements AccessHint {
+    @Nonnull
+    private final KeyExpression primaryKey;
+
+    public PrimaryAccessHint(@Nonnull KeyExpression primaryKey) {
```
hatyo, line 35:
```suggestion
    public PrimaryAccessHint(@Nonnull final KeyExpression primaryKey) {
```

hatyo (commented):
Great work! few minor comments from my side

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
pengpeng-lu: so currently `AccessHints` contains 1 logic `containsAll`, it is a little different than a traditional `Set.containsAll` (a query contains no hint, implies that it can be matched with any index), so it might be clearer to extract it as a class or at least a method. We can potentially put it in `AccessHint` class as well, I don't have a strong preference here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PrimaryAccessHint.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Objects;
+
+/**
+ * Represents a primary key hint.
+ */
+public class PrimaryAccessHint extends AccessHint {
+    @Nonnull
+    private final KeyExpression primaryKey;
+
+    public PrimaryAccessHint(@Nonnull final KeyExpression primaryKey) {
```
normen662, line 35: Primary Key is not needed here is it?

Primary Access just means to read a table directly without using an index plus fetch. I am not sure why you need a `KeyExpression` here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/FullUnorderedScanExpression.java
```diff
@@ -134,7 +144,15 @@ public String toString() {
     @Nonnull
     @Override
     public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candidateExpression, @Nonnull final AliasMap aliasMap, @Nonnull final IdentityBiMap<Quantifier, PartialMatch> partialMatchMap) {
-        return exactlySubsumedBy(candidateExpression, aliasMap, partialMatchMap);
+        if (getClass() != candidateExpression.getClass()) {
+            return ImmutableList.of();
+        }
+        // if query doesnot contain candidate's indexes, the query cannot be subsumed by the candidate
+        if (getAccessHints().containsAll(((FullUnorderedScanExpression)candidateExpression).getAccessHints())) {
```
normen662, line 151: rename `containsAll()` to something like `satisfies()`. We want to encapsulate what it does. In fact it's not a a straight up `constainsAll()`, it has special logic for empty which is required but kind of clashing with `constainsAll`

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AccessHints.java
```diff
@@ -0,0 +1,59 @@
...
+    @Nonnull
+    public Set<AccessHint> getAccessHintSet() {
+        return accessHintSet;
+    }
+
+    public int size() {
+        return accessHintSet.size();
+    }
+
+    public boolean satisfies(@Nonnull AccessHints other) {
```
normen662, line 48: Nice!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PrimaryAccessHint.java
```diff
@@ -0,0 +1,39 @@
...
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Represents reading a table directly without using an index plus fetch.
+ */
+public class PrimaryAccessHint implements AccessHint {
+
+    public PrimaryAccessHint() {
```
normen662, line 30: Hmmm -- I am pretty sure Sonar doesn't like this. Please remove the constructor entirely.

normen662 (approved):

Issue #1639, opened 2022-05-03
Support a configurable FDB API version


alecgrieser:
At the moment, the Record Layer picks a single FDB API version, and it passes that version to FDB during initialization. When we update our FDB Java version, we've typically in the past updated the API version to match the maximum supported API version for that FDB version.

We do this for a few reasons. It's main benefit is that it can simplify deployments, as once an API version has been set to version X, it is now impossible to use older FDB client or server versions. By using a fixed API version in the Record Layer, we can claim things like "Record Layer 3.0 requires FDB 6.3 or newer".

If we let the API version be configurable, we'd need to be a bit more careful about how communicate that information. In particular, we'd need to make clear that setting this value restricts what versions of FDB server can be used. This means that today, client/sever connectability is determined by: (1) the FDB server version, (2) the Record Layer version, and (3) the FDB C version(s) that have been configured with the client. If we added this configuration, then it would be all of that plus (4) the configured FDB API version.

When we expose it, we'll also need to make sure that Record Layer code reacts to any differences in the API between different API versions. These are generally published by the FDB project in its documentation on [API version upgrades](https://apple.github.io/foundationdb/api-version-upgrade-guide.html). Because each API version can have different semantics, we should restrict the set of valid API versions that we support to only ones that the code has been configured (and ideally tested) to work with. We will also probably want to routinely deprecate and remove older supported versions, as each version may potentially require compatibility code to account for API differences.

We may also not want to support this, if we don't like the overhead of the feature. It's primary benefit is that we can start to write code against new FDB features before dropping support for older FDB server versions (provided we have the proper safeguards to fall back to alternatives that don't require the new FDB code).

PR #1641, opened 2022-05-04
Resolves #1636: Increase FDB Java version to FDB 7.1 

### LINKED ISSUE #1636: Increase FDB Java version to FDB 7.1
The FDB 7.1 release has a few new APIs, including `getMappedRange`, that we will want to be able to expose through the Record Layer. We should increase our dependency version to 7.1, which will also require adding the new methods to our implementations of the `Transaction` interface.

### LINKED ISSUE #1640: Dependency bump for 3.2 release
As with most of our minor version releases, we should consider bumping the versions of our dependencies (in addition to FDB, see #1636) to newer versions. In particular, making sure our guava and protobuf versions have been updated would be good housekeeping.

alecgrieser:
This increases the FDB Java dependency to FDB 7.1. This keeps the API version at 630 so that the library can still be used to connect to FDB 6.3 and FDB 7.0 clusters, though it would now require using a 7.1 FDB C client as the primary client with the other clients as external clients.

This will allow for additional features to be built on top of FDB 7.1, though we probably want to also complete #1639 so that we can responsibly guard features reliant on 7.1 in environments where the cluster may still be running older FDB server versions.

This also updates the proto3 version to 3.20.1 and along with a few other dependencies. It leaves the guava version at 30.1-jre instead of updating it to 31.1-jre because that causes spotbugs violations regarding nullability problems. I think those errors are because of a possible change in the way Guava handles nullability annotations, but it seems like something we should fix later.

This resolves #1636, and it resolves #1640.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -4,11 +4,11 @@ This document contains a log of changes to the FoundationDB Record Layer. It aim
 
 As the [versioning guide](Versioning.md) details, it cannot always be determined solely by looking at the version numbers whether one Record Layer version contains all changes included in another. In particular, bug fixes and backwards-compatible changes might be back-ported to or introduced as patches against older versions. To track when a patch version has been included in the main release train, some releases will say as a note that they contain all changes from a specific patch.
 
-## 3.1
+## 3.2
 
-### Breaking Changes
+### Breaking changes
 
-This version of the Record Layer changes the Java source and target compatibility to version 11.  Downstream projects need to be aware that the byte code produced is of class file version `55.0` going forward.
+The FoundationDB Java binding dependency has been updated to 7.1 with this release. This means that clients also need to update their main FDB C client to a 7.1 version. Adopters that still wish to connect to an FDB cluster running a 6.3 or 7.0 server version can do so by packaging additional FDB C clients at the appropriate version(s) using the [FDB multi-version client feature](https://apple.github.io/foundationdb/api-general.html#multi-version-client).
 
 <!--
 // begin next release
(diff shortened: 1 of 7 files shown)


hatyo (commented):
LGTM

alecgrieser:
Rebase addresses merge conflicts (which were in `docs/ReleaseNotes.md`)

alecgrieser:
Rebase to resolve build failures from #1643

docs/ReleaseNotes.md
```diff
@@ -4,11 +4,13 @@ This document contains a log of changes to the FoundationDB Record Layer. It aim
 
 As the [versioning guide](Versioning.md) details, it cannot always be determined solely by looking at the version numbers whether one Record Layer version contains all changes included in another. In particular, bug fixes and backwards-compatible changes might be back-ported to or introduced as patches against older versions. To track when a patch version has been included in the main release train, some releases will say as a note that they contain all changes from a specific patch.
 
-## 3.1
+## 3.2
 
-### Breaking Changes
+### Breaking changes
```
MMcM, line 9: Is it intentional that this is different case from the following sections for older releases?

MMcM (changes requested):

docs/ReleaseNotes.md
alecgrieser: No, sorry, that was a typo. Fixed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -192,6 +194,23 @@ public AsyncIterable<KeyValue> getRange(Range range, int limit, boolean reverse,
...
     }
 
+    @Override
+    public AsyncIterable<MappedKeyValue> getMappedRange(final KeySelector begin, final KeySelector end, final byte[] mapper, final int limit, final boolean reverse, final StreamingMode mode) {
+        increment(FDBStoreTimer.Counts.READS);
+        increment(FDBStoreTimer.Counts.RANGE_READS);
+        return underlying.getMappedRange(begin, end, mapper, limit, reverse, mode);
+    }
+
+    @Override
```
ohadzeliger, line 204: Do we know for certain that no counters should be incremented for these calls? (for getMappedRange work has been done on the index prefetch branch, but I am not aware of similar work for the getRangeSplitPoints)

MMcM (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
alecgrieser: I'm...not sure, actually. It looks like that method probably uses the byte sample to get an estimate for relatively equally sized key ranges. My instinct was to treat these like `getEstimatedRangeSizeBytes` as of a similar type, and leave it uninstrumented. That being said, I'm not opposed to adding instrumentation here, either in this PR or in a follow up.

ohadzeliger: Sure. We can leave uninstrumented and change in followup PR.

ohadzeliger (approved):

alecgrieser:
Recommitting identical last commit to re-trigger sonar analysis 🥲

MMcM (approved):

PR #1642, opened 2022-05-04
Resolves #1639: Support a configurable FDB API version

### LINKED ISSUE #1639: Support a configurable FDB API version

alecgrieser:
This makes the FDB API version configurable through the `FDBDatabaseFactory`.

We'll need some better way of testing this, I think. I've tried running the tests at both the 6.3 and 7.1 API versions, and it passed, but this value cannot be changed after the client has been initiated, which makes this somewhat complex to test with our normal approach. We might be able to use a test-only implementation of `FDBDatabaseFactory` if we wanted to, say, test future work that had feature flags on the API version. Another approach might be to run the tests multiple times, once per API version

This resolves https://github.com/FoundationDB/fdb-record-layer/issues/1639.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/FDBTestBase.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/FDBTestBase.java
@@ -26,9 +26,22 @@ import org.junit.jupiter.api.BeforeAll;
  * Base class that all tests touching FoundationDB should inherit from.
  */
 public abstract class FDBTestBase {
+    private static final int MIN_API_VERSION = 630;
+    private static final int MAX_API_VERSION = 710;
+    private static final String API_VERSION_PROPERTY = "com.apple.foundationdb.apiVersion";
+
+    private static int getAPIVersion() {
+        int apiVersion = Integer.parseInt(System.getProperty(API_VERSION_PROPERTY, "630"));
+        if (apiVersion < MIN_API_VERSION || apiVersion > MAX_API_VERSION) {
+            throw new IllegalStateException(String.format("unsupported API version %d (must be between %d and %d)",
+                    apiVersion, MIN_API_VERSION, MAX_API_VERSION));
+        }
+        return apiVersion;
+    }
+
     @BeforeAll
     public static void setupFDB() {
-        FDB fdb = FDB.selectAPIVersion(630);
+        FDB fdb = FDB.selectAPIVersion(getAPIVersion());
         fdb.setUnclosedWarning(true);
     }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
@@ -44,12 +44,14 @@ import javax.annotation.Nonnull;
  * <p>
  * A note on Record Layer API stability: the elements of this enum are expected to shift as support
  * for FDB versions are deprecated and removed. As a result, the elements of this enum will be added
- * and removed as support for old FDB versions is dropped and new FDB versions is added.
+ * and removed as support for old FDB versions is dropped and new FDB versions is added. Setting the
+ * API version to a non-default value is also currently an experimental feature as additional testing
+ * and validation work to evaluate the library at various FDB API versions is developed.
  * </p>
  *
  * @see FDBDatabaseFactory#setAPIVersion(APIVersion)
  */
-@API(API.Status.MAINTAINED)
+@API(API.Status.EXPERIMENTAL)
 public enum APIVersion {
     /**
      * API version corresponding to FoundationDB 6.3.
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
@@ -1,0 +1,108 @@
+/*
+ * APIVersion.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.RecordCoreArgumentException;
+
+/**
+ * An enum representing the different supported FDB API versions. The
+ * <a href="https://apple.github.io/foundationdb/api-general.html#api-versions">FDB API version</a>
+ * controls the minimum FDB server version that can be successfully connected to, as well as
+ * what features are supported by the FDB client. It is therefore important that in scenarios where
+ * there are FDB servers on different versions that the API version be configured to not exceed the
+ * smallest FDB server version. If the API version is set too high, the FDB client may hang when
+ * trying to connect to the server.
+ *
+ * <p>
+ * Because FDB API versions also control which features are guaranteed to be supported, there may
+ * be Record Layer features that are gated on specific FDB API versions. For that reason, if there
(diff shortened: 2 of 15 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -238,6 +242,23 @@ public synchronized FDBLocalityProvider getLocalityProvider() {
...
+     * internal method that should only be called by internal modules, as higher level Record
+     * Layer APIs should shield adopters from needing to know this configuration value except
+     * possibly during application initialization.
+     *
+     * @return the API version of this database
+     * @see APIVersion
+     * @see FDBDatabaseFactory#setAPIVersion(APIVersion)
+     */
+    @API(API.Status.INTERNAL)
+    public APIVersion getAPIVersion() {
```
MMcM, line 258: How about some convenience method, like `FDBTransactionContext.atLeastVersion(APIVersion)`? Or does that go against the idea of keeping this internal?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
```diff
@@ -0,0 +1,108 @@
...
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.RecordCoreArgumentException;
+
+/**
+ * An enum representing the different supported FDB API versions. The
+ * <a href="https://apple.github.io/foundationdb/api-general.html#api-versions">FDB API version</a>
+ * controls the minimum FDB server version that can be successfully connected to, as well as
+ * what features are supported by the FDB client. It is therefore important that in scenarios where
+ * there are FDB servers on different versions that the API version be configured to not exceed the
+ * smallest FDB server version. If the API version is set too high, the FDB client may hang when
```
ohadzeliger, line 32:
```suggestion
 * lowest FDB server version. If the API version is set too high, the FDB client may hang when
```

ohadzeliger:
One way to maybe help testing is to add a (arguably useless) 6_2_0 version to the enum and then set that as the version in the tests. This would help us ensure that the setting of the api version to a level other than the default works.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
alecgrieser: I think that's a reasonable convenience method to have. I've marked the new methods as `INTERNAL` as well, under the spirit of keeping this internal, but it seems like a method we might want ourselves.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
alecgrieser: 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactoryImpl.java
```diff
@@ -101,7 +103,7 @@ protected synchronized FDB initFDB() {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug(KeyValueLogMessage.of("Starting FDB"));
             }
-            fdb = FDB.selectAPIVersion(API_VERSION);
+            fdb = FDB.selectAPIVersion(apiVersion.getVersionNumber());
```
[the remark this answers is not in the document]
alecgrieser: We don't (currently). I think it's probably a good idea to, though. I've added that to the "Starting FDB" log, as well as the other FDB client configuration parameters here

alecgrieser:
Regarding testing, one thing I added in this most recent update is an API version gradle property that can set the API version used during tests. I can run the tests through gradle now using multiple API versions and look at them pass. If we wanted to, I think we could do something like have multiple PRBs, one per API version, and make sure the tests run at all of those different versions

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
```diff
@@ -0,0 +1,121 @@
...
+@API(API.Status.MAINTAINED)
+public enum APIVersion {
+    /**
+     * API version corresponding to FoundationDB 6.3.
+     */
+    API_VERSION_6_3(630),
+    /**
+     * API version corresponding to FoundationDB 7.0.
+     */
+    API_VERSION_7_0(700),
```
ohadzeliger, line 61: Record layer hasn't been tested with FDB 700, can there be any risks by setting the API_VERSION to this?

ohadzeliger (approved):

fdb-extensions/src/test/java/com/apple/foundationdb/FDBTestBase.java
```diff
@@ -26,9 +26,22 @@
  * Base class that all tests touching FoundationDB should inherit from.
  */
 public abstract class FDBTestBase {
+    private static final int MIN_API_VERSION = 630;
+    private static final int MAX_API_VERSION = 710;
+    private static final String API_VERSION_PROPERTY = "com.apple.foundationdb.apiVersion";
+
+    private static int getAPIVersion() {
+        int apiVersion = Integer.parseInt(System.getProperty(API_VERSION_PROPERTY, "630"));
```
ohadzeliger, line 34: Question: Should the default be set to the `FDBFactoryImpl` default (so that when the system default changes, the tests will change automatically)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
alecgrieser: I think the risks of setting the API version to 700 and to 710 are about the same, which is that we don't (currently) run our PRBs at any API version other than 630. There were API differences between 630 and 700, though not for features that seemed relevant to the Record Layer's usage of the FDB API: https://apple.github.io/foundationdb/api-version-upgrade-guide.html#api-version-700

I could see the argument that we should mark `.setAPIVersion` as `EXPERIMENTAL`, and maybe the same for this enum, for at least until we have a more robust test story for supporting multiple API versions.

fdb-extensions/src/test/java/com/apple/foundationdb/FDBTestBase.java
alecgrieser: Unfortunately, not here. The `FDBDatabaseFactory` and `FDBDatabaseFactoryImpl` are in `fdb-record-layer-core`, and this is in `fdb-extensions`, and so the dependencies point in the wrong direction.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/APIVersion.java
ohadzeliger: Multiple PRBs (per FDB version) seem like a good (and not expensive) way to resolve this. FDB versions are not so frequently adopted to make this a time sink for the development effort.

alecgrieser: Okay, I've marked the new APIs as EXPERIMENTAL and noted that that's at least partially because we want some better testing of non-default API versions

ohadzeliger (approved):

PR #1644, opened 2022-05-04
Fixes #1643: Remove JCenter from build repositories

### LINKED ISSUE #1643: Remove JCenter from build repositories
The JCenter maven repository site has been shut down. See: https://blog.gradle.org/jcenter-shutdown

We have a few dependencies that are still being pulled from JCenter that will need to be removed to account for this change.

alecgrieser:
This removes the references to `jcenter()` from our gradle configuration. There was one dependency, one of the plugins, that didn't seem to have an up-to-date version on `mavenCentral()` (which was potentially why we were using `jcenter()` for anything), but they are now in the `gradlePluginsPortal()`, and so I switched those over to using the new gradle plugins format. The two remaining plugins did not seem to have entries in the `gradlePluginsPortal()`, so I left them as is for now.

I was able to redownload the dependencies locally, but a CI run would be good to know to double check that there wasn't accidentally some cache that made the build possible.

This fixes #1643.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,22 +25,21 @@ buildscript {
             logger.warn 'Warning: you have enabled maven local repository. This should only be used on local workstations to test undeployed library dependencies.'
         }
         mavenCentral()
-        jcenter()
     }
 
     dependencies {
-        classpath 'com.github.ben-manes:gradle-versions-plugin:0.38.0'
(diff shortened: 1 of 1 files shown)


ohadzeliger (approved):
LGTM

jleach4 (approved):

PR #1646, opened 2022-05-05
Resolves #1645: FDBRecordContext::newRunner drops property storage

### LINKED ISSUE #1645: FDBRecordContext::newRunner drops property storage
The method `FDBRecordContext::newRunner` creates a new `FDBDatabaseRunner` from an existing `FDBRecordContext` that is supposed to copy over the configuration from the first transaction (things like the priority, the weak-read semantics, the MDC context, the timer, etc.) so that each new transaction inherits those values from the base transaction. It appears that the `RecordLayerPropertyStorage` field is being dropped, though, so any of those configuration values aren't being passed in. See:

https://github.com/FoundationDB/fdb-record-layer/blob/93db7e3282554ebbc85fa77a8aa8368c0d7a7ac8/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java#L1191-L1199

tian-yizuo:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -1195,6 +1195,7 @@ public FDBDatabaseRunner newRunner() {
         runner.setWeakReadSemantics(weakReadSemantics);
         runner.setPriority(priority);
         runner.setTransactionTimeoutMillis(timeoutMillis);
+        runner.setRecordLayerPropertyStorage(propertyStorage);
```
alecgrieser, line 1198: It's minor, but we should probably update the javadoc on this method to mention that it transfers the property storage, too

alecgrieser (changes requested):
Just one minor javadoc suggestion

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
tian-yizuo: Updated the javadoc

alecgrieser (approved):

PR #1647, opened 2022-05-05
Fix a couple of new Checkstyle failures from 10.2 (missing blank lines).

MMcM:


.idea/checkstyle-idea.xml
```diff
@@ -3,13 +3,13 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/gradle/codequality/checkstyle.xml:Record Layer Checks" />
-        <entry key="checkstyle-version" value="7.5.1" />
+        <entry key="active-configuration-0" value="PROJECT_RELATIVE:$PROJECT_DIR$/gradle/codequality/checkstyle.xml:Record Layer Checks;All" />
+        <entry key="checkstyle-version" value="10.2" />
```
alecgrieser, line 7: Do we also need to update the `toolVersion` value to 10.2 in `check.gradle`:

https://github.com/FoundationDB/fdb-record-layer/blob/7ba7202931ec0eef36c6351469ed4daf79380d86/gradle/check.gradle#L51

I'm not actually entirely clear how those config values are supposed to interact with each other

alecgrieser (changes requested):
The actual code changes look good to me. One question about how this interacts with our larger check style configuration

.idea/checkstyle-idea.xml
MMcM: Yes, that's probably right. Updated.

PR #1650, opened 2022-05-05
Resolves #1649: Only merge up to 1 Lucene index when commit if there …

### LINKED ISSUE #1649: Only merge up to 1 Lucene index when commit if there are multiple

tian-yizuo:
…are multiple


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -412,21 +412,24 @@ private CompletableFuture<Void> loadFileReferenceCacheForMemoization() {
...
+                displayList.add(entry.getKey());
+                totalSize += entry.getValue().getSize();
+                actualTotalSize += entry.getValue().getActualSize();
             }
+            final String logMessage = getLogMessage("listAllFiles",
+                    LuceneLogMessageKeys.FILE_COUNT, displayList.size(),
+                    LuceneLogMessageKeys.FILE_LIST, displayList,
+                    LuceneLogMessageKeys.FILE_TOTAL_SIZE, totalSize,
+                    LuceneLogMessageKeys.FILE_ACTUAL_TOTAL_SIZE, actualTotalSize);
+            LOGGER.debug(logMessage);
```
alecgrieser, line 428: Given that it's kind of expensive to compute this debug message, it seems like we don't want to remove the `if (LOGGER.isDebugEnabled())` guard

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -412,21 +412,24 @@ private CompletableFuture<Void> loadFileReferenceCacheForMemoization() {
...
+
+            System.out.println("****************************************List All Files************************************************");
+            System.out.println(logMessage);
+            System.out.println("******************************************************************************************************");
```
alecgrieser, line 432: Is this some kind of debugging thing you added that should be removed before merge?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -435,6 +438,7 @@ public void commit() {
     public CompletableFuture<Void> commitAsync() {
         final long startTimeNanos = System.nanoTime();
         ensureActive();
+        mergeLuceneIndexIfNeeded();
```
alecgrieser, line 441: Rather than having the logic for only issuing one merge live here in `FDBRecordContext`, it seems like it would be better if the logic lived in the commit hook instead. Given that the `FDBRecordContext` already supports named commit checks, could you do something like:

1. Have a named commit check called something like "LUCENE_EXECUTE_MERGE" or something like that
2. It has an implementation that has a list of directories that need merging
3. In the merge implementations, the merge commit check is read, and then the merge operation is added to the queue
4. At commit check execution time, the commit check itself handles picking and executing one of the merges

I think this would allow all of the Lucene-specific logic to leave the `FDBRecordContext`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -98,8 +102,19 @@ public IndexWriter getWriter(LuceneAnalyzerWrapper analyzerWrapper) throws IOExc
...
-                                    LOGGER.trace("mergeSource={}", mergeSource);
-                                    super.merge(mergeSource, trigger);
+                                    state.context.addLuceneIndexMergeHook(state.index.getName(), () -> {
+                                        try {
+                                            LOGGER.trace(KeyValueLogMessage.of("Basic index mergeSource",
+                                                    LuceneLogMessageKeys.MERGE_SOURCE, mergeSource,
+                                                    LuceneLogMessageKeys.MERGE_TRIGGER, trigger,
+                                                    LuceneLogMessageKeys.INDEX_NAME, state.index.getName(),
+                                                    LuceneLogMessageKeys.INDEX_SUBSPACE, state.indexSubspace));
+                                            super.merge(mergeSource, trigger);
```
alecgrieser, line 112: I think there are a few ways in which this is a little delicate vis-à-vis the scheduling of other operations:

1. We close the index writer in a commit hook. So, as written, the merges are executed before the writer is closed, though I have a proposal that would put the merges in a vanilla commit hook, which would mean that that's no longer the case. That means that right now, if closing the writer ever triggers a merge (which, I'm not sure if it does or not), then those merges won't ever get run. In the future, it gets a little more subtle
2. If there's a range delete (i.e., a `deleteRecordsWhere`) over a directory, then I think the merges associated with that delete aren't cleaned up properly
3. Ideally, it would be good to verify that the `LuceneMergeScheduler` is okay with `merge()` being a delayed call. I have no idea what semantics it expects of its implementations

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: @alecgrieser sorry this is just my local change for debugging, just included in the commit accidentally.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
tian-yizuo: right this is not supposed to be committed.

PR #1662, opened 2022-05-10
Fix the bug with the condition to renew FDBDirectoryWrapper's suggester

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -149,6 +143,15 @@ public AnalyzingInfixSuggester getAutocompleteSuggester(@Nonnull LuceneAnalyzerW
...
     }
 
+    private boolean suggesterRenewNeeded(@Nonnull LuceneAnalyzerWrapper indexAnalyzerWrapper,
+                                         @Nonnull LuceneAnalyzerWrapper queryAnalyzerWrapper,
+                                         @Nullable IndexOptions indexOptions) {
+        return suggester == null
+               || !suggesterIndexAnalyzerId.equals(indexAnalyzerWrapper.getUniqueIdentifier())
+               || !suggesterQueryAnalyzerId.equals(queryAnalyzerWrapper.getUniqueIdentifier())
+               || indexOptions != null && !indexOptions.equals(suggesterFieldIndexOptions);
+    }
```
jleach4, line 153: This required me to remember the order of boolean evaluation.  I think this is right but a parenthesis might make it easier to understand.

tian-yizuo: Added parenthesis

alecgrieser (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -149,6 +143,15 @@ public AnalyzingInfixSuggester getAutocompleteSuggester(@Nonnull LuceneAnalyzerW
         return suggester;
     }
 
+    private boolean suggesterRenewNeeded(@Nonnull LuceneAnalyzerWrapper indexAnalyzerWrapper,
+                                         @Nonnull LuceneAnalyzerWrapper queryAnalyzerWrapper,
+                                         @Nullable IndexOptions indexOptions) {
+        return suggester == null
+               || !suggesterIndexAnalyzerId.equals(indexAnalyzerWrapper.getUniqueIdentifier())
+               || !suggesterQueryAnalyzerId.equals(queryAnalyzerWrapper.getUniqueIdentifier())
+               || (indexOptions != null && !indexOptions.equals(suggesterFieldIndexOptions));
```
MMcM, line 152: I thought this should be `!Objects.equals(indexOptions, suggesterFieldIndexOptions)`. But I now see that this actually implements the correct check for the current behavior, namely that `suggesterFieldIndexOptions` is the sticky options from the last time any were given. That seems unexpected. So, now my question is whether that's really what we want / need.

MMcM (changes requested):

PR #1670, opened 2022-05-13
Making the StoreTimer.Counter class have a couple public constructors

scottfines:
To interface out StoreTimer, but still make use of it, it would be useful if we could create a Subclass of `StoreTimer` that can create and manage it's own `Counter` instances; this PR makes a few of the constructors in `StoreTimer.Counter` public so that they can be instantiated outside the `StoreTimer` class


MMcM:
Although this seems benign, why would one want to manage their own `Counter` instances? Why doesn't the pattern of creating more `enum`s implementing one of the `Event` subinterfaces and then having the base class manage counters for those work?

alecgrieser (approved):

PR #1675, opened 2022-05-16
Resolves #1674: Lucene scans without queries

### LINKED ISSUE #1674: Lucene scans without queries
`planOther` is not called when there is no filter at all. But the only filter(s) might be on the grouping key(s).

The code does the correct thing today. A single filter will not match `getQueryForFilter`. And an `And` will fail to match any of its children. However, that child filter check lacks a comment about why it is there, in as much as this is policy and not correctness.

So mostly this is about documentation and tests.

Still, it is possible to determine that the passed-on filter only had group filters after initially matching them and stop.

MMcM:
Note that this does not change any generated plans. It just tries to make the logic for excluding vacuous queries more clear and better commented. And adds a couple tests.

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
@@ -94,6 +94,7 @@ import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.indexS
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.primaryKeyDistinct;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.scan;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.typeFilter;
+import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.unbounded;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.unorderedUnion;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.aMapWithSize;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -905,4 +906,39 @@ void covering() throws Exception {
...
+            QueryComponent groupFilter = Query.and(
+                    Query.field("group").equalsValue(1L),
+                    Query.field("entry").oneOfThem().matches(Query.field("key").equalsValue("a")));
+            RecordQuery query = RecordQuery.newBuilder()
+                    .setRecordType(MAP_DOC)
+                    .setFilter(groupFilter)
+                    .build();
+            RecordQueryPlan plan = planner.plan(query);
+            Matcher<RecordQueryPlan> matcher = filter(groupFilter, typeFilter(equalTo(Collections.singleton(TextIndexTestUtils.MAP_DOC)), scan(unbounded())));
+            assertThat(plan, matcher);
```
alecgrieser, line 940: Is there a test showing that the plan actually executes correctly (that is, that it returns all results and doesn't trip over the lack of bounds)?

alecgrieser (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: I think the reasoning was that this need only check that a non-Lucene plan is generated and that such tests would exist in the core.

alecgrieser: Oh, I see--the plan in question is a full store scan, not an unbounded scan of the index. Got it

alecgrieser (approved):

PR #1677, opened 2022-05-16
Protect remote fetch from non-value subclasses of IndexMaintainer

### LINKED ISSUE #1679: Protect remote fetch from non-value subclasses of IndexMaintainer
Remote fetch should only be allowed for Index Maintainers that are able to handle the `scanRemoteFetch` method. Ensure that no maintainer picks up the method through inheritance by accident.

ohadzeliger:
Resolves #1679
Only allow remote fetch for ValureIndexManintainer
Add a few more test cases for in- and out-of- range tests

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/ValueIndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/ValueIndexMaintainer.java
@@ -35,7 +35,9 @@ import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.MetaDataException;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.provider.foundationdb.FDBIndexedRawRecord;
 import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
+import com.apple.foundationdb.record.provider.foundationdb.IndexScanBounds;
 import com.apple.foundationdb.tuple.Tuple;
 import com.apple.foundationdb.tuple.TupleHelpers;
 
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/ValueIndexMaintainer.java
```diff
@@ -129,4 +131,9 @@ public CompletableFuture<Tuple> evaluateAggregateFunction(@Nonnull IndexAggregat
                 .thenApply(kvo -> kvo.map(kv -> TupleHelpers.subTuple(kv.getKey(), groupSize, totalSize)).orElse(null));
     }
 
+    @Nonnull
+    @Override
+    public RecordCursor<FDBIndexedRawRecord> scanRemoteFetch(@Nonnull final IndexScanBounds scanBounds, @Nullable final byte[] continuation, @Nonnull final ScanProperties scanProperties, @Nonnull final KeyExpression commonPrimaryKey) {
```
alecgrieser, line 136: I think it would also be safe to add this same call to the `VersionIndexMaintainer`, which has entries that are identical to the `ValueIndexMaintainer`

The `RankIndexMaintainer` also, I think, which for `BY_VALUE` scans operates identically to the `ValueIndexMaintainer`. (Though `BY_RANK` operates differently and could be handled in a different PR.)

alecgrieser (approved):
I had one comment about how I think the `Rank` and `Version` index maintainers could call `scanRemoteFetchByValue` on `StandardIndexMaintainer`, but I think it's also fine to take this as-is, if you'd prefer

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/ValueIndexMaintainer.java
ohadzeliger: Will do, but I'd like to get some tests for these specific index types first. Will create issue.

PR #1678, opened 2022-05-16
Resolves #1676: Lucene merge based on probability

### LINKED ISSUE #1676: Lucene merge based on probability

tian-yizuo:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedBlendedInfixSuggesterWithoutTermVectors.java
```diff
@@ -260,8 +266,31 @@ protected IndexWriterConfig getIndexWriterConfig(Analyzer indexAnalyzer, IndexWr
 
             @Override
             public synchronized void merge(final MergeSource mergeSource, final MergeTrigger trigger) throws IOException {
-                LOGGER.trace("Auto-complete index mergeSource={}", mergeSource);
-                super.merge(mergeSource, trigger);
+                if (state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_MULTIPLE_MERGE_OPTIMIZATION_ENABLED) && trigger == MergeTrigger.FULL_FLUSH) {
+                    if ((new Random()).nextInt(mergeDirectoryCount) == 1) {
```
alecgrieser, line 270: Instead of a `new Random()` here (and in the other `merge`s), this should probably be a `ThreadLocalRandom`

alecgrieser, line 270: Should this check for when the random value is equal to 0 instead of 1? I think if you have 1 directory to merge, you might end up never merging

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -149,4 +159,26 @@ public static FDBDirectoryManager getManager(@Nonnull IndexMaintainerState state
             return newManager;
         }
     }
+
+    private int getMergeDirectoryCount(@Nonnull IndexMaintainerState state) {
+        final AtomicInteger luceneMergeCount = new AtomicInteger();
+        state.store.getRecordMetaData().getAllIndexes().stream().filter(i -> i.getType().equals(LuceneIndexTypes.LUCENE)).forEach(i -> {
```
alecgrieser, line 165: I believe this calculation results in a merge probability of 1/(number of directories) for each directory instead of 1/(number of directories that need merging).

I'm not sure that that's necessarily a problem, but I figured it was worth mentioning so that we're aware that that's what this does.

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
tian-yizuo: yeah that is true, but as far as I know there might not be a good way to figure out which directories need to merge at this level or at the overrides of ```ConcurrentMergeScheduler.merge()```. So this might be the best solution we can take now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedBlendedInfixSuggesterWithoutTermVectors.java
tian-yizuo: Fixed

tian-yizuo: Fixed.

alecgrieser (approved):

MMcM (approved):

alecgrieser (approved):

PR #1683, opened 2022-05-19
Resolves #1682: Support auto-complete via querying main Lucene directory

### LINKED ISSUE #1682: Support auto-complete via querying main Lucene directory
At the moment, we support auto-complete in the Lucene indexes by adding a separate directory with auto-complete specific information. However, that has the disadvantage of requiring duplicate storage of the field and index data for auto-complete purposes. To save space, we should consider using the main Lucene directory as the search index for auto-complete, though this will have the disadvantage that it will require re-loading the original record and then extracting the text from it in order to support auto-complete, but that might still be worth it given the expense of these indexes.

alecgrieser:
This modifies the auto-complete cursor to do a query over the main Lucene directory and then search the original records to recover the text.

This resolves #1682.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1153,7 +1144,6 @@ void testAutoCompleteSearchForPhrase() throws Exception {
                             "states united as a country",
                             "states have been united as a country",
                             "all the states united as a country",
-                            "all the states have been united as a country",
```
alecgrieser, line 1156: This text is from a document that is then updated with a new document. The old behavior was to return both versions of the text, whereas the new behavior only retrieves the text from the current version of the document.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1052,22 +1054,11 @@ void searchForAutoCompleteCrossingMultipleFields(boolean withTermVectors) throws
...
             // Assert the suggestions' keys
             List<String> suggestions = results.stream().map(i -> (String) i.getKey().get(i.getKeySize() - 1)).collect(Collectors.toList());
-            assertEquals(ImmutableList.of("good evening", "Good night", "Good morning", "Good afternoon", "I'm good", "That's really good!"), suggestions);
+            assertThat(suggestions, containsInAnyOrder("good evening", "Good night", "Good morning", "Good afternoon", "I'm good", "That's really good!"));
 
             // Assert the corresponding field for the suggestions
             List<String> fields = results.stream().map(i -> (String) i.getKey().get(i.getKeySize() - 2)).collect(Collectors.toList());
             assertEquals(ImmutableList.of("text", "text", "text", "text", "text2", "text2"), fields);
-
-            // Assert the suggestions are sorted according to their values, which are determined by the position of the term into the indexed text
```
alecgrieser, line 1061: The results are now returned in the same order as the original Lucene query, so these asserts sorting by score is no longer true. If we wanted to continue this, we would need to add some kind of sort step, which might actually be possible given that we have bounded results. But I haven't done that in this PR

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -258,14 +258,9 @@ public CompletableFuture<RecordCursorResult<IndexEntry>> onNext() {
...
-            return ByteArrayContinuation.fromNullable(null);
+            return RecordCursorEndContinuation.END;
         } else {
-            LuceneContinuationProto.LuceneIndexContinuation continuation = LuceneContinuationProto.LuceneIndexContinuation.newBuilder()
-                    .setDoc(searchAfter.doc)
-                    .setScore(searchAfter.score)
-                    .setShard(searchAfter.shardIndex)
-                    .build();
-            return ByteArrayContinuation.fromNullable(continuation.toByteArray());
+            return LuceneCursorContinuation.fromScoreDoc(searchAfter);
```
alecgrieser, line 263: I abstracted this kind of continuation into its own class. This change preserves the continuation value, but it means we will no longer need to serialize the proto to a byte array (if the continuation isn't needed)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -169,12 +170,325 @@ private void performLookup() throws IOException {
...
+        IndexReader indexReader = getIndexReader();
+        Set<String> fieldNames = new HashSet<>();
+        indexReader.leaves().forEach(leaf -> leaf.reader().getFieldInfos().forEach(fieldInfo -> fieldNames.add(fieldInfo.name)));
+        fieldNames.remove(LuceneIndexMaintainer.PRIMARY_KEY_FIELD_NAME);
+        fieldNames.remove(LuceneIndexMaintainer.PRIMARY_KEY_SEARCH_NAME);
+
+        final Set<String> tokenSet = new HashSet<>(tokens);
+        Query finalQuery = phraseQueryNeeded
+                             ? buildQueryForPhraseMatching(fieldNames, tokens, prefixToken)
+                             : buildQueryForTermsMatching(fieldNames, tokenSet, prefixToken);
```
tian-yizuo, line 302: If I understand correctly, you refactored the original code to only get the ```tokens``` in ```getQueryTokens``` and then build the ```finalQuery``` in separate methods, to avoid duplicate code. And you have to pass in a list for phrase query building and set for terms query building. This looks fine to me, but maybe it is worthy to have a comment for '''buildQueryForPhraseMatching``` method to explain why it needs a List but the ```buildQueryForTermsMatching``` uses Set?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -169,12 +170,325 @@ private void performLookup() throws IOException {
...
+                }
+                Tuple key = Tuple.from(matchedField.getFieldName(), matchedText);
+                if (groupingKey != null) {
+                    key = groupingKey.addAll(key);
+                }
+                // TODO: Add the primary key to the index entry
+                // Not having the primary key is fine for auto-complete queries that just want the
+                // text, but queries wanting to do something with both the auto-completed text and the
+                // original record need to do something else
+                IndexEntry indexEntry = new IndexEntry(state.index, key, Tuple.from(scoreDoc.score));
```
tian-yizuo, line 479: is it expected to use the doc's score rather than the concept "weights", and get rid of the logic to re-calculate the weights based on the positions of the matches in their texts and sort the results based on that?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -169,12 +170,325 @@ private void performLookup() throws IOException {
...
+                    if (fieldValue instanceof String) {
+                        String text = (String) fieldValue;
+                        if (text.length() > maxTextLength) {
+                            // Apply the text length filter before searching through the text for the
+                            // matched terms
+                            continue;
+                        }
+                        String match = searchAllMaybeHighlight(text, queryTokens, prefixToken, highlight);
+                        if (match != null) {
+                            matchedField = documentField;
```
tian-yizuo, line 461: I am not very sure it is expected semantics to only return one field when a doc has multiple fields that match with search key.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -169,12 +170,325 @@ private void performLookup() throws IOException {
...
+        lookupResults = lookup().skip(skip);
         if (timer != null) {
             timer.recordSinceNanoTime(LuceneEvents.Events.LUCENE_AUTO_COMPLETE_SUGGESTIONS_SCAN, startTime);
-            timer.increment(LuceneEvents.Counts.LUCENE_SCAN_MATCHED_AUTO_COMPLETE_SUGGESTIONS, lookupResults.size());
+        }
+    }
+
+    @SuppressWarnings("squid:S3776") // Cognitive complexity is too high. Candidate for later refactoring
+    @Nullable
+    private String searchAllMaybeHighlight(String text, Set<String> matchedTokens, @Nullable String prefixToken, boolean highlight) {
```
tian-yizuo, line 182: I think this method is worth unit tests.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -169,12 +170,325 @@ private void performLookup() throws IOException {
...
+                        if (match != null) {
+                            matchedField = documentField;
+                            matchedText = match;
+                            break;
+                        }
+                    }
+                }
+
+                if (matchedField == null) {
+                    return null;
```
tian-yizuo, line 469: It might make more sense to throw exception here rather than returning ```null```. We need to understand why no matching fields are found from the matched doc.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Continuation from scanning a Lucene index. This wraps the LuceneIndexContinuation protobuf message,
+ * which contains enough information to use the Lucene
+ * {@link org.apache.lucene.search.IndexSearcher#searchAfter(ScoreDoc, org.apache.lucene.search.Query, int) searchAfter}
+ * feature to resume a query.
+ */
+class LuceneCursorContinuation implements RecordCursorContinuation {
```
tian-yizuo, line 36: It is great to have a class for Lucene's cursor continuation!

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -364,14 +305,8 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexable
...
-    /**
-     * Get the {@link AnalyzingInfixSuggester} for indexing or query, from the session of the context if there exists a corresponding one, or by creating a new one.
-     * @param indexOptions the {@link IndexOptions} for suggester's {@link FieldType}. This only matters for when the suggester is for indexing.
-     * The one for query can just use an arbitrary one, so just pass in a NULL when getting a suggester for query, so the existing one from session of context can be reused.
-     */
-    private AnalyzingInfixSuggester getSuggester(@Nullable Tuple groupingKey, @Nonnull List<String> texts, @Nullable IndexOptions indexOptions) throws IOException {
-        return directoryManager.getAutocompleteSuggester(groupingKey, autoCompleteIndexAnalyzerChooser.chooseAnalyzer(texts),
-                autoCompleteQueryAnalyzerChooser.chooseAnalyzer(texts), highlightForAutoCompleteIfEnabled, indexOptions);
+    private Analyzer getAutocompleteQueryAnalyzer(@Nonnull List<String> texts) {
+        return autoCompleteQueryAnalyzerChooser.chooseAnalyzer(texts).getAnalyzer();
```
tian-yizuo, line 309: Considering the auto-complete uses the same index as basic index now, not sure whether it still needs a separate query analyzer. I think this depends on the behavior we want to support and we can further discuss this offline.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
tian-yizuo: I am not sure I am getting why this test data is not needed anymore.

tian-yizuo (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
alecgrieser: Sure. I added `LuceneAutoCompleteResultCursorTest` and put some unit tests of this functionality in here.

We may want to refactor the tokenize-and-search logic, at some point, because I could see us wanting to re-use this code somewhere else, but I didn't do that here.

MMcM (changes requested):

MMcM (approved):

alecgrieser:
We should be sure to squash this when merging, I think

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
tian-yizuo: @alecgrieser sorry that was a mistake. My original purpose for that was not to use the same record ID. That text was supposed to be a separate record to verify the phrase search for auto-complete. Could you please revert this change and change its recordID to ```1631L```?

alecgrieser: Sure thing. Made that change

tian-yizuo (approved):

Issue #1623, opened 2022-04-21
Need to upgrade PMD


MMcM:
We are at 6.10; there is a 6.44. A quick check shows 444 new warnings for core, so some cleanups (or suppressions) will need to accompany this change.

PR #1694, opened 2022-05-25
Resolves #1691: Lucene negative queries

### LINKED ISSUE #1691: Lucene negative queries
Queries with `Query.not` should generate appropriate `Query`s, trying to avoid introducing a wildcard when possible.

MMcM:
At root a new clause class that produces a negated boolean.
But since Lucene represents that as another `Occur`, the clause actually implements set subtraction.
The planner tries to maximize the cases where the subtraction is from a positive query rather than just `*:*`.

Includes #1675 because of conflicts.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
@@ -100,6 +100,10 @@ public class LucenePlanner extends RecordQueryPlanner {
             if (!groupingMatch.getType().equals((QueryToKeyMatcher.MatchType.EQUALITY))) {
                 return null;
             }
+            if (filterMask.allSatisfied()) {
+                // If filter is only group predicates, can skip trying to find non-trivial Lucene scan.
+                return null;
+            }
             groupingComparisons = new ScanComparisons(groupingMatch.getEqualityComparisons(), Collections.emptySet());
         } else {
             groupingComparisons = ScanComparisons.EMPTY;
@@ -257,6 +261,7 @@ public class LucenePlanner extends RecordQueryPlanner {
         if (filterMask != null && filterMask.getUnsatisfiedFilters().isEmpty()) {
             filterMask.setSatisfied(true);
         }
+        // Don't do Lucene scan if none are satisfied, though.
         if (childClauses.isEmpty()) {
             return null;
         }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
@@ -300,35 +300,37 @@ public class LucenePlanner extends RecordQueryPlanner {
     private static LuceneQueryClause negate(@Nonnull LuceneQueryClause clause) {
         if (clause instanceof LuceneBooleanQuery) {
             final LuceneBooleanQuery booleanQuery = (LuceneBooleanQuery)clause;
-            final BooleanClause.Occur occur = booleanQuery.getOccur();
-            if (occur == BooleanClause.Occur.MUST) {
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
```diff
@@ -255,19 +260,79 @@ private LuceneQueryClause getQueryForAndOr(@Nonnull LucenePlanState state, @Nonn
...
+                }
+            } else {
+                childClauses.add(childClause);
+            }
         }
         if (filterMask != null && filterMask.getUnsatisfiedFilters().isEmpty()) {
             filterMask.setSatisfied(true);
         }
+        if (!negatedChildren.isEmpty()) {
+            return new LuceneNotQuery(childClauses, negatedChildren);
```
alecgrieser, line 276: If someone did something like `lucene_clause_a OR NOT lucene_clause_b`, could this transformation lead to this being translated into something like `(lucene_clause_a) -(lucene_clause_b)` (effectively turning the `OR` to an `AND`)?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
```diff
@@ -255,19 +260,79 @@ private LuceneQueryClause getQueryForAndOr(@Nonnull LucenePlanState state, @Nonn
...
+                }
+                if (clause instanceof LuceneNotQuery) {
+                    LuceneNotQuery notQuery = (LuceneNotQuery)clause;
+                    if (clauses.isEmpty() && notQuery.getNegatedChildren().size() == 1) {
+                        return notQuery.getNegatedChildren().get(0);
+                    }
+                    clauses.addAll(notQuery.getNegatedChildren());
+                }
+                return new LuceneBooleanQuery(clauses, BooleanClause.Occur.SHOULD);
+            } else {
```
alecgrieser, line 316: Is there a missing case here for if `occur` is `SHOULD` instead of `MUST` or `MUST_NOT`? I'm a little worried that if we're not careful, this could result in accidentally including `(*:*)` in more queries than we'd like, but maybe that's a danger with negative queries more broadly

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
MMcM: I do not think so. `LuceneNotQuery` always has occur = `MUST`, so it is not possible for `negatedChildren` to be added to in the `SHOULD` case. Note how flattening only happens when occur matches.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
MMcM: A boolean query only has `SHOULD` or `MUST`. Those are the two branches of the `if`. `NOT` is not represented as one with `MUST_NOT`, but rather with as `MUST` with a set of negated children.

I think this can be made clearer here by using `switch`. I will do that.

More broadly, I tried a few approaches to reconcile the Boolean algebra trees that queries have with the Boolean queries with occurrence requirements that Lucene has. None of them were entirely satisfactory. One whose behavior was slightly more obvious was where the flattening and the addition of any `*:*` happened at `bind` time rather than plan time. In the end, though, I decided it was true to the intent of the API to have all such manipulation happen early and to have `bind` only be about deferring the concrete representation until the concrete values where known and no other rewriting.

alecgrieser (dismissed):

alecgrieser:
I think the stuff in here regarding negative queries looks good. This includes the PR #1675, which might need to be resolved before we merge this one. I'm also a little concerned about merge skew with #1623, so getting another PRB run would probably be good.

alecgrieser (approved):

PR #1696, opened 2022-05-25
Unicode combining acute accent follows the base character.

MMcM:
I don't _think_ these were deliberately put in a funny place.


alecgrieser (approved):
whoops, yeah, those should have been over the vowels rather than the consonants ¯\\\_(ツ)_/¯

PR #1702, opened 2022-05-26
Refactor the LuceneRecordCursor and pipeline the records loading and …

tian-yizuo:
…indexEntry building

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -132,6 +132,7 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         if (scanType == LuceneScanTypes.BY_LUCENE) {
             LuceneScanQuery scanQuery = (LuceneScanQuery)scanBounds;
             return new LuceneRecordCursor(executor, state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_EXECUTOR_SERVICE),
+                    state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_INDEX_CURSOR_PAGE_SIZE),
                     scanProperties, state, scanQuery.getQuery(), scanQuery.getSort(), continuation,
                     scanQuery.getGroupKey(), scanQuery.getStoredFields(), scanQuery.getStoredFieldTypes());
         }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
@@ -21,15 +21,18 @@
 package com.apple.foundationdb.record.lucene;
 
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.async.AsyncUtil;
 import com.apple.foundationdb.record.IndexEntry;
+import com.apple.foundationdb.record.PipelineOperation;
 import com.apple.foundationdb.record.RecordCoreException;
+import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.RecordCursorContinuation;
-import com.apple.foundationdb.record.RecordCursorEndContinuation;
 import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordCursorVisitor;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.cursors.BaseCursor;
 import com.apple.foundationdb.record.cursors.CursorLimitManager;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectoryManager;
 import com.apple.foundationdb.record.lucene.search.LuceneOptimizedIndexSearcher;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
@@ -54,11 +57,13 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.function.Function;
 
 /**
  * This class is a Record Cursor implementation for Lucene queries.
@@ -68,7 +73,7 @@ import java.util.concurrent.ExecutorService;
 class LuceneRecordCursor implements BaseCursor<IndexEntry> {
     private static final Logger LOGGER = LoggerFactory.getLogger(LuceneRecordCursor.class);
     // pagination within single instance of record cursor for lucene queries.
-    private static final int MAX_PAGE_SIZE = 201;
+    private final int pageSize;
     @Nonnull
     private final Executor executor;
     @Nullable
@@ -78,6 +83,14 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
     @Nullable
     private final FDBStoreTimer timer;
     private int limitRemaining;
+    /**
+     * The records to skip.
+     */
+    private final int skip;
+    /**
+     * To track the count of records left to skip.
+     */
+    private int leftToSkip;
     @Nullable
     private RecordCursorResult<IndexEntry> nextResult;
     final IndexMaintainerState state;
@@ -85,9 +98,10 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
     private final Query query;
     private final Sort sort;
     private IndexSearcher searcher;
-    private TopDocs topDocs;
-    private int currentPosition;
+    private RecordCursor<IndexEntry> lookupResults = null;
+    private int currentPosition = 0;
     private final List<KeyExpression> fields;
+    @Nullable
     private ScoreDoc searchAfter = null;
     private boolean exhausted = false;
     @Nullable
@@ -102,6 +116,7 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
     @SuppressWarnings("squid:S107")
     LuceneRecordCursor(@Nonnull Executor executor,
                        @Nullable ExecutorService executorService,
+                       int pageSize,
                        @Nonnull ScanProperties scanProperties,
                        @Nonnull final IndexMaintainerState state,
                        @Nonnull Query query,
@@ -112,11 +127,14 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
                        @Nullable final List<LuceneIndexExpressions.DocumentFieldType> storedFieldTypes) {
         this.state = state;
         this.executor = executor;
+        this.pageSize = pageSize;
         this.executorService = executorService;
         this.storedFields = storedFields;
         this.storedFieldTypes = storedFieldTypes;
         this.limitManager = new CursorLimitManager(state.context, scanProperties);
         this.limitRemaining = scanProperties.getExecuteProperties().getReturnedRowLimitOrMax();
+        this.skip = scanProperties.getExecuteProperties().getSkip();
+        this.leftToSkip = scanProperties.getExecuteProperties().getSkip();
         this.timer = state.context.getTimer();
         this.query = query;
         this.sort = sort;
@@ -127,10 +145,6 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
             } catch (Exception e) {
                 throw new RecordCoreException("Invalid continuation for Lucene index", "ContinuationValues", continuation, e);
             }
-        }
-        this.currentPosition = 0;
-        if (scanProperties.getExecuteProperties().getSkip() > 0) {
-            this.currentPosition += scanProperties.getExecuteProperties().getSkip();
         }
         this.fields = state.index.getRootExpression().normalizeKeyForPositions();
         this.groupingKey = groupingKey;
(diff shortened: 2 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -128,141 +134,29 @@ class LuceneRecordCursor implements BaseCursor<IndexEntry> {
...
         this.groupingKey = groupingKey;
     }
 
     @Nonnull
     @Override
     public CompletableFuture<RecordCursorResult<IndexEntry>> onNext() {
-        if (nextResult != null && !nextResult.hasNext()) {
-            // Like the KeyValueCursor, it is necessary to memoize and return the first result where
-            // hasNext is false to avoid the NoNextReason changing.
-            return CompletableFuture.completedFuture(nextResult);
```
alecgrieser, line 145: I think we want to keep this code that memoizes the `nextResult` once we hit a reason that stops the cursor from returning data.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -289,10 +183,16 @@ private synchronized IndexReader getIndexReader() throws IOException {
     }
 
     private void performScan() throws IOException {
+        if (lookupResults != null && currentPosition < pageSize) {
```
alecgrieser, line 186: Should there be modular arithmetic here? Something like `currentPosition % pageSize != 0`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -306,14 +206,119 @@ private void performScan() throws IOException {
...
+                // Initialized with values that aren't really legal in a Tuple to find offset bugs.
+                List<Object> fieldValues = Lists.newArrayList(fields);
+                if (groupingKey != null) {
+                    // Grouping keys are at the front.
+                    for (int i = 0; i < groupingKey.size(); i++) {
+                        fieldValues.set(i, groupingKey.get(i));
+                    }
+                }
+                if (storedFields != null) {
+                    for (int i = 0; i < storedFields.size(); i++) {
```
alecgrieser, line 267: I think this is copied over from the previous code, and it wasn't changed, but this looks wrong to me. In particular, the position `i` indicates the index of the _stored_ field array, but then the value is placed into the `fieldValues` array, and I don't think those need to align.

I don't think we necessarily need to block merging this PR on that, though, especially if this bug is preëxisting.

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
alecgrieser: Oh, I see--I think this is fine

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
MMcM: `storedFields` is sparse, I believe.

alecgrieser: I see. If it aligns, there may not be a problem, then

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -306,14 +206,119 @@ private void performScan() throws IOException {
...
             exhausted = true;
         }
-        topDocs = newTopDocs;
-        if (topDocs.scoreDocs.length != 0) {
-            searchAfter = topDocs.scoreDocs[topDocs.scoreDocs.length - 1];
+        if (newTopDocs.scoreDocs.length != 0) {
+            searchAfter = newTopDocs.scoreDocs[newTopDocs.scoreDocs.length - 1];
         }
+
+        lookupResults = RecordCursor.fromIterator(executor, Arrays.stream(newTopDocs.scoreDocs).iterator())
```
MMcM, line 213: The way this pipelining is done, there won't be overlap between the last few stored fields of one search and the next search. Is that okay?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: Done

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -132,8 +132,10 @@ public RecordCursor<IndexEntry> scan(@Nonnull final IndexScanBounds scanBounds,
         if (scanType == LuceneScanTypes.BY_LUCENE) {
             LuceneScanQuery scanQuery = (LuceneScanQuery)scanBounds;
             return new LuceneRecordCursor(executor, state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_EXECUTOR_SERVICE),
+                    state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_INDEX_CURSOR_PAGE_SIZE),
                     scanProperties, state, scanQuery.getQuery(), scanQuery.getSort(), continuation,
-                    scanQuery.getGroupKey(), scanQuery.getStoredFields(), scanQuery.getStoredFieldTypes());
+                    scanQuery.getGroupKey(), scanQuery.getStoredFields(), scanQuery.getStoredFieldTypes())
+                    .skip(scanProperties.getExecuteProperties().getSkip());
```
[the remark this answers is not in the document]
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -289,10 +183,16 @@ private synchronized IndexReader getIndexReader() throws IOException {
     }
 
     private void performScan() throws IOException {
+        if (lookupResults != null && currentPosition < pageSize) {
+            return;
+        }
```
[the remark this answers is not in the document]
tian-yizuo: Renamed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -306,14 +206,119 @@ private void performScan() throws IOException {
...
-        topDocs = newTopDocs;
-        if (topDocs.scoreDocs.length != 0) {
-            searchAfter = topDocs.scoreDocs[topDocs.scoreDocs.length - 1];
+        if (newTopDocs.scoreDocs.length != 0) {
+            searchAfter = newTopDocs.scoreDocs[newTopDocs.scoreDocs.length - 1];
         }
+
+        lookupResults = RecordCursor.fromIterator(executor, Arrays.stream(newTopDocs.scoreDocs).iterator())
+                .mapPipelined(scoreDoc -> buildIndexEntryFromScoreDocAsync(scoreDoc), state.store.getPipelineSize(PipelineOperation.KEY_TO_RECORD))
+                .filter(Objects::nonNull)
```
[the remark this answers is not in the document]
tian-yizuo: Right it is not supposed to be null. removed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: Let's probably just use the current approach for now, considering it is just about the performance when matches count is larger than page size, which happens rarely?

alecgrieser (dismissed):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -144,125 +150,45 @@ public CompletableFuture<RecordCursorResult<IndexEntry>> onNext() {
...
+        // Scan all the pages within skip range firstly
+        CompletableFuture<Void> scanPages = AsyncUtil.whileTrue(() -> {
+            if (skip < pageSize) {
+                return CompletableFuture.completedFuture(false);
             }
-        }
-        if (topDocs.scoreDocs.length - 1 < currentPosition && limitRemaining > 0 && !exhausted) {
             try {
-                performScan();
+                searchForTopDocs(pageSize);
```
MMcM, line 160: Doesn't this need to do something with the result of `searchForTopDocs`? Specifically to deal with the case where the requested `skip` is greater than the _total_ number of matches.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -144,125 +150,45 @@ public CompletableFuture<RecordCursorResult<IndexEntry>> onNext() {
...
-                                fieldValues.set(keyPos[i], setPrimaryKey.get(i));
-                            } else {
-                                leftovers.add(setPrimaryKey.get(i));
-                            }
-                        }
-                        tuple = Tuple.fromList(fieldValues).addAll(leftovers);
-                    } else {
-                        tuple = Tuple.fromList(fieldValues).addAll(setPrimaryKey);
-                    }
+            skip -= pageSize;
```
MMcM, line 169: Probably should reduce by _actual_ number of results that `searchAfter` has advanced.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -288,16 +214,52 @@ private synchronized IndexReader getIndexReader() throws IOException {
         return FDBDirectoryManager.getManager(state).getIndexReader(groupingKey);
     }
 
-    private void performScan() throws IOException {
+    private void maybePerformScan() throws IOException {
+        if (lookupResults != null && currentPosition + skip < pageSize) {
```
MMcM, line 218: `currentPosition` is an absolute position, not an in-current-page position. Therefore I do not believe it is comparable with `pageSize` after the first page.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -288,16 +214,52 @@ private synchronized IndexReader getIndexReader() throws IOException {
...
+            return;
+        }
+        if (lookupResults != null) {
+            lookupResults.close();
+        }
+
+        TopDocs newTopDocs = searchForTopDocs(Math.min(limitRemaining + skip, pageSize));
+
+        lookupResults = RecordCursor.fromIterator(executor, Arrays.stream(newTopDocs.scoreDocs).iterator()).skip(skip)
```
MMcM, line 227: Doesn't something need to clear `skip` for subsequent pages?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
MMcM: I know we decided not to attempt to further refactoring to parallelize across pages, but I will point out that having a paged cursor of docs and then skipping that and then mapping that to `IndexEntry` might make it easier to get the `skip` logic correct.

alecgrieser (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: No need to handle its results. we just want to skip the first pages which completely fall into skip range.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: Since this is only to handle the pages which fall completely into skip range. this should be fine. The first page which has both records to skip and records to return, will be handled by ```maybePerformScan();``` below this.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: yeah this if condition can be completely removed now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
tian-yizuo: Not it has one final ```skip``` and one mutable ```leftToSkip``` which tracks the left of records to skip. it is cleaned up after skipping all records now. And there are new added test to verify this.

MMcM (dismissed):

alecgrieser (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
MMcM: In the case where `skip` is greater than the page size, but there are less than the page size matching documents total, we do an extra search. We also falsely advance `leftToSkip`, but perhaps that doesn't matter because `searchAfter` will be at the end of them all and so the extra search will always be empty.

MMcM (approved):

PR #1704, opened 2022-05-31
resolves #1703: publish test jar artifact

### LINKED ISSUE #1703: Publish test jar as part of the regular distribution
Publish test jar as part of the regular distribution. This is useful in order to benefit from test classes of the record layer in test modules in downstream projects.

normen662:


hatyo (commented):
Tested it locally and the generated test JAR looks good.

ammolitor (approved):
changes look good, I'm re-running the CI builds, it looks like `sonarcloud.io` was out to lunch for the last run. Once CI passes, I'm comfortable with merging.

PR #1708, opened 2022-06-02
Resolve #1707: Enable index remote fetch for version index

### LINKED ISSUE #1707: Enable remote fetch for version index
Version index behaves much like VALUE index, so the same remote fetch scan can apply.
Enable the version scan remote fetch in the index maintainer and test.

ohadzeliger:
Enable remote fetch for version index.
Add tests.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
@@ -2476,4 +2476,5 @@ public class VersionIndexTest extends FDBTestBase {
                 .build());
         return planner.plan(query);
     }
+
 }
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
@@ -109,7 +109,6 @@ import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.bounds;
-import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.fetch;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.hasTupleString;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.indexName;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.indexScan;
@@ -1816,7 +1815,7 @@ public class VersionIndexTest extends FDBTestBase {
                             .setFilter(Query.field("num_value_2").equalsValue(0))
                             .setSort(VersionKeyExpression.VERSION)
                             .build();
-                    RecordQueryPlan plan = planner.plan(query);
+                    RecordQueryPlan plan = plan(query, fetchMethod);
                     assertEquals("Index(MySimpleRecord$num2-version [[0],[0]])", plan.toString());
                     cursor = recordStore.executeQuery(plan, null, ExecuteProperties.newBuilder().setReturnedRowLimit(3).build())
                             .asIterator();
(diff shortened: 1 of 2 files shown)


hfu94:
which fdb image this PR test uses? does it use the `HEAD` in `apple/foundationdb` repo?

ohadzeliger:
Not sure which FDB image this is using: @ammolitor would likely know.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
```diff
@@ -2354,4 +2429,37 @@ private <M extends Message> void validateUsingNewerVersionFormat(@Nonnull List<F
...
+    @Nonnull
+    private List<Tuple> scanIndexToKeys(final RecordQueryPlannerConfiguration.IndexFetchMethod fetchMethod, final String indexName, final ScanProperties direction) throws Exception {
+        if (fetchMethod == RecordQueryPlannerConfiguration.IndexFetchMethod.SCAN_AND_FETCH) {
+            return recordStore.scanIndex(metaData.getIndex(indexName), IndexScanType.BY_VALUE, TupleRange.ALL, null, direction)
+                    .map(IndexEntry::getKey)
+                    .asList().get();
+        } else {
+            IndexScanBounds scanBounds = new IndexScanRange(IndexScanType.BY_VALUE, TupleRange.ALL);
+            KeyExpression primaryKey = recordStore.getRecordMetaData().getRecordType("MySimpleRecord").getPrimaryKey();
+            return recordStore.scanIndexRemoteFetch(indexName, scanBounds, primaryKey, null, direction, IndexOrphanBehavior.ERROR)
```
alecgrieser, line 2442: Maybe I'm missing it, but would these tests pass on API versions less than `API_VERSION_7_1`? It looks like they don't include anything gating these tests on the API version.

It also looks like this doesn't actually fallback if the `fetchMethod` is `USE_REMOTE_FETCH_WITH_FALLBACK`. Which maybe is fine, but I'm not sure if that was the intention or not.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
```diff
@@ -2354,4 +2429,37 @@ private <M extends Message> void validateUsingNewerVersionFormat(@Nonnull List<F
...
+                    .map(FDBIndexedRecord::getIndexEntry)
+                    .map(IndexEntry::getKey)
+                    .asList().get();
+        }
+    }
+
+    @Nonnull
+    private List<FDBIndexedRecord<Message>> scanIndexToRecords(final String indexName, final ScanProperties direction) throws Exception {
```
alecgrieser, line 2450: I missed initially that this method only uses remote fetch. Is there a reason this doesn't take the `IndexFetchMethod` as a parameter?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
```diff
@@ -2300,6 +2331,50 @@ public void upgradeFormatVersions(int testFormatVersion, boolean splitLongRecord
         }
     }
 
+    @Test
+    void testScanVersionIndex() throws Exception {
```
alecgrieser, line 2335: It feels like this test should either be `testScanRemoteFetchVersionIndex()` or should take the `IndexFetchMethod` as a parameter (and adjust its behavior accordingly).

alecgrieser (changes requested):
Mostly LGTM. Just a few comments about some of the tests

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/VersionIndexTest.java
ohadzeliger: Good catch - these tests will fail for `API_VERSION` less than `7.1`. With Gradle and CI both at `7.1` we didn't see it but for adopters that set it to `6.3` the test would fail. Added Assumptions on the `API_VERSION`.

In this particular set of tests the index scan by itself does not support FALLBACK (only query does), and so the behavior of FALLBACK is the same as REMOTE_FETCH. We can change the index scan tests to only use two of the fetch modes, which would mean enumerate the two values, which would mean that if we ever add another value to the enum, this test will not cover it. Maybe it is better to use the entire enum (and waste some cycles on redundant runs) and be future-looking?
See change to line 248.

alecgrieser: Sure, that makes sense. I think having the `_WITH_FALLBACK` and not `_WITH_FALLBACK` methods do the same thing is fine for now. Maybe in the future, if we added some `scanRemoteFetchOrFallBack` method to scan an index, we'd want to switch this over to use that, but that doesn't need to happen now.

alecgrieser (approved):

alecgrieser:
Looks like the proto3 PRB failed with:

```
OnlineIndexerSimpleTest > testConfigLoader() FAILED
    org.opentest4j.AssertionFailedError: expected: <93> but was: <92>
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
        at com.apple.foundationdb.record.provider.foundationdb.OnlineIndexerSimpleTest.testConfigLoader(OnlineIndexerSimpleTest.java:883)
```

Which is another instance of #936, I think

alecgrieser (approved):

PR #1711, opened 2022-06-03
Resolve #1710 UserVersionChecker should be getting RecordMetaDataProt…

### LINKED ISSUE #1710: UserVersionChecker should be getting RecordMetaDataProto.DataStoreInfo
We need a variant of `UserVersionChecker.checkUserVersion` that receives the entire `RecordMetaDataProto.DataStoreInfo`. That will give implementations access to `formatVersion` and more.

sergey-mkhitaryan:
…o.DataStoreInfo

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
@@ -36,6 +36,7 @@ import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.RecordFunction;
 import com.apple.foundationdb.record.RecordIndexUniquenessViolation;
 import com.apple.foundationdb.record.RecordMetaDataBuilder;
+import com.apple.foundationdb.record.RecordMetaDataProto;
 import com.apple.foundationdb.record.RecordMetaDataProvider;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TupleRange;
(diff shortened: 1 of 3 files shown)


alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -169,11 +170,41 @@ default FDBStoreTimer getTimer() {
...
+         * If this store is being created for the first time, the store header will be set to the default instance. In
+         * particular, the format version will be zero, whereas the format version on all stores that have already been
+         * created will be greater than zero.
+         *
+         * @param storeHeader store header
+         * @param metaData the meta-data provider that will be used to get meta-data
+         *
+         * @return the user version to store in the record info header
+         */
+        default CompletableFuture<Integer> checkUserVersion(@Nonnull RecordMetaDataProto.DataStoreInfo storeHeader,
```
alecgrieser, line 183:
```suggestion
        @SuppressWarnings("deprecation")
        default CompletableFuture<Integer> checkUserVersion(@Nonnull RecordMetaDataProto.DataStoreInfo storeHeader,
```

I believe this is needed because it now calls the deprecated `checkUserVersion`. (I also think our PRBs are currently configured _not_ to check this, which is a bit of a problem.)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
sergey-mkhitaryan: Added, though that resulted in "redundant suppression" warning in my Intellij (and there was no deprecated call warning in the first place). Maybe something is up with my configuration.

PR #1715, opened 2022-06-04
Resolves #1714: LocatableResolver::reverseLookup should support borrowing read versions

### LINKED ISSUE #1714: LocatableResolver::reverseLookup should support borrowing read versions
The `LocatableResolver` APIs currently support using a parent transaction's read version during forward lookups (i.e., when resolving an interned value), but not for reverse lookups (i.e., when resolving the original value from an interned value). This means that if an operation does a lot of reverse lookups, each lookup needs to start its own transaction, which means paying the price of an additional get read version request.

alecgrieser:
This adds arguments to `reverseLookup` to allow a transaction to be included that can be used to get a read version, much like the `resolve` methods.

This resolves #1714.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
@@ -31,7 +31,9 @@ import com.apple.foundationdb.record.cursors.AutoContinuingCursor;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.provider.foundationdb.FDBDatabase;
+import com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContextConfig;
 import com.apple.foundationdb.record.provider.foundationdb.FDBReverseDirectoryCache;
 import com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer;
 import com.apple.foundationdb.tuple.ByteArrayUtil2;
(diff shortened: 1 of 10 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
```diff
@@ -75,18 +77,22 @@ public static void validate(@Nullable final FDBStoreTimer timer,
...
-                3)) {
-            resolver.getDatabase().asyncToSync(timer, FDBStoreTimer.Waits.WAIT_VALIDATE_RESOLVER,
-                    cursor.forEach(validatedEntry -> {
-                        if (!badEntriesOnly || validatedEntry.getValidationResult() != ValidationResult.OK) {
-                            entryListener.accept(validatedEntry);
-                        }
-                    }));
+        try (FDBDatabaseRunner runner = resolver.database.newRunner()) {
+            runner.setContextConfigBuilder(FDBRecordContextConfig.newBuilder()
+                    .setSaveOpenStackTrace(true));
```
ScottDugas, line 82: Is this intentional? It seems like we wouldn't want to always do this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ScopedDirectoryLayer.java
```diff
@@ -149,6 +149,22 @@ private CompletableFuture<Optional<ResolverResult>> readInternal(@Nonnull FDBRec
...
+    /**
+     * {@inheritDoc}
+     *
+     * @param timer a timer to use to instrument this operation
+     * @param value the value returned by this resolver
+     * @return a future returning an optional set to the key associated with the given value or {@link Optional#empty()}
+     *     if there is no such value in the map
+     * @deprecated implementors should switch to using {@link #readReverse(FDBRecordContext, Long)} instead
+     */
+    @Deprecated
```
ScottDugas, line 167: Should this implementation be removed?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/layers/interning/ScopedInterningLayer.java
```diff
@@ -118,13 +118,27 @@ protected CompletableFuture<ResolverResult> create(@Nonnull FDBRecordContext con
...
+    /**
+     * {@inheritDoc}
+     *
+     * @param timer a timer to use to instrument this operation
+     * @param value the value returned by this resolver
+     * @return a future returning an optional set to the key associated with the given value or {@link Optional#empty()}
+     *     if there is no such value in the map
+     * @deprecated implementors should switch to using {@link #readReverse(FDBRecordContext, Long)} instead
+     */
+    @Deprecated
```
ScottDugas, line 135: Should this implementation be removed?

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolverValidator.java
alecgrieser: Whoops, yeah. I turned this on during some debugging to look into "transaction already closed errors", but we don't need this any more

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ScopedDirectoryLayer.java
alecgrieser: I don't think we can yet... This is maybe a bit messy, but the easiest time to remove it, I think, would be when we remove the method from `LocatableResolver.java`. At present, callers can still end up using this method if they use the (now deprecated) method on `LocatableResolver`.

At present, `LocatableResolver` is written so that the new method (with a context) calls the old method (with a timer) to avoid breaking external extenders. If `LocatableResolver` were changed to instead make the old method (with a timer) call the new method (with a context) and made the new method `abstract`, then we could  in fact get rid of these implementations, though I think they'd be slightly less efficient than they are today (because they'd need to make some contexts just to throw them away without using them).

alecgrieser:
Rebase addresses merge conflict

ScottDugas (approved):

PR #1716, opened 2022-06-07
Resolve #1811: Return null when a field is not set

### LINKED ISSUE #1811: return null when a repeated field is not set
We'll need to wrap repeated field in a nested field.

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
```diff
@@ -105,6 +105,23 @@ private static Message getFieldMessageOnMessage(@Nonnull MessageOrBuilder messag
...
+            }
+            return list;
+        }
+        if (field.hasDefaultValue() || message.hasField(field)) {
+            return message.getField(field);
+        } else {
+            return null;
+        }
+    }
+
```
hatyo, line 124: This method looks very similar to `Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull String fieldName)`, except that the later one takes care of special `TupleFields` which is good I guess to preserve backward compatibility.

Therefore, I would suggest to reuse `getFieldOnMessage ` and refactor it accordingly such that we have an overload of it that returns the requested field value by number, which is what you want as far as I can tell, i.e.:

```java
    @Nullable
    public static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull String fieldName) {
        final Descriptors.FieldDescriptor field = findFieldDescriptorOnMessage(message, fieldName);
        return getFieldOnMessage(message, field);
    }

    @Nullable
    public static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, int fieldNumber) {
        final Descriptors.FieldDescriptor field = findFieldDescriptorOnMessageByNumber(message, fieldNumber);
        return getFieldOnMessage(message, field);
    } 
    
    @Nullable
    private static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull Descriptors.FieldDescriptor field) {
        if (field.isRepeated()) {
            int count = message.getRepeatedFieldCount(field);
            List<Object> list = new ArrayList<>(count);
            for (int i = 0; i < count; i++) {
                list.add(message.getRepeatedField(field, i));
            }
            return list;
        }
        if (field.hasDefaultValue() || message.hasField(field)) {
            if (field.getJavaType() == Descriptors.FieldDescriptor.JavaType.MESSAGE &&
                     TupleFieldsHelper.isTupleField(field.getMessageType())) {
                return TupleFieldsHelper.fromProto((Message)message.getField(field), field.getMessageType());
            } else {
                return message.getField(field);
            }
        } else {
            return null;
        }
    }
```

hatyo (changes requested):
Great work, I think it we should reuse the logic for retrieving a field value from a `Message`, but besides that, the PR looks good to me!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
normen662: 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
```diff
@@ -67,14 +70,32 @@ public static Object getFieldValue(@Nonnull MessageOrBuilder message, @Nonnull L
...
+        return getFieldOnMessage(message, field);
+    }
+
+    @Nullable
+    private static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull Descriptors.FieldDescriptor field) {
         if (field.isRepeated()) {
+            if (!message.getAllFields().containsKey(field)) {
+                // if a repeated field is unset, should return null
+                return null;
+            }
```
hatyo, line 97: ⚠️ This line is seems to change record layer behaviour; Previously and empty repeated field is considered `equal` to an empty list, now it returns null.

(the parametrised test `com.apple.foundationdb.record.query.expressions.QueryExpressionTest#testParameterComparisons` is now red, since this particular test is now failing because of this:

```java
{ "{} EQUALS {}", "repeat_me", Collections.emptyList(), Comparisons.Type.EQUALS, Collections.emptyList(), Boolean.TRUE},
```

I recommend checking with more experienced colleagues whether this is ok.

hatyo: @pengpeng-lu what was the conclusion about this particular change during our nullability discussion?

pengpeng-lu: It was incorrect -> after this change, the field is null if a field is unset or set to empty array. Normen and I discussed about the Type system, and I just sent what I think should be done in our slack.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -281,7 +283,7 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
             // collection type
             // case 1: primitive types -- assumed to be non-nullable array of that type
             if (typeCode.isPrimitive()) {
-                final var primitiveType = primitiveType(typeCode, false);
+                final var primitiveType = primitiveType(typeCode, true);
```
pengpeng-lu, line 286: I set all `isNullable` to `true` in this PR, because I think that's what we support today? We haven't supported Nonnull yet right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1462,6 +1517,8 @@ public void defineProtoType(final TypeRepository.Builder typeRepositoryBuilder)
                     FieldDescriptorProto.Label.LABEL_OPTIONAL);
             typeRepositoryBuilder.addMessageType(helperDescriptorBuilder.build());
             typeRepositoryBuilder.registerTypeToTypeNameMapping(this, typeName);
+
+             */
```
pengpeng-lu, line 1521: so I'm a little confused in this part:
1) before the array wrapper, why do we add a "value" field in this method?
2) maybe I should put what is in `addProtoField` in here? right now the two methods seem to be calling each other which makes me confusing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1475,20 +1532,17 @@ public void addProtoField(@Nonnull final TypeRepository.Builder typeRepositoryBu
                                   @Nonnull final Optional<String> typeNameOptional,
                                   @Nonnull final FieldDescriptorProto.Label label) {
             Objects.requireNonNull(elementType);
-            //
-            // If the inner type is nullable, we need to create a nested helper message to keep track of
-            // nulls.
-            //
-            if (definesNestedProto()) {
```
pengpeng-lu, line 1482: I guess I removed `definesNestedProto` part because we said we don't need to support arrays with nulls in it. Is there any backward compatibility issue?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -150,7 +150,7 @@ void testInQueryNoIndex() throws Exception {
                             .where(predicates(valuePredicate(fieldValue("num_value_2"), new Comparisons.ListComparison(Comparisons.Type.IN, ImmutableList.of(0, 2))))));
 
             assertEquals(997592219, plan.planHash(PlanHashable.PlanHashKind.LEGACY));
-            assertEquals(-951957746, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
+            assertEquals(-1123732652, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
```
pengpeng-lu, line 153: why do we assert `Hashcode` in unit tests? How did we get the expected the `Hashcode`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TypeRepositoryTest.java
```diff
@@ -180,11 +180,13 @@ void createTypeRepositoryFromArrayTypeWorks() {
     void addSameTypeMultipleTimesShouldNotCreateMultipleMessageTypes() {
         final TypeRepository.Builder builder = TypeRepository.newBuilder();
         final Type t = generateRandomStructuredType();
+        System.out.println("t:" + t);
```
pengpeng-lu, line 183: Please ignore this test for reviews, it fails because I deleted the `Array.defineProtoType` method, and I'd like to make sure the change makes sense before I fix the tests.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -672,6 +673,7 @@ void misMatchQueryShouldReturnNoResult(boolean shouldDeferFetch) throws Exceptio
         }
     }
 
+    @Disabled
```
pengpeng-lu, line 676: After marking this test as disabled, my publishing artifact to local works...I don't understand why...

MMcM: `publishToMavenLocal` fails or not depending on whether this test is enabled? That's strange, since it shouldn't run _any_ tests.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
MMcM: This is the most compact way to guard against _any_ change to the structure of plans. We have discussed ways of making it more robust at some point. If the plan has legitimately changed, or if the `planHash` has legitimately changed, then the test failure itself has the new hash code. But care is required to understand the change before any such update.

.idea/compiler.xml
```diff
@@ -63,11 +63,11 @@
...
-          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/org.foundationdb/fdb-java/7.1.10/5b63e331d6eb1758e9eacd946fb4dfd0d4ad03ca/fdb-java-7.1.10.jar" />
+          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/org.foundationdb/fdb-java/7.1.14/a5ba547fd768293b9974f224850344c4e703cd72/fdb-java-7.1.14.jar" />
+          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.squareup/javapoet/1.12.0/459ae6c796db7dcbc723f9536af2ce49d86c1a80/javapoet-1.12.0.jar" />
           <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.google.auto/auto-common/0.10/c8f153ebe04a17183480ab4016098055fb474364/auto-common-0.10.jar" />
           <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.google.guava/guava/30.1-jre/d0c3ce2311c9e36e73228da25a6e99b2ab826f/guava-30.1-jre.jar" />
           <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-api/1.7.36/6c62681a2f655b49963a5983b8b0950a6120ae14/slf4j-api-1.7.36.jar" />
-          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.squareup/javapoet/1.12.0/459ae6c796db7dcbc723f9536af2ce49d86c1a80/javapoet-1.12.0.jar" />
```
pengpeng-lu, line 70: will remove the change in the end. Why do I keep making this change?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -281,7 +282,7 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
             // collection type
             // case 1: primitive types -- assumed to be non-nullable array of that type
             if (typeCode.isPrimitive()) {
-                final var primitiveType = primitiveType(typeCode, false);
+                final var primitiveType = primitiveType(typeCode, true);
```
pengpeng-lu, line 285: I set all `isNullable` to `true`, because Valkyrie doesn't support "Non-nullable" fields right now. Does it conflict with current status?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1359,6 +1379,8 @@ class Array implements Type {
         @Nullable
         private final Type elementType;
 
+        private final boolean needsWrapper;
```
pengpeng-lu, line 1382: I'm not quite sure whether `isNullable` and `needsWrapper` should be the same.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1453,15 +1480,20 @@ boolean isErased() {
...
+                typeRepositoryBuilder.defineAndResolveType(wrapperType);
+            } else {
+                final var helperDescriptorBuilder = DescriptorProto.newBuilder();
+                helperDescriptorBuilder.setName(typeName);
+                elementType.addProtoField(typeRepositoryBuilder,
+                        helperDescriptorBuilder,
+                        1, "value",
+                        typeRepositoryBuilder.defineAndResolveType(elementType),
+                        FieldDescriptorProto.Label.LABEL_OPTIONAL);
+                typeRepositoryBuilder.addMessageType(helperDescriptorBuilder.build());
```
pengpeng-lu, line 1495: This is the original logic, I actually don't quite understand it.

.idea/compiler.xml
hatyo: I think it is because you made some local changes to Gradle libraries (for M1 compatibility), these changes are indirectly reflected by IntelliJ in this file, and for some reason (I think related to IntelliJ recommending checking in that file) we see this file here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        final FieldKeyExpression parent = nestingKeyExpression.getParent();
+        final KeyExpression child = nestingKeyExpression.getChild();
+        if (KeyExpression.FanType.None.equals(parent.getFanType()) && child.toKeyExpression().hasNesting()) {
+            RecordMetaDataProto.Field firstChild = child.toKeyExpression().getNesting().getParent();
+            if (repeatedFieldName.equals(firstChild.getFieldName()) && RecordMetaDataProto.Field.FanType.FAN_OUT.equals(firstChild.getFanType())) {
+                // remove firstChild from the KeyExpression
+                RecordMetaDataProto.KeyExpression newChild = child.toKeyExpression().getNesting().getChild();
+                RecordMetaDataProto.Nesting.Builder newNestingBuilder;
+                if (newChild.hasNesting()) {
+                    // recursively remove all wrapped array
```
pengpeng-lu, line 81: visitExpression is already recursive.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
```diff
@@ -0,0 +1,96 @@
...
+     * Unwrap nested array in keyExpression.
+     *
+     * @param nestingKeyExpression The input keyExpression
+     *
+     * @return a keyExpression without wrapped array
+     */
+    public static NestingKeyExpression unwrapArrayInKeyExpression(NestingKeyExpression nestingKeyExpression) {
+        final FieldKeyExpression parent = nestingKeyExpression.getParent();
+        final KeyExpression child = nestingKeyExpression.getChild();
+        if (KeyExpression.FanType.None.equals(parent.getFanType()) && child.toKeyExpression().hasNesting()) {
```
pengpeng-lu, line 74: ENUM equals -> ==

fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
```diff
@@ -23,7 +23,7 @@ apply plugin: 'com.github.johnrengelman.shadow'
 def coreProject = ":${ext.coreProjectName}"
 
 dependencies {
-  implementation project(coreProject)
+    implementation project(coreProject)
```
normen662, line 26: please undo!

.idea/compiler.xml
```diff
@@ -63,11 +63,11 @@
         <processorPath useClasspath="false">
           <entry name="$PROJECT_DIR$/fdb-extensions/.out/libs/fdb-extensions-3.2-SNAPSHOT.jar" />
           <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/com.google.auto.service/auto-service/1.0-rc6/4586e4191111b24135ffac93d3e1a91e91bf71d3/auto-service-1.0-rc6.jar" />
-          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/org.foundationdb/fdb-java/7.1.10/5b63e331d6eb1758e9eacd946fb4dfd0d4ad03ca/fdb-java-7.1.10.jar" />
+          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/org.foundationdb/fdb-java/7.1.14/a5ba547fd768293b9974f224850344c4e703cd72/fdb-java-7.1.14.jar" />
```
normen662, line 66: please undo

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param nestingKeyExpression The input keyExpression
+     *
+     * @return a keyExpression without wrapped array
+     */
+    @Nonnull
+    public static NestingKeyExpression unwrapArrayInKeyExpression(@Nonnull NestingKeyExpression nestingKeyExpression) {
+        final FieldKeyExpression parent = nestingKeyExpression.getParent();
+        final KeyExpression child = nestingKeyExpression.getChild();
+
+        RecordMetaDataProto.Nesting.Builder newNestingBuilder = RecordMetaDataProto.Nesting.newBuilder()
```
normen662, line 102: Do you really need to build a new `KeyExpression`? Isn't the _new_ key expression you want to synthesize a part of the old one that is passed in?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
```diff
@@ -113,9 +114,18 @@ public <M extends Message> Object eval(@Nonnull final FDBRecordStoreBase<M> stor
         final var fields = Objects.requireNonNull(getResultType().getFields());
         var i = 0;
         for (final var child : getChildren()) {
-            final var childResultElement = child.eval(store, context);
+            var childResultElement = child.eval(store, context);
             if (childResultElement != null) {
-                resultMessageBuilder.setField(descriptorForType.findFieldByNumber(fields.get(i).getFieldIndex()), childResultElement);
+                final var field = fields.get(i);
```
normen662, line 119: Is there an actual test for this in record-layer?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
```diff
@@ -79,8 +79,14 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return RecordCursor.fromList((List<?>)Objects.requireNonNull(collectionValue.eval(store, context)))
-                        .map(QueryResult::ofComputed);
+        final var result = collectionValue.eval(store, context);
```
normen662, line 82:
```
RecordCursor.fromList(result == null ? ImmutableList.of() : (List<?>)result)
                     .map(QueryResult::ofComputed);
```

normen662 (changes requested):
General comment: Please write a test case (probably in simple graph test) that actually creates a QGM with a nullable-array access in it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
pengpeng-lu: No, this is basically remove the "values" in the `KeyExpression`, so for structures like `a.values.b` we'll need to return `a.b`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
pengpeng-lu: No I don't think so, do you know which test class I can add unit test for this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
```diff
@@ -0,0 +1,129 @@
...
+    private NullableArrayTypeUtils() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    /**
+     * Get the fieldName of the repeated field.
+     *
+     * @return fieldName of the repeated field
+     */
+    public static String getRepeatedFieldName() {
```
normen662, line 49: please annotate all your methods/functions with nullability indicators. `@Nonnull` vs `@Nullable`. Parameters as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
normen662: What I mean is that there is no caller for this anymore is there? Whenever you are in need of doing this you are actually in the process of translating this into a query graph, so you should never actually have someone wanting a new key expression. TLDR: you should be able to remove this method altogether.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NullableArrayTypeUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                .setChild(child.getNesting().getChild());
+        return new NestingKeyExpression(newNestingBuilder.build());
+    }
+
+    public static FieldKeyExpression unwrapArrayInFieldKeyExpression(@Nonnull FieldKeyExpression fieldKeyExpression) {
+        return new FieldKeyExpression(fieldKeyExpression.toProto().toBuilder().setFanType(RecordMetaDataProto.Field.FanType.FAN_OUT).build());
+    }
+
+    @Nullable
+    public static Object unwrapIfArray(@Nullable Object wrappedValue, @Nonnull Type type) {
```
normen662, line 115: Javadoc please.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -311,13 +291,46 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
...
         throw new IllegalStateException("unable to translate protobuf descriptor to type");
     }
 
+    /**
+     * translates a repeated field in a protobuf to a {@link Array}.
+     * @param descriptor The protobuf descriptor.
+     * @param protoType The protobuf descriptor type.
+     * @return A {@link Array} object that corresponds to the protobuf {@link com.google.protobuf.Descriptors.Descriptor}.
+     */
+    private static Array fromProtoTypeToArray(@Nullable Descriptors.GenericDescriptor descriptor, @Nonnull Descriptors.FieldDescriptor.Type protoType, @Nonnull TypeCode typeCode, boolean needsWrapper) {
```
normen662, line 314: `@Nonnull`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -311,13 +291,46 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
...
+                    return Record.fromFieldDescriptorsMap(isNullable, Record.toFieldDescriptorMap(messageDescriptor.getFields()));
+                }
             }
         }
 
         throw new IllegalStateException("unable to translate protobuf descriptor to type");
     }
 
+    /**
+     * translates a repeated field in a protobuf to a {@link Array}.
```
normen662, line 309:
```suggestion
     * Translates a repeated field in a protobuf descriptor to a {@link Array}.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
normen662: Maybe you just have `needsWrapper()` to just do `return isNullable;`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
normen662: Remember that the old logic attempts to support arrays that can have `null`s in them. The new logic supports nullable arrays which cannot have `null`s in them. The approaches of how to get that done are eerily similar but not quite the same. I am ok with removing that code that supports the old logic or to completely leave it intact, but please no mix of the two.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
normen662: I don't think so. This feature comes from the Norse prototype but nothing really depends on that particular feature being there. We can remove that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
normen662: resolved offline

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -211,6 +212,18 @@ protected void openNestedRecordStore(FDBRecordContext context, @Nullable RecordM
         createOrOpenRecordStore(context, nestedMetaData(hook));
     }
 
+    protected void openNestedWrappedArrayRecordStore(FDBRecordContext context) throws Exception {
```
normen662, line 215: nullability on the context

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
pengpeng-lu: makes sense. I'll remove this part.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TypeRepositoryTest.java
```diff
@@ -199,30 +217,6 @@ void attemptToCreateArrayConstructorValueWithDifferentChildrenTypesFails() {
         }
     }
 
-    @Test
-    void createArrayConstructorValueWorks() {
-        final Typed value = new AbstractArrayConstructorValue.ArrayFn().encapsulate(null, List.of(INT_1, INT_2));
```
pengpeng-lu, line 204: This test doesn't apply any more, because we removed the code around null elements in array.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -150,8 +150,8 @@ void testInQueryNoIndex() throws Exception {
                             .where(predicates(valuePredicate(fieldValue("num_value_2"), new Comparisons.ListComparison(Comparisons.Type.IN, ImmutableList.of(0, 2))))));
 
             assertEquals(997592219, plan.planHash(PlanHashable.PlanHashKind.LEGACY));
-            assertEquals(-951957746, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
-            assertEquals(-739430123, plan.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS));
+            assertEquals(78097044, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
+            assertEquals(290624667, plan.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS));
```
pengpeng-lu, line 154: All hash changes are caused by changing `isNullable` of arrays from `true` to `false`. because right now `isNullable` represents the whole array is nullable.

normen662 (approved):

hatyo (approved):

PR #1723, opened 2022-06-10
Resolve #1725: Reconstruct KeyWithValue expression correctly.

### LINKED ISSUE #1725: Cascades does not reconstruct a KeyWithValue correctly
When Cascades expands a Index into a `KeyExpression`, there is some logic that examines the index's root expression keys and appends any missing primary keys to the end via concat. This works as expected for `ThenKeyExpression`.

However if the root expression is `KeyWithValue` such as e.g. (`KeyWithValue({RecordType, A, B} 2)`, where `A` is a primary key), then it will pass the `KeyWithValue` as-is to concat causing it to fail, because concat thinks there is only a single key to be concatenated (but in our example above, we have two keys: `RecordType` and `A`).

hatyo:
When Cascades expands a `Index` into a `KeyExpression`, there is some logic that examines the index's root expression keys and appends any missing primary keys to the end via `concat`. This works as expected for `ThenKeyExpression`.

However if the root expression is `KeyWithValue` such as e.g. (`KeyWithValue({RecordType, A, B} 2)`, where `A` is a primary key), then it will pass the `KeyWithValue` as-is to `concat` causing it to fail, because `concat` thinks there is only a single key to be concatenated (but in our example above, we have two keys: `RecordType` and `A`).

To fix this problem, we check first if the root expression is a `KeyWithValue`, if so, we get its key expression, and proceed with the same logic.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
@@ -156,13 +156,16 @@ public class ValueIndexExpansionVisitor extends KeyExpressionExpansionVisitor im
      */
     @Nonnull
     public static KeyExpression fullKey(@Nonnull Index index, @Nullable final KeyExpression primaryKey) {
+        final KeyExpression rootExpression = index.getRootExpression() instanceof KeyWithValueExpression
+                                             ? ((KeyWithValueExpression)index.getRootExpression()).getKeyExpression()
+                                             : index.getRootExpression();
         if (primaryKey == null) {
-            return index.getRootExpression();
+            return rootExpression;
         }
         final var trimmedPrimaryKeyComponents = new ArrayList<>(primaryKey.normalizeKeyForPositions());
         index.trimPrimaryKey(trimmedPrimaryKeyComponents);
         final var fullKeyListBuilder = ImmutableList.<KeyExpression>builder();
-        fullKeyListBuilder.add(index.getRootExpression());
+        fullKeyListBuilder.add(rootExpression);
         fullKeyListBuilder.addAll(trimmedPrimaryKeyComponents);
         return concat(fullKeyListBuilder.build());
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
```diff
@@ -156,13 +156,16 @@ public MatchCandidate expand(@Nonnull final Supplier<Quantifier.ForEach> baseQua
...
+                                             : index.getRootExpression();
         if (primaryKey == null) {
-            return index.getRootExpression();
+            return rootExpression;
         }
         final var trimmedPrimaryKeyComponents = new ArrayList<>(primaryKey.normalizeKeyForPositions());
         index.trimPrimaryKey(trimmedPrimaryKeyComponents);
         final var fullKeyListBuilder = ImmutableList.<KeyExpression>builder();
-        fullKeyListBuilder.add(index.getRootExpression());
+        fullKeyListBuilder.add(rootExpression);
```
normen662, line 168: This is wrong (and it used to be wrong). We should not just blindly use the `rootExpression` here either in the presence of `KeyWithValue`. I think we should also call `normalizeKeyForPositions()` on the `rootExpression` and then properly pick and merge the pieces from normalized root and from normalized primary..

normen662: Do you mind changing that? It's great you found this (admittedly quite old) problem...

normen662 (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
hatyo: It looks like there was some misunderstanding, after some discussion we found the fix is actually correct 😄 .

PR #1724, opened 2022-06-10
Populate primary key for partial records for Lucene auto-complete que…

tian-yizuo:
…ry results

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexEntry.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexEntry.java
@@ -80,9 +80,15 @@ public class IndexEntry {
 
     @API(API.Status.INTERNAL)
     public IndexEntry(@Nonnull Index index, @Nonnull Tuple key, @Nonnull Tuple value) {
+        this(index, key, value, null);
+    }
+
+    @API(API.Status.INTERNAL)
+    public IndexEntry(@Nonnull Index index, @Nonnull Tuple key, @Nonnull Tuple value, @Nullable Tuple primaryKey) {
         this.index = index;
         this.key = key;
         this.value = value;
+        this.primaryKey = primaryKey;
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexEntry.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexEntry.java
@@ -46,6 +46,11 @@ public class IndexEntry {
     private final Tuple key;
     @Nonnull
     private final Tuple value;
+
+    /**
+     * This primaryKey is either supplied in advance through the constructor, or determined by the index's primaryKeyComponentPositions lazily and memorized in it.
+     * Index maintainers are not obligated to supply this in advance, unless their supposed primary keys cannot be obtained from the {@link #key} tuple.
+     */
     @Nullable
     private Tuple primaryKey;
 
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
@@ -183,7 +183,7 @@ public class LuceneDocumentFromRecord {
         public void addField(@Nonnull T source, @Nonnull String fieldName, @Nullable final Object value,
                              LuceneIndexExpressions.DocumentFieldType type,
                              boolean stored, boolean sorted,
-                             @Nonnull List<Integer> overriddenKeyRanges, int groupingKeyIndex,
+                             @Nonnull List<Integer> overriddenKeyRanges, int groupingKeyIndex, int keyIndex,
                              @Nonnull Map<String, Object> fieldConfigs) {
             fields.add(new DocumentField(fieldName, value, type, stored, sorted, fieldConfigs));
         }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
@@ -216,6 +216,8 @@ public class LuceneIndexExpressions {
                                                                         @Nonnull T source, @Nonnull DocumentDestination<T> destination,
                                                                         @Nullable String fieldNamePrefix, int keyIndex, int groupingCount,
                                                                         @Nonnull List<Integer> overriddenKeyRanges) {
+        // Record type evaluation of primary key based on this expression for the partial record is not needed,
+        // because this partial record to build has the correct record type.
         if (expression instanceof RecordTypeKeyExpression) {
             return;
         }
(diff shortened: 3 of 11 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
```diff
@@ -183,7 +183,7 @@ public List<DocumentField> getFields() {
         public void addField(@Nonnull T source, @Nonnull String fieldName, @Nullable final Object value,
                              LuceneIndexExpressions.DocumentFieldType type,
                              boolean stored, boolean sorted,
-                             @Nonnull List<Integer> overriddenKeyRanges, int groupingKeyIndex,
+                             @Nonnull List<Integer> overriddenKeyRanges, int groupingKeyIndex, int keyIndex,
```
alecgrieser, line 186: Is the new `keyIndex` parameter actually used? I don't see a method that actually uses it any where

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
```diff
@@ -311,6 +324,7 @@ public void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Mess
             String value = (String) keyTuple.get(keyTuple.size() - 1);
             Tuple groupingKey = Tuple.fromList(keyTuple.getItems().subList(0, keyTuple.size() - 2));
             buildPartialRecord(kv.getIndex().getRootExpression(), recordDescriptor, recordBuilder, fieldName, value, groupingKey);
+            populatePrimaryKey(primaryKeyExpression, recordDescriptor, recordBuilder, kv.getPrimaryKey());
```
alecgrieser, line 327: It feels like there should be tests somewhere that validate that:

1. The `IndexEntry`s that get produced from auto-complete have primary keys. (Possibly verifying that the primary key matches expectations, possibly just validating that calling `IndexEntry::getPrimaryKey()` doesn't throw an error and returns something non-`null`)
2. The partial records that get produced copy over the primary key from the entry. I'm not sure if we have tests that do the partial record construction, but if we do, making sure the primary key is set correctly sounds like a good check to add

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneDocumentFromRecord.java
tian-yizuo: it's used here: https://github.com/FoundationDB/fdb-record-layer/pull/1724/files#diff-11ef9c019db17a199d457b1610f89a129c31d03a5b14c006f27fd1f1454ba5e8R77

alecgrieser: I see... The idea being that it uses the machinery to translate the primary key data as well.

A lot of this code seems to be duplicating code in the partial record creation code in `fdb-record-layer-core`, but I suppose refactoring this is something that can be done at a later date

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexKeyValueToPartialRecordUtils.java
tian-yizuo: @alecgrieser the tests for 1 can be found from ```LuceneIndexTest``` now. It actually verifies more than just ```getPrimaryKey()``` is null. The tests for 2 can be found from ```LuceneDocumentFromRecordTest```. And the test  ```searchForAutoCompleteForGroupedRecord()``` in ```LuceneIndexTest``` also verified the partial record.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -107,8 +107,19 @@ public static IndexKeyValueToPartialRecord getToPartialRecord(@Nonnull Index ind
...
-        builder.addRegularCopier(new LuceneIndexKeyValueToPartialRecordUtils.LuceneAutoCompleteCopier());
+        builder.addRegularCopier(new LuceneIndexKeyValueToPartialRecordUtils.LuceneAutoCompleteCopier(scanType.equals(LuceneScanTypes.BY_LUCENE_AUTO_COMPLETE),
+                recordType.getPrimaryKey()));
 
         return builder.build();
     }
+
+    /**
+     * Auto-Complete and Spell-Check scan has their own implementation for {@link IndexKeyValueToPartialRecord} to build partial records,
+     * so they are not appropriate for the optimization by {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryCoveringIndexPlan}.
```
alecgrieser, line 118: I believe the consequence of this design is that auto-complete queries will now get partial records that have:

1. Any grouping keys
2. The primary key
3. The completed text

And nothing else. This means that if the user wants other fields from the record, they have to go look up the record by primer key, I suppose. That...may be the best we can do here, in that there's sort of an amibiguity baked into the way that partial records are generated, where the auto-completed text can't be returned in the same record as other text fields but in a way that still allows the auto-completed text to be distinguished. (In SQL, say, this would probably be some pseudo-column added to the result set that was labeled as the result of the function evaluation, I think.)

At some points we should definitely document what the contract is here so that downstream users can reason about this special kind of scan, but this is probably fine for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
```diff
@@ -215,6 +216,10 @@ public static <T extends RecordSource<T>> void getFieldsRecursively(@Nonnull Key
                                                                         @Nonnull T source, @Nonnull DocumentDestination<T> destination,
                                                                         @Nullable String fieldNamePrefix, int keyIndex, int groupingCount,
                                                                         @Nonnull List<Integer> overriddenKeyRanges) {
+        if (expression instanceof RecordTypeKeyExpression) {
+            return;
```
alecgrieser, line 220: Why does this just return immediately on a `RecordTypeKeyExpression`? I would have though that this would need to return the associated record type's value of the record, but maybe that's also not quite right. It might actually be more of a suggestion that the logic here for extracting the copier for primary key expressions isn't quite right (e.g., it should share more code with the normal tuple-to-copier code for VALUE indexes instead of the Lucene index code)

tian-yizuo: Added comment

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
tian-yizuo: Added java doc for the method ```executePlan```.

alecgrieser (dismissed):

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/IndexEntry.java
```diff
@@ -80,9 +80,15 @@ public IndexEntry(@Nonnull Index index, @Nonnull Key.Evaluated key) {
 
     @API(API.Status.INTERNAL)
     public IndexEntry(@Nonnull Index index, @Nonnull Tuple key, @Nonnull Tuple value) {
+        this(index, key, value, null);
+    }
+
+    @API(API.Status.INTERNAL)
+    public IndexEntry(@Nonnull Index index, @Nonnull Tuple key, @Nonnull Tuple value, @Nullable Tuple primaryKey) {
```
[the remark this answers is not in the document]
tian-yizuo: @MMcM added comment above the field.

MMcM (approved):

alecgrieser (approved):

PR #1731, opened 2022-06-15
Resolve #1730: Fix double count

### LINKED ISSUE #1730: Fix double counting for remote fetch
Limit manager counts are incorrect: tryScan is called in both the KeyValueCursor and the KeyValueUnsplitter, creating a double count that skews limits.

ohadzeliger:
Remote fetch was double counting bytes and scanned records in the limit manager.
Fix by giving KeyValueUnsplitter an unlimited limit manager.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -1347,8 +1347,9 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         final RecordCursor<FDBRawRecord> rawRecords;
         if (metaData.isSplitLongRecords()) {
             // Note that we always set "reverse" to false since regardless of the index scan direction, the MappedKeyValue is in non-reverse order
+            // Setting the limit manager to UNTRACKED since the KeyValueCursor already counts records, and we don't want the unsplitter to double-count
             rawRecords = new SplitHelper.KeyValueUnsplitter(context, recordSubspace, rangeCursor, oldVersionFormat,
-                    sizeInfo, false, new CursorLimitManager(scanProperties));
+                    sizeInfo, false, CursorLimitManager.UNTRACKED);
         } else {
             if (omitUnsplitRecordSuffix) {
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1347,8 +1347,9 @@ private FDBRawRecord reconstructSingleRecord(final Subspace recordSubspace, fina
         final RecordCursor<FDBRawRecord> rawRecords;
         if (metaData.isSplitLongRecords()) {
             // Note that we always set "reverse" to false since regardless of the index scan direction, the MappedKeyValue is in non-reverse order
+            // Setting the limit manager to UNTRACKED since the KeyValueCursor already counts records, and we don't want the unsplitter to double-count
             rawRecords = new SplitHelper.KeyValueUnsplitter(context, recordSubspace, rangeCursor, oldVersionFormat,
-                    sizeInfo, false, new CursorLimitManager(scanProperties));
+                    sizeInfo, false, CursorLimitManager.UNTRACKED);
```
alecgrieser, line 1352: Does this need to do the same below (in the case of `!metaData.isSplitLongRecords` but where `omitUnsplitSuffix` is `false`)?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ohadzeliger: Clearly an issue. Trying to figure out why the test that was supposed to catch that (RemoteFetchWithoutSplitRecordsTest) didn't.

ohadzeliger: Test didn't pick this up because we are still in api version `6.3`by default. Locally running with 7.1 does fail.
Should we change the default version to 7.1 to reduce these kinds of issues (as we should be moving to supporting this version across the board)?

alecgrieser (approved):

PR #1744, opened 2022-06-21
Resolves #1743: Update FDB dependency to 7.1.10

### LINKED ISSUE #1743: Update FDB dependency to 7.1.10
The FDB Java dependency should be updated to version 7.1.10. There are a few bugs affecting things like index prefetch in earlier 7.1 versions. Strictly speaking, the FDB C client version (and the FDB server version) are the more important things, but increasing our declared Java dependency is probably good practice here.

alecgrieser:
The FDB Java version has been updated to 7.1.10 in this PR. The more important thing is actually that the client (and server) versions be updated to 7.1.10 or 7.1.11 (depending on whether the environments support AVX instructions) to get the recent bug fixes, but this seems like a good signal to send to let our downstream adopters know they should upgrade.

This resolves #1743.


ohadzeliger (dismissed):
One more file that we may consider is Dockerfile.standalone-tests. It is not required for production artifacts but may be a good idea to keep up to date anyways.

ohadzeliger (approved):

PR #1746, opened 2022-06-22
Resolves #1745: Lucene auto-complete should apply the limit to the re…

### LINKED ISSUE #1745: Lucene auto-complete should apply the limit to the result cursor
The flap map pipelined cursor may have more results than the limit due to multiple matching fields for one record. So the limit should be applied to the cursor as well.

tian-yizuo:
…sult cursor

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
@@ -177,7 +177,7 @@ public class LuceneAutoCompleteResultCursor implements BaseCursor<IndexEntry> {
         }
         long startTime = System.nanoTime();
 
-        lookupResults = lookup().skip(skip);
+        lookupResults = lookup().skip(skip).limitRowsTo(limit);
         if (timer != null) {
             timer.recordSinceNanoTime(LuceneEvents.Events.LUCENE_AUTO_COMPLETE_SUGGESTIONS_SCAN, startTime);
         }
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
```diff
@@ -177,7 +177,7 @@ private void performLookup() throws IOException {
         }
         long startTime = System.nanoTime();
 
-        lookupResults = lookup().skip(skip);
+        lookupResults = lookup().skip(skip).limitRowsTo(limit);
```
alecgrieser, line 180: I'm a little surprised that this passes the tests. This looks right to me, but the fact that none of the Lucene tests failed seems to suggest that we don't have any tests that validate that the limit is enforced, which could be a sign of a gap in our coverage.

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
MMcM: I also suspect that
```
searcher.search(finalQuery, limit)
```
in `lookup` is failing to take account of `skip` properly.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteResultCursor.java
tian-yizuo: @MMcM @alecgrieser  fixed the skip for ```searcher.search```. And added tests to cover this. The new added tests failed without this fix.

PR #1751, opened 2022-06-28
Resolves #1756: Remote fetch index scan wrapper

### LINKED ISSUE #1756: Create index scan with remote fetch
The remote fetch mechanism supports execution from an IndexPlan. A similar mechanism is needed using regular index scan.
A wrapper for an index scan that takes in a IndexFetchMethod and delegates to either regular index scan or a remote fetch index scan is needed.

ohadzeliger:
This PR creates a wrapper around index scan method: A new overload of `scanIndexRecords` that takes in an `IndexScanMethod` and delegates to the appropriate method. This is very similar to the work done in the `RecordQueryIndexPlan` (except that in the plan, the regular scan and fallback methods are managed by the plan and are not plain ValueIndex scans).
Added tests are a subset of the `RemoteFetchTests` adjusted for the new method instead of a plan execution.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
@@ -30,9 +30,11 @@ import com.apple.foundationdb.record.IndexScanType;
 import com.apple.foundationdb.record.IndexState;
 import com.apple.foundationdb.record.IsolationLevel;
 import com.apple.foundationdb.record.PipelineOperation;
+import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCoreStorageException;
 import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordFunction;
 import com.apple.foundationdb.record.RecordIndexUniquenessViolation;
 import com.apple.foundationdb.record.RecordMetaDataBuilder;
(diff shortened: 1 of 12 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -943,6 +951,67 @@ default RecordCursor<FDBIndexedRecord<M>> scanIndexRecords(@Nonnull final String
...
+     * @param commonPrimaryKey the common primary key for the records that would be returned, only required for USE_REMOTE_FETCH and USE_REMOTE_FETCH_WITH_FALLBACK
+     * @param continuation any continuation from a previous scan
+     * @param orphanBehavior how the iteration process should respond in the face of entries in the index for which there is no associated record
+     * @param scanProperties skip, limit and other scan properties
+     * @return a cursor that return records pointed to by the index
+     */
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    default RecordCursor<FDBIndexedRecord<M>> scanIndexRecords(@Nonnull final String indexName,
+                                                               @Nonnull final RecordQueryPlannerConfiguration.IndexFetchMethod fetchMethod,
```
MMcM, line 976: Would it make sense to move this `enum` into this class? It seems kind of abstraction-stretching to depend on something supposedly belonging to the query planner.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
ohadzeliger: Sure - I can make the enum standalone, so the same type can be used from both query and scan.

ohadzeliger: (Unless you find that inappropriate)

MMcM: Yes, that was the idea. It seems reasonable for the planner config to have store's scan setting, just not the other way around.

ohadzeliger: Done.

MMcM (approved):

PR #1755, opened 2022-06-28
Add missing Checkstyle WhitespaceAfter module.

MMcM:
Differs from default by allowing `TYPECAST`, on which there seems to be genuine divergence in the team. I am certainly willing to put that back and also fix everything to require the space, if we think consistency is more important.

Most of the edits involve generic type parameters, for which there isn't any easy way to allow it. So it seems simplest to just add the spaces.


alecgrieser (approved):
This LGTM. Regarding `TYPECAST`, it might be nice to have consistency, but I agree that some people seem to prefer `(Type)val`, and so I think not waiting to come to an agreement on that is fine (and therefore leaving the check off).

PR #1759, opened 2022-06-30
Resolves #1758: Add index scan plan with additional reads to load the RYW cache

### LINKED ISSUE #1758: Add index scan plan with additional reads to load the RYW cache
In order to support certain kinds of nested loop join queries, there's an opportunity to increase efficiency of an overall query by changing how index scans operate slightly. In particular, if you have a query that is executed many times in the same transaction with different bound parameters, then currently, each execution must make requests to read data from the database (as you might expect). However, in cases where these executions touch adjacent, non-overlapping ranges, there's an opportunity here to read a slightly larger range of data in earlier query executions, which will then cache data in an already-maintained cache in the FDB client, and then subsequent reads can use that cache to skip reading ranges across the network.

I had a draft PR (#1717) which tested out a few different possible ways of optimizing such queries, and the one I'm proposing in this GitHub issue had promising looking numbers in local testing (i.e., against a local FDB cluster): it didn't seem to degrade performance much if at all in cases where the cache was not helpful, but it could speed up cacheable queries (about 2x in local testing, though those numbers would be expected to shift if there is additional latency between the Record Layer client and the FDB server).

alecgrieser:
This adds a new `RecordQueryPlan` that has the same semantics (in terms of results returned, including continuations) as a `RecordQueryIndexPlan`, but that scans additional ranges for the purposes
of loading the FDB C client's read-your-writes cache. If the query is executed multiple times in the same transaction with different bound parameters, then loading this cache might result in performance
wins because query executions (after the first) can use the cache to avoid having to make additional requests to the database. At the same time, because FDB paginates range scans, the amount of additional
work that each scan has to do to load the cache should be bounded as well (to approximately one page of results per scan).

This resolves #1758.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -27,7 +27,7 @@ This release also updates downstream dependency versions. Most notably, the prot
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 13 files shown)


alecgrieser:
Closing in favor of #1774

PR #1760, opened 2022-06-30
Fix Gradle deprecations in anticipation of moving to Gradle 7.

MMcM:


gradle/testing.gradle
```diff
@@ -122,7 +122,7 @@ task testJar(type: Jar, dependsOn: testClasses) {
 
 configurations {
     tests {
-        extendsFrom testRuntime
+        extendsFrom testRuntimeOnly
     }
```
MMcM, line 126: This could be replaced by `testFixtures`, with some files moving there from `test`, but I aimed to keep changes to a minimum and limited to Gradle files.

MMcM:
With this branch,
```
./gradlew wrapper --gradle-version 7.4.2 --distribution-type all --gradle-distribution-sha256-sum e6d864e3b5bc05cc62041842b306383fc1fefcec359e70cebb1d470a6094ca82
```
seems to produce something that works, although there are a number of `implicit_dependency` warnings.

MMcM:
Closing in favor of #1764.

PR #1764, opened 2022-07-01
attempt to upgrade builds to gradle 7

ammolitor:
stepping through changes required to make build work with latest gradle 7.4.2


build.gradle
```diff
@@ -140,9 +132,16 @@ allprojects {
     }
 
     // Distribution
-    task createDistribution(type: Sync) {
-        // source directory depends on the subproject
-        into distDir
+    if(project.name.contains('shaded')) {
```
ammolitor, line 135: this feels ugly to me, is there a better way?

build.gradle
```diff
@@ -211,6 +210,7 @@ subprojects {
                 // exclude generated Protobuf classes
                 exclude "**/*Proto.java"
                 exclude "**/*ProtoV3.java"
+                exclude "**/*Protos.java"
```
alecgrieser, line 213: What happened here? Did the name of some of the protobuf files change, or where there always generated files suffixed with `Protos.java`

fdb-record-layer-core/fdb-record-layer-core.gradle
```diff
@@ -28,37 +28,37 @@ if (!hasProperty('coreNotStrict')) {
 }
 
 configurations {
-    repl.extendsFrom testRuntime
+    repl.extendsFrom testRuntimeOnly
 }
 
 dependencies {
-    compile project(':fdb-extensions')
+    api project(':fdb-extensions')
```
alecgrieser, line 35: I _think_ we would want this to be `implementation` instead of `api`, at least at some point, as I don't think the Record Layer needs to/should expose the extended FDB in its API. I'm also fine with keeping this as `api` for now to avoid accidentally disturbing anyone downstream. There are certainly some utilities in `fdb-extensions` that an adopter might be using, though they should arguably just add an explicit dependency.

fdb-record-layer-icu/fdb-record-layer-icu.gradle
```diff
@@ -22,22 +22,24 @@ apply from: rootProject.file('gradle/strict.gradle')
 
 def coreProject = ":${ext.coreProjectName}"
 dependencies {
-    compile project(coreProject)
-    compile "com.ibm.icu:icu4j:${icuVersion}"
-    compile "com.google.protobuf:protobuf-java:${protobufVersion}"
-    compile "org.slf4j:slf4j-api:${slf4jVersion}"
+    implementation project(coreProject)
```
alecgrieser, line 25: I feel like this should probably be `api` instead of `implementation`. This subproject provides an implementation of an interface defined in `fdb-record-layer-core`, so I think it needs to say that it takes on the core project as an `api` dependency, but I'm not actually certain.

fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
```diff
@@ -23,7 +23,7 @@ apply plugin: 'com.github.johnrengelman.shadow'
 def coreProject = ":${ext.coreProjectName}"
 
 dependencies {
-  compile project(coreProject)
+  implementation project(coreProject)
```
alecgrieser, line 26: Is this right for shaded artifacts (I don't actually know)? The relationship here is a little weird, in that a user who wants to use the shaded artifacts should take the `fdb-record-layer-core-shaded` artifact _instead_ of `fdb-record-layer-core`, so maybe `implementation` is correct?

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
```diff
@@ -26,21 +26,25 @@ if (!hasProperty('coreNotStrict')) {
...
 def coreProject = ":${ext.coreProjectName}"
 dependencies {
-    compile project(coreProject)
-    compile "org.apache.lucene:lucene-core:${luceneVersion}"
-    compile "org.apache.lucene:lucene-queryparser:${luceneVersion}"
-    compile "org.apache.lucene:lucene-analyzers-common:${luceneVersion}"
-    compile "org.apache.lucene:lucene-suggest:${luceneVersion}"
-    compile "org.apache.lucene:lucene-highlighter:${luceneVersion}"
-    testCompile project(path: coreProject, configuration: 'tests')
+    implementation project(coreProject)
```
alecgrieser, line 29: For similar reasons to `fdb-record-layer-icu`, I think this should be an `api` dependency

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
```diff
@@ -26,21 +26,25 @@ if (!hasProperty('coreNotStrict')) {
...
+    implementation "org.apache.lucene:lucene-core:${luceneVersion}"
+    implementation "org.apache.lucene:lucene-queryparser:${luceneVersion}"
+    implementation "org.apache.lucene:lucene-analyzers-common:${luceneVersion}"
+    implementation "org.apache.lucene:lucene-suggest:${luceneVersion}"
+    implementation "org.apache.lucene:lucene-highlighter:${luceneVersion}"
+    implementation "org.apache.commons:commons-lang3:${commonsLang3Version}"
```
alecgrieser, line 35: I _think_ this is right (that the `lucene` dependencies are supposed to be implementation only and not leak as a part of the API), but there's some risk here that we accidentally exposed something we shouldn't have. I'm fine with leaving this as `implementation`, though, and then fixing that later if we find something is amiss.

fdb-record-layer-spatial/fdb-record-layer-spatial.gradle
```diff
@@ -26,24 +26,27 @@ if (!hasProperty('coreNotStrict')) {
 
 def coreProject = ":${ext.coreProjectName}"
 dependencies {
-    compile project(coreProject)
-    compile "org.locationtech.jts.io:jts-io-common:${jtsVersion}"
-    compile "com.geophile:geophile:${geophileVersion}"
-    compile "com.google.protobuf:protobuf-java:${protobufVersion}"
-    compile "org.slf4j:slf4j-api:${slf4jVersion}"
-    compile "org.slf4j:jul-to-slf4j:${slf4jVersion}"
+    implementation project(coreProject)
```
alecgrieser, line 29: For similar reasons to `fdb-record-layer-icu`, I think this should be an `api` dependency

build.gradle
```diff
@@ -140,9 +132,16 @@ allprojects {
     }
 
     // Distribution
-    task createDistribution(type: Sync) {
-        // source directory depends on the subproject
-        into distDir
+    if(project.name.contains('shaded')) {
+        task createDistribution(type: Sync, dependsOn: 'shadowJar') {
```
alecgrieser, line 136: It does seem like this should be in `fdb-record-layer-shaded.gradle`. Something like: `createDistribution.dependsOn { shadowJar }`.

alecgrieser (changes requested):

build.gradle
MMcM: It was trying to Javadoc `DescriptorProtos.java`; perhaps that isn't the case any more after other improvements. Should confirm that it's still needed.

fdb-record-layer-core/fdb-record-layer-core.gradle
MMcM: I am happy with `api` for `fdb-java` and `implementation` for `fdb-extensions`. I think it depends on whether you think the "extended" FDB is an "enhanced" version or orthogonal additions.

fdb-record-layer-icu/fdb-record-layer-icu.gradle
MMcM: I am fine with that.

fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
MMcM: I think that was the reasoning. `implementation` get the shader access to the bits, which it then re-publishes, rather than passing through.

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
MMcM: Nothing seemed to break, so I think we did okay.

fdb-record-layer-spatial/fdb-record-layer-spatial.gradle
MMcM: Fine with me.

build.gradle
ammolitor: made this change, appears to work as desired. thanks.

build.gradle
MMcM: `gradle javadoc` does pass in this branch without that change, indeed suggesting that it isn't needed any more.

build.gradle
ammolitor: corrected per suggestion.

fdb-record-layer-core/fdb-record-layer-core.gradle
ammolitor: changed to `implementation` and added `fdb-extensions` (as `implementation`) to projects that needed it.

fdb-record-layer-icu/fdb-record-layer-icu.gradle
ammolitor: changed to `api`

fdb-record-layer-core-shaded/fdb-record-layer-core-shaded.gradle
ammolitor: left as-is

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
ammolitor: changed to `api`

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
ammolitor: leaving as-is

fdb-record-layer-spatial/fdb-record-layer-spatial.gradle
ammolitor: changed to api

alecgrieser (approved):

PR #1767, opened 2022-07-06
sign archives (prep for publish to central)

ammolitor:
add signing plugin to all projects
use environment variables for signing parameters
only sign if GPG_KEY_ID is not null


ammolitor:
Working through this guide: https://central.sonatype.org/publish/publish-gradle/

This PR adds signing of artifacts, I believe the signed artifacts need to be copied to the `.dist` directory to get published.

MMcM (approved):

alecgrieser (approved):

Issue #1771, opened 2022-07-07
FieldCopier.copy hotspots during large covered index reads


jleach4:
<img width="1776" alt="Screen Shot 2022-07-07 at 3 47 46 PM" src="https://user-images.githubusercontent.com/1850293/177884087-2788a660-55f5-4ebf-a55c-8522eee9ca13.png">

PR #1772, opened 2022-07-07
FieldCopier.copy hotspots during large covered index reads #1771

jleach4:
Screen Shot 2022-07-07 at 3 47 46 PM

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
@@ -102,11 +102,15 @@ public class IndexKeyValueToPartialRecord {
         @Nonnull
         private final TupleSource source;
         private final int index;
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -117,7 +119,6 @@ public void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Mess
             if (value == null) {
                 return;
             }
-            final Descriptors.FieldDescriptor fieldDescriptor = recordDescriptor.findFieldByName(field);
```
MMcM, line 120: I do not think that a `final` field descriptor will work with currently allowed use cases where the plan is cached and the metadata deserialized from the meta-data store anew. But a one-deep cache can probably get much of the same benefit. Keep the last `recordDescriptor.findFieldByName(field)`; compare its `getContainingType` next time and reuse if match, else replace.

MMcM (changes requested):

ScottDugas (approved):

PR #1773, opened 2022-07-08
Gradle 7 problems with IntelliJ compilation of client.

MMcM:
Release 3.2.278 was fine for dependent systems compiled with Gradle.
But adding a `@AutoService(javax.annotation.processing.Processor.class)` means all of its runtime implementation also needs to be `api` so that
IntelliJ IDEA can call some of it (to determine that it isn't something it cares about, presumably).

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/annotation/GenerateVisitorAnnotationProcessor.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/annotation/GenerateVisitorAnnotationProcessor.java
@@ -21,7 +21,6 @@
 package com.apple.foundationdb.annotation;
 
 import com.google.auto.service.AutoService;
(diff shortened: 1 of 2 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/annotation/GenerateVisitorAnnotationProcessor.java
```diff
@@ -122,9 +121,7 @@ public SourceVersion getSupportedSourceVersion() {
 
     @Override
     public Set<String> getSupportedAnnotationTypes() {
-        return ImmutableSet.<String>builder()
```
MMcM, line 125: This keeps Guava internal.

ammolitor (approved):

PR #1774, opened 2022-07-09
Resolves #1758: Add index scan plan with additional reads to load the RYW cache

### LINKED ISSUE #1758: Add index scan plan with additional reads to load the RYW cache

tian-yizuo:


MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1415,7 +1415,10 @@ private RecordQueryPlan planScan(@Nonnull CandidateScan candidateScan,
     private RecordQueryPlan valueScan(@Nonnull CandidateScan candidateScan,
                                       @Nullable ScanComparisons scanComparisons,
                                       boolean strictlySorted) {
-        return planScan(candidateScan, IndexScanComparisons.byValue(scanComparisons), strictlySorted);
+        IndexScanType scanType = this.configuration.valueIndexOverScanNeeded(candidateScan.index.getName())
```
MMcM, line 1418: A full scan is still a value scan; `index` might be null.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
tian-yizuo: @MMcM yeah I also noticed this and saw test failures with NPE. I have fixed this now.

MMcM (approved):

tian-yizuo:
test this please

MMcM (approved):

Issue #1777, opened 2022-07-12
EventKeeperTranslator$BaseEvent hotspots when getting counters


jleach4:
<img width="2304" alt="Screen Shot 2022-07-12 at 12 07 11 PM" src="https://user-images.githubusercontent.com/1850293/178574511-2d6e20eb-4b96-429e-ad43-0a3a9d056e1c.png">

PR #1778, opened 2022-07-12
EventKeeperTranslator hotspots when getting counters #1777

jleach4:
Screen Shot 2022-07-12 at 12 07 11 PM

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/EventKeeperTranslator.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/EventKeeperTranslator.java
@@ -130,6 +130,12 @@ class EventKeeperTranslator implements EventKeeper {
             return logKey;
         }
 
+        /**
+         * Compares on class and name.
+         *
+         * @param o Object
+         * @return equals
+         */
         @Override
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/EventKeeperTranslator.java
```diff
@@ -144,7 +146,7 @@ public boolean equals(final Object o) {
 
         @Override
         public int hashCode() {
-            return Objects.hash(name);
+            return hashCode;
```
MMcM, line 149: I actually suspect `name.hashCode` would be sufficient. `String.hashCode` isn't expensive: it has a cache. It's `Objects.hashCode`, which is unnecessary with a single (non-null) argument, that's taking the time.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/EventKeeperTranslator.java
jleach4: agreed.

PR #1783, opened 2022-07-13
Resolve #1782: Log on fallback

### LINKED ISSUE #1782: Add logging for remote fetch fallback
When falling back to fallback cursor, issue a log message for visibility

ohadzeliger:
Add log message when `FallbackCursor` triggers a fallback, for visibility.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FallbackCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FallbackCursor.java
@@ -26,6 +26,8 @@ import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordCursorVisitor;
 import com.apple.foundationdb.util.LoggableException;
+import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FallbackCursor.java
```diff
@@ -113,6 +117,9 @@ public CompletableFuture<RecordCursorResult<T>> onNext() {
                     inner.close();
                     inner = fallbackCursorSupplier.apply(lastSuccessfulResult);
                     nextResultFuture = inner.onNext();
+                    if (LOGGER.isInfoEnabled()) {
+                        LOGGER.info("fallback triggered", throwable);
```
MMcM, line 121: Is the stack trace from `throwable` important? It could be rather large, right?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FallbackCursor.java
ohadzeliger: Done

MMcM (approved):

PR #1789, opened 2022-07-15
Remove the configuration support for Lucene field's OmitNorms parameter

tian-yizuo:


jleach4:
I was thinking you can keep the option but have the default as omitNorms(true);

MMcM (approved):

MMcM (approved):

PR #1796, opened 2022-07-20
Resolve #1791: Support planning logical group by expression

### LINKED ISSUE #1791: Support planning logical group by expression
The physical layer supports group by expression via streaming aggregates. However we miss planning logical group by expressions. To do this, we could introduce a logical group by expression and add some rules to transform it into the physical operator.

In the first version, the logical group by expression will sit on top of select expression and have a sorting requirement of the grouping column(s) that is pushed down to underlying operators forcing the planner to choose an access path that is compatibly-sorted with the grouping column(s). This means that group by expressions can work only when we have e.g. a value index that is compatibly-sorted. Then we could add another rule that transforms the logical group by expression into a streaming aggregation physical operator which will effectively stream the underlying access path and calculate the groups according to the aggregation function(s).

hatyo:
This allows Cascades to plan QGM plans  involving a `group by`, for such a plan Cascades explores part of the search space limited to access path that are compatibly-ordered with the grouping columns. If one is found, it will implement it by instantiating and setting up the physical operator `RecordQueryStreamingAggregationPlan`.

Here are some details:

* A new logical operator `GroupByExpression` is introduced to encapsulate information about the grouping: it consists of two expressions: the grouping expression, and the aggregation expression, the constituents of these expressions are to be used by the upper (select) expressions. It also includes aliases for both of these expressions, the aliases are basically placeholders glueing together the values produced `RecordQueryStreamingAggregationPlan` at runtime together with operators referencing (parts of) these values on top.
* Two rules are added to Cascades: `PushRequestedOrderingThroughGroupByRule` which takes care of pushing ordering requirements down the plan to limit the search space exploration, and `ImplementGroupByRule` which takes care of implementing the `RecordQueryStreamingAggregationPlan` on top of a compatibly ordered plan partition.
* Added a test that creates a logical `group by` plan and examines what Cascades will generate, this required implementing extra binding matchers and extractors to enable checking the physical plan output.
* Removed `Scopes` and `ParserContext` since they are not very relevant to RecordLayer.


hatyo:
Resolves #1791

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
```diff
@@ -368,6 +369,12 @@ public Cardinalities visitFullUnorderedScanExpression(@Nonnull final FullUnorder
         return Cardinalities.unknownCardinalities();
     }
 
+    @Nonnull
+    @Override
+    public Cardinalities visitGroupByExpression(@Nonnull final GroupByExpression element) {
+        return Cardinalities.unknownCardinalities();
```
normen662, line 375: You should at least do the case where you have no grouping columns or a constant grouping column and set:
- `min` and `max` to `1` if you don't have grouping columns (greeting from the count bug)
- `min` to `0` and `max` to `1` if you have a constant grouping expression

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
```diff
@@ -368,6 +369,12 @@ public Cardinalities visitFullUnorderedScanExpression(@Nonnull final FullUnorder
...
+    }
+
```
normen662, line 377: Can you check, and if needed update the implementation for streaming aggs here as well?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementGroupByRule.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.AnyMatcher.any;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.QuantifierMatchers.forEachQuantifierOverRef;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.RelationalExpressionMatchers.groupByExpression;
+
+/**
+ * Rule for implementing logical {@code GROUP BY} into a physical streaming aggregate operator {@link RecordQueryStreamingAggregationPlan}.
+ */
+public class ImplementGroupByRule extends PlannerRule<GroupByExpression> {
```
normen662, line 53: I started using a name convention that uses the name of the `RecordQueryPlan` instead of the `RelationExpression` it is replacing. So `ImplementNestedLoopJoin` instead of `ImplementSelectExpression`. That has the advantage that you can name two different implementations: `ImplementStreamingAggregation` and e.g. `ImplementHashAggregation`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementGroupByRule.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public void onMatch(@Nonnull final PlannerRuleCall call) {
+        final var bindings = call.getBindings();
+
+        final var groupByExpression = bindings.get(root);
+        final var requestedOrderings = Set.of(groupByExpression.getOrderingRequirement());
+        final var innerQuantifier = Iterables.getOnlyElement(groupByExpression.getQuantifiers());
+        final var innerReference = innerQuantifier.getRangesOver();
+        final var planPartitions = PlanPartition.rollUpTo(innerReference.getPlanPartitions(), OrderingProperty.ORDERING);
```
normen662, line 75: You can put this into the pre of the rule so you don't have to iterate over it. We sometimes do it for other reasons but I think in your case it's perfectly fine to use an approach as e.g. in `ImplementIntersectionRule`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PushRequestedOrderingThroughGroupByRule.java
```diff
@@ -0,0 +1,105 @@
...
+        ImmutableSet.Builder<RequestedOrdering> result = ImmutableSet.builder();
+        for (final var requestedOrdering : requestedOrderings) {
+            if (requestedOrdering.getDistinctness() == RequestedOrdering.Distinctness.PRESERVE_DISTINCTNESS) {
+                // ignore preserve as it might cause e.g. an underlying FUSE operator to be substituted with an access path that is incompatibly-ordered.
+                continue; // ignore
+            }
+            // todo this looks too strict, if requested ordering is more specific (e.g. (a,b,c)) than the requested group by ordering (e.g. (a,b)) should still be accepted.
+            if (!new LinkedHashSet<>(requestedOrdering.getOrderingKeyParts()).equals(groupByOrderingAsSet)) {
+                // case 2: if an incompatible ordering is found, fail.
+                return Set.of(); // fail
```
normen662, line 96: Please use `ImmutableSet` for now as everything in the planner does. We need to switch over at some point I assume but it would have to be done in one fell swoop. The danger of mixing and matching is that both `copyOf()` methods (for `ImmutableSet` in guava, for `Set` in the JDK) have their own bypass logic if the set being passed in is already of type `ImmutableSet` respectively `Set`. If we start mixing and matching we will see a whole bunch of copying going on coming out of nowhere.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -169,4 +170,9 @@ private static Type resolveTypeForPath(@Nonnull final Type inputType, @Nonnull f
         }
         return currentType;
     }
+
+    @Nonnull
+    public KeyExpression toKeyExpression() {
```
normen662, line 175: This is dangerous as the `FieldValue` as it only captures the path contained in here, i.e. the source the field value operates on may not be the root message. I would suggest to either allow a prefix to be passed in (which then can be empty) or to not put his code here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1159,6 +1160,11 @@ public int getFieldIndex() {
                 return getFieldIndexOptional().orElseThrow(() -> new RecordCoreException("field index should have been set"));
             }
 
+            @Nonnull
+            public KeyExpression toKeyExpression() {
```
normen662, line 1164: This seems odd to have that here as a `KeyExpression` is a thing that computes something and a `Type` is something that tells you what a. thing looks like. By the same rational you could have a `toFieldValue()` here which is equally odd.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/OrdinalFieldValue.java
```diff
@@ -94,11 +96,19 @@ public String explain(@Nonnull final Formatter formatter) {
...
     @Override
     public <M extends Message> Object eval(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final EvaluationContext context) {
-        final Message childMessage = (Message)child.eval(store, context);
-        if (childMessage == null) {
+        final var result = child.eval(store, context);
+        if (result == null) {
             return null;
         }
-        return MessageValue.getFieldOnMessage(childMessage, field.getFieldIndex());
+        if (result instanceof QueryResult) {
```
normen662, line 103: This cannot happen as discussed. This is always a message.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -92,6 +92,7 @@ public class RecordQueryStreamingAggregationPlan implements RecordQueryPlanWithC
     @Nonnull
     private final Value completeResultValue;
 
+
```
normen662, line 95: nitpick: remove empty line

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -346,14 +346,34 @@ public static RecordQueryStreamingAggregationPlan of(@Nonnull final Quantifier.P
...
+        return RecordQueryStreamingAggregationPlan.of(inner,
                 groupingKeyValue,
                 aggregateValue,
                 groupingKeyAlias,
                 aggregateAlias,
                 completeResultValueSupplier.supply(groupingKeyValue, aggregateValue, groupingKeyAlias, aggregateAlias));
     }
 
+    @Nonnull
+    public static RecordQueryStreamingAggregationPlan of(@Nonnull final Quantifier.Physical inner,
```
normen662, line 358: I think this is the best way of doing this. Do get into the hair of flattened result vs nested result.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/OrdinalFieldValue.java
hatyo: reverted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementGroupByRule.java
hatyo: I did some refactoring to handle the case where a `GroupByExpression`'s `groupingValue` is `null` causing it (now `Optional) `getOrderingRequirement` to become `empty`.

In this case, I made it such that we `yield` for each `planPartition` since we do not really care about ordering anymore as we want to put all tuples in a single large group.

In light of that, I kept the logic as-is, could you please check it again? I am interested whether we can still pull this part and put it in the pre-part of the rule.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
hatyo: You're right, I removed this method from both `Type.Record.Field` and `FieldValue`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
hatyo: Great observation, thank you!

hatyo: Regarding the second case (constant grouping expression). I will take care of it once I am able to bring https://github.com/FoundationDB/fdb-record-layer/pull/1819 in since it provides an easy way of checking whether a `Value` can be evaluated at compile time (I am assuming compile-time === constant although this is generally not true, e.g. `rand()`, but I think that's a separate work stream). Do you agree?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
hatyo: Yes, there was a stub implementation for `CardinalitiesProperty#visitRecordQueryStreamingAggregationPlan` that returns, similar to the above `Cardinalities.unknownCardinalities()`. I modified it in a similar way to the above suggestion for `GroupByExpression`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
normen662: 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -0,0 +1,273 @@
...
+
+    @Override
+    public int getRelationalChildCount() {
+        return 1;
+    }
+
+    @Nonnull
+    @Override
+    public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
+        return ImmutableSet.of();
```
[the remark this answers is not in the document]
normen662: This seems right.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementStreamingAggregationRule.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public void onMatch(@Nonnull final PlannerRuleCall call) {
+        final var bindings = call.getBindings();
+
+        final var groupByExpression = bindings.get(root);
+        final var requestedOrdering = groupByExpression.getOrderingRequirement();
+        final var innerQuantifier = Iterables.getOnlyElement(groupByExpression.getQuantifiers());
+        final var innerReference = innerQuantifier.getRangesOver();
+        final var planPartitions = PlanPartition.rollUpTo(innerReference.getPlanPartitions(), OrderingProperty.ORDERING);
```
normen662, line 73: My point here was that you can move this `rollUpTo` into the pre of the rule. Sometimes it's better here and sometimes it's better in the pre.

normen662 (approved):

Issue #1800, opened 2022-07-22
ContextRestoringExecutor Hotspots during beforeTask and endTask


jleach4:
<img width="1786" alt="Screen Shot 2022-07-22 at 11 17 06 AM" src="https://user-images.githubusercontent.com/1850293/180499998-4dc86033-a12c-4a74-8fe7-bdc9159f9fa0.png">

<img width="1786" alt="Screen Shot 2022-07-22 at 11 17 45 AM" src="https://user-images.githubusercontent.com/1850293/180500047-cc0b5900-2867-4aaf-9a6c-7e559bd4ad88.png">

PR #1801, opened 2022-07-22
ContextRestoringExecutor Hotspots during beforeTask and endTask #1800

jleach4:
Screen Shot 2022-07-22 at 11 17 45 AM

Screen Shot 2022-07-22 at 11 17 06 AM

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/ContextRestoringExecutor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/ContextRestoringExecutor.java
@@ -56,15 +56,14 @@ class ContextRestoringExecutor extends TaskNotifyingExecutor {
     }
 
     static void restoreMdc(@Nonnull Map<String, String> mdcContext) {
-        MDC.clear();
-        for (Map.Entry<String, String> entry : mdcContext.entrySet()) {
-            MDC.put(entry.getKey(), entry.getValue());
-        }
+        MDC.setContextMap(mdcContext);
     }
 
     static void clearMdc(@Nonnull Map<String, String> mdcContext) {
+        Map<String, String> map = MDC.getMDCAdapter().getCopyOfContextMap();
         for (String key : mdcContext.keySet()) {
-            MDC.remove(key);
+            map.remove(key);
         }
+        MDC.setContextMap(map);
     }
 }


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/ContextRestoringExecutor.java
```diff
@@ -56,15 +56,14 @@ public Map<String, String> getMdcContext() {
...
-            MDC.put(entry.getKey(), entry.getValue());
-        }
+        MDC.setContextMap(mdcContext);
     }
 
     static void clearMdc(@Nonnull Map<String, String> mdcContext) {
+        Map<String, String> map = MDC.getMDCAdapter().getCopyOfContextMap();
         for (String key : mdcContext.keySet()) {
-            MDC.remove(key);
+            map.remove(key);
```
jjezra, line 65: Would it be possible/efficient to use something like: 
```
MDC.keySet().removeAll(mdcContext.keySet());
```

jleach4: no

jleach4: It is a weird implementation...

tian-yizuo (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/ContextRestoringExecutor.java
jjezra: https://stackoverflow.com/questions/17675804/remove-multiple-keys-from-map-in-efficient-way

MMcM: This is the limitation of what SLF4J 1.7 can do: https://github.com/qos-ch/slf4j/blob/e9ee55cca93c2bf26f14482a9bdf961c750d2a56/slf4j-api/src/main/java/org/slf4j/spi/MDCAdapter.java

This code is from f5e1b6d23fa18f1c9fa9bfb24ce01ce6cdc758d5 (2017-04-12), when I believe it was still log4j 1. There was some discussion then about advocating for a better API, but nothing came of it. Nothing much has changed in SLF4J 2.

MMcM (approved):

jjezra (approved):

alecgrieser (approved):

alecgrieser:
It might be a good idea to add a release note (probably under the "performance" section) so we know which RL release this improvement ends up in, but otherwise, LGTM

PR #1803, opened 2022-07-25
Resolve #1808: Add NULL to BinaryPhysicalOperator and UnaryPhysicalOperator

### LINKED ISSUE #1808: Add support of null value in comparison operators
support null in comparison operators (=, !=, <, >, <=, >=, is null, is not null)

pengpeng-lu:
* This PR enables comparison operators to work with NULLs.


hatyo (changes requested):
It looks good but it is missing some tests. Please add tests for these new cases in `BooleanValueTest`.

hatyo (approved):
LGTM

PR #1805, opened 2022-07-25
Resolve #1804: Support remote fetch when no primary key can be calculated, remove common primary key from store interface.

### LINKED ISSUE #1804: Add support for remote fetch when there is no common primary key
When common primary key for an index cannot be derived (e.g. when the PKs for the types are not equal) but the PK length is the same, remote fetch can still be useful as the only thing needed to construct the mapped range is the length of the PK.
Add support for this case and remove the commonPrimary Key from the store interface (since it can be calculated based off of the types the index is defined for.

ohadzeliger:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -475,6 +475,27 @@ public static KeyExpression commonPrimaryKey(@Nonnull Collection<RecordType> rec
...
+     */
+    public static int commonPrimaryKeyLength(@Nonnull Collection<RecordType> recordTypes) {
+        int common = 0;
+        boolean first = true;
+        for (RecordType recordType : recordTypes) {
+            if (first) {
+                common = recordType.getPrimaryKey().getColumnSize();
+                first = false;
+            } else if (common != recordType.getPrimaryKey().getColumnSize()) {
+                return 0;
```
alecgrieser, line 493: Instead of returning 0 when a match isn't found, this feels like it should either return -1 or the whole method should return an `OptionalInt`. I think a zero-length primary key might actually be possible (with an `EMPTY` key expression), or at least, it seems safer to act like it could here, just in case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -976,12 +1000,80 @@ default RecordCursor<FDBIndexedRecord<M>> scanIndexRecords(@Nonnull final String
...
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    default RecordCursor<FDBIndexedRecord<M>> scanIndexRecords(@Nonnull final Index index,
+                                                               @Nonnull final IndexFetchMethod fetchMethod,
+                                                               @Nonnull final IndexScanBounds scanBounds,
+                                                               int commonPrimaryKeyLength,
+                                                               @Nullable byte[] continuation,
+                                                               @Nonnull IndexOrphanBehavior orphanBehavior,
+                                                               @Nonnull ScanProperties scanProperties) {
+        if ((fetchMethod != IndexFetchMethod.SCAN_AND_FETCH) && (commonPrimaryKeyLength <= 0)) {
```
alecgrieser, line 1074: Is there a reason that this doesn't remove the `commonPrimaryKeyLength` parameter from here, even, and instead  calculate the `commonPrimaryKeyLength` as the first thing in this function? Arguably, it could be pushed even further down into `scanIndexRemoteFetch`, if we wanted to, I suppose.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
ohadzeliger: This is what the signature at line 1031 does. The reason for exposing both the "with" length and the "without" length methods is so that in the cases where the PK length is known (pre calculated) it can be passed in and not re-calculated.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
ohadzeliger: Done.

alecgrieser (approved):

alecgrieser:
Oh, whoops, after merging, I just noticed that this is missing a release note

PR #1806, opened 2022-07-26
Fix Lucene optimised index searcher

hatyo:
In previous implementation there was a potential for a deadlock which could happen as the following:

- Let's assume we have a `FDBRecordContextConfig` configured with an `ExecutorService` using a thread pool containing 2 threads, (`T1`, `T2`) , execution a `LuceneScanIndex` query starts.
- `LuceneRecordCursor` calls `searchForTopDocs`, this method invokes [`IndexSearcher#search`](https://github.com/apache/lucene/blob/2cf12b8cdcc629617b2d58c0a2a6336679ff9249/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L651) method which will do, among other things, the following:

```java
for (int i = 0; i < leafSlices.length - 1; ++i) {
    final LeafReaderContext[] leaves = leafSlices[i].leaves;
    final C collector = collectors.get(i);
    FutureTask<C> task = new FutureTask<>(() -> {
        search(Arrays.asList(leaves), weight, collector);
        return collector;
    });
    executor.execute(task); // [*]
    topDocsFutures.add(task);
}
```

at `[*]` it could be that we have some tasks scheduled, let's say 3 tasks (Task1, Task2, Task3) in total. Let's say `T1` now works on `Task1` and `T2` starts working on `Task2`.

Execution service status: `{Task1 (running T1), Task2 (running T2), Task3 (waiting)}`.

The call to `search` on that line will delegate to [`LuceneOptimizedIndexSearcher#search`](https://github.com/FoundationDB/fdb-record-layer/blob/2f0b6e7de2fdabee1579f9969a958e145852c003/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java#L86), which might schedule other tasks on the same executor:

```java
leaves.forEach(ctx -> leafScorers.put(ctx, CompletableFuture.supplyAsync(() -> {
    try {
        return Pair.of(collector.getLeafCollector(ctx), weight.bulkScorer(ctx));
    } catch (IOException e) {
        return null;
    }
}, getExecutor())));
```

Let's stay `Task1` spawns `Task4` and `Task5`, while `Task2` spawns `Task6`.

Execution service status: `{Task1 (running T1), Task2 (running T2), Task3 (wait), Task5 (wait), Task6 (wait), Task7 (wait)}`.

The deadlock happens later on in that method when _all_ worker threads block waiting for the newly spawned sub-tasks to finish:

```java
for (LeafReaderContext ctx : leaves) {
    try {
        Pair<LeafCollector, BulkScorer> scorer = leafScorers.get(ctx).join(); [**]
        if (scorer != null && scorer.getLeft() != null && scorer.getRight() != null) {
            scorer.getRight().score(scorer.getLeft(), ctx.reader().getLiveDocs());
        }
    } catch (CollectionTerminatedException ioe) {
        // No Op
    }
}
```

Execution service status: `{Task1 (blocking T1, wait Task5 and Task6), Task2 (blocking T2, wait Task7), Task3 (waiting), Task5(wait), Task6 (wait), Task7 (wait)}`.

All worker threads are busy waiting for some tasks scheduled to their executor service => deadlock.

To solve this problem, we can model these dependencies via [`CompletableFuture#allOf`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-), this would require overriding `search(Query query, CollectorManager<C, T> collectorManager)` instead so we can control how the spawned tasks are modelled and scheduled.

In the new version, the `LuceneOptimizedIndexSearcher` will first run all of the `collector` scoring sub-tasks on parallel using `allOf`, and instead of blocking on them with `join`, it will use `thenApplyAsync` to synchronise the execution.

This way, we avoid the deadlock scenario, as a proof, we repeat the test `threadedLuceneScanDoesntBreakPlannerAndSearch` two times, one using 10 threads (which was sporadically failing before), and another time using 1 thread (which consistently deadlocks), in both cases, the test now passes uses this new approach.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
@@ -26,21 +26,26 @@ import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
-import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.BulkScorer;
-import org.apache.lucene.search.CollectionTerminatedException;
 import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Weight;
+import org.apache.lucene.util.ThreadInterruptedException;
+
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
 
 import static org.apache.lucene.search.QueryHelper.getQueriesTerms;
 
@@ -63,7 +68,7 @@ public class LuceneOptimizedIndexSearcher extends IndexSearcher {
     /**
      * Lower-level search API.
      *
-     * This is optimized into an attempt to parallelize database access in FDB
+     * This is optimized into an attempt to parallelize database access in FDB, for each collector in {@code collectorManager}.
      *
      * <p>
      * {@link LeafCollector#collect(int)} is called for every document. <br>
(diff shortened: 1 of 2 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -72,47 +77,92 @@ public LuceneOptimizedIndexSearcher(final IndexReader r, final Executor executor
...
+                final List<CompletableFuture<Pair<LeafReaderContext, Pair<LeafCollector, BulkScorer>>>> dependencies = Arrays.stream(leaves).map(ctx -> CompletableFuture.supplyAsync(() -> {
+                    try {
+                        final var result = Pair.of(collector.getLeafCollector(ctx), weight.bulkScorer(ctx));
+                        return Pair.of(ctx, result);
+                    } catch (IOException e) {
+                        return null;
                     }
-                } catch (CollectionTerminatedException ioe) {
-                    // No Op
+                }, getExecutor())).collect(Collectors.toList());
```
MMcM, line 130:
```suggestion
                }, executor)).collect(Collectors.toList());
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -72,47 +77,92 @@ public LuceneOptimizedIndexSearcher(final IndexReader r, final Executor executor
...
+                                if (scorer != null && scorer.getLeft() != null && scorer.getRight() != null) {
+                                    try {
+                                        scorer.getRight().score(scorer.getLeft(), ctx.reader().getLiveDocs());
+                                    } catch (IOException e) {
+                                        // no-op just ignore.
+                                    }
+                                }
+                            });
+                    return collector;
+                }, getExecutor());
```
MMcM, line 147:
```suggestion
                }, executor);
```

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -72,47 +77,92 @@ public LuceneOptimizedIndexSearcher(final IndexReader r, final Executor executor
...
+                        return Pair.of(ctx, result);
+                    } catch (IOException e) {
+                        return null;
                     }
-                } catch (CollectionTerminatedException ioe) {
-                    // No Op
+                }, getExecutor())).collect(Collectors.toList());
+
+                // 2. once we finish processing all of them, we can proceed with the final scoring task.
+                final var future = CompletableFuture.allOf(dependencies.toArray(new CompletableFuture[dependencies.size()])).thenApplyAsync(ignored -> {
```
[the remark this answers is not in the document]
hatyo: I wasn't aware of that, I will use it, thank you.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
hatyo: done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
hatyo: done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -72,47 +77,92 @@ public LuceneOptimizedIndexSearcher(final IndexReader r, final Executor executor
...
+     * {@code collectorManager}.
+     *
+     * @param query
+     *          the search query
+     * @param collectorManager
+     *          manager of collectors that receive hits
      */
     @Override
-    @SuppressWarnings("PMD")
-    protected void search(List<LeafReaderContext> leaves, Weight weight, Collector collector)
```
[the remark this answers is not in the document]
hatyo: > This method that used to be overridden expects to be called from the executor pool and does not itself ever use the executor (in the parent implementation), correct?

Yes that's correct.

> Moreover, all the interesting call paths are from the method that we are now overriding instead, right?

Great catch, I just found another invocation of this method from yet another overload `search(Query, Collector)`, so I did some refactoring and I've overriden it in `LuceneOptimizedIndexSearcher` to use the optimised version.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -115,44 +139,12 @@ public <C extends Collector, T> T search(Query query, CollectorManager<C, T> col
...
-                }, getExecutor());
+                final CompletableFuture<C> future = searchOptimized(executor, weight, Arrays.asList(leaves), collector);
                 topDocsFutures.add(future);
             }
-
-            final LeafReaderContext[] leaves = leafSlices[leafSlices.length - 1].leaves;
-            final C collector = collectors.get(leafSlices.length - 1);
-            // execute the last on the caller thread
-            search(Arrays.asList(leaves), weight, collector);
-            topDocsFutures.add(CompletableFuture.completedFuture(collector));
```
hatyo: This was inherited from Lucene's implementation, but I find it more consistent to delegate work to the desginated thread pool. What's your thoughts on this @MMcM?

MMcM: I thought the same thing. It seems silly to not just make a consistent set of futures. Who cares if the calling thread blocks on them all instead of (first) doing one of them? On the other hand, I do see the benefit of (mostly) copying the code for future maintainability.

MMcM (approved):

PR #1807, opened 2022-07-26
Resolve #1718: Return null when a field is not set

### LINKED ISSUE #1718: Return null when a non-repeated field is not set
Currently in OrdinalFieldValue and QuantifiedColumnValue classes, when a field is not set, it is returned as its default value. In this case null should be returned.

pengpeng-lu:
* For repeated field, it returns empty array if not set.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
@@ -96,7 +96,7 @@ public class MessageValue {
         }
         if (field.hasDefaultValue() || message.hasField(field)) {
             if (field.getJavaType() == Descriptors.FieldDescriptor.JavaType.MESSAGE &&
-                TupleFieldsHelper.isTupleField(field.getMessageType())) {
+                    TupleFieldsHelper.isTupleField(field.getMessageType())) {
                 return TupleFieldsHelper.fromProto((Message)message.getField(field), field.getMessageType());
             } else {
                 return message.getField(field);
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
```diff
@@ -74,6 +75,17 @@ public static Object getFieldValue(@Nonnull MessageOrBuilder message, @Nonnull L
...
     public static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull String fieldName) {
         final Descriptors.FieldDescriptor field = findFieldDescriptorOnMessage(message, fieldName);
+        return getFieldOnMessage(message, field);
+    }
+
+    @Nullable
+    public static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, int fieldNumber) {
+        final Descriptors.FieldDescriptor field = findFieldDescriptorOnMessage(message, fieldNumber);
+        return getFieldOnMessage(message, field);
+    }
```
hatyo, line 85: It would be really nice to add some documentation if possible (same for other newly-added methods).

hatyo (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageValue.java
```diff
@@ -74,6 +75,25 @@ public static Object getFieldValue(@Nonnull MessageOrBuilder message, @Nonnull L
...
+     * @return the value of the field as described above
+     */
+    @Nullable
+    public static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, int fieldNumber) {
+        final Descriptors.FieldDescriptor field = findFieldDescriptorOnMessage(message, fieldNumber);
+        return getFieldOnMessage(message, field);
+    }
+
+    @Nullable
+    private static Object getFieldOnMessage(@Nonnull MessageOrBuilder message, @Nonnull Descriptors.FieldDescriptor field) {
```
normen662, line 96: Wouldn't you have to unnest a wrapped value here?

normen662 (changes requested):

normen662 (changes requested):
So the thing that is missing here is that we want to unnest the wrapped value if it was in fact nested for this null-valued collections. Also if we do that we will also need to wrap it back into it's nested form for `RecordConstructorValue`s. Sorry, if you wanted to keep this separated into a different PR.

normen662 (approved):

PR #1810, opened 2022-07-27
Resolve #1809: Trigger a log message for every index state change

### LINKED ISSUE #1809: Trigger a log message for every index state change
Hunting a specific issue, it seems that some indexes states had been changed from WRITE_ONLY to READABLE without any log message. Since it's a rare event, it seems harmless to add a low level one.

jjezra:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -771,7 +771,14 @@ public Subspace indexSubspaceFromMaintainer(@Nonnull Index index) {
     }
 
     @Nonnull
-    public Subspace indexStateSubspace() {
+    public Subspace indexStateSubspace(String indexName, String callingMethod, String targetState) {
+        if (LOGGER.isInfoEnabled()) {
```
ScottDugas, line 775: Why are you adding this here, rather than on `updateIndexState`. Doing it there would remove the unnecessary `indexStateSubspace("<none>", "deleteAllRecords", "<none>")` in `deleteAllRecords`.
To me, that also seems more semantically correct. I'd be surprised if anyone is using this method outside of record-layer-core, but, if they are, there's a good chance they're not using it to change the index state.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: I wanted to expose anyone who access the state space (directly or via `updateIndexSpace`). But I don't mind changing it.

ScottDugas (approved):

PR #1816, opened 2022-08-09
Add tests for the behavior of the Lucene planner when faced with various combinations …

MMcM:
… of `AND` and `OR` queries and separate / combined Lucene indexes for the clauses' fields.

Enhance `planOther` so that intersection of compatibly (meaning, explicitly) ordered scans is attempted.

Add a planner configuration setting (default off) that gives the Lucene planner a chance at the whole filter before normalization, `OR`, and `IN` handling. With this on, `AND` and `OR` are handled symmetrically: using a combined Lucene boolean query when a single index is available with both requested fields.

Note that no changes are made to the Lucene planner itself, just when and how it is invoked.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
@@ -637,6 +637,13 @@ public class RecordQueryPlanner implements QueryPlanner {
                 if (p != null) {
                     p = computeIndexFilters(planContext, p);
                 }
+                if (p != null && p.getNumNonSargables() > 0) {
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -637,6 +637,13 @@ private ScoredPlan planIndex(@Nonnull PlanContext planContext, @Nonnull QueryCom
                 if (p != null) {
                     p = computeIndexFilters(planContext, p);
                 }
+                if (p != null && p.getNumNonSargables() > 0) {
```
tian-yizuo, line 654: By looking at the ```planOther```, is this just for the plan for TEXT search?

tian-yizuo (commented):
Looks good from perspective of Lucene

normen662 (approved):

PR #1817, opened 2022-08-09
Resolve #1432: Merge online indexer sub modules - "by records" and 'm…

### LINKED ISSUE #1432: Merge online indexer sub modules - "by records" and 'multi target by records"
After gaining some confidence in the "multi target by records" indexing path, change the code to let the "single target by records" be a private case of one item in the target index list.

jjezra:
…ulti target by records"
At this step, we still keep the old `IndexingByRecords` module - used for implementing `buildRange`, `splitIndexBuildRange`, `buildUnbuiltRange`,`buildEndpoints`. Maybe it'll be possible to delete it after implementing parallel indexing in the new modules.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/TupleRange.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/TupleRange.java
@@ -214,6 +214,18 @@ public class TupleRange {
     }
 
     /**
+     * Create a <code>TupleRange</code> over all keys between the given {@link Tuple}s, including both endpoints.
+     * @param low the inclusive start of the range
+     * @param high the inclusive end of the range
+     * @return a <code>TupleRange</code> between <code>start</code> and <code>end</code>, inclusive
+     */
+    public static TupleRange betweenInclusive(@Nullable Tuple low, @Nullable Tuple high) {
+        EndpointType lowEndpoint = (low == null) ? EndpointType.TREE_START : EndpointType.RANGE_INCLUSIVE;
+        EndpointType highEndpoint = (high == null) ? EndpointType.TREE_END : EndpointType.RANGE_INCLUSIVE;
+        return new TupleRange(low, high, lowEndpoint, highEndpoint);
+    }
+
+    /**
      * Create a <code>TupleRange</code> over all keys prefixed by some {@link String}. This
      * is a shortcut for creating a <code>TupleRange</code> where both the low- and high-endpoints
      * are <code>Tuple</code>s containing a single <code>String</code> where both the low
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -823,6 +823,12 @@ public abstract class IndexingBase {
                 getRunner().getExecutor());
     }
 
+    protected static boolean allRangesExhausted(Tuple cont, Tuple end) {
+        // if cont isn't null, it means that the cursor was not exhausted
+        // if end isn't null, it means that the range is a segment (i.e. closed or half-open interval) - the rangeSet may contain more unbuilt ranges
+        return end == null && cont == null;
+    }
+
     // rebuildIndexAsync - builds the whole index inline (without committing)
     @Nonnull
     public CompletableFuture<Void> rebuildIndexAsync(@Nonnull FDBRecordStore store) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
@@ -44,7 +44,6 @@ import com.google.protobuf.ByteString;
 import com.google.protobuf.Message;
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -130,24 +129,21 @@ public class IndexingByIndex extends IndexingBase {
                             .thenCompose(vignore -> {
                                 SubspaceProvider subspaceProvider = common.getRecordStoreBuilder().getSubspaceProvider();
                                 return subspaceProvider.getSubspaceAsync(context)
-                                        .thenCompose(subspace -> buildIndexFromIndex(subspaceProvider, subspace, null, null));
+                                        .thenCompose(subspace -> buildIndexFromIndex(subspaceProvider, subspace));
                             });
                 }), common.indexLogMessageKeyValues("IndexingByIndex::buildIndexInternalAsync"));
     }
 
     @Nonnull
-    private CompletableFuture<Void> buildIndexFromIndex(@Nonnull SubspaceProvider subspaceProvider, @Nonnull Subspace subspace, @Nullable byte[] start, @Nullable byte[] end) {
-        final List<Object> additionalLogMessageKeyValues = Arrays.asList(LogMessageKeys.CALLING_METHOD, "buildIndexFromIndex",
-                LogMessageKeys.RANGE_START, start,
-                LogMessageKeys.RANGE_END, end);
-
+    private CompletableFuture<Void> buildIndexFromIndex(@Nonnull SubspaceProvider subspaceProvider, @Nonnull Subspace subspace) {
+        final List<Object> additionalLogMessageKeyValues = Arrays.asList(LogMessageKeys.CALLING_METHOD, "buildIndexFromIndex");
         return iterateAllRanges(additionalLogMessageKeyValues,
-                (store, recordsScanned) -> buildRangeOnly(store, start, end , recordsScanned),
+                (store, recordsScanned) -> buildRangeOnly(store, recordsScanned),
                 subspaceProvider, subspace);
     }
 
     @Nonnull
-    private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store, byte[] startBytes, byte[] endBytes, @Nonnull AtomicLong recordsScanned) {
+    private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store, @Nonnull AtomicLong recordsScanned) {
         // return false when done
 
         validateSameMetadataOrThrow(store);
@@ -161,7 +157,7 @@ public class IndexingByIndex extends IndexingBase {
         validateOrThrowEx(store.isIndexScannable(srcIndex), "source index is not scannable");
 
         RangeSet rangeSet = new RangeSet(store.indexRangeSubspace(index));
-        AsyncIterator<Range> ranges = rangeSet.missingRanges(store.ensureContextActive(), startBytes, endBytes).iterator();
+        AsyncIterator<Range> ranges = rangeSet.missingRanges(store.ensureContextActive()).iterator();
 
         final ExecuteProperties.Builder executeProperties = ExecuteProperties.newBuilder()
                 .setIsolationLevel(IsolationLevel.SNAPSHOT)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
@@ -48,7 +48,6 @@ import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
@@ -192,7 +191,7 @@ public class IndexingByIndex extends IndexingBase {
                                           lastResult.get().get().getIndexEntry().getKey() :
                                           rangeEnd)
                     .thenCompose(cont -> rangeSet.insertRange(store.ensureContextActive(), packOrNull(rangeStart), packOrNull(cont), true)
-                                .thenApply(ignore -> !Objects.equals(cont, rangeEnd)));
+                                .thenApply(ignore -> !allRangesExhausted(cont, rangeEnd)));
 
         });
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
@@ -39,7 +39,6 @@ import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.Key;
-import com.apple.foundationdb.record.metadata.RecordType;
 import com.apple.foundationdb.subspace.Subspace;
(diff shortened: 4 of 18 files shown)


jjezra:
Test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -192,7 +191,7 @@ private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store,
                                           lastResult.get().get().getIndexEntry().getKey() :
                                           rangeEnd)
                     .thenCompose(cont -> rangeSet.insertRange(store.ensureContextActive(), packOrNull(rangeStart), packOrNull(cont), true)
-                                .thenApply(ignore -> !Objects.equals(cont, rangeEnd)));
+                                .thenApply(ignore -> !allRangesExhausted(cont, rangeEnd)));
```
alecgrieser, line 194: This feels like a somewhat weird change to make. It would seem to assume that `rangeEnd` is always `null`. If that's true, we should probably change the method to reject non-`null` values to `rangeEnd`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -165,8 +198,7 @@ private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store,
                                           lastResult.get().get().getPrimaryKey() :
                                           rangeEnd)
                     .thenCompose(cont -> insertRanges(store.ensureContextActive(), targetRangeSets, packOrNull(rangeStart), packOrNull(cont))
-                                .thenApply(ignore -> !Objects.equals(cont, rangeEnd)));
-
+                                .thenApply(ignore -> !allRangesExhausted(cont, rangeEnd)));
```
alecgrieser, line 201: This also feels like maybe it works in practice because the wrong kind of range is never put into the range set...but that feels a little delicate to me. It's possible I'm missing part of some bigger picture, but it seems like we'd want to retain the ability in general to have a non-contiguous range built so that we can split up the records subspace and build the index in parallel, and a change like this would make that more difficult

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -202,8 +238,7 @@ CompletableFuture<Void> rebuildIndexInternalAsync(FDBRecordStore store) {
 
     @Nonnull
     @SuppressWarnings("PMD.CloseResource")
-    private CompletableFuture<Tuple> rebuildRangeOnly(@Nonnull FDBRecordStore store, Tuple cont, @Nonnull AtomicLong recordsScanned) {
-
+    private CompletableFuture<Tuple> rebuildRangeOnly(@Nonnull FDBRecordStore store, Tuple cont, @Nonnull AtomicLong recordsScanned, @Nullable Tuple rangeEnd) {
```
alecgrieser, line 241: This API doesn't seem right to me. I _think_ I understand what's going on, which is that it is assuming that index builds either require:

1. Scanning all records, in which case `rangeEnd` is `null` or
2. Scanning a set of range of records to cover all indexed types, in which case `rangeEnd` is the type key of the lexicographically last type

The problem I have with it is that the `rangeEnd` tuple here is a little too divorced from its origin for that to be clear to the user, especially because in most of the rest of the API, a bare `Tuple` is not assumed to be an inclusive right endpoint.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -823,6 +823,12 @@ protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMes
                 getRunner().getExecutor());
     }
 
+    protected static boolean allRangesExhausted(Tuple cont, Tuple end) {
```
alecgrieser, line 826: In each case this function is used, it seems to be baking in an implicit assumption that `end` will always be `null`, as otherwise, the iteration of the loops won't ever stop, but I'm not sure that that's true in all cases

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
```diff
@@ -77,7 +76,8 @@ public class IndexingByRecords extends IndexingBase {
 
     IndexingByRecords(@Nonnull IndexingCommon common, @Nonnull OnlineIndexer.IndexingPolicy policy) {
         super(common, policy);
-        this.recordsRange = computeRecordsRange();
+        final TupleRange range = common.computeRecordsRange();
+        this.recordsRange = range == null ? TupleRange.ALL : range;
```
alecgrieser, line 80: Rather than retaining both `IndexingByRecords` and `IndexingMultiTargetByRecords` until the range based APIs can be implemented on both (or removed), would it instead make sense to try and isolate the logic that truly is different between single- and multi-index situations, and then delete one of the classes? From what I can tell, the differences are:

1. When computing the records range, all record types from all indexes need to be included so as to cover all needed records
2. When indexing, all index maintainers need to be updated
3. When updating index states, each index needs to be updated

There are maybe a few other things, but this PR seems to also touch code related to picking ranges to scan, which I think should actually be common between the indexing methods

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -102,20 +110,45 @@ CompletableFuture<Void> buildIndexInternalAsync() {
...
+            rangeEnd = range.end;
+        }
+
+        final CompletableFuture<FDBRecordStore> maybePresetRangeFuture =
+                rangeStart == null ?
+                CompletableFuture.completedFuture(null) :
+                buildCommitRetryAsync((store, recordsScanned) -> {
+                    // Here: only records inside the defined records-range are relevant to the index. Hence, the completing range
+                    // can be preemptively marked as indexed.
+                    final RangeSet rangeSet = new RangeSet(store.indexRangeSubspace(common.getIndex()));
```
alecgrieser, line 137: Is it correct that only `common.getIndex()`'s index range subspace gets updated? That doesn't seem right for non-idempotent indexes, but maybe those are excluded from multi-target index builds?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
jjezra: That's a bug. Will fix.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
jjezra:
```
            if (Boolean.FALSE.equals(hasNext)) {
            }
            final Range range = ranges.next();
            final Tuple rangeStart = RangeSet.isFirstKey(range.begin) ? null : Tuple.fromBytes(range.begin);
            final Tuple rangeEnd = RangeSet.isFinalKey(range.end) ? null : Tuple.fromBytes(range.end);
```
It is always null at the last unbuilt range (if the loop hadn't stop few lines above for no ranges). What am I missing?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
jjezra: I was hoping to avoid the use of predefined ranges that are provided by the users in parallel indexing. At the current indexing state, the iteration should stop only when the index is fully built.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
jjezra: Would a comment and renaming help?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
jjezra: Sorry, but I'm not sure that I understand this comment. Can you clarify?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByRecords.java
jjezra: (after re-reading) 
Yes, I do not wish to retain both modules but to (eventually) delete `IndexingByRecords`, or just leave it with minimal functionality for API backward compatibility - if we decide to keep it. 
The multi-target module also has a slightly different flow, which is simpler and more compatible with the other new modules. It's probably a good idea to sharing code when it's possible and easy - like `computeRecordsRange` here..

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
alecgrieser: I believe that the new condition effectively always assumes it is building the entire range (which is I suppose the "implicit assumption" that I thought it was "baking in"). That's wrong in the cases where we only want to build a range, though it sounds like we're trying to get rid of that in favor of an alternative approach anyway. But if that's the plan, I'd rather we just remove and deprecate those endpoints, possibly waiting until we have a better story around parallel index builds.

Now that I'm looking closer, I think the _original_ condition was also wrong, though maybe in a different way (I think). I'm mainly looking at `buildRangeOnly`, which is supposed to take a range, and ensure that that whole range is built. Before this PR, that code:

1. Reads the range set to compute unbuilt ranges in the given range then
2. Looks in the results and builds the range over multiple transactions until
3. It builds all the way until it reaches `rangeEnd`

However, `rangeEnd` there is a range from the range set, meaning it effectively only builds the first (contiguous) range of data from the target range. The desired behavior, I think, is that it build all the way until it gets to the end of the target range. So, for `buildRangeOnly`, I think a better condition is something like "until the tuple given matches a tuple constructed from `rangeEndBytes` (or is `null` if `rangeEndBytes` is null)".

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/TupleRange.java
```diff
@@ -213,6 +213,18 @@ public static TupleRange between(@Nullable Tuple low, @Nullable Tuple high) {
         return new TupleRange(low, high, lowEndpoint, highEndpoint);
     }
 
+    /**
+     * Create a <code>TupleRange</code> over all keys between the given {@link Tuple}s, including both endpoints.
+     * @param low the inclusive start of the range
+     * @param high the exclusive end of the range
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/TupleRange.java
```diff
@@ -213,6 +213,18 @@ public static TupleRange between(@Nullable Tuple low, @Nullable Tuple high) {
...
+     * @return a <code>TupleRange</code> between <code>start</code> and <code>end</code>, includes.
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: (assume:
'old path' = `IndexingByRecords` indexer - which was kept "as is" as much as possible. I'll be happy to change it as little as possible until reaching the stage of deleting unused code. 
'new path' = `IndexingMultiTargetByRecords`, 'IndexingByIndex`, `IndexingScrub*` indexers - a simplified path with no endpoint)

I believe that the current stop condition of the 'new path' was, indeed, a bug. My bug. It was harmless because these new indexers could not never have caused or 'inherit' a fragmented rangeSet. The name of the new stop condition - `allRangesExhausted` - should clarify that it return true if no more missing ranges exist, which is what these new indexers expect (I think). When the time comes, a different stop condition (`thisRangeIsExhausted`?) can be used to support a single range indexing. 
Although the 'old path' is hard to follow, I believe that its stop condition is fine because we do have some fragmented indexing tests. While developing, these tests had immediately failed when `IndexingMultiTargetByRecords` was defaulted to index single targets.
I'm not sure that I understood your last paragraph. IIUIC, `buildRangeOnly`'s return value is the stop condition of `iterateAllRanges`'s loop. For the current 'new path' indexers it returns true until `allRangesExhausted` - which means either continue indexing the same range or move on to the next unbuilt one. Am I missing something?

jjezra:
Rebased, resolved conflicts, and squashed...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
alecgrieser: My problem with that is that this is not the current API of the indexer, and so making these changes seems premature unless you _also_ remove the `startBytes` and `endBytes` from the API (or require that `startBytes` and `endBytes` are both `null` at runtime). Which could certainly be done, and I'm fine with merging these changes after removing the endpoints we don't want to support any more.

I think if we want to understand why the fragmented indexing tests failed with the initial version of this PR but succeeded with the change to `allRangesExhausted`, I think what's going on is that when there is a range in the range set/an end range passed to this method, `allRangesExhausted` will always return `false`, so it will keep looping, until such time as the range set has no more ranges within it, at which point the earlier check (commented "no more missing ranges") fires. This wasn't a problem with the single index indexer because it uses the `hasMore` in a different way from the multi-target case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
alecgrieser, line 194: This condition is one of the places where using `allRangesExhausted` looks wrong to me (both now and before this PR). Note that `rangeEnd` was computed from reading the `rangeSet` above, so this lambda will now never return `false` if there's every any range in the range set. Before this PR, it was also wrong, but it would return `false` as soon as it completes the first range. But I think it's working, in practice, because `IndexingByIndex` builds would always be structured in a way that they'd have a `null` value as the end of at least one range. The way it reads the range set and only appears to grab the first element is....also kind of suspicious, though that's not affected by this change as directly

alecgrieser (changes requested):

alecgrieser (approved):

PR #1822, opened 2022-08-12
Ignore CollectionTerminatedException.

hatyo:
This PR made the LuceneOptimizedIndexSearcher, however it mistakenly allowed `CollectionTerminatedException` to be propagated to the outside.

In the original code, that exception was [properly ignored](https://github.com/FoundationDB/fdb-record-layer/commit/e98cf6aa234563acc7c45e57c3a6f19e22af0a1f#diff-f59fd6db59846daffb972ed37a5e1e7352965c1905e97b1b75c418bef7c61275L112) as suggested in Lucene Docs:

> Note: IndexSearcher swallows this exception and never re-throws it. As a consequence, you should not catch it when calling [IndexSearcher.search(org.apache.lucene.search.Query, int)](https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/search/IndexSearcher.html#search-org.apache.lucene.search.Query-int-) as it is unnecessary and might hide misuse of this exception.

This PR attempts to restore that behaviour, in addition, it cascades `IOException` instead of ignoring it within asynchronous execution blocks.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
@@ -29,6 +29,7 @@ import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.BulkScorer;
+import org.apache.lucene.search.CollectionTerminatedException;
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.IndexSearcher;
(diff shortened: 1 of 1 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
```diff
@@ -180,8 +205,10 @@ private <C extends Collector> CompletableFuture<C> searchOptimized(@Nonnull fina
                         if (scorer != null && scorer.getLeft() != null && scorer.getRight() != null) {
                             try {
                                 scorer.getRight().score(scorer.getLeft(), ctx.reader().getLiveDocs());
-                            } catch (IOException e) {
+                            } catch (CollectionTerminatedException cte) {
```
MMcM, line 208: Based on `IndexSearcher.search`, this can be thrown either in `getLeafCollector` or `score`. So, I think the similar `catch` 14 lines up needs the same treatment.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/LuceneOptimizedIndexSearcher.java
hatyo: Thanks, I added the same `catch` logic there, and provided some documentation.

MMcM (approved):

PR #1823, opened 2022-08-18
Add a new kind of FDBDirectory that uses a separate FDBRecordContext …

MMcM:
… and is allowed to commit it along the way.

Because of the way merging happens in multiple threads, it is possible for there to be outstanding reads at the same time as a periodic commit or requested fsync. These all need to be completed before closing the transaction so that they don't fail. A read-write lock prevents more being started before this is all done.

Use this to implement force merge as an index operation.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -131,7 +131,7 @@ public class FDBDirectory extends Directory {
      */
     private final AtomicReference<Map<String, FDBLuceneFileReference>> fileReferenceCache;
 
-    private final AtomicLong fileSequenceCounter;
+    protected final AtomicLong fileSequenceCounter;
     private final Cache<Pair<Long, Integer>, CompletableFuture<byte[]>> blockCache;
 
     private final boolean compressionEnabled;
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -24,6 +24,7 @@ import com.apple.foundationdb.KeyValue;
 import com.apple.foundationdb.MutationType;
 import com.apple.foundationdb.ReadTransaction;
 import com.apple.foundationdb.StreamingMode;
+import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncIterable;
 import com.apple.foundationdb.async.AsyncUtil;
@@ -70,8 +71,11 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import static com.apple.foundationdb.record.lucene.codec.LuceneOptimizedCompoundFormat.DATA_EXTENSION;
@@ -103,7 +107,7 @@ public class FDBDirectory extends Directory {
     private static final int META_SUBSPACE = 1;
     private static final int DATA_SUBSPACE = 2;
     private final AtomicLong nextTempFileCounter = new AtomicLong();
-    private final FDBRecordContext context;
+    protected FDBRecordContext context;
     private final Subspace subspace;
     private final Subspace metaSubspace;
     private final Subspace dataSubspace;
@@ -125,7 +129,7 @@ public class FDBDirectory extends Directory {
      * the {@link #listAll()} method requires that file names be returned in sorted order.
      * @see FDBLuceneFileReference
      */
-    private final AtomicReference<ConcurrentSkipListMap<String, FDBLuceneFileReference>> fileReferenceCache;
+    private final AtomicReference<Map<String, FDBLuceneFileReference>> fileReferenceCache;
 
     private final AtomicLong fileSequenceCounter;
     private final Cache<Pair<Long, Integer>, CompletableFuture<byte[]>> blockCache;
@@ -190,7 +194,7 @@ public class FDBDirectory extends Directory {
         this.sharedCachePending = sharedCacheManager != null && sharedCacheKey != null;
     }
 
-    private long deserializeFileSequenceCounter(@Nullable byte[] value) {
+    private static long deserializeFileSequenceCounter(@Nullable byte[] value) {
         return value == null ? 0L : Tuple.fromBytes(value).getLong(0);
     }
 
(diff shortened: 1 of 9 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -206,10 +210,10 @@ private CompletableFuture<Void> loadFileSequenceCounter() {
             // Already loaded
             return AsyncUtil.DONE;
         }
-        return context.ensureActive().get(sequenceSubspaceKey).thenAccept(serializedValue -> {
+        CompletableFuture<Long> loadValue = readTransaction(tr -> tr.get(sequenceSubspaceKey).thenApply(FDBDirectory::deserializeFileSequenceCounter));
+        return loadValue.thenAccept(loadedValue -> {
             // Replace the counter value in the counter unless something else has already
             // updated it, in which case we want to keep the cached value
-            long loadedValue = deserializeFileSequenceCounter(serializedValue);
             fileSequenceCounter.compareAndSet(originalValue, loadedValue);
```
alecgrieser, line 217: One of the premises of the way that the `FDBDirectory` was written before was that because of how everything was, the only thing that could update the file sequence counter was either (1) in another transaction or (2) using the same `FDBDirectory` object. And because the file counter (and `fileReferenceCache`) was re-read with every transaction, it would never be stale in a problematic way. However, it seems like that could happen with the multi-transactional directory, if (for example), there's a multi-transactional directory open, and the file sequence counter is updated between transactions. Do we need to rethink that somehow?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWithSeparateContext.java
```diff
@@ -0,0 +1,185 @@
...
+    @Override
+    protected void countWriteBytes(long count) {
+        pendingBytesCount += count;
+    }
+
+    @Override
+    @SuppressWarnings("PMD.EmptyCatchBlock")
+    protected void syncTransaction() {
+        Lock lock = readWriteLock.writeLock();
+        lock.lock();
```
alecgrieser, line 138: It took me a while to understand how the locking here works, but I think I get it, if I may:

1. Reads can return a future. When a read is scheduled, it gets a handle to the lock, adds the future to a "pending reads" lists, and then releases the lock
2. Writes do not return a future. A write grabs the lock, applies the mutation, and then releases the lock.
3. When `syncTransaction` is called, the lock is acquired and all outstanding reads are allowed to finish. This ensures that transaction is now unused and safe to commit. The transaction is committed, then the object's current transaction is replaced by a new one to use in future reads and writes.

I think these semantics make sense, but I wanted to make sure I had that right.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1984,6 +1984,34 @@ void luceneCursorTestAllMatchesSkipped() throws Exception {
         }
     }
 
+    @Test
+    void forceMerge() {
```
alecgrieser, line 1988: I'm not sure how we'd right dependable unit tests here, but it seems like having tests that show that regular reads, writes, and updates can all happen while the first thread is doing a force merge.

Maybe there needs to be a way to that it to only begin to bulk merge, or something, to write tests.

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWithSeparateContext.java
```diff
@@ -0,0 +1,185 @@
...
+                    .setInfoStream(new LuceneLoggerInfoStream(LOGGER));
+            indexWriter = new IndexWriter(this, indexWriterConfig);
+        }
+        return indexWriter;
+    }
+
+    @Nonnull
+    private static FDBRecordContext openLikeContext(@Nonnull FDBRecordContext context) {
+        return context.getDatabase().openContext(context.getMdcContext(), context.getTimer(),
+                context.getWeakReadSemantics(), context.getPriority(), context.getTransactionId());
```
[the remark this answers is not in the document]
MMcM: #1830

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -103,7 +107,7 @@ public class FDBDirectory extends Directory {
     private static final int META_SUBSPACE = 1;
     private static final int DATA_SUBSPACE = 2;
     private final AtomicLong nextTempFileCounter = new AtomicLong();
-    private final FDBRecordContext context;
+    protected FDBRecordContext context;
```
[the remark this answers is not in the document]
MMcM: Note that `close` clears this field now, meaning that it is possible to detect use of a closed directory, which seems useful in its own right.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: The all or nothing isolation of changes that include counter updates is lost, of course, now that there are multiple transactions. But within each transaction, I believe the invariants still hold: the counter is initialized as of the start of the transaction and therefore sees updates from earlier in the same directory as well as other transactions in the meantime. So that new references won't be sequential, but I don't know that they need to be.

There is the problem of transaction conflicts after the first commit, which may fail the entire merge process. This will leave some partial writes. But that is the same as if a merging process crashed after some files had be written to the file system. There is orphan junk, but I don't think anything worse than that. Do you?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWithSeparateContext.java
```diff
@@ -0,0 +1,185 @@
...
+    public void close() {
+        if (indexWriter != null) {
+            try {
+                indexWriter.close();
+            } catch (IOException ex) {
+                throw new RecordCoreException("error closing writer", ex);
+            }
+        }
+        if (context != null) {
+            context.commit();
```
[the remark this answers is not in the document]
MMcM: In a file system directory, writes will have already been performed to files that will, at some unspecified time in the future, appear on disk. There are no pending buffers. But neither does `close` do `fsync`; there is another method for that.

Still, this seems the more reasonable choice, as opposed to saying that the last set of writes is lost on `close` unless `sync` is done beforehand.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWithSeparateContext.java
MMcM: Yes. That is all correct. Additionally, it is a read-write lock, so there can be multiple reads in parallel (not just the futures, but also the thing creating the future). But only one write. And that write can decide on its own to sync, based on size / time, independent of explicit requests for it.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
MMcM: Perhaps it can hook `openLikeContext` to do another transaction in between those of the directory.

MMcM: `IndexWriter` expects that a lock will prevent another writer, normally in the file system, on the same directory.

MMcM: * https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L948 acquire lock.
* https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L1151-L1163 delete any "orphan" files.

Since our locks are noops, it can delete a file that is part of a parallel merge.

alecgrieser (approved):
I believe there are some outstanding questions about concurrency here that probably need answering, but I'm not really the right person to review that as I'm not really working on the Lucene sub-project anymore and other people are. So I'm going to "approve" as a way of dismissing my old "request changes" review, but someone else will probably be needed to take a real look at that part of the code later

MMcM:
Superseded by #2344 and follow ons.

PR #1825, opened 2022-08-18
Resolves #1824: Use different analyzers for different fields in one L…

### LINKED ISSUE #1824: Use different analyzers for different fields in one Lucene index, and enable exclusion of fields for auto-complete

tian-yizuo:
…ucene index, and enable exclusion of fields for auto-complete

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
@@ -1,0 +1,96 @@
+/*
+ * LuceneAnalyzerCombinationProvider.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import org.apache.lucene.analysis.miscellaneous.PerFieldAnalyzerWrapper;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+/**
+ * Provide a combination of analyzers for multiple fields of one Lucene index.
+ * The combinations of analyzers for index time and query time are potentially different.
+ * The default analyzer chooser is used for all fields of one Lucene index except the fields which has overrides in the analyzer chooser per field mapping.
+ */
+public class LuceneAnalyzerCombinationProvider {
+    private AnalyzerChooser defaultIndexAnalyzerChooser;
+    private AnalyzerChooser defaultQueryAnalyzerChooser;
+    private Map<String, AnalyzerChooser> indexAnalyzerChooserPerFieldOverride;
+    private Map<String, AnalyzerChooser> queryAnalyzerChooserPerFieldOverride;
+
+    public LuceneAnalyzerCombinationProvider(@Nonnull AnalyzerChooser defaultIndexAnalyzerChooser, @Nonnull AnalyzerChooser defaultQueryAnalyzerChooser,
+                                             @Nullable Map<String, AnalyzerChooser> indexAnalyzerChooserPerFieldOverride, @Nullable Map<String, AnalyzerChooser> queryAnalyzerChooserPerFieldOverride) {
+        this.defaultIndexAnalyzerChooser = defaultIndexAnalyzerChooser;
+        this.defaultQueryAnalyzerChooser = defaultQueryAnalyzerChooser;
+        this.indexAnalyzerChooserPerFieldOverride = indexAnalyzerChooserPerFieldOverride;
+        this.queryAnalyzerChooserPerFieldOverride = queryAnalyzerChooserPerFieldOverride;
+    }
+
+    public LuceneAnalyzerWrapper provideIndexAnalyzer(@Nonnull String text) {
+        return provideIndexAnalyzer(Collections.singletonList(text));
+    }
+
+    public LuceneAnalyzerWrapper provideIndexAnalyzer(@Nonnull List<String> texts) {
+        return buildAnalyzerWrapper(texts, defaultIndexAnalyzerChooser, indexAnalyzerChooserPerFieldOverride);
+    }
+
+    public LuceneAnalyzerWrapper provideQueryAnalyzer(@Nonnull String text) {
+        return provideQueryAnalyzer(Collections.singletonList(text));
+    }
+
+    public LuceneAnalyzerWrapper provideQueryAnalyzer(@Nonnull List<String> texts) {
+        return buildAnalyzerWrapper(texts, defaultQueryAnalyzerChooser, queryAnalyzerChooserPerFieldOverride);
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private static LuceneAnalyzerWrapper buildAnalyzerWrapper(@Nonnull List<String> texts, @Nonnull AnalyzerChooser defaultAnalyzerChooser,
+                                                              @Nullable Map<String, AnalyzerChooser> customizedAnalyzerChooserPerField) {
+        final LuceneAnalyzerWrapper defaultAnalyzerWrapper = defaultAnalyzerChooser.chooseAnalyzer(texts);
+        if (customizedAnalyzerChooserPerField != null) {
+            // The order of keys matters because the identifier for each map needs to be consistent
+            SortedMap<String, LuceneAnalyzerWrapper> analyzerWrapperMap = new TreeMap<>(customizedAnalyzerChooserPerField.entrySet().stream()
+                    .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().chooseAnalyzer(texts))));
+
+            PerFieldAnalyzerWrapper analyzerWrapper = new PerFieldAnalyzerWrapper(defaultAnalyzerWrapper.getAnalyzer(),
+                    analyzerWrapperMap.entrySet().stream().collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().getAnalyzer())));
+            final String analyzerId = buildAnalyzerIdentifier(defaultAnalyzerWrapper, analyzerWrapperMap);
+            return new LuceneAnalyzerWrapper(analyzerId, analyzerWrapper);
+        } else {
+            return defaultAnalyzerWrapper;
+        }
+    }
+
+    private static String buildAnalyzerIdentifier(@Nonnull LuceneAnalyzerWrapper defaultAnalyzerWrapper, @Nonnull Map<String, LuceneAnalyzerWrapper> analyzerWrapperMap) {
+        final StringBuilder builder = new StringBuilder();
+        builder.append(defaultAnalyzerWrapper.getUniqueIdentifier());
+        for (String id : analyzerWrapperMap.keySet()) {
+            builder.append("/");
+            builder.append(id);
+        }
+        return builder.toString();
+    }
+}
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
@@ -37,6 +37,9 @@ import java.util.stream.Collectors;
  * The default analyzer chooser is used for all fields of one Lucene index except the fields which has overrides in the analyzer chooser per field mapping.
  */
 public class LuceneAnalyzerCombinationProvider {
+    public static final String DELINEATOR_BETWEEN_KEY_AND_VALUE = ":";
+
+    public static final String DELINEATOR_BETWEEN_KEY_VALUE_PAIRS = ",";
     private AnalyzerChooser defaultIndexAnalyzerChooser;
     private AnalyzerChooser defaultQueryAnalyzerChooser;
     private Map<String, AnalyzerChooser> indexAnalyzerChooserPerFieldOverride;
@@ -88,7 +91,7 @@ public class LuceneAnalyzerCombinationProvider {
         final StringBuilder builder = new StringBuilder();
         builder.append(defaultAnalyzerWrapper.getUniqueIdentifier());
         for (String id : analyzerWrapperMap.keySet()) {
-            builder.append("/");
+            builder.append(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS);
             builder.append(id);
         }
         return builder.toString();
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
@@ -100,16 +100,12 @@ public class LuceneAnalyzerRegistryImpl implements LuceneAnalyzerRegistry {
 
         Map<String, AnalyzerChooser> indexAnalyzerChooserPerFieldOverride = new TreeMap<>();
         Map<String, AnalyzerChooser> queryAnalyzerChooserPerFieldOverride = new TreeMap<>();
-        final String[] elements = analyzerPerFieldName.strip().split(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS);
 
-        for (String element : elements) {
-            final String[] keyValuePair = element.strip().split(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE);
-            final String fieldName = keyValuePair[0];
-            final String analyzerName = keyValuePair[1];
+        LuceneIndexOptions.parseKeyValuePairOptionValue(analyzerPerFieldName).forEach((fieldName, analyzerName) -> {
             Pair<AnalyzerChooser, AnalyzerChooser> perFieldAnalyzerChooserPair = getAnalyzerChooser(index, analyzerName, type);
             indexAnalyzerChooserPerFieldOverride.put(fieldName, perFieldAnalyzerChooserPair.getLeft());
             queryAnalyzerChooserPerFieldOverride.put(fieldName, perFieldAnalyzerChooserPair.getRight());
-        }
+        });
 
         return new LuceneAnalyzerCombinationProvider(defaultAnalyzerChooserPair.getLeft(), defaultAnalyzerChooserPair.getRight(),
                 indexAnalyzerChooserPerFieldOverride, queryAnalyzerChooserPerFieldOverride);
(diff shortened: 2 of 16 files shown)


MMcM (approved):
Although I think the scheme discussed in the meeting of having per-field option _names_ with a common prefix (followed by the field name) would be more straightforward, it looks like this ought to work okay.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
```diff
@@ -60,16 +60,32 @@ public class LuceneIndexOptions {
...
+     * The format of the value should be "fieldName1/analyzerName1/fieldName2/analyzerName2...".
+     */
+    public static final String LUCENE_ANALYZER_NAME_PER_FIELD_OPTION = "luceneAnalyzerNamePerField";
+    /**
+     * The name of the Lucene analyzer to use for auto-complete query.
      */
     public static final String AUTO_COMPLETE_ANALYZER_NAME_OPTION = "autoCompleteAnalyzerName";
+    /**
+     * The override mapping from fields to Lucene analyzers for auto-complete query, if they want to use different analyzers than the default one.
+     * The format of the value should be "fieldName1/analyzerName1/fieldName2/analyzerName2...", with keys and values splitted by "/".
```
alecgrieser, line 75:
```suggestion
     * The format of the value should be "fieldName1/analyzerName1/fieldName2/analyzerName2...", with keys and values split by "/".
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
```diff
@@ -60,16 +60,32 @@ public class LuceneIndexOptions {
...
+     */
+    public static final String AUTO_COMPLETE_ANALYZER_NAME_PER_FIELD_OPTION = "autoCompleteAnalyzerNamePerField";
     /**
      * The name of the synonym set to use in Lucene.
      */
     public static final String TEXT_SYNONYM_SET_NAME_OPTION = "textSynonymSetName";
 
+    /**
+     * A list of fields to be excluded for auto-complete query.
+     * The format of the value should be "fieldName1/fieldName2/...", with field names splitted by "/".
```
alecgrieser, line 85:
```suggestion
     * The format of the value should be "fieldName1/fieldName2/...", with field names split by "/".
```

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+/**
+ * Provide a combination of analyzers for multiple fields of one Lucene index.
+ * The combinations of analyzers for index time and query time are potentially different.
+ * The default analyzer chooser is used for all fields of one Lucene index except the fields which has overrides in the analyzer chooser per field mapping.
+ */
+public class LuceneAnalyzerCombinationProvider {
```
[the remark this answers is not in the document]
tian-yizuo: The ```AnalyzerChooser``` is designed for the provision of an Lucene analyzer given a text, so it is per Lucene analyzer. The only reason we need it for now, is for the capability needed to use different one for different languages. Whereas the ```LuceneAnalyzerCombinationProvider``` is more like a single object to provide a pair of analyzers for index time and query time based on the options config, because we want to make is simple for clients to configure for the analyzers to use, without worrying about which to use on index time and which to use on query time.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -86,17 +88,41 @@ private LuceneAnalyzerRegistryImpl() {
...
+        Pair<AnalyzerChooser, AnalyzerChooser> defaultAnalyzerChooserPair = getAnalyzerChooser(index, defaultAnalyzerName, type);
+
+        if (analyzerPerFieldName == null) {
+            return new LuceneAnalyzerCombinationProvider(defaultAnalyzerChooserPair.getLeft(), defaultAnalyzerChooserPair.getRight(),
+                    null, null);
+        }
+
+        Map<String, AnalyzerChooser> indexAnalyzerChooserPerFieldOverride = new TreeMap<>();
+        Map<String, AnalyzerChooser> queryAnalyzerChooserPerFieldOverride = new TreeMap<>();
+        String[] elements = analyzerPerFieldName.split("/");
```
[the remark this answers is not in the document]
tian-yizuo: @alecgrieser yeah it sounds better to use different delineators for that between key-value pairs and that between keys and values. So can I assume no field names will have ```:``` or ```,```?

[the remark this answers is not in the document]
alecgrieser: I don't know how valid an assumption that is; I think `:` and `,` are better delineator characters for maps, but I don't know whether they're currently allowed in field names (or analyzer names) or not. That's one of the reasons that for, say, the "replaced by" index name options, rather than coming up with a scheme by which the index names could be put together in a single option, we went with instead a single option _prefix_, and then collected all IndexOptions that began with that prefix to produce the list.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerCombinationProvider.java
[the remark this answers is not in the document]
alecgrieser: I think I like the abstraction, for the most part, and that it makes sense to encapsulate index-and-query-time-analyzers into an object other than a pair. However, once you have that abstraction, I'm not sure it makes sense to combine that with the behavior that constructs the per-field Lucene analyzers.

I also think the abstraction is fine for now, though, and something we could iterate on later if we really want to

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -44,5 +46,49 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+                    || analyzerNamePerFieldOption.startsWith(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS)
+                    || analyzerNamePerFieldOption.endsWith(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE)
+                    || analyzerNamePerFieldOption.endsWith(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS)) {
+                throw new MetaDataException("Index " + index.getName() + " has invalid " + optionKey + ": " + analyzerNamePerFieldOption);
+            }
+            String[] keyValuePairs = analyzerNamePerFieldOption.split(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS);
+            for (String keyValue : keyValuePairs) {
+                if (keyValue.indexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE) < 1
+                        || keyValue.indexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE) > keyValue.length() - 2
+                        || keyValue.indexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE) != keyValue.lastIndexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE)) {
```
alecgrieser, line 74: This isn't exactly performance critical code, but this recomputes `keyValue.indexOf(':')` multiple times. Ideally, this would compute the value once (for each `keyValue`) and store it in a local variable.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -44,5 +46,49 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+                    throw new MetaDataException("Index " + index.getName() + " has invalid " + optionKey + ": " + analyzerNamePerFieldOption);
```
alecgrieser, line 75:
```suggestion
                    throw new MetaDataException("Index " + index.getName() + " has invalid option value " + optionKey + ": " + analyzerNamePerFieldOption);
```

Likewise for the other `MetaDataExceptions` that are thrown in this class

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -44,5 +46,49 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+    private static void validateAnalyzerNamePerFieldOption(@Nonnull String optionKey, @Nonnull Index index) {
+        String analyzerNamePerFieldOption = index.getOption(optionKey);
+        if (analyzerNamePerFieldOption != null) {
+            analyzerNamePerFieldOption = analyzerNamePerFieldOption.strip();
+            if (analyzerNamePerFieldOption.startsWith(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE)
+                    || analyzerNamePerFieldOption.startsWith(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS)
+                    || analyzerNamePerFieldOption.endsWith(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE)
+                    || analyzerNamePerFieldOption.endsWith(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS)) {
+                throw new MetaDataException("Index " + index.getName() + " has invalid " + optionKey + ": " + analyzerNamePerFieldOption);
+            }
```
alecgrieser, line 69: I think these checks are all extraneous, as they'd all get validated by the later check of each key-value pair.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -44,5 +46,49 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+    public static void validateIndexOptions(@Nonnull Index index) {
+        validateAnalyzerNamePerFieldOption(LuceneIndexOptions.LUCENE_ANALYZER_NAME_PER_FIELD_OPTION, index);
+        validateAnalyzerNamePerFieldOption(LuceneIndexOptions.AUTO_COMPLETE_ANALYZER_NAME_PER_FIELD_OPTION, index);
+        validateAutoCompleteExcludedFields(index);
+    }
+
+    private static void validateAnalyzerNamePerFieldOption(@Nonnull String optionKey, @Nonnull Index index) {
+        String analyzerNamePerFieldOption = index.getOption(optionKey);
+        if (analyzerNamePerFieldOption != null) {
+            analyzerNamePerFieldOption = analyzerNamePerFieldOption.strip();
```
alecgrieser, line 63: I don't think the way that `strip()` is handled here is quite right, because `strip` is not called when the index options are actually evaluated later

alecgrieser (changes requested):
In my original review, I also proposed stripping out whitespace from the field and analyzer name options, but that wasn't done here, except somewhat during validation. I think it would be reasonable to either go with my initial suggestion (and strip out whitespace for each field and analyzer name) or to have tighter validation that required that there not be whitespace between elements.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
tian-yizuo: Changed it to also call the ```strip()``` when actually evaluating the index options.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
tian-yizuo: Good catch, removed these

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
```diff
@@ -1,2 +1,88 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class LuceneIndexValidatorTest {
```
alecgrieser, line 36: I'm not sure this belongs in the `LuceneIndexValidatorTest`, but we probably want positive test cases somewhere as well. I...suppose this doesn't need to block this PR, though

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -100,11 +100,13 @@ public LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@N
...
+        final String[] elements = analyzerPerFieldName.strip().split(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS);
 
-        for (int i = 0; i < elements.length / 2; i++) {
-            String fieldName = elements[2 * i];
-            Pair<AnalyzerChooser, AnalyzerChooser> perFieldAnalyzerChooserPair = getAnalyzerChooser(index, elements[2 * i + 1], type);
+        for (String element : elements) {
+            final String[] keyValuePair = element.strip().split(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE);
+            final String fieldName = keyValuePair[0];
+            final String analyzerName = keyValuePair[1];
+            Pair<AnalyzerChooser, AnalyzerChooser> perFieldAnalyzerChooserPair = getAnalyzerChooser(index, analyzerName, type);
```
alecgrieser, line 109: As written, this parser algorithm would turn the following string:

```
   a:b,  c: d,e :f
```

Into:

```
{"a": "b", "c": " d", "e ", "f"}
```

That is, it strips out whitespace at the beginning of each key and the end of each value, but not at the end of each key and the beginning of each value. Is that intentional?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
```diff
@@ -1,2 +1,88 @@
...
+    @Test
+    void testInvalidIndexOptions() {
```
alecgrieser, line 38: It might also make sense to add test cases where there is whitespace, and then validate that the whitespace is stripped out as expected

alecgrieser (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
tian-yizuo: The 1st assertion is just a positive test case.

alecgrieser: Right, but I meant more, like, tests that validate that, for example, whitespace is handled correctly and that the strings are parsed into the expected sets or maps, etc.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
tian-yizuo: Fixed it

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -44,5 +46,54 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+                throw new MetaDataException("Index " + index.getName() + " has invalid option value for " + optionKey + ": " + analyzerNamePerFieldOption);
+            }
+            final String[] keyValuePairs = analyzerNamePerFieldOption.split(LuceneIndexOptions.DELIMITER_BETWEEN_ELEMENTS);
+            for (int i = 0 ; i < keyValuePairs.length; i++) {
+                final String keyValue = keyValuePairs[i].strip();
+                int firstIndex = keyValue.indexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE);
+                if (keyValue.isEmpty()
+                        || firstIndex < 1 || firstIndex > keyValue.length() - 2
+                        || firstIndex != keyValue.lastIndexOf(LuceneIndexOptions.DELIMITER_BETWEEN_KEY_AND_VALUE)) {
+                    throw new MetaDataException("Index " + index.getName() + " has invalid option value for " + optionKey + ": " + analyzerNamePerFieldOption);
```
alecgrieser, line 75: Should this validate that the (stripped) key and value are non-empty?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
```diff
@@ -0,0 +1,88 @@
...
+        // valid options
+        final Map<String, String> options = ImmutableMap.of(LuceneIndexOptions.LUCENE_ANALYZER_NAME_PER_FIELD_OPTION, " text1 :NGRAM,text2:SYNONYM",
+                LuceneIndexOptions.AUTO_COMPLETE_EXCLUDED_FIELDS, "text2,text3");
+        validateIndexOptions(options);
```
alecgrieser, line 42: This should probably should be in an `assertDoesNotThrow` assertion to underscore why this is calling `validateIndexOptions`

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
tian-yizuo: I think empty key or value should cause exception because the check ```keyValue.isEmpty() || firstIndex < 1 || firstIndex > keyValue.length() - 2```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
tian-yizuo: done

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexValidatorTest.java
tian-yizuo: Done

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
alecgrieser: Oh, I see. It's a little complicated, but I think you're right: when the `keyValue` is stripped, if there's a completely blank key or value, then either `firstIndex` will be < 1 (for a blank key) or `firstIndex` will be > length - 2 (for a blank value), assuming there's only one delimiter. But I had to convince myself that was enough

alecgrieser (approved):
Approved, pending a successful PRB (which it seems like is a style thing)

alecgrieser (approved):

MMcM (approved):

PR #1828, opened 2022-08-30
Resolves #1827: Lower log severity for remote fetch when index type is unsupported

### LINKED ISSUE #1827: Remote fetch issues a warning on unsupported index maintainer
This is not really a warning (See `scanRemoteFetch operation is not supported by this index maintainer for Index`), since it is a known case where the index maintainer does not yet support the operation. 
This warning needs to be turned down so as not to trigger errors for anyone parsing the logs.

ohadzeliger:
When the index type is not supported for remote fetch (e.g. RANK, BITMAP, TEXT), the log message should not be WARN so as not to pollute the logs - this is a case where future development will increasingly support more index types and is not considered an error. 
In the case of `REMOTE_FETCH_WITH_FALLBACK`, the fallback mechanism would perform the scan using the legacy scan and fetch. 
In the case of `USE_REMOTE_FETCH`, the exception thrown (`UnsupportedRemoteFetchIndexException`) should indicate that this is a "not yet implemented" issue.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/UnsupportedRemoteFetchIndexException.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/UnsupportedRemoteFetchIndexException.java
@@ -1,0 +1,46 @@
+/*
+ * UnsupportedFormatVersionException.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2018 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/UnsupportedRemoteFetchIndexException.java
```diff
@@ -0,0 +1,46 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Exception thrown when remote fetch is attempted on an index that does not support remote fetch.
+ */
+@SuppressWarnings({"serial", "java:S110"})
+@API(API.Status.EXPERIMENTAL)
+public class UnsupportedRemoteFetchIndexException extends RecordCoreException {
```
alecgrieser, line 34: It's maybe worth noting somewhere that index maintainers outside of this project should also throw this exception when they don't support remote fetch in the javadoc

alecgrieser (changes requested):
It looks like this needs a release note, and then I had one suggestion for another Javadoc comment that might be good to add.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/UnsupportedRemoteFetchIndexException.java
ohadzeliger: Right. Done.

ohadzeliger:
Arg. Sorry about the release notes. Done.

alecgrieser (approved):

PR #1831, opened 2022-08-30
Resolves #1830: Make FDBRecordContext remember its config

### LINKED ISSUE #1830: Make FDBRecordContext remember its config
Rather than copying much of it into instance fields and needing to know what to copy for `newRunner`, etc.

MMcM:
See https://github.com/FoundationDB/fdb-record-layer/pull/1823#discussion_r956450035

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
@@ -145,10 +145,8 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
     private ConcurrentNavigableMap<byte[], Integer> localVersionCache;
     @Nonnull
     private ConcurrentNavigableMap<byte[], Pair<MutationType, byte[]>> versionMutationCache;
-    @Nullable
-    private final FDBDatabase.WeakReadSemantics weakReadSemantics;
     @Nonnull
-    private final FDBTransactionPriority priority;
+    private final FDBRecordContextConfig config;
     private final long timeoutMillis;
     @Nullable
     private Consumer<FDBStoreTimer.Wait> hookForAsyncToSync = null;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -205,7 +201,7 @@ protected FDBRecordContext(@Nonnull FDBDatabase fdb,
                 tr.options().setPrioritySystemImmediate();
                 break;
             default:
-                throw new RecordCoreArgumentException("unknown priority level " + priority);
+                throw new RecordCoreArgumentException("unknown priority level " + config.getPriority());
         }
 
         // Set the transaction timeout based on the config (if set) and the database factory otherwise
```
MMcM, line 207: The `timeoutMillis` field remains _only_ for the `STABLE` `getTimeoutMillis` API in the case where the config has default and the database factory value changes _after_ the context is created.

If we were willing to have `getTimeoutMillis` do what `getTimeoutMillisToSet` does now and accept that edge case, it could be removed as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -1171,36 +1176,20 @@ public void timeReadSampleKey(byte[] key) {
...
      * Get a new {@link FDBDatabaseRunner} that will run contexts similar to this one.
      * <ul>
      * <li>Same {@linkplain FDBDatabase database}</li>
-     * <li>Same {@linkplain FDBStoreTimer timer}</li>
-     * <li>Same {@linkplain #getMdcContext() MDC context}</li>
-     * <li>Same {@linkplain FDBDatabase.WeakReadSemantics weak read semantics}</li>
-     * <li>Same {@linkplain FDBTransactionPriority priority}</li>
-     * <li>Same {@linkplain #getTimeoutMillis() transaction timeout}</li>
-     * <li>Same {@linkplain #getPropertyStorage()}  property storage}</li>
+     * <li>Same {@linkplain FDBRecordContextConfig cnofig}</li>
```
alecgrieser, line 1186:
```suggestion
     * <li>Same {@linkplain FDBRecordContextConfig config}</li>
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
alecgrieser: I think leaving it like this (at least for now) seems fine to me. It also seems clear (from the comments and the code) why it's separated out from the config, so it seems maintainable

alecgrieser (changes requested):

alecgrieser (approved):

PR #1835, opened 2022-09-02
fix calls to compileTimeEval for relops

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -158,19 +157,30 @@ public Optional<QueryPredicate> toQueryPredicate(@Nonnull final CorrelationIdent
...
+
+            if (leftChildCorrelatedTo.isEmpty() && rightChildCorrelatedTo.isEmpty()) {
                 return tryBoxSelfAsConstantPredicate();
             }
+            if (rightChildCorrelatedTo.contains(innermostAlias) && !leftChildCorrelatedTo.contains(innermostAlias)) {
+                if (leftChildCorrelatedTo.isEmpty()) {
+                    final Object comparand = compileTimeEvalFn.apply(leftChild);
+                    return comparand == null
+                           ? Optional.empty()
+                           : Optional.of(new ValuePredicate(rightChild, new Comparisons.SimpleComparison(swapBinaryComparisonOperator(comparisonType), comparand)));
```
hatyo, line 169: Since `null` is a permissible value, we should return something else so the caller can distinguish this case.

hatyo (approved):

PR #1840, opened 2022-09-08
Remove key expressions from ordering

normen662:
- replaced `KeyExpression` with `Value` in `KeyPart` and `BoundKeyPart`
- as a consequence `Ordering` and `RequestedOrdering` now both use indirectly values to manage ordering
- comparison keys and primary keys that are used by `Plans are now `Value`s as well,
- generalization of transformation rules and transformation rules sets
- new mini optimizer (destructive) used to simplify `Value` trees
- lots of logic around translation of `Value` trees
- optimizations around the computation of properties in `PropertiesMap` (now done lazily); needed to be done as some properties cannot be computed if references are created from the old planner.


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueSimplificationTest.java
```diff
@@ -0,0 +1,417 @@
...
+                Type.Record.Field.of(Type.primitiveType(Type.TypeCode.STRING), Optional.of("z"))));
+    }
+
+    @Nonnull
+    private static Value defaultSimplify(@Nonnull final Value toBeSimplified) {
+        return toBeSimplified.simplify(DefaultValueSimplificationRuleSet.ofSimplificationRules(), AliasMap.emptyMap(), ImmutableSet.of());
+    }
+
+    @Nonnull
+    private static Value simplifyWithConstantAliases(@Nonnull final Value toBeSimplified, @Nonnull Set<CorrelationIdentifier> constantAliases) {
```
MMcM, line 409: Where is this used?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/EliminateArithmeticValueWithConstantRule.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.common.collect.ImmutableList;
+
+import javax.annotation.Nonnull;
+
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.MultiMatcher.all;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ValueMatchers.anyValue;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ValueMatchers.arithmeticValue;
+
+/**
+ * A rule that eliminates an arithmetic operation. One of the following conditions have to be met:
```
MMcM, line 37: These conditions are reasonable (up to, say, floating point loss of precision) for _ordering_ values.

But the package / naming / something seems to be missing that relates / restricts it to that case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantValue.java
```diff
@@ -0,0 +1,138 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * A wrapper around a constant.
```
MMcM, line 41: Needs some expansion of why / how this is used. It isn't immediately obvious what folding constants into one of these wrappers buys. What is the emergent behavior from this wrapper?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
```diff
@@ -0,0 +1,176 @@
...
+    @Nonnull
+    @SuppressWarnings("PMD.CompareObjectsWithEquals")
+    private static <R, C extends AbstractValueRuleCall<R, C>> Value executeRuleSet(@Nonnull final Value root,
+                                                                                   @Nonnull Value current,
+                                                                                   @Nonnull final AbstractValueRuleSet<R, C> ruleSet,
+                                                                                   @Nonnull final RuleCallCreator<R, C> ruleCallCreator,
+                                                                                   @Nonnull final Function<Collection<R>, Value> onResultsFunction) {
+        final boolean isRoot = current == root;
+        Value newCurrent = current;
+        do {
```
hatyo, line 120: does it make sense to add an upper bound of iterations to safe-guard against (in)directly-cyclic rules, i.e.:

rule1: v1 -> v2
rule2: v2 -> v3
rule3: v3 -> v4
...
rulen: vn -> v1

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyPart.java
```diff
@@ -20,32 +20,37 @@
...
 import javax.annotation.Nonnull;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Supplier;
 
 /**
- * A key expression that can be bound by a comparison.
+ * A value that is used to express ordered-ness.
  */
 public class KeyPart {
```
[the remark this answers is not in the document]
normen662: Yeah -- it does! I also want to rename `PartialOrder` into `PartiallyOrderedSet`, however, in a separate PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
```diff
@@ -226,6 +229,108 @@ public Optional<List<KeyPart>> satisfiesRequestedOrdering(@Nonnull final List<Ke
...
+    public Ordering pullUp(@Nonnull Value value, @Nonnull AliasMap aliasMap, @Nonnull Set<CorrelationIdentifier> constantAliases) {
+        //
+        // Need to pull every participating value of this ordering through the value.
+        //
+        final var orderingKeyParts = getOrderingKeyParts();
+        final var orderingKeyValues =
+                orderingKeyParts
+                        .stream()
+                        .map(KeyPart::getValue)
+                        .collect(ImmutableList.toImmutableList());
```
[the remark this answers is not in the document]
normen662: Same as above! Can do it -- just not in this PR as there is a big other PR after this one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
normen662: Hmm, we don't have a good way of configuring that although it is useful. On the other hand, you can make the case that simplification is monotonic (or should be). I think we can agree that this should only be possible if there is an actual bug in the rules, not if the candid `Value` is too complex (which can be the reason for the Cascades plane). Let me think about how to do that.

MMcM: Maybe a relatively high total time limit on planning? Just to avoid stuck threads due to bugs.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueSimplificationTest.java
normen662: Removed. I could swear there is a PMD rule that should have barked.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantValue.java
normen662: I can take that back out as it is as of today not actually used in any useful ways. It will allow the simplification logic to fold a `Value` tree into a simpler one where this _nested_ `Value` is considered extractable to a higher level, i.e. outside of the current query block into a higher one, or, if no further correlations are present, into some sort of a all-encompassing NLJN of a `SELECT inner FROM VALUES(all the constants as columns) v, ...thequery... as inner` where all the constants are now values of the shape `v.something`. This can help us with literal extractions, etc.

normen662: Should I take it back out?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
normen662: Opened https://github.com/FoundationDB/fdb-record-layer/issues/1854 to track this general problem.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/EliminateArithmeticValueWithConstantRule.java
normen662: I added a disclaimer to this rule. It is also only used in `OrderingValueSimplificationPerPartRuleSet`.

hatyo (approved):

MMcM (approved):

PR #1843, opened 2022-09-14
Allow carried columns to be returend from Lucene scans

scottfines:
so that LUCENE_STORED records can be returned during the scan. Future work may involve only doing this if it's explicitly required.


MMcM (changes requested):
I am not completely clear what problem this is solving that `LucenePlanner.getStoredFields` doesn't already handle. I don't see any fundamental objection to moving the stored fields from the pretend key tuple to the pretend value tuple, but I am not sure what I am missing. Likewise, I suspect that `getStoredFields` will need to be adjusted to keep `FDBLuceneQueryTest.covering` passing.

scottfines:
Closed as unnecessary

PR #1848, opened 2022-09-20
Resolves #1847: TypeReposistory does not allow for named enums

### LINKED ISSUE #1847: TypeReposistory does not allow for named enums
The TypeRepository currently always auto-generates the names for enums. However, in practice, I think we'll need enums to be able to take a user-supplied name. For example, that would be needed to look up the enum type associated with a field from its field's type name.

alecgrieser:
This adds a "name" field to the enum type used by the type repository so that the user can associate a name for the enum type. When a user later looks at field, the enum type can then be looked up via this name.

This resolves #1847.


normen662 (changes requested):
I think we need to discuss if we actually want this, as we didn't want to have named referential types. SQL in its heart uses equality by structure not by name. It introduced `UDTs` but I think that was a poor OR approach. What happens if the name is identical but the structure (that is the enums values in your case) differ somehow. We are building redundant information into the system that we would then have to keep in sync. I would propose to do exactly what we did for named types, i.e. allow them in the DDL via `CREATE STRUCT` but not on the query side.

alecgrieser:
I'm not sure I have too many opinions on the parts of this PR around the `equals` and `hashCode` implementations, and if we want to change those to be more consistent with structural equality, we can do that. I guess, in particular, if we wanted to make this to use a more "structural equality" approach, I can make the `.equals` method on the enum type exclude the name, and then two enum types will be considered equal if and only if they have the same fields.

I think it sort of depends on whether enums are considered by the type system to be more like strings that have a restricted set of values (and have a special ordering property), or if enums are more "their own thing", and we really do want two enums with the same values but different names to be different enums. I think I prefer the latter (and that they are different from structs in that way), but I could see it going either way.

In any case, my actual goal with this change was more around making sure that one could create an `Enum` type that could set a user-defined name, which is necessary for use for DDL use cases (if not query use cases).

alecgrieser:
Okay, this PR should now set enums up to be similar to records, in that they have a name field, but that is not used when constructing the enum unless explicitly set by the user using a somewhat special constructor. The name field is also not included in the `equals` and `hashCode` methods, so two enums with identical fields should compare equally. However, a user who uses the new constructor can set the name if they need to keep track of it (for referential reasons, say).

arnaud-lacurie:
Let's go ahead and merge this PR. This is localised enough that we can get back to the discussion and add further modifications down the line if need be.

normen662 (approved):

PR #1850, opened 2022-09-23
Implementing a JoinRecordPlan over concatenated fields

scottfines:
Honestly, I don't really know if this is the "right" way to solve this problem--it passes the test that I wrote, but I would welcome commentary on whether or not this keeps with the general design principles.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/ThenQueryKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/ThenQueryKeyExpression.java
@@ -1,0 +1,114 @@
+/*
+ * ThenQueryKeyExpression.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -741,14 +741,67 @@ public void aggregateJoinIndex() throws Exception {
         }
     }
 
+    @Test
+    public void concatenatedFields() throws Exception {
```
MMcM, line 745: I don't feel strongly, but suspect this could be a parameterized version of `multiFieldKeys`.

All these tests need plausibility assertions on `JoinedRecordPlan.queries`. Specifically, that they don't do full table scans. I believe this test will fail that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/ThenQueryKeyExpression.java
```diff
@@ -0,0 +1,114 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+
+/**
+ * A simple {@link QueryableKeyExpression} that covers a {@link ThenKeyExpression}, translating
+ * into a Tuple for comparison operations with {@link QueryKeyExpressionWithComparison}.
+ */
+public class ThenQueryKeyExpression implements QueryableKeyExpression {
```
MMcM, line 42: This approach obligates the planner to deal with `WHERE (a,b) = ?`, as opposed to `WHERE (a,b) = (?,?)` or `WHERE a = ? AND b = ?`. Today it only really handles the last one.

A `QueryableKeyExpression` needs to match an _element_ of the index's root expression. And this will likely be a subsequence.

An alternative approach that is perhaps more expedient is to treat this as two joins, either in `addJoin` right away or in `JoinedRecordPlanner`.

MMcM (changes requested):

scottfines:
Closed because it's unnecessary

PR #1851, opened 2022-09-27
Fix broke "Getting Started": add note on protobuf versions, use the `basic` gradle type, update dependencies, and mix in some missing info

saintstack:
This "Getting Started" page didn't work for me and confused. Here are some cleanups:
* Acknowledge we use protobuf3 but syntax is proto2.
 * Gradle `--type java-application` creates an 'app' subdir now with its own build.gradle -- so two build.gradles -- which complicates the example (example won't work if you edit the wrong build.properties). Lets do 'basic' to keep things simple.
 * Add answers for the questions gradle `init` asks.
 * plugins need to be added at the top of the build.gradle -- the example implies the clause can go anywhere -- and our protobuf plugin won't run if no mention of the java plugin.
 * Example code was missing the line where we build the RecordMetaData.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -37,13 +37,16 @@ Be sure you see the message `The database is available.` type `exit` to quit the
 fdb> exit
 ```
 
-Let’s create a fresh demo project. For this example, we’ll be creating a sample Java app with Gradle:
+Let’s create a fresh demo project. For this example, we’ll be creating a basic Java app with Gradle:
 
 ```bash
 $ mkdir record-layer-demo
 $ cd record-layer-demo
-$ gradle init --type java-application
+$ gradle init --type basic
 ```
+Gradle 'init' will ask you some questions on which DSL to use in build scripts,
+whether to use new APIs and behaviors, and how you would like to name your project.
+Choose `groovy`, `yes` to new APIs, and take the proffered project name `record-layer-demo`.
 
 Now we should add the Record Layer as a dependency of our project. The Record Layer dependencies
 are published to Maven Central, so we can declare the dependency by adding the following to our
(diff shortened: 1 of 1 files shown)


MMcM (changes requested):
It may well be that the best thing to do is to keep things simple, in which case this is fine. But, do note https://github.com/FoundationDB/fdb-record-layer/blob/main/docs/Overview.md#indexing-and-querying-of-missing--null-values.

Which means that even when using the pb3 library variant, it is still sometimes sensible to have `syntax = proto2` in order to get "explicit presence," which plays more nicely with 3VL.

Another possibility would be to use the proto 3 version via `optional` (https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3), though I am not confident that is properly supported everywhere.

In the same vein, there is the question of whether the default `Color` should be `RED` or a new `UNKNOWN = 0`.

Again, I'm actually okay with saying it's inappropriate to try to introduce such advanced notions in a "Getting Started" context. You're probably in the best position to make that judgment, while it's fresh.

docs/GettingStarted.md
```diff
@@ -228,10 +239,17 @@ Add a secondary index on `price`:
 metaDataBuilder.addIndex("Order", new Index("priceIndex", Key.Expressions.field("price")));
 ```
 
+
 The primary and secondary index definitions take a `Key.Expression` as an argument. It is possible to have more
 complex index definitions (e.g., compound primary keys or fan-out indexes on repeated fields).
 For more advanced examples, see the [Record Layer Overview](Overview.md).
 
+Now we have finished with configuraiton, build the `RecordMetaData`:
```
MMcM, line 247:
```suggestion
Now we have finished with configuration, build the `RecordMetaData`:
```

MMcM (changes requested):

docs/GettingStarted.md
saintstack: Fixed in latest push.

PR #1855, opened 2022-09-29
resolves #1868: update auto-service, guava; enable incremental gralde builds for deve…

### LINKED ISSUE #1868: enable incremental builds
Due to our use of `auto-service`, grade disables incremental builds. A newer version of `auto-service` fixes this issue, however, also changes the retention policy of `AutoService` to `CLASS` which causes problems for downstream consumers. In a sense, this causes us to be stuck with the old auto service, therefore forcing full rebuilds for single line changes which is not desirable.

normen662:
…lopment builds

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetHashFunctions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetHashFunctions.java
@@ -40,7 +40,8 @@ public class RankedSetHashFunctions {
     public static final String RANDOM = "RANDOM";
     public static final String MURMUR3 = "MURMUR3";
 
-    private static final RankedSet.HashFunction MURMUR3_HASH_FUNCTION = new GuavaHashFunction(Hashing.murmur3_32());
+    @SuppressWarnings("UnstableApiUsage")
+    private static final RankedSet.HashFunction MURMUR3_HASH_FUNCTION = new GuavaHashFunction(Hashing.murmur3_32_fixed());
 
     private static final BiMap<String, RankedSet.HashFunction> extent = ImmutableBiMap.<String, RankedSet.HashFunction>builder()
             .put(JDK, RankedSet.JDK_ARRAY_HASH)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/BoundRecordQuery.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/BoundRecordQuery.java
@@ -43,10 +43,10 @@ import static com.apple.foundationdb.record.query.expressions.BooleanComponent.g
  * A bound record query wraps an instance of {@link RecordQuery} and {@link RecordStoreState} in order to associate
  * the query with the state of a store. That allows for differentiation of queries depending on the store state, e.g.
  * the availability of indexes at the time the query was bound to the state, etc.
- *
+ * <br>
  * Bound record queries also define {@link #hashCode()} and {@link #equals(Object)} in a way that allows for queries to
  * be considered compatible to each other which is useful with respect to pre-bound parameter markers.
- *
+ * <br>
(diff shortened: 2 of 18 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/BoundRecordQuery.java
```diff
@@ -134,7 +134,7 @@ public int hashCode() {
 
     @SuppressWarnings("UnstableApiUsage")
     private int computeHashCode() {
-        final HashFunction hashFunction = Hashing.murmur3_32();
+        final HashFunction hashFunction = Hashing.murmur3_32_fixed();
```
alecgrieser, line 137: I suspect you know this better than me, but this hash function is slightly different from before (due to a bug fix), so if there are problems caused by this function being unstable across releases, it's possible we need to continue using the older (deprecated) function.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankedSetHashFunctions.java
```diff
@@ -40,7 +40,8 @@ public class RankedSetHashFunctions {
     public static final String RANDOM = "RANDOM";
     public static final String MURMUR3 = "MURMUR3";
 
-    private static final RankedSet.HashFunction MURMUR3_HASH_FUNCTION = new GuavaHashFunction(Hashing.murmur3_32());
+    @SuppressWarnings("UnstableApiUsage")
+    private static final RankedSet.HashFunction MURMUR3_HASH_FUNCTION = new GuavaHashFunction(Hashing.murmur3_32_fixed());
```
alecgrieser, line 44: I figured I'd mention it here since I mentioned it with the other usage: here I'm pretty sure that there isn't a requirement for stability across releases. In fact, there's a "random" hash function implementation here that is unstable across executions, let alone releases, so this change here should be okay. (That's also, to my understanding, how the `RankedSetHashFunctions` are supposed to work.)

gradle.properties
```diff
@@ -39,10 +39,12 @@ commonsLang3Version=3.12.0
 commonsMath3Version=3.6.1
 javaPoetVersion=1.12.0
 log4jVersion=2.17.2
-guavaVersion=30.1-jre
+guavaVersion=31.0.1-jre
 hamcrestVersion=2.2
 # AutoService kept on 1.0-rc6 to avoid annotation being retained in CLASS. See: https://github.com/FoundationDB/fdb-record-layer/issues/1281
 autoServiceVersion=1.0-rc6
+# AutoService for development
```
alecgrieser, line 46: Might be worth noting in a comment here that this is bumped during development to enable incremental builds.

gradle.properties
```diff
@@ -39,10 +39,12 @@ commonsLang3Version=3.12.0
 commonsMath3Version=3.6.1
 javaPoetVersion=1.12.0
 log4jVersion=2.17.2
-guavaVersion=30.1-jre
+guavaVersion=31.0.1-jre
```
alecgrieser, line 42: It's possible we should increment the minor version of the Record Layer prior to merging this and warn people about the guava version bump

alecgrieser (changes requested):

gradle.properties
normen662: Upgraded as discussed to `31.1-jre`

gradle.properties
```diff
@@ -19,7 +19,7 @@
 #
 
 rootProject.name=fdb-record-layer
-version=3.2
+version=3.3
```
alecgrieser, line 22: I believe there are some more changes that are needed to properly update the minor version. For example, see how it was done in this commit: https://github.com/alecgrieser/fdb-record-layer/commit/1d17f61572db7465250733de34910c344da77802

In particular, the Release Notes should be updated to add a new header for the new minor version, and the `docs/Versioning.md` gets updated with the most recent version just to avoid that getting stale. It's also a good idea to add a paragraph to the new section so that it's easy for users to see what has changed.

PR #1866, opened 2022-10-12
Allowing JoinedRecordTypes to have two different KeyExpressions for each side

scottfines:
for use when constructing the join and also when constructing the index


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
```diff
@@ -62,18 +62,25 @@ public static class Join {
...
-        private final KeyExpression rightExpression;
+        private final KeyExpression rightSourceExpression;
+
+        @Nonnull
+        private final KeyExpression rightValueExpression;
 
-        protected Join(@Nonnull JoinConstituent left, @Nonnull KeyExpression leftExpression,
-                       @Nonnull JoinConstituent right, @Nonnull KeyExpression rightExpression) {
+        protected Join(@Nonnull JoinConstituent left, @Nonnull KeyExpression leftSourceExpression, @Nonnull KeyExpression leftValueExpression,
+                       @Nonnull JoinConstituent right, @Nonnull KeyExpression rightSourceExpression, @Nonnull KeyExpression rightValueExpression) {
```
alecgrieser, line 77: Could you go into a bit more detail into what the idea behind these new expressions is, and why they were necessary?

fdb-record-layer-core/src/main/proto/record_metadata.proto
```diff
@@ -195,9 +195,11 @@ message JoinedRecordType {
...
   message Join {
     optional string left = 1;
-    optional KeyExpression left_expression = 2;
-    optional string right = 3;
-    optional KeyExpression right_expression = 4;
+    optional KeyExpression left_source_expression = 2;
+    optional KeyExpression left_value_expression = 3;
+    optional string right = 4;
+    optional KeyExpression right_source_expression = 5;
+    optional KeyExpression right_value_expression = 6;
```
MMcM, line 202: Are we confident that there aren't any serialized anywhere? Wouldn't it be better to have the old fields align?

alecgrieser (commented):
I still need to do a full review, but I figured I'd start with this, because I'm a bit confused what this PR is trying to accomplish, and so getting clarity on this would be useful

MMcM (changes requested):

fdb-record-layer-core/src/main/proto/record_metadata.proto
alecgrieser: It probably would be good to align the fields/field numbers, I think, and to consider what to do if the new fields are missing to handle backwards compatibility

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
scottfines: If you want to join a string field to a numeric field, then you have to be able to supply a function key expression which can be applied when performing the join; that FKE is only useful when you are doing the join and shouldn't be used when actually writing the index, so you need two expressions--one for the index storage construction, and one to apply when performing the join

scottfines: Added documentation and changed variable names to reflect their relative roles

fdb-record-layer-core/src/main/proto/record_metadata.proto
scottfines: done

scottfines:
it appears that there might be a race condition when writing and/or deleting lucene data, causing one of the tests to fail here.

.idea/compiler.xml
```diff
@@ -39,17 +39,16 @@
       <profile name="Gradle Imported" enabled="true">
```
MMcM, line 39: It looks like this was included by accident. But maybe it's correct to update it if dependencies have changed (from an earlier pull request)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
```diff
@@ -73,17 +74,46 @@ public static class Join {
...
+        private final KeyExpression rightJoinExpression;
 
-        public Join(@Nonnull String left, @Nonnull KeyExpression leftExpression, @Nonnull String right, @Nonnull KeyExpression rightExpression) {
+        /**
+         * Create a new join from left to right.
+         *
+         * @param left the name of the type on the LHS of the join.
+         * @param leftIndexExpression the KeyExpression to use when constructing indexes. The entries in this
+         * KeyExpression
+         * must line up correctly with how the join index will store records on disk.
```
MMcM, line 93: This doesn't seem to quite capture it. In particular, there is. no _requirement_ that these keys' values be stored in the index at all. Often they are the primary key on one side, which is stored, but nothing prevents joining on some kind of surrogate instead.

I'm trying to think what the most succinct way to describe what's going on it, without introducing a bunch of higher level machinery down here in the model. Perhaps the missing concept is the _synthesis_. Given one of the (assume two, but it extends to n) constituents, how to we find the other to build the synthetic joined record? It's actually as simple as could be: take an _index_ key expression from the record we have and evaluate it into a query binding. Then build a query where the _join_ key expression from the other side equals that binding. The results of this query are the other records we need to synthesize.

All indexing (which is not restricted by index type -- "join indexes" are a classification of the derivation, not a type of index) happens on the synthetic records, which have nested fields for each component and a compound primary key.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
```diff
@@ -122,7 +162,8 @@ public JoinedRecordTypeBuilder(@Nonnull RecordMetaDataProto.JoinedRecordType typ
             addConstituent(joinConstituent.getName(), metaDataBuilder.getRecordType(joinConstituent.getRecordType()), joinConstituent.getOuterJoined());
         }
         for (RecordMetaDataProto.JoinedRecordType.Join join : typeProto.getJoinsList()) {
-            addJoin(join.getLeft(), KeyExpression.fromProto(join.getLeftExpression()), join.getRight(), KeyExpression.fromProto(join.getRightExpression()));
+            addJoin(join.getLeft(), KeyExpression.fromProto(join.getLeftExpression()), KeyExpression.fromProto(join.getLeftJoinExpression()),
+                    join.getRight(), KeyExpression.fromProto(join.getRightExpression()), KeyExpression.fromProto(join.getRightJoinExpression()));
```
alecgrieser, line 166: I don't think this is backwards compatible. In particular, if an existing `JoinedRecordType` does not have the `left_join_expression` and `right_join_expression` set (because it was serialized before those fields existed), this will now give `KeyExpression.toProto` a default instance of the proto which will throw a `DeserializationException`. I believe the right behavior is to instead insert the `left_expression` and `right_expression` if the proto does not have the join expression set.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
```diff
@@ -167,22 +213,59 @@ public Join addJoin(@Nonnull String left, @Nonnull KeyExpression leftExpression,
...
+     * @return a representation of the join.
+     */
+    @Nonnull
+    public Join addJoin(@Nonnull String left, @Nonnull KeyExpression leftIndexExpression, @Nonnull KeyExpression leftJoinExpression,
+                        @Nonnull String right, @Nonnull KeyExpression rightIndexExpression, @Nonnull KeyExpression rightJoinExpression) {
+        if (leftIndexExpression.getColumnSize() != rightIndexExpression.getColumnSize()) {
+            throw new RecordCoreArgumentException("Two sides of join are not the same size and will never match")
+                    .addLogInfo("left", leftIndexExpression, "right", rightIndexExpression);
+        }
+        //TODO(bfines) validate column sizes for the value expressions also
```
alecgrieser, line 248: If I'm understanding this PR correctly, are the actual requirements that `leftIndexExpression.getColumnSize() == rightJoinExpression.getColumnSize()` and `leftJoinExpression.getColumnSize() == rightIndexExpression.getColumnSize()`? Because I _think_ the only things that get compared are the index expressions from one side and the join expressions from the other.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
scottfines: I worked out an example case and tried to explain what was going on that way.

.idea/compiler.xml
scottfines: I only added it to force a rebuild in rec-layer, will remove

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
scottfines: that is correct--I adjusted the column size checks to account for this

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordTypeBuilder.java
scottfines: fixed.

MMcM:
Alec gives a more compartmentalized way of thinking about it.

If the desired join is
```
tl.cl = func(rl.rc)
```

we sometimes need the equivalent
```
func ⁻¹(tl.cl) = rl.rc
```

The two new key expressions are the function and its inverse.

Alternatively, if there were some way for a single key expression to give its inverse or a registry of them, there could just be one in the persisted form.

MMcM:
> One is if there's an index on `func(rec1.some_field)`, then the condition can be satisfied by scanning that index. 

You would expect there to be an index on `rec1.some_field`, so this function index is no worse than that, unless the simpler one is also useful for other queries.

alecgrieser:
Yeah, that's what I was thinking, regarding using function key expressions. Given that the planner currently doesn't support matching a function key expression, the way I was thinking of doing it in version 0 is to always assume the query needs to be `tr.cr = func^-1(tl.cl)`, which means that an explicit index on `func(tr.cr)` wouldn't be usable, but further iterations on the join planner and query planner would allow us to relax that constraint.

MMcM:
> Given that the planner currently doesn't support matching a function key expression ...

Note that the planner _does_ support matching a `QueryableKeyExpression`, provided that is what is _stored_ in the index. So the case where the function is applied to match the primary key, say, and that primary key matches a stored version of the same function application for the inverse direction, should already work and we should try not to break it. The case under discussion here is where the available index does not include any function, so that its inverse needs to be applied to the record in hand.

alecgrieser:
Ah, right, I guess I overlooked `QueryableKeyExpressions`. I think it should be possible to add support for those, too.

I've filed #1872 about this, and I suspect whichever PR we take, we'll want to associate that issue. I've got a draft PR, #1873, which is designed to resolve it. It got somewhat more complicated with `IN` comparisons due to how the `InExtractor` works, so the PR ballooned a bit more than I expected. I have some local test failures (hence it being in draft), but I am working on resolving those

alecgrieser:
Should this be closed now that #1872 is closed?

scottfines:
Closing because it was superseded by InvertibleFunctionKeyExpressions

PR #1867, opened 2022-10-12
Upsert

normen662:


hatyo (approved):
Good work!

I think there is a room for improving the implementation a bit, i.e. during construction, less recursive calls, less maybe using mutable data structures to avoid too many small allocations stressing GC.

I think it is also worth checking whether we can use a more advanced structure such as a compressed trie (easy win) or perhaps a suffix tree (trickier I think), but this should be driven by measurements of real workloads.

PR #1873, opened 2022-10-17
Resolves #1872: Joined record types should support function key expressions in matching expression

### LINKED ISSUE #1872: Joined record types should support function key expressions in matching expression
Joined record types currently don't support function key expressions in their join match expressions. So, that means that you can't create a joined record type representing `FROM left JOIN right WHERE left.col1 = f(right.col2)` (and variations). There are cases that one might want to do this. For example, suppose that `left.col1` and `right.col2` are of different types, and `f` is a function that casts from one type to another.

alecgrieser:
This improves the `JoinedRecordPlanner` by adding matching logic to handle invertible functions. The idea there is that for some functions, we can transform query predicates like `SELECT * FROM t WHERE f(t.col) = $value` into `SELECT * FROM t WHERE t.col = f^-1($value)`. This is then used within the join planner to construct the queries for join matches, which in turn allows for `JoinedRecordTypes` to be constructed (and updated) with join conditions like `t1 JOIN t2 ON t1.col1 = f(t2.col2)`. At join resolution time, when updating a `t2` record, the matching `t1`s will be found by executing a query like `SELECT * FROM t1 WHERE t1.col1 = $val` where `$val` is the result of evaluating `f(t2.col2)` (this worked before this PR), and when updating a `t1` record, the matching `t2`s will be found by executing a query like `SELECT * FROM t2 WHERE t2.col2 = f^-1($val)` where `$val` is `t1.col1` (this is what's updated in this PR).

This still isn't perfect, though, as it is currently limited to unary functions, i.e., functions with a single column argument that produce a single column result. It does have support for both injective functions (where the pre-image of a single value is guaranteed to have at most one element) and non-injective functions (where the pre-image can have multiple values) so long as the entire pre-image can be enumerated.

This ballooned a little bit, mainly because of how `InExtraction` worked, which necessitated creating additional query plans to handle the new kind of comparison. 

This resolves #1872.


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -741,6 +751,298 @@ public void aggregateJoinIndex() throws Exception {
         }
     }
 
+    @Test
+    void joinOnNestedKeysWithDifferentTypes() throws Exception {
+        metaDataBuilder.getRecordType("CustomerWithHeader").setPrimaryKey(Key.Expressions.concat(field("___header").nest("z_key"), field("___header").nest("int_rec_id")));
+        metaDataBuilder.getRecordType("OrderWithHeader").setPrimaryKey(Key.Expressions.concat(field("___header").nest("z_key"), field("___header").nest("rec_id")));
+
```
MMcM, line 758: It might be more realistic to include something like
```
        metaDataBuilder.addIndex("OrderWithHeader", "order$custRef", field("custRef").nest("string_value"));
```
around here.

It does not affect correctness and the interesting case for this branch is the other direction, where a primary key scan's bounds come from a function inverse.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/InvertibleFunctionKeyExpression.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+
+/**
+ * Key expression representing an invertible function. A function is invertible if it is possible to determine
+ * the pre-image of function application given the result. For example, if the function represents the squaring
+ * function, then the inverse would represent the square root. Note that this class allows the inverse to return
+ * multiple values, if there are multiple values that may map to the same output.
+ */
+@API(API.Status.EXPERIMENTAL)
+public abstract class InvertibleFunctionKeyExpression extends FunctionKeyExpression {
```
MMcM, line 38: Perhaps a couple of unit tests of this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/JoinedRecordPlanner.java
```diff
@@ -261,9 +271,138 @@ private static QueryComponent buildCondition(@Nonnull KeyExpression expression,
...
+        if (function.getColumnSize() != 1) {
+            throw new RecordCoreException("invertible functions in joins must return a single element");
+        }
+        if (function.getMinArguments() != 1 || function.getMaxArguments() != 1) {
+            throw new RecordCoreException("invertible functions in joins must operate on a single element");
+        }
+        return new InvertFunctionComparisonWithParameter(function, comparison);
+    }
+
+    private static class InvertFunctionComparisonWithParameter implements Comparisons.Comparison {
```
MMcM, line 295: Is it worth trying to have this outside of just the joined record planner? I guess the main question is how you'd ever get one of these comparisons otherwise, absent improvements to `QueryKeyExpressionWithComparison`.

But it would allow for a unit test that planning a suitable inverted comparison gives an index scan with bounds computed by inverting a parameter value at execution time.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/InvertibleFunctionKeyExpression.java
alecgrieser: Added a few to `InvertibleFunctionKeyExpressionTest`.

MMcM (approved):

alecgrieser:
New push rebases on top of #1875 and then adds asserts to the new tests based on new the `SyntheticPlanMatcher`s that that PR introduces

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -825,6 +831,363 @@ public void aggregateJoinIndex() throws Exception {
...
+            final SyntheticRecordPlanner planner = new SyntheticRecordPlanner(recordStore);
+            JoinedRecordType joinedRecordType = (JoinedRecordType) recordStore.getRecordMetaData().getSyntheticRecordType(joined.getName());
+            assertConstituentPlansMatch(planner, joinedRecordType, Map.of(
+                    "order",
+                    SyntheticPlanMatchers.joinedRecord(List.of(
+                            PlanMatchers.inComparand(PlanMatchers.hasTypelessString("wrap_int^-1($_j2)"),
+                                    PlanMatchers.typeFilter(Matchers.contains("CustomerWithHeader"),
+                                            PlanMatchers.scan(PlanMatchers.bounds(PlanMatchers.hasTupleString("[EQUALS $_j1, EQUALS $__in_int_rec_id__0]"))))))),
+                    "cust",
+                    // This should be able to use the OrderWithHeader$cc index, but for some reason, is favoring a full scan
```
alecgrieser, line 1018: I looked into this, and I believe the problem is #1876 (and not anything to do with joins, *per se*). I have a proposed fix in #1877, though it deals with nested repeated structures, so there may be dragons I'm missing. (That PR is also built on top of this one so that it can change the plan in this test.) AFAICT, this is only a problem with the old planner, though perhaps it needs more testing to verify that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -2159,4 +2163,173 @@ public String toString() {
...
+                            .filter(eval -> eval.size() == 1)
+                            .forEach(eval -> finalValues.add(eval.getObject(0)));
+                }
+                return finalValues;
+            } else {
+                Key.Evaluated evaluated = Key.Evaluated.scalar(originalComparandValue);
+                List<Key.Evaluated> inverse = function.evaluateInverse(evaluated);
+                if (getType() == Type.IN) {
+                    return inverse.stream()
+                            .filter(eval -> eval.size() == 1)
```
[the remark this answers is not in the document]
alecgrieser: Yeah, that sounds like a good idea. I've refactored this a bit so that all branches throw the same `RecordCoreException` in this case.

alecgrieser:
Rebase addresses merge conflicts (mainly in `SyntheticRecordPlannerTest`)

hatyo (approved):
LGTM

alecgrieser:
Rebase addresses merge conflict in release notes

MMcM (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1036,6 +1042,363 @@ public void aggregateJoinIndex() throws Exception {
...
+            try (RecordCursor<IndexEntry> cursor = recordStore.scanIndex(joinIndex, IndexScanType.BY_VALUE, TupleRange.ALL, null, ScanProperties.FORWARD_SCAN)) {
+                List<IndexEntry> entries = cursor.asList().get();
+                assertEquals(1, entries.size());
+                final IndexEntry indexEntry = entries.get(0);
+                assertEquals("Alec", indexEntry.getKey().getString(0), "Incorrect customer name");
+                assertEquals(42L, indexEntry.getKey().getLong(1), "Incorrect order number");
+            }
+
+            // Insert an order with no associated customer
+            TestRecordsJoinIndexProto.OrderWithHeader orderWithNoCustomer = TestRecordsJoinIndexProto.OrderWithHeader.newBuilder()
```
scottfines, line 1144: Not a priority really, but these feel like we would want a separate test for these cases, so that we can see which parts of this test fail. Otherwise, the test itself could fail for a lot of different reasons that are hard to parse.

scottfines (approved):
LGTM. I added one minor request to split out the test in `SyntheticRecordPlannerTest` into smaller tests to make it easier to track, but I'm ok merging as is

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
alecgrieser: Sure. I think my preference would be to do this in a separate PR. I've filed this issue to track it: #1887

PR #1874, opened 2022-10-19
Make time-window leaderboard rebuild when directory first added …

### LINKED ISSUE #1878: TimeWindowLeaderboardWindowUpdate rebuild = NEVER is not honored for new directory
It currently rebuilds anyway. Really, though, this is no different from the case where there are existing records overlapping new windows, for which `NEVER` is allowed with the understanding that the caller will arrange for the rebuild to happen.

MMcM:
… conditional in the same way as when existing records overlap a new window.

Specifically, if `rebuild` = `NEVER`, log a message but do not rebuild anyway. Caller will need to take care of it.

Fixes  #1878.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -15,7 +15,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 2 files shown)


alecgrieser (changes requested):
LGTM, other than that this should probably have a release note.

It also took me a while to entirely understand what this was doing, but I think I understand now: before this change, when the `UpdateState` operation was performed on a time window leaderboard index with an empty directory, it would cause the index to be rebuilt even if the rebuild enum was set to `NEVER`. Now, it'll treat the creation of the directory as a change, and so rebuild if the enum is set to `IF_OVERLAPPING_CHANGED`, but not if set to `NEVER`.

I think that makes sense, though I'm not super fluent with the time-window leaderboard index.

MMcM:
> I think I understand now

That all sounds right, yes.

MMcM:
> this should probably have a release note.

Added.

alecgrieser (approved):

PR #1875, opened 2022-10-19
Add Matchers for SyntheticRecordPlanner outputs.

MMcM:
This comment https://github.com/FoundationDB/fdb-record-layer/pull/1873#discussion_r997918275 suggested that matching on the generated plans would be beneficial. This adds that for all the direct calls to the planner in the test class.

The matchers only go against the generated queries inside the synthetic plan objects. It would be straightforward to extend them to match on other pieces of state, if any are deemed important enough. I felt that while they are, of course, necessary for execution, they weren't. But this wasn't strongly felt.


alecgrieser (approved):

alecgrieser:
This looks great! And it should be helpful for being confident that the plans there are as we expect them in #1873

PR #1877, opened 2022-10-20
Fixes #1876: AndWithThenPlanner does not handle OneOfThemWithComponent expressions

### LINKED ISSUE #1876: AndWithThenPlanner does not handle OneOfThemWithComponent expressions
Within the `RecordQueryPlanner`, the `AndWithThenPlanner` does not handle `OneOfThemWithComponent`s. In practice, this means that if you had a record type like:

```proto
message ExampleType {
    message NestedType {
        optional string value = 1;
    }
    optional int64 id = 1;
    repeated NestedType parent = 2;
    optional int64 other = 3;
}
```

And a value index with an expression like:

```java
concat(field("other"), field("parent", FanType.FanOut).nest("value"))
```

Then you'd expect that index to satisfy the following query:

```java
RecordQuery.newBuilder()
    .setRecordType("ExampleType")
    .setFilter(Query.and(
        Query.field("other").equalsParameter("o"),
        Query.field("parent").oneOfThem().matches(Query.field("value").equalsParameter("v"))
     ))
     .build();
```

However, because the `AndWithThenPlanner` ignores `OneOfThemWithComponent`s, this will get planned some other way (for example, using a less efficient index or scanning the record store).

alecgrieser:
This adds support to the old planner to address a shortcoming where certain queries were choosing to use less efficient indexes due to a shortcoming with the `AndWithThenPlanner`. Based on the tests in the `NestedFieldQueryTest`, it looks like the new planner was already matching such expressions in this case, and so this shortcoming should only be in the old planner.
    
This fixes #1876.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -17,7 +17,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 
 * **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** The `RecordQueryPlanner` now chooses more efficient indexes for and-queries with one-of-them predicates on repeated nested fields [(Issue #1876)](https://github.com/FoundationDB/fdb-record-layer/issues/1876)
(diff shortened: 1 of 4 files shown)


MMcM (commented):
Other than needing to be pulled out into its own PR, this looks reasonably straightforward and relatively safe. The new code is similar to the old in all the expected ways.

alecgrieser:
Rebased now that #1873 is in. I also ended up needing to change a plan assertion in the modified test because of a (I suspect unexpected) consequence of #1888. I'm not sure how much we _really_ care, because (as the test demonstrates), the new planner handles that case better anyway, though maybe there are some more subtly worsened plans.

normen662 (approved):

PR #1879, opened 2022-10-24
plan index scans and covering index scans over synthetic record types

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1424,14 +1426,38 @@ private RecordQueryPlan planScan(@Nonnull CandidateScan candidateScan,
...
+            plan = new RecordQueryIndexPlan(candidateScan.index.getName(), candidateScan.planContext.commonPrimaryKey, indexScanComparisons, indexFetchMethod, fetchIndexRecords, candidateScan.reverse, strictlySorted);
             possibleTypes = getPossibleTypes(candidateScan.index);
         }
         // Add a type filter if the query plan might return records of more types than the query specified
         plan = addTypeFilterIfNeeded(candidateScan, plan, possibleTypes);
         return plan;
     }
 
+    @Nonnull
+    RecordQueryIndexPlan.FetchIndexRecords resolveFetchIndexRecords(@Nonnull CandidateScan candidateScan) {
```
alecgrieser, line 1445: I'm not sure I follow the logic in this method. I take it, it's something like:

1. If there are no synthetic record types in the meta-data, then fetch by primary key (okay)
2. If there are synthetic record types and all of the types in the query are synthetic, then fetch by synthetic constituents (okay)
3. Otherwise, if there are no types being queried (I guess in which case the fetch method doesn't matter? I'm not sure I follow) or at least one queried type is not synthetic, fetch by primary key

It's mainly that third condition that I'm not sure I follow. Should this throw an error if there are both synthetic and non-synthetic record types in a single query (or handle it somehow differently)? And what does an empty list of queried record types signify (and thus why is it okay to fetch by primary key here)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/RecordQueryPlannerSubstitutionVisitor.java
```diff
@@ -102,6 +102,7 @@ public static RecordQueryPlan removeIndexFetch(@Nonnull RecordMetaData recordMet
                 // Need the primary key, even if it wasn't one of the explicit result fields.
                 fields.addAll(commonPrimaryKey.normalizeKeyForPositions());
             }
+            fields.removeIf(KeyExpression::canBeComputedWithoutMessage);
```
alecgrieser, line 105: What is this transformation attempting to accomplish? Is it just that you can get the record type key (and any literal or empty expressions) without needing to fetch the record, so if those are the only fields missing from the required results, then this should still allow the index plan to be used as a covering index?

I'm not sure this is totally right. I guess the other way of doing this would be to somehow tell the `AvailableFields` data structure how to resolve a `recordType` key expression. That might be necessary if this is ever generalized to work on multi-type indexes (which I see that it's not designed for at the moment).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -685,6 +685,15 @@ default CompletableFuture<FDBStoredRecord<M>> loadRecordAsync(@Nonnull final Tup
...
     CompletableFuture<Void> preloadRecordAsync(@Nonnull Tuple primaryKey);
 
+    /**
+     * Load a {@link FDBSyntheticRecord synthetic record} by loading its stored constituent records and synthesizing it from them.
+     * @param primaryKey the primary key of the synthetic record, which includes the primary keys of the constituents
+     * @return a future which completes to the synthesized record
+     */
+    @Nonnull
+    @API(API.Status.EXPERIMENTAL)
+    CompletableFuture<FDBSyntheticRecord> loadSyntheticRecord(@Nonnull Tuple primaryKey);
```
alecgrieser, line 695: Is the `FDBRecordStore` implementation missing from the PR right now? I couldn't find it, but the PR still compiled, which surprised me.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -619,6 +639,40 @@ public void joinIndex() throws Exception {
...
+
+        try (FDBRecordContext context = openContext()) {
+            final FDBRecordStore recordStore = recordStoreBuilder.setContext(context).open();
+
+            final QueryPlanner planner = setupPlanner(recordStore, null);
+
+            RecordQuery query = RecordQuery.newBuilder()
+                    .setRecordType("Simple_Other")
+                    .setFilter(Query.field("simple").matches(Query.field("str_value").equalsValue("even")))
+                    //.setRequiredResults(ImmutableList.of(field("simple").nest("str_value"), field("other").nest("num_value_3")))
```
alecgrieser, line 652: What is the behavior if this is uncommented?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -619,6 +639,40 @@ public void joinIndex() throws Exception {
...
+                    .build();
+
+            // Scan([[photos],[photos]])
+            RecordQueryPlan plan = planner.plan(query);
```
alecgrieser, line 656: Is that the plan that gets chosen by the planner?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBTypedRecordStore.java
```diff
@@ -136,6 +136,12 @@ public CompletableFuture<Void> preloadRecordAsync(@Nonnull Tuple primaryKey) {
         return untypedStore.preloadRecordAsync(primaryKey);
     }
 
+    @Nonnull
+    @Override
+    public CompletableFuture<FDBSyntheticRecord> loadSyntheticRecord(@Nonnull final Tuple primaryKey) {
+        return untypedStore.loadSyntheticRecord(primaryKey);
```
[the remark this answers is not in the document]
normen662: Unfortunately, all `execute()` on a `RecordQueryPlan` get handed a `FDBRecordStoreBase<M>` and I don't want to `default` add its implementation in `FDBRecordStoreBase<M>` as there is some logging stuff that does not carry over. I will leave it as this unless you feel strongly about this.

fdb-record-layer-spatial/src/main/java/com/apple/foundationdb/record/spatial/geophile/GeophilePointWithinDistanceQueryPlan.java
```diff
@@ -60,7 +60,8 @@
  */
 @API(API.Status.EXPERIMENTAL)
 @SuppressWarnings({"squid:S1206", "squid:S2160", "PMD.OverrideBothEqualsAndHashcode"})
-public class GeophilePointWithinDistanceQueryPlan extends GeophileSpatialObjectQueryPlan {
+public
+class GeophilePointWithinDistanceQueryPlan extends GeophileSpatialObjectQueryPlan {
```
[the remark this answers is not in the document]
normen662: whoops, how did that happen?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -135,32 +141,47 @@ public static AvailableFields fromIndex(@Nonnull RecordType recordType,
...
-        index.trimPrimaryKey(primaryKeys);
-        keyFields.addAll(primaryKeys);
+        final ListMultimap<KeyExpression, ImmutableIntArray> primaryKeyPartsToTuplePathMap;
+        if (commonPrimaryKey == null) {
+            primaryKeyPartsToTuplePathMap = ImmutableListMultimap.of();
+        } else {
+            final BitSet coveredPrimaryKeyPositions = index.getCoveredPrimaryKeyPositions();
+            final ExpressionToTuplePathVisitor expressionToTuplePathVisitor =
+                    new ExpressionToTuplePathVisitor(commonPrimaryKey, keyFields.size(), coveredPrimaryKeyPositions);
+            primaryKeyPartsToTuplePathMap = expressionToTuplePathVisitor.compute();
```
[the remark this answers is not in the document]
normen662: Mike and I have been discussing what should be getting this treatment, full index keys, or just the primary key. In the name of stability, I decided to only do this with minimum exposure to get the use case working.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -144,9 +148,25 @@ public void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Mess
...
         }
 
+        @Nullable
+        private static Object getFromTupleLikeForOrdinalPath(@Nonnull Tuple tuple, @Nonnull final ImmutableIntArray indexPath) {
+            Object value = tuple;
+            for (int i = 0; i < indexPath.length(); i ++) {
+                if (value instanceof Tuple) {
+                    value = ((Tuple)value).get(indexPath.get(i));
+                } else {
+                    value = ((List<?>)value).get(indexPath.get(i));
```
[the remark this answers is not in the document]
normen662: The top level `Tuple` cannot be `null`. All nested ones can be `null` but that is checked in `159`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -144,9 +148,25 @@ public void copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Mess
...
+                }
+                if (value == null) {
+                    return null;
```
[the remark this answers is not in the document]
normen662: That's a really good catch. I will try to see what's happening in the nested copier but I think that you are right and in the case all fields are `null` I will still have to force the sub record to be set to an empty message.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: No types (empty list) being queried is a special case of query everything. I will change the universal index maintainers to ignore synthetic types, so that in this case no synthetic records will ever be returned. Will add comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/RecordQueryPlannerSubstitutionVisitor.java
normen662: Fields of `RecordTypeKeyExpression` cannot be in `AvailableFields` unless we reimplement that logic in there as that logic wants to create copiers which you cannot do for this case. Also, it's really not made available in the index, but just an idiosyncrasy that some key expressions can be evaluated regardless of what the index defines, e.g. constants, record type.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
normen662: it's implemented in `FDBRecordStore`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
normen662: I made this better in the meantime to cause covering index scan, regular index scan, and fetch over covering to be tested. Will remove this left-over.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
normen662: yeah

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/UnionVisitor.java
```diff
@@ -93,16 +95,30 @@ public RecordQueryPlan postVisit(@Nonnull final RecordQueryPlan recordQueryPlan)
                     }
                     oldPlan = ((RecordQueryFilterPlan) oldPlan).getChild();
                 }
+                @Nullable RecordQueryFetchFromPartialRecordPlan.FetchIndexRecords currentFetchIndexRecords = resolveFetchIndexRecordsFromPlan(oldPlan);
```
alecgrieser, line 98: It is a little awkward that this logic needs to be repeated in the `IntersectionVisitor` and the `UnionVisitor`. I'm not sure if there's a better place, though, to put a helper method that extracts the common fetch type (or returns `null`).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -619,6 +638,63 @@ public void joinIndex() throws Exception {
...
+            final FDBRecordStore recordStore = recordStoreBuilder.setContext(context).open();
+
+            final QueryPlanner planner = setupPlanner(recordStore, null);
+
+            RecordQuery query = RecordQuery.newBuilder()
+                    .setRecordType("Simple_Other")
+                    .setFilter(Query.field("simple").matches(Query.field("str_value").equalsValue("even")))
+                    .build();
+
+            RecordQueryPlan plan = planner.plan(query);
```
alecgrieser, line 652: Should there be assertions here on the plan (e.g., that it uses the join index)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1424,14 +1427,37 @@ private RecordQueryPlan planScan(@Nonnull CandidateScan candidateScan,
...
+            plan = new RecordQueryIndexPlan(candidateScan.index.getName(), candidateScan.planContext.commonPrimaryKey, indexScanComparisons, indexFetchMethod, fetchIndexRecords, candidateScan.reverse, strictlySorted);
             possibleTypes = getPossibleTypes(candidateScan.index);
         }
         // Add a type filter if the query plan might return records of more types than the query specified
         plan = addTypeFilterIfNeeded(candidateScan, plan, possibleTypes);
         return plan;
     }
 
+    @Nonnull
+    private FetchIndexRecords resolveFetchIndexRecords(@Nonnull CandidateScan candidateScan) {
```
alecgrieser, line 1446: I still feel like this method needs a bit more explanation in comments in the code, given that it's a little confusing. In particular, noting what empty lists here signifies is probably a good idea. I also am inclined to say that it should handle (even if it means throwing an error) the case where some of the queried record types are synthetic and some of the queried record are not.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
```diff
@@ -230,6 +230,15 @@ default boolean hasLosslessNormalization() {
         return !createsDuplicates();
     }
 
+    /**
+     *
+     * @return
+     */
```
alecgrieser, line 236: I assume this was accidentally left blank. I'm not sure how important it is that internal APIs have Javadocs, but this should probably be removed if not filled in

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBTypedRecordStore.java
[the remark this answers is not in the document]
alecgrieser: Ah, I see. Looking at the way synthetic types interact with the "typed" variety of objects, there does seem to be a bit of mismatch between what the type parameters want and how synthetic types work. I think my problem with leaving this like this is that typed record stores (by the nature of them being "typed") can have partial serializers that reject records of the wrong type, which doesn't play well with the concept of synthetic records in general.

There's part of me that actually thinks the right call is to remove the `FDBTypedRecordStore` abstraction all together (as I don't think it's quite as useful as we'd hoped). Another possibility might be to throw an error here, or require that synthetic record type resolves only happen on `FDBRecordStore`s, which...also might allow some of the `FDBQueriedRecord` stuff to be cleaned up. It also might not, in which case, I guess this is fine, though it does break the abstractions somewhat.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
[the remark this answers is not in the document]
alecgrieser: Okay, I see. I think that's reasonable. I think I'm inclined to say that we do want to collapse these code paths eventually, but if you think that should be deferred to a later time, that's fine.

alecgrieser (changes requested):
Okay, I went through and gave this another pass. I also resolved any comments from the first pass that I think are resolved now, but there are a few things left over from the first round of review.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/RecordQuery.java
```diff
@@ -139,7 +139,7 @@ public List<KeyExpression> getRequiredResults() {
      */
     public void validate(@Nonnull RecordMetaData metaData) {
         for (String recordTypeName : recordTypes) {
-            final RecordType recordType = metaData.getRecordType(recordTypeName);
+            final RecordType recordType = metaData.getIndexableRecordType(recordTypeName);
```
MMcM, line 142: I would somewhat rather introduce a new `getQueryableRecordType` method or some name like that for this case, even if the present definition would be the same as `getIndexableRecordType`, that is, physical or synthetic. This isn't valid because it's indexable, really, but because this PR fixes support for planning it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -783,7 +786,7 @@ private PlanContext getPlanContext(@Nonnull RecordQuery query) {
             if (query.getRecordTypes().isEmpty()) { // ALL_TYPES
                 commonPrimaryKey = RecordMetaData.commonPrimaryKey(metaData.getRecordTypes().values());
             } else {
-                final List<RecordType> recordTypes = query.getRecordTypes().stream().map(metaData::getRecordType).collect(Collectors.toList());
+                final List<RecordType> recordTypes = query.getRecordTypes().stream().map(metaData::getIndexableRecordType).collect(Collectors.toList());
```
MMcM, line 789: `getQueryableRecordType`, again.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
```diff
@@ -453,6 +454,20 @@ public void trimPrimaryKey(List<?> primaryKeys) {
...
+            for (int i = 0; i < primaryKeyComponentPositions.length; i++) {
+                if (primaryKeyComponentPositions[i] >= 0) {
+                    resultSet.set(i);
+                }
+            }
+        }
+        return resultSet;
+    }
+
+
```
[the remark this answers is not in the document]
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBTypedRecordStore.java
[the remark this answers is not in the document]
normen662: I am throwing an error now if called from `FDBTypedRecordStore`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
```diff
@@ -230,6 +230,15 @@ default boolean hasLosslessNormalization() {
...
+    @API(API.Status.INTERNAL)
+    default boolean canBeComputedWithoutMessage() {
+        return false;
```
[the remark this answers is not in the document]
normen662: Added all mentioned overrides.

fdb-record-layer-spatial/src/main/java/com/apple/foundationdb/record/spatial/geophile/GeophilePointWithinDistanceQueryPlan.java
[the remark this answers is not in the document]
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/AvailableFields.java
```diff
@@ -225,33 +246,34 @@ public static AvailableFields intersection(@Nonnull List<AvailableFields> toInte
...
         }
 
         @Nonnull
         public IndexKeyValueToPartialRecord.TupleSource getSource() {
             return source;
         }
 
-        public int getIndex() {
-            return index;
+        public ImmutableIntArray getOrdinalPath() {
```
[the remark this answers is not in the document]
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/ExpressionToTuplePathVisitor.java
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        return new Result(resultMapBuilder.build(), 0);
+    }
+
+    @Nonnull
+    public final Result visitDefault(@Nonnull final KeyExpression keyExpression) {
+        final var currentState = getCurrentState();
+        final var path = currentState.getOrdinalWithParent().toPath();
+        if (path.length() == 1 && skipSet.get(currentState.getOrdinalWithParent().getOrdinal() - startOrdinal + currentState.getNumSkippedMappings())) {
```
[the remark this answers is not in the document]
normen662: Wrote a comment explaining the calculation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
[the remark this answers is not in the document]
normen662: This caused some major rework of the code. I changed the copiers to be cognizant of `null` sub-tuples and their meaning in the context of index entries versus partial records. `Copier`s are now able to _refuse_ to copy, which in turn causes encompassing messages to not be created. This is used in conjunction with the specific info about synthetic record types. As a result, the employed copier's logic is can now handle `null` sub-tuples vs `null`-valued fields when necessary.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/RecordQueryPlannerSubstitutionVisitor.java
normen662: Do you have a suggested name for this property?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -61,6 +63,14 @@
     @Nullable
     public abstract IndexEntry getIndexEntry();
 
+    @Nonnull
+    public Message getRecordAsMessage() {
```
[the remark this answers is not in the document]
normen662: Mike and I discussed this and the solution I came up with seemed to be the best one as it allows me to fake the `M` types without actually fulfilling the contract which is impossible as the type of a synthetic type is dynamic. A lot of stuff in `RecordQueryPlan` land is `M` typed but that's poison for everything that is actually touching `DynamicMessage` code. I agree this needs to change sooner than later.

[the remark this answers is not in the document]
normen662: BTW, I had already attempted pretty much all of your suggestion (including folding synthetic records into `QueryResult`) but this seemed to be the best solution.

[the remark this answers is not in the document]
normen662: If you are wondering how a covering indexes scan can pull off this feat: It doesn't: In `QueryPlanUtils.getCoveringIndexEntryToPartialRecordFunction()` there is a cast to `(M)`. I really didn't want to do that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -61,6 +63,14 @@
...
+        return getRecord();
+    }
+
+    @Nullable
+    public abstract Map<String, FDBStoredRecord<? extends Message>> getConstituents();
```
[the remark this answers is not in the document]
normen662: All of the methods is `FDBQueriedRecord` work like that. Things it doesn't have are `null`-ed out. `stored()` has an index-flavored methods which all return `null`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -251,6 +277,90 @@ public FDBRecordVersion getVersion() {
...
+        @Override
+        public Index getIndex() {
+            return null;
+        }
+
+        @Nullable
+        @Override
+        public IndexEntry getIndexEntry() {
+            return null;
+        }
```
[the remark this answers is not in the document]
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/UnionVisitor.java
normen662: `resolveFetchIndexRecordsFromPlan()` is in the base class. `Union` and `Intersection` are slightly differently but still need to walk the legs and do some (different) logs.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: Added a comment. We also now reject mixed types.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/KeyExpression.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/RecordQuery.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: changed all those new refs to that method to use the new method

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/visitor/RecordQueryPlannerSubstitutionVisitor.java
alecgrieser: I almost feel like `InferrableOnSingleType` or something might make sense. It's a little unfortunate, because the property is something like `Inferrable` (for literals and empty) or "is record type key". It also seems pretty specific to this exact circumstance, but ¯\\\_(ツ)_/¯ I guess that happens

alecgrieser (changes requested):

scottfines (dismissed):

fdb-extensions/src/main/java/com/apple/foundationdb/tuple/TupleHelpers.java
```diff
@@ -159,6 +160,39 @@ public static boolean isPrefix(@Nonnull Tuple t1, @Nonnull Tuple t2) {
         return true;
     }
 
+    @SuppressWarnings("UnstableApiUsage")
+    @Nullable
+    public static Object getForOrdinalPath(@Nonnull Tuple tuple, @Nonnull final ImmutableIntArray ordinalPath) {
```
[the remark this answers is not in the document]
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/ExpressionToTuplePathVisitor.java
```diff
@@ -42,13 +44,24 @@
...
 /**
- * Visitor to find mappings of {@link KeyExpression}s to tuple access paths.
+ * Visitor to find mappings of {@link KeyExpression}s to tuple access paths that are encoded as {@link FieldData}.
+ * <br>
+ * This visitor is exclusively used by {@link AvailableFields} to compute the mappings from a key expression
+ * which is expressed based on a record to an ordinal path (a dewey id) that encodes the access path to the same data
+ * but held in e.g. a primary key of an index.
+ * <br>
+ * Example: A record {@code (a: 10, (b: 30, c: 40), d: 50, x: (y: (e: 60, f: 70), g: 80), h: 90)} can be encoded as
+ * a tuple {@code (10, (30, 40), 50, ((60, 70), 80), h: 90)} which can be addressed by ordinals. For instance,
```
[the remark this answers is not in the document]
normen662: you are, fixed.

alecgrieser (dismissed):
LGTM. We should probably squash this when it gets merged in

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -241,6 +244,22 @@ public RecordType getIndexableRecordType(@Nonnull String name) {
         return recordType;
     }
 
+    /**
+     * Get a record type or synthetic record type by name as used in an index.
```
MMcM, line 248:
```suggestion
     * Get a record type or synthetic record type by name as used in a query.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
```diff
@@ -453,6 +454,19 @@ public void trimPrimaryKey(List<?> primaryKeys) {
         }
     }
 
+    @API(API.Status.INTERNAL)
+    public BitSet getCoveredPrimaryKeyPositions() {
+        final BitSet resultSet = new BitSet();
```
MMcM, line 459:
```suggestion
        final BitSet resultSet = new BitSet(primaryKeyComponentPositions.length);
```

alecgrieser (dismissed):

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
normen662: done

MMcM (approved):

PR #1892, opened 2022-11-01
resolves #1893 add copiers for required fields for covering index scans

### LINKED ISSUE #1893: ensure that partial records are created as proper protobuf messages
With #1879, a bug was introduced that proto messages having required fields did not cause us to properly initialize these messages. This needs to be addressed.

normen662:


alecgrieser (dismissed):
LGTM. In a perfect world, we'd probably also have a test (seeing as this slipped past our current test coverage) and a release note

MMcM (dismissed):

MMcM (approved):

PR #1888, opened 2022-10-27
add more plan kinds to the tunneling logic in PanOrderingKey

normen662:

PR #1894, opened 2022-11-01
Revert #1888

alecgrieser:
This reverts #1888, which could lead to plan instability problems with certain queries that had complex intersections or unions in them. These are, for the most part, preëxisting issues that were being hidden before #1888, though sometimes it would result in errors (not just inefficient plans). We do eventually want something equivalent to #1888, but we'll need to consider how that will happen precisely.


alecgrieser:
This ended up going into #1892

PR #1904, opened 2022-11-09
Resolves #1430: Support non-idempotent target indexes while indexing by index

### LINKED ISSUE #1430: Support non-idempotent target indexes while indexing by index
To support this, we'll probably need: 
1. Have a way to online index cumulative indexes in SNAPSHOT resolution (https://github.com/FoundationDB/fdb-record-layer/issues/1429) 
2. Allow a range to be defined by the source index keys (instead of the record's)

With the combination of these two, a write only index can be defined with this special mode of maintaining the target index iff the generated source index is within an "already built" range.

alecgrieser:
This adds support for building a non-idempotent target index when indexing from a different source index. This works in a manner that is analogous to the way that this operation works for non-idempotent indexes built by a record scan, except that as the range set contains ranges of index entries from the source index, the maintainer needs to be updated to: (1) check the indexing type stamp and (2) modify the range set check to use the index key instead of the primary key. Then there are some updates to the indexers to adjust logic that assumed the target index type would always be idempotent.

I played around with the tests added in this PR, and I've validated that if I inject a few bugs into the code (like checking the wrong value in the range set) that they fail, so I'm fairly confident in their ability to cover the interesting scenarios. The exception there would be `deleteRecordsWhere`, which is not currently validated in any of the indexing tests, though which I believe to be correctly handled.

This resolves #1430.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -42,6 +42,7 @@ import com.apple.foundationdb.record.EvaluationContext;
 import com.apple.foundationdb.record.ExecuteProperties;
 import com.apple.foundationdb.record.ExecuteState;
 import com.apple.foundationdb.record.FunctionNames;
+import com.apple.foundationdb.record.IndexBuildProto;
 import com.apple.foundationdb.record.IndexEntry;
 import com.apple.foundationdb.record.IndexScanType;
 import com.apple.foundationdb.record.IndexState;
@@ -605,22 +606,12 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         for (Index index : indexes) {
             final IndexMaintainer maintainer = getIndexMaintainer(index);
             final CompletableFuture<Void> future;
-            if (!maintainer.isIdempotent() && isIndexWriteOnly(index)) {
-                // In this case, the index is still being built, so we are not
-                // going to update the record unless the rebuild job has already
-                // gotten to this range.
-                final Tuple primaryKey = newRecord == null ? oldRecord.getPrimaryKey() : newRecord.getPrimaryKey();
-                future = maintainer.addedRangeWithKey(primaryKey)
-                        .thenCompose(present -> {
-                            if (present) {
-                                return maintainer.update(oldRecord, newRecord);
-                            } else {
-                                return AsyncUtil.DONE;
-                            }
-                        });
-                if (!MoreAsyncUtil.isCompletedNormally(future)) {
-                    futures.add(future);
-                }
+            if (isIndexWriteOnly(index)) {
+                // In this case, the index is still being built. For some index
+                // types, the index update needs to check whether indexing
+                // process has already built the relevant ranges, and it
+                // may adjust the way the index is built in response.
+                future = maintainer.updateWhileWriteOnly(oldRecord, newRecord);
             } else {
                 future = maintainer.update(oldRecord, newRecord);
             }
@@ -3497,6 +3488,32 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
      */
     public CompletableFuture<IndexBuildState> getIndexBuildStateAsync(Index index) {
         return IndexBuildState.loadIndexBuildStateAsync(this, index);
+    }
+
+    @API(API.Status.INTERNAL)
+    @Nonnull
+    public CompletableFuture<IndexBuildProto.IndexBuildIndexingStamp> loadIndexBuildStampAsync(Index index) {
+        byte[] stampKey = IndexingBase.indexBuildTypeSubspace(this, index).pack();
+        return ensureContextActive().get(stampKey).thenApply(serializedStamp -> {
+            if (serializedStamp == null) {
+                return null;
+            }
+            try {
+                return IndexBuildProto.IndexBuildIndexingStamp.parseFrom(serializedStamp);
+            } catch (InvalidProtocolBufferException ex) {
+                RecordCoreException protoEx = new RecordCoreException("invalid indexing type stamp",
+                        LogMessageKeys.INDEX_NAME, index.getName(),
+                        LogMessageKeys.ACTUAL, ByteArrayUtil2.loggable(serializedStamp));
+                protoEx.initCause(ex);
+                throw protoEx;
+            }
+        });
+    }
+
+    @API(API.Status.INTERNAL)
+    public void saveIndexBuildStamp(Index index, IndexBuildProto.IndexBuildIndexingStamp stamp) {
+        byte[] stampKey = IndexingBase.indexBuildTypeSubspace(this, index).pack();
+        ensureContextActive().set(stampKey, stamp.toByteArray());
     }
 
     // Remove any indexes that do not match the filter.
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -3492,9 +3492,17 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         return IndexBuildState.loadIndexBuildStateAsync(this, index);
     }
 
+    /**
+     * Load the indexing type stamp for an index. This stamp contains information about the kind of
+     * index build being used to construct a new index. This method is {@link API.Status#INTERNAL}.
+     *
+     * @param index the index being built
+     * @return the indexing type stamp for the index's current build
+     * @see #saveIndexingTypeStamp(Index, IndexBuildProto.IndexBuildIndexingStamp)
+     */
     @API(API.Status.INTERNAL)
     @Nonnull
-    public CompletableFuture<IndexBuildProto.IndexBuildIndexingStamp> loadIndexBuildStampAsync(Index index) {
+    public CompletableFuture<IndexBuildProto.IndexBuildIndexingStamp> loadIndexingTypeStampAsync(Index index) {
         byte[] stampKey = IndexingBase.indexBuildTypeSubspace(this, index).pack();
         return ensureContextActive().get(stampKey).thenApply(serializedStamp -> {
             if (serializedStamp == null) {
@@ -3512,8 +3520,17 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         });
     }
 
+    /**
+     * Update the indexing type stamp for the given index. This is used by the {@link OnlineIndexer}
+     * to document what kind of indexing procedure is being used to build the given index. This method
+     * is {@link API.Status#INTERNAL}.
+     *
+     * @param index the index being built
+     * @param stamp the new value of the index's indexing type stamp
+     * @see #loadIndexingTypeStampAsync(Index)
+     */
     @API(API.Status.INTERNAL)
-    public void saveIndexBuildStamp(Index index, IndexBuildProto.IndexBuildIndexingStamp stamp) {
+    public void saveIndexingTypeStamp(Index index, IndexBuildProto.IndexBuildIndexingStamp stamp) {
         byte[] stampKey = IndexingBase.indexBuildTypeSubspace(this, index).pack();
         ensureContextActive().set(stampKey, stamp.toByteArray());
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -190,11 +190,13 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     public static final int CACHEABLE_STATE_FORMAT_VERSION = 7;
     // 8 - add custom fields to store header
     public static final int HEADER_USER_FIELDS_FORMAT_VERSION = 8;
-
+    // 9 - add READABLE_UNIQUE_PENDING index state
     public static final int READABLE_UNIQUE_PENDING_FORMAT_VERSION = 9;
+    // 10 - check index build type during update
+    public static final int CHECK_INDEX_BUILD_TYPE_DURING_UPDATE_FORMAT_VERSION = 10;
 
     // The current code can read and write up to the format version below
-    public static final int MAX_SUPPORTED_FORMAT_VERSION = READABLE_UNIQUE_PENDING_FORMAT_VERSION;
+    public static final int MAX_SUPPORTED_FORMAT_VERSION = CHECK_INDEX_BUILD_TYPE_DURING_UPDATE_FORMAT_VERSION;
 
     // By default, record stores attempt to upgrade to this version
     // NOTE: Updating this can break certain users during upgrades.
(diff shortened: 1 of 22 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -401,35 +400,32 @@ public void enforceStampOverwrite() {
     @SuppressWarnings("PMD.CloseResource")
     private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boolean continuedBuild) {
         // continuedBuild is set if this session isn't a continuation of a previous indexing
-        Transaction transaction = store.getContext().ensureActive();
         IndexBuildProto.IndexBuildIndexingStamp indexingTypeStamp = getIndexingTypeStamp(store);
 
-        return forEachTargetIndex(index -> setIndexingTypeOrThrow(store, continuedBuild, transaction, index, indexingTypeStamp));
```
jjezra, line 407: Is there a way to change all the indexes' typestamp with a single transaction? (to avoid changing few indexes but not all..)

alecgrieser: Yes, the transactionality of the operation here hasn't changed. The only thing that changed was that the `Transaction` object is now no longer passed separately from the store, which I made because that way, it is clearer that the store and transaction are linked (i.e., that someone wasn't doing something sneaky like passing a store that was created with a _different_ transaction)

alecgrieser:
One thing that hasn't been addressed by this PR (yet): how should this be rolled out? The main concern here with roll out is that if someone starts an index-from-index build on a non-idempotent index, but there are still instances accessing the same record store that are running an older version of the code, then the record update won't check the correct values in the range set, which can result in index corruption. Note that the data written by builds are unchanged, so any ongoing index build can continue

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -251,6 +252,96 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable final FDBInd
...
+                    return updateWriteOnlyByIndex(sourceIndex, oldRecord, newRecord);
+                default:
+                    throw new RecordCoreException("unable to update write-only index with current type stamp")
+                            .addLogInfo("stamp", stamp);
+            }
+        });
+    }
+
+    private <M extends Message> CompletableFuture<Void> updateWriteOnlyByRecords(@Nullable final FDBIndexableRecord<M> oldRecord, @Nullable final FDBIndexableRecord<M> newRecord) {
+        // Check if the record has been built be checking its primary key in the range set. Update the index
```
jjezra, line 285: "be" = "by"?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -251,6 +252,96 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable final FDBInd
...
+                // built, and update the index if so
+                return addedRangeWithKey(oldEntryKey).thenCompose(inRange ->
+                        inRange ? update(oldRecord, newRecord) : AsyncUtil.DONE);
+            } else {
+                // The old and new record use different keys in the source index. Check each one individually,
+                // and then simulate deleting the old record (if needed) and adding the new record (if needed)
+                return addedRangeWithKey(oldEntryKey)
+                        .thenCompose(oldInRange -> oldInRange ? update(oldRecord, null) : AsyncUtil.DONE)
+                        .thenCompose(ignore -> addedRangeWithKey(newEntryKey))
+                        .thenCompose(newInRange -> newInRange ? update(null, newRecord) : AsyncUtil.DONE);
```
jjezra, line 308: Would `allOf` make sense here?

jjezra (changes requested):
Nice!

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -115,12 +115,30 @@ public RecordCursor<IndexEntry> scan(@Nonnull IndexScanBounds scanBounds,
...
+     * @return a future that is complete when the index update is done
      */
     @Nonnull
     public abstract <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexableRecord<M> oldRecord,
                                                                        @Nullable FDBIndexableRecord<M> newRecord);
 
+    /**
+     * Update the associated index for a changed record while the index is in
+     * {@link com.apple.foundationdb.record.IndexState#WRITE_ONLY} mode. For most indexes, this should do the
+     * same thing that a normal update does, but if the index is not {@linkplain #isIdempotent() idempotent},
```
ScottDugas, line 127: I think, at some point, the existence of `updateWhileWriteOnly` could allow us to remove `isIdempontent`, although we may want to expand the documentation to more clearly describe the behavior that an implementor needs to handle for both `update` and `updateWhileWriteOnly`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -115,12 +115,30 @@ public RecordCursor<IndexEntry> scan(@Nonnull IndexScanBounds scanBounds,
...
+     * associated with the index build to check to see if the record has already been indexed, and then decide
+     * to update (or not update) the index as appropriate.
+     *
+     * @param oldRecord the previous stored record or <code>null</code> if a new record is being created
+     * @param newRecord the new record or <code>null</code> if an old record is being deleted
+     * @param <M> type of message
+     * @return a future that is complete when the index update is done
+     */
+    @Nonnull
+    public abstract <M extends Message> CompletableFuture<Void> updateWhileWriteOnly(@Nullable FDBIndexableRecord<M> oldRecord,
```
ScottDugas, line 139: I can't recall what the stance is, but this is a breaking change for implementors of `IndexMaintainer`. Should this have a default implementation that is what we previously used to do in the record store, and then, comment that the default implementation will be removed in the future?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3499,6 +3492,49 @@ public CompletableFuture<IndexBuildState> getIndexBuildStateAsync(Index index) {
...
+     * index build being used to construct a new index. This method is {@link API.Status#INTERNAL}.
+     *
+     * @param index the index being built
+     * @return the indexing type stamp for the index's current build
+     * @see #saveIndexingTypeStamp(Index, IndexBuildProto.IndexBuildIndexingStamp)
+     */
+    @API(API.Status.INTERNAL)
+    @Nonnull
+    public CompletableFuture<IndexBuildProto.IndexBuildIndexingStamp> loadIndexingTypeStampAsync(Index index) {
+        byte[] stampKey = IndexingBase.indexBuildTypeSubspace(this, index).pack();
```
ScottDugas, line 3506: It seems a bit weird that `IndexingBase` continues to manage what the subspace is, but not how to write to it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -251,6 +252,100 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable final FDBInd
...
+        if (record == null) {
+            return null;
+        }
+        List<IndexEntry> entries = maintainer.filteredIndexEntries(record);
+        if (entries == null || entries.isEmpty()) {
+            // If there is an IndexMaintenanceFilter on the index, this can return null/an empty list even if the
+            // index key expression always returns a single value. In this case, the record is excluded from the index
+            return null;
+        } else if (entries.size() != 1) {
+            throw new RecordCoreException("index produced incorrect number of entries for use as source index");
```
ScottDugas, line 342: Does this restriction apply for idempotent indexes, or is this a new restriction for non-idempotent indexes?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingByIndex.java
```diff
@@ -243,12 +237,22 @@ private CompletableFuture<Tuple> rebuildRangeOnly(@Nonnull FDBRecordStore store,
...
+
+    private void validateIdempotenceIfNecessary(@Nonnull FDBRecordStore store, @Nonnull IndexMaintainer maintainer) {
+        // idempotence - Non-idempotent indexes can only be built from a source if the store format version supports it
+        // Prior to this format version, record updates to non-idempotent indexes would check to see if the record
+        // had been built already by looking for its primary key in the range set, which is incorrect for most source
+        // indexes. After this format version, we are assured that all updates will check the index type first and
+        // respond appropriately
+        if (store.getFormatVersion() < FDBRecordStore.CHECK_INDEX_BUILD_TYPE_DURING_UPDATE_FORMAT_VERSION) {
+            validateOrThrowEx(maintainer.isIdempotent(), "target index is not idempotent");
+        }
```
jjezra, line 256: Would 
```
validateOrThrowEx(maintainer.isIdempotent() || 
(store.getFormatVersion() >= FDBRecordStore.CHECK_INDEX_BUILD_TYPE_DURING_UPDATE_FORMAT_VERSION), 
"target index is not idempotent")
```
be a little more elegant? 
(not a big deal..)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildGroupedCountIndexTest.java
```diff
@@ -0,0 +1,504 @@
...
+
+        try (OnlineIndexer indexer = newIndexerBuilder()
+                .setIndex(COUNT_BY_NUM_VALUE_2)
+                .setIndexingPolicy(policyBuilder.build())
+                .setLimit(10)
+                .setMaxRetries(Integer.MAX_VALUE)
+                .setRecordsPerSecond(OnlineIndexer.DEFAULT_RECORDS_PER_SECOND * 100)
+                .build()) {
+            CompletableFuture<?> buildFuture = indexer.buildIndexAsync(true);
+            recordsAfter = updater.update(recordsBefore, otherRecordsBefore);
```
ScottDugas, line 206: There's definitely a risk that these won't really end up testing concurrent problems, at least with the numbers in the test, but I don't have a good solution, other than make a very large test that has a ton of records, takes hours to build, and randomly does all sorts of mutations while doing so, and then asserts that the result is correct. But, the record layer doesn't really have the infrastructure to do that kind of testing.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/TerribleIndexMaintainer.java
```diff
@@ -94,6 +94,12 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexable
         }
     }
 
+    @Nonnull
+    @Override
+    public <M extends Message> CompletableFuture<Void> updateWhileWriteOnly(@Nullable final FDBIndexableRecord<M> oldRecord, @Nullable final FDBIndexableRecord<M> newRecord) {
+        return AsyncUtil.DONE;
```
ScottDugas, line 100: Should this just call `update`? or return a failed exception

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -162,29 +169,64 @@ public void testIndexFromIndexContinuation() {
...
         openSimpleMetaData(hook);
         buildSrcIndex(srcIndex);
 
         openSimpleMetaData(hook);
-        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
-                .setDatabase(fdb).setMetaData(metaData).addTargetIndex(tgtIndex).setSubspace(subspace)
+        try (OnlineIndexer indexBuilder = newIndexerBuilder()
+                .addTargetIndex(tgtIndex)
+                // Set a format version on the store that does not allow index-from-index builds
+                // Because the store already has this format version, though it should be allowed
```
jjezra, line 228: I'm not sure that I understand the second line of the comment. 
Did you mean to mention that the indexing fallback to "by records" scan? 
(fallsback? fallbacks? falls back?)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -605,22 +608,12 @@ private <M extends Message> void updateSecondaryIndexes(@Nullable final FDBIndex
...
-                            if (present) {
-                                return maintainer.update(oldRecord, newRecord);
-                            } else {
-                                return AsyncUtil.DONE;
-                            }
-                        });
-                if (!MoreAsyncUtil.isCompletedNormally(future)) {
-                    futures.add(future);
-                }
+            if (isIndexWriteOnly(index)) {
```
ScottDugas, line 611: There are other calls to `update`, do any of those need to be updated to do this check? They weren't doing the previous check, so they may have already been broken for non-idempotent use cases.

jjezra (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
alecgrieser: This is not a new restriction: there was already a check in the indexer to make sure the index did not create duplicates. However, it is more important for non-idempotent target indexes.

Along those lines, there are actually two codepaths, a transactional build and a non-transactional build, and it turns out this check was missed in the transactional case. I've added the check in the the other case, which also added a few other missed checks.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/TerribleIndexMaintainer.java
alecgrieser: I had this return `DONE` rather than throw an error because this index maintainer always returns `false` to `isIdempotent`, so this is behavior preserving. I didn't try throwing an error, but IIRC, this maintainer is used in some somewhat delicate tests, so it could cause some tests to break, I think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
alecgrieser: Yeah, that seems to be the case. The other calls are:

1. Within the `TextIndexMaintainer`, which will only get called on (idempotent) text indexes, so we're fine
2. Within the indexer, which wants to update the index maintainer unconditionally
3. Within `FDBRecordStore::runSyntheticMaintainers`, which I'm less certain about, but that already was in a bit of a suspect state

So nothing seems to be updated in response to this, but maybe we should revisit this after this is in

alecgrieser:
There are a few cleanup/comment/refactoring things that I will try and get into a follow up, but I've focused on addressing things that might affect correctness in this PR right now, but I don't want to delay this too much given that without it, affected index builds could be delayed significantly

MMcM (approved):

PR #1908, opened 2022-11-10
Fixes for planning aggregate index COUNT and COUNT_NOT_NULL in Cascades

hatyo:
This fixes the way we construct a matching `RelationalExpression` out of a an aggregate key expression containing e.g. an empty grouped and / or empty grouping.

It also refactors the construction of the `RelationalExpression` such that we avoid using field names and rely on their ordinals.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
@@ -31,8 +31,8 @@ import com.apple.foundationdb.record.query.plan.cascades.expressions.SelectExpre
 import com.apple.foundationdb.record.query.plan.cascades.predicates.ValueComparisonRangePredicate;
 import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
 import com.apple.foundationdb.record.query.plan.cascades.typing.TypeRepository;
-import com.apple.foundationdb.record.query.plan.cascades.values.AggregateValue;
 import com.apple.foundationdb.record.query.plan.cascades.values.CountValue;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
```diff
@@ -183,34 +178,32 @@ private Quantifier constructSelectWhere(@Nonnull final Quantifier.ForEach baseQu
...
+        // we have a single base quantifier which is _always_ fixed at position-1 in the underlying select-where.
+        final var baseQuantifierReference = FieldValue.ofOrdinalNumber(selectWhereQun.getFlowedObjectValue(), 1);
+        final var arguments = groupedValue.map(value -> {
+            if (value instanceof EmptyValue) {
+                return RecordConstructorValue.ofColumns(List.of());
+            } else {
+                return FieldValue.ofFields(selectWhereQun.getFlowedObjectValue(), baseQuantifierReference.getFieldPath().withSuffix(((FieldValue)value).getFieldPath()));
+            }
+        }).orElse(RecordConstructorValue.ofColumns(List.of()));
+        final var aggregateValue = (Value)aggregateMap.get().get(index.getType()).encapsulate(TypeRepository.newBuilder(), List.of(arguments));
```
normen662, line 195: Don't want to start this debate but all the `List.of()` should be `ImmutableList.of()`

normen662:
contained in https://github.com/FoundationDB/fdb-record-layer/pull/1909

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
hatyo: Agreed, all should be `ImmutableList`.

PR #1863, opened 2022-10-12
Resolves #1862: Lucene search with highlighting the terms

tian-yizuo:

PR #1911, opened 2022-11-14
Resolves #1862: Lucene search with highlighting the terms

### LINKED ISSUE #1862: Lucene search with highlighting the terms

MMcM:
Follow-up to #1863.

Instead of introducing a new way of intercepting the loading of a record from its index entry, add a new plan operator to do highlighting. This makes changes to `-core` (hopefully) uncontroversial.

Other than that, the changes are intentionally minimal from the earlier PR.

Note that Union plans are different: there is no case of `Union` of `Covering` with deferred `Fetch`. The `Highlight` prevents this. It should be possible to defer the `Highlight` to after the `Fetch` and so allow a union of index entries inside. However, that requires planner support. Moreover, all these plans, including the one in the earlier PR, have a problem when results from two differently shaped searches are combined. The desired outcome is that the record's fields highlight terms from either side. But the `Union` throws one side away on the assumption that they are equivalent when the primary-key matches. Since the highlighting params are currently stored in the index entry, the `Union` would need to produce a merged highlighting params to be carried on with its output's index entry and used by `Highlight`.


scottfines (approved):

scottfines:
Thanks for taking care of this!

PR #1913, opened 2022-11-15
Fix RemoveSortRule

hatyo:
Make sure to rebase `sortValues` before attempting to check order satisfiability with plan partition.


hatyo:
closing this PR after clarifying how `LogicalSortExpression` is supposed to work. Bottom line is, the `Value`s in the `LSE` operator are to be bound to `CURRENT` quantifier, i.e. they are _not_ deriving from any specific underlying operator, they are just imposing a restriction on the output order.

PR #1914, opened 2022-11-15
Some more Lucene highlighting fixes

MMcM:
* More `Query` classes for term extraction.
* String index out of bounds when finding matched terms.

Note that there are still problems with `searchAllMaybeHighlight` regarding `highlightedTextConnector` if an ngram tokenizer is used. The same word will be seen several times from the min ngram up to its length, confusing the `currentPos` count. The result of this is stray ellipses. It seems better to get this in without crashes than wait for @tian-yizuo to be available to review anything more substantial.


scottfines:
LGTM, thanks!

Issue #1921, opened 2022-11-17
Consider tighter protobuf performance by removing array copies


jleach4:
Move backward work that is in UnsafeByteOperations in Protobuf 3 to record layer in attempt to reduce array copies.

alecgrieser:
Closed by #1922

PR #1924, opened 2022-11-18
Resolves #1923: Avoid copying continuation byte arrays by using ByteString internally

### LINKED ISSUE #1923: Avoid copying continuation byte arrays by using ByteString internally
The `RecordCursorContinuation` class has a `toBytes()` method which returns a `byte[]` and a `toByteString()` method which returns a `ByteString`. Internally, there are quite a few cases where we construct the byte representation by constructing a protobuf, often with child continuations that serialize their continuation to a `byte[]`, which we then wrap in a `ByteString` and hand to the protobuf serializer.

We should be able to skip that intermediate state by using `toByteString` on the continuation object. Some care does need to be taken, though, as some implementations of `toByteString()` are (more-or-less) `ByteString.copy(toBytes())`, which could induce additional copying. But assuming we can tidy that up, this should remove some byte copies and object creation.

alecgrieser:
This updates the logic in the `RecordCursorContinuation` objects to avoid serializing a continuation all the way to a `byte[]` if it is not necessary. For example, when constructing protobuf objects that back continuations, child continuations can be constructed by calling `.toByteString()` on the continuation and passing that to the protobuf builder instead of serializing all the to a `byte[]` and then wrapping it back in a `ByteString`. This also makes sure that (as much as possible) the cursor implementations are more efficient when constructing a `ByteString`. For example, for the proto-based continuations, it should be able to return `.toByteString()` on the protobuf object instead of serializing the _protobuf_ all the way to a `byte[]` and then wrapping it back up, which some implementations did previously.

Adopters who need a `ByteString` rather than a `byte[]` (for example, to pack the continuation into a protobuf) are also encouraged to switch to the newer method.

This resolves #1923. This also follows up from #1921, which introduced the zero copy `ByteString` to avoid copying immutable arrays.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -23,7 +23,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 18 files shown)


MMcM (approved):

MMcM (changes requested):
Changes look good, but some test failures look to be actual regressions.

MMcM (approved):

PR #1932, opened 2022-11-23
changes needed for inserts-updates-deletes

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/InsertExpression.java
```diff
@@ -192,12 +189,7 @@ private static Type.Record computeResultType(@Nonnull final Type inType, @Nonnul
     }
 
     @Nonnull
-    private static Value makeComputationValue(@Nonnull final Quantifier inner, @Nonnull final Type targetType) {
```
hatyo, line 195: The now not-used method `computeResultType` is not used anymore, this is causing PMD check to fail.

hatyo (commented):
LGTM, but we must fix the PMD failure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/InsertExpression.java
hatyo: ... pushed a commit to address that.

PR #1938, opened 2022-12-06
Make random UUID a compliant type name in CorrelationIdentifier.uniqueID

arnaud-lacurie:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CorrelationIdentifier.java
```diff
@@ -89,7 +90,7 @@ public static CorrelationIdentifier uniqueID(@Nonnull final Class<?> clazz, @Non
         final CorrelationIdentifier id =
                 Debugger.getIndexOptional(clazz)
                         .map(i -> CorrelationIdentifier.of(prefix + i))
-                        .orElseGet(() -> new CorrelationIdentifier(UUID.randomUUID().toString()));
+                        .orElseGet(() -> new CorrelationIdentifier(Type.uniqueCompliantTypeName()));
```
MMcM, line 93: Doesn't this just want half of that method? This id needs to be protobuf-friendly but isn't a `__type__`, is it?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/ProtoUtils.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * MapUtils.java
```
MMcM, line 2:
```suggestion
 * ProtoUtils.java
```

PR #1944, opened 2022-12-09
IN function in RelOp values

### LINKED ISSUE #1955: Support for IN predicates in Cascades planner
Provide support for queries like: 
```
SELECT * FROM T1 WHERE T1.A IN (1, 2, 3, 4);
```

g31pranjal:
fixes #1955


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -149,15 +150,19 @@ public Optional<QueryPredicate> toQueryPredicate(@Nonnull final CorrelationIdent
                 // it seems this is a constant expression, try to evaluate it.
                 return tryBoxSelfAsConstantPredicate();
             }
-        } else if (childrenCount == 2) {
+        }
+        Verify.verify(childrenCount == 2);
```
hatyo, line 154: The contract of `toQueryPredicate` is best-effort to transform `this` value into a `QueryPredicate`, this is reflected by returning an _Optional_ object, so we can propagate failure to find a mapping by returning `Optional.empty()`.

Theoretically, a component on top, be it a plan generator, could choose to generate a different `Value`, however now that we add a `Verify` the contract has been altered.

Therefore, I suggest to return `Optional.empty()` if `childrenCount != 2`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -149,15 +150,19 @@ public Optional<QueryPredicate> toQueryPredicate(@Nonnull final CorrelationIdent
...
+        final var leftChild = it.next();
+        final var rightChild = it.next();
+        if (comparisonType == Comparisons.Type.IN) {
+            final var values = Objects.requireNonNull(((LiteralValue<?>) rightChild).getLiteralValue());
+            return Optional.of(new ValuePredicate(leftChild, new Comparisons.ListComparison(comparisonType, ((List<?>)values))));
```
hatyo, line 159: In the spirit of the comment above, I think we should do similar graceful processing here w.r.t. casts.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -149,15 +150,19 @@ public Optional<QueryPredicate> toQueryPredicate(@Nonnull final CorrelationIdent
                 // it seems this is a constant expression, try to evaluate it.
                 return tryBoxSelfAsConstantPredicate();
             }
-        } else if (childrenCount == 2) {
+        }
+        Verify.verify(childrenCount == 2);
+        final var leftChild = it.next();
+        final var rightChild = it.next();
+        if (comparisonType == Comparisons.Type.IN) {
+            final var values = Objects.requireNonNull(((LiteralValue<?>) rightChild).getLiteralValue());
```
hatyo, line 158: We miss an important chance of optimisation here.

Say we have the following `in` predicate: `(3+4) in (1, 1+1, 3, 5, 5+2)`, we could pre-evaluate this to `true` which could cascade other execution optimisations.

E.g. if we can prove that the predicate is always `false` we can prune an entire execution.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -290,8 +304,6 @@ private static Value encapsulate(@Nonnull TypeRepository typeRepository, @Nonnul
                         return physicalOperator.eval(it.next(), it.next());
                     });
         }
-        // TODO: we should also allow encapsulating variadic comparison functions such as IN
```
hatyo, line 293: It is nice to see some progress 🙂

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -270,9 +273,20 @@ private static Value encapsulate(@Nonnull TypeRepository typeRepository, @Nonnul
                         Verify.verify(Iterables.size(objects) == 1);
                         return physicalOperator.eval(objects.iterator().next());
                     });
+        }
+        Verify.verify(arguments.size() == 2);
+        final Typed arg1 = arguments.get(1);
+        final Type res1 = arg1.getResultType();
+        if (comparisonType == Comparisons.Type.IN) {
+            SemanticException.check(res1.getTypeCode() == Type.TypeCode.ARRAY, SemanticException.ErrorCode.COMPARAND_TO_COMPARISON_IS_OF_COMPLEX_TYPE);
```
hatyo, line 281: The error code doesn't seem to fit with the error message right?

Maybe you want to check two things for this to work:
1. that `res1` is indeed an array (throw an error like type-mismatch if otherwise).
2. the `elementType` of the array is primitive (throw an error with `SemanticException.ErrorCode.COMPARAND_TO_COMPARISON_IS_OF_COMPLEX_TYPE` error code if otherwise).

hatyo (changes requested):
Nice work @g31pranjal ! 

I noticed the PR misses some unit tests, could you please add tests for `IN` predicate to the file `BooleanValueTest`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: That's right! while we cannot evaluate it completely in the compile-time (because we do not have a physical implementation of IN function) we can actually as an optimization evaluate the left and right sides of it. so that the example you demonstrated reduces to `7 in (1, 2, 3, 5, 7)`. Adding that with other changes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: The check for the `elementType` being one of the primitive type (boolean, byteArray, double, Float, Long, Int, String, Enum) and that all the elements are of the same type happens in the `ListComparison`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: changed the error type for 1

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -150,39 +152,57 @@ public Optional<QueryPredicate> toQueryPredicate(@Nonnull final CorrelationIdent
...
+                final var literalValue = ((LiteralValue<?>) rightChild).getLiteralValue();
+                if (!(literalValue instanceof List)) {
+                    return Optional.empty();
+                }
+                final var evaluatedValues = new ArrayList<>();
+                ((List<?>) literalValue).forEach(value -> {
+                    final var operand = compileTimeEvalFn.apply((Value) value);
+                    if (operand != null) {
+                        evaluatedValues.add(operand);
+                    }
```
hatyo, line 171: I don't think this is correct, I think if `null` (unknown) exists as an operand in an `in-list` then the result is `null`. i.e.

```
null in (1, 2, null) --> null
2 in (1,2,null) -> null
(null in (1,2,3, null)) is null --> true
```

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
MMcM:
```
2 in (1,2,null) -> true
3 in (1,2,null) -> null
```

hatyo: I stand corrected, thanks!

Since the `IN` in a shorthand for disjunctions, and in Kleene logic `true or null` is `null`:

```
2 in (1, 2, null) <=> (2 = 1 OR 2 = 2 OR 2 = null)
                  <=> (false OR true OR null) (execution is free to shortcut execution at this point I think,...)
                  <=> (true OR null) 
                  <=> (true)
```

g31pranjal:
We now have a physical function that is capable of short-circuiting the IN expression for cases where the left operand is a Literal constant.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -440,6 +450,32 @@ private static Value encapsulate(@Nonnull TypeRepository.Builder typeRepositoryB
...
+        }
+
+        public static Object evaluate(Object left, Object right) {
+            Verify.verify(right instanceof List);
+            if (((List<?>) right).stream().anyMatch(object -> object != null && object.equals(left))) {
+                return true;
+            } else if (((List<?>) right).stream().anyMatch(Objects::isNull)) {
+                return null;
+            } else {
+                return false;
```
hatyo, line 474: This is subtle (in a good way); at first glance I thought it is somewhat incorrect, but it seems to be actually correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -440,6 +450,32 @@ private static Value encapsulate(@Nonnull TypeRepository.Builder typeRepositoryB
...
+        public InFn() {
+            super("in",
+                    List.of(new Type.Any(), new Type.Array()), InFn::encapsulate);
+        }
+
+        private static Value encapsulate(@Nonnull TypeRepository.Builder typeRepositoryBuilder, @Nonnull BuiltInFunction<Value> builtInFunction, @Nonnull final List<Typed> arguments) {
+            return RelOpValue.encapsulate(typeRepositoryBuilder.build(), builtInFunction.getFunctionName(), Comparisons.Type.IN, arguments);
+        }
+
+        public static Object evaluate(Object left, Object right) {
```
hatyo, line 467:
```suggestion
        @Nullable
        public static Object evaluate(@Nullable final Object left, @Nonnull final Object right) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/BooleanValueTest.java
```diff
@@ -548,13 +572,39 @@ public Stream<? extends Arguments> provideArguments(final ExtensionContext conte
...
                     Arguments.of(List.of(INT_2, INT_NULL), new RelOpValue.GteFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(INT_NULL, INT_NULL), new RelOpValue.GteFn(), ConstantPredicate.NULL),
-
                     Arguments.of(List.of(BOOL_NULL, BOOL_NULL), new RelOpValue.EqualsFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(BOOL_NULL, BOOL_FALSE), new RelOpValue.EqualsFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(BOOL_NULL, BOOL_TRUE), new RelOpValue.EqualsFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(BOOL_NULL, BOOL_NULL), new RelOpValue.NotEqualsFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(BOOL_NULL, BOOL_FALSE), new RelOpValue.NotEqualsFn(), ConstantPredicate.NULL),
                     Arguments.of(List.of(BOOL_NULL, BOOL_TRUE), new RelOpValue.NotEqualsFn(), ConstantPredicate.NULL),
+                    /*IN*/
```
hatyo, line 581: Nice! 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -1339,8 +1339,7 @@ public static class ListComparison implements Comparison {
         @Nullable
         private final Descriptors.FieldDescriptor.JavaType javaType;
 
-        @SuppressWarnings({"rawtypes", "unchecked"})
-        public ListComparison(@Nonnull Type type, @Nonnull List comparand) {
+        public ListComparison(@Nonnull Type type, @Nonnull List<?> comparand, boolean ignoreNulls) {
```
hatyo, line 1342: I probably missed it but is there a place where we actually set `ignoreNulls` to `true`?

hatyo:
Great work @g31pranjal!  the subsequent commits makes this PR much better IMHO 👍 .

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: :)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: DONE

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
g31pranjal: no we don't. we just need it to validate the List correctly

hatyo: could you please revert since this is a `STABLE` API change which is not allowed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -1349,22 +1348,25 @@ public ListComparison(@Nonnull Type type, @Nonnull List comparand) {
                 case IN:
                     break;
                 default:
-                    throw new RecordCoreException("ListComparison only supports EQUALS, NOT_EQUALS and STARTS_WITH");
+                    throw new RecordCoreException("ListComparison only supports EQUALS, NOT_EQUALS, STARTS_WITH and IN");
             }
-            if (comparand == null || (this.type == Type.IN && comparand.stream().anyMatch(o -> o == null))) {
-                throw new NullPointerException("List comparand is null, or contains null");
+            if (!ignoreNulls && this.type == Type.IN && comparand.stream().anyMatch(Objects::isNull)) {
```
normen662, line 1353: I am not sure this is right. If we rewrite into an in join and we encounter a a `NULL` in the in list, that is currently not correctly treated (a non-null return in a method would then return `NULL`).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -672,7 +709,16 @@ private enum BinaryPhysicalOperator {
         GTE_UL(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.LONG, (l, r) -> null),
         GTE_UF(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.FLOAT, (l, r) -> null),
         GTE_UD(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.DOUBLE, (l, r) -> null),
-        GTE_US(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.STRING, (l, r) -> null);
+        GTE_US(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.STRING, (l, r) -> null),
+
+        IN_I(Comparisons.Type.IN, Type.TypeCode.INT, Type.TypeCode.ARRAY, InFn::evaluate),
```
normen662, line 714: Who is promoting the input to the correct common type? I.e. `(1, 1.5, 3)` should be promoted to `(1.0, 1.5, 3.0)`.

normen662 (changes requested):
Hi, I think we should make `IN`s separate from `RelOps`. That will give us more freedom further down the line. A relop value should only compare two primitives and not a primitive and an array (for instance relapse can be inverted `x > y` is the same as `y < x` but `IN` cannot.

MMcM:
> Hi, I think we should make `IN`s separate from `RelOps`.

How close are we to just doing it right from the start?

```
x IN (2,3)
```

```
(x) IN ((2),(3))
(x,y) IN ((2,20),(3,30))
```

```
(x) = ANY (VALUES (2),(3))
(x,y) = ANY (VALUES (2,20),(3,30))
(x,y) < ANY (VALUES (2,20),(3,30))
```

```
EXISTS (SELECT * FROM (VALUES (2),(3)) AS _v(_v1) WHERE (x) = (_v1))
EXISTS (SELECT * FROM (VALUES (2,20),(3,30)) AS _v(_v1,_v2) WHERE (x,y) < (_v1,_v2))
```

Is the problem recovering the simple `List.contains` behavior when no indexes are present from that most general form? That seems like the most motivating reason to perhaps model it like is done here for now.

normen662:
> Hi, I think we should make INs separate from RelOps.

This PR aims to express an `IN` as a `Value` which I think is right as subsequent transformations can do all the things with it that are potentially useful. I do not think, however, it is right, to make this `InValue` the same as a `RelOpValue`. That's the extent of my earlier comment.

I think the representation of `IN` in the more generic forms as a `Value` is useful as it may be picked up by a transformation and expanded into a more data flow-oriented representation. The `Value` can drive such an expansion that may later turn into a proper sub plan. The IN-to-JOIN related transformations to that, except that they are currently driven by a `Comparison` inside of a `ValuePredicate` in `InComparisonToExplodeRule`. With this PR, the chain of transformations goes:

`Value` created by query processor -> call to `toQueryPredicate()` during semantic processing ->  `ValuePredicate` ->  `InComparisonToExplodeRule` -> data flow

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
hatyo: The provided `TypeCode` is the type code of the list item(s), in other words, all list items must share the same type.

In `RelOpValue` we do "implicit" type promotion by the ability to choose a physical operator. The lack of a physical operator represents improper comparison (e.g. `foo` > 5).

I am not sure we can do the same trick here though, as individual array items must "agree" on a common max-type before we can form the in-list during plan generation, for that to happen, we can probably re-use the promotion API you introduced in `insert` code stream. But I think this can be done in a subsequent PR if I am right.

hatyo:
> > Hi, I think we should make `IN`s separate from `RelOps`.
> 
> How close are we to just doing it right from the start?
> 
> ```
> x IN (2,3)
> ```
> 
>  ```
> (x) IN ((2),(3))
> (x,y) IN ((2,20),(3,30))
> ```
> 
>  ```
> (x) = ANY (VALUES (2),(3))
> (x,y) = ANY (VALUES (2,20),(3,30))
> (x,y) < ANY (VALUES (2,20),(3,30))
> ```
> 
>  ```
> EXISTS (SELECT * FROM (VALUES (2),(3)) AS _v(_v1) WHERE (x) = (_v1))
> EXISTS (SELECT * FROM (VALUES (2,20),(3,30)) AS _v(_v1,_v2) WHERE (x,y) < (_v1,_v2))
> ```
> 
> Is the problem recovering the simple `List.contains` behavior when no indexes are present from that most general form? That seems like the most motivating reason to perhaps model it like is done here for now.

To answer the first part of your question, I think currently we do not support comparisons of complex types. Since we only compare primitive types we are able to get away with doing type castings implicitly in the physical operator. It is not possible to use this approach for complex types. As I suggested in my comment to @normen662, we can address this by using the type promotion API he introduced in his insert PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
g31pranjal: corrected

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Nonnull
+    @Override
+    public String explain(@Nonnull final Formatter formatter) {
+        return "(" + leftChild.explain(formatter) + " IN " + values.explain(formatter) + ")";
+    }
+
+    @Override
+    public String toString() {
+        return "IN(" + leftChild + ", " + values + ")";
```
normen662, line 164: Let's do `(x IN [1, 2, 3])` so the infix-ness is preserved.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
```diff
@@ -0,0 +1,220 @@
...
+ * A {@link Value} that checks if the left child is in the list of values.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class InOpValue implements BooleanValue {
+    private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("In-Op-Value");
+
+    @Nonnull
+    private final Value leftChild;
+    @Nonnull
+    private final AbstractArrayConstructorValue values;
```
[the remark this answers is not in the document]
normen662: Also don't you just care that this `Value` is of type `Array`? Is it important that you actually have an array constructor at hand? (I looked ahead and I see you are tunneling through that but I guess we should talk about that).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -439,12 +439,9 @@ static Type maximumType(@Nonnull final Type t1, @Nonnull final Type t2) {
             if (PromoteValue.resolvePromotionFunction(t1, t2) != null) {
                 return t2.withNullability(isResultNullable);
             }
-
-            if (PromoteValue.isPromotionNeeded(t2, t1)) {
-                return t2.withNullability(isResultNullable);
+            if (PromoteValue.resolvePromotionFunction(t2, t1) != null) {
+                return t1.withNullability(isResultNullable);
             }
-
```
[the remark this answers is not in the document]
g31pranjal: I just assumed that the above cases are exhaustive wrt Type being primitive. I have put it back.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
g31pranjal: As per our talk, reverted the support for NULLs

normen662 (approved):

normen662 (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
g31pranjal: We now actually have a way of type promotion for elements in the Array by means of using `AbstractArrayConstructorValue`. It does implicit promotion, if possible, and hence provides with an array of homogeneous type.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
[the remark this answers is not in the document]
g31pranjal: just using `Value` for the arrays, as discussed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
```diff
@@ -0,0 +1,220 @@
...
+        builder.addAll(values.getChildren());
+        return builder.build();
+    }
+
+    @Nonnull
+    @Override
+    public InOpValue withChildren(final Iterable<? extends Value> newChildren) {
+        Verify.verify(Iterables.size(newChildren) == 2);
+        Verify.verify(Iterables.get(newChildren, 1) instanceof AbstractArrayConstructorValue);
+        return new InOpValue(Iterables.get(newChildren, 0),
```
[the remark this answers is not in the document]
g31pranjal: reverted.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/InOpValueTest.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+class InOpValueTest {
+
+    private static final TypeRepository.Builder typeRepositoryBuilder = TypeRepository.newBuilder().setName("foo").setPackage("a.b.c");
+
+    @SuppressWarnings("ConstantConditions")
+    static class ThrowsValue implements BooleanValue {
```
hatyo, line 54: This class is duplicate from `BooleanTestValue`, could you please factor it out?

hatyo (changes requested):

normen662 (approved):

hatyo (approved):
Nice work @g31pranjal.

PR #1949, opened 2022-12-16
Delegate accumulator finalisation to children in RCV.

### LINKED ISSUE #1950: accumulation logic in `RecordConstructorValue` prematurely finalises group
In `RecordConstructorValue`, the group finalisation logic returns `null` immediately if the children accumulators are not built, this does not allow child accumulators to express their, potentially special, semantics in case of empty groups, such as `count` which returns `0` in case of empty group instead of `null`.

hatyo:
This allows the children accumulators in `RecordConstructorValue` to finalise their group instead of prematurely return `null`, this allows some special accumulator semantics to kick in such as `count()` (which returns `0` instead of `null` in case of empty group).

Fixes #1950.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
@@ -282,12 +282,12 @@ public class RecordConstructorValue implements Value, AggregateValue, CreatesDyn
     @Override
     public Accumulator createAccumulator(final @Nonnull TypeRepository typeRepository) {
         return new Accumulator() {
-            List<Accumulator> childAccumulators = null;
+
+            @Nonnull
+            private final List<Accumulator> childAccumulators = buildAccumulators();
+
             @Override
             public void accumulate(@Nullable final Object currentObject) {
-                if (childAccumulators == null) {
-                    childAccumulators = buildAccumulators();
-                }
                 if (currentObject == null) {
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
```diff
@@ -286,12 +286,7 @@ public Accumulator createAccumulator(final @Nonnull TypeRepository typeRepositor
             @Override
             public void accumulate(@Nullable final Object currentObject) {
                 if (childAccumulators == null) {
-                    final ImmutableList.Builder<Accumulator> childAccumulatorsBuilder = ImmutableList.builder();
-                    for (final var child : getChildren()) {
-                        Verify.verify(child instanceof AggregateValue);
-                        childAccumulatorsBuilder.add(((AggregateValue)child).createAccumulator(typeRepository));
-                    }
-                    childAccumulators = childAccumulatorsBuilder.build();
+                    childAccumulators = buildAccumulators();
```
MMcM, line 289: Can this just be done in the "constructor," that is, make `childAccumulators` final? I believe `buildAccumulators` works in the parent class, if `typeRespository` is made an argument.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
hatyo: done.

MMcM (approved):

PR #1954, opened 2023-01-06
Fixes #1953: LuceneRecordCursor.getTerms missing PrefixQuery

### LINKED ISSUE #1953: LuceneRecordCursor.getTerms missing PrefixQuery
Because of the way this code is structured, it needs to know every possible Lucene search node.

MMcM:
* Add a test for the missing case.
* Properly extract the term from that class.
* Generalize the auto-complete prefix highlighting to handle this new term.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
@@ -53,6 +53,7 @@ import org.apache.lucene.search.BoostQuery;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.MultiPhraseQuery;
 import org.apache.lucene.search.PhraseQuery;
+import org.apache.lucene.search.PrefixQuery;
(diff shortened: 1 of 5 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -434,6 +435,11 @@ private static void getTerms(Query query, Map<String, Set<String>> map) {
             Term term = spanTermQuery.getTerm();
             map.putIfAbsent(term.field(), new HashSet<>());
             map.get(term.field()).add(term.text().toLowerCase(Locale.ROOT));
+        } else if (query instanceof PrefixQuery) {
+            PrefixQuery termQuery = (PrefixQuery) query;
+            Term term = termQuery.getPrefix();
+            map.putIfAbsent(term.field(), new HashSet<>());
+            map.get(term.field()).add(term.text().toLowerCase(Locale.ROOT) + "*");
```
MMcM, line 442: I don't have any problem making this a structure instead of a special string, although it makes the diff larger.

scottfines (approved):

PR #1956, opened 2023-01-09
Restore generating a prefixed correlation name

hatyo:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/ProtoUtils.java
```diff
@@ -43,7 +43,7 @@ public static String uniqueTypeName() {
      * @return a unique type name.
      */
     public static String uniqueCorrelationName() {
-        return uniqueName("");
+        return uniqueName("__corr__");
```
hatyo, line 46: @normen662 this restores a [reverted](https://github.com/FoundationDB/fdb-record-layer/commit/0e960461eb848c3373d6cc650c8459fb83cb2072) fix: a `CorrelationIdentifier` can be used indirectly to generate a ProtoBuf descriptor by having its name used for a `Record` type name in some plans. Should we prefix it differently to avoid `__corr___corr__` is names?

normen662: where in the photo layer is this needed be an identifier? The only reference to this method I found was to create

normen662 (approved):

PR #1958, opened 2023-01-10
Resolve  #1957: Create base class for KeyValueCursor and Builder, use with IndexPrefetchRangeKeyValueCursor

### LINKED ISSUE #1957: Refactor KeyValueCursor to allow extension
From the Remote fetch code review:
It kind of seems like rather than having IndexPrefetchRangeKeyValueCursor extend KeyValueCursor that both should extend a new abstract class called something like BaseKeyValueCursor<K extends KeyValue> implements BaseCursor<K>.

Our general convention would be to put the newBuilder method on the to-be-built class (so you call Foo.newBuilder() to get a new Foo.Builder), though it looks like we took a somewhat non-standard approach with KeyValueCursor and made people call withSubspace. Maybe it's fine, then, if this is like this, or if it is something like .withSubspaceAndMapper

ohadzeliger:
This is clearing some technical debt from the original remote fetch implementation. One of the PR comments was to implement a base class for KeyValueCursor and extend it for the base KeyValueCursor and the RemoteFetch variant.
A few other fixes were made in support of SonarQube errors, but the overall logic should not have changed.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
@@ -177,7 +177,7 @@ public abstract class KeyValueCursorBase<K extends KeyValue> extends AsyncIterat
      * A builder for {@link KeyValueCursorBase}.
      * @param <T> the type of the concrete subclass of the builder
      * This follows a pattern for nested builder class, hence is generic, so that subclasses can pass in their type (and
-     * implement {@link #getThis()}) to return the subclass builder correct type.
+     * implement {@link #self()}) to return the subclass builder correct type.
      *
      * <pre><code>
      * KeyValueCursorSubclass.Builder.withSubspace(subspace)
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -0,0 +1,410 @@
...
+     * KeyValueCursorSubclass.Builder.withSubspace(subspace)
+     *                     .setContext(context)
+     *                     .setRange(TupleRange.ALL)
+     *                     .setContinuation(null)
+     *                     .setScanProperties(ScanProperties.FORWARD_SCAN)
+     *                     .build()
+     * </code></pre>
+     */
+    @API(API.Status.MAINTAINED)
+    public abstract static class Builder<T extends Builder<T>> {
```
alecgrieser, line 192: Should this also be parameterized on the thing it builds, and then should there be an abstract `build` method that returns that thing? In theory, that would allow someone to write a method that took a `KeyValueCursorBase.Builder` and then it could build the cursor, but I'm not sure how useful that actually is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -0,0 +1,410 @@
...
+            } else if (propertiesStreamingMode == CursorStreamingMode.SMALL) {
+                return StreamingMode.SMALL;
+            } else if (limit == ReadTransaction.ROW_LIMIT_UNLIMITED) {
+                return StreamingMode.WANT_ALL;
+            } else {
+                return StreamingMode.EXACT;
+            }
+        }
+
+        protected abstract T getThis();
```
alecgrieser, line 408: minor: I believe I've seen this referred to as "self" and the type parameter as "self type" (for example, in this PEP: https://peps.python.org/pep-0673/). It may be useful to refer to it as such here

alecgrieser (changes requested):
I left a few minor things, but overall, looks good!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
ohadzeliger: That's a good idea (™) but I couldn't make it work due to some strange quirks of the Java typing system. It appears that you can't add a second generic type marker to the Builder class for some reason (bug in Java interpreter?) and trying to achieve the same goal by introducing an abstract `build()` method on a different `<V extends KeyValue>` introduces `unchecked` errors.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
ohadzeliger: Done.

alecgrieser (approved):

PR #1960, opened 2023-01-11
resolve fetch index method for lucene index scans correctly

normen662:


scottfines (approved):

arnaud-lacurie (dismissed):

scottfines (dismissed):

arnaud-lacurie (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1458,6 +1458,13 @@ private boolean avoidScanPlan(@Nonnull PlanContext planContext) {
                queriedRecordTypes.stream().anyMatch(syntheticRecordTypes::contains);
     }
 
+    @Nonnull
+    protected FetchIndexRecords resolveFetchIndexRecords(@Nonnull PlanContext planContext) {
```
MMcM, line 1462: If I am not mistaken, rather than adding a second signature like this, the existing method could be replaced by the new one. The only other caller of the existing method gets the two necessary arguments in the same way and does not use them for anything other than invoking the method in question. Am I misreading?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: Done.

arnaud-lacurie (approved):

PR #1963, opened 2023-01-12
Expose version change info from FDBRecordStore

### LINKED ISSUE #1965: Expose version change info from FDBRecordStore

g31pranjal:
Adds a boolean flag `VersionChanged` to indicate if the version has been changed while creating/opening a new store. 

When a RecordStore is created via `createOrOpen()`, `create()` or `open()` methods in `FDBRecordStore.Builder`, the `checkVersion()` is called underneath to update the store versions to the newest possible values and returns a boolean value. `VersionChanged` persists this boolean value that can be retrieved from the `Builder` _after_ creating/opening a (new) store. 

The version changed info can therefore be made available anywhere the store is built.

fixes #1965

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -26,7 +26,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Refactor KeyValueCursor to allow extension [(Issue #1957)](https://github.com/FoundationDB/fdb-record-layer/issues/1957)
-* **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Expose version change info from FDBRecordStore [(Issue #1965)](https://github.com/FoundationDB/fdb-record-layer/issues/1965)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Indexing: Throw PartlyBuiltException when appropriate [(Issue #1961)](https://github.com/FoundationDB/fdb-record-layer/issues/1961)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -354,11 +354,12 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     }
 
     /**
-     * Get the provider for the record store's meta-data.
-     * @return the meta-data source to use
+     * Returns {@code true} if the RecordMetadata version is changed in the RecordStore. Alternatively, returns
+     * {@code false} if the version is either not checked (checkVersion() not called) or it is up-to-date.
+     * @return the versionChanged boolean
      */
     @Nonnull
-    public Boolean getVersionChanged() {
+    public boolean getVersionChanged() {
         return versionChanged;
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -2031,7 +2031,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         }
         CompletableFuture<Boolean> result = storeHeaderFuture.thenCompose(storeHeader -> checkVersion(storeHeader, userVersionChecker));
         return context.instrument(FDBStoreTimer.Events.CHECK_VERSION, result).thenApply(versionChanged -> {
-            this.versionChanged |= versionChanged
+            this.versionChanged |= versionChanged;
             return versionChanged;
         });
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -2031,9 +2031,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         }
         CompletableFuture<Boolean> result = storeHeaderFuture.thenCompose(storeHeader -> checkVersion(storeHeader, userVersionChecker));
         return context.instrument(FDBStoreTimer.Events.CHECK_VERSION, result).thenApply(versionChanged -> {
-            if (!this.versionChanged && versionChanged) {
-                this.versionChanged = true;
-            }
+            this.versionChanged |= versionChanged
             return versionChanged;
         });
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -242,6 +242,9 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     @Nonnull
     protected final RecordMetaDataProvider metaDataProvider;
 
+    @Nullable
+    private boolean versionChanged;
+
     @Nonnull
     protected final AtomicReference<MutableRecordStoreState> recordStoreStateRef = new AtomicReference<>();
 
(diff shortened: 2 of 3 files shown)


alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4620,6 +4620,9 @@ public static class Builder implements BaseBuilder<Message, FDBRecordStore> {
         @Nullable
         private FDBRecordStoreStateCache storeStateCache = null;
 
+        @Nullable
+        private Boolean versionChanged;
```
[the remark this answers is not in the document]
g31pranjal: My idea was to make the version change information available in case the consumer is calling vanilla versions of `create()`, `open()`, `createOrOpen` - which all route to `createOrOpenAsync()` and also "guarantees" checking version on startup. 

I am not quite sure of the contract here though, but this is what I perceive. `build()` is essentially doing the bare minimum thing of unintelligently initializing the store and `unchecked__()` are indicative of skipping version checks IMO. If the consumer goes on to call `checkVersion` on the returned store, the version changing information is readily available. However, extending your point, there is definitely info loss if there are multiple calls to `build()` / `createOrOpen()`, which is unusual in general. 

Maybe we can just make the API to be such that the versionChanged information is available _only_ when the store is initialized with checks and is not guaranteed to be always in sync. 

I am fine doing it the other way but the contract there will be more unclear as checkVersion() is heavily called in conjunction with other basic operations of the store and version change information can be lost even there.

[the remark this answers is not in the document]
alecgrieser: Yeah, that's right when it comes to the contract of `build()` and `uncheckedOpen()`. They are mainly there so that users who want to have more control over when check version gets called, but it's debatable how wise they are, given how important it is to call `checkVersion`.

The way I was imagining it, `versionChanged` on the store would be updated at the end of `checkVersion`, and it would be guaranteed to be in sync if and only if (1) a `checkVersion` call has completed and (2) no other `checkVersion` call has started. If we really cared about cases where we call `checkVersion` multiple times, we could also make it a counter instead of a Boolean, though that seems like it could be overkill.

What kind of loss-of-version-change information are you envisioning? Just the case where check version is called multiple times, or something else? To my knowledge, the logic in `FDBRecordStore` that makes sure `checkVersion` has been called (such as it is) won't try and call it a second time if it has already been called.

On the other hand, my reservation about putting this information in the builder is mainly that people don't expect builders to retain state from the thing they've built, and in particular, use cases that re-use builders (perhaps swapping out the transaction to open otherwise identical record stores with different transactions) can unexpectedly carry over state between calls to `open()` (and variants).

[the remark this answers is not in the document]
g31pranjal: That's exactly what I am worried about, the consumer calls `checkVersion` multiple times. In that case, we might lose the relevant state of the flag from a previous `checkVersion` call. I think this doesn't happen in the `FDBRecordStore` as such, but we should be wary of it being misused. 

However, we can decide on the contract such as`VersionChanged` is true if there is a version update at least once in the lifetime of the transaction (and `FDBRecordStore`).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -350,6 +353,15 @@ public RecordMetaDataProvider getMetaDataProvider() {
         return metaDataProvider;
     }
 
+    /**
+     * Get the provider for the record store's meta-data.
+     * @return the meta-data source to use
+     */
+    @Nonnull
+    public Boolean getVersionChanged() {
```
alecgrieser, line 361: Should this be a `boolean`? It looks like the member variable is already

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -350,6 +353,15 @@ public RecordMetaDataProvider getMetaDataProvider() {
         return metaDataProvider;
     }
 
+    /**
+     * Get the provider for the record store's meta-data.
```
alecgrieser, line 357: It looks like this description needs to be updated after a copy/paste error

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreOpeningTest.java
```diff
@@ -118,6 +119,7 @@ public void open() throws Exception {
             RecordMetaDataBuilder metaDataBuilder = RecordMetaData.newBuilder().setRecords(TestRecords1Proto.getDescriptor());
 
             FDBRecordStore recordStore = storeBuilder(context, metaDataBuilder).createOrOpen();
+            assertTrue(recordStore.getVersionChanged());
```
alecgrieser, line 122: Maybe I'm missing it, but is there a test for what happens if you call `checkVersion` twice, and the version is changed during the first call but then not in the second? I think you were concerned an update could be lost in that case. (It's also possible that that's already covered in this test.)

alecgrieser (changes requested):
A few minor things, but looking good. This probably should also update the release notes: https://github.com/FoundationDB/fdb-record-layer/blob/main/docs/ReleaseNotes.md In general, changes that a user might care about (that is, not counting things like test-only changes or build changes) should update one of the commented out release notes at the top of the file, and then the release process updates that file when a new version is cut

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
g31pranjal: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
g31pranjal: done

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreOpeningTest.java
g31pranjal: added this.

alecgrieser (dismissed):

docs/ReleaseNotes.md
```diff
@@ -25,7 +25,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Expose version change info from FDBRecordStore [(Issue #1965)](https://github.com/FoundationDB/fdb-record-layer/issues/1965)
```
alecgrieser, line 28: This will re-direct, but technically the issue number is #1963, and the PR is #1965

g31pranjal: #1965 is the issue. I timed it wrong

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2017,7 +2030,12 @@ private CompletableFuture<Boolean> checkVersion(@Nullable UserVersionChecker use
             storeHeaderFuture = subspacePreloadFuture.thenCombine(storeHeaderFuture, (vignore, storeHeader) -> storeHeader);
         }
         CompletableFuture<Boolean> result = storeHeaderFuture.thenCompose(storeHeader -> checkVersion(storeHeader, userVersionChecker));
-        return context.instrument(FDBStoreTimer.Events.CHECK_VERSION, result);
+        return context.instrument(FDBStoreTimer.Events.CHECK_VERSION, result).thenApply(versionChanged -> {
+            if (!this.versionChanged && versionChanged) {
+                this.versionChanged = true;
+            }
+            return versionChanged;
+        });
```
ScottDugas, line 2038: I think this could be simplified to 
```suggestion
        return context.instrument(FDBStoreTimer.Events.CHECK_VERSION, result).thenApply(versionChanged -> {
            this.versionChanged |= versionChanged
            return versionChanged;
        });
```

But `|=` is rarely used, so being more explicit here, might improve readibility.

ScottDugas (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
g31pranjal: done

alecgrieser (approved):

alecgrieser (approved):

ScottDugas (approved):

PR #1966, opened 2023-01-13
Allowing boolean terms to be stored and queried using a single byte in Lucene

scottfines:
Just a quick bit of parser magic so that boolean fields in the lucene index can be represented using either `0x00` (for false) or `0x01` (for true)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/search/BooleanPointsConfig.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.text.NumberFormat;
+
+/**
+ * A subclass of PointsConfig to allow the Parser the ability to translate boolean terms to binary ones.
+ */
+public class BooleanPointsConfig extends PointsConfig {
+
+    public static final byte[] TRUE_BYTES = {0x01};
+    public static final byte[] FALSE_BYTES = {0x00};
```
MMcM, line 33:
```
    @SuppressWarnings("squid:S2386")
    @SpotBugsSuppressWarnings("MS_MUTABLE_ARRAY")
```

MMcM (changes requested):

MMcM (approved):

PR #1968, opened 2023-01-16
Resolves #1967: Consider alternative mechanism of loading records which favors gets over scans

### LINKED ISSUE #1967: Consider alternative mechanism of loading records which favors gets over scans
For LSM-backed storage engines, single key gets can be more efficient than small scans over the same key, mainly due to how many files the LSM will need to consult to serve the read. At present, the Record Layer does almost no single key gets, partially because the system was designed and implemented with a B-tree based storage engine in mind and partially because the ordered key-value API of FDB makes that design easy to implement in many cases.

One place we could think about changing this might be with single-record reads. For most record stores, record reads even of a single record by primary key currently require a small scan. This is because FDB values are limited in size to 100 kB, and so to support larger records than that, the record must somehow be split across multiple keys. The format we chose was that:

1. For unspilt records, we suffix the primary key with a tuple-encoded 0 and store all of the record data at that key
2. For split records, we split the record into *n* 100 kB chunks and assign each chunk a number from 1 to *n*
3. For records with a version (which we store separately from the record because it is manipulated in a somewhat special way), the version is stored at split point -1

When loading a single record, the client can then scan a subspace that includes the record's primary key at the end to list all of the split points for that record and then combine the results.

For an LSM, this could be significantly worse performance than getting the key(s). There are a few strategies we might pursue in making this better, but one that is fairly straightforward to implement would be:

1. The client reads split points -1, 0, and 1 in parallel.
2. If split point 0 contains data, then that's the only record data we need to care about
3. If split point 1 contains data, then a scan needs to be issued to load split points 2 to _n_
4. The version (read from split point -1) can then be attached and returned with the read data

Note that the only time we need to issue a scan is in the case where there really is a split record. If there is no record at all, then all three reads should return empty (which, with Bloom filters on the LSM SSTable files should be fairly efficient, at least most of the time). If the record is present and not split, then that can be determined without any scans. The downside is that this needs to make three requests, and in the split case, that means issuing and waiting on the reads and _then_ issuing the scan. So, even for LSM-backed clusters, this could be problematic if there are lots of split records. Additionally, the extra requests would effectively triple the number of read requests on workloads with many load records, which could increase network traffic and storage server CPU costs. There are some ways (with different APIs) that we may be able to get around those problems, but doing the most straightforward thing to see how it performs may allow us to get insight into whether this is an optimization we want to pursue further.

alecgrieser:
This adds a property-controlled method for reading single records from FDB. It issues multiple single-key gets instead of one scan (for reasons detailed in #1967). This is off by default, because it is expected that this will perform worse on B-tree based storage engines (including the default SSD storage engine), and it is highly experimental. But once in, users can adjust this configuration to configure how reads are orchestrated in their own experiments.

This resolves #1967.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelper.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelper.java
@@ -58,6 +58,8 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
 
 /**
  * Helper classes for splitting records across multiple key-value pairs.
@@ -289,6 +291,10 @@ public class SplitHelper {
         if (!splitLongRecords && missingUnsplitRecordSuffix) {
             return loadUnsplitLegacy(tr, context, subspace, key, sizeInfo);
         }
+        if (context.getPropertyStorage().getPropertyValue(FDBRecordStoreProperties.LOAD_RECORDS_VIA_GETS)) {
+            return loadSplitViaGets(tr, context, subspace, key, sizeInfo);
+        }
+
         // Even if long records are not split, then unless we are using the old format, it might be the case
         // that there is a record version associated with this key, hence the range read.
         // It is still better to do a range read in that case than two point reads (probably).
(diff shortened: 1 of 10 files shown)


MMcM (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreSplitRecordsTest.java
```diff
@@ -800,22 +869,52 @@ private void checkForConflicts(int formatVersion, boolean splitLongRecords, @Non
...
+                FDBRecordStore recordStore = getStoreBuilder(context, simpleMetaData(hook))
+                        .setFormatVersion(formatVersion)
+                        .createOrOpen();
+                // Insert a record here so that the delete logic (1) sees the record is there and then
+                // (2) issues a real delete
+                recordStore.saveRecord(TestRecords1Proto.MySimpleRecord.newBuilder().setRecNo(1066L).build());
+                commit(context);
+            }
+            checkForConflicts(formatVersion, splitLongRecords,
+                    store1 -> store1.deleteRecord(Tuple.from(1066L)),
```
ScottDugas, line 913: Would there be benefit to adding another version that does a save, and then gets a conflict on the delete?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
```diff
@@ -122,13 +125,90 @@ public void setSubspace() {
         }
     }
 
+    public static Stream<Pair<Boolean, Boolean>> splitAndSuffixArgs() {
```
ScottDugas, line 128: I think this and `splitSuffixAndUnrollArgs` can be removed now.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
```diff
@@ -633,25 +725,33 @@ private FDBRawRecord loadWithSplit(@Nonnull FDBRecordContext context, @Nonnull T
...
 
     @Nullable
-    private FDBRawRecord loadWithSplit(@Nonnull FDBRecordContext context, @Nonnull Tuple key, boolean splitLongRecords, boolean omitUnsplitSuffix,
+    private FDBRawRecord loadWithSplit(@Nonnull FDBRecordContext context, @Nonnull Tuple key, SplitHelperTestConfig testConfig,
                                        @Nullable FDBStoredSizes expectedSizes, @Nullable byte[] expectedContents) {
-        return loadWithSplit(context, key, splitLongRecords, omitUnsplitSuffix, expectedSizes, expectedContents, null);
+        return loadWithSplit(context, key, testConfig, expectedSizes, expectedContents, null);
+    }
+
+    private static Stream<Arguments> loadWithSplit() {
```
ScottDugas, line 733: I think this can be removed, as it is unused

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelper.java
```diff
@@ -289,19 +291,104 @@ public static CompletableFuture<FDBRawRecord> loadWithSplit(@Nonnull final ReadT
...
+                                                                    @Nullable SizeInfo sizeInfo) {
+        final SizeInfo storedSizes = sizeInfo == null ? new SizeInfo() : sizeInfo;
+        storedSizes.reset();
+
+        final Subspace recordSubspace = subspace.subspace(key);
+        final byte[] versionKey = recordSubspace.pack(RECORD_VERSION);
+        final byte[] unsplitKey = recordSubspace.pack(UNSPLIT_RECORD);
+        final byte[] startSplitKey = recordSubspace.pack(START_SPLIT_RECORD);
+
+        // Cover the whole record range in a singe read conflict range to decrease size of final commit request
```
ScottDugas, line 325:
```suggestion
        // Cover the whole record range in a single read conflict range to decrease size of final commit request
```

ScottDugas (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreSplitRecordsTest.java
alecgrieser: Sure, added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelper.java
alecgrieser: ✅

ScottDugas (dismissed):

alecgrieser:
Rebased to fix up merge conflict in `ReleaseNotes.md`

alecgrieser:
Okay, after the rebase, the test `testVersionChangedFlagAfterOpening` started failing. I eventually tracked it down to state in a static `Index` option changing when added to a meta-data builder object. I...am not quite sure how my other changes induced this change (maybe something to do with test execution order), but fixes for that and some other tidying of that test are in 787a730 now to avoid breaking the build

alecgrieser:
Rebase addresses merge conflicts (in `docs/ReleaseNotes.md`)

ScottDugas (approved):

PR #1969, opened 2023-01-17
Some improvements for Lucene queries against joined indexes

MMcM:
1. Allow longer field names for Lucene validation.
    In effect, any path prefix is accepted.
    The fields in the Lucene query component field list sometimes reflect more of the path, particularly with a Lucene join index.

2. Allow highlighting to work on a constituent record from a joined record Lucene index.
    For this we need the constituent field name and to adjust the index expression and terms map accordingly.

3. Generalize constituent handling around `FDBQueriedRecord` beyond just `getConstituents`.
    Can also get a particular constituent, which requires a little work in the case of covering.
    This adds some non-abstract methods, but there doesn't seem any philosophical objection to that.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
@@ -31,7 +31,6 @@ import com.google.protobuf.Message;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -62,12 +65,68 @@
...
+    @Nullable
+    public abstract String getConstituentName();
+
     /**
      * The constituents of this record if it is a synthetic record type.
      * @return a map of constituents from name to {@link FDBStoredRecord}
      */
     @Nullable
-    public abstract Map<String, FDBStoredRecord<? extends Message>> getConstituents();
+    public Map<String, FDBStoredRecord<? extends Message>> getConstituents() {
```
normen662, line 87: Can we put this into `SyntheticRecord`? Seems like a bunch of stuff logic that nulls out if `getSyntheticRecord()` nulls out.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -62,12 +65,68 @@
...
+    }
+
+    /**
+     * Get a constituent record of this record if it is a synthetic record type.
+     * @param constituentName name of the constituent
+     * @return a constituent record of {@code null}
+     */
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public FDBQueriedRecord<M> getConstituent(@Nonnull String constituentName) {
```
normen662, line 102: I guess this needs to stay to deal with the covered records that can be synthetic records? Not too happy with that ambiguity but I don't have anything better.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
```diff
@@ -278,7 +353,45 @@ public FDBStoredRecord<M> getStoredRecord() {
...
+        @Nullable
+        @Override
+        public String getConstituentName() {
+            return null;
+        }
+
+        @Nullable
+        @Override
+        @SuppressWarnings("unchecked")
+        public FDBQueriedRecord<M> getConstituent(@Nonnull String constituentName) {
```
normen662, line 369: Ah, so we have synthetic records that are covered records. That's why you expose `getConstituent()` directly. Correct?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBQueriedRecord.java
MMcM: Correct. There are already `FDBQueriedRecord`s that aren't `FDBStoredRecord`, but implement some similar functionality when coming from covering scans. And there are already `FDBQueriedRecord`s for synthetic record types that aren't `FDBSyntheticRecord` when coming from covering scans. These changes add some additional constituent-related operations around these, staying within the `FDBQueriedRecord` realm.

Fundamentally, having to pretend that covering scans produce records at all is problematic. It gets somewhat worse with join indexes. But fixing that is a comprehensive change.

scottfines (approved):

normen662 (approved):

PR #1972, opened 2023-01-20
Resolves #1997: Sparse index support in Cascades.

### LINKED ISSUE #1997: Add support for planning sparse (filtered) indexes in Cascades.
record layer supports sparse (filtered) indexes, i.e. an index that indexes only a subset of records based on a user-defined predicate.

We should provide support for planning these indexes properly since using them might boost performance as they usually contain much less entries compared to non-filtered indexes and therefore faster index scans.

This requires though two things:
1. supporting serialisation of index filter metadata so Cascades can use this information to model index filters as predicates in the corresponding match candidate.
2. perform range implication between query predicate and respective match candidate placeholders and only allow a filtered index to be used it is range(s) of tuples implies the query's requested range(s) of records.

hatyo:
This implements support for serialising and planning filtered value- and aggregate-indexes in Cascades.

**Details:**
record layer supports sparse (filtered) indexes, i.e. an index that indexes only a subset of records based on a user-defined predicate.

We should provide support for planning these indexes properly since using them might boost performance as they usually contain much less entries compared to non-filtered indexes and therefore faster index scans.

This PR is split into two major parts:

A. supporting serialisation of index filter metadata so Cascades can use this information to model index filters as predicates in the corresponding match candidate.

This step is done as the following:
* a new message names `Predicate` is added to `record_metadata.proto` which represents a simple, recursive, boolean expression. It is added as an `optional` field to `Index` message. It is marked, like all other fields, for backward- and forward-compatibility.
* a Plain old Java object (POJO) type hierarchy `IndexPredicate` (and `IndexComparison`) is added to wrap (de)serialisation logic from/to the above `Predicate` message. It resembles very closely a subset of types in `QueryPredicate` (resp. `Comparisons.Comparison`) and offer a way to convert to it as well.

The deserialised predicate is assumed to be in Disjunctive Normal Form (DNF) whose terms are either flat or a conjunction of flat predicates belonging to the same predicate. It is the consumer's responsibility to guarantee that the predicate is serialised in this form (because it can be expensive to reconstruct it).

B. perform range implication between query predicate and respective match candidate placeholders and only allow a filtered index to be used it is range(s) of tuples implies the query's requested range(s) of records.

This enables Cascades to reason about predicate range matching better and potentially choose the better (filtered) index. The idea is, if a query contains a predicate whose range is: A) evaluable at runtime and B) implied by a range predicate of a filtered index, then that index is matched (without loss of generality), conversely, if an index defines a predicate on a non-indexed field, (a.k.a value restriction) it will _not_ be chosen if the query has a predicate on that field that is neither compile-time nor compile-time but not implied but the value restriction.

Finally, the `OrPredicate` matching is improved such that it works better with compile-time ranges; if two `OrPredicate` are being checked for matching, and both contain compile time ranges in their disjuncted children, and _each_ `or` child on the left is implied by _at least_ one child from the `or` on the right, we establish a mapping between both. If that does not work, then Cascades tries other matching strategies including splitting the `or` into a `union`.

This helps optimising queries with a disjunction predicate to use a _single_ filtered index scan who is defined with an _almost_ matching disjunction filter instead of a, potentially, more expensive union plan.

When deserialising an index predicate, we check if the index is in DNF form, and can be transformed into a set of values with disjuncted ranges (`ValueWithRanges`), if this is not possible, e.g. because the DNF has a term which is a conjunction of more than variable predicate, then we add the predicate as-is, this effectively means that such as index can be matched against a query if and only if the query had a predicate that is semantically (and structurally) equal to the index predicate itself. In future iterations we'll add better support to multi-dimensional ranges (with free variables).

This resolves #1997.


fdb-record-layer-core/src/main/proto/record_metadata.proto
```diff
@@ -48,7 +48,6 @@ message Index {
     RANK = 3;
     RANK_UNIQUE = 4;    // (Meaning the score is unique.)
   };
-  // TODO: Generalize to predicate on records in storage area to be included in index.
```
MMcM, line 51: That comment is from 2015.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
```diff
@@ -646,12 +696,13 @@ public boolean equals(Object o) {
                 && this.addedVersion == that.addedVersion
                 && this.lastModifiedVersion == that.lastModifiedVersion
                 && Arrays.equals(this.primaryKeyComponentPositions, that.primaryKeyComponentPositions)
-                && this.options.equals(that.options);
+                && this.options.equals(that.options)
+                && (Objects.equals(this.predicate, that.predicate));
```
MMcM, line 700:
```suggestion
                && Objects.equals(this.predicate, that.predicate);
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
```diff
@@ -646,12 +696,13 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public int hashCode() {
         // Within the context of a single RecordMetaData, this should be sufficient
```
MMcM, line 705: Why isn't this still true? If you had a `Map` from `Index`, presumably all from the same `RecordMetaData`, wouldn't this be good enough?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
```diff
@@ -99,7 +99,7 @@ default List<MatchedOrderingPart> computeMatchedOrderingParts(@Nonnull MatchInfo
             builder.add(
                     MatchedOrderingPart.of(value,
                             comparisonRange == null ? ComparisonRange.Type.EMPTY : comparisonRange.getRangeType(),
-                            queryPredicate,
+                            (comparisonRange == null || comparisonRange.isEmpty()) ? null : queryPredicate,
```
normen662, line 102: Not sure what that does.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
```diff
@@ -107,6 +113,140 @@ public OrPredicate withChildren(final Iterable<? extends QueryPredicate> newChil
...
+     * - no match.
+     *
+     * @param aliasMap the current alias map.
+     * @param candidatePredicate another predicate to match.
+     * @return optional match mapping.
+     */
+    @Nonnull
+    @Override
+    public Optional<PredicateMultiMap.PredicateMapping> impliesCandidatePredicate(@NonNull final AliasMap aliasMap, @Nonnull final QueryPredicate candidatePredicate) {
+        final var valueWithRangesMaybe = toValueWithRangesMaybe(this);
```
normen662, line 198: Not good I think, this should at least be memoized. But this is also in contrast to how `SelectExpression` currently groups their predicates.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/Value.java
```diff
@@ -174,15 +173,15 @@ default ValuePredicate withComparison(@Nonnull Comparisons.Comparison comparison
...
      * that is used solely for matching query predicates against.
      * @param parameterAlias alias to uniquely identify the parameter in the
      *        {@link com.apple.foundationdb.record.query.plan.cascades.MatchCandidate} this placeholder will be a part of.
-     * @return a new {@link Placeholder}
+     * @return a new {@link ValueWithRanges} that has {@code parameterAlias} alias.
      */
     @Nonnull
-    default Placeholder asPlaceholder(@Nonnull final CorrelationIdentifier parameterAlias) {
-        return ValueComparisonRangePredicate.placeholder(this, parameterAlias);
+    default ValueWithRanges asPlaceholder(@Nonnull final CorrelationIdentifier parameterAlias) {
```
normen662, line 183: naming could be updated.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -61,6 +62,9 @@ public interface CompensatePredicateFunction {
                 (partialMatch, boundPrefixMap) -> {
                     throw new RecordCoreException("should not be called");
                 };
+
+        CompensatePredicateFunction EMPTY =
```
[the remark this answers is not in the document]
normen662: Maybe make this `private` and add a new static method returning this `empty()`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -168,13 +179,11 @@ public Optional<CorrelationIdentifier> getParameterAliasOptional() {
 
         @NonNull
         public Optional<ComparisonRange> getComparisonRangeOptional() {
-            if (parameterAliasOptional.isEmpty() || !(queryPredicate instanceof Sargable)) {
+            if (parameterAliasOptional.isEmpty() || !(queryPredicate instanceof ValueWithRanges && ((ValueWithRanges)queryPredicate).getRanges().size() == 1)) {
```
normen662, line 182: Maybe a helper.

fdb-record-layer-core/src/main/proto/record_metadata.proto
hatyo: Was it rightful to remove it now that we have a `Predicate` member of `Index`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Index.java
hatyo: I thought about it again, you're right, it is actually still true even with the addition of the predicate. I will restore it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AliasMap.java
```diff
@@ -287,6 +287,15 @@ public CorrelationIdentifier getSource(final CorrelationIdentifier target) {
         return map.inverse().get(target);
     }
 
+    /**
+     * Returns the inverse of this {@link AliasMap}.
+     * @return The inverse of this {@link AliasMap}.
+     */
+    @Nonnull
+    public AliasMap inverse() {
+        return AliasMap.copyOf(map.inverse());
```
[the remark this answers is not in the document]
hatyo: `ImmutableBiMap#invserse()` does not make a deep copy, it only offers a _view_:

> Returns the inverse view of this bimap, which maps each of this bimap's values to its associated key. The two bimaps are backed by the same data; any changes to one will appear in the other.

I am using the private constructor of `AliasMap` as suggested so make sure we don't copy the view itself.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+
+import static com.apple.foundationdb.record.query.plan.cascades.predicates.RangeConstraints.EvalResult.FALSE;
+import static com.apple.foundationdb.record.query.plan.cascades.predicates.RangeConstraints.EvalResult.TRUE;
+
+/**
+ * A special predicate used to represent a parameterized tuple range.
+ */
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+@API(API.Status.EXPERIMENTAL)
+public class ValueWithRanges implements PredicateWithValue {
```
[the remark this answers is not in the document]
hatyo: As discussed offline, I factored out the alias-related logic into a separate `Placeholder` type that extends this type and implements a trait `WithAlias`. I also added more JavaDoc to the remaining code here.

fdb-record-layer-core/src/main/proto/record_metadata.proto
MMcM: Seems fine. The generalized predicate implements the spirit of the comment, I believe.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    @Nonnull
+    private final Value value;
+
+    @Nonnull
+    private final Optional<CorrelationIdentifier> alias;
```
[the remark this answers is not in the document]
hatyo: As discussed offline, I factored out the alias-related logic into a separate `Placeholder` subtype that extends this type and implements a trait `WithAlias`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+
+    @Nonnull
+    @Override
+    public ValueWithRanges withValue(@Nonnull final Value value) {
+        return new ValueWithRanges(value, ranges, alias);
+    }
+
+    @Nonnull
+    public ValueWithRanges withExtraRanges(@Nonnull final Set<RangeConstraints> ranges) {
+        return new ValueWithRanges(value, Stream.concat(ranges.stream(), this.ranges.stream()).collect(Collectors.toSet()), alias);
```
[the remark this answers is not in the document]
hatyo: Thanks, yes, I will change it to that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+
+    @Nullable
+    @Override
+    public <M extends Message> Boolean eval(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final EvaluationContext context) {
+        throw new RecordCoreException("this method should not ever be reached");
+    }
+
+    @Nonnull
+    @Override
+    public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
```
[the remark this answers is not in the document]
hatyo: ditto.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return semanticHashCode();
+    }
+
+    @Override
+    public int semanticHashCode() {
+        return Objects.hash(value.semanticHashCode());
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    @Override
+    public int planHash(@Nonnull final PlanHashKind hashKind) {
+        return PlanHashable.planHash(hashKind, value);
```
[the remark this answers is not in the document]
hatyo: 👍 thanks, I missed that, I will make it just throw.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    @Nonnull
+    public CorrelationIdentifier getAlias() {
+        Verify.verify(alias.isPresent(), "attempt to retrieve non-existing alias");
+        return alias.get();
+    }
+
+    @Nonnull
+    @Override
+    public ValueWithRanges translateLeafPredicate(@Nonnull final TranslationMap translationMap) {
+        return new ValueWithRanges(value.translateCorrelations(translationMap), ranges.stream().map(range -> range.translateCorrelations(translationMap)).collect(Collectors.toSet()), alias);
```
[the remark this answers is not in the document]
hatyo: ditto.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+            if (!left.isCompileTimeEvaluable() || !left.isEmpty().equals(FALSE) || rightDisjunction.stream().noneMatch(right -> right.implies(left).equals(TRUE))) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Nonnull
+    public static ValueWithRanges placeholder(@Nonnull Value value, @Nonnull CorrelationIdentifier parameterAlias) {
+        return new ValueWithRanges(value, Set.of(), Optional.of(parameterAlias));
```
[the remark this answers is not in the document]
hatyo: done (same for other occurrences).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+            if (candidate.getRanges().isEmpty()) {
+                if (candidate.hasAlias()) {
+                    return Optional.of(new PredicateMapping(this, candidatePredicate, (ignore, boundParameterPrefixMap) -> {
+                        if (boundParameterPrefixMap.containsKey(candidate.getAlias())) {
+                            return Optional.empty();
+                        }
+                        return injectCompensationFunctionMaybe();
+                    }, candidate.getAlias()));
+                } else {
+                    return Optional.empty(); // we should probably throw.
```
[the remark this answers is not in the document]
hatyo: Right, I missed that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    public static ValueWithRanges constraint(@Nonnull final Value value, @Nonnull final Set<RangeConstraints> ranges) {
+        return new ValueWithRanges(value, ranges, Optional.empty());
+    }
+
+    @Nonnull
+    @Override
+    public Optional<PredicateMapping> impliesCandidatePredicate(@NonNull final AliasMap aliasMap,
+                                                                @Nonnull final QueryPredicate candidatePredicate) {
+        if (candidatePredicate.isContradiction()) {
+            return Optional.empty();
```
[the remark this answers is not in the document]
hatyo: I want to cover a particular case where the index contains an impossible predicate `WHERE FALSE`, which is set be the consumer as it wants to indirectly signal to the planner that this particular index may _not_ be used for this query.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+                            return Optional.empty();
+                        }
+                        return injectCompensationFunctionMaybe();
+                    }, candidate.getAlias()));
+                } else {
+                    return Optional.empty(); // we should probably throw.
+                }
+            }
+
+            // if this and candidate and compile-time evaluable.
```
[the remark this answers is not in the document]
hatyo: thanks, removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/RangeConstraints.java
```diff
@@ -0,0 +1,706 @@
...
+     * <li>{@code this = [1, 10), other = () => this.implies(other) = FALSE}</li>
+     * <li>{@code this = {[1, 10), deferred = STARTS_WITH}, other = (2, 5) => this.implies(other) = UNKNOWN}</li>
+     * </ul>
+     *
+     * @param other The other range to compare against.
+     * @return {@code true} if the range is known, at compil-time, to imply {@code other} or {@code false} if not. Otherwise
+     * it returns {@code unknown}.
+     */
+    @Nonnull
+    public EvalResult implies(@Nonnull final RangeConstraints other) {
```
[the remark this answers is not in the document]
hatyo: Good idea, thanks.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public ValueWithRanges translateLeafPredicate(@Nonnull final TranslationMap translationMap) {
+        return new ValueWithRanges(value.translateCorrelations(translationMap), ranges.stream().map(range -> range.translateCorrelations(translationMap)).collect(Collectors.toSet()), alias);
+    }
+
+    private boolean impliedBy(@Nonnull final ValueWithRanges other) {
+        final var leftDisjunction = getRanges();
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/RangeConstraints.java
```diff
@@ -0,0 +1,706 @@
...
+/**
+ * Represents a compile-time range that can be evaluated against other compile-time ranges with an optional list
+ * of deferred ranges that can not be evaluated at compile-time but can still be used as scan index prefix.
+ */
+public class RangeConstraints implements PlanHashable, Correlated<RangeConstraints> {
+
+    /**
+     * Evaluation result, ternary logic.
+     */
+    public enum EvalResult {
```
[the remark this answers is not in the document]
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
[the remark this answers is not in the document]
hatyo: Since it is an interface, we can't hide it. But I added a static method `empty()` and I use it throughout the codebase instead. Thanks for the suggestion.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
```diff
@@ -135,8 +171,9 @@ public MatchCandidate expand(@Nonnull final Supplier<Quantifier.ForEach> baseQua
         }
 
         final var completeExpansion = GraphExpansion.ofOthers(allExpansionsBuilder.build());
-        final var parameters = completeExpansion.getPlaceholderAliases();
-        final var matchableSortExpression = new MatchableSortExpression(parameters, isReverse, completeExpansion.buildSelect());
+        final var sealedExpansion = completeExpansion.seal();
+        final var parameters = sealedExpansion.getPlaceholders().stream().map(ValueWithRanges::getAlias).collect(Collectors.toList());
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
```diff
@@ -149,6 +186,85 @@ public MatchCandidate expand(@Nonnull final Supplier<Quantifier.ForEach> baseQua
...
+     * and list of corresponding {@link RangeConstraints}.
+     * <br>
+     * For example: {@code OR(AND(v1, <3), AND(v2 >4), AND(v1<4))} will be transformed to the following:
+     * {@code v1 -> [(-∞,3), (-∞, 4)], v2 -> [(4, +∞)]}.
+     *
+     * @param predicate The predicate to transform.
+     * @return A mapping from a {@link Value} and list of corresponding {@link RangeConstraints}.
+     */
+    @Nonnull
+    public static Optional<Multimap<Value, RangeConstraints>> dnfPredicateToRanges(@Nonnull final QueryPredicate predicate) {
```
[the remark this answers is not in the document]
hatyo: Good idea, I made a trait `IndexPredicateExpansion` that is plugged in the `ValueIndexExpansionVisitor`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
hatyo: Memoization done, will create a follow-up ticket to have a unified interface for `OrPredicate`/`AndPredicate` with `ValueWithRanges`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/Value.java
hatyo: restored.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
hatyo: Here is the ticket: #2009

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GraphExpansion.java
```diff
@@ -189,27 +190,34 @@ public Sealed seal() {
...
             final var localPredicates = ImmutableSet.copyOf(getPredicates());
-            final var resultPlaceHolders = Lists.newArrayList(placeholders);
+            final var resultPlaceHolders = Lists.newArrayList(deDupPlaceholders);
             final var localPlaceHolderPairs =
-                    IntStream.range(0, placeholders.size())
-                            .mapToObj(i -> Pair.of(placeholders.get(i), i))
-                            .filter(p -> localPredicates.contains(p.getKey()))
+                    IntStream.range(0, deDupPlaceholders.size())
+                            .mapToObj(i -> Pair.of(deDupPlaceholders.get(i), i))
+                            .filter(p -> localPredicates.stream().anyMatch(predicate -> predicate instanceof ValueWithRanges && ((ValueWithRanges)predicate).getValue().equals(p.getKey().getValue())))
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
hatyo: It is a left-over from a previous iteration ... removed it, thanks for spotting it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
normen662: Thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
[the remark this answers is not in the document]
normen662: Thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
```diff
@@ -48,7 +51,7 @@
  * Class to expand value index access into a candidate graph. The visitation methods are left unchanged from the super
  * class {@link KeyExpressionExpansionVisitor}, this class merely provides a specific {@link #expand} method.
  */
-public class ValueIndexExpansionVisitor extends KeyExpressionExpansionVisitor implements ExpansionVisitor<KeyExpressionExpansionVisitor.VisitorState> {
+public class ValueIndexExpansionVisitor extends KeyExpressionExpansionVisitor implements ExpansionVisitor<KeyExpressionExpansionVisitor.VisitorState>, IndexPredicateExpansion {
```
normen662, line 54: I don't care really one way or another, but if the methods in  `IndexPredicateExpansion` are all _non_selfish, why not make the interface a class with a bunch of statics and just call them from here. The original comment was just about the extraction of the general code that can be applied in the context of any kind of index versus the code that can just be applied for value indexes. TLDNR, we don't have to use a trait mix-in approach. Up to you. Of course all of this is moot if any of the methods are selfish.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
```diff
@@ -98,8 +98,7 @@ default List<MatchedOrderingPart> computeMatchedOrderingParts(@Nonnull MatchInfo
 
             builder.add(
                     MatchedOrderingPart.of(value,
-                            comparisonRange == null ? ComparisonRange.Type.EMPTY : comparisonRange.getRangeType(),
-                            queryPredicate,
+                            comparisonRange == null ? ComparisonRange.Type.EMPTY : comparisonRange.getRangeType(), queryPredicate,
```
normen662, line 101: Didn't you say this was a remnant of some previous iteration?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
```diff
@@ -542,14 +542,14 @@ public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candid
...
-                    unmappedOtherPredicates
-                            .removeIf(queryPredicate -> queryPredicate instanceof Placeholder || queryPredicate.isTautology());
-
-                    if (!unmappedOtherPredicates.isEmpty()) {
+                    // Last chance for unmapped candidate predicates - if there is a placeholder without compile-time range
+                    // or a tautology on the candidate side that is still unmapped, we can (and should) remove it from the
+                    // unmapped other set now. The reasoning is that this predicate is not filtering, so it does not cause
+                    // records to be filtered that are not filtered on the query side.
+                    remainingUnmappedCandidatePredicates
+                            .removeIf(queryPredicate -> queryPredicate.isTautology() || (queryPredicate instanceof ValueWithRanges && ((ValueWithRanges)queryPredicate).getRanges().isEmpty()));
```
normen662, line 550: Didn't you already create a helper for this or something similar?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
```diff
@@ -596,13 +596,17 @@ public String toString() {
         return "SELECT " + resultValue + " WHERE " + AndPredicate.and(getPredicates());
     }
 
-    private static List<? extends QueryPredicate> partitionPredicates(final List<? extends QueryPredicate> predicates) {
+    /**
+     * Breaks a list of predicates into a number of saragbles and value predicates.
```
normen662, line 600:
```suggestion
     * Breaks a list of predicates into a number of sargbles and value predicates.
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/SparseIndexTest.java
```diff
@@ -0,0 +1,189 @@
+/*
+ * SparseIndexConstructionTest.java
```
normen662, line 2:
```suggestion
 * SparseIndexTest.java
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/match/RangeConstraintsTest.java
```diff
@@ -0,0 +1,137 @@
+/*
+ * CompileTimeEvaluableRangeTest.java
```
normen662, line 2:
```suggestion
 * RangeConstraintsTest.java
```

normen662 (changes requested):
Left a bunch of single comments. They are all cosmetics. Please address and I can approve.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/match/RangeConstraintsTest.java
hatyo: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/SparseIndexTest.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ValueWithRanges.java
```diff
@@ -0,0 +1,322 @@
...
+        final var rightDisjunction = other.getRanges();
+        for (final var left : leftDisjunction) {
+            if (!left.isCompileTimeEvaluable() || !left.isEmpty().equals(FALSE) || rightDisjunction.stream().noneMatch(right -> right.implies(left).equals(TRUE))) {
```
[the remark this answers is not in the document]
hatyo: I forgot to reply here, done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
hatyo: hmm ... I just created a small helper method in `ValueWithRanges`:

```
    public boolean isEmpty() {
        return getRanges().isEmpty();
    }
```

are you suggesting we create a helper function for the entire condition? `queryPredicate -> queryPredicate.isTautology() || (queryPredicate instanceof ValueWithRanges && ((ValueWithRanges)queryPredicate).isEmpty())` ?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
hatyo: Not sure what you mean, I reverted the remnant part, the revert left the code styled differently apparently, I can also revert that unless you mean something else?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java
hatyo: That's a good point, thanks. I don't have a strong opinion one way of the other, but I like the _non_selfish analogy. So, I will just refactor it into a class with a bunch of static methods instead.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
normen662: Nevermind. I saw ghosts.

normen662 (approved):

MMcM (approved):

PR #1976, opened 2023-01-25
Resolves #1973: Clearing index data can skip uniqueness violations space on non-unique indexes

### LINKED ISSUE #1973: Clearing index data can skip uniqueness violations space on non-unique indexes
At the moment, the `clearIndexData` method will clear out the uniqueness violations space on all indexes. However, this space is only updated if the index is unique, and it only needs to be read when updating a unique index. We should be able to avoid a few clears and a range scan on non-unique indexes.

alecgrieser:
When clearing out index data, we don't really need to clear out the list of uniqueness violations on non-unique indexes, as that data is only consulted by unique indexes. This skips clearing out that space for those indexes, and it also adjusts the `scan` method to only ask unique indexes for violations.

This resolves #1973.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -20,7 +20,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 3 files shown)


MMcM:
Is there any issue for an index that is being rebuilt due to a metadata version change that removed a uniqueness constraint?

alecgrieser:
> Is there any issue for an index that is being rebuilt due to a metadata version change that removed a uniqueness constraint?

I believe that this PR actually makes the situation for indexes like that better. In particular, under the old code, if you had an index with uniqueness violations, then if you dropped the uniqueness constraint (without changing the `last_modified_version`), any indexes that were in the middle of a build would still have the uniqueness violations in their uniqueness space, and they would therefore still need to be rebuilt just to clear out the uniqueness violations. Under the new code, that space is ignored for non-unique indexes, so the index can be marked as `READABLE` without needing a rebuild. I've added a new test to that affect, and the test fails without the `FDBRecordStore` changes.

There's still some room for improvement, though. (1) The uniqueness violation data for that index will not be cleared out, so it will eat up space. I guess the worst case for that data size is something on the order of the index itself (if every entry contained a uniqueness violation), so not great, but this situation should be fairly rare. (2) The index still needs to be manually marked as readable. One potential fix would be to have `checkVersion` look for indexes in this state and mark them as `READABLE`. It could also clear out the uniqueness violation space for good measure.

ScottDugas (dismissed):

alecgrieser:
Rebase addresses merge conflicts in `docs/ReleaseNotes.md`

ScottDugas (approved):

PR #1979, opened 2023-01-27
avoid in-joins if correlations not used in sargs

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1951,34 +1999,56 @@ public List<QueryComponent> combineNonSargables() {
...
 
+        public Set<String> getSargedInBindings() {
+            return sargedInBindingsSupplier.get();
+        }
+
+        private Set<String> computeSargedInBindings() {
+            return sargedComparisons.stream()
+                    .filter(comparison -> comparison instanceof Comparisons.ParameterComparison)
+                    .map(comparison -> (Comparisons.ParameterComparison)comparison)
+                    .filter(parameterComparison -> Bindings.Internal.IN.isOfType(parameterComparison.getParameter()))
```
alecgrieser, line 2022: Maybe it's too much trouble, but there's part of me that thinks that rather than using the name of the parameter here to compute the number of sarged IN-bindings, the list of comparisons in the `IN` extractor could be compared to the sarged comparisons. The current approach has the unfortunate side-effect that if a user creates a parameter called `__in_blah_blah`, it will be treated as an IN-parameter by this code. I'm not actually sure if we ban such parameters or not

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/InExtractor.java
```diff
@@ -281,6 +242,78 @@ public List<InSource> unionSources() {
...
+
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static InExtractor fromFilter(@Nonnull final QueryComponent filter, @Nonnull Predicate<String> inBindingFilter) {
+        final AtomicInteger bindingIndex = new AtomicInteger();
+        final List<InClause> inClauses = Lists.newArrayList();
+        final QueryComponent subFilter = mapClauses(filter, (withComparison, fields) -> {
+            if (withComparison.getComparison().getType() == Comparisons.Type.IN) {
+                final String bindingName = Bindings.Internal.IN.bindingName(
+                        withComparison.getName() + "__" + bindingIndex.getAndIncrement());
```
alecgrieser, line 257: Are these binding names stable enough? In particular, I think the requirement here is that if `inBindingFilter.test(bindingName)` is `false`, then skipping the logic below must not change any of the in-bindings, if I understand how this is used.

The only alternative I could think of would be to have the `InExtractor` run once to generate the transformed query, and then when we want to turn off certain `IN`-joins, we have to have the `InExtractor` replace the set of clauses with their original comparisons. But in this alternative reality, it doesn't re-run the assignment of clauses to binding names.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -516,12 +521,22 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+                final RecordQueryPlan wrapped = inExtractor.wrap(planContext.rankComparisons.wrap(bestPlan.plan, bestPlan.includedRankComparisons, metaData));
+                final ScoredPlan scoredPlan = new ScoredPlan(bestPlan.score, wrapped);
+                if (needOrdering) {
+                    scoredPlan.planOrderingKey = inExtractor.adjustOrdering(bestPlan.planOrderingKey, false);
+                }
+                return scoredPlan;
+            } else {
+                // create a new in extractor that only uses the in-clauses we were actually able to use
+                final InExtractor restrictedInExtractor = inExtractor.filter(sargedInBindings::contains);
+                // recursively call to re-plan with fewer in-clauses. This is monotonous and therefore must end.
```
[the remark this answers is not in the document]
normen662: Changed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -1571,4 +1572,24 @@ void testInQueryEmptyList() throws Exception {
...
+        RecordQuery query = RecordQuery.newBuilder()
+                .setRecordType("MySimpleRecord")
+                .setFilter(Query.and(
+                        Query.field("num_value_2").greaterThan(42),
+                        Query.field("num_value_3_indexed").in(List.of(1, 3, 5, 7, 9))
+                ))
+                .build();
+
+        RecordQueryPlan plan = planner.plan(query);
+        System.out.println(plan);
```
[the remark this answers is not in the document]
normen662: I hadn't added any proper testing for that commit. It's a little better now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
```diff
@@ -151,7 +151,7 @@ private ScoredPlan planLucene(@Nonnull CandidateScan candidateScan,
         if (hasRewrite) {
             plan = new LuceneHighlightTermsPlan(plan);
         }
-        return new ScoredPlan(plan, filterMask.getUnsatisfiedFilters(), Collections.emptyList(),  11 - filterMask.getUnsatisfiedFilters().size(),
+        return new ScoredPlan(plan, filterMask.getUnsatisfiedFilters(), Collections.emptyList(), Collections.emptySet(),  11 - filterMask.getUnsatisfiedFilters().size(),
```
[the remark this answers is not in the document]
normen662: I added some Lucent planner-specific code to extract the comparisons from such plans as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: We do ban such parameters.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -1314,7 +1349,7 @@ private ScoredPlan planRank(@Nonnull CandidateScan candidateScan,
                 final ScanComparisons scanComparisons = rankComparison.getScanComparisons();
                 final RecordQueryPlan scan = rankScan(candidateScan, filterComparison, scanComparisons);
                 final boolean createsDuplicates = RankComparisons.createsDuplicates(index, indexExpr);
-                return new ScoredPlan(scan, Collections.emptyList(), Collections.emptyList(), 1, createsDuplicates, Collections.singleton(rankComparison));
+                return new ScoredPlan(scan, Collections.emptyList(), Collections.emptyList(), Collections.emptySet(), 1, createsDuplicates, Collections.singleton(rankComparison));
```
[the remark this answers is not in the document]
normen662: There are complications around rank() functionality such as the `__in_rank_XXX` correlation predicate (treating that equi-predicate just like a regular `__in_XXX` will cause the rank functionality code path to break).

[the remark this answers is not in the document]
normen662: I ignore those predicates coming from  rank meaning they will stay in the `Compassion` set.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -604,6 +619,10 @@ public boolean isBetterThanOther(@Nonnull final PlanContext planContext,
             return true;
         }
 
+        if (plan.getSargedInBindings().size() > otherPlan.getSargedInBindings().size()) {
+            return true;
```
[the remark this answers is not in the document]
normen662: Okay.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -513,17 +518,63 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+                if (inBindings.size() == inExtractor.getInBindings().size()) {
+                    // we were unable to make progress
+                    progress = false;
+                    break;
+                }
+                
+                // Continue to re-plan with fewer in-clauses or exit the loop.
+                numReplan ++;
+            } else {
+                return null;
```
MMcM, line 554: Since this is an early exit, I suggest inverting the `if` and getting back some indentation.

It looks like this covers the case we discussed with IN-Union plans due to forced ordering, since that is stable `null` no matter how the filters are changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -513,17 +518,63 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+                    break;
+                }
+
+                // create a new in extractor that only uses the in-clauses we were actually able to use
+                inExtractor = inExtractor.filter((componentWithComparison, inBinding) -> {
+                    if (sargedInBindings.contains(inBinding)) {
+                        return true;
+                    }
+                    // we cannot filter out INs that are defined using more complicated things like "rank(...) IN (...)"
+                    return componentWithComparison instanceof QueryRecordFunctionWithComparison;
```
MMcM, line 542: This predicate can be tightened up to cover that case (transformed version of component is in context's rank comparisons).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -603,7 +654,7 @@ public boolean isBetterThanOther(@Nonnull final PlanContext planContext,
         if (otherPlan == null) {
             return true;
         }
-
+
```
MMcM, line 657:
```suggestion

```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryPlanWithComparisons.java
```diff
@@ -21,19 +21,38 @@
...
 @API(API.Status.EXPERIMENTAL)
 public interface RecordQueryPlanWithComparisons extends RecordQueryPlan {
-    @Nonnull
-    ScanComparisons getComparisons();
 
     default boolean hasComparisons() {
+        return hasScanComparisons();
+    }
+
+    default Set<Comparisons.Comparison> getComparisons() {
```
MMcM, line 43: This method has a rather strange contract. The main issue is that comparisons are associated with a particular position. For equality, their position in the list, for inequality, the number of equalities that proceed. And this just flattens that.

As used by `computeSargedInBindings` and `compareInOperator`, only certain equality comparisons are looked at and their parameter names picked out.

It seems that we need to know for a given plan how it uses any parameters that it depends on.
* Filter plan evaluates expression including the parameter.
* Index plan binds part of the scan range to the parameter.
* Lucene plan builds the search query to include the parameter's value.

An (imagined) prefix string index scan would have both the first and second. In this case, I believe we consider the index part enough to justify its use.

MMcM (changes requested):

docs/ReleaseNotes.md
```diff
@@ -28,7 +28,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** https://github.com/FoundationDB/fdb-record-layer/issues/1985 [(Issue #1985)](https://github.com/FoundationDB/fdb-record-layer/issues/1985)
```
alecgrieser, line 31: Could this include a description instead of the URL? Something like "Avoid planning IN-predicates as IN-joins when the IN predicate is not optimized by an index" may be good to have to make the release notes a bit more readable

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -1325,19 +1328,15 @@ void testInQueryOrDifferentCondition() throws Exception {
...
-                                    inValuesJoinPlan(
-                                            filterPlan(
-                                                    indexPlan().where(indexName("MySimpleRecord$num_value_unique")).and(scanComparisons(range("([990],>"))))
-                                    ).where(inValuesList(equalsObject(Arrays.asList(0, 2)))))));
-            assertEquals(-97067043, plan.planHash(PlanHashable.PlanHashKind.LEGACY));
-            assertEquals(942676960, plan.planHash(PlanHashable.PlanHashKind.FOR_CONTINUATION));
-            assertEquals(417180157, plan.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS));
+                                    filterPlan(
+                                            indexPlan().where(indexName("MySimpleRecord$num_value_unique")).and(scanComparisons(range("([990],>"))))
+                                            .where(queryComponents(only(equalsObject(Query.field("num_value_2").in(Arrays.asList(2, 0))))))));
```
alecgrieser, line 1336: Nice!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -484,6 +492,18 @@ public Builder setPlanOtherAttemptWholeFilter(final boolean planOtherAttemptWhol
...
         }
 
+        /**
+         * Set the maximum number of replanning-attempts during IN-to-JOIN transformations in the planner. This
+         * option only applies to {@link RecordQueryPlanner}.
+         * @param maxNumReplansForInToJoin the number of replanning attempts; defaults to {@code 0} for no
+         *        replanning attempts
+         * @return this builder
+         */
+        public Builder setMaxNumReplansForInToJoin(final int maxNumReplansForInToJoin) {
```
alecgrieser, line 502: Does this new planner configuration property allow for reverting back to the old behavior? I'm pretty sure the old behavior is worse, so I don't think we'd want to go back once this has rolled out, but arguably we might want it for plan stability during roll out.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -841,10 +912,14 @@ private ScoredPlan handleNonSargables(@Nonnull ScoredPlan bestPlan,
             comparisonKey = getKeyForMerge(sort, comparisonKey);
             ScoredPlan intersectionPlan = planIntersection(intersectionCandidates, comparisonKey);
             if (intersectionPlan != null) {
-                if (intersectionPlan.unsatisfiedFilters.isEmpty()) {
-                    return intersectionPlan;
-                } else if (bestPlan.getNumNonSargables() > intersectionPlan.getNumNonSargables()) {
-                    bestPlan = intersectionPlan;
+                // number of sarged in-bindings in intersection plan can only be lower or equal to sarged in-bindings
+                // of best plan
+                if (bestPlan.getSargedInBindings().size() == intersectionPlan.getSargedInBindings().size()) {
```
alecgrieser, line 917: I'm struggling a bit with this assertion ("number of sarged in-bindings in intersection plan can only be lower or equal to sarged in-bindings of best plan"). Is it saying that the only kinds of intersection plans that we expect are those for which the same sets of `IN`-bindings are sargable on all legs, and that the `bestPlan` from before will either be on one of those indexes (and thus have the same number of sargable in bindings) or will be on some other index with more sargable `IN`s?

Even if it is true, do we still want to favor the `intersectionPlan` in this case? This seems somewhat analogous to the previous decision to favor sargable `IN`s over saragable predicates that used to be in this PR but was removed

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryPlanWithComparisons.java
normen662: I view this method as the `getCorrelatedTo()` of sorts in the old planner realm. The old `getComparisons()` has been renamed to `getScanComparisons()`. So the old API is still intact.
There are `Comparisons` inside `ScanComparisons` (inequalities) that are a set already.
I would like to keep the comparisons in a set here and then filter them later on by use case rather than only filtering them here and then have to open up this collector for other use cases in the future.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: In the latest commit I am using the version you sent me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: done

docs/ReleaseNotes.md
normen662: oh a cp went wrong

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
normen662: I think I don't have the confidence to provide a path where you get the exact same plan as before which as far as we know is squeal or word to the new behavior. I can provide yet another configuration that allows for enforcing to return the suboptimal plan after the initial planning knowing full well we create a worse plan. I am not a fan of that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
normen662: The intersection logic for finding the sargables intersects the sets of comparisons for all legs. Since the best plan is among those, the intersection set can only be at most that.

I reverted the code to be in line with the preferences for individual plan comparison i.e. in `isBetterThanOther()`

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryPlanWithComparisons.java
MMcM: Maybe it's harmless noise, but compared to `getCorrelatedTo` concretely, `getComparisons` is full of a bunch of comparisons to literal values. And, to me, the set cardinality difference when you happen to have two `= 1` versus `= 1` and `= 2` or `= 1` and `>= 1` suggests that it isn't really meaningful. I suppose the idea put forward is that this is just part of some API and every particular use case ends up wrapping the result in what it takes as important before doing any reasoning.

MMcM (approved):

PR #1981, opened 2023-01-28
Resolves #1980: Indexing typestamp: add query + flow control

### LINKED ISSUE #1980: Indexing typestamp: add query + flow control
During indexing, allow: 
1. query indexing type stamp 
2. block (pause) indexing - optional: a) string id/description b) expiration in seconds
3. unblock - if id is given, unblock only if matches 
4. indexing policy - allow atomic unblock and index operation (with optional id - to index only if matches)

These operations should also be supported for multiple indexes

jjezra:
Add indexing control flow. Allowing: 
1. Set frequency of checking the indexing stamp (never, always, or minimal time between checks) 
2. Throw an exception of the indexing stamp had changed 
3. Provide a way to query, block, and unblock the indexing stamp
4. Provide an option to unblock during indexing 
5. Provide an option to add a block id during block, that can be used for verification during unblock 
6. Provide a time-to-leave option to the indexing block

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -398,7 +398,7 @@ public class FDBStoreTimer extends StoreTimer {
         /** Wait for an {@link IndexOperation} to complete. */
         WAIT_INDEX_OPERATION("wait for index operation"),
         /** Wait for indexing type stamp operation. */
-        WAIT_INDEX_TYPESTAMP_OP("wait for indexing type stamp operation"),
+        WAIT_INDEX_TYPESTAMP_OPERATION("wait for indexing type stamp operation"),
         /** Wait for adding an index. */
         WAIT_ADD_INDEX("wait for adding an index"),
         /** Wait for dropping an index. */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -397,6 +397,8 @@ public class FDBStoreTimer extends StoreTimer {
         WAIT_REVERSE_DIRECTORY_LOCATE("wait for finding reverse directory location"),
         /** Wait for an {@link IndexOperation} to complete. */
         WAIT_INDEX_OPERATION("wait for index operation"),
+        /** Wait for indexing type stamp operation. */
+        WAIT_INDEX_TYPESTAMP_OP("wait for indexing type stamp operation"),
         /** Wait for adding an index. */
         WAIT_ADD_INDEX("wait for adding an index"),
         /** Wait for dropping an index. */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -55,13 +55,12 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.text.SimpleDateFormat;
+import java.time.DateTimeException;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.IdentityHashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -463,16 +462,13 @@ public abstract class IndexingBase {
     }
 
     private boolean shouldAllowTakeoverContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
+        if (isTypeStampBlocked(savedStamp) && !policy.shouldAllowUnblock(savedStamp.getBlockID())) {
+            return false;
+        }
         if (areSimilar(newStamp, savedStamp)) {
-            return !isTypeStampBlocked(savedStamp) || policy.shouldAllowUnblock(savedStamp.getBlockID());
+            return true;
         }
-        if (policy.shouldAllowTakeoverContinue()) {
-            return (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-                     savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) ||
-                    (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-                     savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
-        }
-        return false;
+        return policy.shouldAllowTakeoverContinue(newStamp.getMethod(), savedStamp.getMethod());
     }
 
     private static boolean areSimilar(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
@@ -1095,9 +1091,11 @@ public abstract class IndexingBase {
                 }
                 long expirationMillis = stamp.getBlockExpireEpochMilliSeconds();
                 if (expirationMillis > 0) {
-                    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-mm-dd_HH:mm:ss", Locale.US);
-                    Date expirationDate = new Date(expirationMillis);
-                    str.append(", blockExpires{").append(sdf.format(expirationDate)).append("}");
+                    try {
+                        str.append(", blockExpires{").append(Instant.ofEpochMilli(expirationMillis)).append("}");
+                    } catch (DateTimeException ignore) {
+                        str.append(", blockExpires{value=").append(expirationMillis).append("}");
+                    }
                 }
             }
             return str.append(")").toString();
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -48,6 +48,7 @@ import com.apple.foundationdb.record.query.plan.synthetic.SyntheticRecordPlanner
 import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.ByteArrayUtil2;
 import com.apple.foundationdb.tuple.Tuple;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -55,6 +56,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.IdentityHashMap;
@@ -97,6 +99,7 @@ public abstract class IndexingBase {
     private StoreTimerSnapshot lastProgressSnapshot = null;
     private boolean forceStampOverwrite = false;
     private final long startingTimeMillis;
+    private long lastTypeStampCheckMillis;
 
     IndexingBase(@Nonnull IndexingCommon common,
                  @Nonnull OnlineIndexer.IndexingPolicy policy) {
@@ -113,6 +116,7 @@ public abstract class IndexingBase {
         IndexState expectedIndexState = isScrubber ? IndexState.READABLE : IndexState.WRITE_ONLY;
         this.throttle = new IndexingThrottle(common, expectedIndexState);
         this.startingTimeMillis = System.currentTimeMillis();
+        this.lastTypeStampCheckMillis = startingTimeMillis;
     }
 
     // helper functions
@@ -450,7 +454,7 @@ public abstract class IndexingBase {
                                         savedStamp, continuedBuild));
                     }
                     // fall down to exception
-                    throw newPartlyBuildException(continuedBuild, savedStamp, indexingTypeStamp, index);
+                    throw newPartlyBuiltException(continuedBuild, savedStamp, indexingTypeStamp, index);
                 });
     }
 
@@ -485,7 +489,7 @@ public abstract class IndexingBase {
                     .toString());
         }
         final IndexBuildProto.IndexBuildIndexingStamp fakeSavedStamp = IndexingByRecords.compileIndexingTypeStamp();
-        throw newPartlyBuildException(true, fakeSavedStamp, indexingTypeStamp, index);
+        throw newPartlyBuiltException(true, fakeSavedStamp, indexingTypeStamp, index);
     }
 
     @Nonnull
@@ -502,7 +506,7 @@ public abstract class IndexingBase {
             return AsyncUtil.DONE;
         }
         // A force overwrite cannot be allowed when partly built
-        throw newPartlyBuildException(continuedBuild, savedStamp, indexingTypeStamp, index);
+        throw newPartlyBuiltException(continuedBuild, savedStamp, indexingTypeStamp, index);
     }
 
     @Nonnull
@@ -561,17 +565,17 @@ public abstract class IndexingBase {
         });
     }
 
-    RecordCoreException newPartlyBuildException(boolean continuedBuild,
+    RecordCoreException newPartlyBuiltException(boolean continuedBuild,
                                                 IndexBuildProto.IndexBuildIndexingStamp savedStamp,
-                                                IndexBuildProto.IndexBuildIndexingStamp indexingTypeStamp,
+                                                IndexBuildProto.IndexBuildIndexingStamp expectedStamp,
                                                 Index index) {
-        return new PartlyBuiltException(savedStamp,
+        return new PartlyBuiltException(savedStamp, expectedStamp, index.getName(),
                 "This index was partly built by another method",
                 LogMessageKeys.INDEX_NAME, index,
                 LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
                 LogMessageKeys.INDEXER_ID, common.getUuid(),
                 LogMessageKeys.CONTINUED_BUILD, continuedBuild,
-                LogMessageKeys.EXPECTED, indexingTypeStamp,
+                LogMessageKeys.EXPECTED, expectedStamp,
                 LogMessageKeys.ACTUAL, savedStamp);
     }
 
@@ -698,76 +702,79 @@ public abstract class IndexingBase {
         AtomicLong recordsScannedCounter = new AtomicLong();
 
         final AtomicReference<RecordCursorResult<T>> nextResult = new AtomicReference<>(null);
-        return AsyncUtil.whileTrue(() -> cursor.onNext().thenCompose(result -> {
-            RecordCursorResult<T> currResult;
-            final boolean isExhausted;
-            if (result.hasNext()) {
-                // has next, process one previous item (if exists)
-                currResult = nextResult.get();
-                nextResult.set(result);
-                if (currResult == null) {
-                    // that was the first item, nothing to process
-                    return AsyncUtil.READY_TRUE;
-                }
-                isExhausted = false;
-            } else {
-                // end of the cursor list
-                timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_COUNT);
-                if (!result.getNoNextReason().isSourceExhausted()) {
-                    nextResultCont.set(nextResult.get());
-                    hasMore.set(true);
-                    return AsyncUtil.READY_FALSE;
-                }
-                // source is exhausted, fall down to handle the last item and return with hasMore=false
-                currResult = nextResult.get();
-                if (currResult == null) {
-                    // there was no data
-                    hasMore.set(false);
-                    return AsyncUtil.READY_FALSE;
-                }
-                // here, process the last item and return
-                nextResult.set(null);
-                isExhausted = true;
-            }
-
-            // here: currResult must have value
-            timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED);
-            recordsScannedCounter.incrementAndGet();
-
-            return getRecordToIndex.apply(store, currResult)
-                    .thenCompose(rec -> {
-                        if (null == rec) {
-                            if (isExhausted) {
-                                hasMore.set(false);
-                                return AsyncUtil.READY_FALSE;
-                            }
-                            return AsyncUtil.READY_TRUE;
-                        }
-                        // This record should be indexed. Add it to the transaction.
-                        if (isIdempotent) {
-                            store.addRecordReadConflict(rec.getPrimaryKey());
-                        }
-                        timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
-
-                        final CompletableFuture<Void> updateMaintainer = updateMaintainerBuilder(store, rec);
-                        if (isExhausted) {
-                            // we've just processed the last item
-                            hasMore.set(false);
-                            return updateMaintainer.thenApply(vignore -> false);
-                        }
-                        return updateMaintainer.thenCompose(vignore ->
-                                context.getApproximateTransactionSize().thenApply(size -> {
-                                    if (size >= common.config.getMaxWriteLimitBytes()) {
-                                        // the transaction becomes too big - stop iterating
-                                        timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_SIZE);
-                                        nextResultCont.set(nextResult.get());
-                                        hasMore.set(true);
-                                        return false;
+        return validateTypeStamp(store)
+                .thenCompose(ignore ->
+                        AsyncUtil.whileTrue(() -> cursor.onNext()
+                                .thenCompose(result -> {
+                                    RecordCursorResult<T> currResult;
+                                    final boolean isExhausted;
+                                    if (result.hasNext()) {
+                                        // has next, process one previous item (if exists)
+                                        currResult = nextResult.get();
+                                        nextResult.set(result);
+                                        if (currResult == null) {
+                                            // that was the first item, nothing to process
+                                            return AsyncUtil.READY_TRUE;
+                                        }
+                                        isExhausted = false;
+                                    } else {
+                                        // end of the cursor list
+                                        timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_COUNT);
+                                        if (!result.getNoNextReason().isSourceExhausted()) {
+                                            nextResultCont.set(nextResult.get());
+                                            hasMore.set(true);
+                                            return AsyncUtil.READY_FALSE;
+                                        }
+                                        // source is exhausted, fall down to handle the last item and return with hasMore=false
+                                        currResult = nextResult.get();
+                                        if (currResult == null) {
+                                            // there was no data
+                                            hasMore.set(false);
+                                            return AsyncUtil.READY_FALSE;
+                                        }
+                                        // here, process the last item and return
+                                        nextResult.set(null);
+                                        isExhausted = true;
                                     }
-                                    return true;
-                                }));
-                    });
-        }), cursor.getExecutor())
+
+                                    // here: currResult must have value
+                                    timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED);
+                                    recordsScannedCounter.incrementAndGet();
+
+                                    return getRecordToIndex.apply(store, currResult)
+                                            .thenCompose(rec -> {
+                                                if (null == rec) {
+                                                    if (isExhausted) {
+                                                        hasMore.set(false);
+                                                        return AsyncUtil.READY_FALSE;
+                                                    }
+                                                    return AsyncUtil.READY_TRUE;
+                                                }
+                                                // This record should be indexed. Add it to the transaction.
+                                                if (isIdempotent) {
+                                                    store.addRecordReadConflict(rec.getPrimaryKey());
+                                                }
+                                                timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED);
+
+                                                final CompletableFuture<Void> updateMaintainer = updateMaintainerBuilder(store, rec);
+                                                if (isExhausted) {
+                                                    // we've just processed the last item
+                                                    hasMore.set(false);
+                                                    return updateMaintainer.thenApply(vignore -> false);
+                                                }
+                                                return updateMaintainer.thenCompose(vignore ->
+                                                        context.getApproximateTransactionSize().thenApply(size -> {
+                                                            if (size >= common.config.getMaxWriteLimitBytes()) {
+                                                                // the transaction becomes too big - stop iterating
+                                                                timerIncrement(timer, FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_SIZE);
+                                                                nextResultCont.set(nextResult.get());
+                                                                hasMore.set(true);
+                                                                return false;
+                                                            }
+                                                            return true;
+                                                        }));
+                                            });
+                                }), cursor.getExecutor()))
                 .thenApply(vignore -> {
                     long recordsScannedInTransaction = recordsScannedCounter.get();
                     if (recordsScanned != null) {
@@ -782,6 +789,50 @@ public abstract class IndexingBase {
                     }
                     return null;
                 });
+    }
+
+    private CompletableFuture<Void> validateTypeStamp(@Nonnull FDBRecordStore store) {
+        final long minimalInterval = policy.getCheckIndexingMethodFrequencyMilliseconds();
+        if (minimalInterval < 0 || isScrubber) {
+            return AsyncUtil.DONE;
+        }
+        if (minimalInterval > 0) {
+            final long now = System.currentTimeMillis();
+            if (now < lastTypeStampCheckMillis + minimalInterval) {
+                return AsyncUtil.DONE;
+            }
+            lastTypeStampCheckMillis = now;
+        }
+        final IndexBuildProto.IndexBuildIndexingStamp expectedTypeStamp = getIndexingTypeStamp(store);
+        return forEachTargetIndex(index ->
+                store.loadIndexingTypeStampAsync(index)
+                        .thenAccept(typeStamp -> validateTypeStamp(typeStamp, expectedTypeStamp, index)));
+    }
+
+
+    private void validateTypeStamp(final IndexBuildProto.IndexBuildIndexingStamp typeStamp,
+                                   final IndexBuildProto.IndexBuildIndexingStamp expectedTypeStamp,
+                                   Index index) {
+        if (typeStamp == null && expectedTypeStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+            // special case - null type stamp is considered a BY_RECORD
+            return;
+        }
+        if (typeStamp == null || typeStamp.getMethod() != expectedTypeStamp.getMethod() || isTypeStampBlocked(typeStamp)) {
+            throw new PartlyBuiltException(typeStamp, expectedTypeStamp,
+                    index.getName(), "Indexing stamp had changed",
+                    LogMessageKeys.INDEX_NAME, index,
+                    LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
+                    LogMessageKeys.INDEXER_ID, common.getUuid(),
+                    LogMessageKeys.EXPECTED, expectedTypeStamp,
+                    LogMessageKeys.ACTUAL, typeStamp);
+        }
+    }
+
+    private static boolean isTypeStampBlocked(final IndexBuildProto.IndexBuildIndexingStamp typeStamp) {
+        return typeStamp.hasBlock() && typeStamp.getBlock() &&
+               (!typeStamp.hasBlockExpireEpochSeconds() ||
+                typeStamp.getBlockExpireEpochSeconds() == 0 ||
+                typeStamp.getBlockExpireEpochSeconds() > (System.currentTimeMillis() / 1000L));
     }
 
     private CompletableFuture<Void> updateMaintainerBuilder(@Nonnull FDBRecordStore store,
@@ -846,7 +897,9 @@ public abstract class IndexingBase {
             // indexes know to update the index on all record saves.
             IndexingRangeSet rangeSet = IndexingRangeSet.forIndexBuild(store, index);
             return rangeSet.insertRangeAsync(null, null);
-        })).thenCompose(vignore -> rebuildIndexInternalAsync(store));
+        }))
+                .thenCompose(vignore -> setIndexingTypeOrThrow(store, false))
+                .thenCompose(vignore -> rebuildIndexInternalAsync(store));
     }
 
     abstract CompletableFuture<Void> rebuildIndexInternalAsync(FDBRecordStore store);
@@ -880,6 +933,53 @@ public abstract class IndexingBase {
         if ( recordMetaDataProvider == null || !metaData.equals(recordMetaDataProvider.getRecordMetaData())) {
             throw new MetaDataException("Store does not have the same metadata");
         }
+    }
+
+    CompletableFuture<Void> indexingStamp(@Nonnull AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
+                                          @Nullable OnlineIndexer.IndexingStampOperation op,
+                                          @Nullable String id,
+                                          @Nullable  Integer timeoutSeconds) {
+        return getRunner().runAsync(context -> openRecordStore(context).thenCompose(store ->
+            forEachTargetIndex(index -> store.loadIndexingTypeStampAsync(index)
+                    .thenApply(stamp -> indexingStamp(oldStamps, store, index, stamp, op, id, timeoutSeconds)))
+        ));
+    }
+
+    boolean indexingStamp(@Nonnull AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
+                          @Nonnull FDBRecordStore store,
+                          @Nonnull Index index,
+                          @Nullable IndexBuildProto.IndexBuildIndexingStamp stamp,
+                          @Nullable OnlineIndexer.IndexingStampOperation op,
+                          @Nullable String id,
+                          @Nullable  Integer timeoutSeconds
+    ) {
+        oldStamps.put(index.getName(), stamp);
+        if (op == null || stamp == null || op.equals(OnlineIndexer.IndexingStampOperation.QUERY)) {
+            return false;
+        }
+
+        IndexBuildProto.IndexBuildIndexingStamp.Builder builder;
+        try {
+            builder = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(stamp.toByteArray()).toBuilder();
+        } catch (InvalidProtocolBufferException e) {
+            return false;
+        }
+
+        if (op == OnlineIndexer.IndexingStampOperation.BLOCK) {
+            builder.setBlock(true);
+            if (id != null) {
+                builder.setBlockDescription(id);
+            }
+            if (timeoutSeconds != null && timeoutSeconds > 0) {
+                builder.setBlockExpireEpochSeconds((int)(System.currentTimeMillis() / 1000) + timeoutSeconds);
+            }
+        }
+        if (op == OnlineIndexer.IndexingStampOperation.UNBLOCK &&
+                (id == null || id.isEmpty() || id.equals(stamp.getBlockDescription()))) {
+            builder.setBlock(false);
+        }
+        store.saveIndexingTypeStamp(index, builder.build());
+        return true;
     }
 
     /**
@@ -930,10 +1030,18 @@ public abstract class IndexingBase {
     @SuppressWarnings("serial")
     public static class  PartlyBuiltException extends RecordCoreException {
         final IndexBuildProto.IndexBuildIndexingStamp savedStamp;
+        final IndexBuildProto.IndexBuildIndexingStamp expectedStamp;
+        final String indexName;
 
-        public PartlyBuiltException(@Nonnull IndexBuildProto.IndexBuildIndexingStamp savedStamp, @Nonnull String msg, @Nullable Object ... keyValues) {
+        public PartlyBuiltException(IndexBuildProto.IndexBuildIndexingStamp savedStamp,
+                                    IndexBuildProto.IndexBuildIndexingStamp expectedStamp,
+                                    String indexName,
+                                    @Nonnull String msg,
+                                    @Nullable Object ... keyValues) {
             super(msg, keyValues);
             this.savedStamp = savedStamp;
+            this.expectedStamp = expectedStamp;
+            this.indexName = indexName;
         }
 
         public IndexBuildProto.IndexBuildIndexingStamp getSavedStamp() {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -56,13 +56,14 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
@@ -459,11 +460,25 @@ public abstract class IndexingBase {
     }
 
     private boolean shouldAllowTakeoverContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
-        return policy.shouldAllowTakeoverContinue() &&
+        return
+            areSimilar(newStamp, savedStamp)
+            ?
+            (!isTypeStampBlocked(savedStamp) || policy.shouldAllowUnblock(savedStamp.getBlockDescription()))
+            :
+            policy.shouldAllowTakeoverContinue() &&
                (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
                 savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) ||
                (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-               savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+                savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+    }
+
+    private static boolean areSimilar(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
+        return savedStamp.getMethod() == newStamp.getMethod()
+               && savedStamp.getTargetIndexList().size() == newStamp.getTargetIndexList().size()
+               && new HashSet<>(savedStamp.getTargetIndexList()).containsAll(newStamp.getTargetIndexList())
+               && (savedStamp.getMethod() != IndexBuildProto.IndexBuildIndexingStamp.Method.BY_INDEX
+                   || (savedStamp.getSourceIndexSubspaceKey().equals(newStamp.getSourceIndexSubspaceKey())
+                       && savedStamp.getSourceIndexLastModifiedVersion() == newStamp.getSourceIndexLastModifiedVersion()));
     }
 
     @Nonnull
@@ -570,6 +585,8 @@ public abstract class IndexingBase {
                                                 IndexBuildProto.IndexBuildIndexingStamp expectedStamp,
                                                 Index index) {
         return new PartlyBuiltException(savedStamp, expectedStamp, index.getName(),
+                savedStamp.hasBlock() && savedStamp.getBlock() ?
+                "This index was partly built, and blocked" :
                 "This index was partly built by another method",
                 LogMessageKeys.INDEX_NAME, index,
                 LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
@@ -935,7 +952,7 @@ public abstract class IndexingBase {
         }
     }
 
-    CompletableFuture<Void> indexingStamp(@Nonnull AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
+    CompletableFuture<Void> indexingStamp(@Nonnull ConcurrentHashMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
                                           @Nullable OnlineIndexer.IndexingStampOperation op,
                                           @Nullable String id,
                                           @Nullable  Integer timeoutSeconds) {
@@ -945,7 +962,7 @@ public abstract class IndexingBase {
         ));
     }
 
-    boolean indexingStamp(@Nonnull AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
+    boolean indexingStamp(@Nonnull ConcurrentHashMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
                           @Nonnull FDBRecordStore store,
                           @Nonnull Index index,
                           @Nullable IndexBuildProto.IndexBuildIndexingStamp stamp,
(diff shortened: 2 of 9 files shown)


jjezra:
test this please

jjezra:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -810,6 +814,18 @@ public CompletableFuture<Boolean> markReadable() {
         return getIndexer().markIndexReadable(true);
     }
 
+    enum IndexingStampOperation {
+        QUERY, BLOCK, UNBLOCK,
```
ScottDugas, line 818: I think it would be better to have separate methods (`queryIndexStamps`, `blockIndexBuilds`, `unblockIndexBuilds`). Specifically for `blockIndexBuilds` and `unblockIndexBuilds`, the fact that it is backed by the indexing stamp could be internal, and doesn't need to be exposed (conceptually) to the user.

That being said, if you feel adamant that these should be an enum as you have it, then:
1. These enum values should have javadoc, specifically important is that `BLOCK` will abort any currently running index builds
2. The method should have a javadoc
3. The method should be very-like, something like `performIndexingStampOperation`
4. The method should document the id. Particularly that it should be unique and provide enough information to identify who and why it was blocked.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1094,7 +1110,7 @@ public static class Builder {
         private IndexingPolicy indexingPolicy = null;
         private IndexingPolicy.Builder indexingPolicyBuilder = null;
         @Nullable
-        private Function<Config, Config> configLoader = null;
+        private UnaryOperator<Config> configLoader = null;
```
ScottDugas, line 1113: This is a superficial change, unless someone is passing in a variable of type `Function<Config, Config>`, correct? I would expect most uses to be a lambda, or method reference, and it's `UNSTABLE`, so I think this makes sense.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -451,16 +456,30 @@ private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boo
...
                 savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) ||
                (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-               savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+                savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+    }
+
+    private static boolean areSimilar(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
+        return savedStamp.getMethod() == newStamp.getMethod()
+               && savedStamp.getTargetIndexList().size() == newStamp.getTargetIndexList().size()
+               && new HashSet<>(savedStamp.getTargetIndexList()).containsAll(newStamp.getTargetIndexList())
```
ScottDugas, line 479: `targetIndex` is supposed to be sorted alphabetically, so you should just be able to do 
```suggestion
               && savedStamp.getTargetIndexList().equals(newStamp.getTargetIndexList())
```

Or are you trying to avoid making that sorting a requirement, and have this code handle the possibility that some code puts it in unsorted.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+
+        Index srcIndex = new Index("src_index", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS);
+        Index tgtIndex = new Index("tgt_index", field("num_value_3_indexed"), IndexTypes.VALUE);
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = myHook(srcIndex, tgtIndex);
+
+        openSimpleMetaData();
+        populateData(numRecords);
+
+        openSimpleMetaData(hook);
+        buildSrcIndex(srcIndex);
```
ScottDugas, line 1132: It would probably make the test faster, and be more representative of the the actual usage if the metadata before populating the data had the source index but not the target index.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+    @Test
+    public void testIndexFromIndexBlock() {
+        // start indexing by Index, verify continuation
+        final FDBStoreTimer timer = new FDBStoreTimer();
+        final int numRecords = 49;
+        final int chunkSize  = 12;
+        final int numChunks  = 1 + (numRecords / chunkSize);
+
+        Index srcIndex = new Index("src_index", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS);
+        Index tgtIndex = new Index("tgt_index", field("num_value_3_indexed"), IndexTypes.VALUE);
```
ScottDugas, line 1125:
```suggestion
        Index targetIndex = new Index("tgt_index", field("num_value_3_indexed"), IndexTypes.VALUE);
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+
+        // query, ensure blocked
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setIndex(tgtIndex)
+                .build()) {
+            final AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> stampMap =
+                    indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.QUERY, null, 0);
+            System.out.println(stampMap);
```
ScottDugas, line 1159: You assert about the values of `stampMap` below, so this seems unnecessary.
```suggestion
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+        // Partly build index
+        openSimpleMetaData(hook);
+        buildIndexAndCrashHalfway(tgtIndex, chunkSize, 1, timer,
+                OnlineIndexer.IndexingPolicy.newBuilder()
+                        .setSourceIndex("src_index")
+                        .forbidRecordScan()
+                        .build());
+
+        openSimpleMetaData(hook);
+        String luka = "my name is Luka";
```
ScottDugas, line 1143: Probably worth giving this a descriptive name like `blockId`.
It might also be good to make this somewhat demonstrative, and use what would be a good blockId in production. such as `Blocked: #1981` or at least `Blocked by Josef`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+            final AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> stampMap =
+                    indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.QUERY, null, 0);
+            System.out.println(stampMap);
+            String indexName = tgtIndex.getName();
+            final IndexBuildProto.IndexBuildIndexingStamp stamp = stampMap.get(indexName);
+            assertEquals(IndexBuildProto.IndexBuildIndexingStamp.Method.BY_INDEX, stamp.getMethod());
+            assertTrue(stamp.getBlock());
+            assertEquals(luka, stamp.getBlockID());
+            assertTrue(stamp.getBlockExpireEpochSeconds() > (System.currentTimeMillis() / 1000));
+            assertTrue(stamp.getBlockExpireEpochSeconds() < 20 + (System.currentTimeMillis() / 1000));
```
ScottDugas, line 1166: You use `/ 1000` in a lot of places to convert from to/from millis. It would be better to use `TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis())` (or it might be better to just have the stamp have millis since epoch, as I commented elsewhere, and avoid all this conversion)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+
+        // Successfully build
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexBuilder = newIndexerBuilder()
+                .addTargetIndex(tgtIndex)
+                .setIndexingPolicy(OnlineIndexer.IndexingPolicy.newBuilder()
+                        .setSourceIndex("src_index")
+                        .forbidRecordScan()
+                        .setAllowUnblock(true)
+                        .build())
```
ScottDugas, line 1192: Not as part of this PR, but this implies to me that it might be good to just have an option to `unblockAndContinue`, which would get any necessary information for the policy from the stamp.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+                .setLimit(chunkSize)
+                .setTimer(timer)
+                .build()) {
+            // now continue building from the last successful range
+            indexBuilder.buildIndex();
+        }
+
+        // counters should demonstrate a continuation to completion
```
ScottDugas, line 1200: Should this also assert that the index is readable. Perhaps also asserting after the previous attempts that it is in WRITE_ONLY.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
```diff
@@ -1110,4 +1112,94 @@ public void testIndexFromIndexRebuild() {
...
+            indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.BLOCK, luka, 10);
+        }
+
+        // query, ensure blocked
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setIndex(tgtIndex)
+                .build()) {
+            final AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> stampMap =
```
ScottDugas, line 1157: Perhaps it would also be good to query before blocking the index build

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -775,4 +777,153 @@ public void benchMarkMultiTarget() {
...
+            indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.BLOCK, luka, 10);
+        }
+
+        // 3. query, ensure blocked
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setTargetIndexes(indexes)
+                .build()) {
+            final AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> stampMap =
```
ScottDugas, line 822: Perhaps it would also be good to query before blocking the index build

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -775,4 +777,153 @@ public void benchMarkMultiTarget() {
...
+            RecordCoreException e = assertThrows(RecordCoreException.class, indexBuilder::buildIndex);
+            assertTrue(e.getMessage().contains("This index was partly built, and blocked"));
+        }
+
+        // 4. sleep
+        try {
+            Thread.sleep(2000);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
```
ScottDugas, line 915: Does the record layer have any mechanisms for mocking out time? If we did, we wouldn't need to put a 2 second sleep in this test.

Otherwise, using millis for the expiration time would allow an expiration of 1 millisecond which wouldn't require so much sleeping here.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -976,4 +978,196 @@ void testMultiTargetMultiType() {
...
+        final FDBStoreTimer timer = new FDBStoreTimer();
+        IntStream.rangeClosed(0, 4).parallel().forEach(id -> {
+            if (id == 3) {
+                FDBRecordStoreTestBase.RecordMetaDataHook localHook = allIndexesHook(indexes);
+                openSimpleMetaData(localHook);
+                try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                        .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                        .setTargetIndexes(indexes)
+                        .build()) {
+                    Thread.sleep(500); // this sleep is required to make sure that indexing had started and the indexes contain type stamps
```
ScottDugas, line 1012: Could you do this with a semaphore instead?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -976,4 +978,196 @@ void testMultiTargetMultiType() {
...
+                    indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.BLOCK, null, 0);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            } else {
+                oneThreadIndexingCrashHalfway(indexes, timer, boundariesList, 1);
+            }
+        });
```
ScottDugas, line 1020: It seems like we should assert here that none of the indexes are built.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -976,4 +978,196 @@ void testMultiTargetMultiType() {
         });
         validateIndexes(indexes);
     }
+
+    @Test
+    void testMutualIndexingBlocker() {
```
ScottDugas, line 983: Each type of index build gets a different set of tests. They're all testing the block logic, which should mostly be orthogonal to the type of index build.
The code doesn't have the right abstractions so that we can test this independently from the type of index build, and changing it, i think, would be a fairly sizeable undertaking, that we shouldn't do now.

That being said, perhaps it would make sense to create an abstract test class for this, with some overrides for setting up and doing the different build attempts that could test all the behavior you have tests for and run it for all the different types of index builds.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -976,4 +978,196 @@ void testMultiTargetMultiType() {
...
+        }
+
+        // Unblock, the return value is the old stamp - validate it
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setTargetIndexes(indexes)
+                .build()) {
+            final AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> stampMap =
+                    indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.UNBLOCK, null, 0);
```
ScottDugas, line 1061: I didn't notice, but it seems like there should be a test of trying to unblock with the wrong id.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -775,4 +777,153 @@ public void benchMarkMultiTarget() {
...
+
+        // 2. Block indexing
+        FDBRecordStoreTestBase.RecordMetaDataHook localHook = allIndexesHook(indexes);
+        openSimpleMetaData(localHook);
+        String luka = "my name is Luka";
+        try (OnlineIndexer indexer = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setTargetIndexes(indexes)
+                .build()) {
+            indexer.indexingStamp(OnlineIndexer.IndexingStampOperation.BLOCK, luka, 2);
```
ScottDugas, line 897: I think it's possible; would it make sense to test the possibility where some indexes are blocked, but others are not? I think that might have different behavior for multi-target vs mutual indexing.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -775,4 +777,153 @@ public void benchMarkMultiTarget() {
...
+
+        openSimpleMetaData();
+        populateData(numRecords);
+
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        // 1. partly build multi
+        buildIndexAndCrashHalfway(chunkSize, 2, timer, newIndexerBuilder()
```
ScottDugas, line 802: Is there any way to test aborting a build?
I think in order to do that, record-layer would need infrastructure to have longer-running tests, but if you know of a way that would be great. Perhaps it could be done with the config loader?

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -382,6 +382,8 @@ public enum Waits implements Wait {
         WAIT_REVERSE_DIRECTORY_LOCATE("wait for finding reverse directory location"),
         /** Wait for an {@link IndexOperation} to complete. */
         WAIT_INDEX_OPERATION("wait for index operation"),
+        /** Wait for indexing type stamp operation. */
+        WAIT_INDEX_TYPESTAMP_OP("wait for indexing type stamp operation"),
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -451,16 +456,30 @@ private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boo
...
                     // fall down to exception
-                    throw newPartlyBuildException(continuedBuild, savedStamp, indexingTypeStamp, index);
+                    throw newPartlyBuiltException(continuedBuild, savedStamp, indexingTypeStamp, index);
                 });
     }
 
     private boolean shouldAllowTakeoverContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
-        return policy.shouldAllowTakeoverContinue() &&
+        return
+            areSimilar(newStamp, savedStamp)
```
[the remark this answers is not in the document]
jjezra: My intention was to declare timestamps as similar if their indexing methods (including targets and sources) are the same. However, their control flow fields (block, etc.) may differ. Should the function be named `areTheSame`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
                 });
     }
 
+    private CompletableFuture<Void> validateTypeStamp(@Nonnull FDBRecordStore store) {
+        final long minimalInterval = policy.getCheckIndexingMethodFrequencyMilliseconds();
+        if (minimalInterval < 0 || isScrubber) {
```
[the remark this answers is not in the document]
jjezra: Nope. The scrubber has its internal timestamp for code validation, but it never publishes it in the DB. 
The reasoning, IIRC, was that scrubbing is only allowed on readable indexes - hence it is not exclusive and all possible scrubbing types may run concurrently.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+                return AsyncUtil.DONE;
+            }
+            lastTypeStampCheckMillis = now;
+        }
+        final IndexBuildProto.IndexBuildIndexingStamp expectedTypeStamp = getIndexingTypeStamp(store);
+        return forEachTargetIndex(index ->
+                store.loadIndexingTypeStampAsync(index)
+                        .thenAccept(typeStamp -> validateTypeStamp(typeStamp, expectedTypeStamp, index)));
+    }
+
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+    private void validateTypeStamp(final IndexBuildProto.IndexBuildIndexingStamp typeStamp,
+                                   final IndexBuildProto.IndexBuildIndexingStamp expectedTypeStamp,
+                                   Index index) {
+        if (typeStamp == null && expectedTypeStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+            // special case - null type stamp is considered a BY_RECORD
+            return;
+        }
+        if (typeStamp == null || typeStamp.getMethod() != expectedTypeStamp.getMethod() || isTypeStampBlocked(typeStamp)) {
+            throw new PartlyBuiltException(typeStamp, expectedTypeStamp,
```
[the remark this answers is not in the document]
jjezra: Will `PartlyBuiltException.isBlocked()` be enough, or would rather make it a whole different exception? 

Callers can use something like:
```
.. .catch (PartlyBuiltException ex) { 
  if (! ex.isBlocked()) {
    method = ex..getSavedStamp().getMethod();
    targetIndexes = ex..getSavedStamp()..getTargetIndexList();
 .. do something with it ..
 } 
```
(I feel that blocked state means partly built, but continuation is not allowed.  Hence the `PartlyBuiltException`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+                    index.getName(), "Indexing stamp had changed",
+                    LogMessageKeys.INDEX_NAME, index,
+                    LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
+                    LogMessageKeys.INDEXER_ID, common.getUuid(),
+                    LogMessageKeys.EXPECTED, expectedTypeStamp,
```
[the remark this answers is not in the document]
jjezra: I'm not sure - we can expose the timestamps with `getSavedStamp` and `getExpectedStamp`, but exposing each of their fields will be hairy and would require constant maintenance for every future Stamp changes (+ the need to translate the `method` to another enum to avoid returning a protobuf value)
How about: 
```
ex.isBlocked()
ex.getSavedStamp()
ex.getExpectedStamp() 
```
(We currently only have the second one)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+                                   final IndexBuildProto.IndexBuildIndexingStamp expectedTypeStamp,
+                                   Index index) {
+        if (typeStamp == null && expectedTypeStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+            // special case - null type stamp is considered a BY_RECORD
+            return;
+        }
+        if (typeStamp == null || typeStamp.getMethod() != expectedTypeStamp.getMethod() || isTypeStampBlocked(typeStamp)) {
+            throw new PartlyBuiltException(typeStamp, expectedTypeStamp,
+                    index.getName(), "Indexing stamp had changed",
+                    LogMessageKeys.INDEX_NAME, index,
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+    }
+
+    boolean indexingStamp(@Nonnull ConcurrentHashMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
+                          @Nonnull FDBRecordStore store,
+                          @Nonnull Index index,
+                          @Nullable IndexBuildProto.IndexBuildIndexingStamp stamp,
+                          @Nullable OnlineIndexer.IndexingStampOperation op,
+                          @Nullable String id,
+                          @Nullable  Integer timeoutSeconds
+    ) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+        oldStamps.put(index.getName(), stamp != null ? stamp :
+                                       IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                                               .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.INVALID)
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+                                               .build());
+        if (op == null || stamp == null || op.equals(OnlineIndexer.IndexingStampOperation.QUERY)) {
+            return false;
+        }
+
+        IndexBuildProto.IndexBuildIndexingStamp.Builder builder;
+        try {
+            builder = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(stamp.toByteArray()).toBuilder();
```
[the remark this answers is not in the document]
jjezra: ✅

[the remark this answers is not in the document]
jjezra: (Seems like a late night coding syndrome)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+            return false;
+        }
+
+        if (op == OnlineIndexer.IndexingStampOperation.BLOCK) {
+            builder.setBlock(true);
+            if (id != null) {
+                builder.setBlockID(id);
+            }
+            if (timeoutSeconds != null && timeoutSeconds > 0) {
+                builder.setBlockExpireEpochSeconds((int)(System.currentTimeMillis() / 1000) + timeoutSeconds);
```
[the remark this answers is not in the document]
jjezra: 🤷‍♂️ 
Seconds since epoch in 31 bits seemed reasonable for this expiration. 
Changed it (if no other reason, for the sake of Jan 19, 2038).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+            }
+        }
+        if (op == OnlineIndexer.IndexingStampOperation.UNBLOCK &&
+                (id == null || id.isEmpty() || id.equals(stamp.getBlockID()))) {
+            builder.setBlock(false);
+        }
+        store.saveIndexingTypeStamp(index, builder.build());
```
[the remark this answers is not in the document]
jjezra: Yes. 
It can be especially useful as a way to clear halt mutual indexing, then continue with fewer workers.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+            builder.setBlock(true);
+            if (id != null) {
+                builder.setBlockID(id);
+            }
+            if (timeoutSeconds != null && timeoutSeconds > 0) {
+                builder.setBlockExpireEpochSeconds((int)(System.currentTimeMillis() / 1000) + timeoutSeconds);
+            }
+        }
+        if (op == OnlineIndexer.IndexingStampOperation.UNBLOCK &&
+                (id == null || id.isEmpty() || id.equals(stamp.getBlockID()))) {
```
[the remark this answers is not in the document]
jjezra: If the users wish, they can always use real id in their side of the API. I'm not sure that we should enforce it at the first experimental version (but maybe later, if it seems right..).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+        IndexBuildProto.IndexBuildIndexingStamp.Builder builder;
+        try {
+            builder = IndexBuildProto.IndexBuildIndexingStamp.parseFrom(stamp.toByteArray()).toBuilder();
+        } catch (InvalidProtocolBufferException e) {
+            return false;
+        }
+
+        if (op == OnlineIndexer.IndexingStampOperation.BLOCK) {
+            builder.setBlock(true);
+            if (id != null) {
```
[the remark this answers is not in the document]
jjezra: 🤷 
I really don't know. This is the first version - I tried to keep everything possible. But sure, we can enforce id to avoid having a legacy of `block=true, blockId=<undef>`. 
Note that the id may still be an empty string.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: How about: 
```
indexingSessionQuery() 
indexingSessionBlock(@Nullable id, @Nullable ttlSeconds) 
indexingSessionUnblock(@Nullable id)
```
?

jjezra: Converted to the functions above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I believe so. 
https://stackoverflow.com/questions/54441751/use-of-unaryoperator-and-binaryoperator-when-i-have-function-in-java-8

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: I admit that I hadn't thought about it. But since the code runs rarely, maybe it's better pay the performance penalty and assume nothing about the sorting.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -885,6 +954,56 @@ protected void validateSameMetadataOrThrow(FDBRecordStore store) {
...
+                          @Nullable OnlineIndexer.IndexingStampOperation op,
+                          @Nullable String id,
+                          @Nullable  Integer timeoutSeconds
+    ) {
+        oldStamps.put(index.getName(), stamp != null ? stamp :
+                                       IndexBuildProto.IndexBuildIndexingStamp.newBuilder()
+                                               .setMethod(IndexBuildProto.IndexBuildIndexingStamp.Method.INVALID)
+                                               .build());
+        if (op == null || stamp == null || op.equals(OnlineIndexer.IndexingStampOperation.QUERY)) {
+            return false;
```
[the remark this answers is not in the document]
alecgrieser: Even if you wanted to collect the results in parallel, it is probably better to return the map rather than have it as an argument, especially into the `indexingStamp` method that creates the transaction. If you want it to run in parallel, you can still do that by constructing the map within the `runAsync` lambda and then either passing it to this function or having this function return just the stamp for the index it works on, and then the results can be collected in one spot

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
alecgrieser: Is the information there (whatever is used to "stop mutual indexing") dropped from the type stamp in the `BLOCK` or `UNBLOCK` case? I feel like you'd want to get the value of the type stamp from after the operation has applied on these non-read only cases, but I could be wrong.

There's also some weirdness here with transaction retry loops: if the first transaction updates the stamp, and then commits, then if it hits `commit_unknown_result`, it can see the new value on retry. So, the exact thing that this method returns is a little more unpredictable than you'd like.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: Why `indexingSession` instead of `indexingStamp`? I feel like I prefer `queryIndexingStamps` better. I can kind of see `blockIndexingSession`, but `blockIndexBuilds` and `unblockIndexBuilds` seems to be more indicative of what these methods do from a semantic sense (whereas something like `indexingStampBlock` is clearer on the mechanics (setting a flag in the indexing stamp), but it doesn't necessarily inform the user why they might want to call this method).

I mostly feel strong that the verbs should come first in those names, especially `block` and `unblock`, as that highlights more clearly that these are operations that have side-effects

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: Sound like a good idea!
However, calling indexer without specifying which index to build might be better performed by a layer on top of `OnlineIndexer`. Maybe `StoreIndexer`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -2429,14 +2492,37 @@ public Builder setMutualIndexingBoundaries(final List<Tuple> primaryKeysBoundari
                 return this;
             }
 
+            /**
+             * If the index is partly built and blocked, allowed (or disallow) unblocking before indexing.
+             * @param allowUnblock if true, unblock (if blocked) and continue.
+             * @param allowUnblockId if non-null and non-empty, unblock only if the indexing stamp's id matches it. Will not continue if not blocked.
```
alecgrieser, line 2498: What happens if `allowUnblock = false` and `allowUnblockId` is not `null`?

It's also a little weird to me that if `allowUnblockId` is not `null`, then it won't build on an unblocked index build, but if `allowUnblockId` is `null`, it will. But on the other hand, that doesn't necessarily seem to be what `shouldAllowUnblock` is doing

alecgrieser (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: What's the time mocking mechanism?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: That makes sense. One of the things in my todo list is to restructure the tests derived from `OnlineIndexerTest` and share common code in the base module.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: Added `testMutualIndexingBlockFewIndexingUnblock`

jjezra: The behavior is the same. Since both allow a `BY_RECORD` takevoer there is a value in blocking few indexes but not all.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: Added `testMutualIndexingBlockerOnTheFly`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -445,16 +450,30 @@ private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boo
...
+                savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+    }
+
+    private static boolean areSimilar(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
+        return savedStamp.getMethod() == newStamp.getMethod()
+               && savedStamp.getTargetIndexList().size() == newStamp.getTargetIndexList().size()
+               && new HashSet<>(savedStamp.getTargetIndexList()).containsAll(newStamp.getTargetIndexList())
+               && (savedStamp.getMethod() != IndexBuildProto.IndexBuildIndexingStamp.Method.BY_INDEX
+                   || (savedStamp.getSourceIndexSubspaceKey().equals(newStamp.getSourceIndexSubspaceKey())
+                       && savedStamp.getSourceIndexLastModifiedVersion() == newStamp.getSourceIndexLastModifiedVersion()));
```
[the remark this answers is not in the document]
jjezra: Is there any way to make a conditional checks? Source index should only be checked for BY_INDEX stamps.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -451,16 +456,30 @@ private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boo
...
                     }
                     // fall down to exception
-                    throw newPartlyBuildException(continuedBuild, savedStamp, indexingTypeStamp, index);
+                    throw newPartlyBuiltException(continuedBuild, savedStamp, indexingTypeStamp, index);
                 });
     }
 
     private boolean shouldAllowTakeoverContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
-        return policy.shouldAllowTakeoverContinue() &&
+        return
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+        if (minimalInterval < 0 || isScrubber) {
+            return AsyncUtil.DONE;
+        }
+        if (minimalInterval > 0) {
+            final long now = System.currentTimeMillis();
+            if (now < lastTypeStampCheckMillis + minimalInterval) {
+                return AsyncUtil.DONE;
+            }
+            lastTypeStampCheckMillis = now;
+        }
```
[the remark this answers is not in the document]
jjezra: Since blocking is not expected to be immediate or atomic, it seems reasonable to let the caller choose the desire frequency - every transaction, clock, or never (zero, positive, or negative numbers respectively) according to his needs. 
Since we were struggling with indexing performance recently, it seems reasonable to set the default to a non-disruptive 1 minute wait between checks. `SynchronizedSession`, for comparison, has a default of 10 seconds before expiring. 
note that `SynchronizedSession` is only applied by the runner and only to the primary index. This check is performed for every target index.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -786,6 +810,50 @@ protected  <T> CompletableFuture<Void> iterateRangeOnly(@Nonnull FDBRecordStore
...
+                    LogMessageKeys.INDEXER_ID, common.getUuid(),
+                    LogMessageKeys.EXPECTED, expectedTypeStamp,
+                    LogMessageKeys.ACTUAL, typeStamp);
+        }
+    }
+
+    private static boolean isTypeStampBlocked(final IndexBuildProto.IndexBuildIndexingStamp typeStamp) {
+        return typeStamp.hasBlock() && typeStamp.getBlock() &&
+               (!typeStamp.hasBlockExpireEpochSeconds() ||
+                typeStamp.getBlockExpireEpochSeconds() == 0 ||
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
alecgrieser: I see. I think that's fine, as long as this isn't critical for correctness, though the exact semantics were a bit unclear to me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Added a custom toString.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: Are there retries in this path?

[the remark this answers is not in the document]
alecgrieser: Yes, because this is called within a lambda that gets passed to `getRunner().runAsync`, which will then retry the inner operation

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: I changed to your suggestion, but wouldn't `blockIndexingContinuation` be more accurate?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: > what happens if
Unblocking is not allowed, so blockId is meaningless. 
Yet, the API had changed and the users cannot call this combination.

jjezra:
Rebased..

alecgrieser (changes requested):
There are a few things that I've suggested here, but I think the most important one before merging is changing the return type from `AbstractMap` to `Map`. Everything else, if we want it in today, could be changed later, I think, in a way that is opaque to users

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: The way I see it, blocking prevents continuing an index build (for future and ongoing indexing processes). 
`.setAllowUnblock(true)` does not ignore the block but unblocks and continues indexing in one "atomic" transaction. The motivation was to prevent other indexing processes (possibly with undesired arguments) from taking over the indexing task between the `unblock` and `buildIndex` user calls. 
But it's not a big deal. I can live with both names.

alecgrieser: Ah, I see. I think there's maybe something missing, then, because I could see a scenario where we, for example, want to run an index build (maybe using one specific set of parameters), but we don't want other builds to try and pre-empt it. In that case, I'd expect us to set the `block` on the index, and then configure the one-specific index builder to ignore the block, but to leave it in place. Is that not supported with the current API, then?

jjezra: No, it is not. But we can always use `setUseSynchronizedSession` to ensure a single indexing process. 
(But not a handful of mutual ones...)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -944,14 +947,14 @@ enum IndexingStampOperation {
         QUERY, BLOCK, UNBLOCK,
     }
 
-    CompletableFuture<Void> indexingStamp(@Nonnull ConcurrentHashMap<String, IndexBuildProto.IndexBuildIndexingStamp> oldStamps,
-                                          @Nullable IndexingStampOperation op,
-                                          @Nullable String id,
-                                          @Nullable Long ttlSeconds) {
+    CompletableFuture<AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp>> indexingStamp(@Nullable IndexingStampOperation op,
```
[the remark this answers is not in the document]
jjezra: ✅ 
and 
✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -810,6 +813,46 @@ public CompletableFuture<Boolean> markReadable() {
...
     }
 
+    /**
+     * Query the current indexing session.
+     * Note that this query cannot tell if another process is actively indexing now, but whether or not an index is in
+     * the process of being built (and how).
+     * @return a map of target indexes and their "indexing stamps".
+     */
+    @API(API.Status.EXPERIMENTAL)
+    AbstractMap<String, IndexBuildProto.IndexBuildIndexingStamp> queryIndexingStamps() {
```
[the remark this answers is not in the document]
jjezra: ✅

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -930,15 +1040,72 @@ public static boolean isTimeLimitException(@Nullable Throwable ex) {
...
+                str.append(", blocked");
+                String id = stamp.getBlockID();
+                if (id != null && !id.isEmpty()) {
+                    str.append(", blockId{").append(id).append("} ");
+                }
+                long expirationMillis = stamp.getBlockExpireEpochMilliSeconds();
+                if (expirationMillis > 0) {
+                    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-mm-dd_HH:mm:ss", Locale.US);
+                    Date expirationDate = new Date(expirationMillis);
+                    str.append(", blockExpires{").append(sdf.format(expirationDate)).append("}");
```
ScottDugas, line 1100: IIUC Date drops the time component, so you won't get the time, which seems important here. 
```suggestion
                    str.append(", blockExpires{").append(Instant.ofEpochMilli(expirationMillis)).append("}");
```
(that will format in ISO-8601, which is in UTC, which seems most appropriate)

Or you could just have the millis since epoch here, and require users to parse it on the logging side which might also allow them to pick their own time zone.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerIndexFromIndexTest.java
ScottDugas: Yeah, adding more to `OnlineIndexer` might be excessive, and perhaps something around, or utilities for doing common operations around building indexes would probably be useful. There's been a lot of moderately organic growth in the indexing part of fdb-record-layer, so perhaps revisiting the architecture could result in some better, and hopefully more testable abstractions.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -976,4 +980,449 @@ void testMultiTargetMultiType() {
...
+        }
+
+        // Attempt continue while blocked with the wrong id, ensure failure
+        openSimpleMetaData(hook);
+        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                .setDatabase(fdb).setMetaData(metaData).setSubspace(subspace)
+                .setTargetIndexes(indexes)
+                .setIndexingPolicy(OnlineIndexer.IndexingPolicy.newBuilder()
+                        .setMutualIndexingBoundaries(boundariesList)
+                        .setAllowUnblock(true, "Blocked by Leonardo")
```
ScottDugas, line 1176: This tests trying to start an index build with the wrong block id,  but it would probably be good to call `unblockIndexBuilds` with the wrong id and make sure that doesn't work, and gives back right block info.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -450,16 +458,34 @@ private CompletableFuture<Void> setIndexingTypeOrThrow(FDBRecordStore store, boo
...
     private boolean shouldAllowTakeoverContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp, IndexBuildProto.IndexBuildIndexingStamp savedStamp) {
-        return policy.shouldAllowTakeoverContinue() &&
-               (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-                savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) ||
-               (newStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-               savedStamp.getMethod() == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS);
+        if (areSimilar(newStamp, savedStamp)) {
+            return !isTypeStampBlocked(savedStamp) || policy.shouldAllowUnblock(savedStamp.getBlockID());
+        }
+        if (policy.shouldAllowTakeoverContinue()) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: ✅

jjezra: I had a dream, where all timestamps's time zones are equal. And all users expect them to be UTC.

ScottDugas (approved):

PR #1987, opened 2023-01-31
Add RecordCursor.isClosed()

arnaud-lacurie:


MMcM:
This seems to be want to have the invariant that after calling `cursor.close()`, `cursor.isClosed()` is `true` and maybe other things aren't possible. But it does it as a best effort only (in some cases going to great lengths, such as checking for deleted files from the sorter). In some cases it isn't possible.

It seems better, if we want such a contract, to require it explicitly. In most cases, that would mean adding another state member to the cursor instance to track calls to `close` and return that.

MMcM:
> other things aren't possible

A good example would be that `onNext` returns an error. While such a check can never be perfect in the face of asynchronous work, a bit of care in implementing the check can improve it. For instance, a map-like cursor could check _after_ calling `inner.onNext`, so that if the close happens while that is pending there is no further propagation through.

MMcM (approved):

Issue #1989, opened 2023-01-31
ThenKeyExpression#validateHotSpots


jleach4:
Small nit...

PR #1990, opened 2023-01-31
ThenKeyExpression#validate HotSpots #1989

jleach4:
Small nit, saw a bit of actively in a profiled application.


MMcM:
I suspect most callers don't care about the return value. So, if this is getting hit too often, perhaps validating and getting field references need to be separated.

It may also be that validating is happening too often. I think we probably do want to do it out of a long-term meta-data store. But perhaps not out of relative hot caches.

PR #1992, opened 2023-02-01
Upgrade to Gradle 7.6

MMcM:
There is a good chance that this will OOM, based on https://github.com/gradle/gradle/issues/23215. This is just to confirm one way or the other.


MMcM:
Yes, it does
```
Execution failed for task ':fdb-record-layer-core:checkstyleMandatory'.
> A failure occurred while executing org.gradle.api.plugins.quality.internal.CheckstyleAction
   > An unexpected error occurred configuring and executing Checkstyle.
      > java.lang.Error: Error was thrown while processing /codebuild/output/src742659734/src/github.com/FoundationDB/fdb-record-layer/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
```

PR #1994, opened 2023-02-02
Resolves #1993: Add store timer metrics for retry and indexing delays

### LINKED ISSUE #1993: Add store timer metrics for retry and indexing delays
We should record in the store timer how much time is being spent during deliberate delays (for example, during retry delays and in the indexer, which injects delays to avoid overwhelming FDB server). This is useful during performance investigations to isolate how much time is being spent doing work versus how much time is spent doing deliberately nothing

alecgrieser:
This adds metrics recording the amount of time that is spent during retries and indexing during delays. There was already a `WAIT_RETRY_DELAY` metric, but nothing that captured the event in asynchronous contexts, so I added `Event`s for `RETRY_DELAY` and `INDEXER_DELAY`, and then added instrumentation for those hooks. I also looked into testing, but testing that these events are hit is actually kind of difficult. I tried to add a test to `FDBDatabaseRunnerTest` that asserted on the metrics, but it was difficult to write one that was both reliable and actually asserted on anything, because the values for the counters can be legitimately zero if there isn't any actual delaying going on. Giving that this is just some metrics, the most important thing would seem to be making sure that it doesn't cause NPEs, which I think our tests already cover in that they set the `timer` to `null` in some runs.

This resolves #1993.


jjezra (approved):
Nice!

ScottDugas (approved):

PR #2003, opened 2023-02-09
Fixes #2002: "EMPTY_SCANS" metric double counts empty ranges if hasNext() called multiple times

### LINKED ISSUE #2002: "EMPTY_SCANS" metric double counts empty ranges if hasNext() called multiple times
At the moment, the `EMPTY_SCANS` metric can double count empty scans if `hasNext()` is called multiple times, as this method will increment the counter every time `hasNext` is called if the iterator has not returned any results:

https://github.com/FoundationDB/fdb-record-layer/blob/98251cea702e8f08528e8719c4a24aa2b62dcabc/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java#L390-L396

This can lead to the metric appearing to exceed, for example, the `RANGE_READS` metric, which can give the unexpected result of the number of range reads returning no results appearing to exceed the total number of range reads.

There's another way in which the metric may show that same unintuitive result: the `RANGE_READS` and `READS` metric are incremented whenever `getRange` is called, but that returns an `AsyncIterable`, which doesn't actually start reading until the user calls `.asList()` or `.iterator()`. So, if someone creates one (1) `AsyncIterable` and then calls `.asList()` or `.iterator()` multiple times, then `EMPTY_SCANS` might likewise exceed `RANGE_READS`.

alecgrieser:
At the end of a range scan, it is legal to call `hasNext()` mutliple times. The old code would allow `hasNext()` to be called multiple times, and then it would increment the `EMPTY_SCANS` counter each time, which could result in the metric appearing the exceed the `RANGE_READS` metric. This also fixes another way in which `RANGE_READS` could be undercounted, as it is legal to start multiple `AsyncIterator`s from one `AsyncIterable`, though I don't think we actually do that within the Record Layer. Those two fixes together should ensure that `EMPTY_SCANS` is never higher than `RANGE_READS`.

This fixes #2002.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -19,7 +19,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** The `EMPTY_SCANS` metric is no double counts scans that call `hasNext` multiple times after completing an empty range scan [(Issue #2002)](https://github.com/FoundationDB/fdb-record-layer/issues/2002)
 * **Performance** Additional instrumentaiton of the `RangeSet` is added to account for time spent during index builds  [(Issue #1995)](https://github.com/FoundationDB/fdb-record-layer/issues/1995)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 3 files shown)


docs/ReleaseNotes.md
```diff
@@ -19,7 +19,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** The `EMPTY_SCANS` metric is no double counts scans that call `hasNext` multiple times after completing an empty range scan [(Issue #2002)](https://github.com/FoundationDB/fdb-record-layer/issues/2002)
```
ScottDugas, line 22: This doesn't seem to have proper grammar.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/InstrumentedReadTransaction.java
```diff
@@ -112,86 +113,61 @@ public CompletableFuture<byte[]> getKey(KeySelector keySelector) {
...
 
     @Override
     public AsyncIterable<KeyValue> getRange(Range range, int limit, boolean reverse) {
-        increment(FDBStoreTimer.Counts.READS);
-        increment(FDBStoreTimer.Counts.RANGE_READS);
         return new ByteCountingAsyncIterable<>(underlying.getRange(checkKey(range), limit, reverse));
     }
 
     @Override
     public AsyncIterable<KeyValue> getRange(Range range, int limit, boolean reverse, StreamingMode streamingMode) {
```
ScottDugas, line 170: Should `getMappedRange` also be updated

ScottDugas (changes requested):

docs/ReleaseNotes.md
alecgrieser: Whoops, yeah, I got two different versions of this comment jumbled as I was writing it.

ScottDugas (approved):

Issue #1853, opened 2022-09-28
Mutual concurrent indexing


jjezra:
Allow mutual concurrent multi target indexing

PR #2005, opened 2023-02-14
Cont. #1853: Mutual concurrent indexing - null pointer bug fix

jjezra:
The bug is in the `IndexingMutuallyByRecords.java` module, using `null` as `additionalLogMessageKeyValues`. But creating a general solution seems to be a good idea.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -593,7 +593,7 @@ public abstract class IndexingBase {
                     LogMessageKeys.LIMIT, limit,
                     LogMessageKeys.DELAY, toWait,
                     LogMessageKeys.RECORDS_PER_SECOND, recordsPerSecond)
-                    .addKeysAndValues(additionalLogMessageKeyValues)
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -593,7 +593,7 @@ protected CompletableFuture<Boolean> throttleDelayAndMaybeLogProgress(SubspacePr
                     LogMessageKeys.LIMIT, limit,
                     LogMessageKeys.DELAY, toWait,
                     LogMessageKeys.RECORDS_PER_SECOND, recordsPerSecond)
-                    .addKeysAndValues(additionalLogMessageKeyValues)
+                    .addKeysAndValues(additionalLogMessageKeyValues != null ? additionalLogMessageKeyValues : Collections.emptyList())
```
ScottDugas, line 596: This should probably have a new issue, and an update to the release notes.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: ✅

ScottDugas (approved):

PR #2016, opened 2023-02-20
Fixes #2015: ValueIndexExpansionVisitor does not handle when primary key subsumed by index key

### LINKED ISSUE #2015: ValueIndexExpansionVisitor does not handle when primary key subsumed by index key
The `ValueIndexExpansionVisitor` has logic to add in the primary key fields to the index's root expression to produce a wider key that contains both the indexed fields and the primary key fields:

https://github.com/FoundationDB/fdb-record-layer/blob/5ad62bacf7b390e64ded1d9f3c5bea926c49ac7a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexExpansionVisitor.java#L190-L203

However, if the whole primary key is in the index key, then the final `fullKeyList` will contain only the index's root expression, which then causes problems when wrapped into a `ThenKeyExpression`. We should be able to fix this by checking the size of the list before calling `concat`.

alecgrieser:
This adds a check while adding the primary key fields to the root expression so that if the primary key is entirely contained within the index expression and therefore doesn't need to be added to the key expression. There's a new test of the planner that fails without this fix, as the new planner needs to consider an expression of that kind. The test also runs under the old planner, which succeeded before and after this fix.

This fixes #2015.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -15,7 +15,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Cascades planner no longer trips over `VALUE` indexes which contain all fields in the primary key [(Issue #2015)](https://github.com/FoundationDB/fdb-record-layer/issues/2015)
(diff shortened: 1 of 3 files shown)


alecgrieser:
It looks like the sonar errors here are for cognitive complexity, and not in the function that I modified, so I think we should ignore it, but up to the reviewer

alecgrieser:
Rebase addresses merge conflicts (in `docs/ReleaseNotes.md`)

normen662 (approved):

PR #2018, opened 2023-02-20
Resolve #2017: Make constructors public

### LINKED ISSUE #2017: Index predicates: Make constructors public
In order to make use of the new IndexPredicate, constructors need to be made public

ohadzeliger:


hatyo (approved):
Nice work @ohadzeliger, it looks good to me, but I would ask @MMcM for a second review if that's ok.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -99,6 +99,25 @@ public class RecordMetaData implements RecordMetaDataProvider {
...
+                Collections.unmodifiableMap(orig.recordTypeKeyToSyntheticTypeMap),
+                Collections.unmodifiableMap(orig.indexes),
+                Collections.unmodifiableMap(orig.universalIndexes),
+                Collections.unmodifiableList(orig.formerIndexes),
+                orig.splitLongRecords,
+                orig.storeRecordVersions,
+                orig.version,
+                orig.subspaceKeyCounter,
+                orig.usesSubspaceKeyCounter,
+                orig.recordCountKey, // TODO: Do we need to protect from modification?
```
MMcM, line 117: `KeyExpression`s are immutable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -99,6 +99,25 @@ public class RecordMetaData implements RecordMetaDataProvider {
             RecordMetaDataProto.getDescriptor(), RecordMetaDataOptionsProto.getDescriptor(), TupleFieldsProto.getDescriptor()
     };
 
+    protected RecordMetaData(@Nonnull RecordMetaData orig) {
```
MMcM, line 102: This doesn't seem to be used in this branch, but I wonder whether a `toBuilder` / `asBuilder` would be more idiomatic.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
ohadzeliger: This gets used in the consumer, where a copy-constructor is used to create an isolated instance from the original.
I thought of using a `toBuilder` pattern but ran into difficulties as the building of `RecordMetaDataBuilder` from `RecordMetadaData` is not trivial: For example, the reverse engineering of a a `RecordTypeBuilder` and the Indexes seemed too involved. I would be happy to do so if there is a path though.

MMcM: I suppose it can wait, then.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
ohadzeliger: Sure. Removed TODO

PR #2021, opened 2023-02-22
resolves #2020: improve logic to push plans through a fetch

### LINKED ISSUE #2020: improve logic to push plans through a fetch
Currently we only allow very simple predicates (as part of a filter plan) of the shape

`q.a = 5`

to be pushed through a fetch (if possible). The logic does not allow for `q.a * q.b < 100` to be pushed (where both `a` and `b` are provided by a covering index scan). We should support this case.

We also should support pushing the result value of a `RecordQueryMapPlan` to be pushed through a fetch, if all fields are available in which case the fetch can be removed entirely.

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -74,10 +75,24 @@
      */
     @Nonnull
     default Iterable<? extends T> inPreOrder() {
+        return inPreOrder(self -> true);
+    }
+
+    /**
+     * Method that returns an {@link Iterable} of nodes as encountered in pre-order traversal of this tree-like.
+     * @param descentIntoPredicate a predicate that determines of the traversal enters the subtree underneath the
```
hatyo, line 83:
```suggestion
     * @param descentIntoPredicate a predicate that determines if the traversal enters the subtree underneath the
```

hatyo (approved):

PR #2025, opened 2023-02-22
Fixes #2024: Listing missing ranges in a RangeSet has overly large conflict ranges

### LINKED ISSUE #2024: Listing missing ranges in a RangeSet has overly large conflict ranges
When inserting into a range set, we take care to only insert read and write conflict ranges over the range we're inserting. See:

https://github.com/FoundationDB/fdb-record-layer/blob/77cc0c5328bea2722482a3e6562869b7b1b56d4e/fdb-extensions/src/main/java/com/apple/foundationdb/async/RangeSet.java#L231-L243

Note that the two `AsyncIterator`s are created at snapshot, and there's a read conflict range added over just the range we're inserting. Because of the data format of the `RangeSet`, it does have to read outside the range, but it will discard any information it gets from keys that aren't in the range actually being inserted.

However, the logic for reading missing ranges from the range set doesn't work like that:

https://github.com/FoundationDB/fdb-record-layer/blob/77cc0c5328bea2722482a3e6562869b7b1b56d4e/fdb-extensions/src/main/java/com/apple/foundationdb/async/RangeSet.java#L514-L517

Here, read conflict ranges can be added for values outside the range, which means that (for example) a transaction that inserts a range can cause a transaction that reads a disjoint range to fail with a conflict.

alecgrieser:
This adjusts the way that the `RangeSet` adds conflict ranges when listing missing ranges so that it only adds a read conflict range if the range is relevant to the final answer. This means that, for example, disjoint ranges can be modified in concurrent transacitons and all commit. There was already logic that adjusted these ranges in `insertRange`, so only the list method needed to be adjusted.

This fixes #2024.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
@@ -261,7 +261,7 @@ public class RangeSetTest extends FDBTestBase {
             // In separate transactions, read the ranges r1 and r3. One of them
             // is before r2 and the other other after
             Transaction tr2 = multi.get(1);
-            assertEquals(List.of(r2), rs.missingRanges(tr2, r1).asList().join());
+            assertEquals(List.of(r1), rs.missingRanges(tr2, r1).asList().join());
             Transaction tr3 = multi.get(2);
             assertEquals(List.of(r3), rs.missingRanges(tr3, r3).asList().join());
 
@@ -406,13 +406,13 @@ public class RangeSetTest extends FDBTestBase {
             List<Range> missing = rs.missingRanges(tr, fullRange).asList().join();
             assertEquals(gaps, missing);
 
-            List<Range> allMissing = rs.missingRanges(tr).asList().join();
             List<Range> expectedAll = new ArrayList<>();
             expectedAll.add(new Range(new byte[]{0}, fullRange.begin));
             if (!gaps.isEmpty()) {
                 expectedAll.addAll(gaps.subList(0, gaps.size() - 1));
             }
             expectedAll.add(new Range(lastInRange, new byte[]{(byte)0xff}));
+            List<Range> allMissing = rs.missingRanges(tr).asList().join();
             assertEquals(expectedAll, allMissing);
 
             rs.insertRange(tr, fullRange, false).join();
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
@@ -261,14 +261,14 @@ public class RangeSetTest extends FDBTestBase {
             Range r2 = new Range(new byte[]{0x02}, new byte[]{0x03});
             Range r3 = new Range(new byte[]{0x04}, new byte[]{0x05});
 
-            // Insert r1
+            // Insert r2
             Transaction tr1 = multi.get(0);
-            assertTrue(rs.insertRange(tr1, r1, true).join());
+            assertTrue(rs.insertRange(tr1, r2, true).join());
 
-            // In separate transactions, read the ranges r2 and r3. One of them
-            // is before r1 and the other other after
+            // In separate transactions, read the ranges r1 and r3. One of them
+            // is before r2 and the other other after
             Transaction tr2 = multi.get(1);
-            assertEquals(List.of(r2), rs.missingRanges(tr2, r2).asList().join());
+            assertEquals(List.of(r2), rs.missingRanges(tr2, r1).asList().join());
             Transaction tr3 = multi.get(2);
             assertEquals(List.of(r3), rs.missingRanges(tr3, r3).asList().join());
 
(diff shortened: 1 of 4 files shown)


fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
```diff
@@ -233,66 +240,182 @@ public void insert() {
...
+            Range r1 = new Range(new byte[]{0x00}, new byte[]{0x01});
+            Range r2 = new Range(new byte[]{0x02}, new byte[]{0x03});
+            Range r3 = new Range(new byte[]{0x04}, new byte[]{0x05});
+
+            // Insert r1
+            Transaction tr1 = multi.get(0);
+            assertTrue(rs.insertRange(tr1, r1, true).join());
+
+            // In separate transactions, read the ranges r2 and r3. One of them
+            // is before r1 and the other other after
```
jjezra, line 269: Aren't both after r1?

jjezra (dismissed):

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
ScottDugas: Yeah, it seems like this is missing coverage of querying before the inserted range.

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
```diff
@@ -233,66 +240,182 @@ public void insert() {
...
+            Transaction tr2 = multi.get(1);
+            assertEquals(List.of(r), rs.missingRanges(tr2, r).asList().join());
+
+            // The second transaction should fail to commit
+            multi.commit(Set.of(1));
+        }
+    }
+
+    @Test
+    void insertNonOverlappingRanges() {
```
ScottDugas, line 347: It seems like it would be good to also have a test that does `[i, i+1)`, which would also be non-overlapping, but abutting. Or would that result in them merging, and thus conflicting?

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
```diff
@@ -233,66 +240,182 @@ public void insert() {
...
+
+    @Test
+    void insertOverlappingRanges() {
+        try (MultipleTransactions multi = MultipleTransactions.create(db, 10)) {
+            for (int i = 0; i < multi.size(); i++) {
+                Transaction tr = multi.get(i);
+                assertTrue(rs.insertRange(tr, new Range(new byte[]{(byte)i}, new byte[]{(byte)(i + 1), 0x00}), false).join());
+            }
+
+            // Even transactions should be able to commit
```
ScottDugas, line 367: even transactions can commit because it commits in order, so in chronological order we have:
1. tr0 commits
2. tr1 conflicts with tr0
3. tr2 does not overlap with tr0's range, so commits
4. tr3 conflicts with tr2
5. tr4 does not overlap with the two that committed, so succeeds
...

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
```diff
@@ -430,4 +541,78 @@ public void missingRanges() {
         }
     }
 
+    private static class MultipleTransactions implements AutoCloseable, Iterable<Transaction> {
```
ScottDugas, line 544: It doesn't have to be done in this PR, but I feel like this class would be pretty generally useful in other tests, so perhaps worth pulling out.

ScottDugas (changes requested):

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
alecgrieser: Added a test with `[i, i+1)`. This does succeed, because each one inserts a different key representing a different range. This means that it isn't coalesced, but given the goal here, it's probably better to have the lack of coalescence rather than to have transactions that fail.

In this test (and the others), I've added some additional tests to make sure that the set makes sense, i.e., there are gaps in the expected places

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
alecgrieser: Yeah, that's right. Expanded this comment

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
alecgrieser: Yeah, that makes sense. I've extracted the class from outside of the `RangeSetTest` into its own top-level class. I haven't migrated any other tests to use it, but now it's available.

alecgrieser:
Rebase was to protect against merge skew, but otherwise didn't make any changes. Responded to comments in 30d2e75 and 153553c

PR #2029, opened 2023-02-23
Resolves #2026: Protect Index.options from modification

### LINKED ISSUE #2026: Protect `Index.options` from modification
Index options (`Map<String,String>`) should be read only to ensure indexes are immutable. While this is true when there are no options, if the index has options defined, the data structure is mutable, allowing code like `index.getOptions().put("A", "B")` to modify the instance

ohadzeliger:
This PR modifies index to ensure Options are immutable. 
The index semantics are meant to be preserved, where an original Index' Options are cloned using ImmutableMap.copyOf(). This method would try to avoid extra copying if the given map (original) is immutable as well (which would normally be the case).
Two tests had to be changed, and a `IndexWithOptions` class was created to facilitate tests that need to modify Options on an existing index.
Risk: Low, most notably:
- Cases where the `equals` calculation has changed (between the `TreeMap` and `ImmutableMap`) may impact overall `equals`
- Cases where a method is given an index and expects to change the Options inline and someone else is relying on this side effect. This is mitigated by failing the options change once it becomes immutable.


alecgrieser (approved):
We may want to call this out in a release note, but otherwise, LGTM

alecgrieser (approved):

alecgrieser (approved):

PR #2032, opened 2023-03-02
Fix Lucene index matching for predicates involving one leg of a join lucene index

hatyo:
If a `LuceneQueryPredicate` is set inside a `NestedField`, representing it a residual, we immediately give up matching it. However if the `NestedField` plus the field(s) in the `LuceneQueryPredicate` _do_ exist in the matched Lucene index we _could_ still match it by removing the `NestedField` and replacing it with a prefix set on each field of the underlying `LuceneQueryPredicate`.

This PR does exactly that.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
```diff
@@ -263,6 +264,15 @@ private LuceneScanParameters getSpecialScan(@Nonnull LucenePlanState state, @Non
         return null;
     }
 
+    @Nonnull
+    private LuceneQueryComponent tryPushResidual(@Nonnull LuceneQueryComponent luceneQueryComponent, @Nonnull final List<String> prefix) {
+        if (prefix.isEmpty()) {
+            return luceneQueryComponent;
+        }
+        final var name = String.join("_", prefix);
+        return new LuceneQueryComponent(luceneQueryComponent.getQuery(), luceneQueryComponent.getFields().stream().map(field -> name + "_" + field).collect(Collectors.toList()));
```
MMcM, line 273: This can lose things like `luceneQueryHighlightParameters`, which seems orthogonal to the fix. Perhaps `LuceneScanQueryParameters` needs a `withOtherFields` or something like that to do it maintainably.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePlanner.java
hatyo: Fixed.

MMcM (approved):

MMcM (approved):

PR #2036, opened 2023-03-06
primary keys for lucene auto complete scans need to be trimmed

normen662:


alecgrieser (changes requested):
The actual changes look good to me. Should there be tests added to make sure auto-complete handles its primary key correctly? Also, it would be nice to have a release note

alecgrieser (approved):

PR #2037, opened 2023-03-06
Remove unused Lucene highlight features

arnaud-lacurie:


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -501,17 +467,18 @@ void simpleLuceneScans(boolean shouldDeferFetch, boolean withHighlight) throws E
             Matcher<RecordQueryPlan> matcher = indexScan(allOf(indexScan("Complex$text_index"),
                     indexScanType(LuceneScanTypes.BY_LUCENE),
                     scanParams(query(hasToString("MULTI civil blood makes civil hands unclean")))));
-            if (withHighlight) {
-                matcher = highlight(matcher);
-            }
+            // TODO(alacurie) WHY?
```
MMcM, line 470: `LuceneHighlightTermsPlan` wrapping has been removed from `LucenePlanner`. Perhaps the whole class (and the associated matcher can now be removed?

The `withHighlight` parameter of this test doesn't do much that one can tell now. It exists so that the `LuceneQueryHighlightParameters` can get attached to the plan, even though they aren't used in the direct scan results. I guess either it can be removed altogether or the test should do some additional checking that the `FDBQueriedRecord` have access to custom highlighting settings.

Or is this just meant to be a step on the way to larger changes and so somewhat odd in ways like this?

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
arnaud-lacurie: That's a good point, I'll remove that!

MMcM (approved):

PR #2038, opened 2023-03-07
Implementing a BitSetQuery in Lucene

scottfines:
This does the programmatic work for implementing Bitset operations in Lucene. It doesn't add tests though, and it isn't integrated in with the rest of lucene, but this PR exists to make it easy to move the work from one person to another


scottfines:
This was done to ease communication, rather than to be committed. Since that communication is cleared, closing the PR

PR #2045, opened 2023-03-08
Copy Lucene query parser.

hatyo:
This copies Lucene query parser into our code base to achieve the following goals:

 * more fine-grained control of fields of certain types w.r.t. used analysers.
 * extend the parser to support new features such as bitset indexing.
 * have more control over parsing workflow, and be more explicit about error messages, if possible.

a new Gradle task `compileJavacc` is added to `fdb-record-layer-lucene.gradle` to perform code (re)generation.


fdb-record-layer-lucene/src/main/java/org/apache/lucene/queryparser/classic/QueryParser.java
```diff
@@ -0,0 +1,1050 @@
+/* QueryParser.java */
+/* Generated By:JavaCC: Do not edit this line. QueryParser.java */
+package org.apache.lucene.queryparser.classic;
```
alecgrieser, line 3: It looks like we're checking in both `QueryParser.java` and `QueryParser.jj`. Does running javacc generation on `QueryParser.jj` generate `QueryParser.java`? And is the idea that if people edit `QueryParser.jj` that they check in the new `QueryParser.java`?

alecgrieser (changes requested):
I think this mostly LGTM. I think it would be good to separate the javacc source files from the Java files, and to avoid intermingling the generated and copied over Java files.

There are a few places where I'm not quite sure I see how things fit together. For example, I'm not sure I totally see why even if we want to take over the grammar that we need to also copy over, say, `QueryParserBase` and `FastCharStream`. It also wasn't clear to me if it was better to import the `javacc` generated Java files rather than copying over the base grammar files, if that's what's going on

hatyo:
It seems we don't really need the parser (for now, at least 😄) as we were able to address all the bugs we wanted to fix without having to extend the parser:

* Extension for bitsets is done here #2049 , admittedly it is a bit hacky, but still much better than importing the entire parser to support it.
* Fix for handling stored/sorted Lucene document fields is done here #2047 . Again this doesn't need parser extensions as it sets the analyser in earlier stages.
* I looked around a bit, and it seems that improving the error messages (namely not ignoring a predicate because it wasn't tokenized correctly) can be done in an extension of `QueryParser` such as `LuceneOptimizedMultiFieldQueryParser` and `LuceneOptimizedQueryParser`.

Neverthless, thank you @alecgrieser for the extensive code review, we'll probably revisit your comments if we attempt to make this effort in the future.

PR #2046, opened 2023-03-09
add BITSET_CONTAINS to query parser without forking the lucene parser logic

scottfines:
This is a draft version which passes Arnaud's test, but doesn't make use of a forked query parser from lucene (which is advantageous for us).

However, I think the semantics on this are still kinda wonky, at least from a user's perspective. Forcing the user to pass through the entire bit mask in one go may be OK, but it's not particularly ergonomic. It would be _nicer_ if we could do something like `BITSET_CONTAINS(1,3,7,...)` or something, so that you can just specify the indexes you want to be one, and let us magically create the correct bit mask from it. But that may be difficult to implement directly here, so possibly this is the best short-term solution


arnaud-lacurie:
Closing this in favour of #2049

PR #2047, opened 2023-03-09
Add internal Lucene analysers to store and sorted document fields

hatyo:
This fixes a problem with potentially parsing a Lucene query field predicate incorrectly if the field is stored or sorted. The issue is caused because the same text analyser, that is specified in the schema by the customer, is used to analyse the RHS of a stored field predicate which could lead to incorrect results.

When dealing with stored and sorted fields, the index maintainer will store their values as-is, therefore, when it is expected that a predicate on these fields gets the same treatments for its operands. This PR addresses this behaviour by doing the following changes:

* Implement a very simple no-op analyser, analyser factory, and tokeniser, that pass-through the analysed terms.
* In `LuceneAnalyzerRegistry`, add a overload of the method `getLuceneAnalyzerCombinationProvider` that accepts a third parameter which is an auxiliary mapping between each index field and its `DocumentFieldDerivation`.
* During the analysis of the index and creating its set of analysers, we examine the auxiliary map, and if feasible, add the above no-op analyser to each field that is stored or sorted.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
@@ -23,6 +23,7 @@ package com.apple.foundationdb.record.lucene;
 import com.apple.foundationdb.record.metadata.Index;
 
 import javax.annotation.Nonnull;
+import java.util.Map;
 
 /**
  * Registry for {@link AnalyzerChooser}s. This registry allows for full-text indexes to specify
@@ -39,4 +40,9 @@ import javax.annotation.Nonnull;
 public interface LuceneAnalyzerRegistry {
     @Nonnull
     LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index, @Nonnull LuceneAnalyzerType type);
+
+    @Nonnull
+    LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index,
+                                                                           @Nonnull LuceneAnalyzerType type,
+                                                                           @Nonnull Map<String, LuceneIndexExpressions.DocumentFieldDerivation> auxiliaryFieldInfo);
 }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
@@ -38,8 +38,6 @@ import java.util.Map;
  */
 @SuppressWarnings("unused")
 public interface LuceneAnalyzerRegistry {
-    @Nonnull
-    LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index, @Nonnull LuceneAnalyzerType type);
 
     @Nonnull
     LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index,
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.lucene;
 
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
+import com.apple.foundationdb.record.lucene.exact.ExactTokenAnalyzerFactory;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.MetaDataException;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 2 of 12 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/exact/ExactTokenAnalyzerFactory.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * Constructs a new instance of {@link ExactTokenAnalyzer}.
+ */
+@AutoService(LuceneAnalyzerFactory.class)
+public class ExactTokenAnalyzerFactory implements LuceneAnalyzerFactory  {
+
+    public static final String NAME = "QUERY_ONLY_EXACT_ANALYZER_10";
```
alecgrieser, line 39: Why is there a 10 at the end of this name?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -88,29 +89,63 @@ private LuceneAnalyzerRegistryImpl() {
...
+                throw new MetaDataException("could not retrieve analyzer",
+                        LuceneLogMessageKeys.ANALYZER_NAME, ExactTokenAnalyzerFactory.NAME,
+                        LuceneLogMessageKeys.ANALYZER_TYPE, LuceneAnalyzerType.FULL_TEXT);
+            }
+            chooserPerFieldOverride.put(fieldName, registry.get(LuceneAnalyzerType.FULL_TEXT).get(ExactTokenAnalyzerFactory.NAME).getIndexAnalyzerChooser(index));
+        }
+    }
+
+    private static boolean isEligibleForNoOpAnalyzer(@Nonnull final LuceneIndexExpressions.DocumentFieldDerivation fieldInfo) {
+        return fieldInfo.isSorted() || fieldInfo.isStored();
```
alecgrieser, line 146: Rather than looking at the `sorted` and `stored` of the field, should it look at the `fieldInfo`'s type instead? In particular, I'd think you'd want to use the exact analyzer unless the type is `TEXT` (as only `TEXT` fields are going to actually bo tokenized in the index)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
```diff
@@ -39,4 +40,9 @@
 public interface LuceneAnalyzerRegistry {
     @Nonnull
     LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index, @Nonnull LuceneAnalyzerType type);
+
+    @Nonnull
+    LuceneAnalyzerCombinationProvider getLuceneAnalyzerCombinationProvider(@Nonnull Index index,
```
alecgrieser, line 45: Should the old overload of this class be removed? It seems like the user should always use this new one to get the right analyzer anyway

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -88,29 +89,63 @@ private LuceneAnalyzerRegistryImpl() {
...
+                                                @Nonnull final LuceneIndexExpressions.DocumentFieldDerivation fieldInfo,
+                                                @Nonnull final Index index) {
+        if (chooserPerFieldOverride.containsKey(fieldName)) {
+            // do not override already chosen field analyzer.
+            return;
+        }
+        if (isEligibleForNoOpAnalyzer(fieldInfo)) {
+            if (registry.isEmpty()
+                    || registry.get(LuceneAnalyzerType.FULL_TEXT) == null
+                    || registry.get(LuceneAnalyzerType.FULL_TEXT).get(ExactTokenAnalyzerFactory.NAME) == null) {
```
alecgrieser, line 136: Is this right if the user has specified that they're looking for the `AUTO_COMPLETE` analyzer?

alecgrieser (changes requested):
Mostly LGTM. It would be nice if there was some kind of test at this level, though one could also be added later. This also could use a release note

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
hatyo: That's interesting, I changed the check to:
```java
return fieldInfo.getType() != LuceneIndexExpressions.DocumentFieldType.TEXT;
```
I am a bit worried about `STRING` type though. and more generally whether this no-op analyser can substitute, correctly, the `StandardAnalyser` (for cases where the analyser is not defined, and the fields are _not_ `stored` or `sorted`).

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistry.java
hatyo: done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
hatyo: Good point! fixed it. Thanks!

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
alecgrieser: I think for `STRING` type fields, we want the no-op analyzer, because the field shouldn't have been analyzed before being indexed...if it is indexed, somehow. At least, that's my understanding

alecgrieser:
Is this missing a push with review comment responses?

hatyo:
> Is this missing a push with review comment responses?

Oh, I usually reply as I address comments, sorry about the confusion, I just pushed a new commit.

alecgrieser (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
MMcM: Ideally the PR description and summary would be changed to reflect this new, simpler determining factor.

MMcM (approved):
Just since I was confused:

> This fixes a problem with potentially parsing a Lucene query field predicate incorrectly if the field is ~~stored or sorted~~ not full-text.

hatyo:
> Just since I was confused:
> 
> 
> 
> > This fixes a problem with potentially parsing a Lucene query field predicate incorrectly if the field is ~~stored or sorted~~ not full-text.
> 
> 
> 
> 

Apologies for the confusion, I updated the PR description.

PR #2049, opened 2023-03-09
Support `bitset_contains` in Lucene query

hatyo:
Original comment as in #2048:

> This introduces a BITSET_CONTAINS function to the lucene query language:
>
> `flag:BITSET_CONTAINS(mask)`
> This returns true if the mask is wholly contained in flag.

This mainly copies #2046 and #2048 removing all dependencies on parser integration work.


alecgrieser (changes requested):
In general, LGTM. This looks like it could also use a release note

alecgrieser (approved):

arnaud-lacurie (approved):

PR #2054, opened 2023-03-10
Resolves #2053: Default protoMajorVersion to 3 in build

### LINKED ISSUE #2053: Default protoMajorVersion to 3 in build
Given that proto2 is getting a little long in the tooth, we should consider switching what we have left on proto2 to using it. For a variety of reasons (detailed here: https://foundationdb.github.io/fdb-record-layer/Overview.html#indexing-and-querying-of-missing--null-values), we'll probably want to continue supporting `syntax="proto2"` files for a while, but we can/should upgrade our Java dependency to proto3.

An easy first step would be to default the build to use proto3. We'd still be able to ask for proto2 builds (via the existing `PROTO_VERSION` environment variable), but calling `./gradle build` would now produce proto3-based builds. This also interacts a bit better with IDEs, which are harder to thread through custom build environments through

alecgrieser:
This updates the build script so that if one runs `./gradlew build` with no other arguments, it will build the proto3-based artifacts. If one sets the `PROTO_VERSION` to 2, it should still build the proto2-based artifacts, so this won't change the published artifacts, but it may affect flows of a developer working on the project.

This resolves #2053.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -201,12 +201,13 @@ subprojects {
         if (System.getenv('PROTO_VERSION') != null) {
             protoMajorVersion = System.getenv('PROTO_VERSION')
         } else {
-            protoMajorVersion = '2'
+            protoMajorVersion = '3'
         }
         if (protoMajorVersion == "2") {
             protobufVersion = protobuf2Version
         } else if (protoMajorVersion == "3") {
             protobufVersion = protobuf3Version
+            setProperty('coreNotStrict', 'true')
         } else {
             throw new GradleException("Unknown protobuf major version: ${protoMajorVersion}")
         }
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -201,12 +201,13 @@ subprojects {
...
             protoMajorVersion = System.getenv('PROTO_VERSION')
         } else {
-            protoMajorVersion = '2'
+            protoMajorVersion = '3'
         }
         if (protoMajorVersion == "2") {
             protobufVersion = protobuf2Version
         } else if (protoMajorVersion == "3") {
             protobufVersion = protobuf3Version
+            setProperty('coreNotStrict', 'true')
```
alecgrieser, line 210: This is necessary so that if someone runs `./gradlew build`, they don't get validation errors from the proto3 generated code. I thought about removing the `coreNotStrict` flag entirely and replacing it with checks on the `protoMajorVersion`, but I figured this was less invasive, though I could be argued either way

MMcM (approved):

PR #2057, opened 2023-03-16
Fix stack overflow in LuceneNotQuery.planHash

arnaud-lacurie:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneNotQuery.java
```diff
@@ -91,7 +91,7 @@ public void getPlannerGraphDetails(@Nonnull ImmutableList.Builder<String> detail
 
     @Override
     public int planHash(@Nonnull final PlanHashKind hashKind) {
-        return super.planHash() - PlanHashable.iterablePlanHash(hashKind, negatedChildren);
+        return PlanHashable.iterablePlanHash(hashKind, negatedChildren);
```
normen662, line 94: what about the impact of the regular `children`. Looks like they are not hashed anymore.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -843,6 +843,7 @@ void testLimitForAutoCompleteWithMultipleMatchingFields() throws Exception {
                             false,
                             null,
                             COMPLEX_MULTI_GROUPED_WITH_AUTO_COMPLETE_STORED_FIELDS);
+            assertEquals(-35583570, luceneIndexPlan.planHash());
```
normen662, line 846: Why are all the hashes in this file identical? Are they always the same plan?

normen662 (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneNotQuery.java
arnaud-lacurie: Good point, we do need this super function after all, but we forgot to provide the hashKind

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
arnaud-lacurie: The hashes are identical for four test cases which seem to have identical queries, but use different data sets and execution properties.

normen662: Okeydok

normen662 (approved):

PR #2058, opened 2023-03-16
resolves #2061: improve join enumeration and plan memoization

### LINKED ISSUE #2061: improve join enumeration and plan memoization
The current join-enumeration algorithm (via `ImplementNestedLoopJoin`) is inflexible and wasteful during replanning due to bad memoization. We essentially transition from a `SelectExpression` owning `n` quantifiers to a tree of nested-loop joins (`RecordQueryFlatMapPlan`). While that tree step-by-step, we pick two quantifiers of the select and join them and then re-add the remainder of the original `SelectExpression` as a new `SelectExpression` ranging over the e.g. nested loop join. That causes enumeration of that rest, but also that rest is already enumerated in the original `SelectExpression`. The rest cannot easily be memoized. Also, if the `SelectExpression` needs to be re-implemented (re-planned) due to e.g. other ordering requirements this explorations starts again. This is bad and needs to be improved.

normen662:
Contains the following features:
- profiling support for the planner, i.e. metrics that keep track of time spent in various transformations, memoization; total time vs. own time; various counts
- reimplementation of join enumeration to separate the explorative aspect of join enumeration from the physical implementation, i.e. the join-partitioning of n-way joins into a bush of binary joins which is common for all replanning VS the physical data access patterns that can be utilized and _implemented_
- implementation of a switch that allows the user to defer the planning of cross products in order to eliminate exploration space
- reimplementation of `EXISTS()` predicate in a way that an existential quantifier now acts more like a scalar quantifier that just happens to have a predicate on top of it rather than some construct that needs constant special casing
- improvements for matching of indexes


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -134,14 +141,53 @@ public void yield(@Nonnull ExpressionRef<? extends RelationalExpression> express
...
+                    traversal.addExpression(expressionReference, member);
                 }
             }
         } else {
             throw new RecordCoreArgumentException("found a non-group reference in an expression used by the Cascades planner");
         }
     }
 
+    @Nonnull
+    public ExpressionRef<? extends RelationalExpression> ref(@Nonnull final RelationalExpression expression) {
```
normen662, line 153: This method allows memoization of a new expression, by looking at the references directly owned by the new expression. If there already is another expression using those references, it can be memoized. This approach works really well for algorithms following the classic dynamic programming paradigm such as during join enumeration and incremental predicate normalization.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -302,6 +302,8 @@ static Compensation impossibleCompensation() {
     /**
      * Returns a specific compensation object that uses a mapping between predicates from a query to predicates that
      * are used for the application of the compensation.
+     * @param isImpossible an indicator if this compensation can be applied, if not, this compensation
```
normen662, line 305: The changes in this file fix a problem that only got exposed by the other changes. Consider two matches and two predicates in a `SelectExpression`. Each match can sarg one predicate and need to be compensated for a remaining predicate. Let's further say that match `1` and match `2` sarg the respective other predicate and therefore need to compensate the respective other predicate. Further there are cases where compensations are impossible (e.g. `EXISTS()` compensations may need to reapplied). Let's say match `1` has such an impossible compensation (previously we just wouldn't match), it cannot be planned by itself _BUT_ the compensation can be intersected with other compensations which may eliminate the need for the impossible compensation to be planned at all.

normen662: Addendum: The approach here is a little bit like complex numbers and real numbers. A match can lead to a compensation which is considered impossible (or not plannable). That's like a complex number with an imaginary part. Performing operations other than planning the compensation, such intersecting this compensation with other compensations may lead you back to a plannable compensation (a real number).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GroupExpressionRef.java
```diff
@@ -181,6 +181,8 @@ private boolean insert(@Nonnull final T newValue, @Nullable final Map<PlanProper
         } finally {
             Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(newValue, Debugger.Location.END)));
         }
+        Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(newValue, containsInMemo ? Debugger.Location.REUSED : Debugger.Location.NEW)));
```
normen662, line 184: All of this code that is sprinkled throughout the planner package is used to do some bookkeeping in order to pin-point problems with memoization, over-exploration, over-implementation of groups. There is a new `showStats()` method on `Debugger.getDebugger()` that can be evaluated from the IDE to bring up a stats page about the last planning cycle which can be quite helpful.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueIndexLikeMatchCandidate.java
```diff
@@ -147,7 +151,10 @@ default Ordering computeOrderingFromScanComparisons(@Nonnull final ScanCompariso
                     new ScalarTranslationVisitor(normalizedKeyExpression).toResultValue(Quantifier.current(),
                             getBaseType());
 
-            result.add(OrderingPart.of(normalizedValue, isReverse));
+            if (!normalizedValues.contains(normalizedValue)) {
```
normen662, line 154: Independent fix for bug exposed by the other code.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/Debugger.java
```diff
@@ -215,7 +215,9 @@ enum Location {
         MATCH_PRE,
         YIELD,
         FAILURE,
-        COUNT
+        COUNT,
+        NEW,
+        REUSED
```
normen662, line 220: `REUSED` is a metric to show how many `yield()` were actually not inserted but reused.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -238,6 +241,24 @@ public boolean isRuleEnabled(@Nonnull CascadesRule<?> rule) {
...
+     * Return whether the query planner should defer cross products during join enumeration.
+     * <br>
+     * Cross products are joins of data sources that are completely independent of one another, i.e., the data sources
+     * are not <em>connected</em> via a join predicate, nor is one correlated to the other. It is usually better to
+     * plan a cross product as the last join in a join tree as a cross product tends to increase the cardinality.
+     * Thus, such a join does not need to be planned among all the other joins and can therefore avoid useless
+     * exploration and over-enumeration.
+     * <br>
+     * Unfortunately, there are cases, however, where it may be beneficial to treat a cross product like any other
+     * join in a query.
```
MMcM, line 254: Is there a one sentence example that would fit in this comment, rather than leaving it elusive?

I think it's something like two (or more) low cardinality dimension tables and some (presumably very selective) predicates on the fact table.

MMcM (changes requested):

MMcM (changes requested):

PR #2070, opened 2023-03-23
Use Index predicate when appropriate 

jjezra:
FDBRecodStore is currently using a predefined IndexMaintenanceFilter as filter (defaulted to IndexMaintenanceFilter.NORMAL).
With the new Index.predicate field, change the logic to be:

If the user had explicitly set his own IndexMaintenanceFilter, use it.
Else if the index predicate was defined, use it.
Else use IndexMaintenanceFilter.NORMAL.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
@@ -46,6 +46,7 @@ import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
+import com.apple.foundationdb.record.metadata.IndexPredicate;
 import com.apple.foundationdb.record.metadata.IndexRecordFunction;
 import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.RecordType;
@@ -363,13 +364,35 @@ public abstract class StandardIndexMaintainer extends IndexMaintainer {
         if (savedRecord == null) {
             return null;
         }
+        // Apply both filters:
+        // 1. Index predicates (if exist)
+        // 2. IndexMaintenanceFilter
+        // In the longer term, we will probably think about deprecating the index maintenance filter.
+        final FDBStoreTimer timer = state.store.getTimer();
+        final IndexPredicate predicate = state.index.getPredicate();
+        if (predicate != null) {
+            final long startTime = System.nanoTime();
+            final boolean useMe = predicate.shouldIndexThisRecord(state.store, savedRecord);
+            // Note: for now, IndexPredicate will not support filtering of certain index entries
+            if (timer != null) {
+                final FDBStoreTimer.Events event =
+                        useMe ?
+                        FDBStoreTimer.Events.USE_INDEX_RECORD_BY_PREDICATE :
+                        FDBStoreTimer.Events.SKIP_INDEX_RECORD_BY_PREDICATE;
+                timer.recordSinceNanoTime(event, startTime);
+            }
+            if (!useMe) {
+                // Here: index predicate filters out this record
+                return null;
+            }
+        }
         final Message record = savedRecord.getRecord();
         long startTime = System.nanoTime();
         boolean filterIndexKeys = false;
         switch (state.filter.maintainIndex(state.index, record)) {
             case NONE:
-                if (state.store.getTimer() != null) {
-                    state.store.getTimer().recordSinceNanoTime(FDBStoreTimer.Events.SKIP_INDEX_RECORD, startTime);
+                if (timer != null) {
+                    timer.recordSinceNanoTime(FDBStoreTimer.Events.SKIP_INDEX_RECORD, startTime);
                 }
                 return null;
             case SOME:
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -27,7 +27,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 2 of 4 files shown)


MMcM:
It seems to me that a simpler contract would be that both the record store's filter from `Builder.setIndexMaintenanceFilter` and the `Index`'s predicate from `Index.getPredicate` are used. The former can be made a no-op, if desired, with `NORMAL` and the latter is already `Nullable`. This can be done either in `filteredIndexEntries`, as here, by calling both, as appropriate, or when the `IndexMaintainerState` is setup, by merging them.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4647,8 +4647,8 @@ public static class Builder implements BaseBuilder<Message, FDBRecordStore> {
         @Nonnull
         private IndexMaintainerRegistry indexMaintainerRegistry = IndexMaintainerRegistryImpl.instance();
 
-        @Nonnull
-        private IndexMaintenanceFilter indexMaintenanceFilter = IndexMaintenanceFilter.NORMAL;
+        @Nullable
+        private IndexMaintenanceFilter indexMaintenanceFilter = null;
```
normen662, line 4651: Why are we changing for a special element `NORMAL` specific to `IndexMaintenanceFilter` (that allows us to be `@Nonnull` in a lot of places, thus allowing us to call methods on it without prior checking) to a non-specific `null`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -363,10 +372,20 @@ public <M extends Message> List<IndexEntry> filteredIndexEntries(@Nullable final
         if (savedRecord == null) {
             return null;
         }
+        IndexMaintenanceFilter filter = state.filter;
+        if (filter == null) {
+            final IndexPredicate predicate = state.index.getPredicate();
+            if (predicate != null) {
+                return filteredIndexEntriesByPredicate(savedRecord);
+            } else {
+                filter = IndexMaintenanceFilter.NORMAL;
```
normen662, line 381: here we set it back to `NORMAL`? I think it's doing something subtle but I think this is lost without a good comment somewhere.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
jjezra: The requirement, as specified by @arnaud-lacurie, was to prefer the `indexMaintenanceFilter` over the new predicates iff it was explicitly set the caller (for backward compatibility, I suppose). Setting it to `NORMAL` preemptively hides the information, prevents the code below from knowing if it was explicit or not (as the user may have also set it to `NORMAL`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
jjezra: It is explained in the PR comment. I'll add a comment here too and refresh your review request .
✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -400,6 +423,29 @@ public <M extends Message> List<IndexEntry> filteredIndexEntries(@Nullable final
...
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug(KeyValueLogMessage.of("index predicate used",
+                    LogMessageKeys.INDEX_NAME, state.index.getName(),
+                    LogMessageKeys.FILTER, predicate));
+        }
+        CorrelationIdentifier objectQuantifier = Quantifier.current();
+        final RecordType recordType = state.store.getRecordMetaData().getRecordType(savedRecord.getRecordType().getName());
+        Type.Record typeRecord = Type.Record.fromDescriptor(recordType.getDescriptor());
+        Value recordValue = QuantifiedObjectValue.of(objectQuantifier, typeRecord);
+        QueryPredicate queryPredicate = predicate.toPredicate(recordValue);
```
hatyo, line 440: This looks expensive, is it possible to cache the generated predicate (`queryPredicate`) and only (re)generate it if the incoming `savedRecord` has a different `typeRecord`?

hatyo (commented):
LGTM, but I would ask @alecgrieser to also have a look as he is more familiar with index filtering logic than I.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
jjezra: I like this idea!

jjezra: From what I'm seeing, the standard maintainer is regenerate by `FDBRecordStore` for every record update. I wonder if the cache should be saved at the index predicate..

hatyo: > I wonder if the cache should be saved at the index predicate..

I think this is fine, although I rather put it in a separate small (caching) factory class so we can keep the POJO contract of the `IndexPredicate` class.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -427,22 +419,21 @@ public <M extends Message> List<IndexEntry> filteredIndexEntries(@Nullable final
     private <M extends Message> List<IndexEntry> filteredIndexEntriesByPredicate(@Nonnull final FDBIndexableRecord<M> savedRecord) {
         // Note: for now, IndexPredicate will not support filtering of certain index entries
         final IndexPredicate predicate = state.index.getPredicate();
-        // Already checked: saved record and predicate are not null,
+        if (predicate == null) {
+            // Already checked: saved record and predicate are not null, yet keeping java happy.
+            return null;
+        }
+        boolean useMe = predicate.indexThisRecord(state.store, savedRecord);
```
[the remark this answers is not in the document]
jjezra: ✅

MMcM (approved):

normen662 (approved):

PR #2076, opened 2023-03-30
Refactor the highlight code into two passes: find & cut

arnaud-lacurie:
First pass with the query analyzer: we find all matches and where they are in the original text
Second pass with the standard analyzer: we extract snippets from the original text

Before this change, we would advance the query analyzer and the standard analyzer respective tokenStream at the same time. The assumption was that the standard analyzer would always be behind the query analyzer because the standard analyzer looks at stop words. 
However, because of an issue with the way synonyms were handled, we ended up with the query analyzer running behind, which could lead to StringIndexOutOfBounds issues.


MMcM (approved):
At some point, this needs more tests. #2077 restores some tests of highlighting proper, but those tests do not enable snippets.

MMcM (approved):

PR #2078, opened 2023-03-31
resolves #2061: fixes for join enumeration

### LINKED ISSUE #2061: improve join enumeration and plan memoization

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -156,7 +143,7 @@ public boolean equalsWithoutChildren(@Nonnull final RelationalExpression other,
 
     @Override
     public int hashCodeWithoutChildren() {
-        return Objects.hash(computeResultSupplier, getGroupingValue(), getAggregateValue());
```
hatyo, line 159: 🤦

hatyo (approved):
LGTM

PR #2079, opened 2023-04-03
Resolves #2083: Lucene auto complete v2

### LINKED ISSUE #2083: Remove old Lucene `auto-complete` functionality and create a proper composable query clause for auto-complete functionality
The original implementation of auto-complete used to build on top of a special Lucene index. That is no longer the case. As a remnant we see specific auto-complete-related implementations for index plans, cursors, etc. The Lucene index maintainer also uses special hooks to facilitate auto-complete-driven index scans.

Aside from an obsolete implementation, the semantics of the auto-complete functionality (with the pivoting logic) made hard to impossible to compose with other query fragments in order to formulate rich queries.

normen662:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteQueryClause.java
```diff
@@ -0,0 +1,316 @@
+/*
+ * LuceneQueryMultiFieldSearchClause.java
```
MMcM, line 2:
```suggestion
 * LuceneAutoCompleteQueryClause.java
```

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAutoCompleteQueryClause.java
normen662: done

MMcM (approved):

PR #2087, opened 2023-04-05
Fix bad error

scottfines:
We accidentally used the wrong error type in the LuceneQuerySearchClause, so the error was getting replaced with a construction error.

While I was at it, I added some .gitignores to avoid accidentally committing VSCode folders


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQuerySearchClause.java
```diff
@@ -83,7 +83,7 @@ public Query bind(@Nonnull FDBRecordStoreBase<?> store, @Nonnull Index index, @N
         try {
             return parser.parse(searchString);
         } catch (Exception ioe) {
-            throw new RecordCoreArgumentException("Unable to parse search given for query", ioe);
+            throw new RecordCoreException("Unable to parse search given for query", ioe);
```
alecgrieser, line 86: Could you go into a little more detail as to why this change makes a difference? Is there something above this that is `catch`-ing `RecordCoreArgumentException`s, but you want this error to be rethrown?

In principle, this doesn't seem like a bad change, given that this is only really a "bad argument" using some lateral thinking.

alecgrieser (approved):

PR #2088, opened 2023-04-07
Mutual indexing: smarter ANY iteration - skip fragments that produce rangeSet conflicts

jjezra:
Implement `anyJumper` 
During the ANY iteration: 
  If `throttledRunAsync` gets an exception, the mutual indexer will store it and return quietly 
  Then, during the next outer iteration:
        If at the same fragment, same range: re-throw (as the problem was not related to a rangeSet conflicts)
        If at the same fragment, different range: jump to the next fragment

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
@@ -341,6 +341,8 @@ public class IndexingMutuallyByRecords extends IndexingBase {
         if (fragmentIterationType == FragmentIterationType.RECOVER) {
             // At this point, recovery should be done manually after investigation. In the future we may fallback
             // to a regular multi target indexing - with or without a type stamp change.
+            // Note that if there are less unbuilt fragments than active indexers, we will expect the surplus indexer to
+            // reach this iteration (see anyJumper below).
             throw new ValidationException("Mutual indexing failure - third iteration");
         }
         validateSameMetadataOrThrow(store);
@@ -552,19 +554,19 @@ public class IndexingMutuallyByRecords extends IndexingBase {
                     .addKeysAndValues(fragmentLogMessageKeyValues())
                     .toString());
         }
-        if (anyJumperCurrent == fragmentCurrent) {
-            if (anyJumperRange.equals(rangeToBuild)) {
-                // Here: in hindsight, this exception was not caused by a rangeSet conflict. Rethrow it.
-                throw anyJumperEx;
-            }
-            // Here: Another indexer is working on this fragment, jump to the next one
-            timerIncrement(FDBStoreTimer.Counts.MUTUAL_INDEXER_ANY_JUMP);
+        if (anyJumperCurrent != fragmentCurrent) {
+            // Here: this way or another, the conflicting fragment is fully built now. Irrelevant.
             anyJumperEx = null;
-            return true;
+            return false;
         }
-        // Here: this way or another, the conflicting fragment is fully built now. Irrelevant.
+        if (anyJumperRange.equals(rangeToBuild)) {
+            // Here: in hindsight, this exception was not caused by a rangeSet conflict. Rethrow it.
+            throw anyJumperEx;
+        }
+        // Here: Another indexer is processing this fragment, jump to the next one.
+        timerIncrement(FDBStoreTimer.Counts.MUTUAL_INDEXER_ANY_JUMP);
         anyJumperEx = null;
-        return false;
+        return true;
     }
 
     @Nullable
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
@@ -369,16 +369,22 @@ public class IndexingMutuallyByRecords extends IndexingBase {
                 }
                 return AsyncUtil.READY_TRUE; // let the caller handle the recovery
             }
-            Range rangeToBuild = fragmentIterationType == FragmentIterationType.FULL ?
+            final boolean isFull = fragmentIterationType == FragmentIterationType.FULL;
+            Range rangeToBuild = isFull ?
                                  fullyUnBuiltRange(missingRanges, fragmentRange) :
                                  partlyUnBuiltRange(missingRanges, fragmentRange);
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug(KeyValueLogMessage.of("range to build",
-                        LogMessageKeys.SCAN_TYPE, fragmentIterationType,
-                        LogMessageKeys.RANGE, rangeToBuild,
-                        LogMessageKeys.ORIGINAL_RANGE, fragmentRange));
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info(KeyValueLogMessage.build("range to build",
+                                LogMessageKeys.SCAN_TYPE, fragmentIterationType,
+                                LogMessageKeys.RANGE, rangeToBuild,
+                                LogMessageKeys.ORIGINAL_RANGE, fragmentRange)
+                        .addKeysAndValues(fragmentLogMessageKeyValues())
+                        .toString());
             }
             if (rangeToBuild != null && !anyJumperSaysJump(rangeToBuild)) {
+                timerIncrement(isFull ?
+                               FDBStoreTimer.Counts.MUTUAL_INDEXER_FULL_START :
+                               FDBStoreTimer.Counts.MUTUAL_INDEXER_ANY_START);
                 infiniteLoopProtection(rangeToBuild, missingRanges);
                 final List<Object> additionalLogMessageKeyValues = new ArrayList<>(
                         Arrays.asList(LogMessageKeys.CALLING_METHOD, "mutualMultiTargetIndex",
@@ -392,6 +398,9 @@ public class IndexingMutuallyByRecords extends IndexingBase {
                 ).thenCompose(ignore -> AsyncUtil.READY_TRUE);
             }
             if (anyJumperEx == null) {
+                timerIncrement(isFull ?
+                               FDBStoreTimer.Counts.MUTUAL_INDEXER_FULL_DONE :
+                               FDBStoreTimer.Counts.MUTUAL_INDEXER_ANY_DONE);
                 fragmentPlusPlus();
             }
         }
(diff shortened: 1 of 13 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -84,15 +84,13 @@ public class IndexingThrottle {
     }
 
     public <R> CompletableFuture<R> buildCommitRetryAsync(@Nonnull BiFunction<FDBRecordStore, AtomicLong, CompletableFuture<R>> buildFunction,
```
jjezra, line 86: After deleting indexingByRecord, we may be able to simplify the code a bit by converting the <R> to a Boolean.

alecgrieser (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -1322,8 +1322,7 @@ void testMutualIndexingBlockerWhileActivelyIndexing() {
         }
 
         // After unblocking, finish building
-        IntStream.rangeClosed(0, 3).parallel()
-                .forEach(ignore -> oneThreadIndexing(indexes, timer, boundariesList));
+        oneThreadIndexing(indexes, timer, boundariesList);
```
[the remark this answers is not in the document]
jjezra: For the record - this change is not fully related to the any jumper - I did notice that this test fails about 5% of the time and learnt that it fails here because sometimes one of the indexers, while initializing, expects a WRITE_ONLY state when it's already READABLE  - just a timing issue. Since a single thread doesn't seem to matter to the test at this point, I preferred to reduce the noise. 
When it comes to the anyJumper, I could not create a reliable, repeateable test. The timing is very tricky..

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -514,6 +536,59 @@ private void infiniteLoopProtection(final Range range, final List<Range> missing
...
+        }
+        return ex -> {
+            if (anyJumperEx != null) {
+                // Here: anyJumperSaysJump is re-throwing. Do not interrupt.
+                anyJumperEx = null;
+                return Optional.empty();
+            }
+            // Here: this exception may or may not be a rangeSet conflict.
+            // Memorize this info and handle it during the next iteration.
+            anyJumperEx = ex;
```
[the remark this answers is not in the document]
alecgrieser: I see. I suppose that decision makes sense

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -362,16 +371,22 @@ private CompletableFuture<Boolean> buildNextRangeOnly(List<Range> missingRanges,
...
+            Range rangeToBuild = isFull ?
                                  fullyUnBuiltRange(missingRanges, fragmentRange) :
                                  partlyUnBuiltRange(missingRanges, fragmentRange);
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug(KeyValueLogMessage.of("range to build",
-                        LogMessageKeys.SCAN_TYPE, fragmentIterationType,
-                        LogMessageKeys.RANGE, rangeToBuild,
-                        LogMessageKeys.ORIGINAL_RANGE, fragmentRange));
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info(KeyValueLogMessage.build("range to build",
```
ScottDugas, line 379: I'm worried this might be too chatty in general. It's probably not a big problem since building mutually is still experimental, but perhaps this indicates that we should (as a followup) add an option to change the logging level for an individual index build.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -514,6 +536,59 @@ private void infiniteLoopProtection(final Range range, final List<Range> missing
         }
     }
 
+    // anyJumper:
+    //  During the ANY iteration, we wish to avoid cases of two indexers competing on the same fragment.
+    //  It is done by recording the exception and the fragment info, then - during the next iteration:
+    //      If at the same fragment, same range: re-throw
+    //      If at the same fragment, different range: jump to the next fragment
```
ScottDugas, line 543: This would be better as a javadoc

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -514,6 +536,59 @@ private void infiniteLoopProtection(final Range range, final List<Range> missing
...
+    boolean anyJumperSaysJump(Range rangeToBuild) {
+        if (anyJumperEx == null || fragmentIterationType != FragmentIterationType.ANY) {
+            return false;
+        }
+        if (LOGGER.isInfoEnabled()) {
+            LOGGER.info(KeyValueLogMessage.build("anyJumper: check if should jump",
```
ScottDugas, line 549: This seems like it should be changed to DEBUG once mutual indexing is a bit less experimental.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -514,6 +536,59 @@ private void infiniteLoopProtection(final Range range, final List<Range> missing
...
+                    .addKeysAndValues(fragmentLogMessageKeyValues())
+                    .toString());
+        }
+        if (anyJumperCurrent != fragmentCurrent) {
+            // Here: this way or another, the conflicting fragment is fully built now. Irrelevant.
+            anyJumperEx = null;
+            return false;
+        }
+        if (anyJumperRange.equals(rangeToBuild)) {
+            // Here: in hindsight, this exception was not caused by a rangeSet conflict. Rethrow it.
```
ScottDugas, line 563: Is this because, if the conflict where a rangeSet conflict, on retry, we would have gotten a different `rangeToBuild`, so if we got this, it means we conflicted on something else?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -84,15 +84,13 @@ public class IndexingThrottle {
...
-                                                          boolean limitControl,
```
ScottDugas, line 87: Why was this removed?
Is `IndexingByRecords.buildEndpoints` the only thing that was passing `false` here? Would that mean that it could retry building the endpoints more than it needs to?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -202,7 +202,7 @@ void testMultiTargetWithTimeQuota() {
         assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
         assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
         // cannot predict exactly how many times the transaction time quota will be reached, but surely more than 0
-        assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_SIZE));
+        assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RANGES_BY_DEPLETION));
```
ScottDugas, line 205: This comment doesn't seem to align with the assertion here

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -135,6 +135,48 @@ void testMutualIndexingNoBoundaries() {
...
+        long numRecords = 180;
+        populateData(numRecords);
+
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        IntStream range = IntStream.rangeClosed(0, 7);
+        AtomicInteger count = new AtomicInteger(0);
+        range.parallel().forEach(ignore -> {
```
ScottDugas, line 158: Since there's only one fragment, this won't actually jump ever, right? Wouldn't we want a test that has like 5 ranges and 20 workers?

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
jjezra: Yes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
jjezra: It was removed to simplify the code a bit. 
`IndexingByRecords.buildEndpoints` will retry just the same, but it *will* increase/decrease limit upon success/failure. It seems harmless to me, especially since  `IndexingByRecords` is now obsolete (single index is now built by the simpler `IndexingMultiTargetByRecords`). 

BTW, when mutual indexing is mature, I hope to retire `IndexingByRecords` altogether. It's only use now is for explicit parallel/range indexing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
jjezra: ✅ 

To reduce noice, will now log a message only when attempting to build this fragment/range.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
jjezra: This only will only log when `anyJumperEx != null`. 
At this point, there are two options: 
1. Jump. In which case, we should probably log something. 
2. Re-throw. In which case, adding an info msg  to a warn one. 
Can we live with it?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: ✅ 

Removed the comment.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: Added. 
However, used 7 workers instead of 20. 
With 20 workers, some of them will not start indexing until the index becomes readable. You had mentioned the need for a framework that supports big db, this may be another use case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
jjezra: ✅

ScottDugas (dismissed):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: Update: 
I had to remove this test. Occasionally it fails (a matter of timing, that we can't control).

ScottDugas (approved):
It would be nice to get better coverage of this, but I think that would require a more powerful testing framework

PR #2090, opened 2023-04-13
Resolves #2089: Cascades planner should support queries on record version

### LINKED ISSUE #2089: Cascades planner should support queries on record version
The Cascades planner currently doesn't support queries on the record version, and it won't look at version indexes during index selection. We should be able to support that in the new planner fairly easily (given how similar version and value indexes are), though it will require some work in the type system.

alecgrieser:
Support for record versions and version indexes are added to the Cascades planner. This does a number of things:

1. A new primitive type, `VERSION`, is added to the type system. This type is used to represent an `FDBRecordVersion`, with some promotion rules to allow it to be compared to `byte[]` and `String` types. It is serialized to `bytes` when included as a field in a message (for example, when included as return type field in a QGM).
1. The existing `VersionValue` is modified to be evaluatble. It now takes a quanitifier and evaluates the version on the record backed by the quantifier.
1. Matching on `version` indexes is added to the index/query matching logic. As `version` indexes behave identically to `value` indexes during query time, this re-uses the matching logic for `value` indexes, with a small number of changes to allow `VersionValue`s to be flowed through. One problem that I hit is that covered index records do not carry their version, even if that was in the index entry, and so some of the logic to support covering index optimizations needed to be modified to stop the optimization from being applied. That is something that we should try to fix, though.

This resolves #2089.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
@@ -186,7 +186,7 @@ public class QueryRecordFunctionWithComparison implements ComponentWithCompariso
 
             return GraphExpansion.ofExists(rankComparisonQuantifier);
         } else if (function instanceof StoreRecordFunction<?> && FunctionNames.VERSION.equals(function.getName())) {
-            VersionValue versionValue = new VersionValue(baseQuantifier.getAlias());
+            final VersionValue versionValue = new VersionValue(baseQuantifier.getAlias());
             return GraphExpansion.builder()
                     .addPredicate(new ValuePredicate(versionValue, comparison))
                     .build();
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
@@ -27,6 +27,7 @@ import com.apple.foundationdb.record.ObjectPlanHash;
 import com.apple.foundationdb.record.PlanHashable;
 import com.apple.foundationdb.record.RecordFunction;
 import com.apple.foundationdb.record.metadata.IndexRecordFunction;
+import com.apple.foundationdb.record.metadata.StoreRecordFunction;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
 import com.apple.foundationdb.record.query.plan.cascades.GraphExpansion;
@@ -38,6 +39,7 @@ import com.apple.foundationdb.record.query.plan.cascades.predicates.ValuePredica
 import com.apple.foundationdb.record.query.plan.cascades.values.FieldValue;
 import com.apple.foundationdb.record.query.plan.cascades.values.QuantifiedObjectValue;
 import com.apple.foundationdb.record.query.plan.cascades.values.RankValue;
+import com.apple.foundationdb.record.query.plan.cascades.values.VersionValue;
 import com.apple.foundationdb.record.util.HashUtils;
 import com.google.common.collect.Lists;
 import com.google.protobuf.Descriptors;
@@ -183,6 +185,11 @@ public class QueryRecordFunctionWithComparison implements ComponentWithCompariso
             }
 
             return GraphExpansion.ofExists(rankComparisonQuantifier);
+        } else if (function instanceof StoreRecordFunction<?> && FunctionNames.VERSION.equals(function.getName())) {
+            VersionValue versionValue = new VersionValue(baseQuantifier.getAlias());
+            return GraphExpansion.builder()
+                    .addPredicate(new ValuePredicate(versionValue, comparison))
+                    .build();
         }
 
         throw new UnsupportedOperationException();
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
@@ -273,6 +273,7 @@ public interface MatchCandidate {
 
         switch (indexType) {
             case IndexTypes.VALUE:
+            case IndexTypes.VERSION:
                 expandIndexMatchCandidate(index,
                     availableRecordTypeNames,
                     availableRecordTypes,
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PromoteValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PromoteValue.java
@@ -56,8 +56,7 @@ public class PromoteValue implements ValueWithChild {
                     Pair.of(Type.TypeCode.INT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Integer)in),
                     Pair.of(Type.TypeCode.LONG, Type.TypeCode.FLOAT), (descriptor, in) -> Float.valueOf((Long)in),
                     Pair.of(Type.TypeCode.LONG, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Long)in),
-                    Pair.of(Type.TypeCode.FLOAT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Float)in)
-            );
+                    Pair.of(Type.TypeCode.FLOAT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Float)in));
     /**
      * The hash value of this expression.
      */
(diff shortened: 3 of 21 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -509,6 +510,7 @@ enum TypeCode {
         INT(Integer.class, FieldDescriptorProto.Type.TYPE_INT32, true, true),
         LONG(Long.class, FieldDescriptorProto.Type.TYPE_INT64, true, true),
         STRING(String.class, FieldDescriptorProto.Type.TYPE_STRING, true, false),
+        VERSION(FDBRecordVersion.class, FieldDescriptorProto.Type.TYPE_BYTES, true, false),
```
normen662, line 513: What happens if you try to put a `VERSION` into a `RECORD`? Is the normal _primitive type_ logic sufficient to deal with this new type so it can figure out how to create the photo descriptors, etc.? Are there tests for this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
```diff
@@ -165,15 +168,23 @@ public static Object deepCopyIfNeeded(@Nonnull TypeRepository typeRepository,
...
-            return field;
+            return wrapPrimitive(field);
         }
 
         if (fieldType instanceof Type.Array) {
+            final var objects = (List<?>)field;
             final var elementType = Verify.verifyNotNull(((Type.Array)fieldType).getElementType());
             if (elementType.isPrimitive()) {
-                return field;
+                if (elementType.getTypeCode() == Type.TypeCode.BYTES || elementType.getTypeCode() == Type.TypeCode.VERSION) {
```
normen662, line 178: Shouldn't this code go to the `eval()` just like the `ARRAY` case?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBVersionsQueryTest.java
```diff
@@ -0,0 +1,411 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests of queries involving predicates on {@link FDBRecordVersion}s. These tests are around to facilitate testing
+ * the planners' version queries, including things like making sure if the version is
+ * For additional tests, see {@link com.apple.foundationdb.record.provider.foundationdb.indexes.VersionIndexTest}.
+ */
+@Tag(Tags.RequiresFDB)
+public class FDBVersionsQueryTest extends FDBRecordStoreQueryTestBase {
```
normen662, line 76: Can we get a test that creates the query graph directly?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
```diff
@@ -183,6 +185,11 @@ public GraphExpansion expand(@Nonnull final Quantifier.ForEach baseQuantifier,
             }
 
             return GraphExpansion.ofExists(rankComparisonQuantifier);
+        } else if (function instanceof StoreRecordFunction<?> && FunctionNames.VERSION.equals(function.getName())) {
+            VersionValue versionValue = new VersionValue(baseQuantifier.getAlias());
```
hatyo, line 189:
```suggestion
            final VersionValue versionValue = new VersionValue(baseQuantifier.getAlias());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
```diff
@@ -273,6 +273,7 @@ static Iterable<MatchCandidate> fromIndexDefinition(@Nonnull final RecordMetaDat
 
         switch (indexType) {
             case IndexTypes.VALUE:
+            case IndexTypes.VERSION:
```
hatyo, line 276: This is more of a design question, why do we need a dedicated index type for version? In other words, isn't serialising the record version along with the record an internal function that can be applied to _any_ value index?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/VersionValue.java
```diff
@@ -22,32 +22,87 @@
...
-public class VersionValue implements LeafValue, Value.CompileTimeValue {
+public class VersionValue implements QuantifiedValue {
     private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Version-Value");
 
-    public VersionValue() {
-        // explicit default constructor
+    @Nonnull
+    private final CorrelationIdentifier baseAlias;
+
+    public VersionValue(@Nonnull CorrelationIdentifier baseAlias) {
```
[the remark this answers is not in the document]
hatyo: > My guess is that you cannot use a regular QuantifiedObjectValue.of(baseAlias) as a child as that one only gives you a Message but you need the actual record? That makes sense and seems to be necessary then but I am not too happy about this.

Me neither, it looks more correct to me to have `QuantifiedObjectValue` as a child. I wonder how we can do this though, does it make sense to (implicitly) flow, when possible, the `version` information of a record in a special `Message` such that `VersionValue` will only project it?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBVersionsQueryTest.java
hatyo: here is an example of such a test `FDBSimpleQueryGraphTest`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
hatyo: Another design question: is it possible to avoid introducing the `VERSION` at all as a primitive type? it seems to me that `version` is merely a function produces a `byte[]` array.

Can't we model it as such, e.g. to make `VersionValue` return a `byte[]`?

alecgrieser:
Rebase addresses merge conflicts. There are some test refactoring of the `FDBSimpleQueryGraphTest` in 9112d44 that is somewhat separable from the rest of the work. Otherwise, all responses to comments are in f11e644 (which makes use of some of the new methods added in 9112d44).

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PromoteValue.java
```diff
@@ -56,7 +56,8 @@ public class PromoteValue implements ValueWithChild {
                     Pair.of(Type.TypeCode.INT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Integer)in),
                     Pair.of(Type.TypeCode.LONG, Type.TypeCode.FLOAT), (descriptor, in) -> Float.valueOf((Long)in),
                     Pair.of(Type.TypeCode.LONG, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Long)in),
-                    Pair.of(Type.TypeCode.FLOAT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Float)in));
+                    Pair.of(Type.TypeCode.FLOAT, Type.TypeCode.DOUBLE), (descriptor, in) -> Double.valueOf((Float)in)
+            );
```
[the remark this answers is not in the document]
alecgrieser: Whoops, looks like that's carried over from when I made more substantive changes to this file. Reverting

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ValueMatchers.java
```diff
@@ -150,4 +151,9 @@ public static BindingMatcher<StreamableAggregateValue> streamableAggregateValue(
                 Extractor.of(StreamableAggregateValue::getChildren, name -> "children(" + name + ")"),
                 downstreamValues);
     }
+
+    @Nonnull
+    public static BindingMatcher<VersionValue> versionValue() {
+        return typedWithDownstream(VersionValue.class, Extractor.identity(), anyValue());
```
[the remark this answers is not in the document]
alecgrieser: Switched to just using `typed`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -141,7 +144,29 @@ public <M extends Message> Object eval(@Nonnull final FDBRecordStoreBase<M> stor
         // If the last step in the field path is an array that is also nullable, then we need to unwrap the value
         // wrapper.
         //
-        return NullableArrayTypeUtils.unwrapIfArray(fieldValue, getResultType());
+        return wrapPrimitive(getResultType(), NullableArrayTypeUtils.unwrapIfArray(fieldValue, getResultType()));
+    }
+
+    @Nullable
+    private static Object wrapPrimitive(@Nonnull Type type, @Nullable Object fieldValue) {
```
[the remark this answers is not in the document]
alecgrieser: Yeah, I'm not sure either `wrap` or `unwrap` are quite right...it's almost more of a "cast" primitive. But I've switched to unwrap so that it's at least consistent with `NullableArrayTypeUtils.unwrapIfArray`

alecgrieser (commented):
Rebased to get on top of latest `main`, though it ended up merging cleanly. Changes in response to comments are in af3850ebf7cec7bb878779797d2b78b374

normen662 (approved):

alecgrieser:
Discussed with @hatyo offline, and he's okay with merging. He may take another look later

PR #2092, opened 2023-04-13
Add support for `int32` in record metadata `Value`

### LINKED ISSUE #2091: record metadata's `Value` does not support `int32`
Currently decimal values are treated as `int64`, this could silently change the semantics of an `IndexPredicate` with a `SimpleComparison` that has an operand of type `Integer` as it is widening it to long.

hatyo:
Currently decimal values are treated as `int64`, this could silently change the semantics of an `IndexPredicate` with a `SimpleComparison` that has an operand of type `Integer` as it is widening it into `long`.

This adds support for `int32`, it also adds a unit test that verifies that the serialisation / deserialisation of comparison parameters yield identical results. it fixes #2091 and unblocks https://github.com/FoundationDB/fdb-record-layer/pull/2070.


MMcM (changes requested):
There is an `incorrectIntegerSerializationTest`, which can now be merged into `serializationTest`.

hatyo:
> There is an `incorrectIntegerSerializationTest`, which can now be merged into `serializationTest`.

Done.

normen662 (approved):

MMcM (approved):

PR #2093, opened 2023-04-13
Ensure that Synthetic Record Types are not updated if the underlying indexes are all disabled

scottfines:
Addresses https://github.com/FoundationDB/fdb-record-layer/issues/2099

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -15,7 +15,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Don't write to disabled Synthetic Indexes [(Issue #2099)](https://github.com/FoundationDB/fdb-record-layer/issues/2099)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -655,6 +655,10 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                 return;
             }
             final Map<RecordType, Collection<IndexMaintainer>> maintainers = getSyntheticMaintainers(plan.getSyntheticRecordTypes());
+            if (maintainers.isEmpty()) {
+                //don't exec anything if there are no maintainers
+                return;
+            }
             final Map<Tuple, FDBSyntheticRecord> oldRecords = new ConcurrentHashMap<>();
             CompletableFuture<Void> future = plan.execute(this, oldRecord).forEach(syntheticRecord -> oldRecords.put(syntheticRecord.getPrimaryKey(), syntheticRecord));
             @Nonnull final FDBStoredRecord<M> theNewRecord = newRecord; // @SpotBugsSuppressWarnings("NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE", justification = "https://github.com/spotbugs/spotbugs/issues/552")
@@ -682,14 +686,18 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                 final SyntheticRecordFromStoredRecordPlan plan = planner.fromStoredType(oldRecord.getRecordType(), true);
                 if (plan != null) {
                     final Map<RecordType, Collection<IndexMaintainer>> maintainers = getSyntheticMaintainers(plan.getSyntheticRecordTypes());
-                    futures.add(plan.execute(this, oldRecord).forEachAsync(syntheticRecord -> runSyntheticMaintainers(maintainers, syntheticRecord, null), pipelineSize));
+                    if (!maintainers.isEmpty()) {
+                        futures.add(plan.execute(this, oldRecord).forEachAsync(syntheticRecord -> runSyntheticMaintainers(maintainers, syntheticRecord, null), pipelineSize));
+                    }
                 }
             }
             if (newRecord != null) {
                 final SyntheticRecordFromStoredRecordPlan plan = planner.fromStoredType(newRecord.getRecordType(), true);
                 if (plan != null) {
                     final Map<RecordType, Collection<IndexMaintainer>> maintainers = getSyntheticMaintainers(plan.getSyntheticRecordTypes());
-                    futures.add(plan.execute(this, newRecord).forEachAsync(syntheticRecord -> runSyntheticMaintainers(maintainers, null, syntheticRecord), pipelineSize));
+                    if (!maintainers.isEmpty()) {
+                        futures.add(plan.execute(this, newRecord).forEachAsync(syntheticRecord -> runSyntheticMaintainers(maintainers, null, syntheticRecord), pipelineSize));
+                    }
                 }
             }
         }
@@ -701,8 +709,8 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         final RecordMetaData metaData = getRecordMetaData();
         return syntheticRecordTypes.stream().map(metaData::getSyntheticRecordType).collect(Collectors.toMap(Function.identity(), syntheticRecordType -> {
             List<IndexMaintainer> indexMaintainers = new ArrayList<>();
-            syntheticRecordType.getIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
-            syntheticRecordType.getMultiTypeIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+            syntheticRecordType.getIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+            syntheticRecordType.getMultiTypeIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
             return indexMaintainers;
         }));
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -707,12 +707,17 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     @API(API.Status.EXPERIMENTAL)
     private Map<RecordType, Collection<IndexMaintainer>> getSyntheticMaintainers(@Nonnull Set<String> syntheticRecordTypes) {
         final RecordMetaData metaData = getRecordMetaData();
-        return syntheticRecordTypes.stream().map(metaData::getSyntheticRecordType).collect(Collectors.toMap(Function.identity(), syntheticRecordType -> {
-            List<IndexMaintainer> indexMaintainers = new ArrayList<>();
-            syntheticRecordType.getIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
-            syntheticRecordType.getMultiTypeIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
-            return indexMaintainers;
-        }));
+        Map<RecordType, Collection<IndexMaintainer>> map = syntheticRecordTypes.stream()
+                .map(metaData::getSyntheticRecordType).collect(Collectors.toMap(Function.identity(), syntheticRecordType -> {
+                    List<IndexMaintainer> indexMaintainers = new ArrayList<>();
+                    syntheticRecordType.getIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+                    syntheticRecordType.getMultiTypeIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+                    return indexMaintainers;
+                }));
+
+        map.entrySet().removeIf(entry -> entry.getValue().isEmpty());
+
+        return map;
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
@@ -38,6 +38,7 @@ import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.IndexOptions;
 import com.apple.foundationdb.record.metadata.IndexRecordFunction;
 import com.apple.foundationdb.record.metadata.IndexTypes;
+import com.apple.foundationdb.record.metadata.IndexValidator;
 import com.apple.foundationdb.record.metadata.JoinedRecordType;
 import com.apple.foundationdb.record.metadata.JoinedRecordTypeBuilder;
 import com.apple.foundationdb.record.metadata.Key;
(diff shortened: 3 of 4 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1040,6 +1049,90 @@ public void aggregateJoinIndex() throws Exception {
...
+
+            TestRecordsJoinIndexProto.OrderWithHeader order = TestRecordsJoinIndexProto.OrderWithHeader.newBuilder()
+                    .setHeader(TestRecordsJoinIndexProto.Header.newBuilder().setZKey(1).setRecId("23"))
+                    .setOrderNo(10)
+                    .setQuantity(23)
+                    .setCustRef(TestRecordsJoinIndexProto.Ref.newBuilder().setStringValue("i:1"))
+                    .build();
+            recordStore.saveRecord(order);
+        }
+    }
```
alecgrieser, line 1134: This test seems like it mostly tests what we want, in that the test should fail if the buggy maintainer gets called. The other way I can think to test this would be to create a normal join index, disable it, then insert some records that would generate entries if the index wasn't disabled, and then call `uncheckedMarkIndexReadable` to force the index into a readable state without issuing a build. Then, the index should be scannable (and the test can validate no entries were written).

The other thing that we may want to test: I could be wrong, but I _think_ the current approach will not even try and find the join constituents on a record if the index(es) it would be used for are `DISABLED`. If we wanted to test that, then I think we could do that in one of two ways:

1. Create a join that the `SyntheticRecordPlanner` can't correctly calculate joins for. For example, using non-invertible `FunctionKeyExpression` in the join constituent might do it
2. Modify the `SyntheticRecordPlanner` so that it will return plans that throw errors when run. That's a little more difficult, given how the `SyntheticRecordPlanner` is constructed, though maybe adding in enough hooks is possible. An alternative alternative to that might be to add metrics surrounding the amount of time being spent loading synthetic records (which are metrics we may want anyway), and then asserting that these metrics are 0 if the index is DISABLED

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -701,8 +709,8 @@ private Map<RecordType, Collection<IndexMaintainer>> getSyntheticMaintainers(@No
         final RecordMetaData metaData = getRecordMetaData();
         return syntheticRecordTypes.stream().map(metaData::getSyntheticRecordType).collect(Collectors.toMap(Function.identity(), syntheticRecordType -> {
             List<IndexMaintainer> indexMaintainers = new ArrayList<>();
-            syntheticRecordType.getIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
-            syntheticRecordType.getMultiTypeIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+            syntheticRecordType.getIndexes().stream().filter(index->!isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
```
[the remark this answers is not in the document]
scottfines: it does not appear to be called outside of `updateSyntheticIndexes`, no

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1504,4 +1597,20 @@ private static void assertConstituentPlansMatch(SyntheticRecordPlanner planner,
             assertThat(plan, matcher);
         }
     }
+
+    private static class FailIndexMaintainer extends TerribleIndexMaintainer{
```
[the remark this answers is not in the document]
scottfines: That feels like an extremely heavy-weight solution to the problem of wanting to quickly fail.

Just as a quick attempt, I ended up writing three classes and several hundred lines of code to get it into a ServiceLoader, whereas this particular change is fairly straightforward and obvious. 

Perhaps a cleaner approach would be to push this mocking logic into some kind of utility class, so that it's not sitting in the test code itself?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
scottfines: the current structure of the code requires the underlying RL planner to be able to create a plan for both sides of the join _before_ it determines if the indexes are enabled or not. This is arguably the incorrect order--we probably want to not even attempt to plan the join elements if all the destination indexes are disabled. But that is how it currently stands.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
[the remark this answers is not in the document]
alecgrieser: Hm, it's maybe not the lightest weight solution I can imagine, but I'm surprised you needed three classes... It would seem that you'd need:

1. An `IndexMaintainer` implementation. For this purpose, sub-classing `ValueIndexMaintainer` and overriding just `updateIdexKeys` would seem to be sufficient
2. An `IndexMaintainerFactory` implementation for the new maintainer. The only tricky thing about that one is it needs to be marked as `@AutoService(IndexMaintainerFactory.class)`.

With those two things, the index type should now be picked up by the registry. Not counting imports and whitespace, I wouldn't expect it to exceed...100 lines? About the same order of magnitude as what is here, and more in line with the other tests that do this kind of thing

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
alecgrieser: But the plans don't get executed in that case, right? So both my proposals to either have plans that throw errors when run or asserting that timer values associated with running the plans are zero would be consistent with the current code structure, I'd think

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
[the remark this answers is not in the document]
scottfines: I mispoke--I had modified three files (the test META-INF file and the other two). And a lot of lines were documentation to explain how it works and why, because it felt very disconnected from the one test it was going to be used in, so I felt it needed additional explanation.

But also, for one test it just seems like a lot to be adding in the service loader pattern here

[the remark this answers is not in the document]
alecgrieser: I see. Well, using a mocked out registry like this isn't the worst, but it does give me a bit of pause, mainly for code uniformity with the other times we wanted a non-standard index maintainer.

In terms of polluting the service loader, we actually do that fairly often in tests. For example, there are a fair number of `FunctionKeyExpression`s that get added to the service loader for one or two tests.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -701,8 +709,8 @@ private Map<RecordType, Collection<IndexMaintainer>> getSyntheticMaintainers(@No
         final RecordMetaData metaData = getRecordMetaData();
         return syntheticRecordTypes.stream().map(metaData::getSyntheticRecordType).collect(Collectors.toMap(Function.identity(), syntheticRecordType -> {
             List<IndexMaintainer> indexMaintainers = new ArrayList<>();
-            syntheticRecordType.getIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
-            syntheticRecordType.getMultiTypeIndexes().stream().map(this::getIndexMaintainer).forEach(indexMaintainers::add);
+            syntheticRecordType.getIndexes().stream().filter(index -> !isIndexDisabled(index)).map(this::getIndexMaintainer).forEach(indexMaintainers::add);
```
ScottDugas, line 712: This got me thinking about other index states, do we need to call `updateWhileWriteOnly` if the index is in `writeOnly`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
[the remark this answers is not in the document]
scottfines: I rewrote the test to force-mark the index as readable

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1504,4 +1589,20 @@ private static void assertConstituentPlansMatch(SyntheticRecordPlanner planner,
             assertThat(plan, matcher);
         }
     }
+
+    private static class FailIndexMaintainer extends TerribleIndexMaintainer {
```
alecgrieser, line 1593: Do we even need this class any more?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
alecgrieser: Yes, we probably do want to call `updateWhileWriteOnly` in that case. For most maintainers, if the index is idempotent, that works out to be the same thing, but yeah, I think if you had a `COUNT` index (say) defined on a synthetic record type, then this won't correctly update the index while the index is being built.

That being said, I'm not _actually_ sure that `COUNT` indexes on synthetic record types can be built at all. In particular, I think it might actually double count any existing record, if I'm reading the logic in `IndexingBase` correctly:

https://github.com/FoundationDB/fdb-record-layer/blob/d4478d6d8c39760a096f7217c94f54a617adf0cb/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java#L880-L897

And in any case, the logic to determine whether to index a record or not would have to be adjusted, essentially because synthetic record types aren't indexed in primary key order.

We may want to add `IndexValidation` logic to stop us from building non-idempotent indexes on synthetic types. That feels like it may belong in a separate PR. It may be worth switching `runSyntheticMaintainers` in this PR to call `update` or `updateWhileWriteOnly` as appropriate, though.

alecgrieser (changes requested):

alecgrieser:
Could you associate this PR with a GitHub issue, and update the release notes? This feels like the kind of bug fix we'd want to publicize as well as easily validate (from the release notes) which builds have the fix

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
scottfines: Considering the time constraints on this, and current supported indexes (basically only lucene), I'd suggest that we file a GitHub issue on this and approach it at the same time we expand support for COUNT indexes over synthetic types

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1040,6 +1045,86 @@ public void aggregateJoinIndex() throws Exception {
         }
     }
 
+    @Test
+    void wontUpdateSyntheticTypeIfUnderlyingIndexesAreDisabled() throws Exception {
```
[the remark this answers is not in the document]
scottfines: This is a potential bug, but in practice the usage pattern of testing this in other locations does this pattern I believe. Would suggest that we add additional testing in a separate PR

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1040,6 +1045,86 @@ public void aggregateJoinIndex() throws Exception {
...
+        )));
+
+        try (FDBRecordContext context = openContext()) {
+            final FDBRecordStore recordStore = recordStoreBuilder.setContext(context).create();
+            final Index joinIndex = recordStore.getRecordMetaData().getIndex("joinNestedConcat");
+            recordStore.markIndexDisabled(joinIndex).get();
+            context.commit();
+        }
+
+        try (FDBRecordContext context = openContext()) {
```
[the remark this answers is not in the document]
scottfines: Added two tests--one for delete, and one for update when indexes are write-only

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
scottfines: Removed, thanks for the catch

docs/ReleaseNotes.md
```diff
@@ -41,6 +41,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 
 ### 3.3.365.0
 
+* **Bug fix** Don't write to disabled Synthetic indexes [(Issue #2099)](https://github.com/FoundationDB/fdb-record-layer/issues/2099)
```
alecgrieser, line 44: This should go in the commented out section above rather than here (given that it is _not_ a feature of 3.3.365.0)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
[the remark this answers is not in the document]
scottfines: I agree that's worth doing in a follow-up PR

alecgrieser (changes requested):

docs/ReleaseNotes.md
```diff
@@ -39,6 +39,9 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // end next release
 -->
 
+### 3.3.366.0
+* **Bug fix** Don't write to disabled Synthetic indexes [(Issue #2099)](https://github.com/FoundationDB/fdb-record-layer/issues/2099)
```
alecgrieser, line 43: I think this still isn't quite right. This needs to go in the commented out section, replacing one of the bullets that is already there. The release process generates the appropriate bullet point with the correct version number (which probably will be 3.3.366.0, but there's no guarantee). You can see an example of how that's done here, for instance: https://github.com/FoundationDB/fdb-record-layer/pull/2064/files

alecgrieser (changes requested):

docs/ReleaseNotes.md
scottfines: oof. got it

alecgrieser (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1040,6 +1041,225 @@ public void aggregateJoinIndex() throws Exception {
...
+            final FDBRecordStore recordStore = recordStoreBuilder.setContext(context).open();
+            TestRecordsJoinIndexProto.CustomerWithHeader customer = TestRecordsJoinIndexProto.CustomerWithHeader.newBuilder()
+                    .setHeader(TestRecordsJoinIndexProto.Header.newBuilder().setZKey(1).setIntRecId(2L))
+                    .setName("Scott")
+                    .setCity("Toronto")
+                    .build();
+            recordStore.saveRecord(customer);
+
+            TestRecordsJoinIndexProto.OrderWithHeader order = TestRecordsJoinIndexProto.OrderWithHeader.newBuilder()
```
ScottDugas, line 1140: I like this test, but I don't think it covers the case of updating a record while in write-only mode. It creates a second record. In order to cover updates, you'd have to save a record with the same header, but change the name or orderNo. And in order to actually cover updates on a write-only index you have to assert about the values of the index.
For this PR, I think the tests as you have them are probably fine, but we should add more coverage of indexes on synthetic types in various situations.
I was a bit thrown off because you have `updateSyntheticIndexesWhenInWriteOnly` and `deleteFromSyntheticIndexWhenIndexIsWriteOnly` so I thought the former was for updating a record, rather than simply updating the index.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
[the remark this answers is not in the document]
ScottDugas: I guess I wasn't clear, but I think the priority is to add a test that updates a record while the index is disabled, and a test that deletes a record while the index is disabled. There are 3 paths that you changed, and the test here only covers one of them. 

In my head, I thought the write-only tests were a lower priority, but there here, so worth keeping.

ScottDugas (changes requested):

alecgrieser (approved):

ScottDugas (approved):

alecgrieser (approved):

ScottDugas (approved):

PR #2097, opened 2023-04-18
add static method to load store header

pengpeng-lu:
Add this static method to load store header without opening the store.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -2107,6 +2107,30 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                 });
     }
 
+    private static RecordMetaDataProto.DataStoreInfo checkAndParseStoreHeader(@Nullable KeyValue firstKeyValue,
+                                                                              @Nonnull StoreExistenceCheck existenceCheck,
+                                                                              @Nonnull FDBRecordContext context,
+                                                                              @Nonnull SubspaceProvider subspaceProvider,
+                                                                              @Nonnull Subspace subspace) {
+        RecordMetaDataProto.DataStoreInfo info;
+        if (firstKeyValue == null) {
+            info = RecordMetaDataProto.DataStoreInfo.getDefaultInstance();
+        } else if (!checkFirstKeyIsHeader(firstKeyValue, context, subspaceProvider, subspace, existenceCheck)) {
+            // Treat as brand new, although there is no way to be sure that what was written is compatible
+            // with the current default versions.
+            info = RecordMetaDataProto.DataStoreInfo.getDefaultInstance();
+        } else {
+            try {
+                info = RecordMetaDataProto.DataStoreInfo.parseFrom(firstKeyValue.getValue());
+            } catch (InvalidProtocolBufferException ex) {
+                throw new RecordCoreStorageException("Error reading version", ex)
+                        .addLogInfo(subspaceProvider.logKey(), subspaceProvider.toString(context));
+            }
+        }
+        checkStoreHeaderInternal(info, context, subspaceProvider, existenceCheck);
+        return info;
+    }
+
     @Nonnull
     private RecordMetaDataProto.DataStoreInfo checkAndParseStoreHeader(@Nullable KeyValue firstKeyValue,
                                                                        @Nonnull StoreExistenceCheck existenceCheck) {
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2380,6 +2404,12 @@ private void endRecordStoreStateWrite() {
         });
     }
 
+    public static CompletableFuture<RecordMetaDataProto.DataStoreInfo> loadStoreHeaderAsync(@Nonnull FDBRecordContext context, @Nonnull KeySpacePath keySpacePath) {
+        SubspaceProvider subspaceProvider1 = new SubspaceProviderByKeySpacePath(keySpacePath);
+        Subspace subspace1 = subspaceProvider1.getSubspace(context);
+        return readStoreFirstKey(context, subspace1, IsolationLevel.SERIALIZABLE).thenApply(keyValue -> checkAndParseStoreHeader(keyValue, StoreExistenceCheck.NONE, context, subspaceProvider1, subspace1));
```
ScottDugas, line 2410: If this takes the `StoreExistenceCheck` as a parameter, acn more of this be shared between the static and non-static version of `loadStoreHeaderAsync`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2380,6 +2404,12 @@ private void endRecordStoreStateWrite() {
         });
     }
 
+    public static CompletableFuture<RecordMetaDataProto.DataStoreInfo> loadStoreHeaderAsync(@Nonnull FDBRecordContext context, @Nonnull KeySpacePath keySpacePath) {
```
ScottDugas, line 2407: This should have a javadoc
It should probably also be marked as `INTERNAL`.

ScottDugas, line 2407: This change should come with a github issue, and an update to the release notes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3237,6 +3267,49 @@ private CompletableFuture<RecordStoreState> loadRecordStoreStateInternalAsync(@N
         return context.instrument(FDBStoreTimer.Events.LOAD_RECORD_STORE_STATE, storeHeaderFuture.thenCombine(loadIndexStates, RecordStoreState::new));
     }
 
+    @Nonnull
+    @SuppressWarnings("PMD.CloseResource")
+    public static CompletableFuture<Map<String, IndexState>> loadIndexStatesAsync(@Nonnull Subspace subspace, @Nonnull FDBRecordContext context) {
```
ScottDugas, line 3272: It would be better to have the non-static version call this version, and remove the duplication of logic here.
Also, this method should have a javadoc and probably be marked as `INTERNAL`

ScottDugas: Particularly of note for the javadoc is that readable indexes are not included in the returned map

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -209,6 +210,24 @@ public void prefixPrimaryKeysWithNullByteAfterPrefix(int formatVersion, boolean
         }
     }
 
+    @Test
+    public void testLoadStoreHeader() throws Exception {
```
ScottDugas, line 214: There should also be a test for loading the index statuses

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -209,6 +210,24 @@ public void prefixPrimaryKeysWithNullByteAfterPrefix(int formatVersion, boolean
...
+        try (FDBRecordContext context = openContext()) {
+            openSimpleRecordStore(context);
+            // load store header with new static method
+            RecordMetaDataProto.DataStoreInfo dataStoreInfo1 = FDBRecordStore.loadStoreHeaderAsync(context, path).get();
+            // load store header with old method
+            RecordMetaDataProto.DataStoreInfo dataStoreInfo2 = recordStore.getRecordStoreState().getStoreHeader();
+            // assert
```
ScottDugas, line 221: `RecordMetaDataProto.DataStoreInfo` is a protobuf so it would be better to just assert that the two `datastoreInfo`s are equals, so that this test maintains coverage if those protobuf definitions add fields.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -209,6 +210,24 @@ public void prefixPrimaryKeysWithNullByteAfterPrefix(int formatVersion, boolean
...
+            Assertions.assertEquals(FDBRecordStore.MAX_SUPPORTED_FORMAT_VERSION, dataStoreInfo1.getFormatVersion());
+            Assertions.assertEquals(FDBRecordStore.MAX_SUPPORTED_FORMAT_VERSION, dataStoreInfo2.getFormatVersion());
+            Assertions.assertEquals(recordStore.getRecordMetaData().getVersion(), dataStoreInfo1.getMetaDataversion());
+            Assertions.assertEquals(recordStore.getRecordMetaData().getVersion(), dataStoreInfo2.getMetaDataversion());
+            Assertions.assertEquals(dataStoreInfo1.getLastUpdateTime(), dataStoreInfo2.getLastUpdateTime());
```
ScottDugas, line 226: It's probably overkill, but it might be good to change the store header and assert that getting it still gets the same value.
Perhaps more relevant for getting the index states than the store header

ScottDugas (changes requested):

PR #2098, opened 2023-04-19
Rebuild Lucene Analyzers and Highlighting

scottfines:
This PR does the following primary tasks:

1. Rebuilds Lucene Highlighting to rely on Lucene's `UnifiedHighlighter` concept, which makes use of the actual query searched on to provide highlighting logic. It allows for improved features like limiting the number of matched highlight passages and the ability to use different analyzers, and is generally more reliable overall.
2. Simplifies and systematizes the API for Lucene-based highlighting, so that the call stack in `LuceneHighlighting` is clearer and more consistent with its usage. 
3. Provides a final configuration pattern for acceptable Lucene analyzer behavior (in terms of email addresses, long tokens, non-english characters, and so forth).
3. Adds significant testing for Highlighting logic, checking for a variety of different necessary conditions in a compact form. This testing has the side-effect of revealing and explaining different lucene analyzer behavior, which further helps to cement the expected behavior of both highlighting _and_ query matching.
4. Moves integration highlighting tests into its own class and providing shared utility classes to make testing simpler.

Along the way, this resolves a number of smaller bugs both within highlighting and within our usage pattern of analyzers in general.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/highlight/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Highlighting of matched terms found in using record-layer lucene integration.
+ */
+package com.apple.foundationdb.record.lucene.highlight;
```
normen662, line 24: can you annotate with `INTERNAL`?

normen662 (commented):
I think we should have a small chat about it when you get back to the office. I think this is structurally sound as in I think this is how this should be done and integrated into the existing code but I don't claim I understand the semantic nuances when it comes to these analyzers, etc.

normen662 (approved):

PR #2101, opened 2023-04-20
Resolve #2100: A bug in mutual indexing may have cause low performance

### LINKED ISSUE #2100: A bug in mutual indexing may have cause low performance
When throttle catches a non-fdb exception, it passes null to mutual indexing callback - which causes many repetitions on the same rangeSet snapshot (limited by the retry counter).

jjezra:
Changes are: 
1. Apply anyJumper to FULL iteration as well. This is done to prevent the inner loop from repeating indexing attempt when the range set is changing. 
2. Handle potential null fdb error provided to the anyJumper callback. 
3. Force anyJumperSaysBuild to either throw an exception or clear anyJumperEx.
4. Clear anyJumperEx if the new range is null. 
5. Fix a bug in the infinite loop protection.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -15,7 +15,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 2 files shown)


jjezra:
Rebased, resolved conflicts.

jjezra:
`gw fdb-extensions:compileTestJava ` works in my environment. 

Test this please

jjezra:
Test this please

alecgrieser (approved):
This looks fine to me, though ideally it would use a few more tests. Given the experimental nature of the code, it's probably fine, but we should probably consider at some point how to increase the coverage of this class

ScottDugas (approved):

Issue #2104, opened 2023-04-27
Captured constraints of filtered index are too restrictive


hatyo:
When planning a query and matching its predicate against a filtered index predicate, we capture the filtered index predicate as a plan constraint. The way this works is a bit involved, but it can be summerized using the following example:

consider we have the following filtered index:

```sql
select * from t1 where col1 > 42
```

and the following query:
```sql
select * form t1 where col1 > 50;
```

In the query plan, the predicate, which is something like `col1 > ConstantObjectValue[0] | EvaluationContext = [const, [50]]`, is clearly implied by the index filter predicate, so we generate a constraint that is effectively a predicate with `ConstantObjectValue[0]` on the LHS and the index filter `> 42` on the RHS, i.e. `ConstantObjectValue[0] > 42`.

At first glance, this looks ok, however note that this constraint prevents a query like `select * from t1 where col1 > 42` from reusing the physical plan, although it matches the filtered index definition one to one!

The reason behind this is that we simple reorganize the predicate above without realizing what we _really_ want to achieve is this implication (for the example above): `col1 > ConstantObjectValue[0] => col1 > 42`, solving this implication for `ConstantObjectValue[0]` gives the result `ConstantObjectValue[0] : [42, +inf)` (and not `(42, +inf)`).

PR #2105, opened 2023-04-27
Fixes for filtered indexes

hatyo:
This provides few fixes pertaining filtered indexes:

* Fix for an issue with a flat conjunction not being recognized as a DNF.
* Fix for another issue with the captured constraints of filtered indexes being very strict (solves #2104).
* Finally, it adds `toString` to some of constructs used during filtered index matching, making debugging somewhat easier.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/RecordTypeKeyComparison.java
```diff
@@ -180,6 +180,12 @@ public Comparisons.Type getType() {
             return Comparisons.Type.EQUALS;
         }
 
+        @Nonnull
+        @Override
+        public Comparisons.Comparison withType(@Nonnull final Comparisons.Type newType) {
```
normen662, line 185: This should error out if it is used for anything other than the current type.

normen662 (changes requested):
Left comment as a single comment. Ugh!

normen662 (approved):

PR #2110, opened 2023-05-03
Resolve #2109: FDBRecordStore: keep userVersionChecker during copyFro…

### LINKED ISSUE #2109: FDBRecordStore: keep `userVersionChecker` during copyFrom store

jjezra:
…m store

 Without this PR, creating a store from another store (by something like `recordStore.asBuilder()`) will omit `userVersionChecker`.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -263,6 +263,9 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     protected final FDBRecordStoreStateCache storeStateCache;
 
     @Nullable
+    private final FDBRecordStoreBase.UserVersionChecker userVersionChecker;
+
+    @Nullable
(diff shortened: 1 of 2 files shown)


MMcM (approved):

jjezra:
Test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4711,6 +4716,7 @@ public void copyFrom(@Nonnull FDBRecordStore store) {
             this.metaDataProvider = store.metaDataProvider;
             this.context = store.context;
             this.subspaceProvider = store.subspaceProvider;
+            this.userVersionChecker = store.userVersionChecker;
```
ScottDugas, line 4719: I think technically, there could be implementations of `UserVersionChecker` that store state, that could be confused about being reused on multiple store objects. That being said, re-reading the documentation for that interface, it seems like all the javadoc implies that it shouldn't be reused. 

It would also be good to follow up with more interesting coverage of `UserVersionChecker`, it is clear that we don't have good coverage if we drop the checker when converting to a builder and then back.

ScottDugas (approved):

PR #2117, opened 2023-05-09
A couple of additions to FDBRecordContextConfig

### LINKED ISSUE #2115: Expose tags in FDBRecordContextConfig
Allow caller to specify string(s) to be placed in underlying transaction.

### LINKED ISSUE #2116: Expose conflict tracking to FDBRecordContextConfig
Allow caller to request that conflict keys are tracked by the underlying transaction and preserved in the process of throwing any not-committed exception.

MMcM:
* Resolves #2115: Add throttling tags to context config.
* Resolves #2116: Add optional conflict tracking to context config.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
@@ -23,6 +23,7 @@ package com.apple.foundationdb.record.provider.foundationdb;
 import com.apple.foundationdb.FDBError;
 import com.apple.foundationdb.FDBException;
 import com.apple.foundationdb.MutationType;
+import com.apple.foundationdb.Range;
 import com.apple.foundationdb.ReadTransaction;
 import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.annotation.API;
@@ -52,6 +53,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.nio.charset.Charset;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -60,6 +62,7 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutionException;
@@ -159,6 +162,8 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
     private long trackOpenTimeNanos;
     @Nonnull
     private final Map<Object, Object> session = new LinkedHashMap<>();
+    @Nullable
+    private List<Range> notCommittedConflictingKeys = null;
 
     @SuppressWarnings("PMD.CloseResource")
     protected FDBRecordContext(@Nonnull FDBDatabase fdb,
@@ -188,6 +193,10 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
             for (String tag : config.getTags()) {
                 tr.options().setTag(tag);
             }
+        }
+
+        if (config.isReportConflictingKeys()) {
+            tr.options().setReportConflictingKeys();
         }
 
         this.config = config;
@@ -468,6 +477,21 @@ public class FDBRecordContext extends FDBTransactionContext implements AutoClose
                 return AsyncUtil.DONE;
             }
         });
+        if (config.isReportConflictingKeys()) {
+            // Do this before the close (below).
+            commit = MoreAsyncUtil.composeWhenCompleteAndHandle(commit, (v, ex) -> {
+                if (ex instanceof CompletionException) {
+                    final Throwable cause = ex.getCause();
+                    if (cause instanceof FDBException) {
+                        final FDBException fdbException = (FDBException)cause;
+                        if (FDBError.fromCode(fdbException.getCode()) == FDBError.NOT_COMMITTED) {
+                            return readConflictingKeys();
+                        }
+                    }
+                }
+                return AsyncUtil.DONE;
+            }, ex -> ex instanceof RuntimeException ? (RuntimeException)ex : new RecordCoreException(ex));
+        }
         return commit.whenComplete((v, ex) -> {
             StoreTimer.Event event = FDBStoreTimer.Events.COMMIT;
             try {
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
```diff
@@ -597,6 +628,39 @@ public Builder setRecordContextProperties(@Nonnull final RecordLayerPropertyStor
...
+         */
+        public boolean isReportConflictingKeys() {
+            return reportConflictingKeys;
+        }
+
+        /**
+         * Set whether to report conflicting key rangs on commit conflict.
+         * @param reportConflictingKeys {@code true} to report conflicting keys
+         */
+        public void setReportConflictingKeys(final boolean reportConflictingKeys) {
```
alecgrieser, line 660: Should these two new setters return `Builder`s to make them more fluent?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
```diff
@@ -597,6 +628,39 @@ public Builder setRecordContextProperties(@Nonnull final RecordLayerPropertyStor
...
+        public Set<String> getTags() {
+            return tags;
+        }
+
+        /**
+         * Set tags used for throttling.
+         * @param tags new set of tags
+         */
+        public void setTags(@Nonnull final Set<String> tags) {
+            this.tags = tags;
```
alecgrieser, line 645: I guess I was a little surprised that the builder here took a `Set<String>` instead of having an `addTag` method. As written, it looks like this has the advantage that it doesn't need to allocate any objects if the (default) empty set is used, which nice, so maybe this is preferable.

Reading the docs, it also sounds like each transaction can have at most 5 tags, and each tag should be 16 characters or less. Given that FDB's own error handling here makes it hard to associate an error from option value validation with the origin, should this method validate the options (and throw if invalid)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -1397,4 +1427,31 @@ public RecordLayerPropertyStorage getPropertyStorage() {
...
+    @Nullable
+    public List<Range> getNotCommittedConflictingKeys() {
+        return notCommittedConflictingKeys;
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private CompletableFuture<Void> readConflictingKeys() {
+        notCommittedConflictingKeys = new ArrayList<>();
+        final Transaction tr = ensureActive();
+        tr.options().setReadSystemKeys();
```
alecgrieser, line 1443: Is this option necessary? I thought the special keyspace (i.e., the `\xff\xff` keyspace) didn't need it, unlike the non-special system keyspace (i.e., the `\xff` keyspace). I could have that wrong, though

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -1397,4 +1427,31 @@ public RecordLayerPropertyStorage getPropertyStorage() {
...
+     * @return list of conflict ranges
+     */
+    @Nullable
+    public List<Range> getNotCommittedConflictingKeys() {
+        return notCommittedConflictingKeys;
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private CompletableFuture<Void> readConflictingKeys() {
+        notCommittedConflictingKeys = new ArrayList<>();
```
alecgrieser, line 1441: It seems a bit strange that this sets the instance variable at the beginning of the method and then updates the instance variable asynchronously. It feels like this should have a local variable, and then when the async callback completes, ether it returns the list (and `commit()` sets the instance variable) or the instance variable gets set within this method. That would then mean that the `notCommittedConflictingKeys` variable would either be `null` or would contain a complete set of conflicts (rather than certain incorrect usage resulting in the viewer getting a non-`null` but incomplete set of conflicts).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -462,6 +477,21 @@ public CompletableFuture<Void> commitAsync() {
                 return AsyncUtil.DONE;
             }
         });
+        if (config.isReportConflictingKeys()) {
+            // Do this before the close (below).
+            commit = MoreAsyncUtil.composeWhenCompleteAndHandle(commit, (v, ex) -> {
+                if (ex instanceof CompletionException) {
+                    final Throwable cause = ex.getCause();
+                    if (cause instanceof FDBException) {
```
alecgrieser, line 485: Is it guaranteed that the `commit` error is always a `CompletionExcpetion` wrapping an `FDBException`? Or should this use something like `FDBExceptions.getFDBCause()` to extract the causal `FDBException` (if any) from more complicated error hierarchies?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -462,6 +477,21 @@ public CompletableFuture<Void> commitAsync() {
...
+                        final FDBException fdbException = (FDBException)cause;
+                        if (FDBError.fromCode(fdbException.getCode()) == FDBError.NOT_COMMITTED) {
+                            return readConflictingKeys();
```
alecgrieser, line 488: It would be nice to have a test of this. Something like a test that constructs two transactions with the appropriate options and then orchestrates a conflict and validates that the conflicting ranges read align

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
MMcM: Yes, they should. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
MMcM: Added the validation. I have no objection to `addTag`, if that seems cleaner. I just imagined that they all happened at once.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
MMcM: No, you're right. It isn't necessary.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
MMcM: I switched it to a value-returning async method instead of that strange side-effecting one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
MMcM: Even if it's guaranteed now, using that is more future proof. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContextConfig.java
alecgrieser: I think either is fine, really.

alecgrieser (approved):

PR #2119, opened 2023-05-10
Resolves #2112: Add a length-limited query plan printer

### LINKED ISSUE #2112: Add a length-limited query plan printer
For more complicated `RecordQueryPlan`s, it may not make sense to print out the entire plan. To facilitate that, we should add a method to print out a length-limited query plan. While that could be done with something like:

```java
final String fullString = plan.toString();
if (fullString.length() > maxSize) {
    return fullString.substring(0, maxSize) + "...";
} else {
    return fullString;
}
```

That has the downside of realizing the full query plan string in memory. To save memory, it would be nice to halt execution early in that case.

alecgrieser:
This adds a new `RecordQueryPlanVisitor` called `PlanStringRepresentation`. It encapsulates the job of constructing the query plan string for a given query plan tree, and it has a parameter to allow for the maximum size of such a string to be capped. This allows for arbitrarily large and complex plans to be packed into a string of fixed size. This can then in turn be used to limit the amount data logged for those larger plans.

An effort was made to keep plan strings the same. Note that the `toString` implementations on the plan objects has been replaced with the new visitor to avoid having two implementations of plan-object-to-explain-string implementations. To test that strings were kept the same, the randomized tester that was added here was run with the old `toString` implementations. Then the implementations were switched over to use the new visitor without changing the test. The exceptions here are:

* Previously, explode plans had an unmatched `]` that appears to have been a typo. In particular, it would generate strings like `explode(@2])`. I've added in a new `[` to match it, so the strings now look like `explode([@2])`
* For `RecordQueryInValuesJoinPlan`, the explain string didn't include whether the values were sorted (ascending or descending). All of the other `InSource`s did, so the new string representation aligns that `InSource` with the other `RecordQueryInJoinPlan` string representations

This resolves #2112.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
@@ -34,6 +34,7 @@ import com.apple.foundationdb.record.RecordCursorVisitor;
 import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
 import com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer;
+import com.apple.foundationdb.record.query.plan.PlanStringRepresentation;
 import com.apple.foundationdb.record.query.plan.cascades.AliasMap;
 import com.apple.foundationdb.record.query.plan.cascades.ExpressionRef;
 import com.apple.foundationdb.record.query.plan.cascades.GroupExpressionRef;
@@ -60,7 +61,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.stream.Collectors;
 
 /**
  * A {@link RecordQueryChooserPlanBase} that selects one of its children to be executed.
@@ -158,7 +158,7 @@ public class RecordQuerySelectorPlan extends RecordQueryChooserPlanBase {
 
     @Override
     public String toString() {
-        return "SELECTOR OF " + getChildStream().map(RecordQueryPlan::toString).collect(Collectors.joining(" "));
+        return PlanStringRepresentation.toString(this);
     }
 
     @Override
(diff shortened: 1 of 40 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
```diff
@@ -158,7 +158,7 @@ public int planHash(@Nonnull final PlanHashKind hashKind) {
 
     @Override
     public String toString() {
-        return "SELECTOR OF " + getChildStream().map(RecordQueryPlan::toString).collect(Collectors.joining(" "));
```
ScottDugas, line 161: This changed to `SELECT OR`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/PlanStringRepresentationTest.java
```diff
@@ -0,0 +1,642 @@
...
+            boolean showComparisonKey = r.nextBoolean();
+            String delimiter = String.format(" ∪%s ", showComparisonKey ? comparisonKey : "");
+            return Pair.of(RecordQueryUnionPlan.from(plans, comparisonKey, showComparisonKey), String.join(delimiter, strings));
+        }
+    }
+
+    @Nonnull
+    private static Pair<RecordQueryPlan, String> randomPlanAndString(@Nonnull Random r, double decay) {
+        // Generate a random plan, but use the decay argument to avoid creating trees with too many
+        // levels (and potentially hitting the maximum stack depth). Every time we a plan with
```
ScottDugas, line 424:
```suggestion
        // levels (and potentially hitting the maximum stack depth). Every time a plan with
```

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQuerySelectorPlan.java
alecgrieser: Whoops. Fixed

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/PlanStringRepresentationTest.java
```diff
@@ -497,7 +497,17 @@ void doNotEvaluateOutsideOfLimit() {
...
         for (int i = 0; i < 3; i++) {
-            plans.add(randomPlanAndString(r).getLeft());
+            RecordQueryPlan plan;
+            while (true) {
+                plan = randomPlanAndString(r).getLeft();
+                try {
+                    plan.isReverse();
+                    break;
+                } catch (RecordCoreException e) {
+                    // Not all plans have a defined reverse-ness. If this happens, generate a new plan
```
ScottDugas, line 507: This seems like we might be missing something here, but I think it should be covered by the other tests

ScottDugas (approved):

PR #2126, opened 2023-05-11
Resolves #2108: Cascades should match record type key field with record type restrictions

### LINKED ISSUE #2108: Cascades should match record type key field with record type restrictions
The Cascades planner currently doesn't match record type keys at all during planning. This can be a problem for operations like querying for a single type when the primary key is prefixed by the record type key. In that case, we'd want a query equivalent to `SELECT * FROM aTable` to be executed by a scanning just the records of that type:

```
Scan([IS aTable])
```

(Which is how the old planner plans such a query.) Instead, the planner doesn't match the record type key, and so the plan is executed as:

```
Scan(<,>) | [aTable]
```

That is, as a full scan with a type-filter on `aTable` after it.

alecgrieser:
This adds record type key matching to the Cascades planner. This is designed to handle two scenarios:

1. Record type key fields in an index. This makes some amount of sense for multi-type (including universal) indexes, but this only tests single-type indexes mainly as a stepping stone.
2. In the primary key. As is documented in the FAQ on how to make record types act like tables, this is good to have if the user has put record type key prefixes on all primary keys, as this allows scans of the primary index for a given type to be executed efficiently.

There are actually two approaches in here, which are separated out into separate commits. The first adds additional matching to the `SelectExpression::isSubsumedBy` method so that if there are record type fields in a match candidate, the value can be fixed based on the value of record types that can be returned by the relevant quantifier. The second commit pushes the type filter predicate down into the `LogicalTypeFilterExpression`, and then consolidates record type matching to that method.

There's some benefit to consolidating the type matching stuff to a single type of node. However, it may also be that matching within the `SelectExpression` is a bit more flexible. In particular, it seems like it's more likely to be more resilient to more exotic query trees. But both are presented here in this PR for view.

This resolves #2108.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/QueryPredicateMatchers.java
```diff
@@ -118,4 +119,11 @@ public static <P extends QueryPredicate> BindingMatcher<OrPredicate> orPredicate
                 Extractor.of(OrPredicate::getChildren, name -> "children(" + name + ")"),
                 downstream);
     }
+
+    @Nonnull
+    public static <V extends Value> BindingMatcher<PredicateWithValue> predicateWithValue(@Nonnull final BindingMatcher<V> downstreamValue) {
```
normen662, line 124: can you use the matchers on `ValuePredicate`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/QuantifierMatchers.java
```diff
@@ -78,6 +79,13 @@ public static BindingMatcher<Quantifier> anyQuantifierOverRef(@Nonnull final Bin
         return ofTypeRangingOverRef(Quantifier.class, downstream);
     }
 
+    @Nonnull
+    public static BindingMatcher<Quantifier> anyQuantifierWithAlias(@Nonnull final BindingMatcher<CorrelationIdentifier> alias) {
```
normen662, line 83: Matchers on aliases will only work if the debugger is installed. So if you have a test scenario where you need this, it's fine, it won't work if `DebuggerWithSymbolTables` is not present as the alias is not stable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ValueMatchers.java
```diff
@@ -150,4 +152,16 @@ public static BindingMatcher<StreamableAggregateValue> streamableAggregateValue(
...
                 downstreamValues);
     }
+
+    @Nonnull
+    public static BindingMatcher<RecordTypeValue> recordTypeValue() {
+        return typed(RecordTypeValue.class);
+    }
+
+    @Nonnull
+    public static BindingMatcher<RecordTypeValue> recordTypeValue(@Nonnull BindingMatcher<CorrelationIdentifier> alias) {
```
normen662, line 162: Same here, you get into stability problems if this is called outside the test cases (which have a debugger installed). If that's ok, then please go ahead (maybe add a comment into the java doc).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/RecordTypesProperty.java
```diff
@@ -139,6 +144,30 @@ public Set<String> evaluateAtExpression(@Nonnull RelationalExpression expression
...
+        } else if (byComparisons.isEmpty()) {
+            // No type comparisons in the scan. Return the original set of record types
+            return recordTypes;
+        } else {
+            // One type comparison in the scan. Limit the set of types to just those consistent with the scan
+            return Sets.intersection(byComparisons, recordTypes);
+        }
+    }
+
+    private Set<String> comparisonLimitedTypes(@Nonnull ScanComparisons scanComparisons) {
```
normen662, line 163:
```suggestion
    @Nonnull
    private Set<String> comparisonLimitedTypes(@Nonnull ScanComparisons scanComparisons) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/PlannerRepl.java
```diff
@@ -249,8 +249,6 @@ public void onEvent(final Event event) {
         if (lineReader == null) {
             return;
         }
-        Objects.requireNonNull(queryAsString);
```
normen662, line 252: Not sure why this was removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalTypeFilterExpression.java
```diff
@@ -134,7 +161,47 @@ public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candid
                                           @Nonnull final AliasMap aliasMap,
                                           @Nonnull final IdentityBiMap<Quantifier, PartialMatch> partialMatchMap,
                                           @Nonnull final EvaluationContext evaluationContext) {
-        return exactlySubsumedBy(candidateExpression, aliasMap, partialMatchMap);
+        final Iterable<MatchInfo> subsumedMatches = exactlySubsumedBy(candidateExpression, aliasMap, partialMatchMap);
```
normen662, line 164: I would probably do this:

1. check if both are `LogicalTypeFilterExpression`s. If not return empty.
2. check if the candidate expression has a place holder, if not straight out call exactlySubsumedBy() and return that
3. all the special stuff when there is a placeholder

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalTypeFilterExpression.java
```diff
@@ -134,7 +161,47 @@ public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candid
...
+
+        // Bind any predicate that can be bound to the record type key
+        final String recordTypeName = Iterables.getOnlyElement(recordTypes);
+        Comparisons.Comparison comparison = new RecordTypeKeyComparison.RecordTypeComparison(recordTypeName);
+        Optional<ComparisonRange> comparisonRangeMaybe = ComparisonRange.from(comparison);
+        if (comparisonRangeMaybe.isEmpty()) {
+            return subsumedMatches;
+        }
+        ComparisonRange comparisonRange = comparisonRangeMaybe.get();
+        QueryPredicate predicate = new ValuePredicate(new RecordTypeValue(inner.getAlias()), comparison);
```
normen662, line 181: Shouldn't this predicate already be in your set of query predicates that are part of the query side? You can create it when the type filter expression with the appropriate `recordTypes` is created.

normen662 (changes requested):
1. Do not create variations on the candidate side. Defer the call to the expansionBuilders build() method if possible. Recreate the appropriate `LogicalTypeFilterExpression` by calling a new `LogicalTypeFilterExpression.withPlaceholder(...)` and then call the expansion's `build()` after that.
2. Create a `LogicalTypeFilterExpression.withPlaceholder(...)` that encapsulates the creation of the predicates within the type filter.
3. Write a test case that make sure that an index `a, b, rTK, c, d` can be used on an `ORDER BY a, b, c, d`

alecgrieser:
Superceded by #4032

PR #2128, opened 2023-05-17
Resolves #2131: Predicate normalizations and generalized OrToLogicalUnionRule

### LINKED ISSUE #2131: cascades planner needs to enumerate OR factorizations properly
Currently we take a `SelectExpression` and normalize its `QueryPredicate`s to its CNF and DNF. Usually, the DNF then drives transformations of the DNF into a `UNION` which in turn helps to trigger index matching on the constituent terms.

All that is necessary but not general enough. Especially with the introduction of index filters index matching became more demanding.
It is necessary to create proper intermediate representations of predicates that are neither CNF nor DNF but partially multiplied forms.

The idea is as follows:
1. We still generate the CNF in `NormalizePredicatesRule` if the CNF is not too big.
2. We then generalize `OrToLogicalUnionRule`. The rule should fix some factors in the CNF created in 1. and create the DNF of the remaining factors. If that DNF is a logical OR we then create a UNION in the old way but also push all fixed predicates to each arm.

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesCostModel.java
```diff
@@ -80,25 +84,58 @@ public int compare(@Nonnull RelationalExpression a, @Nonnull RelationalExpressio
             return 1;
         }
 
-        int unsatisfiedFilterCompare = Integer.compare(PredicateCountProperty.evaluate(a),
-                PredicateCountProperty.evaluate(b));
+        final Map<Class<? extends RelationalExpression>, Set<RelationalExpression>> planOpsMapA =
```
normen662, line 87: To the reviewer: Changes to this file may be important!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -590,6 +590,27 @@ public boolean isUnary() {
         }
     }
 
+    @Nullable
+    public static Type invertComparisonType(@Nonnull final Comparisons.Type type) {
+        if (type.isUnary()) {
+            return null;
```
MMcM, line 596: `IS_NULL` and `NOT_NULL` are their respective inverses. There doesn't seem to be anything about how this is used that restricts to binary comparisons.

MMcM: Does not need to block this, though.

MMcM (approved):

PR #2129, opened 2023-05-23
introduce RecordQueryMemoryDamPlan

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -596,6 +596,7 @@ public enum Counts implements Count {
         PLAN_COMPOSED_BITMAP_INDEX("number of composed bitmap plans", false),
         /** The number of query plans that include a {@link com.apple.foundationdb.record.query.plan.sorting.RecordQuerySortPlan}. */
         PLAN_SORT("number of sort plans", false),
+        PLAN_DAM("number of dam plans", false),
```
alecgrieser, line 599: Related to Jean-Claude?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Nonnull
+    @Override
+    @SuppressWarnings({"PMD.CloseResource", "resource"})
+    public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull FDBRecordStoreBase<M> store,
+                                                                     @Nonnull EvaluationContext context,
+                                                                     @Nullable byte[] continuation,
+                                                                     @Nonnull ExecuteProperties executeProperties) {
+        // Since we are sorting, we need to feed through everything from the inner plan,
+        // even just to get the top few.
```
alecgrieser, line 89: Is this comment about sorting accurate? I thought, based on what it appears that this PR is doing, that the dam was more for Halloween problems during update operations rather than sorting. If that's the case, it also seems like it would be safe to pass the skip and limit along

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemoryDam.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.function.Supplier;
+
+/**
+ * Collect keyed values into a {@link TreeMap} so that they end up sorted.
```
alecgrieser, line 40: I assume this is copied and pasted incorrectly, given that this uses a `LinkedHashMap` internally, not a `TreeMap`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
```diff
@@ -0,0 +1,216 @@
...
+ * A query plan implementing sorting in-memory, possibly spilling to disk.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class RecordQueryDamPlan implements RecordQueryPlanWithChild {
+    private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Record-Query-Dam-Plan");
+
+    @Nonnull
+    private final Quantifier.Physical inner;
+    @Nonnull
+    private final RecordQuerySortKey key;
```
alecgrieser, line 70: What's the semantic meaning of a "sort key" here? Given that data are kept in insertion order, it seems like the semantics here are a little fuzzy

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQuerySortAdapter.java
```diff
@@ -71,10 +72,11 @@
     private Key encryptionKey;
     private static final Supplier<SecureRandom> RANDOM = Suppliers.memoize(SecureRandom::new);
 
-    protected RecordQuerySortAdapter(int memoryLimit, boolean memoryOnly,
+    protected RecordQuerySortAdapter(int memoryLimit, boolean memoryOnly, boolean isInsertionOrder,
                                      @Nonnull RecordQuerySortKey key, @Nonnull FDBRecordStoreBase<M> recordStore) {
         this.memoryLimit = memoryLimit;
         this.memoryOnly = memoryOnly;
+        this.isInsertionOrder = isInsertionOrder;
```
alecgrieser, line 79: Having this be a Boolean flag is a little awkward. It almost seems like this suggests that there needs to be more separation between the sorted cursor and the dam cursor, at the mechanics level, though I'm not sure I really believe that

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemorySortCursor.java
```diff
@@ -73,15 +75,21 @@ public CompletableFuture<RecordCursorResult<V>> onNext() {
...
+        return scratchpad.load(inputCursor, minimumKey).thenApply(loadResult -> {
             inputContinuation = loadResult.getSourceContinuation();
             if (loadResult.getSourceNoNextReason().isOutOfBand()) {
                 // The input cursor did not complete; we must save the sorter state in the continuation so can pick up after.
-                MemorySortCursorContinuation<K, V> continuation = new MemorySortCursorContinuation<>(adapter, false, sorter.getMap().values(), minimumKey, inputContinuation);
+                final MemorySortCursorContinuation<K, V> continuation =
+                        new MemorySortCursorContinuation<>(adapter, false, scratchpad.getMap().values(),
+                                adapter.isInsertionOrder() && loadResult.hasSeenMinimumKey()
+                                ? null
+                                : minimumKey,
```
alecgrieser, line 86: I'm not sure the whole "minimum key" thing makes all that much sense for the dam case, though perhaps continuations for dams like this are more nuanced, given that they exist specifically for cases where data has to be loaded into memory, but continuations are designed for incremental processing. In any case, this is another area where having a bit more separation between the in-memory sort and dam code might be nice

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQuerySortAdapter.java
normen662: Yeah -- so if `isInsertionOrder` then we only use the comparator to find the current minimum key and treat everything that came before as smaller and everything that came afterwards as bigger. It makes the whole process stateful.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemorySortCursor.java
normen662: The minimum key is more like the last consumed item in the dam case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
normen662: The sort key is only needed in its function to identify a singular item (for dams). We do use the comparator to do that. It is however not used for `<` and `>` as it is in the sort case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemorySortCursor.java
normen662: This is now implemented in a different way where all the stuff that is really different between the two cases is extracted into their own `MemorySortComparator`s which are created by the adapter and which can be stateful.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQuerySortAdapter.java
normen662: I have removed this clunkiness in exchange for hopefully something better and less clunky.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
alecgrieser: It might be good to note that in a comment in the class so the semantics are more permanently documented

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
alecgrieser: Asking about this comment again. And related to that, whether clearing the skip and limit for dam plans makes sense, or if it should pass down the skip and limit to its child

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemoryDam.java
alecgrieser: I believe this is still out of date

alecgrieser (changes requested):
I like the way the new comparator class separates out concerns between memory buffers for dams and sorts 👍

Left a few comments here, mainly about documentation/comments that I think could be cleaned up a bit

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/sorting/MemoryDam.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryDamPlan.java
normen662: So for a dam, we just need to ensure that the entire inner is consumed before any records are returned. It's not so much about the ordering but the before-after this introduces into the data flow. If the `MemoryDam` is in-memory-limitded, we may actually go over the input multiple times which is weird for a dam. However, in reality we use that plan operator in a way that we are unlimited which in turn ensures that we go over the input exactly once. 

I added that bit of information in the comment.

alecgrieser (approved):

MMcM (approved):

PR #2133, opened 2023-05-24
Fixes #2132: Plan string changed for RecordQueryMapPlan

### LINKED ISSUE #2132: Plan string changed for RecordQueryMapPlan
With the new plan representation visitor added in #2119, the representation for the `RecordQueryMapPlan` changed. Previously, it was `map(${innerPlan}[${resultValue}])`, but the new representation drops the `map()` wrapper. This means plan strings for those plans are changed (unintentionally). We should revert to the old behavior.

alecgrieser:
This reverts the plan string for `RecordQueryMapPlan`s, which were accidentally changed as a part of #2119, to their older format to keep plan strings stable.

This fixes #2132.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -15,7 +15,7 @@ The Guava dependency version has been updated to 31.1. Projects may need to chec
 // begin next release
 ### NEXT_RELEASE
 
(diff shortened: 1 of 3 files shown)


hatyo (approved):
LGTM

PR #2134, opened 2023-05-24
Allow placeholders in aggregate index match candidate's underlying SelectExpression

### LINKED ISSUE #2135: Allow placeholders in aggregate index match candidate's underlying `SelectExpression`
Currently we expose placeholders only in the upper `SelectExpression` of the aggregate index match candidate. This matches well with group-by queries that have a predicate on the grouping column(s) in its upper `SelectExpression`. However this fails to match when the group-by query has a predicate in its underlying select because we do not expose any placeholders in the corresponding `SelectExpression` of the aggregate index plan.

hatyo:
Currently we expose placeholders only in the upper `SelectExpression` of the aggregate index match candidate. This matches well with group-by queries that have a predicate on the grouping column(s) in its upper `SelectExpression`. However this fails to match when the group-by query has a predicate in its underlying select because we do not expose any placeholders in the corresponding `SelectExpression` of the aggregate index plan.

This PR addresses this issues by exposing the _same_ set of placeholders in both `SelectExpressions` of the aggregate index plan. I had to do some adaptations to make this work:
* In `MatchInfo`, when trying to merge the parameter binding map, it is now cognizant to the fact that a binding could be associated with multiple `ComparisonRange`s and it attempts to merge them. There was a small bug in `ComparisonRange` merge logic that is fixed as a byproduct of this.
* In `computeMatchedOrderingParts` we used to create a parameter binding predicate _map_ and stash the `QueryPredicate` of each parameter binding in the resulting `MatchedOrderedPart`. This code broke because now it is possible to associate a parameter binding with multiple predicates instead of one. When attempting to refactor this logic, it turned out that stashing the `QueryPredicate` inside hat `MatchedOrderedPart` is not really necessary and is a remnant of previous work on ordering logic, so it is removed, and it caused a bunch of dead code to be cleaned up as a side-effect.
* Refactored the `GroupByTest` to reflect the fact that we can now plan aggregate indexes for more complex groups by plans correctly.

This fixes #2135.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
@@ -76,12 +76,13 @@ public class MatchedOrderingPart {
             return false;
         }
         final MatchedOrderingPart that = (MatchedOrderingPart)o;
-        return getOrderingPart().equals(that.getOrderingPart());
+        return getOrderingPart().equals(that.getOrderingPart()) &&
+               comparisonRangeType.equals(that.comparisonRangeType);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(getOrderingPart().hashCode());
+        return Objects.hash(getOrderingPart(), comparisonRangeType);
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
@@ -20,12 +20,9 @@
 
 package com.apple.foundationdb.record.query.plan.cascades;
 
-import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
-import com.google.common.base.Preconditions;
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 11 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
```diff
@@ -92,20 +76,18 @@ public boolean equals(final Object o) {
...
         final MatchedOrderingPart that = (MatchedOrderingPart)o;
-        return getOrderingPart().equals(that.getOrderingPart()) &&
-               Objects.equals(getQueryPredicate(), that.getQueryPredicate());
+        return getOrderingPart().equals(that.getOrderingPart());
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(getOrderingPart().hashCode(), getQueryPredicate());
+        return Objects.hash(getOrderingPart().hashCode());
```
normen662, line 84: Can you add the `isReverse` bit of information here and in the `equals()`. I hope it was just an oversight and there is no hidden reason to not have it in here.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchedOrderingPart.java
hatyo: `isReverse` is member of `OrderingPart` and it is included in calculating equality and hash code. So I think it is not necessary to have it included here again, right?

hatyo: I added `comparisonRangeType` to both equality and hash code calculation.

normen662: Sorry, I was being blind, I meant `comparisonRangeType` but you already picked up on my mistake.

hatyo: no worries at all

PR #2139, opened 2023-05-30
flatmap should honor skips and limits

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFlatMapPlan.java
```diff
@@ -125,7 +127,7 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                             });
                 },
                 continuation,
-                5);
+                5).skipThenLimit(executeProperties.getSkip(), executeProperties.getReturnedRowLimit());
```
alecgrieser, line 130: At some point, we should also change this `5` constant to use the store's `PipelineSizer`, but that doesn't have to be now

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFlatMapPlan.java
normen662: Agreed.

Issue #2145, opened 2023-06-02
LuceneRecordCursor should only load stored values needed for query


jleach4:
The current implementation returns all stored values even if all you need is the primary key.

PR #2146, opened 2023-06-02
LuceneRecordCursor should only load stored values needed for query #2145

jleach4:
Test coverage in FDBLuceneQueryTest.covering().  This is just a performance optimization to not load all stored values unless query requires them.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
@@ -43,6 +43,7 @@ import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
 import com.apple.foundationdb.tuple.Tuple;
 import com.apple.foundationdb.tuple.TupleHelpers;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.IndexNotFoundException;
 import org.apache.lucene.index.IndexReader;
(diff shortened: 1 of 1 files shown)


MMcM (approved):
We could perhaps trim off a tiny bit more by making a `DocumentStoredFieldVisitor` with the target fields directly, but the one made internally probably goes on the stack anyway.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -162,6 +165,10 @@ public class LuceneRecordCursor implements BaseCursor<IndexEntry> {
         this.pageSize = pageSize;
         this.executorService = executorService;
         this.storedFields = storedFields;
+        this.storedFieldsToReturn = Sets.newHashSet(LuceneIndexMaintainer.PRIMARY_KEY_FIELD_NAME);
+        if (this.storedFields != null) { // Only return StoredValues required for primary key and to support query.
```
ohadzeliger, line 169: Trying to understand the role of the `storedFields` argument: Running queries with various `SELECT` fields (or `.*`) resulted in this parameter unchanged: it always contains 13 elements, 11 of which are `null` and the remaining 2 are `msgstate_uid` and `msgstate_flags`.

ohadzeliger: So I guess my question is what should be the intended impact of this optimization?

jleach4: The optimization will not materialize the stored values that are not-needed based on the query.  We need primary key for each value but if you had a field for example that stored another value, it is not needed to be retrieved for the user unless the query user needs to return it.

ohadzeliger (approved):

PR #2148, opened 2023-06-05
resolves #2147: only break ORs into UNIONs if beneficial for index matching

### LINKED ISSUE #2147: `PredicateToLogicalUnionRule` should only yield if breaking the `UNION` is beneficial for index matching
```
SELECT *
FROM (a = 5 OR b = 10) AND c > 100
```

is currently broken into a UNION

```
   SELECT *
   FROM a = 5 AND c > 100
UNION DISTINCT
   SELECT *
   FROM b = 10 AND c > 100 
```

This causes a lot of enumeration and only leads to a beneficial outcome when at least one of the two UNION legs can be planned in a way that `a` or `b` participate in an index scan as a sargable. If that's not case then we do twice the work by scanning either the entire table twice or at least records satisfying `c > 100` in some index scan that is agnostic to `a` respectively `b` being broken.

This issue should only allow breaking the `OR` into a UNION if at least one of the `OR` terms can help in an index scan.

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -163,16 +148,101 @@ public Optional<CorrelationIdentifier> getParameterAliasOptional() {
...
+                                                     @Nonnull final CompensatePredicateFunction compensatePredicateFunction) {
+            return orTermMapping(queryPredicate, candidatePredicate, compensatePredicateFunction, Optional.empty(), Optional.empty());
+        }
+
+        @Nonnull
+        public static PredicateMapping orTermMapping(@Nonnull final QueryPredicate queryPredicate,
+                                                     @Nonnull final QueryPredicate candidatePredicate,
+                                                     @Nonnull final CompensatePredicateFunction compensatePredicateFunction,
+                                                     @Nonnull final Optional<CorrelationIdentifier> parameterAliasOptional,
+                                                     @Nonnull final Optional<QueryPlanConstraint> constraintOptional) {
```
hatyo, line 192: Is the constraint relevant for this type of mapping?

hatyo (approved):
Nice optimization.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: Not really -- the worst that can happen is that we think an or term matches but then later on it doesn't. So it's not the end of the world.

Issue #2152, opened 2023-06-07
Reference Mail Implementation for the Lucene Package


jleach4:
It would be nice to have an example schema with load characteristics...

PR #2153, opened 2023-06-07
Reference Mail Implementation for the Lucene Package #2152

jleach4:
Reference implementation for mail...


fdb-record-layer-core/src/test/proto/test_records_text.proto
```diff
@@ -87,3 +88,21 @@ message NestedMapDocument {
...
 }
+
+message Email {
+    optional int64 doc_id = 1 [(field).primary_key = true];
+    message MailHeader {
+        required string key = 1;
+        required string value = 2;
+    }
+    repeated MailHeader header = 2;
+    required string mailbox = 3;
```
alecgrieser, line 99: In general, we prefer `optional` to `required` for fields. Given that this is test-only, it doesn't _really_ matter, but it would be good to be consistent if possible.

fdb-record-layer-lucene/fdb-record-layer-lucene.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"
     testRuntimeOnly "org.apache.logging.log4j:log4j-core:${log4jVersion}"
 
+    testImplementation("com.sun.mail:javax.mail:${javaxMail}")
```
alecgrieser, line 41: Is the `javaxMail` property defined? I'd also suggest `javaxMailVersion`, again for consistency with the other dependency version properties (and to put the definition in `gradle.properties`)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneMailExampleTest.java
```diff
@@ -0,0 +1,244 @@
...
+                    mailBoxIndex++;
+                }
+                if (index >= recordCountToLoad) {
+                    break;
+                }
+            } finally {
+                totalWriteTime += (System.currentTimeMillis() - start);
+                if (index % 100 == 0) {
+                    start = System.currentTimeMillis();
+                    //zone.flushWithoutReset();
```
alecgrieser, line 194: This should probably be removed

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneMailExampleTest.java
```diff
@@ -0,0 +1,244 @@
...
+                    totalWriteTime+= (System.currentTimeMillis()-start);
+                    LOGGER.info("Number of records that have been loaded: {}. Time taken: {} ", index, Duration.ofMillis(totalWriteTime));
+                }
+            }
+        }
+        LOGGER.info("Number of records that have been loaded: {}. Time taken: {} ", index, Duration.ofMillis(totalWriteTime));
```
alecgrieser, line 200: The canonical way to do this (a log message with a static message and some key-value pairs) would be with a `KeyValueLogMessage`. I'd also also suggest including the unit of measure in the logs--something like `load_time_micros`

Issue #2169, opened 2023-06-13
LuceneOptimizedCompoundReader always reads block 0, even when not needed...


jleach4:
We always read block 0, even when not needed.  This adds about 33% more reads to the simple insert record case.

alecgrieser:
Resolved in #2170

PR #2170, opened 2023-06-13
LuceneOptimizedCompoundReader always reads block 0, even when not needed... #2169

jleach4:
This removes a read to block 0 for compound files when we do a single row insert.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
@@ -71,7 +71,10 @@ final class LuceneOptimizedCompoundReader extends CompoundDirectory {
         this.segmentName = si.name;
         String dataFileName = IndexFileNames.segmentFileName(segmentName, "", LuceneOptimizedCompoundFormat.DATA_EXTENSION);
         String entriesFileName = IndexFileNames.segmentFileName(segmentName, "", LuceneOptimizedCompoundFormat.ENTRIES_EXTENSION);
-        handle = directory.openInput(dataFileName, context); // async
+        if ( !(directory instanceof LuceneOptimizedWrappedDirectory) ) {
+            throw new IOException("Directory Must Be Wrapped");
+        }
+        handle = ((LuceneOptimizedWrappedDirectory) directory).openLazyInput(dataFileName, context, 0, 0L); // attempting not to read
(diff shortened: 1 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexInput.java
```diff
@@ -73,20 +73,49 @@ public FDBIndexInput(@Nonnull final String resourceDescription, @Nonnull final F
...
+     * This is currently used by the LuceneOptimizedCompoundReader to not have to read its first block unless needed.
+     *
+     * @param resourceDescription opaque description of file; used for logging
+     * @param fdbDirectory FDB directory mapping
+     * @param initialOffset initialOffset
+     * @param position currentBlockPosition
+     * @throws IOException exception
+     */
+    public FDBIndexInput(@Nonnull final String resourceDescription, @Nonnull final FDBDirectory fdbDirectory, long initialOffset, long position) throws IOException {
+        super(resourceDescription);
```
ohadzeliger, line 91: I think it would be easier to follow if this constructor would call 
`this(resourceDescription, fdbDirectory, dbDirectory.getFDBLuceneFileReferenceAsync(resourceDescription), initialOffset, position, -1, CompletableFuture.completedFuture(new byte[0]))`
And, maybe add a boolean parameter to control whether to do a seek.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedCompoundReader.java
```diff
@@ -71,7 +71,10 @@ public LuceneOptimizedCompoundReader(Directory directory, SegmentInfo si, IOCont
         this.segmentName = si.name;
         String dataFileName = IndexFileNames.segmentFileName(segmentName, "", LuceneOptimizedCompoundFormat.DATA_EXTENSION);
         String entriesFileName = IndexFileNames.segmentFileName(segmentName, "", LuceneOptimizedCompoundFormat.ENTRIES_EXTENSION);
-        handle = directory.openInput(dataFileName, context); // async
+        if ( !(directory instanceof LuceneOptimizedWrappedDirectory) ) {
```
ohadzeliger, line 74: Can we change the constructor to take a parameter of the right type and ensure we detect this failure at compile time?

ohadzeliger (changes requested):

ohadzeliger (approved):

PR #2205, opened 2023-06-30
Support Greatest, Least and Coalesce scalar functions

arnaud-lacurie:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+            SemanticException.check(argType.isPrimitive(), SemanticException.ErrorCode.ARGUMENT_TO_ARITHMETIC_OPERATOR_IS_OF_COMPLEX_TYPE);
+            if (resultType.isUnresolved()) {
+                resultType = argType;
+            } else if (!argType.isUnresolved()) {
+                resultType = Type.maximumType(resultType, argType);
+            }
+            SemanticException.check(resultType != null, SemanticException.ErrorCode.INCOMPATIBLE_TYPE);
+        }
+
+        final Optional<ScalarFunction> scalarFunction = Enums.getIfPresent(ScalarFunction.class, builtInFunction.getFunctionName().toUpperCase(Locale.getDefault())).toJavaUtil();
```
normen662, line 172: Wouldn't it be better if the `BuiltInFunction` also declared which enum to use so you don't have to look it up through enum manipulations?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * A {@link Value} that applies an arithmetic operation on its child expressions.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class ScalarFunctionValue extends AbstractValue {
```
normen662, line 63: I think this should be named something that indicated the variadic nature of these functions, maybe `ScalarVariadicFunctionValue`. `foo(string, int, double)` also is a scalar function but follows different encapsulation and logic in general.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+            return type;
+        }
+
+        @Nullable
+        public Object eval(List<Object> args) {
+            return evaluateFunction.apply(args);
+        }
+
+        @SuppressWarnings("unchecked")
+        private static <T extends Comparable<T>> T greatest(final List<Object> args, Class<T> clazz) {
```
normen662, line 295: what do you need the `Class<T>` for? The `(T)` cast is `unchecked` anyway.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+            T max = (T) args.get(0);
+            for (Object i : args) {
+                if (i == null) {
+                    return null;
+                }
+                if (((T) i).compareTo(max) > 0) {
```
normen662, line 301: I think it does not matter much but the original idea behind physical operators was to not do dispatch on the type but to statically encode what they need to do. All of these physical implementations dynamically dispatch and are effectively the same for all types.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+        LEAST,
+        COALESCE;
+    }
+
+    /**
+     * Physical operators.
+     */
+    @VisibleForTesting
+    public enum PhysicalOperator {
+        GREATEST_INT(ScalarFunction.GREATEST, TypeCode.INT, args -> greatest(args, Integer.class)),
```
normen662, line 241: Maybe you should read this comment after the ones further down but the idea was to hard code the physical action here, e.g. this one would be doing hard-coded integer comparisons, not delegate to `compare()`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+        Type resultType = arguments.get(0).getResultType();
+        for (final var arg : arguments) {
+            Type argType = arg.getResultType();
+            SemanticException.check(argType.isPrimitive(), SemanticException.ErrorCode.ARGUMENT_TO_ARITHMETIC_OPERATOR_IS_OF_COMPLEX_TYPE);
+            if (resultType.isUnresolved()) {
+                resultType = argType;
+            } else if (!argType.isUnresolved()) {
+                resultType = Type.maximumType(resultType, argType);
+            }
+            SemanticException.check(resultType != null, SemanticException.ErrorCode.INCOMPATIBLE_TYPE);
```
normen662, line 169: This can only trigger if the `else if` one line up was taken. Maybe move this check into that branch.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
```diff
@@ -0,0 +1,332 @@
...
+    private static Map<Pair<ScalarFunction, TypeCode>, PhysicalOperator> getOperatorMap() {
+        return operatorMapSupplier.get();
+    }
+
+    @Nonnull
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
+    private static Value encapsulate(@Nonnull BuiltInFunction<Value> builtInFunction,
+                                     @Nonnull final List<? extends Typed> arguments) {
+        Verify.verify(arguments.size() >= 2);
+        Type resultType = arguments.get(0).getResultType();
```
normen662, line 160: Why not make this line here `resultType = null` and then check on that in the loop rather than also executing the loop on `arguments.get(0)`

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
arnaud-lacurie: Good point :+1:

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
arnaud-lacurie: Is there a better way to do this without duplicating the code for all types?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ScalarFunctionValue.java
normen662: Can we rename this ppplllleeeeaassee. :-)

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/VariadicFunctionValue.java
```diff
@@ -0,0 +1,393 @@
+/*
+ * ScalarFunctionValue.java
```
normen662, line 2:
```suggestion
 * VariadicFunctionValue.java
```

normen662 (changes requested):

PR #2208, opened 2023-07-03
Lucene skip data integrity static

jjezra:
This will: 
  1. Revert commit 3918fd5b184b222571c34317351a85132340476d (which doesn't work as expected - apparently because often a default constructor is called via the `@AutoService(PostingsFormat.class)`. 
  2. Add static var to control `checkIntergrity` +  allow callers to set it for all future transactions

   Note to reviewer: the use of a static var is ugly. Please consider that a quick n dirty way to test skipping the checksums test. If it does works we'll find a better way to do it.


ohadzeliger (approved):
This is OK as a temporary measure, please create an issue to track the replacement by a more permanent solution

PR #2211, opened 2023-07-05
Resolve #2210: Add more metrics to Lucene.

### LINKED ISSUE #2210: Add some metrics to Lucene operations
The following metrics are missing from the Lucene store timer:
- Merge
- Create file
- Rename file
- Delete file
-

ohadzeliger:
Created new events for Merge and CreateOutput.
Some metrics (delete/rename) have a WAIT event associated with them. The event doesn't always trigger (likely due to the use of context.asynctosync) so a counter event was added for them.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -580,6 +580,7 @@ public class FDBDirectory extends Directory {
      */
     @Override
     @Nonnull
+    @SuppressWarnings("java:S2093")
     public IndexOutput createOutput(@Nonnull final String name, @Nullable final IOContext ioContext) {
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(getLogMessage("createOutput",
@@ -682,7 +683,7 @@ public class FDBDirectory extends Directory {
         return new FDBIndexInput(name, this);
     }
 
-    public IndexInput openLazyInput(@Nonnull final String name, @Nonnull final IOContext ioContext, long initialOffset, long position) throws IOException {
+    public IndexInput openLazyInput(@Nonnull final String name, long initialOffset, long position) throws IOException {
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(getLogMessage("openInput",
                     LuceneLogMessageKeys.FILE_NAME, name));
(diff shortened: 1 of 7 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -576,12 +580,18 @@ public long fileLength(@Nonnull String name) throws NoSuchFileException {
...
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(getLogMessage("createOutput",
                     LuceneLogMessageKeys.FILE_NAME, name));
         }
-        return new FDBIndexOutput(name, name, this);
+        long startTime = System.nanoTime();
+        try {
+            return new FDBIndexOutput(name, name, this);
+        } finally {
+            context.record(LuceneEvents.Waits.WAIT_LUCENE_CREATE_OUTPUT, System.nanoTime() - startTime);
```
ScottDugas, line 593: Why are you adding tracking for this? I don't immediately see anything that should be expensive

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -669,7 +683,7 @@ public IndexInput openInput(@Nonnull final String name, @Nonnull final IOContext
         return new FDBIndexInput(name, this);
     }
 
-    public IndexInput openLazyInput(@Nonnull final String name, @Nonnull final IOContext ioContext, long initialOffset, long position) throws IOException {
```
ScottDugas, line 672: I'm assuming this was just removed because you noticed nothing was using it?

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: There were FindBugs or PMD errors that mandated that I remove it all through the call chain (as the parameter wasn't used).

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: There are some potentially meaningful things to capture in `fdbDirectory.getIncrement()`which is called from the constructor. Plus, incrementing the counter (even without the WAIT) would give us meaningful data of the flow of file creation.

ScottDugas (approved):

Issue #2182, opened 2023-06-21
FieldInfosFormat requires each segment to fetch their schema definition


jleach4:
It should be possible to have a small map of the possible different schemas in case of missing fields etc, without taking the scalability penalty of fetching for each segment.

alecgrieser:
Resolved originally in #2183. Reverted by #2200, but then reintroduced by #2214 (see #2213).

PR #2214, opened 2023-07-07
Resolve #2213: Reintroduce #2182

### LINKED ISSUE #2213: Reintroduce #2182
PR #2182 was reverted due to a bug. It needs to be reintroduced.

ohadzeliger:
This PR is a reintroduction of #2182 with the addition of a fix to the issue (failed `verify()`)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -420,6 +513,22 @@ void simpleInsertAndSearch() {
         }
     }
 
+    @Test
+    void largeMetadataTest() {
```
ScottDugas, line 517: Given the behavior of this change, it seems like there should be tests where we have different documents with different sets of fields. @jleach4 Did you find that those tests already existed. 
Perhaps changing this test to use `MAP_DOC` and have a variety of different keys in the map, for each document would cover that.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ohadzeliger: Added tests that use various fields out of the ManyFieldsDoc.

ScottDugas (approved):

Issue #2220, opened 2023-07-11
LuceneOptimizedCodec should be static, wrong scope.


jleach4:
LuceneOptimizedCodec is currently created per indexWriter.

alecgrieser:
Resolved in #2221.

PR #2221, opened 2023-07-11
LuceneOptimizedCodec should be static, wrong scope. #2220

jleach4:
LuceneOptimizedCodec is currently created per indexWriter.


ohadzeliger (dismissed):
I think this is OK as the java doc for Codec mentions that it is loaded via SPI service loader. One note though (maybe to be fixed in the future) is that it should be resolved via `Codec.forName()` as other places may have a need to get to an instance of our codec)

ohadzeliger:
> I believe this handles it, right?

> super("RL");

Right. probably with the `@AutoService(Codec.class)` and the `getName()` it is sufficient.

ohadzeliger (approved):

PR #2225, opened 2023-07-12
Resolve #2224: Lucene: allow skipping checkIntegrity for all Postings…

### LINKED ISSUE #2224: Lucene: allow skipping `checkIntegrity` for all PostingsFormat inheritors

jjezra:
…Format inheritors

Note to reviewer: The implementation is a little ugly. 
The idea is that: 
1. This is compatible with the existing optimization in `LuceneOptimizedPostingsFormat`  (no changes are required from the users)
2. After testing, if this optimization fixes the performance problems, we may remove the condition and make our `checkIntegrity` implementations a constant no-op.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedPostingsFormat.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedPostingsFormat.java
@@ -44,7 +44,7 @@ import java.util.function.Supplier;
 @AutoService(PostingsFormat.class)
 public class LuceneOptimizedPostingsFormat extends PostingsFormat {
     PostingsFormat postingsFormat;
-    private static boolean allowCheckDataIntegrity = true;
+    static boolean allowCheckDataIntegrity = true;
 
     public LuceneOptimizedPostingsFormat() {
         this(new Lucene84PostingsFormat());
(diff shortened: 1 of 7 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedPostingsFormat.java
```diff
@@ -44,7 +44,7 @@
 @AutoService(PostingsFormat.class)
 public class LuceneOptimizedPostingsFormat extends PostingsFormat {
     PostingsFormat postingsFormat;
-    private static boolean allowCheckDataIntegrity = true;
+    static boolean allowCheckDataIntegrity = true;
```
ScottDugas, line 47: If this sticks around for any amount of time, it would probably be good to move it to some sort of `LuceneGlobalConfig` class or something like that.

fdb-record-layer-lucene/src/main/java/org/apache/lucene/codecs/lucene84/LuceneOptimizedPostingsReader.java
```diff
@@ -2016,6 +2016,7 @@ public long ramBytesUsed() {
 
     @Override
     public void checkIntegrity() throws IOException {
+        // Should this checksum verification be skipped for fdb?
```
ScottDugas, line 2019: Based on the other places, I think that would make sense to align this with rest of our implementations for `checkIntegrity`

ScottDugas (approved):

PR #2231, opened 2023-07-18
Avoid introducing duplicates when splitting disjunction to union

### LINKED ISSUE #2230: Splitting disjunction into union produces duplicates
The `PredicateToLogicalUnionRule` splits a disjunction into an unordered union that might introduce duplicates. E.g.:

```sql
create table t1(col1 int, col2 int);
insert into t1 values (100, 200);
select * from t1 where col1 = 100 or col2 = 200;
```

The disjunction `col1 = 100 or col2 = 200` is split into an unordered union:

```sql
       UnorderedUnion
       /             \
   Filter             Filter
   col1 = 100         col2 = 200
       |                   |
     Scan                Scan
      T1                  T1
```

The above plan outputs the row `(100, 200)` twice since it satisfies the filters of each union leg. Note that more duplicates could be potentially introduced if we have more (predicated) union legs (that are satisfied).

hatyo:
This fixes the problem of potentially introducing duplicates when splitting a disjunction into a union as explained in #2230. The fix mainly introduces a new `LogicalUniqueExpression` that sits on top of each union leg and can only be implemented when underlying expression is distinct with primary key. a `LogicalDistinctExpression` is put on of the union expression to make sure we eliminate any duplicates.

This fixes #2230.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
@@ -154,12 +154,6 @@ public class NodeInfo {
             NodeIcon.COMPUTATION_OPERATOR,
             "Unordered Distinct",
             "An unordered distinct operator processes its input records and returns the set of distinct records. It does not require its input records to be ordered but does impose a memory overhead as it needs to track records it has already encountered.");
-
-    public static final NodeInfo UNORDERED_UNIQUE_OPERATOR = new NodeInfo(
-            "Unique",
-            NodeIcon.COMPUTATION_OPERATOR,
-            "Unique",
-            "A unique operator that outputs a stream of unique records. It does not require input records to be ordered and does not reorder output records.");
     public static final NodeInfo UNORDERED_PRIMARY_KEY_DISTINCT_OPERATOR = new NodeInfo(
             "UnorderedPrimaryKeyDistinct",
             NodeIcon.COMPUTATION_OPERATOR,
@@ -248,7 +242,6 @@ public class NodeInfo {
                 TYPE_FILTER_OPERATOR,
                 UNION_OPERATOR,
                 UNORDERED_DISTINCT_OPERATOR,
-                UNORDERED_UNIQUE_OPERATOR,
                 UNORDERED_PRIMARY_KEY_DISTINCT_OPERATOR,
                 UNORDERED_UNION_OPERATOR);
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
@@ -154,6 +154,12 @@ public class NodeInfo {
             NodeIcon.COMPUTATION_OPERATOR,
             "Unordered Distinct",
             "An unordered distinct operator processes its input records and returns the set of distinct records. It does not require its input records to be ordered but does impose a memory overhead as it needs to track records it has already encountered.");
+
+    public static final NodeInfo UNORDERED_UNIQUE_OPERATOR = new NodeInfo(
+            "Unique",
+            NodeIcon.COMPUTATION_OPERATOR,
+            "Unique",
+            "A unique operator that outputs a stream of unique records. It does not require input records to be ordered and does not reorder output records.");
     public static final NodeInfo UNORDERED_PRIMARY_KEY_DISTINCT_OPERATOR = new NodeInfo(
             "UnorderedPrimaryKeyDistinct",
             NodeIcon.COMPUTATION_OPERATOR,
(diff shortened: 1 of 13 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
```diff
@@ -154,6 +154,12 @@ public class NodeInfo {
             NodeIcon.COMPUTATION_OPERATOR,
             "Unordered Distinct",
             "An unordered distinct operator processes its input records and returns the set of distinct records. It does not require its input records to be ordered but does impose a memory overhead as it needs to track records it has already encountered.");
+
+    public static final NodeInfo UNORDERED_UNIQUE_OPERATOR = new NodeInfo(
```
normen662, line 158: I don't think there is an operator like that planned. Can't we just remove this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalUniqueExpression.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return semanticHashCode();
+    }
+
+    @Nonnull
+    @Override
+    public PlannerGraph rewriteInternalPlannerGraph(@Nonnull final List<? extends PlannerGraph> childGraphs) {
```
normen662, line 115: This may also be wrong in for `LogicalDistinctExpression` but I don't think we need to rewrite it. Can we just remove this override and then not implement the rest. interface?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalUniqueExpression.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
+    @Override
+    public boolean equals(final Object other) {
+        return semanticEquals(other);
+    }
+
+    @Override
+    public int hashCodeWithoutChildren() {
+        return 31;
```
normen662, line 105: Can you use a different prime? Half of our expressions use that number :-). Ideally we should use something like the `BASE_HASH` or maybe we can incorporate the hash of the class here since this hash code is never externalized and does not need to be stable, like plan hashes need to be.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementUniqueRule.java
```diff
@@ -0,0 +1,74 @@
...
+public class ImplementUniqueRule extends CascadesRule<LogicalUniqueExpression> {
+
+    @Nonnull
+    private static final CollectionMatcher<PlanPartition> anyPlanPartitionMatcher = all(anyPlanPartition());
+
+    @Nonnull
+    private static final BindingMatcher<ExpressionRef<? extends RelationalExpression>> innerReferenceMatcher = planPartitions(
+            where(planPartition -> planPartition.getAttributesMap().containsKey(DistinctRecordsProperty.DISTINCT_RECORDS)
+                                   && planPartition.getAttributeValue(PrimaryKeyProperty.PRIMARY_KEY).isPresent(),
+                    rollUpTo(anyPlanPartitionMatcher, ImmutableSet.of())));
```
normen662, line 60: There is a matcher `rollUp(...)` that uses `ImmutableSet.of()`. I think that would be more succinct.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementUniqueRule.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalUniqueExpression.java
hatyo: Good point! I am using a different prime number (`251`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalUniqueExpression.java
hatyo: Done, I thought that ideally, we should a graphical representation of all expressions?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/explain/NodeInfo.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalUniqueExpression.java
normen662: We do get a default which is fine here. This interface is just there to allow you to cognate the default.

normen662 (approved):

PR #2232, opened 2023-07-19
Constrain pipeline size to 1 for UPDATE operations

arnaud-lacurie:
The update pipeline is not thread safe and should not use a default size of 10

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -163,7 +163,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     private static final Logger LOGGER = LoggerFactory.getLogger(FDBRecordStore.class);
 
     public static final int DEFAULT_PIPELINE_SIZE = 10;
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -163,7 +163,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     private static final Logger LOGGER = LoggerFactory.getLogger(FDBRecordStore.class);
 
     public static final int DEFAULT_PIPELINE_SIZE = 10;
-    public static final PipelineSizer DEFAULT_PIPELINE_SIZER = pipelineOperation -> DEFAULT_PIPELINE_SIZE;
+    public static final PipelineSizer DEFAULT_PIPELINE_SIZER = pipelineOperation -> pipelineOperation == PipelineOperation.UPDATE ? 1 : DEFAULT_PIPELINE_SIZE;
```
normen662, line 166: This I think is too large of a hammer. The problem only occurs for inserts as we only implement `UPDATE` and `DELETE` over distinct sources. All three codepaths use `PipelineOperation.UPDATE`. Can you split that up into separate pipeline operations? Once that's done we can limit the `PipelineOperation.UPDATE` and set the others to `DEFAULT_PIPELINE_SIZE` in the sizer.

normen662 (changes requested):

PR #2235, opened 2023-07-20
Resolve #2234: Fix readSchema deadlock.

### LINKED ISSUE #2234: ReadSchema deadlock
The Lucene code for FieldInfos' read_schema is deadlocking. It looks like the caching mechanism is causing threads to block and for the readSchema call to time out.

ohadzeliger:
Make the byte[] schema perform a non-atomic cache population.
Remove the parsed fieldInfo caching (so it always reads from the byte[] cache).

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -45,7 +45,6 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.tuple.Pair;
-import org.apache.lucene.index.FieldInfos;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.store.ChecksumIndexInput;
 import org.apache.lucene.store.Directory;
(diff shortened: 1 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -488,7 +478,14 @@ private CompletableFuture<byte[]> readSchemaAsync(List<Long> bitSetWords) {
 
     public byte[] readSchema(List<Long> bitSetWords) throws IOException {
         BitSet bitSet = BitSet.valueOf(ArrayUtils.toPrimitive(bitSetWords.toArray(new Long[0])));
-        return fieldInfosDataCache.asMap().computeIfAbsent(bitSet, ignore -> context.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_GET_SCHEMA, readSchemaAsync(bitSetWords)));
+        // In order to avoid a deadlock, and since the readSchema is idempotent, perform a non-blocking, non-atomic cache
+        // population. There may be a few threads that make calls to FDB, but they should all be returning the same result
+        byte[] value = fieldInfosDataMap.get(bitSet);
+        if (value == null) {
```
ScottDugas, line 484: You have this backwards
```suggestion
        if (value != null) {
```

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: This is the keeper for the first check - the one below has != as it should

ScottDugas (approved):

PR #2236, opened 2023-07-20
Resolve #2234: Fix readSchema deadlock.

### LINKED ISSUE #2234: ReadSchema deadlock

ohadzeliger:
Create a reproducing test for the Lucene thread deadlock

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
@@ -94,7 +94,6 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import static com.apple.foundationdb.record.TestHelpers.assertLoadRecord;
-import static com.apple.foundationdb.record.lucene.LuceneIndexTest.generateRandomWords;
 import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.SIMPLE_TEXT_SUFFIXES;
 import static com.apple.foundationdb.record.lucene.LucenePlanMatchers.group;
 import static com.apple.foundationdb.record.lucene.LucenePlanMatchers.query;
@@ -103,7 +102,6 @@ import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.concatenateFields;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
-import static com.apple.foundationdb.record.provider.common.text.TextSamples.ROMEO_AND_JULIET_PROLOGUE;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.bounds;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.coveringIndexScan;
 import static com.apple.foundationdb.record.query.plan.match.PlanMatchers.fetch;
@@ -133,6 +131,9 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 @Tag(Tags.RequiresFDB)
 public class FDBLuceneQueryTest extends FDBRecordStoreQueryTestBase {
 
+    // The fork/join pool parallelism factor (the number of active threads in the pool).
+    private static final int PARALLELISM = 8;
+
     @BeforeAll
     public static void setup() {
         //set up the English Synonym Map so that we don't spend forever setting it up for every test, because this takes a long time
@@ -144,14 +145,14 @@ public class FDBLuceneQueryTest extends FDBRecordStoreQueryTestBase {
             TextSamples.AETHELRED,
             TextSamples.PARTIAL_ROMEO_AND_JULIET_PROLOGUE,
             TextSamples.FRENCH,
-            ROMEO_AND_JULIET_PROLOGUE,
+            TextSamples.ROMEO_AND_JULIET_PROLOGUE,
             TextSamples.ROMEO_AND_JULIET_PROLOGUE_END
     ));
 
     final List<String> textSamples = Arrays.asList(
-            ROMEO_AND_JULIET_PROLOGUE,
+            TextSamples.ROMEO_AND_JULIET_PROLOGUE,
             TextSamples.AETHELRED,
-            ROMEO_AND_JULIET_PROLOGUE,
+            TextSamples.ROMEO_AND_JULIET_PROLOGUE,
             TextSamples.ANGSTROM,
             TextSamples.AETHELRED,
             TextSamples.FRENCH
@@ -807,7 +808,7 @@ public class FDBLuceneQueryTest extends FDBRecordStoreQueryTestBase {
         for (int i = 0; i < 200; i++) {
             try (FDBRecordContext context = openContext()) {
                 openRecordStore(context);
-                String[] randomWords = generateRandomWords(500);
+                String[] randomWords = LuceneIndexTestUtils.generateRandomWords(500);
                 final TestRecordsTextProto.SimpleDocument dylan = TestRecordsTextProto.SimpleDocument.newBuilder()
                         .setDocId(i)
                         .setText(randomWords[1])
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -138,14 +144,14 @@ public static void setup() {
             TextSamples.AETHELRED,
             TextSamples.PARTIAL_ROMEO_AND_JULIET_PROLOGUE,
             TextSamples.FRENCH,
-            TextSamples.ROMEO_AND_JULIET_PROLOGUE,
```
ScottDugas, line 141: We should keep this consistent with the rest, and avoid the static import

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1271,4 +1277,55 @@ void continuations(boolean sorted) throws Exception {
             }
         }
     }
+
+    /**
+     * This tests is reproducing a thread deadlock that happened with Lucene threads when reading schema information.
```
ScottDugas, line 1282:
```suggestion
     * This test is reproducing a thread deadlock that happened with Lucene threads when reading schema information.
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1271,4 +1277,55 @@ void continuations(boolean sorted) throws Exception {
...
+     * The fix for the deadlock has removed the {@code Map.computeIfAbsent} call (with the {@code synchronized} block)
+     * that blocked the {@code asyncToSync} calls.
+     * This test creates a large enough set of {@link org.apache.lucene.index.SegmentReader} futures so as to flood the thread pool,
+     * and then queries Lucene. THe large number of futures in the pool, together with the {@code synchronized} block
```
ScottDugas, line 1286:
```suggestion
     * and then queries Lucene. The large number of futures in the pool, together with the {@code synchronized} block
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1271,4 +1277,55 @@ void continuations(boolean sorted) throws Exception {
...
+     * would cause a deadlock.
+     */
+    @Test
+    void testQueryWithManyDocuments() {
+        String[] words = generateRandomWords(1000);
```
ScottDugas, line 1291: It's probably worth commenting here that we generate random words, rather than using the `TextSamples`, or choosing random words from the `TextSamples`, because Lucene is too efficient at compressing those, so even with a lot of documents (100k+ If I remember correctly), it still falls short of creating enough segments to cause a problem.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1271,4 +1277,55 @@ void continuations(boolean sorted) throws Exception {
...
+        FDBDatabaseFactory.instance().setExecutor(new ForkJoinPool(Runtime.getRuntime().availableProcessors(),
+                ForkJoinPool.defaultForkJoinWorkerThreadFactory,
+                null, false));
```
ScottDugas, line 1294: Setting it to `availableProcessors` might more accurately represent how someone would use a `ForkJoinPool` as the executor, but it makes this test less reproducible. For example if you ran this on a machine with 128 cores, it would probably pass even with the deadlock, but reproduce the issue on a machine with 4 cores (like a docker container, or a rather old computer).
4 might be too small and detect other deadlocks, so maybe 8 or 16 is more appropriate for now, although I think, ideally we would be able to use lucene with a million segments with a parallelism of 1.
```suggestion
        FDBDatabaseFactory.instance().setExecutor(new ForkJoinPool(4,
                ForkJoinPool.defaultForkJoinWorkerThreadFactory,
                null, false));
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1271,4 +1277,55 @@ void continuations(boolean sorted) throws Exception {
...
+        FDBDatabaseFactory.instance().getDatabase().setAsyncToSyncTimeout( event -> {
+            // Make AsyncToSync calls timeout after one second
```
ScottDugas, line 1296:
```suggestion
            // Make AsyncToSync calls timeout after one second, otherwise a deadlock would just result in the test taking forever
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ScottDugas, line 1291: IF `generateRandomWords` is `public`, and used by a variety of tests, it should probably be moved to `LuceneIndexTestutils`

ScottDugas, line 1291: I don't understand the api for `generateRandomWords`, but you're not using it correctly here.
IIUC it generates an array of two elements:
`words[1]` is a string of 1,000 randomly generated words, separated by a string
`words[0]` is the last word in `words[1]`

The otherr two usages just use `words[1]` and completely forget `words[0]`.

It might also be simpler by having a `String generateRandomWord(int minLength, int maxLength)`
And then you could have two methods that easily use that:
```
Stream.generate(() -> generateRandomWord(3, 10)).limit(1000).collect(Collectors.joining(" "));
```
and
```
Stream.generate(() -> generateRandomWords(3, 10)).limit(1000).collect(Collectors.toList())
```
and use the later in this test, and then randomly chose from that. Or you could perhaps use the former repeatedly, which would basically result in every document having completely unique sets of words, at least until you generate 141 trillion words. It would be very detached from legitimate use cases of lucene, but might force it to make many segments.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ohadzeliger: Sure, done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ohadzeliger: Done.

ohadzeliger: Done.

ohadzeliger: Sorry, my mistake. The actual usage of generateRandomWords(500) would generate the string that we eventually need, so no need to process further. The code was the result of having the prior random word selector in place.
Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ohadzeliger: Good point. The test should be reproducible under all conditions and the existing code would make it unpredictable.
Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
ohadzeliger: Done.

ScottDugas (approved):

PR #2238, opened 2023-07-21
Resolves #2237: Remove unnecessary async code from writeSchema

### LINKED ISSUE #2237: FDBDirectory.writeAsync uses supplyAsync unnecessarily
FDBDirectory.writeAsync uses supplyAsync but the only usage immediately calls `get`

https://github.com/FoundationDB/fdb-record-layer/blob/832e2d3baa8a722778167ee92cacdf184086ae6e/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java#L390

ScottDugas:
it used supplyAsync, but always called get immediately on the result

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -387,18 +387,16 @@ public class FDBDirectory extends Directory  {
         });
     }
 
-    public CompletableFuture<Integer> writeSchema(@Nonnull List<Long> bitSetWords, @Nonnull final byte[] value) {
-        return CompletableFuture.supplyAsync( () -> {
-            context.increment(LuceneEvents.Counts.LUCENE_WRITE_SIZE, value.length);
-            context.increment(LuceneEvents.Counts.LUCENE_WRITE_CALL);
-            if (LOGGER.isTraceEnabled()) {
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -387,18 +387,16 @@ public CompletableFuture<Integer> writeData(final long id, final int block, @Non
...
-        });
+    public int writeSchema(@Nonnull List<Long> bitSetWords, @Nonnull final byte[] value) {
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_SIZE, value.length);
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_CALL);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(getLogMessage("Write lucene data",
+                    LuceneLogMessageKeys.DATA_SIZE, value.length,
+                    LuceneLogMessageKeys.ENCODED_DATA_SIZE, value.length));
+        }
+        context.ensureActive().set(schemaSubspace.pack(Tuple.from(bitSetWords)), value);
```
ohadzeliger, line 398: IIUC, calls to `set` are synchronous (deferred until commit) so this change will not make a change to the flow esp. because of the get() call immediately following

ohadzeliger (approved):

PR #2241, opened 2023-07-25
Resolve #2240: Indexing throttle: reduce limit faster during repeatin…

### LINKED ISSUE #2240: Indexing throttle: reduce limit faster during repeating failures
The current algorithm has a hidden conjecture - assuming that the majority of the indexing time is spent during records iterations. Lucene seems to change that, and spend a lot of time at transaction commit (merging?). To address slow responses for repeating failures, track consecutive indexing failures and decrease limit more aggressively when it happens.

jjezra:
…g failures

   The current algorithm has a hidden conjecture - assuming that the majority of the indexing time is spent during records iterations. Lucene index
   seems to change that, and spend a lot of time at transaction commit (merging?). To address slow responses for repeating failures, track consecutive
   indexing failures and decrease limit more aggressively when it happens.
   Also: for debugging, log store timer diff between failures.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
@@ -28,6 +28,8 @@ import com.apple.foundationdb.record.IndexState;
 import com.apple.foundationdb.record.RecordCoreStorageException;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
+import com.apple.foundationdb.record.provider.common.StoreTimer;
+import com.apple.foundationdb.record.provider.common.StoreTimerSnapshot;
 import com.apple.foundationdb.record.provider.foundationdb.runners.ExponentialDelay;
 import com.apple.foundationdb.util.LoggableException;
 import com.google.common.annotations.VisibleForTesting;
@@ -88,6 +90,9 @@ public class IndexingThrottle {
         private int consecutiveSuccessCount = 0;
         private long forcedDelayTimestampMilliSeconds = 0;
         private long recordsScannedSinceForcedDelayMilliSeconds = 0;
+        private long consecutiveFailureCount = 0;
+        private StoreTimerSnapshot lastFailureSnapshot = null;
+
 
         Booker(@Nonnull IndexingCommon common) {
             this.common = common;
@@ -153,8 +158,9 @@ public class IndexingThrottle {
                            @Nullable List<Object> additionalLogMessageKeyValues) {
             // TODO: decrease the limit only for certain errors
             countFailedTransactions++;
+            consecutiveFailureCount++;
             long oldLimit = recordsLimit;
-            recordsLimit = Math.max(1, Math.min(lastFailureRecordsScanned - 1, ((lastFailureRecordsScanned * 9) / 10)));
+            recordsLimit = Math.max(1, Math.min(lastFailureRecordsScanned - 1, ((lastFailureRecordsScanned * oneToNineFactor(consecutiveFailureCount)) / 10)));
             if (LOGGER.isInfoEnabled()) {
                 final KeyValueLogMessage message = KeyValueLogMessage.build("Lessening limit of online index build",
                                 LogMessageKeys.ERROR, fdbException.getMessage(),
@@ -165,8 +171,19 @@ public class IndexingThrottle {
                 if (additionalLogMessageKeyValues != null) {
                     message.addKeysAndValues(additionalLogMessageKeyValues);
                 }
+                addStoreTimerAtFailure(message);
                 LOGGER.info(message.toString(), fdbException);
             }
+        }
+
+        private static long oneToNineFactor(long count) {
+            if (count > 7) {
+                return 1; // panic mode after the 7th failure
+            }
+            if (count > 3) {
+                return 5; // 50% after the third failure
+            }
+            return 10 - Math.max(1, count);
         }
 
         void handleLimitsPostRunnerTransaction(@Nullable Throwable exception,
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -153,8 +158,9 @@ void decreaseLimit(@Nonnull FDBException fdbException,
                            @Nullable List<Object> additionalLogMessageKeyValues) {
             // TODO: decrease the limit only for certain errors
             countFailedTransactions++;
+            consecutiveFailureCount++;
             long oldLimit = recordsLimit;
-            recordsLimit = Math.max(1, Math.min(lastFailureRecordsScanned - 1, ((lastFailureRecordsScanned * 9) / 10)));
+            recordsLimit = Math.max(1, Math.min(lastFailureRecordsScanned - 1, ((lastFailureRecordsScanned * oneToNineFactor(consecutiveFailureCount)) / 10)));
```
ScottDugas, line 163: This approach doesn't work around the retries by the underlying runner, correct? So if the runner is set to retry 10 times, you'll still do 10 retries before decreasing the limit. Is that correct?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -247,6 +265,15 @@ void refreshConfigLimits() {
                 recordsLimit = maxLimit;
             }
         }
+
+        private void addStoreTimerAtFailure(KeyValueLogMessage message) {
+            final FDBStoreTimer timer = common.getRunner().getTimer();
+            if (timer != null) {
+                StoreTimer metricsDiff = lastFailureSnapshot == null ? timer : StoreTimer.getDifference(timer, lastFailureSnapshot);
+                lastFailureSnapshot = StoreTimerSnapshot.from(timer);
```
ScottDugas, line 273: I don't think this will get the snapshot you want. Namely, this gathers the snapshot since the last failure, so if you have 1000 transactions successfully, and then fail, it will have the store timer for all 1001 transactions.
You could gather the snapshot at the beginning of every transaction, but that might be expensive. It might make sense to clear the snapshot after a success, and if you don't have one, gather it.

This way, if you have two failures in a row, it will include the snapshot, but if you have sporadic failures, that succeed on retry we won't log any snapshots.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -165,10 +171,21 @@ void decreaseLimit(@Nonnull FDBException fdbException,
...
         }
 
+        private static long oneToNineFactor(long count) {
+            if (count > 7) {
+                return 1; // panic mode after the 7th failure
+            }
+            if (count > 3) {
+                return 5; // 50% after the third failure
+            }
+            return 10 - Math.max(1, count);
```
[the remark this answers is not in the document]
jjezra: It makes little difference, I think: 
0.9, 0.72, 0.504, 0.252, 0.126, 0.063, 0.0315, 0.00315, 0.000315...
vs. 
0.9, 0.72, 0.504, 0.3024, 0.1512, 0.06048, 0.018144, 0.0036288, 0.00036288, 0.00036288...
(I liked the cleaner 1/2 and 1/10 factors, but can change it if you want)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
jjezra: Yes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
jjezra: IIUIC, by clearing the snapshot after success the next displayed counters will be since the beginning of time (well, the indexing session). I had assumed that it will be better to have it this way - when you can see how many "things" had happened between failures and have a precise measurement for two consecutive ones.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -165,10 +171,21 @@ void decreaseLimit(@Nonnull FDBException fdbException,
                 if (additionalLogMessageKeyValues != null) {
                     message.addKeysAndValues(additionalLogMessageKeyValues);
                 }
+                addStoreTimerAtFailure(message);
                 LOGGER.info(message.toString(), fdbException);
             }
         }
 
+        private static long oneToNineFactor(long count) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -325,6 +332,7 @@ public <R> CompletableFuture<R> buildCommitRetryAsync(@Nonnull final BiFunction<
             loadConfig();
             return common.getRunner().runAsync(context -> common.getRecordStoreBuilder().copyBuilder().setContext(context).openAsync().thenCompose(store -> {
                 expectedIndexStatesOrThrow(store, context);
+                booker.resetStoreTimerSnapshot();
```
ScottDugas, line 335: IIUC this will be off from the transaction. So, it could be something like:
1. getReadVersion
2. resetStoreTimerSnapshot
3. indexSome records
4. commit (successfully)
5. getReadVersion & open store -> times out

The metrics would include everything except the setup for the previous transaction, but not the current transaction.


If you did this instead in `booker.handleLimitsPostRunnerTransaction` when there's success, and before calling `AsyncUtil.whileTrue`, I think it would have the timer for the entire duration of the transaction. 

All of this is an indication that we should stop using the runner within index building, but that's a much bigger change.

ScottDugas (changes requested):

ScottDugas (approved):

PR #2244, opened 2023-07-28
Add `ConditionSelectorValue`.

hatyo:
This introduces `ConditionSelectorValue` which can be used in conjunction with `PickValue` to implement features such as SQL `case-when`

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
@@ -1,0 +1,101 @@
+/*
+ * ConditionSelectorValue.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2023 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades.values;
+
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.ObjectPlanHash;
+import com.apple.foundationdb.record.PlanHashable;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
+import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+/**
+ * This returns the index of the first implication that is satisfied.
+ *
+ * @see PickValue for more information.
+ */
+public class ConditionSelectorValue extends AbstractValue {
+
+    private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Implication-Selector-Value");
```
normen662, line 42: Can you make this the same-ish as the class name?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
```diff
@@ -0,0 +1,101 @@
...
+            if (Boolean.TRUE.equals(result)) {
+                return i;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public String toString() {
+        return "ImplicationSelector";
```
normen662, line 89: same here

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public int planHash(@Nonnull final PlanHashKind hashKind) {
+        return PlanHashable.objectsPlanHash(hashKind, BASE_HASH, implications);
+    }
+
+    @Override
+    public int hashCodeWithoutChildren() {
```
normen662, line 98: shouldn't there also be a `equalsWithoutChildren`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConditionSelectorValue.java
hatyo: It is delegating to `AbstractValue` which is doing it correctly:

```java
    default boolean equalsWithoutChildren(@Nonnull final Value other,
                                          @Nonnull final AliasMap equivalenceMap) {
        if (this == other) {
            return true;
        }

        return other.getClass() == getClass();
    }
```

normen662 (approved):

PR #2245, opened 2023-08-01
Support dry updates

pengpeng-lu:


.idea/compiler.xml
hatyo, line 1: Changes to this file don't seem to be related to your PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -137,22 +137,32 @@
...
+ * {@link Subspace}
+ * within the context of a single transaction, i.e., with the same {@link FDBRecordContext}. This is because the
+ * {@code FDBRecordStore}
  * object maintains state about certain uncommitted operations, and concurrent access through two objects will not see
- * changes to this in-memory state. See <a href="https://github.com/FoundationDB/fdb-record-layer/issues/489">Issue #489</a>
+ * changes to this in-memory state. See <a href="https://github.com/FoundationDB/fdb-record-layer/issues/489">Issue
+ * #489</a>
  * for more details. Note also that the record stores returned by {@link #getTypedRecordStore(RecordSerializer)} and
- * {@link #getUntypedRecordStore()} will share an {@code FDBRecordStore} with the record store on which they are called,
- * so it <em>is</em> safe to have a typed- and untyped-record store open over the same {@code Subspace} within the context
```
hatyo, line 155: Could you please revert these changes since they are irrelevant?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -331,10 +343,13 @@ public int getFormatVersion() {
...
+     * <p>
+     * After calling {@link FDBRecordStore.Builder#open} or {@link #checkVersion} directly, this will be the value
+     * stored in the store's info header.
+     * <p>
+     * This version is returned from {@link UserVersionChecker#checkUserVersion} and does not have any meaning within
+     * the Record Layer core.
      *
-     * After calling {@link FDBRecordStore.Builder#open} or {@link #checkVersion} directly, this will be the value stored in the store's info header.
-     *
-     * This version is returned from {@link UserVersionChecker#checkUserVersion} and does not have any meaning within the Record Layer core.
```
hatyo, line 337: Ditto (and all other doc-reformatting changes).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -476,6 +498,47 @@ protected <M extends Message> CompletableFuture<FDBStoredRecord<M>> saveTypedRec
...
+                            LogMessageKeys.EXPECTED_TYPE, recordType.getName());
+                }
+            }
+            // (TODO): recordBuilder.setSizeInfo(sizeInfo);
+            // (TODO): update indexState in recordStoreState
+            return CompletableFuture.completedFuture(recordBuilder.build());
+        });
+        // (TODO): timer should be changed here
+        return context.instrument(FDBStoreTimer.Events.SAVE_RECORD, result);
+    }
```
hatyo, line 540: This is a verbatim copy of `saveTypedRecord` without this line:

```java
final FDBStoredRecord<M> newRecord = serializeAndSaveRecord(typedSerializer, recordBuilder, metaData, oldRecord);
```

I think it makes sense to think about some refactoring here to reduce code duplication, e.g. by means of introducing a boolean flag in `saveTypedRecord`, or some sort of `read-only` attribute defined in `FDBRecordStore` that is checked in `saveTypedRecord` s.t. we conditionally call `serializeAndSaveRecord`.

Regardless, I think this should be well documented.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/UpdateExpression.java
```diff
@@ -83,12 +83,14 @@ public class UpdateExpression implements RelationalExpressionWithChildren, Plann
...
     private final Supplier<Integer> hashCodeWithoutChildrenSupplier;
+    private final boolean dryRun;
 
     public UpdateExpression(@Nonnull final Quantifier.ForEach inner,
                             @Nonnull final String targetRecordType,
                             @Nonnull final Type.Record targetType,
                             @Nonnull final Descriptors.Descriptor targetDescriptor,
-                            @Nonnull final Map<FieldValue.FieldPath, Value> transformMap) {
+                            @Nonnull final Map<FieldValue.FieldPath, Value> transformMap,
+                            final boolean dryRun) {
```
hatyo, line 93: I wouldn't model `dryRun` in plan operators since this is not relevant for planning. It is just a physical property that controls an execution workflow.

From planner's perspective, the update plan will (and should be) planned exactly the same way. I am thinking about adding this value to [`ExecuteProperties`](https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ExecuteProperties.java) and set it when we're about to execute the physical plan.

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -463,15 +481,19 @@ protected <M extends Message> CompletableFuture<FDBStoredRecord<M>> saveTypedRec
                             LogMessageKeys.EXPECTED_TYPE, recordType.getName());
                 }
             }
-            final FDBStoredRecord<M> newRecord = serializeAndSaveRecord(typedSerializer, recordBuilder, metaData, oldRecord);
-            if (oldRecord == null) {
-                addRecordCount(metaData, newRecord, LITTLE_ENDIAN_INT64_ONE);
+            if (isDryRun) {
+                return CompletableFuture.completedFuture(recordBuilder.build());
```
hatyo, line 485: The record being built here will not have its (fake) new size set correctly. This is probably needed as the customer is interested in knowing the byte-size of the records.

pengpeng-lu: good point, I changed the `serializeAndSaveRecord` method to include a dryRun option.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -513,18 +535,22 @@ private <M extends Message> CompletableFuture<FDBStoredRecord<M>> loadExistingRe
 
     @Nonnull
     private <M extends Message> FDBStoredRecord<M> serializeAndSaveRecord(@Nonnull RecordSerializer<M> typedSerializer, @Nonnull final FDBStoredRecordBuilder<M> recordBuilder,
-                                                                          @Nonnull final RecordMetaData metaData, @Nullable FDBStoredSizes oldSizeInfo) {
+                                                                          @Nonnull final RecordMetaData metaData, @Nullable FDBStoredSizes oldSizeInfo, final boolean isDryRun) {
```
alecgrieser, line 538: Why is this method called at all if `isDryRun` is true?

It seems like it would be less invasive to instead have a method to generate the new `FDBStoredRecord<M>` and then call that from `saveTypedRecord`. It also appears like the hard part here is mainly calculating the `SizeInfo`. Arguably, we could just say that dry run updates are not guaranteed to have accurate `SizeInfo` information

alecgrieser (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -1177,6 +1177,29 @@ public void testNoUnion() {
         }
     }
 
+    @Test
+    public void testDryRunSaveRecord() throws Exception {
```
alecgrieser, line 1181: It would be nice to have more tests here so that we're more confident we're not accidentally update the database in dry-run mode. One here that seems like we'd want to check is indexes: if you save a record with an indexed field in dry-run mode, then you shouldn't find index entries from that record when scanning the index.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -1177,6 +1177,29 @@ public void testNoUnion() {
...
+        CompletableFuture<FDBStoredRecord<Message>> future1;
+        TestRecords1Proto.MySimpleRecord.Builder recBuilder = TestRecords1Proto.MySimpleRecord.newBuilder();
+        TestRecords1Proto.MySimpleRecord record1 = recBuilder.setRecNo(1).build();
+        TestRecords1Proto.MySimpleRecord record2 = recBuilder.setRecNo(2).setNumValue3Indexed(3).build();
+        try (FDBRecordContext context = openContext()) {
+            openSimpleRecordStore(context);
+            future1 = recordStore.dryRunSaveRecordAsync(record1, FDBRecordStoreBase.RecordExistenceCheck.ERROR_IF_EXISTS);
+            // assert returns the message to be saved
+            assertEquals(record1, future1.get().getRecord());
```
alecgrieser, line 1190: If we care enough about the `SizeInfo` to calculate it correctly, it feels like we should also assert on it. We could, for example, take the `SizeInfo` from an `FDBStoredRecord` that we "saved" with a dry update, and then compare it to the `SizeInfo` of an actual save of the same record that we do later.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryUpdatePlan.java
```diff
@@ -88,7 +88,11 @@ public PipelineOperation getPipelineOperation() {
     @Nonnull
     @Override
     public <M extends Message> CompletableFuture<FDBStoredRecord<M>> saveRecordAsync(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final M message) {
-        return store.saveRecordAsync(message, FDBRecordStoreBase.RecordExistenceCheck.ERROR_IF_NOT_EXISTS_OR_RECORD_TYPE_CHANGED);
+        if (isDryRun()) {
```
alecgrieser, line 91: Rather than adding an `isDryRun()` check here, I think we should update `saveRecordAsync` in `RecordQueryAbstractDataModificationPlan` to take a `dryRun` Boolean (or an `ExecuteProperties`). Then the `executePlan` method in the abstract class can pass in the correct value from the `ExecuteProperties` it's given. That way, we don't have to mutate the state of the plan object (which generally should be immutable, as it can be cached and shared across multiple threads). Also, it looks like `RecordQueryInsertPlan` doesn't honor the dry-run flag, which I think would become more apparent if the abstract method was modified.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -1177,6 +1177,29 @@ public void testNoUnion() {
...
+            commit(context);
+        }
+        // assert dryRun doesn't save the record
+        try (FDBRecordContext context = openContext()) {
+            openSimpleRecordStore(context, metaDataBuilder -> metaDataBuilder.updateRecords(TestRecords1Proto.getDescriptor()));
+            assertNull(recordStore.loadRecord(Tuple.from(1)));
+            assertEquals(record2, recordStore.loadRecord(Tuple.from(2)).getRecord());
+            commit(context);
+        }
+    }
```
alecgrieser, line 1201: We should probably have tests that validate that the update (and insert and potentially delete) plans honor the dry-run flag in the `ExecuteProperties`. There are some tests of those plans in `FDBModificationQueryTest`, and it would be nice to execute either those same tests or similar new tests in dry-run mode and validate that database mutations are not made.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAbstractDataModificationPlan.java
```diff
@@ -126,6 +126,7 @@ public abstract class RecordQueryAbstractDataModificationPlan implements RecordQ
     private final Supplier<Integer> planHashForContinuationSupplier;
     @Nonnull
     private final Supplier<Integer> planHashForWithoutLiteralsSupplier;
+    private boolean dryRun;
```
alecgrieser, line 129: I have a suggestion for refactoring that I left in `RecordQueryUpdatePlan` that would allow this field to be removed. In general, having mutable state in the plan is dangerous, because plan objects can be cached and re-used across threads, and if different threads are run with different values of the dry-run flag, the value could get corrupted.

alecgrieser (changes requested):
I left a bit more of a review, but to summarize:

1. I'm a little concerned about the number of places that needed to be modified in `FDBRecordStore` and `SaveWithSplitHelper` in order to make this work. There's a small concern about performance here (due to the extra branches), but my bigger concern is that we would have accidentally missed a spot. My suggestion for that is to refactor `FDBRecordStore` a bit, and have there be a single guard on `isDryRun` that either executes `saveAndSerializeRecord` OR calculates an "estimated `SizeInfo`" and attaches it to the new record. Ideally, I think that means that `SaveWithSplitHelper` can be reverted to its old state (potentially with a new method to get the estimated size info)
2. We should avoid having mutable state in the plan objects. I go into how I'd suggest changing that a bit more in the comments
3. There should be tests of the query modification plans to ensure that they honor the dry-run flags.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
alecgrieser: I guess I'm less sure that we care about having the exact right `SizeInfo` values here. If we can get the right values, that's great (and may make comparisons with alternative update approaches cleaner), but if it ends up being too difficult to get right, it doesn't seem like the worst thing to jettison.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
pengpeng-lu: this method is called when `isDryRun` = true to set `SizeInfo`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInsertPlan.java
```diff
@@ -77,7 +77,7 @@ public PipelineOperation getPipelineOperation() {
 
     @Nonnull
     @Override
-    public <M extends Message> CompletableFuture<FDBStoredRecord<M>> saveRecordAsync(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final M message) {
+    public <M extends Message> CompletableFuture<FDBStoredRecord<M>> saveRecordAsync(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final M message, final boolean isDryRun) {
         return store.saveRecordAsync(message, FDBRecordStoreBase.RecordExistenceCheck.ERROR_IF_EXISTS);
```
alecgrieser, line 81: Shouldn't this call `dryRunSaveRecordAsync` if `isDryRun` is true? I guess the PR only claims to be about using dry run for updates, but it seems weird that it isn't also implemented for inserts and delete plans

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
alecgrieser: Yeah. I think it would be better, though, if this were only called if `isDryRun` is false, and we find a different way of calculating the `SizeInfo` information (or say that we don't care about getting the `SizeInfo` right, given that dry updates aren't really about testing if the size metrics collection logic here in the Record Layer is accurate or not). Ideally, it would seem like we want only one place in the serialize pipeline that needs to check the value of `isDryRun`. (My main concern here being that we either missed a spot, or that someone in the future modifying this code will miss a spot, and that's harder to do if all of the places that _actually_ write to the database are coalesced into a single method.)

alecgrieser (changes requested):
I think there are still three outstanding things:

1. I still think it would be better to refactor this so that we have fewer conditionals on `isDryRun`, namely avoiding calling `saveAndSerializeRecord` at all in dry-run mode
2. I believe insert and delete plans still actually do database operations in dry-run mode
3. I don't think there are tests that validate that indexes aren't written to during a dry-run build

Based on visual inspection of the code, I believe that that last one is done correctly, we just don't test it, so in theory, we could take this and then add the test. The second point is maybe more serious, though it speaks as much to the API as anything else. In theory, we could say that dry-run mode only applies to updates, and not inserts or deletes, though that feels...weird. That first point is perhaps the most matter-of-opinion, though I think we'd be more confident in this with the suggested refactoring

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTest.java
```diff
@@ -1177,6 +1177,56 @@ public void testNoUnion() {
...
+            openSimpleRecordStore(context, metaDataBuilder -> metaDataBuilder.updateRecords(TestRecords1Proto.getDescriptor()));
+            assertNotNull(result2.getRecord(), "record2 was missing");
+            assertEquals(record1, result2.getRecord());
+            assertEquals(Collections.singletonList(Tuple.from("abc", 1)),
+                    recordStore.scanIndex(recordStore.getRecordMetaData().getIndex("MySimpleRecord$str_value_indexed"),
+                            IndexScanType.BY_VALUE, TupleRange.ALL, null, ScanProperties.FORWARD_SCAN).map(IndexEntry::getKey).asList().join());
+
+            commit(context);
+        }
+        // (TODO) assert dryRun doesn't change index
```
alecgrieser, line 1227: I think this `TODO` can be removed now

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
```diff
@@ -393,13 +441,13 @@ public void saveWithSplitAndIncompleteVersions(SplitHelperTestConfig testConfig)
 
             commit(context);
             versionstamp = context.getVersionStamp();
-            if (!testConfig.omitUnsplitSuffix) {
+            if (!testConfig.omitUnsplitSuffix && !testConfig.isDryRun) {
```
alecgrieser, line 444: Why is this necessary? It seems like the whether the config is dry-run shouldn't affect this

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBModificationQueryTest.java
```diff
@@ -207,31 +214,33 @@ public void testPlanInsertExpression() throws Exception {
...
-                    case 200:
-                        Assertions.assertEquals("Heirloom Cafe", record.getField(name));
-                        break;
-                    default:
-                        Assertions.fail("unexpected record");
-                }
-                return record;
-            }, c -> { });
-            Assertions.assertEquals(2, resultValues.size());
+            for (ExecuteProperties p : List.of(ExecuteProperties.newBuilder().setDryRun(true).build(), ExecuteProperties.SERIAL_EXECUTE)) {
```
alecgrieser, line 217: If I'm reading this correctly, this doesn't actually assert that the update doesn't happen if `dryRun` mode is true. It would be nice if, for example, the `selectPlan` were executed after each execution, and it validated that the select plan returned no results when `dryRun` was `true`.

alecgrieser (commented):
I still think that we want to also have delete plans honor the `dryRun` flag, as otherwise, that flag means "run a plan that would modify the data, but don't actually write any changes, unless it's deletes, in which case do actually issue the deletes", which seems less understandable than "run a plan that would modify the data, but don't actually modify the data". It may not actually make sense to issue dry delete plans, if we don't surface enough information for the user to know what would have been deleted, but it would at least make the dry-run flag safer

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/SplitHelperTest.java
pengpeng-lu: I think `versionStamp` is set when `transaction.commit()`, so if it is dry run, no versionStamp is set.

alecgrieser: Ah, i see. I’m actually a little surprised this returns `null` now if `omitVersionSuffix` is set, but maybe I’m misremembering how that value in the context gets populated. I would have thought it would show up on every commit that isn’t read only. Hmm

alecgrieser (approved):
LGTM. We should be sure to squash during the merge

alecgrieser (approved):

hatyo (approved):

PR #2255, opened 2023-08-09
Support promoting null to another type in PromoteValue.inject

arnaud-lacurie:


arnaud-lacurie:
@normen662 , where is a good place for me to test that additional piece of code?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
```diff
@@ -97,6 +98,19 @@ public boolean equalsWithoutChildren(@Nonnull final Value other, @Nonnull final
         return ordinal == otherConstantObjectValue.ordinal;
     }
 
+    @Nonnull
+    @Override
+    public boolean canResultInType(@Nonnull final Type type) {
+        return resultType.isUnresolved() || resultType.getTypeCode() == Type.TypeCode.NULL;
```
normen662, line 104: can this happen?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
arnaud-lacurie: Removed the `isUnresolved`

normen662 (approved):

PR #2260, opened 2023-08-15
Add a test that fills a store and then does updates

ScottDugas:


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
```diff
@@ -0,0 +1,409 @@
...
+
+@Tag(Tags.RequiresFDB)
+@Tag(Tags.Slow)
+@Timeout(value = 8, unit = TimeUnit.HOURS)
+public class LuceneScaleTest extends FDBRecordStoreTestBase {
+    private static final Logger logger = LogManager.getLogger(LuceneScaleTest.class);
+
+    private static final String RECORD_COUNT_COLUMN = "recordCount";
+    private static final String TOTAL_TIME_COLUMN = "totalTime";
+    static final List<String> CSV_COLUMNS = List.of(RECORD_COUNT_COLUMN, TOTAL_TIME_COLUMN, "bytes_deleted_count",
```
ScottDugas, line 72: TODO at total run time of the test so far

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
```diff
@@ -0,0 +1,409 @@
...
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag(Tags.RequiresFDB)
+@Tag(Tags.Slow)
+@Timeout(value = 8, unit = TimeUnit.HOURS)
+public class LuceneScaleTest extends FDBRecordStoreTestBase {
```
MMcM, line 67: This should probably do like `FDBRecordStoreRepairTest` and pick a different keyspace. It suppresses `clear` but can be affected by running other tests.

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
```diff
@@ -0,0 +1,409 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag(Tags.RequiresFDB)
+@Tag(Tags.Slow)
+@Timeout(value = 8, unit = TimeUnit.HOURS)
```
MMcM, line 66: This turns out not to be long enough here, so maybe multiply by some.

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
```diff
@@ -0,0 +1,409 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static com.apple.foundationdb.record.metadata.Key.Expressions.concat;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.function;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag(Tags.RequiresFDB)
+@Tag(Tags.Slow)
```
MMcM, line 65: Is this enough to keep from running during CI or do we have / need a stronger one that only does manual runs?

MMcM (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
ScottDugas: I added the total test run time to the columns.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
ScottDugas: Done. I thought about adding an option so that you can easily add additional bits to differentiate separate runs, but opted against it, at least for now.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
ScottDugas: It turns out there is a `Tags.Performance`, so I switched to that. It has some trickery so that when you run in intelliJ, it will run, but shouldn't run in CI/PRB

ScottDugas:
Something got weird with my branch, so I re-opened a squashed version of this in: https://github.com/FoundationDB/fdb-record-layer/pull/2310

I didn't want to squash here, because I wanted to leave around the commits that were reverted in: https://github.com/FoundationDB/fdb-record-layer/pull/2260/commits/775c42f80b2884c3589932dfe0826f4699bea369

PR #2264, opened 2023-08-18
Resolve #2266: Update .gitignore

### LINKED ISSUE #2266: Update .gitignore to exclude .idea/palantir-java-format.xml and .DS_Store
To exclude these files:

- .idea/palantir-java-format.xml (because this is autogenerated)
- .DS_Store (because this is a Mac specific file and does not need to be committed)

butlermh:


.gitignore
```diff
@@ -74,4 +76,7 @@ protogen/
 fdb-environment.properties
 
 # Docker local support
-run/
+run/i
```
alecgrieser, line 79: Was this `i` intentional?

.gitignore
```diff
@@ -42,6 +42,8 @@ atlassian-ide-plugin.xml
 .idea/gradle.xml
 .idea/jarRepositories.xml
 .idea/shelf
+# don't include autogenerated files
+.idea/palantir-java-format.xml
```
alecgrieser, line 46: What's this for? As far as I know, we don't use the Palantir Java format plugin

alecgrieser (changes requested):

PR #2267, opened 2023-08-21
Get scan order from query rather than indexes

arnaud-lacurie:
Indexes can be read top down or bottom up. The reverse scan requirement should not come from indexes but from the query itself

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
@@ -61,6 +61,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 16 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -259,13 +260,27 @@ protected Set<? extends RelationalExpression> dataAccessForMatchPartition(@Nonnu
...
     private static List<PartialMatchWithCompensation> maximumCoverageMatches(@Nonnull final Collection<? extends PartialMatch> matches,
                                                                              @Nonnull final Set<RequestedOrdering> interestedOrderings) {
-        final var partialMatchesWithCompensation =
-                matches
-                        .stream()
-                        .filter(partialMatch -> !satisfiedOrderings(partialMatch, interestedOrderings).isEmpty())
-                        .map(partialMatch -> new PartialMatchWithCompensation(partialMatch, partialMatch.compensate()))
-                        .sorted(Comparator.comparing((Function<PartialMatchWithCompensation, Integer>)p -> p.getPartialMatch().getBindingPredicates().size()).reversed())
-                        .collect(ImmutableList.toImmutableList());
+        final List<PartialMatchWithCompensation> partialMatchesWithCompensation = new ArrayList<>();
```
normen662, line 263: This looks better than before! Thank you!

normen662 (changes requested):
This looks good in general, except it would be nice to have a dedicated test case.

PR #2277, opened 2023-08-29
resolves #2276: highlighting improvements

### LINKED ISSUE #2276: segregate different types of lucene querying
Currently, a `LuceneQueryComponent` annotated a query type, e.g. a regular query, auto complete, or a highlighting query. We lose that information when the query component gets translated into a `LuceneQueryClause`-tree. When we then execute specific logic to e.g. highlighting we have no idea which particular sub clause corresponded to highlighting and which one corresponded to a regular lucene search or auto complete. This issue is to keep track of the query types within the clauses hierarchy and to use that information to make specific (e.g. highlighting or auto complete) logic to only recognize the clauses specific to the specific use case at hand.

normen662:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQueryClause.java
```diff
@@ -23,21 +23,180 @@
...
+    protected BoundQuery toBoundQuery(@Nonnull final Query luceneQuery) {
+        return BoundQuery.ofLuceneQueryWithQueryType(luceneQuery, getQueryType());
+    }
+
+    /**
+     * Parse the Lucene query to get all the mapping from field to terms.
+     * @param query lucene query to extract all terms from
+     */
+    @Nonnull
+    protected static Map<String, Set<String>> getHighlightingTermsMap(@Nonnull final Query query) {
```
normen662, line 89: This is code I took from the highlighting code path and just morphed it into this shape without changing any of its conceptual logic.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQueryComponent.java
```diff
@@ -53,19 +53,8 @@
 public class LuceneQueryComponent implements QueryComponent, ComponentWithNoChildren {
     private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Lucene-Query");
 
-    /**
```
normen662, line 56: moved out into its own top-level enum

MMcM (approved):

PR #2278, opened 2023-08-30
Fixes #2275: Version mutations aren't cleared out by delete records where

### LINKED ISSUE #2275: Version mutations aren't cleared out by delete records where
Versionstamp-related mutations are currently stored in a buffer in the JVM and flushed during commit. This is to support multiple updates to the same record happening during a single transaction. Right before commit, this data is flushed to disk.

However, this data structure is not cleared out when there is a `deleteRecordsWhere` operation. This means that if one:

1. Inserts a record with a versionstamp then
2. Issues a `deleteRecordsWhere` operation for that record, the result is
3. The versionstamp (and versionstamp indexes) get inserted into the database, but the record itself does not

This leads to a missing split point exception when the record is read. As a fix, the versionstamp keys should be cleared out during `deleteRecordsWhere` as well. Alternatively, we could ban `deleteRecordsWhere` if there are outstanding versionstamp operations.

alecgrieser:
The main record store and index maintainers now clear out uncommitted version mutations when performing a record range delete. Previously, this failure to do so meant that it would be dangerous to perform `deleteRecordsWhere` with record versions enabled if there were uncommitted record saves performed in the same transaction. Committing the transaction would lead to the record data not being written, but the version information and the version indexes would be, leading to corrupt record stores and orphaned index entries. This cleans up that data during the clear so that we don't have uncommitted versions.

There's something to be said about the abstractions here potentially being a little off, and that we should improve some of our I/O methods here to make this bug harder to write. For now, I've just tried to patch the bug.

This fixes #2275.


alecgrieser:
Rebased to be on top of latest `main`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
```diff
@@ -800,7 +800,9 @@ public CompletableFuture<Void> deleteWhere(Transaction tr, @Nonnull Tuple prefix
         // NOTE: Range.startsWith(), Subspace.range() and so on cover keys *strictly* within the range, but we sometimes
         // store data at the prefix key itself.
         final byte[] key = state.indexSubspace.pack(prefix);
-        tr.clear(key, ByteArrayUtil.strinc(key));
+        Range indexRange = new Range(key, ByteArrayUtil.strinc(key));
```
ScottDugas, line 803: I think the following also won't work:
1. `FDBRecordStore.deleteAllRecords`
2. `KeySpacePathImpl.deleteAllDataAsync`
3. `FDBRecordStore.rebuildAllIndexes`
4. `FDBRecordStore.clearIndexData`
5. `FDBRecordStore.removeFormerIndex`

I'm guessing this leans into what you were mentioning in the description that perhaps the abstractions are broken

I see that some other index maintainers also do range clears:
1. `TimeWindowLeaderboard`
2. `Multidimensional`
7. `Rank`
8. `PermutedMinMax`
Is there something to be concerned about here? I think, in theory, they should be able to support `version()` in some capacity, even if I'm not sure that it entirely makes sense. Or that their support is already covered by them calling `super.deleteWhere`.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/StandardIndexMaintainer.java
alecgrieser: > I'm guessing this leans into what you were mentioning in the description that perhaps the abstractions are broken

Yeah, that's right. I've tried to rectify this a bit, by adding some new `clear` methods to `FDBRecordContext` that combine both clearing out the underlying data from FDB and cleaning up the version information stored in the record context. Then I switched over callers to this new method, including the methods you listed here. I've added additional tests to cover those, except for `rebuildAllIndexes`, which is a tough one given that the rebuilding of the index within the transaction should add back any of the deleted data.

> I see that some other index maintainers also do range clears. ... Is there something to be concerned about here?

None of the listed indexes currently support version mutations (and their `IndexValidator`s prevent them from being created with one), so those range clears shouldn't cause any additional problems. In most of those cases, the actual index format means it's impossible to order data correctly if there's a version column, with the possible exception of `PermutedMinMax`. That being said, switching over to the new method shouldn't be wrong and sets us up a bit better if we start to enable versions there in the future.

ScottDugas (dismissed):

alecgrieser:
Updated to fix release notes conflict

ScottDugas (approved):

PR #2280, opened 2023-08-31
Resolve #2279: Lucene: avoid segments auto merge during commit

### LINKED ISSUE #2279: Lucene: avoid segments merge at closing
To speed up indexing and record update performance, avoid merge during closing. The store will provide to the caller an indication and merge is needed, and it will be the caller's responsibility to trigger an explicit merge in a separate transaction (possibly in a background process).

jjezra:
To speed up indexing and record update performance, avoid auto merge during commit. The store will provide the caller deferred indexing policy with an indication that merge is needed, and it will be the caller's responsibility to trigger an explicit mergeIndex in a separate, possibly background, transaction

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -276,6 +276,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     private boolean recordsReadConflict;
 
     private boolean storeStateReadConflict;
+    private IndexDeferredMaintenancePolicy indexDeferredMaintenancePolicy;
 
     @Nonnull
     private final Set<String> indexStateReadConflicts = ConcurrentHashMap.newKeySet(8);
@@ -4782,6 +4783,23 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     private UninitializedRecordStoreException uninitializedStoreException(@Nonnull String msg) {
         return new UninitializedRecordStoreException(msg,
                 subspaceProvider.logKey(), subspaceProvider.toString(context));
+    }
+
+
+    /**
+     * For some indexes, there are index maintenance operations that could be either done inline, during a record update
+     * operation, or later in the background. The returned policy object lets the caller indicate to the index maintenance
+     * that he intends to trigger these deferred maintenance in another, possibly background, transaction.
+     * This feature is experimental. The default is to perform all the needed index changes inline.
+     * @return an IndexDeferredMaintenancePolicy object.
+     */
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    public synchronized IndexDeferredMaintenancePolicy getIndexDeferredMaintenancePolicy() {
+        if (indexDeferredMaintenancePolicy == null) {
+            indexDeferredMaintenancePolicy = new IndexDeferredMaintenancePolicy();
+        }
+        return indexDeferredMaintenancePolicy;
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -383,7 +383,7 @@ public class FDBStoreTimer extends StoreTimer {
         WAIT_STOP_ONLINE_INDEX_BUILD("wait for stopping ongoing online index builds"),
         /** Wait for {@link OnlineIndexer} to checking ongoing online index builds. */
         WAIT_CHECK_ONGOING_ONLINE_INDEX_BUILD("wait for checking ongoing online index builds"),
-        /** Wait for {@link OnlineIndexer} to stop ongoing online index merge(s). */
+        /** Wait for {@link OnlineIndexer} to complete ongoing online index merge(s). */
         WAIT_ONLINE_MERGE_INDEX("wait for online merge index"),
         /** Wait for {@link OnlineIndexer} to build endpoints. */
         WAIT_BUILD_ENDPOINTS("wait for building endpoints"),
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
@@ -383,6 +383,8 @@ public class FDBStoreTimer extends StoreTimer {
         WAIT_STOP_ONLINE_INDEX_BUILD("wait for stopping ongoing online index builds"),
         /** Wait for {@link OnlineIndexer} to checking ongoing online index builds. */
         WAIT_CHECK_ONGOING_ONLINE_INDEX_BUILD("wait for checking ongoing online index builds"),
+        /** Wait for {@link OnlineIndexer} to stop ongoing online index merge(s). */
+        WAIT_ONLINE_MERGE_INDEX("wait for online merge index"),
         /** Wait for {@link OnlineIndexer} to build endpoints. */
         WAIT_BUILD_ENDPOINTS("wait for building endpoints"),
         /** Wait for a record scan without an index. */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
@@ -55,7 +55,7 @@ public class IndexDeferredMaintenancePolicy {
     /**
      * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
      * If the user sets it to false, he is responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
-     * function if {@link #isMergeRequired()} returns true after the operation.
+     * function if {@link #isMergeRequired()} returns true after the requested operation.
      * @param autoMergeDuringCommit if true (default) and applicable, automatically merge during commit
      */
     public void setAutoMergeDuringCommit(final boolean autoMergeDuringCommit) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
@@ -1,0 +1,64 @@
+/*
+ * IndexDeferredMaintenancePolicy.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2023 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+/**
+ * Some store's indexes may need merging on some occasions. This helper module should allow the caller
+ * to set and probe the merge policy and merge requests.
+ */
+public class IndexDeferredMaintenancePolicy {
+    private boolean mergeRequired = false;
+    private boolean autoMergeDuringCommit = true;
+
+    /**
+     * Indicate to the caller if a deferred index merge operation is required.
+     * @return if true, an index merge if required
+     */
+    public boolean isMergeRequired() {
+        return mergeRequired;
+    }
+
+    /**
+     * Indicate to the caller if a deferred merge operation is required.
+     * @param mergeRequired if true, indicate that an index merge if required
+     */
+    public void setMergeRequired(final boolean mergeRequired) {
+        this.mergeRequired = mergeRequired;
+    }
+
+    /**
+     * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
+     * @return true if should merge
+     */
+    public boolean shouldAutoMergeDuringCommit() {
+        return autoMergeDuringCommit;
+    }
+
+    /**
+     * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
+     * If the user sets it to false, he is responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
+     * function if {@link #isMergeRequired()} returns true after the operation.
+     * @param autoMergeDuringCommit if true (default) and applicable, automatically merge during commit
+     */
+    public void setAutoMergeDuringCommit(final boolean autoMergeDuringCommit) {
+        this.autoMergeDuringCommit = autoMergeDuringCommit;
+    }
+}
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
@@ -29,7 +29,8 @@ public class IndexDeferredMaintenancePolicy {
     private boolean autoMergeDuringCommit = true;
 
     /**
-     * Indicate to the caller if a deferred index merge operation is required.
+     * Indication that a deferred index merge operation is required. This function may be used by the
+     * caller, to check  if the index maintainer requested a deferred merge.
      * @return if true, an index merge if required
      */
     public boolean isMergeRequired() {
@@ -37,7 +38,7 @@ public class IndexDeferredMaintenancePolicy {
     }
 
     /**
-     * Indicate to the caller if a deferred merge operation is required.
+     * Indicate to the caller if a deferred merge operation is required. This function is used by the index maintainer.
      * @param mergeRequired if true, indicate that an index merge if required
      */
     public void setMergeRequired(final boolean mergeRequired) {
@@ -54,7 +55,7 @@ public class IndexDeferredMaintenancePolicy {
 
     /**
      * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
-     * If the user sets it to false, he is responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
+     * If the user sets it to false, they are responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
      * function if {@link #isMergeRequired()} returns true after the requested operation.
      * @param autoMergeDuringCommit if true (default) and applicable, automatically merge during commit
      */
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
@@ -20,33 +20,41 @@
 
 package com.apple.foundationdb.record.provider.foundationdb;
 
+import com.apple.foundationdb.record.metadata.Index;
+
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * Some store's indexes may need merging on some occasions. This helper module should allow the caller
  * to set and probe the merge policy and merge requests.
  */
 public class IndexDeferredMaintenancePolicy {
-    private boolean mergeRequired = false;
+    private Set<Index> mergeRequiredIndexes = null;
     private boolean autoMergeDuringCommit = true;
 
     /**
-     * Indication that a deferred index merge operation is required. This function may be used by the
-     * caller, to check  if the index maintainer requested a deferred merge.
-     * @return if true, an index merge if required
+     * Return a set of indexes that need a deferred index merge operation. This function may be used by the
+     * caller, to check which index maintainer requested a deferred merge.
+     * @return set of indexes to be merged, null if no merge was requested.
      */
-    public boolean isMergeRequired() {
-        return mergeRequired;
+    public Set<Index> getMergeRequiredIndexes() {
+        return mergeRequiredIndexes;
     }
 
     /**
      * Indicate to the caller if a deferred merge operation is required. This function is used by the index maintainer.
-     * @param mergeRequired if true, indicate that an index merge if required
+     * @param mergeRequiredIndex an index that should be merged
      */
-    public void setMergeRequired(final boolean mergeRequired) {
-        this.mergeRequired = mergeRequired;
+    public synchronized void setMergeRequiredIndexes(final Index mergeRequiredIndex) {
+        if (mergeRequiredIndexes == null) {
+            this.mergeRequiredIndexes = new HashSet<>();
+        }
+        this.mergeRequiredIndexes.add(mergeRequiredIndex);
     }
 
     /**
-     * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
+     * If thrue, indicates to the index maintenance to automatically merge indexes during commit.
      * @return true if should merge
      */
     public boolean shouldAutoMergeDuringCommit() {
@@ -56,7 +64,7 @@ public class IndexDeferredMaintenancePolicy {
     /**
      * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
      * If the user sets it to false, they are responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
-     * function if {@link #isMergeRequired()} returns true after the requested operation.
+     * function with the set of indexes returned by {@link #getMergeRequiredIndexes()} as target indexes.
      * @param autoMergeDuringCommit if true (default) and applicable, automatically merge during commit
      */
     public void setAutoMergeDuringCommit(final boolean autoMergeDuringCommit) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
@@ -38,7 +38,7 @@ public class IndexDeferredMaintenancePolicy {
      * caller, to check which index maintainer requested a deferred merge.
      * @return set of indexes to be merged, null if no merge was requested.
      */
-    public Set<Index> getMergeRequiredIndexes() {
+    public synchronized Set<Index> getMergeRequiredIndexes() {
         return mergeRequiredIndexes;
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
@@ -301,4 +301,12 @@ public abstract class IndexMaintainer {
         // Not implemented by default - needs to be overridden by individual maintainers
         throw new UnsupportedRemoteFetchIndexException("scanRemoteFetch operation is not supported by this index maintainer for Index " + state.index.getName());
     }
+
+    /**
+     * If the index is merge'Able, perform an index merge (within a single transaction limit).
+     * @return a future that completes the merge operation
+     */
+    @API(API.Status.EXPERIMENTAL)
+    public abstract CompletableFuture<Void> mergeIndex();
+
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -61,6 +61,7 @@ import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
@@ -105,7 +106,6 @@ public abstract class IndexingBase {
     private boolean forceStampOverwrite = false;
     private final long startingTimeMillis;
     private long lastTypeStampCheckMillis;
-    private List<Index> luceneIndexes;
 
     IndexingBase(@Nonnull IndexingCommon common,
                  @Nonnull OnlineIndexer.IndexingPolicy policy) {
@@ -123,14 +123,6 @@ public abstract class IndexingBase {
         this.throttle = new IndexingThrottle(common, expectedIndexState);
         this.startingTimeMillis = System.currentTimeMillis();
         this.lastTypeStampCheckMillis = startingTimeMillis;
-        for (Index index: common.getTargetIndexes()) {
-            if (index.getType().contains("lucene")) { // Is there a better way to identify Lucene indexes without generating index maintainers?
-                if (luceneIndexes == null) {
-                    luceneIndexes = new ArrayList<>();
-                }
-                luceneIndexes.add(index);
-            }
-        }
     }
 
     // helper functions
@@ -934,9 +926,9 @@ public abstract class IndexingBase {
                     throttle.buildCommitRetryAsync(iterateRange, shouldReturnQuietly, additionalLogMessageKeyValues, true)
                             .handle((hasMore, ex) -> {
                                 if (ex == null) {
-                                    if (throttle.isIndexMergeNeeded()) {
-                                        throttle.resetIndexMergeNeeded();
-                                        return mergeIndexes().thenCompose(ignore -> doneOrThrottleDelayAndMaybeLogProgress(!hasMore, subspaceProvider, additionalLogMessageKeyValues));
+                                    final Set<Index> indexSet = throttle.getAndResetMergeRequiredIndexes();
+                                    if (indexSet != null && !indexSet.isEmpty()) {
+                                        return mergeIndexes(indexSet).thenCompose(ignore -> doneOrThrottleDelayAndMaybeLogProgress(!hasMore, subspaceProvider, additionalLogMessageKeyValues));
                                     }
                                     return doneOrThrottleDelayAndMaybeLogProgress(!hasMore, subspaceProvider, additionalLogMessageKeyValues);
                                 }
@@ -951,7 +943,11 @@ public abstract class IndexingBase {
     }
 
     public CompletableFuture<Void> mergeIndexes() {
-        return AsyncUtil.whenAll(luceneIndexes.stream()
+        return mergeIndexes(new HashSet<>(common.getTargetIndexes()));
+    }
+
+    private CompletableFuture<Void> mergeIndexes(Set<Index> indexSet) {
+        return AsyncUtil.whenAll(indexSet.stream()
                 .map(index2 -> getRunner().runAsync(context ->
                         openRecordStore(context).thenCompose(store ->
                     store.getIndexMaintainer(index2).mergeIndex()
@@ -959,7 +955,7 @@ public abstract class IndexingBase {
     }
 
     private void maybeDeferAutoMergeDuringCommit(FDBRecordStore store) {
-        if (luceneIndexes != null && !luceneIndexes.isEmpty() && policy.shouldDeferLuceneMergeDuringIndexing()) {
+        if (policy.shouldDeferMergeDuringIndexing()) {
             store.getIndexDeferredMaintenancePolicy().setAutoMergeDuringCommit(false);
         }
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -105,6 +105,7 @@ public abstract class IndexingBase {
     private boolean forceStampOverwrite = false;
     private final long startingTimeMillis;
     private long lastTypeStampCheckMillis;
+    private List<Index> luceneIndexes;
 
     IndexingBase(@Nonnull IndexingCommon common,
                  @Nonnull OnlineIndexer.IndexingPolicy policy) {
@@ -122,6 +123,14 @@ public abstract class IndexingBase {
         this.throttle = new IndexingThrottle(common, expectedIndexState);
         this.startingTimeMillis = System.currentTimeMillis();
         this.lastTypeStampCheckMillis = startingTimeMillis;
+        for (Index index: common.getTargetIndexes()) {
+            if (index.getType().contains("lucene")) { // Is there a better way to identify Lucene indexes without generating index maintainers?
+                if (luceneIndexes == null) {
+                    luceneIndexes = new ArrayList<>();
+                }
+                luceneIndexes.add(index);
+            }
+        }
     }
 
     // helper functions
@@ -596,7 +605,10 @@ public abstract class IndexingBase {
     }
 
     // Helpers for implementing modules. Some of them are public to support unit-testing.
-    protected CompletableFuture<Boolean> throttleDelayAndMaybeLogProgress(SubspaceProvider subspaceProvider, List<Object> additionalLogMessageKeyValues) {
+    protected CompletableFuture<Boolean> doneOrThrottleDelayAndMaybeLogProgress(boolean done, SubspaceProvider subspaceProvider, List<Object> additionalLogMessageKeyValues) {
+        if (done) {
+            return AsyncUtil.READY_FALSE;
+        }
         long toWait = throttle.waitTimeMilliseconds();
 
         if (LOGGER.isInfoEnabled() && shouldLogBuildProgress()) {
@@ -721,6 +733,7 @@ public abstract class IndexingBase {
         // Copying the state also guards against changes made by other online building from check version.
         AtomicLong recordsScannedCounter = new AtomicLong();
         final AtomicReference<RecordCursorResult<T>> nextResult = new AtomicReference<>(null);
+        maybeDeferAutoMergeDuringCommit(store);
 
         return validateTypeStamp(store)
                 .thenCompose(ignore ->
@@ -921,11 +934,11 @@ public abstract class IndexingBase {
                     throttle.buildCommitRetryAsync(iterateRange, shouldReturnQuietly, additionalLogMessageKeyValues, true)
                             .handle((hasMore, ex) -> {
                                 if (ex == null) {
-                                    if (Boolean.FALSE.equals(hasMore)) {
-                                        // all done
-                                        return AsyncUtil.READY_FALSE;
+                                    if (throttle.isIndexMergeNeeded()) {
+                                        throttle.resetIndexMergeNeeded();
+                                        return mergeIndexes().thenCompose(ignore -> doneOrThrottleDelayAndMaybeLogProgress(!hasMore, subspaceProvider, additionalLogMessageKeyValues));
                                     }
-                                    return throttleDelayAndMaybeLogProgress(subspaceProvider, additionalLogMessageKeyValues);
+                                    return doneOrThrottleDelayAndMaybeLogProgress(!hasMore, subspaceProvider, additionalLogMessageKeyValues);
                                 }
                                 final RuntimeException unwrappedEx = getRunner().getDatabase().mapAsyncToSyncException(ex);
                                 if (LOGGER.isInfoEnabled()) {
@@ -935,6 +948,20 @@ public abstract class IndexingBase {
                                 throw unwrappedEx;
                             }).thenCompose(Function.identity()),
                 getRunner().getExecutor());
+    }
+
+    public CompletableFuture<Void> mergeIndexes() {
+        return AsyncUtil.whenAll(luceneIndexes.stream()
+                .map(index2 -> getRunner().runAsync(context ->
+                        openRecordStore(context).thenCompose(store ->
+                    store.getIndexMaintainer(index2).mergeIndex()
+                ))).collect(Collectors.toList()));
+    }
+
+    private void maybeDeferAutoMergeDuringCommit(FDBRecordStore store) {
+        if (luceneIndexes != null && !luceneIndexes.isEmpty() && policy.shouldDeferLuceneMergeDuringIndexing()) {
+            store.getIndexDeferredMaintenancePolicy().setAutoMergeDuringCommit(false);
+        }
     }
 
     protected static boolean allRangesExhausted(Tuple cont, Tuple end) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -569,17 +569,18 @@ public class OnlineIndexer implements AutoCloseable {
     }
 
     /**
-     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
-     * {@link IndexDeferredMaintenancePolicy}.
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes
+     * {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and {@link IndexDeferredMaintenancePolicy}.
      * @return a future with the merge index operation
      */
+    @API(API.Status.EXPERIMENTAL)
     public CompletableFuture<Void> mergeIndexAsync() {
         return indexingLauncher(() -> getIndexer().mergeIndexes());
     }
 
     /**
-     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
-     * {@link IndexDeferredMaintenancePolicy}.
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes
+     * {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and {@link IndexDeferredMaintenancePolicy}.
      */
     @API(API.Status.EXPERIMENTAL)
     public void mergeIndex() {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -2132,7 +2132,7 @@ public class OnlineIndexer implements AutoCloseable {
         private final List<Tuple> mutualIndexingBoundaries;
         private final boolean allowUnblock;
         private final String allowUnblockId;
-        private final boolean deferLuceneMergeDuringIndexing;
+        private final boolean deferMergeDuringIndexing;
 
         /**
          * Possible actions when an index is already partially built.
@@ -2163,7 +2163,7 @@ public class OnlineIndexer implements AutoCloseable {
                                DesiredAction ifDisabled, DesiredAction ifWriteOnly, DesiredAction ifMismatchPrevious, DesiredAction ifReadable,
                                boolean allowUniquePendingState, boolean allowTakeoverContinue, long checkIndexingMethodFrequencyMilliseconds,
                                boolean mutualIndexing, List<Tuple> mutualIndexingBoundaries,
-                               boolean allowUnblock, String allowUnblockId, boolean deferLuceneMergeDuringIndexing) {
+                               boolean allowUnblock, String allowUnblockId, boolean deferMergeDuringIndexing) {
             this.sourceIndex = sourceIndex;
             this.forbidRecordScan = forbidRecordScan;
             this.sourceIndexSubspaceKey = sourceIndexSubspaceKey;
@@ -2178,7 +2178,7 @@ public class OnlineIndexer implements AutoCloseable {
             this.mutualIndexingBoundaries = mutualIndexingBoundaries;
             this.allowUnblock = allowUnblock;
             this.allowUnblockId = allowUnblockId;
-            this.deferLuceneMergeDuringIndexing = deferLuceneMergeDuringIndexing;
+            this.deferMergeDuringIndexing = deferMergeDuringIndexing;
         }
 
         /**
@@ -2343,13 +2343,13 @@ public class OnlineIndexer implements AutoCloseable {
         }
 
         /**
-         * If true, attempt to merge Lucene indexes in separate transactions. This feature may be used to avoid repeating
-         * transaction timeout (Lucene only).
-         * @return true if should defer Lucene index merge during indexing.
+         * If true, attempt to merge indexes in separate transactions. This feature may be used to avoid
+         * transaction timeout(s) for indexes that require merge and support deferred merge.
+         * @return true if should defer index merge during the indexing process.
          */
         @API(API.Status.EXPERIMENTAL)
-        public boolean shouldDeferLuceneMergeDuringIndexing() {
-            return this.deferLuceneMergeDuringIndexing;
+        public boolean shouldDeferMergeDuringIndexing() {
+            return this.deferMergeDuringIndexing;
         }
 
         /**
@@ -2381,7 +2381,7 @@ public class OnlineIndexer implements AutoCloseable {
             private List<Tuple> useMutualIndexingBoundaries = null;
             private boolean allowUnblock = false;
             private String allowUnblockId = null;
-            private boolean deferLuceneMergeDuringIndexing = false;
+            private boolean deferMergeDuringIndexing = false;
 
             protected Builder() {
             }
@@ -2630,15 +2630,15 @@ public class OnlineIndexer implements AutoCloseable {
             }
 
             /**
-             *  If set to true, attempt to merge Lucene indexes in separate transactions. This feature may be used to avoid
-             *  repeating transaction timeout, and only effects Lucene indexes.
+             *  If set to true, attempt to merge indexes in separate transactions. This feature may be used to avoid
+             *  transaction timeout(s), and only affects indexes that require merge and support deferred merge.
              * The default value is false.
-             * @param deferLuceneMergeDuringIndexing if true, attempt to merge in separate transactions.
+             * @param deferMergeDuringIndexing if true, attempt to merge indexes in separate transactions.
              * @return this builder
              */
             @API(API.Status.EXPERIMENTAL)
-            public Builder setDeferLuceneMergeDuringIndexing(boolean deferLuceneMergeDuringIndexing) {
-                this.deferLuceneMergeDuringIndexing = deferLuceneMergeDuringIndexing;
+            public Builder setDeferMergeDuringIndexing(boolean deferMergeDuringIndexing) {
+                this.deferMergeDuringIndexing = deferMergeDuringIndexing;
                 return this;
             }
 
@@ -2650,7 +2650,7 @@ public class OnlineIndexer implements AutoCloseable {
                         ifDisabled, ifWriteOnly, ifMismatchPrevious, ifReadable,
                         doAllowUniqueuPendingState, doAllowTakeoverContinue, checkIndexingStampFrequency,
                         useMutualIndexing, useMutualIndexingBoundaries, allowUnblock, allowUnblockId,
-                        deferLuceneMergeDuringIndexing);
+                        deferMergeDuringIndexing);
             }
         }
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -569,6 +569,24 @@ public class OnlineIndexer implements AutoCloseable {
     }
 
     /**
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
+     * {@link IndexDeferredMaintenancePolicy}.
+     * @return a future with the merge index operation
+     */
+    public CompletableFuture<Void> mergeIndexAsync() {
+        return indexingLauncher(() -> getIndexer().mergeIndexes());
+    }
+
+    /**
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
+     * {@link IndexDeferredMaintenancePolicy}.
+     */
+    @API(API.Status.EXPERIMENTAL)
+    public void mergeIndex() {
+        asyncToSync(FDBStoreTimer.Waits.WAIT_ONLINE_MERGE_INDEX, mergeIndexAsync());
+    }
+
+    /**
      * Builds (transactionally) the endpoints of an index. What this means is that builds everything from the beginning of
      * the key space to the first record and everything from the last record to the end of the key space.
      * There won't be any records within these ranges (except for the last record of the record store), but
@@ -2113,6 +2131,7 @@ public class OnlineIndexer implements AutoCloseable {
         private final List<Tuple> mutualIndexingBoundaries;
         private final boolean allowUnblock;
         private final String allowUnblockId;
+        private final boolean deferLuceneMergeDuringIndexing;
 
         /**
          * Possible actions when an index is already partially built.
@@ -2143,7 +2162,7 @@ public class OnlineIndexer implements AutoCloseable {
                                DesiredAction ifDisabled, DesiredAction ifWriteOnly, DesiredAction ifMismatchPrevious, DesiredAction ifReadable,
                                boolean allowUniquePendingState, boolean allowTakeoverContinue, long checkIndexingMethodFrequencyMilliseconds,
                                boolean mutualIndexing, List<Tuple> mutualIndexingBoundaries,
-                               boolean allowUnblock, String allowUnblockId) {
+                               boolean allowUnblock, String allowUnblockId, boolean deferLuceneMergeDuringIndexing) {
             this.sourceIndex = sourceIndex;
             this.forbidRecordScan = forbidRecordScan;
             this.sourceIndexSubspaceKey = sourceIndexSubspaceKey;
@@ -2158,6 +2177,7 @@ public class OnlineIndexer implements AutoCloseable {
             this.mutualIndexingBoundaries = mutualIndexingBoundaries;
             this.allowUnblock = allowUnblock;
             this.allowUnblockId = allowUnblockId;
+            this.deferLuceneMergeDuringIndexing = deferLuceneMergeDuringIndexing;
         }
 
         /**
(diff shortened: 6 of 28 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -159,6 +169,8 @@ public IndexWriter getWriter(LuceneAnalyzerWrapper analyzerWrapper) throws IOExc
                     }
                     writer = new IndexWriter(directory, indexWriterConfig);
                     writerAnalyzerId = analyzerWrapper.getUniqueIdentifier();
+                    // Merge is required when creating an index writer (do we have a better indicator for a required merge?)
```
ScottDugas, line 172: This seems excessive. but if we can't get something better from Lucene itself, it could be any time we do any write or delete, which is when we create the IndexWriter. It's possible that the index maintainer should reuse the `IndexWriter` for more operations.

jjezra: This is what we have today. The overhead would be re-reading the index information (and the means required to do a task in the background, if the record update caller chooses to go this way).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+/**
+ * Some store's indexes may need merging on some occasions. This helper module should allow the caller
+ * to set and prob the merge policy and merge requests.
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class IndexMergePolicy {
```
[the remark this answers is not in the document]
jjezra: As far as I can tell, there is nothing in this class/flow that is thread sensitive.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -937,6 +950,20 @@ protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMes
...
+    public CompletableFuture<Void> mergeIndexes() {
+        return AsyncUtil.whenAll(luceneIndexes.stream()
+                .map(index2 -> getRunner().runAsync(context -> openRecordStore(context).thenApply(store -> {
+                    store.getIndexMaintainer(index2).mergeIndex();
+                    return true;
+                }))).collect(Collectors.toList()));
+    }
+
+    private void preventMergeDuringStoreClose(FDBRecordStore store) {
+        if (luceneIndexes != null && !luceneIndexes.isEmpty() && policy.shouldPreventLuceneMergeDuringIndexing()) {
```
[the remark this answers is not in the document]
jjezra: It seems cleaner this way. Typically, for non-Lucene sessions, the first `luceneIndexes != null` condition will be enough to return.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
```diff
@@ -0,0 +1,46 @@
...
+    private boolean indexNeedsMerge = false;
+    private boolean mergeDuringClose = true;
+
+    public boolean isIndexNeedsMerge() {
```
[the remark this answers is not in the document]
ohadzeliger: With the current implementation, setting the values of the booleans is not thread safe, in the sense that under different threads paths, the results can be inconsistent. It may not be a big deal now, as I guess that we're only ever setting indexNeedsMerge to TRUE and mergeDuringClose to FALSE, but in the future it can be an opening to bugs. 
One way to fix this is to, for example, only allow "flipping" the value and not actually setting it to any boolean you want.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/AtomicMutationIndexMaintainer.java
```diff
@@ -239,4 +239,8 @@ public boolean skipUpdateForUnchangedKeys() {
     // NOTE: It is possible to convert _LONG entries to _TUPLE using something like
     // maintainer1.scan(...).forEach(maintainer2::saveIndexEntryAsKeyValue).
     // This might require multiple scans in transactions if there are too many groups.
```
ScottDugas, line 241: This comment looks out of place, but looks to always have been at the bottom of this file. Perhaps it can be dropped, or kept below your new method

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -107,6 +108,15 @@ private FDBDirectoryMergeScheduler(@Nonnull IndexMaintainerState state, int merg
         })
         @Override
         public synchronized void merge(final MergeSource mergeSource, final MergeTrigger trigger) throws IOException {
+            if (trigger == MergeTrigger.CLOSING) {
+                final IndexMergePolicy indexMergePolicy = state.store.getIndexMergePolicy();
+                if (!indexMergePolicy.isMergeDuringClose()) {
+                    // Here it is done for the record - if this store is being closed it is too late to use it as a messenger
+                    // (See the other code points that set this value)
+                    indexMergePolicy.setIndexNeedsMerge(true);
```
ScottDugas, line 116: Why are we doing this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -937,6 +950,19 @@ protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMes
                 getRunner().getExecutor());
     }
 
+    public CompletableFuture<Void> mergeIndexes() {
+        return AsyncUtil.whenAll(luceneIndexes.stream()
+                .map(index2 -> getRunner().runAsync(context -> openRecordStore(context).thenApply(store ->
+                    store.getIndexMaintainer(index2).mergeIndex().join()
+                ))).collect(Collectors.toList()));
+    }
```
ScottDugas, line 958: This seems like it exposes too many details for other adopters to use. A better way would be to create a new `IndexOperation` to be used here.

Perhaps there's something more opaque we could provide that would serve these needs, but allow the user to avoid having to know what types of maintenance they are deferring (if that's even a good idea).
Something like a an option the `FDBRecordStore`: `allowDeferredIndexMaintenance`, `needsDeferredIndexMaintenance` and `performDeferredIndexMaintenance`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,70 @@ public synchronized void close() throws IOException {
...
+                scanProperties);
+
+        return cursor.forEach(groupingKey -> mergeIndex(analyzerWrapper, groupingKey));
+    }
+
+    private void mergeIndex(LuceneAnalyzerWrapper analyzerWrapper, Tuple groupingKey) {
+        try {
+            getDirectoryWrapper(groupingKey).mergeIndex(analyzerWrapper);
+        } catch (IOException e) {
+            if (LOG.isWarnEnabled()) {
```
ScottDugas, line 113: This should throw a `RecordCoreException`, with `e` as the cause.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,70 @@ public synchronized void close() throws IOException {
...
+                LOG.warn("Lucene mergeIndex failed: error={}", e.getMessage());
```
ScottDugas, line 114: This is not the proper way to log information in the record layer, it should use `KeyValueLogMessage`.
Also, when logging exceptions, the exception should be added as the second parameter to the logging call, so that the stack trace is included.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,70 @@ public synchronized void close() throws IOException {
...
+                        .setRange(range)
+                        .setContinuation(null)
+                        .setScanProperties(scanProperties)
+                        .build()
+                :
+                KeyValueCursor.Builder.withSubspace(subspace)
+                        .setContext(context)
+                        .setContinuation(null)
+                        .setScanProperties(scanProperties)
+                        .setLow(subspace.pack(lastTuple.get().get(0)), EndpointType.RANGE_EXCLUSIVE)
```
ScottDugas, line 139: I don't think this will work correctly when there is more than one grouping key. You'll have to get the size of the grouping key from the root expression of the index, and set that sub-tuple as the low here.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,70 @@ public synchronized void close() throws IOException {
...
+                        .setHigh(range.getHighKey(), range.getHighEndpoint())
+                        .build();
+
+        return cursor.onNext().thenApply(next -> {
+            if (next.hasNext()) {
+                KeyValue kv = next.get();
+                if (kv != null) {
+                    return Optional.of(subspace.unpack(kv.getKey()));
```
ScottDugas, line 147: This doesn't look right, I think this will pick up all of the key after the subspace, not just the grouping key. So you'll be using whatever the first key used in the index is.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,70 @@ public synchronized void close() throws IOException {
         createdDirectories.clear();
     }
 
+    @SuppressWarnings("PMD.CloseResource")
+    public CompletableFuture<Void> mergeIndex(LuceneAnalyzerWrapper analyzerWrapper) {
+
+        final ScanProperties scanProperties = ScanProperties.FORWARD_SCAN.with(
```
ScottDugas, line 93: It's possible that the root expression does not have any grouping keys, in which case you should just provide an empty Tuple for the grouping key.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.NGRAM_LUCENE_INDEX;
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.QUERY_ONLY_SYNONYM_LUCENE_INDEX;
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.SIMPLE_TEXT_SUFFIXES;
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.TEXT_AND_STORED;
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.createSimpleDocument;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.COMPLEX_DOC;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.SIMPLE_DOC;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class LucenOnlineIndexingTest extends FDBRecordStoreTestBase {
```
ScottDugas, line 51: There should be tests of using this feature outside of indexing.
Namely there should be tests of saving and updating records that do the merge outside of the save.
You'll also need to write tests that reach more into the details of lucene, namely doing a few inserts/updates in separate transactions and seeing that the number of segments (i.e. files) grows, and then actually perform the merge operation, and see that the number of segments goes back to 1.

That should run with 0, 1, or 2 grouping keys.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+    private static int oldMaxAttempts;
+
+    @BeforeAll
+    public static void setUpForClass() {
+        oldInitialDelayMillis = FDBDatabaseFactory.instance().getInitialDelayMillis();
+        FDBDatabaseFactory.instance().setInitialDelayMillis(2L);
+        oldMaxDelayMillis = FDBDatabaseFactory.instance().getMaxDelayMillis();
+        FDBDatabaseFactory.instance().setMaxDelayMillis(4L);
+        oldMaxAttempts = FDBDatabaseFactory.instance().getMaxAttempts();
+        FDBDatabaseFactory.instance().setMaxAttempts(100);
```
ScottDugas, line 64: Why are you adjusting all these settings?
It would be good to add a comment

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+    public static void tearDownForClass() {
+        FDBDatabaseFactory.instance().setMaxDelayMillis(oldMaxDelayMillis);
+        FDBDatabaseFactory.instance().setInitialDelayMillis(oldInitialDelayMillis);
+        FDBDatabaseFactory.instance().setMaxAttempts(oldMaxAttempts);
+    }
+
+    @BeforeEach
+    public void setUp() {
+        fdb.run(context -> {
+            FDBRecordStore.deleteStore(context, path);
```
ScottDugas, line 77: Doesn't the base class already do this?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+                    .setIndexingPolicy(OnlineIndexer.IndexingPolicy.newBuilder()
+                            .setPreventLuceneMergeDuringIndexing(preventMergeDuringIndexing))
+                    .build()) {
+                assertTrue(recordStore.isIndexDisabled(index));
+                indexBuilder.buildIndex(true);
+            }
+        }
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            assertTrue(recordStore.isIndexReadable(index));
```
ScottDugas, line 129: It involves accessing internal bits, but it seems like this should assert that there is a reasonable number of segments.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            recordStore.saveRecord(createSimpleDocument(1623L, ENGINEER_JOKE, 2));
+            recordStore.saveRecord(createSimpleDocument(1547L, WAYLON, 1));
+            recordStore.saveRecord(createSimpleDocument(2222L, WAYLON + " who?", 1));
+            context.commit();
+        }
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
```
ScottDugas, line 117: For the number of records in the store, wouldn't it be expected that this will index the entire store in a single transaction? It seems like it would be better to force it to have multiple transactions during indexing.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    @Test
+    void luceneOnlineIndexingTest1() {
+        luceneOnlineIndexingTestAny(QUERY_ONLY_SYNONYM_LUCENE_INDEX, COMPLEX_DOC, 7);
+    }
+
+    @Test
+    void luceneOnlineIndexingTest2() {
+        luceneOnlineIndexingTestAny(QUERY_ONLY_SYNONYM_LUCENE_INDEX, SIMPLE_DOC, 15);
```
ScottDugas, line 140: It seems like there should also be a test with multiple lucene indexes.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+        for (int i = 0; i < numRecords * 2; i ++) {
+            docIds[i] = rn.nextLong();
+        }
+        int middle = 1 + rn.nextInt(numRecords - 2);
+        // write records
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, document, index);
+            for (int i = 0; i < numRecords; i ++) {
+                long docId = docIds[i];
+                recordStore.saveRecord(createSimpleDocument(docId, randomText(rn), randomGroup(rn)));
```
ScottDugas, line 184: This should test with various group counts (0, 1, or 2 columns)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,221 @@
...
+                            .build())
+                    .build()) {
+                assertTrue(recordStore.isIndexDisabled(index));
+                indexBuilder.buildIndex(true);
+            }
+        }
+        // .. and assert readable mode
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, document, index);
+            assertTrue(recordStore.isIndexReadable(index));
```
ScottDugas, line 218: This also seems like it should assert that merge happens.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
jjezra: Just in case the was an explicit "close" call. 
(I don't know if it's possible at all)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: Which would be worse, preventing a record update by repeatedly throwing an exception, or keeping the index unmerged?

jjezra:
test this please

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4784,6 +4785,14 @@ private UninitializedRecordStoreException uninitializedStoreException(@Nonnull S
                 subspaceProvider.logKey(), subspaceProvider.toString(context));
     }
 
+    @Nonnull
+    public synchronized IndexMergePolicy getIndexMergePolicy() {
+        if (indexMergePolicy == null) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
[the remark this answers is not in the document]
jjezra: If we have two threads, one changes a boolean and one reading it, the reader may get the boolean value either before or after the writer changed it - which is fine. 
(Also, this doesn't seem to be very relevant in our IO path)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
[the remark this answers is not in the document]
jjezra: Sure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -122,6 +123,14 @@ public abstract class IndexingBase {
         this.throttle = new IndexingThrottle(common, expectedIndexState);
         this.startingTimeMillis = System.currentTimeMillis();
         this.lastTypeStampCheckMillis = startingTimeMillis;
+        for (Index index: common.getTargetIndexes()) {
+            if (index.getType().contains("lucene")) { // Is there a better way to identify Lucene indexes without generating index maintainers?
```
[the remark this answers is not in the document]
jjezra: We cannot assume that online indexing will be called before the merge does. I'll keep it like this for now and maybe create a lazy evaluator later on.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/AtomicMutationIndexMaintainer.java
jjezra: First I moved it below, then dropped it. 
Perfectionism.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: Can that be a future enhancement? 
The functions `getIndexMergePolicy` and `shouldPreventLuceneMergeDuringIndexing` are marked as experimental, will that allow us future changes?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,475 @@
...
+        try (final FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, document, index);
+            recordStore.markIndexDisabled(index).join();
+            context.commit();
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {false, true})
+    void luceneOnlineIndexingTestSimple(boolean preventMergeDuringIndexing) {
```
ScottDugas, line 115: I'm concerned that all the tests of indexing behave _exactly_ the same if you set `PreventLuceneMergeDuringIndexing` to `true` or `false`; it basically means that you could rip out everything but the setter, and all these tests will still pass.

ScottDugas (changes requested):
Some nitpicks, mostly concerned about increased test coverage of various scenarios

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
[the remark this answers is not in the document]
jjezra: The caller must specify the indexes for any `OnlineIndexer operation`. I wouldn't like to make an exception here. 
Since `mergeIndexes` may be called externally by the API, and since it is a single transaction operation, I wouldn't like to pay the penalty of another read transaction to get a store and index maintainers while creating the `OnlineIndexer` object. 
As first version, identifying Lucene by the index type may be good enough. I just hope it would have been an enum instead of a string.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: Would the code below do the trick? 
```
    cursorBuilder.setLow(subspace.pack(Tuple.fromItems(tuple.getItems().subList(0, groupingCount))), EndpointType.RANGE_EXCLUSIVE)
```

(regardless of creating a low level test)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: It was a copy and paste. Removed it all.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: Ditto.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: Added assertion for `allFiles.length < 10`.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅ 
Added `transactionLimit` param to the `luceneOnlineIndexingTestAny` test.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: I might be missing something - we do not allow merge while indexing in this test.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: Is there any other multi Lucene index test in the code that I can use as an example?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -937,6 +950,19 @@ protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMes
                 getRunner().getExecutor());
     }
 
+    public CompletableFuture<Void> mergeIndexes() {
+        return AsyncUtil.whenAll(luceneIndexes.stream()
+                .map(index2 -> getRunner().runAsync(context -> openRecordStore(context).thenApply(store ->
+                    store.getIndexMaintainer(index2).mergeIndex().join()
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,475 @@
...
+                }
+                commit(context);
+            }
+        }
+        assertTrue(needMerge);
+
+        boolean allDone = false;
+        while (! allDone) {
+            int oldLength;
+            try (FDBRecordContext context = openContext()) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,475 @@
...
+
+    private String[] listFiles(Index index) {
+        try (FDBRecordContext context = openContext()) {
+            final FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context);
+            return directory.listAll();
+        }
+    }
+
+    @Test
+    void testRecordUpdateBgMerge() {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -107,6 +108,16 @@ private FDBDirectoryMergeScheduler(@Nonnull IndexMaintainerState state, int merg
         })
         @Override
         public synchronized void merge(final MergeSource mergeSource, final MergeTrigger trigger) throws IOException {
+            if (trigger == MergeTrigger.FULL_FLUSH || trigger == MergeTrigger.CLOSING) {
+                final IndexMergePolicy indexMergePolicy = state.store.getIndexMergePolicy();
+                if (!indexMergePolicy.isMergeDuringClose()) {
+                    // If this store is being closed, it is too late to use it as a messenger. This flag should also be set
+                    // when the requirements for a "real" merge are met.
```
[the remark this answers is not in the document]
jjezra: Not that I am aware of. But we still wish to skip merge during `MergeTrigger.CLOSING`. Marking the store only during `FULL_FLUSH` will only make the code less readable, I think.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
ScottDugas: Yes, I think that would be the correct code.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ScottDugas: You could try commenting out the code that actually does the merges, and I think it might end up with more than 10 segments, especially if it only indexes one record per transaction.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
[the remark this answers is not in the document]
ScottDugas: 🤦 I misread the code and thought that you were turning off `isMergeDuringClose`. This looks fine to me.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +90,82 @@ public synchronized void close() throws IOException {
...
+                .forEach(groupingKey ->
+                        mergeIndex(analyzerWrapper, groupingKey));
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private static CompletableFuture<Optional<Tuple>> nextTuple(@Nonnull FDBRecordContext context,
+                                                         @Nonnull Subspace subspace,
+                                                         @Nonnull KeyRange range,
+                                                         @Nonnull Optional<Tuple> lastTuple,
+                                                         @Nonnull ScanProperties scanProperties) {
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +90,82 @@ public synchronized void close() throws IOException {
         createdDirectories.clear();
     }
 
+    private void mergeIndex(LuceneAnalyzerWrapper analyzerWrapper, Tuple groupingKey) {
+        try {
+            getDirectoryWrapper(groupingKey).mergeIndex(analyzerWrapper);
+        } catch (IOException e) {
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn(KeyValueLogMessage.build("Lucene mergeIndex failed").toString(), e);
+            }
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,475 @@
...
+                    .build()) {
+                assertTrue(recordStore.isIndexDisabled(index));
+                indexBuilder.buildIndex(true);
+            }
+        }
+        // .. and assert readable mode
+        try (final FDBRecordContext context = openContext()) {
+            openRecordStore(context, hook);
+            assertTrue(recordStore.isIndexReadable(index));
+        }
```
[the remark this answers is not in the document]
jjezra: IIUIC, both the fragmentation and the merge (inlined or not) happens during the online indexing. The only assertion I can think about is that the number of fragments is lower than a certain magic constant. I don't think that it helps much. Do you have any other indication in mind?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: How can we assert that?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ScottDugas: This test calls `.setPreventLuceneMergeDuringIndexing(true)`
IIUC that means:
1. that it won't merge on close during the normal indexing
2. it will, after each transaction call merge, if one is needed (which I think it always will be)

So at the end, it should have done some merges.

I think if you add a check on the number of files, and run it a few times with (2) commented out you will see a very high segment count. If you then run it with that code uncommented, you should see a much lower number.
Repeating both runs a few times should give you an indication of the predictability of the file count, so that you can add an assertion.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ScottDugas: I don't think so. I think this is the first test that does anything that might realistically be impacted by having more than one lucene index.
A simple approach would be to do something similar to what you have in: 
but add all grouping counts at once in a single test: `luceneOnlineIndexingTestGroupingKeys`
```
RecordMetaDataHook hook = metaDataBuilder -> {
    for (int i=0; i<3; i++) {
        metaDataBuilder.addIndex(new Index(
                                         "Map_with_auto_complete$entry-value-" + groupingCount,
                                         new GroupingKeyExpression(field("entry",
                                                                         KeyExpression.FanType.FanOut).nest(concat(LuceneIndexTest.keys)), groupingCount + 1),
                                         LuceneIndexTypes.LUCENE,
                                         ImmutableMap.of()));
    }
    metaDataBuilder.removeIndex(TextIndexTestUtils.SIMPLE_DEFAULT_NAME);
    TextIndexTestUtils.addRecordTypePrefix(metaDataBuilder);
    metaDataBuilder.addIndex(MAP_DOC, index);
};
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅ 

Added `luceneOnlineIndexingTestMulti`.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMergePolicy.java
```diff
@@ -0,0 +1,46 @@
...
+        return indexNeedsMerge;
+    }
+
+    public void setIndexNeedsMerge(final boolean indexNeedsMerge) {
+        this.indexNeedsMerge = indexNeedsMerge;
+    }
+
+    public boolean isMergeDuringClose() {
```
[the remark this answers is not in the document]
jjezra: I will rename this API.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,475 @@
...
+            TextIndexTestUtils.addRecordTypePrefix(metaDataBuilder);
+            metaDataBuilder.addIndex(MAP_DOC, index);
+        };
+        int group = 3;
+
+        // write records
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, hook);
+            for (long i = 1; i < 12; i++) {
+                recordStore.saveRecord(LuceneIndexTest.createMultiEntryMapDoc(77L * i, ENGINEER_JOKE, "sampleTextPhrase", WAYLON, "sampleTextSong", group));
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

jjezra:
```
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[1] STARTED
je: gk:("Text 2, and 0", "Text 3, plus 0", "I am text 4, with 0")
je: gk:("Text 2, and 1", "Text 3, plus 1", "I am text 4, with 1")
je: gk:("Text 2, and 2", "Text 3, plus 2", "I am text 4, with 0")
je: gk:("Text 2, and 3", "Text 3, plus 0", "I am text 4, with 1")
je: gk:("Text 2, and 4", "Text 3, plus 1", "I am text 4, with 0")
je: gk:("Text 2, and 5", "Text 3, plus 2", "I am text 4, with 1")
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[1] PASSED
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[2] STARTED
je: gk:("Text 2, and 0", "Text 3, plus 0")
je: gk:("Text 2, and 1", "Text 3, plus 1")
je: gk:("Text 2, and 2", "Text 3, plus 2")
je: gk:("Text 2, and 3", "Text 3, plus 0")
je: gk:("Text 2, and 4", "Text 3, plus 1")
je: gk:("Text 2, and 5", "Text 3, plus 2")
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[2] PASSED
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[3] STARTED
je: gk:("Text 2, and 0")
je: gk:("Text 2, and 1")
je: gk:("Text 2, and 2")
je: gk:("Text 2, and 3")
je: gk:("Text 2, and 4")
je: gk:("Text 2, and 5")
LucenOnlineIndexingTest > luceneOnlineIndexingTestGroupingKeys(int) > com.apple.foundationdb.record.lucene.LucenOnlineIndexingTest.luceneOnlineIndexingTestGroupingKeys(int)[3] PASSED
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: Making the code pass through both possible paths seems like a good idea. Isn't it?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,604 @@
...
+        try (final FDBRecordContext context = openContext()) {
+            recordStore = LuceneIndexTestUtils.openRecordStore(context, path, hook);
+            for (Index index: indexes) {
+                assertTrue(recordStore.isIndexReadable(index));
+            }
+        }
+        // assert number of segments
+        for (Index index: indexes) {
+            String[] allFiles = listFiles(index);
+            assertTrue(allFiles.length < 12);
```
ohadzeliger, line 310: I may have missed it, but would it make sense to test the logic by creating multiple stores - one with the defer option set to `TRUE` and one with the defer option as `FALSE`, add a few documents to the first, check the number of segments, add more documents to the other and assert that the number of segments is smaller?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -383,6 +383,8 @@ public enum Waits implements Wait {
         WAIT_STOP_ONLINE_INDEX_BUILD("wait for stopping ongoing online index builds"),
         /** Wait for {@link OnlineIndexer} to checking ongoing online index builds. */
         WAIT_CHECK_ONGOING_ONLINE_INDEX_BUILD("wait for checking ongoing online index builds"),
+        /** Wait for {@link OnlineIndexer} to stop ongoing online index merge(s). */
```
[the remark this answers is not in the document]
jjezra: I'll change to "checking"

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -73,6 +87,82 @@ public synchronized void close() throws IOException {
...
+            final byte[] lowKey = subspace.pack(Tuple.fromItems(lastTuple.get().getItems().subList(0, groupingCount)));
+            cursorBuilder
+                    .setLow(lowKey, EndpointType.RANGE_EXCLUSIVE)
+                    .setHigh(range.getHighKey(), range.getHighEndpoint());
+        } else {
+            cursorBuilder.setContext(context)
+                    .setRange(range);
+        }
+
+        return cursorBuilder.build().onNext().thenApply(next -> {
```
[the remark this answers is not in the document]
jjezra: Since the merge is performed by another transaction (probably in a background queue), the store that performed the user operation is long gone. Even if we'll find a way to preserve this info, it could not be trusted.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: That can be done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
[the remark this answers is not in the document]
ScottDugas: I agree that this should be a later optimization (if we do it at all), but I think it can be trusted to the point where it can be passed around. This probably isn't the place to discuss how to add more generic support for running delayed work.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ScottDugas: Yes, my concern here was that the indexer encapsulates the optimization to the point where we can't test whether or not it was applied or not. I don't have a good answer for how to solve that, other than having abstractions in place so that the indexer isn't complicated enough to have risk.
I think for now, this is probably good as it is.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,631 @@
...
+            metaDataBuilder.addIndex(MAP_DOC, index);
+        };
+        int group = 3;
+
+        // write/overwrite records
+        boolean needMerge = false;
+        for (int iLast = 60; iLast > 40; iLast --) {
+            try (FDBRecordContext context = openContext()) {
+                openRecordStore(context, hook);
+                // recordStore.markIndexReadable(index);
```
ScottDugas, line 412:
```suggestion
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,631 @@
...
+        }
+        assertTrue(needMerge);
+
+        Tuple tuple = Tuple.from("Text 2, and 1", "Text 3, plus 1", "I am text 4, with 1");
+        String[] allFiles = listFiles(index, tuple, groupingCount);
+        int oldLength = allFiles.length;
+
+        recordStore.getIndexDeferredMaintenancePolicy().setAutoMergeDuringCommit(false);
+        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                .setRecordStore(recordStore)
```
ScottDugas, line 431: You shouldn't really be using the `FDBRecordStore` outside of the context it is bound to, I think, instead you should call `recordStore.toBuilder()` in the previous block, and then call `setRecordStoreBuilder()` here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Indicate to the caller if a deferred index merge operation is required.
+     * @return if true, an index merge if required
+     */
+    public boolean isMergeRequired() {
+        return mergeRequired;
+    }
+
+    /**
+     * Indicate to the caller if a deferred merge operation is required.
```
ScottDugas, line 40: Perhaps worth noting that IndexMaintainer's will call this

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
```diff
@@ -0,0 +1,64 @@
...
+     * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
+     * @return true if should merge
+     */
+    public boolean shouldAutoMergeDuringCommit() {
+        return autoMergeDuringCommit;
+    }
+
+    /**
+     * Indicate to the index maintenance to automatically merge indexes during commit (if applicable).
+     * If the user sets it to false, he is responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
```
ScottDugas, line 57:
```suggestion
     * If the user sets it to false, they are responsible to call, possibly in the background, the {@link OnlineIndexer#mergeIndex()}
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -568,6 +568,24 @@ public void rebuildIndex(@Nonnull FDBRecordStore store) {
         asyncToSync(FDBStoreTimer.Waits.WAIT_ONLINE_BUILD_INDEX, rebuildIndexAsync(store));
     }
 
+    /**
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
+     * {@link IndexDeferredMaintenancePolicy}.
```
ScottDugas, line 573:
```suggestion
     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes 
     * {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and {@link IndexDeferredMaintenancePolicy}.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -568,6 +568,24 @@ public void rebuildIndex(@Nonnull FDBRecordStore store) {
...
+     * @return a future with the merge index operation
+     */
+    public CompletableFuture<Void> mergeIndexAsync() {
+        return indexingLauncher(() -> getIndexer().mergeIndexes());
+    }
+
+    /**
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
+     * {@link IndexDeferredMaintenancePolicy}.
```
ScottDugas, line 582:
```suggestion
     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes 
     * {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and {@link IndexDeferredMaintenancePolicy}.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -568,6 +568,24 @@ public void rebuildIndex(@Nonnull FDBRecordStore store) {
         asyncToSync(FDBStoreTimer.Waits.WAIT_ONLINE_BUILD_INDEX, rebuildIndexAsync(store));
     }
 
+    /**
+     * If applicable, merge the target indexes as a background maintenance. This experimental feature completes {@link FDBRecordStore#getIndexDeferredMaintenancePolicy()} and
+     * {@link IndexDeferredMaintenancePolicy}.
+     * @return a future with the merge index operation
+     */
+    public CompletableFuture<Void> mergeIndexAsync() {
```
ScottDugas, line 576:
```suggestion
    @API(API.Status.EXPERIMENTAL)
    public CompletableFuture<Void> mergeIndexAsync() {
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,631 @@
...
+
+    @Test
+    void testRecordUpdateReducedMerge() {
+        // emulate repeating merge until until unchanged.
+        Index index = SIMPLE_TEXT_SUFFIXES;
+        boolean needMerge = populateDataSplitSegments(index, 40, 7);
+        assertTrue(needMerge);
+
+        boolean allDone = false;
+        while (! allDone) {
```
ScottDugas, line 613: It seems worthwhile to assert that it goes through this loop at least once, maybe at least twice.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
jjezra: This setup is not needed for `OnlineIndexer` operations. I've used it for my own experiments. Now removed all these "out of the blue setups" from all the tests. 
✅

ScottDugas (approved):
Added @MMcM as a reviewer, because I didn't have any familiarity with how Lucene does merges before reviewing this code, and getting someone with more expertise here would be good.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
```diff
@@ -0,0 +1,65 @@
...
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+/**
+ * Some store's indexes may need merging on some occasions. This helper module should allow the caller
+ * to set and probe the merge policy and merge requests.
+ */
+public class IndexDeferredMaintenancePolicy {
+    private boolean mergeRequired = false;
+    private boolean autoMergeDuringCommit = true;
```
MMcM, line 29: I am skeptical that treating all index merges as equivalent is going to work out. I know it's hard to say when there is only one example. But it would feel more natural to me to have per-index-type settings. Or even per index if a caller chooses. Either with subclassing of the policy or maps.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -105,6 +105,7 @@ public abstract class IndexingBase {
     private boolean forceStampOverwrite = false;
     private final long startingTimeMillis;
     private long lastTypeStampCheckMillis;
+    private List<Index> luceneIndexes;
```
MMcM, line 108: I think a forward reference to an index type from one of the extension subsystems is an unnecessary violation of modularity, even if it doesn't call any actual code there.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -937,6 +950,20 @@ protected CompletableFuture<Void> iterateAllRanges(List<Object> additionalLogMes
                 getRunner().getExecutor());
     }
 
+    public CompletableFuture<Void> mergeIndexes() {
+        return AsyncUtil.whenAll(luceneIndexes.stream()
```
MMcM, line 954: Here we know about Lucene indexes as supporting merge, even though we pretended elsewhere that it's general.

I think a better approach might be to have the indexer capable of doing throttle-able `IndexOperations` and making non-inline Lucene merge one of those.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexMaintainer.java
```diff
@@ -279,6 +279,11 @@ public CompletableFuture<Tuple> evaluateAggregateFunction(@Nonnull IndexAggregat
                 .thenApply(combined -> Tuple.from(combined.asByteArray()));
     }
 
+    @Override
+    public CompletableFuture<Void> mergeIndex() {
```
MMcM, line 283: Since the contract of this is likely to vary from index type to index type and require additional state, I think index operations (which already exist) are probably better suited for doing Lucene merges.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -107,6 +108,16 @@ private FDBDirectoryMergeScheduler(@Nonnull IndexMaintainerState state, int merg
         })
         @Override
         public synchronized void merge(final MergeSource mergeSource, final MergeTrigger trigger) throws IOException {
+            if (trigger == MergeTrigger.FULL_FLUSH || trigger == MergeTrigger.CLOSING) {
+                final IndexDeferredMaintenancePolicy indexDeferredMaintenancePolicy = state.store.getIndexDeferredMaintenancePolicy();
+                if (!indexDeferredMaintenancePolicy.shouldAutoMergeDuringCommit()) {
+                    // If this store is being closed, it is too late to use it as a messenger. This flag should also be set
+                    // when the requirements for a "real" merge are met.
+                    indexDeferredMaintenancePolicy.setMergeRequired(true);
```
MMcM, line 116: Here we need to "queue" a deferred merge operation. I don't know whether that's literally some kind of operation state or just names / flags.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -107,6 +108,16 @@ private FDBDirectoryMergeScheduler(@Nonnull IndexMaintainerState state, int merg
...
+                    skipMerge(mergeSource, trigger, "user request");
```
MMcM, line 117: `user` seems kind of vague here. Maybe "policy" would be clearer?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenancePolicy.java
ScottDugas: At least one other potential use case would be an extension to allow `deleteWhere` on some value indexes that wouldn't otherwise support the query, but instead to the delete asynchronously, and handle the inconsistency under the hood. That does feel as though it has some difference, in that it *can't* be done in the transaction, whereas the lucene merges can.
That being said, it seems like this is trying to accomplish two things:
1. Declaring to the record layer that the caller will take care of this asynchronous work, and provide a mechanism for how to do so
2. A policy as to what should be done asynchronously

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
ScottDugas: Perhaps the item missing here is the ability to determine whether there is any reason to do this at all. There's no way for the indexer to determine which, if any of the indexes would benefit from calling `mergeIndexes`.
That being said, if they are all no-op, I doubt that would have any noticeable impact on any of the indexing performance.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/BitmapValueIndexMaintainer.java
MMcM: We could possibly defer this to another pull request.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexes/AtomicMutationIndexMaintainer.java
```diff
@@ -236,7 +236,8 @@ public boolean skipUpdateForUnchangedKeys() {
         return !IndexTypes.COUNT_UPDATES.equals(state.index.getType());
     }
 
-    // NOTE: It is possible to convert _LONG entries to _TUPLE using something like
-    // maintainer1.scan(...).forEach(maintainer2::saveIndexEntryAsKeyValue).
-    // This might require multiple scans in transactions if there are too many groups.
```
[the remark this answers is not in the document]
jjezra: I'm not sure where to put it, so I'll just keep this comment 'as is' at the end of this module.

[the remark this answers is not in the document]
jjezra: ✅

ScottDugas (approved):
I agree it would be good to transition this to using IndexOperations, and tidy up the api a bit, but we can get the first pass in, and do that in a follow up.

jjezra:
Rebased and forced push to resolve a conflict.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -0,0 +1,638 @@
...
+                        .setValue(text3)
+                        .setSecondValue(text4)
+                        .setThirdValue(text))
+                .build();
+
+    }
+
+    private String[] listFiles(Index index) {
+        try (FDBRecordContext context = openContext()) {
+            final FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, true);
```
ScottDugas, line 469: Given that you aren't creating the index with the option enabled, it probably would make more sense to pass `false` in here, and in the line below, but it shouldn't really matter in terms of correctness or coverage.

ScottDugas (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainer.java
```diff
@@ -301,4 +301,12 @@ public RecordCursor<FDBIndexedRawRecord> scanRemoteFetch(@Nonnull final IndexSca
         // Not implemented by default - needs to be overridden by individual maintainers
         throw new UnsupportedRemoteFetchIndexException("scanRemoteFetch operation is not supported by this index maintainer for Index " + state.index.getName());
     }
+
+    /**
+     * If the index is merge'Able, perform an index merge (within a single transaction limit).
+     * @return a future that completes the merge operation
+     */
+    @API(API.Status.EXPERIMENTAL)
+    public abstract CompletableFuture<Void> mergeIndex();
```
MMcM, line 310: This can probably just be an index operation, thereby admitting that it's very specific to the kind of index, which I think it is.

MMcM (approved):
I am approving this for now, so that we can get some more testing integrated with other recent changes. I would like to see the experimental index merge converted to an index operation. I think some knock-on refactoring may then make related things cleaner. But perhaps all that can also be informed by learnings from the testing at scale.

PR #2298, opened 2023-09-07
Add a special secondary sub-index for Lucene documents per segment…

MMcM:
… by record primary key.

Use this to speed up deletes / updates.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -32,6 +32,7 @@ import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TupleRange;
+import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectoryManager;
 import com.apple.foundationdb.record.lucene.idformat.LuceneIndexKeySerializer;
 import com.apple.foundationdb.record.lucene.idformat.RecordCoreFormatException;
@@ -50,6 +51,7 @@ import com.apple.foundationdb.record.provider.foundationdb.indexes.InvalidIndexE
 import com.apple.foundationdb.record.provider.foundationdb.indexes.StandardIndexMaintainer;
 import com.apple.foundationdb.record.query.QueryToKeyMatcher;
 import com.apple.foundationdb.tuple.Tuple;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.Message;
 import org.apache.lucene.document.BinaryPoint;
 import org.apache.lucene.document.Document;
@@ -62,6 +64,7 @@ import org.apache.lucene.document.NumericDocValuesField;
 import org.apache.lucene.document.SortedDocValuesField;
 import org.apache.lucene.document.StoredField;
 import org.apache.lucene.document.StringField;
+import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.search.Query;
@@ -264,6 +267,19 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
     @SuppressWarnings("PMD.CloseResource")
     private void deleteDocument(Tuple groupingKey, Tuple primaryKey) throws IOException {
         final IndexWriter oldWriter = directoryManager.getIndexWriter(groupingKey, indexAnalyzerSelector.provideIndexAnalyzer(""));
+        final DirectoryReader directoryReader = DirectoryReader.open(oldWriter);
+        @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = directoryManager.getDirectory(groupingKey).getPrimaryKeySegmentIndex();
+        if (segmentIndex != null) {
+            final LucenePrimaryKeySegmentIndex.DocumentIndexEntry documentIndexEntry = segmentIndex.findDocument(directoryReader, primaryKey);
+            if (documentIndexEntry != null) {
+                state.context.ensureActive().clear(documentIndexEntry.entryKey); // TODO: Only if valid?
+                long valid = oldWriter.tryDeleteDocument(documentIndexEntry.indexReader, documentIndexEntry.docId);
+                if (valid > 0) {
+                    state.context.increment(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_PRIMARY_KEY);
+                    return;
+                }
+            }
+        }
         Query query;
         // null format means don't use BinaryPoint for the index primary key
         if (keySerializer.hasFormat()) {
@@ -281,6 +297,7 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
             query = SortedDocValuesField.newSlowExactQuery(PRIMARY_KEY_SEARCH_NAME, new BytesRef(keySerializer.asPackedByteArray(primaryKey)));
         }
         oldWriter.deleteDocuments(query);
+        state.context.increment(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_QUERY);
     }
 
     @Nonnull
@@ -435,6 +452,11 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         LOG.trace("performOperation operation={}", operation);
         return CompletableFuture.completedFuture(new IndexOperationResult() {
         });
+    }
+
+    @VisibleForTesting
+    protected FDBDirectory getDirectory(@Nonnull Tuple groupingKey) {
+        return directoryManager.getDirectory(groupingKey);
     }
 
     /**
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -264,7 +264,7 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         return map;
     }
 
-    @SuppressWarnings("PMD.CloseResource")
+    @SuppressWarnings({"PMD.CloseResource", "java:S2095"})
     private void deleteDocument(Tuple groupingKey, Tuple primaryKey) throws IOException {
         final IndexWriter oldWriter = directoryManager.getIndexWriter(groupingKey, indexAnalyzerSelector.provideIndexAnalyzer(""));
         final DirectoryReader directoryReader = DirectoryReader.open(oldWriter);
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
@@ -72,6 +72,7 @@ public class LuceneIndexOptions {
     public static final String PRIMARY_KEY_SERIALIZATION_FORMAT = "primaryKeySerializationFormat";
     /**
      * Whether a separate B-tree index gives a direct mapping of live documents to segment by record primary key.
+     * Boolean string ({@code true} or {@code false}).
      */
     public static final String PRIMARY_KEY_SEGMENT_INDEX_ENABLED = "primaryKeySegmentIndexEnabled";
 
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
@@ -70,6 +70,10 @@ public class LuceneIndexOptions {
      */
     public static final String TEXT_SYNONYM_SET_NAME_OPTION = "textSynonymSetName";
     public static final String PRIMARY_KEY_SERIALIZATION_FORMAT = "primaryKeySerializationFormat";
+    /**
+     * Whether a separate B-tree index gives a direct mapping of live documents to segment by record primary key.
+     */
+    public static final String PRIMARY_KEY_SEGMENT_INDEX_ENABLED = "primaryKeySegmentIndexEnabled";
 
     private LuceneIndexOptions() {
     }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
@@ -1,0 +1,335 @@
+/*
+ * LucenePrimaryKeySegmentIndex.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2023 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.ScanProperties;
+import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
+import com.apple.foundationdb.record.provider.foundationdb.KeyValueCursor;
+import com.apple.foundationdb.subspace.Subspace;
+import com.apple.foundationdb.tuple.ByteArrayUtil;
+import com.apple.foundationdb.tuple.Tuple;
+import org.apache.lucene.codecs.StoredFieldsReader;
+import org.apache.lucene.codecs.StoredFieldsWriter;
+import org.apache.lucene.index.DirectoryReader;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.FieldInfos;
+import org.apache.lucene.index.FilterDirectoryReader;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.index.MergeState;
+import org.apache.lucene.index.SegmentInfo;
+import org.apache.lucene.index.SegmentInfos;
+import org.apache.lucene.index.SegmentReader;
+import org.apache.lucene.index.StandardDirectoryReader;
+import org.apache.lucene.index.StoredFieldVisitor;
+import org.apache.lucene.util.Accountable;
+import org.apache.lucene.util.Bits;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Maintain a B-tree index of primary key to segment and doc id.
+ * This allows for efficient deleting of a document given that key, such as when doing index maintenance from an update.
+ */
+public class LucenePrimaryKeySegmentIndex {
+    @Nonnull
+    private final FDBDirectory directory;
+    @Nonnull
+    private final Subspace subspace;
+
+    public LucenePrimaryKeySegmentIndex(@Nonnull FDBDirectory directory, @Nonnull Subspace subspace) {
+        this.directory = directory;
+        this.subspace = subspace;
+    }
+
+    /**
+     * Get all stored primary key index entries.
+     * Really only useful for small-scale debugging.
+     * @return a list of Tuple-decoded key entries
+     */
+    public List<List<Object>> readAllEntries() {
+        List<Tuple> tuples;
+        try (KeyValueCursor kvs = KeyValueCursor.Builder.newBuilder(subspace)
+                .setContext(directory.getContext())
+                .setScanProperties(ScanProperties.FORWARD_SCAN)
+                .build();
+                RecordCursor<Tuple> entries = kvs.map(kv -> subspace.unpack(kv.getKey()))) {
+            tuples = directory.getContext().asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FIND_PRIMARY_KEY, entries.asList());
+        }
+        return tuples.stream().map(t -> {
+            List<Object> items = t.getItems();
+            // Replace segment id with segment name to make inspection easier.
+            String name = directory.primaryKeySegmentName((Long)items.get(items.size() - 2));
+            if (name != null) {
+                items.set(items.size() - 2, name);
+            }
+            items.set(items.size() - 1, ((Long)items.get(items.size() - 1)).intValue());
+            return items;
+        }).collect(Collectors.toList());
+    }
+
+    /**
+     * Result of {@link #findDocument}.
+     */
+    // TODO: Can be a record.
+    public static class DocumentIndexEntry {
+        @Nonnull
+        public final Tuple primaryKey;
+        @Nonnull
+        public final byte[] entryKey;
+        @Nonnull
+        public final IndexReader indexReader;
+        public final int docId;
+
+        public DocumentIndexEntry(@Nonnull final Tuple primaryKey, @Nonnull final byte[] entryKey, @Nonnull final IndexReader indexReader, final int docId) {
+            this.primaryKey = primaryKey;
+            this.entryKey = entryKey;
+            this.indexReader = indexReader;
+            this.docId = docId;
+        }
+    }
+
+    /**
+     * Find document in index for direct delete.
+     * @param directoryReader a NRT reader
+     * @param primaryKey the document's record's primary key
+     * @return an entry with the leaf reader and document id in that segment or {@code null} if not found
+     * @see IndexWriter#tryDeleteDocument
+     */
+    @Nullable
+    public DocumentIndexEntry findDocument(@Nonnull DirectoryReader directoryReader, @Nonnull Tuple primaryKey) {
+        final SegmentInfos segmentInfos = ((StandardDirectoryReader)FilterDirectoryReader.unwrap(directoryReader)).getSegmentInfos();
+        final Subspace keySubspace = subspace.subspace(primaryKey);
+        try (KeyValueCursor kvs = KeyValueCursor.Builder.newBuilder(keySubspace)
+                .setContext(directory.getContext())
+                .setScanProperties(ScanProperties.FORWARD_SCAN)
+                .build();
+                RecordCursor<DocumentIndexEntry> documents = kvs.map(kv -> {
+                    final Tuple segdoc = keySubspace.unpack(kv.getKey());
+                    final long segid = segdoc.getLong(0);
+                    final String segmentName = directory.primaryKeySegmentName(segid);
+                    if (segmentName == null) {
+                        return null;
+                    }
+                    for (int i = 0; i < segmentInfos.size(); i++) {
+                        SegmentInfo segmentInfo = segmentInfos.info(i).info;
+                        if (segmentInfo.name.equals(segmentName)) {
+                            final int docid = (int)segdoc.getLong(1);
+                            return new DocumentIndexEntry(primaryKey, kv.getKey(),
+                                    directoryReader.leaves().get(i).reader(), docid);
+                        }
+                    }
+                    return null;
+                }).filter(Objects::nonNull)) {
+            return directory.getContext().asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FIND_PRIMARY_KEY,
+                    documents.first()).orElse(null);
+        }
+    }
+
+    /**
+     * Hook for getting back segment info during merge.
+     */
+    public interface StoredFieldsReaderSegmentInfo {
+        SegmentInfo getSegmentInfo();
+    }
+
+    /**
+     * Hook the fields writer to also record primary keys in index.
+     * @param storedFieldsWriter normal field writer
+     * @param si segment info for current writer
+     * @return a wrapped writer
+     */
+    @Nonnull
+    public StoredFieldsWriter wrapFieldsWriter(@Nonnull StoredFieldsWriter storedFieldsWriter, @Nonnull SegmentInfo si) throws IOException {
+        final long segmentId = directory.primaryKeySegmentId(si.name, true);
+        return new WrappedFieldsWriter(storedFieldsWriter, segmentId);
+    }
+
+    class WrappedFieldsWriter extends StoredFieldsWriter {
+        @Nonnull
+        private final StoredFieldsWriter inner;
+        @Nonnull
+        private final long segmentId;
+
+        private int documentId;
+
+        WrappedFieldsWriter(@Nonnull StoredFieldsWriter inner, long segmentId) {
+            this.inner = inner;
+            this.segmentId = segmentId;
+        }
+
+        @Override
+        public void startDocument() throws IOException {
+            inner.startDocument();
+        }
+
+        @Override
+        public void finishDocument() throws IOException {
+            inner.finishDocument();
+            documentId++;
+        }
+
+        @Override
+        public void writeField(FieldInfo info, IndexableField field) throws IOException {
+            inner.writeField(info, field);
+            if (info.name.equals(LuceneIndexMaintainer.PRIMARY_KEY_FIELD_NAME)) {
+                final byte[] primaryKey = field.binaryValue().bytes;
+                addOrDeletePrimaryKeyEntry(primaryKey, segmentId, documentId, true);
+            }
+        }
+
+        @Override
+        @SuppressWarnings("PMD.CloseResource")
+        public int merge(MergeState mergeState) throws IOException {
+            final int docCount = inner.merge(mergeState);
+
+            final int segmentCount = mergeState.storedFieldsReaders.length;
+            final PrimaryKeyVisitor visitor = new PrimaryKeyVisitor();
+            for (int i = 0; i < segmentCount; i++) {
+                final StoredFieldsReader storedFieldsReader = mergeState.storedFieldsReaders[i];
+                final SegmentInfo mergedSegmentInfo = ((StoredFieldsReaderSegmentInfo)storedFieldsReader).getSegmentInfo();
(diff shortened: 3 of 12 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+     * The primary keys are gotten from the stored fields, so this must complete and not yet have been removed.
+     * @param reader a reader for the target segment
+     * @param add {@code true} to add entries; {@code false} to delete them
+     * @throws IOException if the stored fields gets an error
+     */
+    @SuppressWarnings("PMD.CloseResource")
+    public void addOrDeleteDocumentPrimaryKeys(@Nonnull SegmentReader reader, boolean add) throws IOException {
+        final int maxDoc = reader.maxDoc();
+        final PrimaryKeyVisitor visitor = new PrimaryKeyVisitor();
+        final long segmentId = directory.primaryKeySegmentId(reader.getSegmentName(), add);
```
ScottDugas, line 278: Instead of doing this can you get the segmentId from `reader.getSegmentInfo`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -61,6 +61,11 @@ public final class LuceneRecordContextProperties {
      */
     public static final RecordLayerPropertyKey<Double> LUCENE_MERGE_MAX_SIZE = RecordLayerPropertyKey.doublePropertyKey("com.apple.foundationdb.record.lucene.mergeMaxSize", 5.0);
 
+    /**
+     * Count of segments after which to merge for ordinary full-text search with Lucene.
```
ScottDugas, line 65: Is limiting this a requirement of the new index? Or is this just needed because merging has additional cost, and we could more easily run into transaction limits?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -152,25 +154,24 @@ public class FDBDirectory extends Directory  {
 
     private final Map<BitSet, byte[]> fieldInfosDataMap;
 
+    private final boolean primaryKeySegmentIndexEnabled;
+    @Nullable
+    private LucenePrimaryKeySegmentIndex primaryKeySegmentIndex;
+
     public FDBDirectory(@Nonnull Subspace subspace, @Nonnull FDBRecordContext context) {
```
ScottDugas, line 161: It looks like this only has a few test-usages, perhaps it would make sense to add `primaryKeySegmentIndexEnabled` as an argument to this constructor

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -596,25 +597,36 @@ public void deleteFile(@Nonnull String name) throws IOException {
...
+            boolean deleted = Objects.requireNonNull(context.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_DELETE_FILE, getFileReferenceCacheAsync()
+                    .thenApply(cache -> deleteFileInternal(cache, name))));
 
             if (!deleted) {
                 throw new NoSuchFileException(name);
             }
+
+            if (isCompoundFile(name)) {
+                Map<String, FDBLuceneFileReference> cache = fileReferenceCache.get();
+                String primaryKeyName = name.substring(0, name.length() - DATA_EXTENSION.length()) + "pky";
```
ScottDugas, line 609: Would it be better to use some of the methods from `IndexFileNames`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -596,25 +597,36 @@ public void deleteFile(@Nonnull String name) throws IOException {
...
+                deleteFileInternal(cache, primaryKeyName);
+                // TODO: If the segment is being deleted because it no longer has any live docs, it won't be merged
+                //  so we might need to delete the primary key entries another way for case where tryDeleteDocument
+                //  wasn't used.
```
ScottDugas, line 613: The trick that I used in https://github.com/FoundationDB/fdb-record-layer/pull/2300/files/f55d9a8ba7c7f449b30463ea5857ea08c47bd096..63fd482f1af4b11c401ed05c92b9c8c80605603b (which has yet to be reviewed, so maybe someone will point out it's a _bad_ idea)
Is to have files that are actually tracked in the filesystem, but have custom delete logic.
This would basically mean that `LucenePrimaryKeySegmentIndex.WrappedFieldsWriter` would open an additional `*.pky` for the segment when writing, etc. Lucene would then now that the file is part of the segment, and when it goes to delete it, it would send the delete to the `FDBDirectory`. This could then call back into `LucenePrimaryKeySegmentIndex` to delete any entries for that segment. That being said, I think you would have to ensure that the `*.pky` is delete _after_ the stored fields data, so that you can efficiently find the primary keys that you have to delete.
We could, instead, have the `LucenePrimaryKeySegmentIndex` baked into our `StoredFieldsFormat` and then when you delete the data for `StoredFields` we will delete the associated `*.pky` data.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -596,25 +597,36 @@ public void deleteFile(@Nonnull String name) throws IOException {
...
         }
     }
 
+    private boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference> cache, @Nonnull String name) {
+        FDBLuceneFileReference value = cache.remove(name);
+        if (value == null) {
+            return false;
+        }
+        context.ensureActive().clear(metaSubspace.pack(name));
+        context.ensureActive().clear(dataSubspace.subspace(Tuple.from(value.getId())).range());
```
ScottDugas, line 626: Are these valid? Do we store anything for `*.pky` in the actual `dataSubspace`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -838,4 +850,50 @@ public ChecksumIndexInput openChecksumInput(final String name, final IOContext c
...
+    /**
+     * Get a primary key segment index if enabled.
+     * @return index or {@code null} if not enabled
+     */
+    @Nullable
+    public LucenePrimaryKeySegmentIndex getPrimaryKeySegmentIndex() {
+        if (!primaryKeySegmentIndexEnabled) {
+            return null;
+        }
+        if (primaryKeySegmentIndex == null) {
```
ScottDugas, line 863: Does this need to be `synchronized`? I'm fairly sure that merging uses a bunch of threads.
`LucenePrimaryKeySegmentIndex` doesn't have any mutable state, so it's probably fine to have multiple different instances exist. Probably worth commenting otherwise.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -838,4 +850,50 @@ public ChecksumIndexInput openChecksumInput(final String name, final IOContext c
...
+            primaryKeySegmentIndex = new LucenePrimaryKeySegmentIndex(this, primaryKeySubspace);
+        }
+        return primaryKeySegmentIndex;
+    }
+
+    // Map segment name to integer id.
+    // TODO: Could store this elsewhere, such as inside compound file.
+    public long primaryKeySegmentId(@Nonnull String segmentName, boolean create) throws IOException {
+        final String fileName = IndexFileNames.segmentFileName(segmentName, "", "pky");
+        FDBLuceneFileReference ref = getFDBLuceneFileReference(fileName);
```
ScottDugas, line 874: So the `.pky` file only exists as a means to map the segmentName to a unique number using the existing `getIncrement`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -144,7 +146,8 @@ public IndexWriter getWriter(LuceneAnalyzerWrapper analyzerWrapper) throws IOExc
             synchronized (this) {
                 if (writer == null || !writerAnalyzerId.equals(analyzerWrapper.getUniqueIdentifier())) {
                     TieredMergePolicy tieredMergePolicy = new TieredMergePolicy()
-                            .setMaxMergedSegmentMB(Math.max(0.0, state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_MERGE_MAX_SIZE)));
+                            .setMaxMergedSegmentMB(state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_MERGE_MAX_SIZE))
```
ScottDugas, line 149: Why drop the `Math.max` part?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -61,6 +61,11 @@ public final class LuceneRecordContextProperties {
...
+     */
+    public static final RecordLayerPropertyKey<Double> LUCENE_MERGE_SEGMENTS_PER_TIER = RecordLayerPropertyKey.doublePropertyKey("com.apple.foundationdb.record.lucene.mergeSegmentsPerTier", 10.0);
```
ScottDugas, line 67: Should this be an `Integer` property?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -144,7 +146,8 @@ public IndexWriter getWriter(LuceneAnalyzerWrapper analyzerWrapper) throws IOExc
...
+                            .setSegmentsPerTier(state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_MERGE_SEGMENTS_PER_TIER));
```
ScottDugas, line 150: Out of curiosity, do you know why this is a double? Based no the description, I don't see how `5.323` would make sense.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
ScottDugas: I suspect that this is just to facilitate testing, and have better control over how many merges happen.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1167,6 +1177,63 @@ void testDataLoad() {
...
+        }
+
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            assertIndexEntryPrimaryKeys(List.of(1002L), // 18
+                    recordStore.scanIndex(index, fullTextSearch(index, "three"), null, ScanProperties.FORWARD_SCAN));
+            assertIndexEntryPrimaryKeys(List.of(1000L, 1004L),  // 16,20
+                    recordStore.scanIndex(index, fullTextSearch(index, "four"), null, ScanProperties.FORWARD_SCAN));
+            assertIndexEntryPrimaryKeys(List.of(),
+                    recordStore.scanIndex(index, fullTextSearch(index, "seven"), null, ScanProperties.FORWARD_SCAN));
```
ScottDugas, line 1218: It seems like it would also be good to do some deletes and test that, but as you noted in a TODO, if a segment runs out of live documents, this code doesn't currently clean them up.

ScottDugas: Also, it would be good to have a test that deletes and recreates a new record with the same primary key.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ScottDugas: I was talking with @ohadzeliger offline about this. When the documents are deleted we immediately delete the entries from the index, so the issue here would only be that the `*.pky` file reference is left around, and that needs to get cleaned up, right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+            items.set(items.size() - 1, ((Long)items.get(items.size() - 1)).intValue());
+            return items;
+        }).collect(Collectors.toList());
+    }
+
+    /**
+     * Result of {@link #findDocument}.
+     */
+    // TODO: Can be a record.
```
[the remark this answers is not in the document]
MMcM: Yes, that's what was meant. We are still on `JavaVersion.VERSION_11` for both `sourceCompatibility` and `targetCompatibility`.

I suppose we should have a more explicit consensus around the use of these when the time comes. I kind of feel like we missed doing that with `var`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -264,6 +267,19 @@ private Map<IndexOptions, List<LuceneDocumentFromRecord.DocumentField>> getIndex
     @SuppressWarnings("PMD.CloseResource")
     private void deleteDocument(Tuple groupingKey, Tuple primaryKey) throws IOException {
         final IndexWriter oldWriter = directoryManager.getIndexWriter(groupingKey, indexAnalyzerSelector.provideIndexAnalyzer(""));
+        final DirectoryReader directoryReader = DirectoryReader.open(oldWriter);
+        @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = directoryManager.getDirectory(groupingKey).getPrimaryKeySegmentIndex();
+        if (segmentIndex != null) {
+            final LucenePrimaryKeySegmentIndex.DocumentIndexEntry documentIndexEntry = segmentIndex.findDocument(directoryReader, primaryKey);
+            if (documentIndexEntry != null) {
+                state.context.ensureActive().clear(documentIndexEntry.entryKey); // TODO: Only if valid?
```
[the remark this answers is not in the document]
MMcM: In practice, I think so, yes. Specifically, I believe that the NRT reader meets the criteria of being up-to-date with the writer. And I do not think it is possible for the document id to have been merged away (because that should replace the index entry with one for the new segment).

If we didn't try to fall back to the slow query, but gave an error, for instance, then this comment wouldn't apply. But if we do, there is a question of whether the existing index entry should be considered invalidated by that slow query delete.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+        public void finishDocument() throws IOException {
+            inner.finishDocument();
+            documentId++;
+        }
+
+        @Override
+        public void writeField(FieldInfo info, IndexableField field) throws IOException {
+            inner.writeField(info, field);
+            if (info.name.equals(LuceneIndexMaintainer.PRIMARY_KEY_FIELD_NAME)) {
+                final byte[] primaryKey = field.binaryValue().bytes;
```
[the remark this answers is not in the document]
MMcM: `PRIMARY_KEY_FIELD_NAME` (`_p`) is always `keySerializer.asPackedByteArray(primaryKey)`, which is just `primaryKey.pack()`. There is always a key serializer, even if it doesn't have a format.
`PRIMARY_KEY_BINARY_POINT_NAME` (`_b`) is `keySerializer.asFormattedBinaryPoint(primaryKey)` which depends on enabling that.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+                    final byte[] primaryKey = visitor.getPrimaryKey();
+                    if (primaryKey != null) {
+                        if (liveDocs == null || liveDocs.get(j)) {
+                            int docId = docMap.get(j);
+                            if (docId >= 0) {
+                                addOrDeletePrimaryKeyEntry(primaryKey, segmentId, docId, true);
+                            }
+                        }
+                        // TODO: Is doing this right away safe enough? Does transaction isolation protect against all cases where
+                        //  the source segment is used after a merge?
```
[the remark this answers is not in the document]
MMcM: Yes, that is the sort of thing I was worried about.

We do some kind of filtering, in that we only return primary key segment index entries when the segment is in the NRT reader. Which I _think_ means it won't include such deleted segments, but I admit there might be a case I am overlooking.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+                        addOrDeletePrimaryKeyEntry(primaryKey, mergedSegmentId, j, false);
+                        visitor.reset();
+                    }
+                }
+            }
+
+            directory.getContext().increment(LuceneEvents.Counts.LUCENE_MERGE_DOCUMENTS, docCount);
+            directory.getContext().increment(LuceneEvents.Counts.LUCENE_MERGE_SEGMENTS, segmentCount);
```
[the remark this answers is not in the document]
MMcM: Do you think it's worth another layer of wrapping just to instrument merges, even if the primary key index isn't enabled? I agree that this isn't an obvious place to do it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+    /**
+     * Reconcile primary key index entries for a segment.
+     * Delete those for a segment that is about to be deleted or add for a segment that has been filled without going through the wrapper.
+     * The primary keys are gotten from the stored fields, so this must complete and not yet have been removed.
+     * @param reader a reader for the target segment
+     * @param add {@code true} to add entries; {@code false} to delete them
+     * @throws IOException if the stored fields gets an error
+     */
+    @SuppressWarnings("PMD.CloseResource")
+    public void addOrDeleteDocumentPrimaryKeys(@Nonnull SegmentReader reader, boolean add) throws IOException {
```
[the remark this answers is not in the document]
MMcM: As of right now, yes. I wanted to have it in case we thought of additional cases that needed to do catch- / clean-up outside the inline ones it does now. Unfortunately, it isn't easy to get a `SegmentReader` in the appropriate places that I can work out.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
MMcM: I added it for testing, though I could imagine that the work going on in parallel to do merges out-of-band could want to control this. It is not necessary with the new index and defaults to what the Lucene default would be.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: I'm not sure that ends up being more clear / stable. It's sort of modeled after the existing `convertToDataFile`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: Yes, currently. Though we might want to use the lifetime of the `.pky` file reference to trigger some additional cleanup if we determine that there are missing cases with the inline one.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: We do not store anything there now; should we just skip this clear?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: Correct. Would it be better to have a separate registry or to do that mapping separate from the file references?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
MMcM: It seems like it would be better to get an error if the property were set to a negative number than to silently use zero. Did I overlook a legit case where the `max` is needed?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
MMcM: No, I do not know why. Other counts seem to be like that, too.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
MMcM: I wondered about that as well. Is it better to be consistent with the Lucene API or to have the configuration API make a little more sense?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1167,6 +1177,63 @@ void testDataLoad() {
...
+                context.commit();
+            }
+        }
+
+        if (primaryKeySegmentIndexEnabled) {
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_MERGE_SEGMENTS), greaterThan(10));
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_QUERY), equalTo(0));
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_PRIMARY_KEY), greaterThan(10));
+        } else {
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_QUERY), greaterThan(10));
```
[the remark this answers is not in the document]
MMcM: My thinking was that it didn't matter in this case, because none of the custom merge logic would apply. Would it be better to be consistent?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -264,6 +267,19 @@ private Map<IndexOptions, List<LuceneDocumentFromRecord.DocumentField>> getIndex
     @SuppressWarnings("PMD.CloseResource")
     private void deleteDocument(Tuple groupingKey, Tuple primaryKey) throws IOException {
         final IndexWriter oldWriter = directoryManager.getIndexWriter(groupingKey, indexAnalyzerSelector.provideIndexAnalyzer(""));
+        final DirectoryReader directoryReader = DirectoryReader.open(oldWriter);
+        @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = directoryManager.getDirectory(groupingKey).getPrimaryKeySegmentIndex();
+        if (segmentIndex != null) {
+            final LucenePrimaryKeySegmentIndex.DocumentIndexEntry documentIndexEntry = segmentIndex.findDocument(directoryReader, primaryKey);
+            if (documentIndexEntry != null) {
```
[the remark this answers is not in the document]
MMcM: If the document to be deleted's segment is in the writer's `runningMerges`, then that segment will not appear in the NRT reader from that writer. (Also the `.cfs` file won't exist yet; the stored fields will still be in`.f*` files. Although that doesn't matter a lot.) So `segmentIndex.findDocument` will find the index entry not it will look like an orphan and `documentIndexEntry` will be `null`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
```diff
@@ -70,6 +70,10 @@ public class LuceneIndexOptions {
      */
     public static final String TEXT_SYNONYM_SET_NAME_OPTION = "textSynonymSetName";
     public static final String PRIMARY_KEY_SERIALIZATION_FORMAT = "primaryKeySerializationFormat";
+    /**
+     * Whether a separate B-tree index gives a direct mapping of live documents to segment by record primary key.
```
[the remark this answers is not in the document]
MMcM: Yes. Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,335 @@
...
+    private final Subspace subspace;
+
+    public LucenePrimaryKeySegmentIndex(@Nonnull FDBDirectory directory, @Nonnull Subspace subspace) {
+        this.directory = directory;
+        this.subspace = subspace;
+    }
+
+    /**
+     * Get all stored primary key index entries.
+     * Really only useful for small-scale debugging.
```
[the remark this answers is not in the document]
MMcM: Sure. Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: Added. I didn't change any of those callers to pass anything other than `false` though.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
MMcM: I added a `synchronized` around this bit, just be safe. There weren't any already in the class, which maybe indicates other problems._

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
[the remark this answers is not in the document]
ScottDugas: I don't know that it's worth wrapping, but there might be a more appropriate Format to track this as we make more custom writers.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
ScottDugas: I thought Lucene had a number associated with segments, but I can't find it now. It has an `id`, but that's based off a random `BigInteger`, so not really helpful.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
MMcM: Added a delete test.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1167,6 +1177,63 @@ void testDataLoad() {
...
+    void testSimpleUpdate(boolean primaryKeySegmentIndexEnabled) {
+        final Index index = primaryKeySegmentIndexEnabled ? SIMPLE_TEXT_SUFFIXES_WITH_PRIMARY_KEY_SEGMENT_INDEX : SIMPLE_TEXT_SUFFIXES;
+        final RecordLayerPropertyStorage contextProps = RecordLayerPropertyStorage.newBuilder()
+                .addProp(LuceneRecordContextProperties.LUCENE_MERGE_SEGMENTS_PER_TIER, 3.0)
+                .build();
+        final String[] nums = { "zero", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine" };
+        for (int i = 0; i < 20; i++) {
+            final int ii = i + 1;
+            try (FDBRecordContext context = openContext(contextProps)) {
+                rebuildIndexMetaData(context, SIMPLE_DOC, index);
```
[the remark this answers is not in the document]
MMcM: Added a complex primary key test.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ScottDugas: I don't think there would be a big win to have a second registry. I mostly just wanted to confirm that I wasn't missing something.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
ScottDugas: I think having it error rather than silently replacing with 0 is the right way to go

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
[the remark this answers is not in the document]
ScottDugas: I think the consistency would be nice.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1167,6 +1186,151 @@ void testDataLoad() {
...
+            assertIndexEntryPrimaryKeys(Set.of(1623L, 1624L),
+                    recordStore.scanIndex(index, fullTextSearch(index, "Vision"), null, ScanProperties.FORWARD_SCAN));
+            assertTrue(recordStore.deleteRecord(Tuple.from(1624L)));
+            assertIndexEntryPrimaryKeys(Set.of(1623L),
+                    recordStore.scanIndex(index, fullTextSearch(index, "Vision"), null, ScanProperties.FORWARD_SCAN));
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_QUERY), equalTo(0));
+            assertThat(timer.getCount(LuceneEvents.Counts.LUCENE_DELETE_DOCUMENT_BY_PRIMARY_KEY), equalTo(1));
+        }
+        timer.reset();
+        try (FDBRecordContext context = openContext()) {
```
ScottDugas, line 1268: Given some of the concerns above, it seems like it would be good to have a test where you delete all the documents, and then commit, and re-open and check that PrimaryKeySegmentIndex is empty

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -2436,7 +2600,7 @@ void manySegmentsParallelOpen() {
             rebuildIndexMetaData(context, SIMPLE_DOC, SIMPLE_TEXT_SUFFIXES);
             assertEquals(20,
                     recordStore.scanIndex(SIMPLE_TEXT_SUFFIXES, fullTextSearch(SIMPLE_TEXT_SUFFIXES, "Vision"), null, ScanProperties.FORWARD_SCAN).getCount().join());
-            try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(SIMPLE_TEXT_SUFFIXES), context)) {
+            try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(SIMPLE_TEXT_SUFFIXES), context, true)) {
```
ScottDugas, line 2603: Why is this with `true`? Should this be parameterized to run with `true` and `false`?

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -0,0 +1,337 @@
...
+                .setContext(directory.getContext())
+                .setScanProperties(ScanProperties.FORWARD_SCAN)
+                .build();
+                RecordCursor<DocumentIndexEntry> documents = kvs.map(kv -> {
+                    final Tuple segdoc = keySubspace.unpack(kv.getKey());
+                    final long segid = segdoc.getLong(0);
+                    final String segmentName = directory.primaryKeySegmentName(segid);
+                    if (segmentName == null) {
+                        return null;
+                    }
```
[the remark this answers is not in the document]
MMcM: Yes. This is not a stable set of the index, that is, one that can be arrived at by committing and not opening read-only. But, right now, it is only at commit time that we guarantee that a merge is not pending. The recently committed scale test will, during the early stages of running, have few enough segments that its multiple-update pass can sometimes update several in the same segment within the same transaction, which (sometimes) does this.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
MMcM: Added.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
MMcM: I made it parameterized.

ScottDugas (approved):

PR #2299, opened 2023-09-08
Detach CascadesPlanner from QueryPlanner interface

arnaud-lacurie:
CascadesPlanner's recommend usage is through `planGraph`. `planQuery` and `plan` were supported for test purpose but make the interface confusing


normen662 (changes requested):
I wasn't thinking about the effect this would have on the dual planner test cases. You are no always creating a `RecordQueryPlanner`. The code in the test cases switches based on `instanceof RecordQueryPlanner` which now makes these test cases to run with the old planner twice rather than using cascading vs. heuristics. I think this item needs to be shelved until we have a test bed that can test the cascades planner independently of the `planQuery()` interface.

PR #2302, opened 2023-09-14
Make `FieldCopier.copy` thread-safe

hatyo:
`FieldCopier.copy` is not thread-safe. This could cause problems in situations where we have multiple threads trying to reuse a physical plan but with slight differences in their record descriptors.

**Setup**
-: We have two threads, `T1` and `T2` and a physical plan that comprises an operator that leverages `IndexKeyValueToPartialRecord` such as `RecordQueryCoveringIndexPlan`.
- `T1` is operating on a version of record store metadata that contains the following record:
```
message R1 {
  optional int32 x = 1;
}
```
- `T2` is operating on another record store metadata that contains an _evolved_ version the same record:
```
message R1 {
  optional int32 x = 1;
  optional int32 y = 2;
}
```
Let's call this message `R1'` for clarity.
Here is a schedule that shows depicts the race-condition (scroll to the write to see the full schedule):
```java
@Override  
public synchronized boolean copy(...) {                                                              
    final Tuple tuple = (source == TupleSource.KEY ? kv.getKey() : kv.getValue());  
    if (!copyIfPredicate.test(tuple)) {  
        return false;  
    }  
  
    Object value = getForOrdinalPath(tuple, ordinalPath);  
    if (value == null) {  
        return true;  
    }  
    if (!containingType.equals(recordDescriptor)) {  
        containingType = recordDescriptor;  
        fieldDescriptor = recordDescriptor.findFieldByName(field);                                      [1] fieldDescriptor=R1.X                                      
                                                                                                                                                                    [2] fieldDescriptor=R1'.X
    }  
    switch (fieldDescriptor.getType()) {  
        case INT32:  
            value = ((Long)value).intValue();  
            break;  
        case BYTES:  
            value = ZeroCopyByteString.wrap((byte[])value);  
            break;  
        case MESSAGE:  
            value = TupleFieldsHelper.toProto(value, fieldDescriptor.getMessageType());  
            break;  
        case ENUM:  
            value = fieldDescriptor.getEnumType().findValueByNumber(((Long)value).intValue());  
            break;  
        default:  
            break;  
    }  
    recordBuilder.setField(fieldDescriptor, value);                                                                                                                 [3] recordBuilder=R1', fieldDescriptor=R1'.X
                                                                                                        [4] recordBuilder=R1, fieldDescriptor=R1'.X -> ERROR
    return true;  
}
```

The error we get at step 4 results from [this](https://github.com/protocolbuffers/protobuf/blob/743bf9214601eddaf13456b1029247e0a6d9a92f/java/core/src/main/java/com/google/protobuf/DynamicMessage.java#L500C35-L500C35) check in ProtoBuf which is trying to compare the field's containing type with the message type:

```java
  /** Verifies that the field is a field of this message. */
  private void verifyContainingType(FieldDescriptor field) {
    if (field.getContainingType() != type) {
      throw new IllegalArgumentException("FieldDescriptor does not match message type.");
    }
  }
```

Although the field correctly belongs to the evolved message descriptor.

The included test exposes this race condition, it will fail if one makes the method `copy` not reentrant.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
@@ -174,7 +174,7 @@ public class IndexKeyValueToPartialRecord {
         }
 
         @Override
-        public boolean copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Builder recordBuilder,
+        public synchronized boolean copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull Message.Builder recordBuilder,
                             @Nonnull IndexEntry kv) {
             final Tuple tuple = (source == TupleSource.KEY ? kv.getKey() : kv.getValue());
             if (!copyIfPredicate.test(tuple)) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
@@ -157,8 +157,7 @@ public class IndexKeyValueToPartialRecord {
         private final Predicate<Tuple> copyIfPredicate;
         @Nonnull
         private final ImmutableIntArray ordinalPath;
-        private Descriptors.FieldDescriptor fieldDescriptor;
-        private Descriptors.Descriptor containingType;
+        private final Descriptors.FieldDescriptor fieldDescriptor;
 
         private FieldCopier(@Nonnull final Descriptors.FieldDescriptor fieldDescriptor,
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -185,27 +183,27 @@ public boolean copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull M
             if (value == null) {
                 return true;
             }
-            if (!containingType.equals(recordDescriptor)) {
-                containingType = recordDescriptor;
-                fieldDescriptor = recordDescriptor.findFieldByName(field);
+            Descriptors.FieldDescriptor mutableFieldDescriptor = this.fieldDescriptor;
```
hatyo, line 186: Is it possible that this is causes too many small object allocations stressing GC if `this.fieldDescriptor` is _always_ different from the `recordDescriptor` for a given execution?

MMcM (approved):

Issue #2303, opened 2023-09-14
Make it possible to reset IndexMaintainerRegistryImpl for testing purposes


butlermh:
IndexMaintainerRegistryImpl can only be initialized once because it uses a static instance. 

This can create problems during testing. 

This change adds a reset method that re-initializes the registry.

PR #2304, opened 2023-09-14
Make it possible to reset IndexMaintainerRegistryImpl for testing purposes #2303

butlermh:
IndexMaintainerRegistryImpl can only be initialized once because it uses a static instance.

This can create problems during testing.

This change adds a reset method that re-initializes the registry.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
@@ -45,7 +45,7 @@ public class IndexMaintainerRegistryImpl implements IndexMaintainerRegistry {
     protected static final IndexMaintainerRegistryImpl INSTANCE = new IndexMaintainerRegistryImpl();
 
     @Nonnull
-    private final Map<String, IndexMaintainerFactory> registry;
+    private Map<String, IndexMaintainerFactory> registry;
 
     @Nonnull
     public static IndexMaintainerRegistry instance() {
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
```diff
@@ -45,7 +45,7 @@ public class IndexMaintainerRegistryImpl implements IndexMaintainerRegistry {
     protected static final IndexMaintainerRegistryImpl INSTANCE = new IndexMaintainerRegistryImpl();
 
     @Nonnull
-    private final Map<String, IndexMaintainerFactory> registry;
+    private Map<String, IndexMaintainerFactory> registry;
```
normen662, line 48: Since this is not `final` anymore, the constructor call does not guarantee visibility in other threads anymore.

https://stackoverflow.com/questions/33702625/java-memory-visibility-in-constructors

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
butlermh: I want to avoid having to do something like this in my test code:
```
    static class ResettableIndexMaintainerRegistryImpl {
        public static void reset() {
            // warning - reflection - reset the static registry everytime we reset Guice
            try {
                IndexMaintainerRegistry registry = IndexMaintainerRegistryImpl.instance();
                Field privateField = registry.getClass().getDeclaredField("registry");
                privateField.setAccessible(true);
                privateField.set(registry, initRegistry());
            } catch (Exception e) {
                //
            }
        }

        protected static Map<String, IndexMaintainerFactory> initRegistry() {
            final Map<String, IndexMaintainerFactory> registry = new HashMap<>();
            for (IndexMaintainerFactory factory : ServiceLoaderProvider.load(IndexMaintainerFactory.class)) {
                for (String type : factory.getIndexTypes()) {
                    registry.put(type, factory);
                }
            }
            return registry;
        }
    }
```

PR #2306, opened 2023-09-21
Support Java UDF.

### LINKED ISSUE #2307: Support Java UDFs in Cascades
We would like to provide an easy way for the users to define their own user-defined functions that can be used in the context of query planning and execution through Cascades.

Similar to other vendors, we would provide an interface that the user is supposed to implement and provide their own business logic. When the function is invoked, the planner looks up the function by name, match it with the provided arguments and provide any necessary type promotions.

Here is an example of the intended usage:

```java
public abstract class JavaUdf extends AbstractValue {

    ....

    @Nonnull
    public static Type getUdfResultType() {
        return Type.primitiveType(Type.TypeCode.UNKNOWN);
    }

    @Nonnull
    public static Iterable<Type> getUdfParameterTypes() {
        return List.of();
    }

    @Nullable
    @Override
    public <M extends Message> Object eval(@Nonnull final FDBRecordStoreBase<M> store, @Nonnull final EvaluationContext context) {
        return call(Streams.stream(children).map(c -> c.eval(store, context)).collect(Collectors.toList()));
    }

    @Nullable
    public abstract Object call(@Nonnull final List<Object> parameters);
}
```

Example implementation from the user:

```java
 public static final class SumUdf extends JavaUdf {
        ....
        @Nonnull
        public static Type getUdfResultType() {
            return Type.primitiveType(Type.TypeCode.LONG);
        }

        @Nonnull
        public static Iterable<Type> getUdfParameterTypes() {
            return List.of(Type.primitiveType(Type.TypeCode.LONG), Type.primitiveType(Type.TypeCode.LONG));
        }

        @Nullable
        @Override
        public Object call(@Nonnull List<Object> parameters) {
            Verify.verify(parameters.size() == 2);
            return (long)(parameters.get(0)) + (long)(parameters.get(1));
        }
```

Example usage:
```sql
select java_call('my.package.SumUdf', col1, 42) from t1)
```

hatyo:
This adds support for Java UDFs in Cascades. It enables defining a custom UDF by subclassing `JavaUdf`, override its method and hide its static methods. It can be encapsulated by `JavaUserDefinedFunction` which takes care of:

1. Looking up the function in the loaded classes via reflection.
2. Match invocation arguments with the UDFs parameters and perform any necessary type promotion(s).

This resolves #2307 .


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUdf.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * JavaUdf.java
```
normen662, line 2:
```suggestion
 * JavaUdfValue.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUdf.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Nonnull
+    @Override
+    public Type getResultType() {
+        return getUdfResultType();
+    }
+
+    @Nonnull
+    public static Type getUdfResultType() {
+        return Type.primitiveType(Type.TypeCode.UNKNOWN);
```
normen662, line 65: The encapsulation process must resolve a result type other than `UNKNOWN`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUdf.java
```diff
@@ -0,0 +1,91 @@
...
+    @Nonnull
+    @Override
+    public Iterable<? extends Value> getChildren() {
+        return children;
+    }
+
+    @Nonnull
+    @Override
+    public Type getResultType() {
+        return getUdfResultType();
```
normen662, line 60: This only returns `UNKNOWN`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUdf.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Nonnull
+    public static Type getUdfResultType() {
+        return Type.primitiveType(Type.TypeCode.UNKNOWN);
+    }
+
+    @Nonnull
+    public static Iterable<Type> getUdfParameterTypes() {
+        return List.of();
```
normen662, line 70: This should always return the declared types of the parameters I suppose. I think we should enforce the implementor to properly define those (same with the result type). On top of that the arity of this returned list (which should be a list not an iterable), should match the number of children I think. At least that would be a desirable invariant.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUserDefinedFunction.java
```diff
@@ -0,0 +1,127 @@
...
+        // sanity / security check
+        if (!JavaUdf.class.isAssignableFrom(clazz)) {
+            throw new RecordCoreException(String.format("expecting '%s' to be a subclass of '%s'", clazz.getName(), JavaUdf.class.getName()));
+        }
+
+        final var fClazz = Verify.verifyNotNull(clazz);
+        // @param arguments The types of the arguments, must be the same type or of a promotable type of the corresponding subclass expected {@link Value} parameter(s).
+        Method method;
+        try {
+            method = fClazz.getMethod("getUdfParameterTypes");
```
normen662, line 76: Why resolve a method using duck typing here on `JavaUdf`? You should resolve the user-supplied descendent of `BuiltinFunction` here, instantiate it or get it from a cache and then call its `encapsulate` method (or a derivate thereof).

hatyo: > Why resolve a method using duck typing here on JavaUdf?

It is not duck-typing, the provided type must be a descendant of `JavaUdf` as per line 68.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/JavaUdf.java
hatyo: In the new commit, this is what happens:
- The `UdfFunction`'s encapsulation will retrieve a list of arguments.
- It will perform necessary promotions of these arguments to match the user-defined set of parameters given by the user, injecting `PromoteValue`(s) as necessary.
- It will then create a new call-site, using the user-defined factory method.
- The user-defined factory method is responsible for instantiating a new `Value`, the return type of that `Value` is determined by the `Value` implementation itself.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/UdfFunction.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Nonnull
+    protected abstract UdfValue newCallsite(@Nonnull List<Value> arguments);
+
+    @SuppressWarnings("unchecked")
+    @Nonnull
+    @Override
+    public final Typed encapsulate(@Nonnull final List<? extends Typed> arguments) {
+        arguments.forEach(argument -> Verify.verify(argument instanceof Value));
```
normen662, line 61: can you either fully support variadic functions or put a check in here somewhere to not allow those? or I just didn't see that :-)...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/UdfValue.java
```diff
@@ -0,0 +1,89 @@
...
+    @Nonnull
+    @Override
+    public abstract Value withChildren(Iterable<? extends Value> newChildren);
+
+    @Nullable
+    public abstract Object call(@Nonnull List<Object> arguments);
+
+    @Override
+    public int hashCodeWithoutChildren() {
+        return PlanHashable.objectsPlanHash(PlanHashKind.FOR_CONTINUATION, BASE_HASH);
```
normen662: use the class name in addition

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/UdfValue.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * This represents a user-defined {@link Value}.
+ */
+public abstract class UdfValue extends AbstractValue {
```
normen662, line 39: Can you write a longish comment how the flow of this is supposed to compile in?

normen662 (changes requested):

normen662 (approved):

normen662 (approved):

PR #2314, opened 2023-10-11
Resolves #2313: Support synthetic records that allow for "unnesting" repeated nested fields

### LINKED ISSUE #2313: Support synthetic records that allow for "unnesting" repeated nested fields
One limitation of key expressions is that the way they handle nested repeated things makes it hard to align different elements from the same repeated element. For example, take this record:

```protobuf
message ARecordType {
    optional int64 id = 1 [(field).primary_key = true];
    
    message Nested {
        optional string foo = 1;
        optional string bar = 2;
   }
   repeated Nested element = 2;
   optional string other_top_level_element = 3;
}
```

Within this element, there's no way currently of expressing a query like "return all the `bar` values from nested elements of `ARecordType` where `foo` is equal to `$foo`". The closest is something like:

```java
RecordQuery.newBuilder()
    .setRecordType("ARecordType")
    .setQuery(Query.field("element").oneOfThem().matches(Query.field("foo").equalsParamater("foo")))
    .setRequiredResults(List.of(Key.Expressions.field("element", FanType.FanOut).nest("bar")))
    .build()
```

However, the `element.bar` in in the required results actually means "get _all_ `element.bar`s", not just the ones associated with the matching `element.foo`s.

Similarly, there's no way of defining an index on `element.foo, other_top_level_element, element.bar` where `element.foo` and `element.bar` come from the _same_ `Nested` entry. (Right now, it's only possible to create an element on the cross-product.)

One possible way to support this is by using synthetic record types. Synthetic record types were introduced to model joins for the purposes of constructing joined indexes. Well, a nested repeated element can be thought of as representing a special kind of join between the nested child type and its parent. By default, this join is denormalized into a nested repeated structure, but we could use a synthetic type to re-transform the operation into a different structure.

alecgrieser:
This introduces a new synthetic record type, the `UnnestedRecordType`. It is constructed from one "parent" constituent and several "nested" constituents. Each one of those nested constituents represents one occurrence of a nested element on the parent record (or on another nested constituent). This has an advantage over traditional key expressions, in that it allows for expressions to be constructed that reference different fields from a nested repeated expression in a way that refers to the fields from the _same_ nested message. For example, this type:

```proto
message ParentRecord {
   message ChildRecord {
      optional int64 foo = 1;
      optional string bar = 2;
   }
   repeated ChildRecord child = 1;
   optional string other = 2;
}
```

Without this new synthetic type, it is impossible to construct an index on `(child.foo, other, child.bar)` where `foo` and `bar` each come from the _same_ child message. With the new type, we can first construct a special synthetic record type:

```java
final UnnestedRecordTypeBuilder typeBuilder = metaDataBuilder.addUnnestedRecordType("ParentAndChild", "ParentRecord");
typeBuilder.addNestedConstituent("c", ParentRecord.ChildRecord.getDescriptor(), UnnestedRecordType.PARENT_CONSTITUENT,
    field("child", FanType.Concatenate), field("child", FanType.FanOut));
```

(By default, every unnested record type has one constituent pointing to the parent record. This adds a second constituent called `c` that points to the child.) Then we can define an index on:

```java
concat(field("c").nest("foo"), field(UnnestedRecordType.PARENT_CONSTITUENT).nest("other"), field("c").nest("bar"))
```

Which gives us what we want.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/UnnestedRecordTypeBuilder.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @Nullable
+    private static Descriptors.Descriptor findDescriptorByName(@Nonnull Descriptors.Descriptor descriptor, @Nonnull String fullName) {
+        if (fullName.startsWith(descriptor.getFullName())) {
+            if (descriptor.getFullName().equals(fullName)) {
+                return descriptor;
+            }
+            for (Descriptors.Descriptor nestedType : descriptor.getNestedTypes()) {
+                Descriptors.Descriptor foundNested = findDescriptorByName(nestedType, fullName);
```
pengpeng-lu, line 210: I found a problem with this piece of code:
message MapRecord {
  int64 id = 1;
  MapFromStringToString map_field = 2;
}
getNestedTypes() returns empty list. Maybe "nestedTypes" here only mean message types defined inside MapRecord.
We should probably do:
for (FieldDescriptor field: descriptor.getFields()) {
  if (field.getType() == MESSAGE) {
    Descriptor foundNested = findDescriptorByName(field.getDescriptor(), fullName);
    if (foundNested != null) { return foundNested; }
  }
}

alecgrieser: Yeah, `descriptor.getNestedTypes()` returns the list of nested messages defined within the type. The goal of this code is to find, by iterating through the list of defined message types, is to find the type that matches the given name. So, if you had something like:

```proto
package com.example.package;

message MapFromStringToString {
    message Entry {
        optional string key = 1;
        optional string value = 2;
    }
    repeated Entry entry = 1;
}

message MapRecord {
    int64 id = 1;
    MapFromStringToString map_field = 2;
}
```

Then I'd expect the `fullName` to be something like `com.example.package.MapFromStringToString.Entry`. When it gets to the `MapRecord`, that name shouldn't match, and it should move on. When it gets to `MapFromStringToString`, it should look at its nested types to find the correct `Entry` type.

Importantly, this can be defined within the same stored record type or outside of it, and it can be within the same file (as in this example) or in one of the dependencies. If you have a potential example of full name and file descriptor that it fails to find, though, that would be good to know.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/UnnestedRecordTypeBuilder.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @API(API.Status.INTERNAL)
+    public UnnestedRecordTypeBuilder(@Nonnull RecordMetaDataProto.UnnestedRecordType typeProto, @Nonnull final RecordMetaDataBuilder metaDataBuilder) {
+        super(typeProto.getName(), LiteralKeyExpression.fromProtoValue(typeProto.getRecordTypeKey()), metaDataBuilder);
+
+        // Parent is a bit special. Go through the constituents and attempt to find it
+        String parentTypeName = null;
+        for (RecordMetaDataProto.UnnestedRecordType.UnnestedConstituent constituentProto : typeProto.getUnnestedConstituentsList()) {
+            if (constituentProto.getName().isEmpty()) {
```
pengpeng-lu, line 127: Instead of empty string, can we use another constant? I tried to use empty string, but having an empty fieldName in keyExpression seemed to break the index validator. (I'm not sure which line of code it broke, but I tried "__parent" and it worked.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOfFunctionKeyExpression.java
```diff
@@ -0,0 +1,151 @@
...
+ * two arguments:
+ *
+ * <ol>
+ *     <li>A list of items.</li>
+ *     <li>A scalar object.</li>
+ * </ol>
+ *
+ * <p>
+ * It returns a single column, which will contain the (zero-indexed) position of the scalar object
+ * within the list. For example, on a message like:
```
MMcM, line 48: The _first occurrence_ of the object in the list, right? Also, `null` if not found?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/NestedRecordTypeBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Builder class for {@link NestedRecordType}s. This is used primarily by the {@link UnnestedRecordTypeBuilder} to keep
+ * track of nested types.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class NestedRecordTypeBuilder extends RecordTypeBuilder {
+    public NestedRecordTypeBuilder(@Nonnull final Descriptors.Descriptor descriptor) {
+        super(descriptor);
+    }
+}
```
MMcM, line 37: Doesn't this need to override `build` to make the subclass? (Never called by `SyntheticRecordTypeBuilder`'s subclasses' `build` I guess.)

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/UnnestedRecordPlanner.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.apple.foundationdb.record.query.plan.synthetic;
+
+import com.apple.foundationdb.record.metadata.SyntheticRecordType;
+import com.apple.foundationdb.record.metadata.UnnestedRecordType;
+
+import javax.annotation.Nonnull;
+
+class UnnestedRecordPlanner {
```
ScottDugas, line 28: Should this be tagged `EXPERIMENTAL`

ScottDugas (changes requested):
I haven't gone through the `plan` or tests, but wanted to give a partial review as I go along.

alecgrieser (commented):
I've begun going through the comments, but I figured I'd start with something a bit experimental. In response to it being pointed out that the `index_of` function doesn't really play nicely with duplicates in the entry, I've suggested an alternative approach where the position information is encoded in a special field within the synthetic record. This special field has a nested type with one field of type integer for each (nested) constituent. That change is within 5bf72339ba9a691fc771c38bad476b6932dd90d3 if you want to look at it separately. I added a test case (that fails with the old code) of duplicate entries in a451f7f48a5a82e29bd92e924ae2fd55c4d6d25f, and I have removed the new function entirely in 372e31110672d367f7df21746ac602a314599a31

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBMapQueryTest.java
```diff
@@ -0,0 +1,1241 @@
...
+            RecordQueryPlan plan = planner.plan(query);
+            assertMatchesExactly(plan, planMatcher);
+
+            final KeyExpression recIdExpr = field(UnnestedRecordType.PARENT_CONSTITUENT).nest("rec_id");
+            final KeyExpression keyAndValueExpr = field("entry").nest(concatenateFields("key", "value"));
+            for (String key : keys) {
+                EvaluationContext evaluationContext = EvaluationContext.forBinding(keyParam, key);
+                final List<TestRecordsNestedMapProto.OuterRecord> queried = plan.execute(recordStore, evaluationContext)
+                        .map(rec -> {
+                            // Nest the data back into an OuterRecord
```
ScottDugas, line 608: It's probably not worth changing, but for future reference stylistic advice, I found this hard to follow. It's not clear why you map it back into an `OuterRecord` rather than simply asserting directly about the synthetic record that comes back.

ScottDugas (changes requested):

alecgrieser (commented):
Followed up on some (but not all) of the comments. I've responded to the ones that I've attempted to address. I also have some local changes that shouldn't have been pushed (yet), and those will hopefully be coming soon

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlanner.java
```diff
@@ -285,6 +297,19 @@ public SyntheticRecordFromStoredRecordPlan forJoinConstituent(@Nonnull JoinedRec
         return new JoinedRecordPlanner(joinedRecordType, queryPlanner).plan(joinConstituent);
     }
 
+    @Nonnull
```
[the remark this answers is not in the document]
alecgrieser: Added

ScottDugas (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
```diff
@@ -0,0 +1,841 @@
...
+                .setOtherId(1)
+                .setMap(TestRecordsNestedMapProto.MapRecord.newBuilder()
+                        .addEntry(TestRecordsNestedMapProto.MapRecord.Entry.newBuilder().setKey("foo").setValue("bar").setIntValue(1))
+                        .addEntry(TestRecordsNestedMapProto.MapRecord.Entry.newBuilder().setKey("baz").setValue("qux").setIntValue(2))
+                        .addEntry(TestRecordsNestedMapProto.MapRecord.Entry.newBuilder().setKey("wid").setValue("get").setIntValue(3))
+                )
+                .build();
+    }
+
+    @Test
```
ScottDugas, line 222: Sonarcloud detected that you have a `@Test` that will be silently ignored
```suggestion
    @Nonnull
```

ScottDugas (changes requested):

ScottDugas (approved):

MMcM (approved):

alecgrieser:
Talked to @pengpeng-lu out of band, and she's okay with this going in, so I'll go ahead and merge this

PR #2322, opened 2023-10-23
Resolves #2321: Lucene highlight: enable Lucene highlight with text and range queries

### LINKED ISSUE #2321: Lucene highlight: enable Lucene highlight with text and range queries
similar to BITSET, range should be enabled as a predicate as well on top of the text predicate for highlighting

qdnguyen0:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQueryClause.java
```diff
@@ -146,6 +147,10 @@ protected static Map<String, Set<String>> getHighlightingTermsMap(@Nonnull final
             BitSetQuery bitsetQuery = (BitSetQuery)query;
             String field = bitsetQuery.getField();
             highlightingTermsMap.computeIfAbsent(field, key -> new HashSet<>()).add(field.toLowerCase(Locale.ROOT));
+        } else if (query instanceof PointRangeQuery) {
```
ohadzeliger, line 150: What is the intent here? IIUC, this change would allow queries with a range of int(long) point to be included. Is the idea to allow them to be matching the doc but then ignored for highlight? Is the idea to do the same for term range ("a" TO "b")?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/FDBLuceneHighlightingTest.java
```diff
@@ -101,6 +101,22 @@ void highlightedBitsetQuery() {
         }
     }
 
+    @Test
```
ohadzeliger, line 104: Other types that are included in the change are the rest of the subclasses of PointRange (BigInteger, bBinary, Double etc). They should be tested.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/FDBLuceneHighlightingTest.java
```diff
@@ -101,6 +101,22 @@ void highlightedBitsetQuery() {
...
+    void highlightedRangeQuery() {
```
ohadzeliger, line 105: What about text range? Is this allowed? Should it be tested?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQueryClause.java
qdnguyen0: Yes, we want to allow matching using a value or a range. If there's no text part in the clause, then the highlighting would be ignored (say if `isExpunged: 0` is the only predicate) and `highlights` would be empty in the response.

We wouldn't send a`group:  ["a" TO "b"]` predicate. But `group: "a"` would be a TermQuery and that's already allowed. Either way, tha `"a"` and `"b"` should be ignored for highlighting

qdnguyen0:
@ohadzeliger added tests for more data range types

ohadzeliger (approved):

PR #2326, opened 2023-10-25
Fix unstable in hash code of `RecordQueryAbstractDataModificationPlan`

### LINKED ISSUE #2325: `RecordQueryAbstractDataModificationPlan` hash code is not stable
The hash code calculation of the `RecordQueryAbstractDataModificationPlan` is not stable. This is causing discrepancies in the optimiser choosing optimal `UPDATE` plans pseudo-randomly when multiple equivalent alternatives exist and the plan hash code is used as a costing tie-breaker.

hatyo:
The hash code calculation of the `RecordQueryAbstractDataModificationPlan` is not stable. This is causing discrepancies in the optimiser choosing optimal `UPDATE` plans pseudo-randomly when multiple equivalent alternatives exist and the plan hash code is used as a costing tie-breaker.

This happens because the `MessageHelpers#CoercionTrieNode` class does not override `equals` and `hashCode`, despite it being used in `computeHashCodeWithoutChildren` of `RecordQueryAbstractDataModificationPlan`.

The provided fix implements `hashCode` and `equals` properly for this class, and adds a number of unit tests for equality and hash code calculation for `MessageHelpers#CoercionTrieNode` and `MessageHelpers#TransformationTrieNode`, it also adds an integration test for that examines the stability of plan hash of an `UPDATE` plan (which leverages the `RecordQueryAbstractDataModificationPlan` physical operator).

This fixes #2325.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
@@ -44,7 +44,6 @@ import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.function.BiFunction;
 
 /**
@@ -490,7 +489,6 @@ public class MessageHelpers {
     /**
      * Trie data structure of {@link Type.Record.Field}s to {@link Value}s.
      */
-    @SuppressWarnings({"squid:S1206", "squid:S2160", "PMD.OverrideBothEqualsAndHashcode"})
     public static class TransformationTrieNode extends TrieNode<Integer, Value, TransformationTrieNode> {
 
         public TransformationTrieNode(@Nullable final Value value, @Nullable final Map<Integer, TransformationTrieNode> childrenMap) {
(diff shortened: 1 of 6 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/TrieNode.java
```diff
@@ -77,4 +78,22 @@ public Collection<T> values() {
                 .flatMap(trie -> trie.getValue() == null ? Stream.of() : Stream.of(trie.getValue()))
                 .collect(ImmutableList.toImmutableList());
     }
+
+    @Override
+    public boolean equals(final Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (!(other instanceof TrieNode)) {
```
MMcM, line 87: Could this compare `getClass` instead and eliminate the need for the subclass to override?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -579,5 +576,19 @@ public CoercionTrieNode(@Nullable final BiFunction<Descriptors.Descriptor, Objec
         public CoercionTrieNode getThis() {
             return this;
         }
+
+        @Override
+        public boolean equals(final Object other) {
+            if (this == other) {
+                return true;
+            }
+            if (!(other instanceof CoercionTrieNode)) {
```
MMcM, line 585: This is the only difference, right?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/TrieNode.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
hatyo: Yes, correct, I refactored it accordingly.

MMcM (approved):

PR #2331, opened 2023-10-31
RecordQueryAbstractDataModificationPlan does not produce stable hash with CoercionTrieNode

### LINKED ISSUE #2330: RecordQueryAbstractDataModificationPlan does not produce stable hash with CoercionTrieNode
CoercionTrieNode takes `BiFunction<Descriptors.Descriptor, Object, Object>` as its value that is used in computing the `equals()` and `hashCode()` [in `TrieNode`]. Since, functional interfaces does not define for semantic equality, the value is compared with ref equality. Hence, using `CoercionTrieNode` is plan hash of `RecordQueryAbstractDataModificationPlan` makes it unstable for even identical coercions. 

We can either remove `CoercionTrieNode` from contributing to the planHash or we can override the equals() and hashCode() semantics in `CoercionTrieNode` to derive hash only out of the structure and not considering the value.

g31pranjal:
fixes #2330

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
@@ -22,6 +22,7 @@ package com.apple.foundationdb.record.query.plan.cascades.values;
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.PlanHashable;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.metadata.expressions.TupleFieldsHelper;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
@@ -33,6 +34,7 @@ import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
 import com.apple.foundationdb.record.util.TrieNode;
 import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.ImmutableIntArray;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.DynamicMessage;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -560,5 +561,10 @@ public CoercionTrieNode(@Nullable final BiFunction<Descriptors.Descriptor, Objec
         public CoercionTrieNode getThis() {
             return this;
         }
+
+        @Override
+        public int planHash(@Nonnull final PlanHashKind hashKind) {
+            return PlanHashable.objectPlanHash(hashKind, getChildrenMap());
```
normen662, line 567: I don't think this is correct as the Map is not `PlanHashable` and will therefore delegate to the regular `hashCode()`.

normen662 (changes requested):
I have a comment in the code and I would also like to see a new unit test or adapted unit test in `FDBModificationQueryTest`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -560,5 +562,10 @@ public CoercionTrieNode(@Nullable final BiFunction<Descriptors.Descriptor, Objec
         public CoercionTrieNode getThis() {
             return this;
         }
+
+        @Override
+        public int planHash(@Nonnull final PlanHashKind hashKind) {
+            return PlanHashable.objectsPlanHash(hashKind, getChildrenMap() == null ? ImmutableSet.of() : getChildrenMap().keySet(), getChildren());
```
normen662, line 568: is the `keySet` stable? I think until we fix a problem with plan hashable that treats sets as an iterable you will have to hashCode the set here directly or you fix the set problem in planhashable.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBModificationQueryTest.java
```diff
@@ -801,6 +802,14 @@ void planHashCodeIsCalculatedCorrectly() throws Exception {
...
+        try (FDBRecordContext context = openContext()) {
+            openNestedRecordStore(context);
+            var plan1 = getUpdateArrayPlan(cascadesPlanner);
+            var plan2 = getUpdateArrayPlan(cascadesPlanner);
+            Assertions.assertEquals(plan1.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS),
+                    plan2.planHash(PlanHashable.PlanHashKind.STRUCTURAL_WITHOUT_LITERALS));
+        }
     }
 
     private RecordQueryPlan getUpdatePlan(final CascadesPlanner cascadesPlanner) {
```
normen662, line 815: `@NonNull`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBModificationQueryTest.java
```diff
@@ -846,6 +855,53 @@ private RecordQueryPlan getUpdatePlan(final CascadesPlanner cascadesPlanner) {
                 EvaluationContext.empty()).getPlan();
     }
 
+    private RecordQueryPlan getUpdateArrayPlan(final CascadesPlanner cascadesPlanner) {
```
normen662, line 858: `@Nonnull`

normen662 (changes requested):

normen662 (dismissed):

normen662 (approved):

PR #2348, opened 2023-11-07
Resolve #2345: Fix BufferOverflowException triggered by FDBIndexOutput

### LINKED ISSUE #2345: Fix BufferOverflowException triggered by FDBIndexOutput

jjezra:
Attempt to rephrase the expression.


jjezra:
Rebased to resolve conflict.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
```diff
@@ -100,8 +98,8 @@ public FDBIndexOutput(@Nonnull String resourceDescription, @Nonnull String name,
         this.fdbDirectory = fdbDirectory;
         blockSize = fdbDirectory.getBlockSize();
         buffer = BUFFERS.poll();
-        if (buffer == null) {
-            buffer = ByteBuffer.allocate((int)blockSize);
+        if (buffer == null || buffer.capacity() != blockSize) {
+            buffer = ByteBuffer.allocate(blockSize);
```
ohadzeliger, line 102: It may be interesting to know how many of these allocations we've done

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
```diff
@@ -216,8 +217,14 @@ private void flush() {
             buffer.flip();
             byte[] arr = new byte[buffer.remaining()];
             buffer.get(arr);
-            flushes.add(fdbDirectory.writeData(id, (int) ( (currentSize - 1) / blockSize), arr));
-            buffer.clear();
+            flushes.add(fdbDirectory.writeData(id, ( (currentSize - 1) / blockSize), arr));
```
ohadzeliger, line 220: I think (we've discussed in another context) that these are synchronous calls and should not be treated as async ones. This may help investigate this issue as turning the call to a sync may simplify the code path.

ohadzeliger (changes requested):
Left 2 comments. Understanding that this is part of the investigation into why we are seeing the issue in iCloud3 we may skip these comments, but I wanted to ring them up for discussion.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
jjezra: It is my understanding that adding log messages here will be a flood. I wouldn't like to do it in the same PR. 
(Note that this buffer can also be created after the first pool initialization)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
jjezra: I'm not sure that I follow. How is it related to this change?

ohadzeliger: What the FDBDirectory does is to create an Async initiation of a synchronous operation. I am not sure as to why this is done that way since (IIUC) the write to the transaction is immediate (in memory only). In another area I disabled that same exact construct.
Disabling that for now may help in troubleshooting since we won't have multi-threading here anymore.
I also don't mind doing this in another PR.

ohadzeliger: Will do in a subsequent PR

ohadzeliger (approved):

PR #2350, opened 2023-11-09
Resolves #2336: PlanOrderingKey.mergedComparisonKey too strict about fields between requested ordering and primary key

### LINKED ISSUE #2336: PlanOrderingKey.mergedComparisonKey too strict about fields between requested ordering and primary key
Given an index on `a, b, c, d` on a record type with primary key `k`, and a query with equality on `a`, disjunct of equality on `b`, and sort on `c`, the subplans are rejected because the candidate key `[a, b, k]` and `[a, b, c, d, k]` don't align well enough for it. But provided `c` and `d` are 1:1 with `k` (that is, not repeated / `createsDuplicates`), the actual (common) `planOrderingKey` will work. Such a Union will do extra comparing of these fields which ideally would be eliminated (by returning the candidate?) but is otherwise harmless. When the same pkey is on both sides, the extraneous fields will always match as well, I believe.

alecgrieser:
This adjusts the way the plan ordering key is constructed for unions and intersections so that it can accept additional columns in the order. Effectively, in the case where you have an index on `[a, b, c, d]` with an equality on `a`, a disjunct of equalities on `b`, and a primary key `k` (the hypothetical raised in the original issue), then this will construct an ordered union between multiple scans of the index (one leg for each member of the disjunction on `b`) and will use the comparison key `[c, d, k]` to ensure that the results are processed in an expected order.

This resolves #2336.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
@@ -2349,7 +2349,7 @@ public class TextIndexTest extends FDBRecordStoreTestBase {
                     queryComplexDocumentsWithOr((OrComponent) Query.or(
                             Query.field("text").text().containsAll("unit named after"),
                             Query.field("text").text().containsPhrase("אן ארמיי און פלאט")
-                    ), 0, -1558384887));
+                    ), 0, -396375489));
 
             assertEquals(Collections.singletonList(Tuple.from(1L, 5L)),
                     queryComplexDocumentsWithIndex(
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
@@ -22,12 +22,18 @@ package com.apple.foundationdb.record.provider.foundationdb.query;
 
 import com.apple.foundationdb.record.ExecuteProperties;
 import com.apple.foundationdb.record.PlanHashable;
+import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.RecordCursorIterator;
+import com.apple.foundationdb.record.RecordCursorResult;
+import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TestRecords1Proto;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.metadata.RecordType;
+import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyWithValueExpression;
 import com.apple.foundationdb.record.provider.foundationdb.FDBQueriedRecord;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
 import com.apple.foundationdb.record.provider.foundationdb.IndexScanComparisons;
 import com.apple.foundationdb.record.provider.foundationdb.IndexScanParameters;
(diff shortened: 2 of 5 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
```diff
@@ -2349,7 +2349,7 @@ void queryComplexDocumentsWithAdditionalFilters() throws Exception {
                     queryComplexDocumentsWithOr((OrComponent) Query.or(
                             Query.field("text").text().containsAll("unit named after"),
                             Query.field("text").text().containsPhrase("אן ארמיי און פלאט")
-                    ), 0, -1558384887));
+                    ), 0, -396375489));
```
alecgrieser, line 2352: This changed because the `comparisonKey` was adjusted from `concatenateFields("group", "rec_no")` (i.e., the primary key) to just `field("rec_no")`, as the `group` field was in the set of equality-bound predicates. I could see an argument that we should include the primary key back in here, I guess, though it's not strictly necessary, but it may decrease the number of changed plans

MMcM (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/TextIndexTest.java
normen662: I would imagine this lines up with the plan given by the Cascades planner which IIRC returns the shortest comparison key.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
```diff
@@ -1077,6 +1092,320 @@ void testOrQueryOrdered() throws Exception {
         }
     }
 
+    /**
+     * Verify that the same index can be matched for multiple legs of an OR query, even if there are (compatible) ordering
+     * constraints.
+     * @throws Exception from test set up
+     */
+    @DualPlannerTest
```
normen662, line 1100: Thank you for making this a dual planner test.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
```diff
@@ -1077,6 +1092,320 @@ void testOrQueryOrdered() throws Exception {
...
+        } else {
+            final BindingMatcher<? extends RecordQueryPlan> planMatcher =
+                    fetchFromPartialRecordPlan(
+                            RecordQueryPlanMatchers.unionOnValuesPlan(
+                                    coveringIndexPlan()
+                                            .where(indexPlanOf(indexPlan().where(indexName("multi_index")).and(scanComparisons(range("[[odd, 0],[odd, 0]]"))))),
+                                    coveringIndexPlan()
+                                            .where(indexPlanOf(indexPlan().where(indexName("multi_index")).and(scanComparisons(range("[[odd, 2],[odd, 2]]"))))))
+                                    // note that the comparison key is on num_value_3_indexed + the primary key
+                                    .where(comparisonKeyValues(exactly(fieldValueWithFieldNames("num_value_3_indexed"), fieldValueWithFieldNames("num_value_2"), fieldValueWithFieldNames("rec_no")))));
```
normen662, line 1213: This is strange as we should drop the `num_value_2` part.

normen662: Maybe I am misremembering the details of planning this in the Cascades Planner but we should track that. Can you please open an issue for me?

normen662 (changes requested):
This looks good. Will approve once we get this question resolved: Why don't you use `Ordering` for all the logic around synthesizing a better comparison key? That class and a bunch of other helpers should have the logic you need already built-in.

alecgrieser:
Rebasing to re-run tests to protect against merge skew

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
alecgrieser: Discussed this a bit offline, but we believe the difference here is that the Cascades planner includes the equality bound field here because it can make the result of the union usable in more cases (e.g., if there's a requirement that the plan be ordered by that field that the planner doesn't (yet) know about). In any case, the additional field is harmless (it effectively means that it will consume one leg of the union first before the other leg), but not necessary. There are other cases where one of the equality bound fields is also an ordering key where the inclusion of the field is _required_, and there are already some tests for that

MMcM (approved):

PR #2352, opened 2023-11-09
Resolve #2351: Convert FDBDirectory.writeData to a regular func

### LINKED ISSUE #2351: Convert FDBDirectory.writeData to a regular func
Currently, this function returns a completable future. This seems to be un-needed and may be causing "access closed objects" failures.

jjezra:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
```diff
@@ -217,7 +196,7 @@ private void flush() {
             buffer.flip();
             byte[] arr = new byte[buffer.remaining()];
             buffer.get(arr);
-            flushes.add(fdbDirectory.writeData(id, ( (currentSize - 1) / blockSize), arr));
+            actualSize += fdbDirectory.writeData(id, ((currentSize - 1) / blockSize), arr);
```
ohadzeliger, line 199: This makes the class non-thread-safe. Are we OK with that at this point?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutput.java
jjezra: Yes. This writing module, by definition, cannot be used asynchronously - as the written data must be consistent.

ohadzeliger (approved):

PR #2362, opened 2023-11-15
LuceneScaleTest: better support for large stores

ScottDugas:
1. An option to disable the index, or build it at the start of the test to more quickly create large stores
2. Only update the search words if doing a command that uses them, as this can take quite some time on large accounts
3. Some code to dump a timer into a json file. This isn't currently used, but I wanted it in some profiling investigations I was doing, and figured it would be worth sharing

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
@@ -442,7 +442,7 @@ public class LuceneScaleTest extends FDBRecordStoreTestBase {
 
         private DataModel() {
             if (!Config.COMMANDS_TO_RUN.contains(Command.Search) &&
-                !Config.COMMANDS_TO_RUN.contains(Command.Insert)) {
+                    !Config.COMMANDS_TO_RUN.contains(Command.Insert)) {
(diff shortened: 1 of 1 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneScaleTest.java
```diff
@@ -372,24 +436,60 @@ private class DataModel {
...
+                    if (!store.isIndexReadable(INDEX.getName())) {
+                        indexBuilder = OnlineIndexer.newBuilder()
+                                .setRecordStoreBuilder(store.asBuilder())
+                                .setTargetIndexesByName(List.of(INDEX.getName()));
+                    }
+                }
+                context.commit();
+            }
+            if (indexBuilder != null) {
+                logger.info("Building index");
```
ohadzeliger, line 475: Is this happenning synchronously, which is why it should not be done inside the transaciton?

ScottDugas: When using this for large stores, as intended, this would take way too long to happen in a single transaction.

ohadzeliger (commented):
.

ohadzeliger (dismissed):

ohadzeliger (approved):

PR #2364, opened 2023-11-15
Lucene: Store simple files as content on File Reference

### LINKED ISSUE #2284: LuceneOptimizedFieldInfosFormat breaks Lucene constraints
LuceneOptimizedFieldInfosFormat incorrectly assumes that the mapping from field names to field numbers is consistent, and that individual segments will have a subset of the global mapping.
This is not a valid assumption, as noted in this comment: https://github.com/apache/lucene/blob/d62ca4a01f3693e0c6043a48080b33d03fdee8b4/lucene/test-framework/src/java/org/apache/lucene/tests/index/MismatchedLeafReader.java#L34-L37

In my investigation, I _think_ this only happens when using `IndexWriter.addIndexes`, so if the `LuceneIndexMaintainer` avoids ever using that it might not be issue. In addition to the possibility that I missed something here, it's possible that future code will do things resulting in different maps.

### LINKED ISSUE #2295: FDBDirectory drops other segment files when .cfs is deleted
It's possible that Lucene will delete the .cfs file without deleting other segment related files (e.g. .si), but FDBDirectory might store these other files on the .cfs and just drop them when the .cfs gets deleted.
This results in a corrupted index.

ScottDugas:
Resolves #2295
Resolves #2284

This resolves two issues by storing the .fnm, .cfe and .si on the LuceneFileReference as `content`.
Previously it would store .si and .cfe directly on the reference for .cfs.
It would also try to deduplicate the .fnm information, but didn't do so correctly, because it assumed that all segments had the same name-to-number mapping. This is a much smarter de-duplication algorithm, that should, in general result in one FieldInfos file, but can support incompatible files.

As a consequence, this removes LuceneOptimizedWrappedDirectory and LuceneOptimizedSegmentInfoFormat as they were focused on special casing this behavior.

This is not a backwards compatible change, but should fail with the following if opening an older index:
```
new RecordCoreException("FileReference has old file content")
```

A note for reviewers: It might make sense to look at the commits one-at-a-time, or it might make more sense to just review the whole thing.
Commit summary:
1. My simplistic attempt to meet the requirements from the Lucene contract, and resolve the issues, but it had substantially worse performance than main.
2. This tries to take advantage of the read-your-write cache and read all FieldInfos as a range read, but that also was insufficient.
3. Change the code to reuse a "global" FieldInfos if possible, but allow for different segments to have different FieldInfos
4. Cleanup from self-review and automatic tools, and some rebase test issues.

I also had a draft PR #2296 when I was iterating over (4) but it had so many comments about failed builds I felt a new PR would be better.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -282,6 +288,69 @@ public FDBLuceneFileReference getFDBLuceneFileReference(@Nonnull final String na
...
+        byte[] key = fieldInfosSubspace.pack(id);
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_SIZE, key.length + value.length);
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_CALL);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(getLogMessage("Write lucene stored fields data",
+                    LuceneLogMessageKeys.DATA_SIZE, value.length,
+                    LuceneLogMessageKeys.ENCODED_DATA_SIZE, value.length));
+        }
+        context.ensureActive().set(key, value);
+        context.asyncToSync(
```
ScottDugas, line 329: This could probably just write to the map, instead of re-reading

ScottDugas: Or at least simplified

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1208,11 +1208,11 @@ void testSimpleUpdate(boolean primaryKeySegmentIndexEnabled) {
                         .getDirectory(Tuple.from())
                         .getPrimaryKeySegmentIndex();
                 assertEquals(List.of(
```
ScottDugas, line 1210: We should add tests that:
1. if we delete all the documents the FieldInfos space is empty
2. after a merge that the reference counting for FieldInfos is correct

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedFieldInfosFormat.java
```diff
@@ -20,36 +20,277 @@
...
+                }
+            }
+        }
+        final long id;
+        if (!canReuseGlobal) {
+            id = fdbDirectory.writeFieldInfo(protobuf.build().toByteArray());
+        } else {
+            id = FDBDirectory.GLOBAL_FIELD_INFOS_ID;
+        }
+        if (globalNeedsUpdating) {
```
ohadzeliger, line 162: It is possible, by the algorithm above, to have both `canReuseGlobal` and `globalNeedsUpdating` as `TRUE`, in which case we will write both the custom version and the revised global versions to the DB. Is that intended?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -280,6 +291,75 @@ public FDBLuceneFileReference getFDBLuceneFileReference(@Nonnull final String na
...
+        }
+        return writeFieldInfo(value, id);
+    }
+
+    public long writeFieldInfo(byte[] value, long id) {
+        byte[] key = fieldInfosSubspace.pack(id);
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_SIZE, key.length + value.length);
+        context.increment(LuceneEvents.Counts.LUCENE_WRITE_CALL);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(getLogMessage("Write lucene stored fields data",
```
ohadzeliger, line 324:
```suggestion
            LOGGER.trace(getLogMessage("Write lucene field info data",
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ScottDugas: I added the tests / assertions, and it found a bug with the delete code, where it was deleting the wrong id. Since we use the same counter to calculate ids, this wouldn't every break anything, but also means that we never deleted FieldInfos data.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedFieldInfosFormat.java
```diff
@@ -20,36 +20,277 @@
...
+ *     for more information about why we need to support incompatible mappings.
+ * </p>
  */
 public class LuceneOptimizedFieldInfosFormat extends FieldInfosFormat {
-    private final FieldInfosFormat fieldInfosFormat;
 
-    public LuceneOptimizedFieldInfosFormat(FieldInfosFormat fieldInfosFormat) {
-        this.fieldInfosFormat = fieldInfosFormat;
-    }
+    public static final String EXTENSION = "fip";
```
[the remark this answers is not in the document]
ohadzeliger: My intent was purely for maintainability purposes, so looking at a file you can tell that `fdb-si` is not a standard Lucene file extension. Having an enum is a good way to capture that since there would be a way to see them all in one place.

ohadzeliger (approved):
LGTM

ScottDugas:
The 6 code smells from sonarqube are all from older code

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -280,6 +292,83 @@ public FDBLuceneFileReference getFDBLuceneFileReference(@Nonnull final String na
...
+                LuceneEvents.Waits.WAIT_LUCENE_READ_FIELD_INFOS,
+                getAllFieldInfos().thenApply(allFieldInfos -> allFieldInfos.get(GLOBAL_FIELD_INFOS_ID)));
+    }
+
+    public long updateGlobalFieldInfos(final byte[] fieldInfos) {
+        return writeFieldInfo(fieldInfos, GLOBAL_FIELD_INFOS_ID);
+    }
+
+    @VisibleForTesting
+    public CompletableFuture<Map<Long, byte[]>> getAllFieldInfos() {
```
MMcM, line 356: Every call to this appears to immediately do a `context.asyncToSync`, so perhaps this ought to do it itself. Or this could be `getAllFieldInfosAsync` and a new method could combine with that.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ScottDugas: changed `getAllFieldInfos` to call `asyncToSync`

MMcM (approved):

PR #2368, opened 2023-11-17
Resolve #2367: Move Lucene StoredFields to K/V Store

### LINKED ISSUE #2367: Move Lucene StoredFields to K/V store
Lucene stored fields should be moved to K/V store from the file-based mechanism to assist in query latency improvements.

ohadzeliger:
Move the Lucene StoredFields to the K/V store from a file. The contents will be stored in its own subspace per segment, and a new empty file marker is created in order to tap into the data life cycle management (.fsf).
The feature is guarded by an index option and the set of options is added to the FDBDirectory to support future options accessibility.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for Lucene StoredFields implementation.
+ */
+@Tag(Tags.RequiresFDB)
+public class LuceneStoredFieldsTest extends FDBRecordStoreTestBase {
+
+    // The index to use when teh optimized stored fields option is enabled (this is the same as the base index, with the
```
ScottDugas, line 73:
```suggestion
    // The index to use when the optimized stored fields option is enabled (this is the same as the base index, with the
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+    // PK index and Optimized stored fields options enabled)
```
ScottDugas, line 74: this doesn't enable either; I suspect the comment is wrong, or supposed to be on the field below

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+    public static final Index TEXT_AND_STORED_COMPLEX_WITHOUT_OPT_STORED_FIELDS = new Index(
+            "Simple$test_stored_complex",
+            concat(function(LuceneFunctionNames.LUCENE_TEXT, field("text")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("text2")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("group")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("score")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("time")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("is_seen"))),
+            LuceneIndexTypes.LUCENE,
+            ImmutableMap.of(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED, "false"));
```
ScottDugas, line 90: should this enable the primarykey index?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.SIMPLE_TEXT_SUFFIXES;
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.TEXT_AND_STORED_COMPLEX;
```
ScottDugas, line 56: Probably worth re-declaring these here, rather than taking them from the base class, to ensure that the only difference is the name, and whether the option is enabled or not

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+    private static final Index SIMPLE_TEXT_SUFFIXES_WITHOUT_OPT_STORED_FIELDS = new Index("Simple$text_suffixes_pky",
+            function(LuceneFunctionNames.LUCENE_TEXT, field("text")),
+            LuceneIndexTypes.LUCENE,
+            ImmutableMap.of(IndexOptions.TEXT_TOKENIZER_NAME_OPTION, AllSuffixesTextTokenizer.NAME,
+                    LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED, "false"));
+
+    public static final Index TEXT_AND_STORED_COMPLEX_WITHOUT_OPT_STORED_FIELDS = new Index(
+            "Simple$test_stored_complex",
+            concat(function(LuceneFunctionNames.LUCENE_TEXT, field("text")),
+                    function(LuceneFunctionNames.LUCENE_STORED, field("text2")),
```
ScottDugas, line 84: Does this cover all the field types (that are relevant for the record layer)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+            context.commit();
+        }
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            final RecordQuery query = buildQuery("Document", Collections.emptyList(), SIMPLE_DOC);
+            queryAndAssertFields(query, "text", Set.of(), Set.of());
+            if (useOptimizedStoredFieldsFormat) {
+                try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions())) {
+                    // When deleting all docs from the index, the last segment (_1) gets removed, leaving the _0 segment with the 3 tombstones
+                    assertDocCountPerSegment(directory, List.of("_0", "_1"), List.of(3, 0));
```
ScottDugas, line 219: If you run a merge like in: https://github.com/FoundationDB/fdb-record-layer/pull/2364/files#diff-eb73467a288b09c88277b4ba28abdf2e1dc24cbae14e25808c479ed69aba6ec8R1306
You should be able to re-scan and see that there are no documents, as `_0` will be deleted.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+    @BooleanSource
+    void testRollbackInsert(boolean useOptimizedStoredFieldsFormat) throws Exception {
+        Index index = getSimpleTextIndex(useOptimizedStoredFieldsFormat);
+
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            recordStore.saveRecord(createSimpleDocument(1623L, "Document 1", 2));
+            recordStore.saveRecord(createSimpleDocument(1624L, "Document 2", 2));
+            recordStore.saveRecord(createSimpleDocument(1547L, "NonDocument 3", 2));
+            context.ensureActive().cancel();
```
ScottDugas, line 236: Is the point of this to future proof against potential caching that could leak between transactions, or is it to test that a non-committed transaction doesn't persist anything. I think if it is the latter, that this test can probably just be removed.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+                try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions())) {
+                    assertDocCountPerSegment(directory, List.of("_0"), List.of(0));
+                }
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testRollbackUpdate(boolean useOptimizedStoredFieldsFormat) throws Exception {
```
ScottDugas, line 252: Same comment here as in `testRollbackInsert`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+                try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions())) {
+                    assertDocCountPerSegment(directory, List.of("_0", "_1"), List.of(3, 0));
+                }
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testRollbackDeleteAllDocuments(boolean useOptimizedStoredFieldsFormat) throws Exception {
```
ScottDugas, line 281: Same comment here as in `testRollbackInsert`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1623L, "Hello", "Hello 2", 5, 12, false, 7.123));
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1624L, "Hello record", "Hello record 2", 6, 13, false, 8.123));
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1625L, "Hello record layer", "Hello record layer 2", 7, 14, true, 9.123));
+            context.commit();
+        }
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            final RecordQuery query = buildQuery("record", Collections.emptyList(), COMPLEX_DOC);
+            // This doc has the OK as (Group, docID)
```
ScottDugas, line 328: What is this comment saying?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+                .setRecordType(docType)
+                .setFilter(filter)
+                .build();
+    }
+
+    private void queryAndAssertFields(RecordQuery query, String fieldName, Set<Long> expectedPrimaryKeys, Set<?> expectedFields) throws Exception {
+        queryAndAssertFieldsT(query, fieldName, expectedPrimaryKeys.stream().map(Tuple::from).collect(Collectors.toSet()), expectedFields);
+    }
+
+    private void queryAndAssertFieldsT(RecordQuery query, String fieldName, Set<Tuple> expectedPrimaryKeys, Set<?> expectedFields) throws Exception {
```
ScottDugas, line 359:
```suggestion
    private void queryAndAssertFieldsTuples(RecordQuery query, String fieldName, Set<Tuple> expectedPrimaryKeys, Set<?> expectedFields) throws Exception {
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+        RecordQueryPlan plan = planner.plan(query);
+        try (RecordCursor<FDBQueriedRecord<Message>> fdbQueriedRecordRecordCursor = recordStore.executeQuery(plan)) {
+            List<FDBQueriedRecord<Message>> result = fdbQueriedRecordRecordCursor.asList().get();
+            Set<Tuple> actualKeyTuples = result.stream().map(FDBQueriedRecord::getIndexEntry).map(IndexEntry::getPrimaryKey).collect(Collectors.toSet());
+            assertEquals(expectedPrimaryKeys, actualKeyTuples);
+            Set<?> fields = result.stream()
+                    .map(FDBQueriedRecord::getStoredRecord)
+                    .filter(Objects::nonNull) //unnecessary, but helps keep the linter happy
```
ScottDugas, line 367: Should this throw if it finds `null`?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+                    .map(FDBStoredRecord::getRecord)
+                    .map(m -> m.getField(m.getDescriptorForType().findFieldByName(fieldName)))
+                    .collect(Collectors.toSet());
+            assertEquals(expectedFields, fields);
```
ScottDugas, line 371: Since you are converting these to sets, you aren't actually asserting that the field values correspond to the correct documents. i.e. I think all of your tests would pass if the doc ids were randomly mixed up. It would be better to have the method take a `Map<Tuple, ?> expected values` and then assert that the `keySet` is the same as the set of primary keys, and that each one has the right value.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,395 @@
...
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, SIMPLE_DOC, index);
+            final RecordQuery query = buildQuery("Document", Collections.emptyList(), SIMPLE_DOC);
+            queryAndAssertFields(query, "text", Set.of(1624L), Set.of("Document 2"));
+        }
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testInsertUpdateDocuments(boolean useOptimizedStoredFieldsFormat) throws Exception {
```
ScottDugas, line 171: It would be good to have a test of actually doing merges to make sure the right stuff ends up in the right place.
I think you should be able to simply use 3 transactions for saving the 3 documents, and then assert about the results, and assert that there is only one segment, and it has 3 documents.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormat.java
```diff
@@ -36,85 +36,59 @@
...
+    @Override
     public StoredFieldsWriter fieldsWriter(final Directory directory, final SegmentInfo si, final IOContext context) throws IOException {
-        @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = ((FDBDirectory)FilterDirectory.unwrap(directory)).getPrimaryKeySegmentIndex();
-        final StoredFieldsWriter storedFieldsWriter = storedFieldsFormat.fieldsWriter(directory, si, context);
+        FDBDirectory fdbDirectory = toFdbDirectory(directory);
+        @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = fdbDirectory.getPrimaryKeySegmentIndex();
+        StoredFieldsWriter storedFieldsWriter;
+
+        // Use TRUE as the default OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED option
+        if (fdbDirectory.getBooleanIndexOption(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED, false)) {
```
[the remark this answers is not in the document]
ohadzeliger: Right - that's incorrect, fixed the comment.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormat.java
```diff
@@ -36,85 +36,59 @@
...
- * This class provides a Lazy reader implementation to limit the amount of
+ * This class provides a custom KeyValue based reader and writer implementation to limit the amount of
  * data needed to be read from FDB.
  *
  */
 public class LuceneOptimizedStoredFieldsFormat extends StoredFieldsFormat {
-
-    private StoredFieldsFormat storedFieldsFormat;
+    public static final String STORED_FIELDS_EXTENSION = "fsf";
+    private final StoredFieldsFormat storedFieldsFormat;
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormat.java
```diff
@@ -36,85 +36,59 @@
...
-            if (LuceneOptimizedPostingsFormat.allowCheckDataIntegrity) {
-                storedFieldsReader.get().checkIntegrity();
-            }
-        }
-
-        @Override
-        public void close() throws IOException {
-            storedFieldsReader.close();
+    @SuppressWarnings("PMD.CloseResource")
+    private FDBDirectory toFdbDirectory(Directory directory) {
```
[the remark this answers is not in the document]
ohadzeliger: Created a radar for followup items: rdar://118941551 (StoredFields PR followup)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsReader.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void visitDocument(final int docID, final StoredFieldVisitor visitor) throws IOException {
+        LuceneStoredFieldsProto.LuceneStoredFields storedFields;
+        if (rangeIterator == null) {
+            storedFields = LuceneStoredFieldsProto.LuceneStoredFields.parseFrom(directory.readStoredFields(segmentName, docID));
+        } else {
+            KeyValue keyValue = rangeIterator.next();
+            if (keyValue == null) {
```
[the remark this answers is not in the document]
ohadzeliger: The range read was a leftover from the scratch PR (From John) that was used for the merge reads. It was not in use anymore, so I removed it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsReader.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public void close() throws IOException {
+        if (LOG.isTraceEnabled()) {
+            LOG.trace("close");
+        }
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        return 1024; // TODO Estimate
```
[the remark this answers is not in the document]
ohadzeliger: Added to followup radar

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsWriter.java
```diff
@@ -0,0 +1,195 @@
...
+    private final String segmentName;
+    private int docId;
+
+    @SuppressWarnings("PMD.CloseResource")
+    public LuceneOptimizedStoredFieldsWriter(final FDBDirectory directory, final SegmentInfo si, IOContext context) {
+        this.directory = directory;
+        this.docId = 0;
+        this.segmentName = si.name;
+        // Create a "dummy" file to tap into the lifecycle management (e.g. be notified when to delete the data)
+        this.directory.createOutput(IndexFileNames.segmentFileName(si.name, "", LuceneOptimizedStoredFieldsFormat.STORED_FIELDS_EXTENSION), context);
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsWriter.java
```diff
@@ -0,0 +1,195 @@
...
+        return docCount;
+    }
+
+    @Override
+    public void close() throws IOException {
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        return 1; // ToDo Fix
```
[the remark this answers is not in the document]
ohadzeliger: Added to followup

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsReader.java
```diff
@@ -0,0 +1,145 @@
...
+        this.segmentName = si.name;
+    }
+
+    @Override
+    public void visitDocument(final int docID, final StoredFieldVisitor visitor) throws IOException {
+        LuceneStoredFieldsProto.LuceneStoredFields storedFields;
+        if (rangeIterator == null) {
+            storedFields = LuceneStoredFieldsProto.LuceneStoredFields.parseFrom(directory.readStoredFields(segmentName, docID));
+        } else {
+            KeyValue keyValue = rangeIterator.next();
```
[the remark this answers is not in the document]
ohadzeliger: Yeah, it was a leftover from before, removed for now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -400,6 +418,39 @@ public int writeSchema(@Nonnull List<Long> bitSetWords, @Nonnull final byte[] va
...
+    /**
+     * Delete stored fields data from the DB.
+     * @param segmentName the segment name to delete the fields from (all docs in the segment will be deleted)
+     */
+    public void deleteStoredFields(@Nonnull final String segmentName) {
+        byte[] key = storedFieldsSubspace.pack(Tuple.from(segmentName));
+        context.increment(LuceneEvents.Counts.LUCENE_DELETE_STORED_FIELDS);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace(getLogMessage("Delete Stored Fields Data",
+                    LuceneLogMessageKeys.RESOURCE, key));
```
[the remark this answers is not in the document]
ohadzeliger: That makes sense, done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -624,6 +691,23 @@ private boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference>
         context.ensureActive().clear(metaSubspace.pack(name));
         // Nothing stored here currently.
         context.ensureActive().clear(dataSubspace.subspace(Tuple.from(value.getId())).range());
+        // Delete K/V data: If the deferredDelete flag is on then delete all content from K/V subspace for the segment
+        // (this is to support CFS deletion, that will be disjoint from the actual file deletion).
+        // Otherwise, delete the data for the specific file immediately.
+        String segmentName = IndexFileNames.parseSegmentName(name);
+        if (deferDeleteToCompoundFile) {
```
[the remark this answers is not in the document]
ohadzeliger: I don't quite follow that one - if the CFS was deleted because this is an abort, then should there be any data saved at all?

fdb-record-layer-lucene/src/main/proto/lucene_stored_fields.proto
```diff
@@ -0,0 +1,38 @@
+/*
+ * lucene_file_system.proto
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Yes, the comment was wrong, fixed.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: No - the PK index is enabled and disabled through the PK-index specific tests. Since the desired state is to have the stored fields option enabled, I opted to have the base index definition as `TRUE` so that all future tests are including it by default. The PK index option is defaulted to `FALSE` so the only way they are tested as both `TRUE` is in the PK index tests.
I think it may be worthwhile to invert the PK index option on the base indexes, later.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: We can, but then we lose the connection to the base definition, so for example, if someone adds a new option to it, we won't see it. I thought it would be better to reuse the common index definition in this case (though maybe it would be good to come up with a way to extend the index definition somehow).

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: I think so:  string, long, short, boolean, double. Am I missing something?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: I'm not seeing the diff. Do you mean like:
```
        try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
                 .setRecordStore(recordStore)
                 .setIndex(index)
                 .build()) {
             indexBuilder.mergeIndex();
         }
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Should be `PK` instead of `OK` - fixed.
This is just to let people know where the values in the assertions are coming from

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Changed to `map(Objects::requireNonNull)`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ScottDugas: arbitrary bytes?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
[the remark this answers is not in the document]
ScottDugas: I'd have to go back to that and investigate what was happening. I _suspect_ that it will only delete the cfs in situations where it will error, and thus not commit, but I'm not sure. I don't think we need to block this change here, but something we'll need to validate.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ScottDugas: Yes, that is what it was linking to, the lines that run the merge.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ScottDugas: If I had to guess what I was thinking back then.... I suspect it was to make sure we don't have cached information leaking between transactions, but since we don't cache anything, I don't think there's any risk here.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -2695,26 +2695,33 @@ void manySegmentsParallelOpen(boolean primaryKeySegmentIndexEnabled) {
             assertEquals(20,
                     recordStore.scanIndex(index, fullTextSearch(index, "Vision"), null, ScanProperties.FORWARD_SCAN).getCount().join());
             try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions())) {
-                assertEquals(21, directory.listAll().length);
+                assertEquals(61, directory.listAll().length);
```
ScottDugas, line 2698: Do you know why this went up with your PR?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -2726,7 +2733,7 @@ private static void assertOnlyCompoundFileExisting(Index index, @Nonnull Subspac
                     || file.endsWith(".pky")) {
                 assertFalse(fileReference.getContent().isEmpty(), "fileName=" + file);
             } else {
-                assertTrue(FDBDirectory.isCompoundFile(file) || file.startsWith(IndexFileNames.SEGMENTS),
+                assertTrue(FDBDirectory.isCompoundFile(file) || file.startsWith(IndexFileNames.SEGMENTS) || FDBDirectory.isStoredFieldsFile(file),
```
ScottDugas, line 2736: Should this be around? Shouldn't the `fsf` be deleted when it is added to the `cfs`?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
```diff
@@ -502,8 +502,19 @@ private TestRecordsTextProto.MapDocument multiEntryMapDoc(long id, String text,
 
     private String[] listFiles(Index index) {
         try (FDBRecordContext context = openContext()) {
-            final FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions());
-            return directory.listAll();
+            if (index.getRootExpression() instanceof GroupingKeyExpression) {
+                List<String> files = new ArrayList<>();
+                for (int i = 0; i < 2; i++) {
```
ScottDugas, line 507: This won't necessarily catch all groups, right?
As a followup we should probably align this more with what we do in `OnlineIndexer.mergeIndex` where we gather all the groups and list them.
It might also make sense to return a `Map<Tuple, String[]>`, returning the results of each `listAll` associated with the group tuple.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: I tried running this after multiple commits and it didn't reduce the segment count. Added a test with a TODO to show the issue. I haven't looked at the other tests that used this snippet to see what is happening though.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Deleted tests

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Deleted tests

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: Added similar test, though the segments do not merge.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ohadzeliger: Hmm. This is the right value, though I don't know why it shows up as a diff. 
Must be something funny with the merge.
I'm going to try merging again as it shows conflicts anyway.

ScottDugas: IT changed on main, and showed up because of the merge.

ScottDugas:
`:fdb-record-layer-lucene-pb3:checkstyleTest` failed

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ohadzeliger: The `fsf` can be found here from another (e.g. future) segment. The `fsf` is created early enough in the process such that the next segments' one is already here, so added the segment name to the method to ensure that the `fsf` from the verified segment is not around

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -2732,8 +2735,14 @@ private static void validateIndexIntegrity(Index index, @Nonnull Subspace subspa
             if (FDBDirectory.isEntriesFile(file) || FDBDirectory.isSegmentInfo(file) || FDBDirectory.isFieldInfoFile(file)
                     || file.endsWith(".pky")) {
                 assertFalse(fileReference.getContent().isEmpty(), "fileName=" + file);
+            } else if (FDBDirectory.isStoredFieldsFile(file) && (segmentName != null)) {
+                // It is OK for a stored fields info for the next segment to be created at this point.
```
ScottDugas, line 2739: This seems somewhat supsicious, I'm not sure why this would be.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -0,0 +1,375 @@
...
+    @BooleanSource
+    void testComplexDocManyFields(boolean useOptimizedStoredFieldsFormat) throws Exception {
+        // Use a complex index with several fields
+        Index index = getComplexTextIndex(useOptimizedStoredFieldsFormat);
+
+        try (FDBRecordContext context = openContext()) {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1623L, "Hello", "Hello 2", 5, 12, false, 7.123));
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1624L, "Hello record", "Hello record 2", 6, 13, false, 8.123));
+            recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1625L, "Hello record layer", "Hello record layer 2", 7, 14, true, 9.123));
```
ScottDugas, line 274: I think this is still missing arbitrary bytes

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ScottDugas: We should probably followup to make sure this assertion covers all groups.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ohadzeliger: There is one test in particular that causes (simpleInsertDeleteAndSearch) `_1.fsf` to be present when validating the `_0` segment.

ScottDugas: ok, I think I might have a very vague understanding of how that could happen....

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenOnlineIndexingTest.java
ohadzeliger: I did not make this change - this is part of the PR you merged yesterday, and is likely some PR issue that confused the diff.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
ohadzeliger: `bytes` support will be added later - will add to followup

ScottDugas (approved):

PR #2379, opened 2023-11-30
Fixes #2377: NPE during maintenance of synthetic indexes if some types have only disabled indexes

### LINKED ISSUE #2377: NPE during maintenance of synthetic indexes if some types have only disabled indexes
There's a potential `NullPointerException` that can arise on this line: https://github.com/FoundationDB/fdb-record-layer/blob/f5b8fb10b266499b7e6a5e43aa97cde17e48482a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java#L784

The scenario is this. There are multiple synthetic record types that all contain a given stored type `T`.  Suppose that each of them have indexes defined on them, but that some of them have only disabled indexes while some of them have at least one non-disabled index. When we get a record of type `T`, we need to go and update all of the relevant synthetic indexes. We do that by:

1. All synthetic types using that type are identified, and then any without any indexes defined on them are filtered out.
2. Plans are constructed for each type to construct all of the synthetic records (e.g., looking up join partners).
3. For each type, we construct a map linking the type to a set of index maintainers, filtering out any disabled index maintainers. Crucially, we _remove any entries from that map if all indexes are disabled_.
4. If our type map is empty, we exit early, not executing any plans.
5. Now, we execute each plan from step 2 to construct the synthetic records.
6. For each synthetic record, we grab the set of relevant index maintainers from the map constructed in step 3. However, if all indexes are disabled, then the entry will have been removed, hence the NPE.

There are a few ways to fix this, but I think the best is to adjust the logic in step 1 so that we only select synthetic records with non-disabled indexes. Then we no longer need to remove any empty entry from the map in step 3 (as there shouldn't be any) and the check on step 4 should be redundant. This should also save us some planning time, as we'll no longer try and construct plans for synthetic types that we end up not executing.

alecgrieser:
Now, when we go to index a record of a synthetic type, we only go through the process of generating a plan if there is at least one non-disabled index defined on it. Previously, we would generate the plan if there was any index defined on it, and then we would filter out any types without any non-disabled indexes later. The logic for that later filtering could result in a null pointer exception, though, so this updated approach aligns what gets planned with what is actually necessary to execute. This also can save on planning time if there are disabled indexes.

This fixes #2377.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
@@ -1217,10 +1217,6 @@ public class SyntheticRecordPlannerTest {
                     .setOtherRecNo(1415L)
                     .setStrValue("foo")
                     .build();
-            TestRecordsJoinIndexProto.MyOtherRecord otherRecord = TestRecordsJoinIndexProto.MyOtherRecord.newBuilder()
-                    .setRecNo(1415L)
-                    .setNumValue3(42)
-                    .build();
             recordStore.saveRecord(simpleRecord);
             assertEquals(2L, timer.getCount(FDBStoreTimer.Counts.PLAN_SYNTHETIC_TYPE));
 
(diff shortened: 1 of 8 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/synthetic/SyntheticRecordPlannerTest.java
```diff
@@ -1229,7 +1403,7 @@ void updateSyntheticIndexesWhenDisabled() throws Exception {
     }
 
     @Test
-    void updateSyntheticIndexesWhenWriteOnly() throws Exception {
+    void updateRecordWhenSyntheticIndexesIsWriteOnly() throws Exception {
```
alecgrieser, line 1406: These last two tests had very similar names. I've renamed them to make their difference more apparent (namely that one is about record updates and the other about record inserts). Other than adding assertions about the number of synthetic type plans, the tests are unchanged.

pengpeng-lu (approved):

Issue #2380, opened 2023-11-30
AUTO-COMPLETE does not complete phrases which contain stop words in them


qdnguyen0:
Simple reproducing test case (using the syntax in LuceneIndexTest.java inside of RecordLayer):

        recordStore.saveRecord(createSimpleDocument(1623L, "united states of america", 1));
        recordStore.saveRecord(createSimpleDocument(1624L, "welcome to the united states of america", 1));
        recordStore.saveRecord(createSimpleDocument(1625L, "united kingdom, france, the states", 1));
        recordStore.saveRecord(createSimpleDocument(1626L, "The countries are united kingdom, france, the states", 1));
        recordStore.saveRecord(createSimpleDocument(1627L, "states united as a country", 1));
        recordStore.saveRecord(createSimpleDocument(1628L, "all the states united as a country", 1));
        recordStore.saveRecord(createSimpleDocument(1629L, "states have been united as a country", 1));
recordStore.saveRecord(createSimpleDocument(1630L, "all the states have been united as a country", 1));
recordStore.saveRecord(createSimpleDocument(1631L, "united states is a country in the continent of america", 1));

queryAndAssertAutoCompleteSuggestionsReturned(index,
                    storedFields,
                    "text",
                    "\"united states of america\””,
                    ImmutableList.of("united states of america",
                            "welcome to the united states of america"));

However, this does not return any results.

MMcM:
Fixed by #2381.

PR #2381, opened 2023-11-30
AUTO-COMPLETE does not complete phrases which contain stop words in them #2380

qdnguyen0:
**Motivation**
To auto-complete phrases with stop words in them. Example: "united states of ameri" would need to match "United States of America" with "of" being indexed as a stopword

**Modification** 
Use Lucene parser when parsing the search input into query to preserve the positions of the stopwords and turn them into Lucene gaps for matching.
Plus, some more enhancement to matching when the phrase ends with a stopword

**Results**
The example mentioned works.


qdnguyen0:
test this please

normen662 (dismissed):
The PR makes it sound like it's meant to leverage from us indexing stop words which may not be something we would like to do. I ll leave it up to @MMcM to comment on that. The plan/rec-layer-specific stuff looks fine, I am not an expert on lucene API so cannot comment on that. Feel free to ignore my review if Mike deems stop-word indexing doable.

qdnguyen0:
> The PR makes it sound like it's meant to leverage from us indexing stop words which may not be something we would like to do

No, we're not leverage the indexing of stopwords. We do take into account the stopwords when saving the position information when indexing though, so this PR takes advantage of the position with stopwords taken into account in the query in order to phrase-match better.

MMcM (approved):

PR #2382, opened 2023-12-01
Resolves #71376: hasLosslessNormalization is too conservative

MMcM:
Add more specific implementations of hasLosslessNormalization.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/ThenKeyExpression.java
```diff
@@ -201,6 +201,16 @@ public List<KeyExpression> normalizeKeyForPositions() {
         return list;
     }
 
+    @Override
+    public boolean hasLosslessNormalization() {
```
normen662, line 205: So for a `Then` we would expect it to be lossless if all children are lossless, so for instance a `Then` we would be ok with `FanOut`s for each child as long as those children are single extraction points looking at `NestingKeyExpression` (i.e. lossless themselves).

MMcM:
Missing #2809 added.

PR #2386, opened 2023-12-01
Resolves #2385: Integrate Lucene test framework for Stored Fields

### LINKED ISSUE #2385: Utilize Lucene's test framework to add coverage
Lucene provides a test framework to validate directory/codec implementations. It would be good to utilize this to add further confidence to the lucene implementation in the record layer.

As a starter, we could extend: https://github.com/apache/lucene/blob/main/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseStoredFieldsFormatTestCase.java to validate the behavior of #2367

ScottDugas:
This integrates an implementation of BaseStoredFieldsFormatTestCase
I limited it to a single seed to prevent flakiness in PRB.
I've also incorporated a few small changes required to get the tests
to pass for the given seed:
1. StoredFields was not unwrapping in a way that supported the
   MockDirectoryWrapper that the tests use, so I extracted the
   unwrapping into FDBDirectoryUtils. One usage of
   `FilteredDirectory.unwrap` is still used to find out whether
   FieldInfos is being used in a CompoundReader.
2. `FDBDirectory.readStoredFields` will throw a helpful message
   if there is no bytes for the segment/docID. This required making
   LuceneOptimizedCompoundReader public, which in turn required
   cleaning up its javadocs.

To support the test code, I also extracted some methods, and changed
some things to be public/protected.

I also fixed one of the configurations in idea's codestyle to align with our checkstyle checks.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/TestFDBDirectory.java
```diff
@@ -0,0 +1,252 @@
...
+    private static boolean fullBufferToSurviveDeletes;
+    private static boolean allowAddIndexes;
+
+    private static AtomicReference<Pair<String, FieldInfos>> previousFieldInfos = new AtomicReference<>();
+    private static AtomicReference<Pair<String, Map<Long, byte[]>>> previousStoredFields = new AtomicReference<>();
+
+    public TestFDBDirectory() {
+        super(new Subspace(Tuple.from("record-test", "unit", "lucene")),
+                FDBDatabaseFactory.instance().getDatabase().openContext(),
+                Map.of(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED, "true"));
```
ScottDugas, line 69: You may note that this does not include `PRIMARY_KEY_SEGMENT_INDEX_ENABLED`, that code broke with the tests because it tries to cast the `StoredFieldsReader`, but the reader is wrapped by the test framework.
https://github.com/FoundationDB/fdb-record-layer/blob/a1e93d2dc293877c8bb200783f5c5a81ef6ead1d/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java#L219

We may want to change the way that code works to work with the tests, but I think it makes sense to try and get the test framework in and then work to test it with more situations.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/TestFDBDirectory.java
```diff
@@ -0,0 +1,252 @@
...
+     * <p>
+     * {@link IndexWriter#addIndexes} is used to copy segments from one directory to another. Under the hood this
+     * uses {@link org.apache.lucene.store.Directory#copyFrom} to copy the data, which we could override, but
+     * {@link org.apache.lucene.store.MockDirectoryWrapper} explicitly does not call into the wrapped function, so
+     * there is no clean way for us to copy non-file data, instead it will always just copy the files. This looks
+     * to be intentional (<a href="https://issues.apache.org/jira/browse/LUCENE-6299">LUCENE-6299</a> /
+     * <a
+     * href="https://github.com/apache/lucene/commit/4e53fae38e993fad0c8a5466e1e0c6121d49ad5f#diff-6c29e90918d4bab6eab63cb94f9c0ae55fb14bbbb07446ee0568bb33343947f7">
+     * commit</a>.
+     * The code that is run when this is enabled analyzes the stack trace and copies the data that is not stored in
```
ohadzeliger, line 115: Cool solution!

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/TestingCodec.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+@AutoService(Codec.class)
+public class TestingCodec extends Codec {
+    private LuceneOptimizedCodec underlying;
+    private static boolean disableLaziness;
+    private static boolean disableLazinessForLiveDocs;
+
+    public TestingCodec() {
+        super("RLT");
+        underlying = new LuceneOptimizedCodec(Lucene87Codec.Mode.BEST_SPEED);
```
ohadzeliger, line 68: Use the no-param constructor so as to keep up with any production code changes.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/TestingCodec.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Override
+    public StoredFieldsFormat storedFieldsFormat() {
+        final StoredFieldsFormat storedFieldsFormat = underlying.storedFieldsFormat();
+        if (disableLaziness) {
+            return new StoredFieldsFormat() {
+                @Override
+                public StoredFieldsReader fieldsReader(final Directory directory, final SegmentInfo si, final FieldInfos fn, final IOContext context) throws IOException {
+                    final StoredFieldsReader storedFieldsReader = storedFieldsFormat.fieldsReader(directory, si, fn, context);
+                    storedFieldsReader.ramBytesUsed();
```
ohadzeliger, line 124: This would only work for the "traditional" reader, right? The Optimized one will not be affected by `ramBytesUsed` call. On the other hand, maybe its lazy initialization is not problematic?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/TestFDBDirectory.java
ScottDugas: https://github.com/FoundationDB/fdb-record-layer/issues/2407

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
```diff
@@ -0,0 +1,432 @@
...
+        }
+    }
+
+    @Override
+    public void testMismatchedFields() throws Exception {
+        TestFDBDirectory.setAllowAddIndexes(true);
+        try {
+            super.testMismatchedFields();
+        } finally {
+            TestFDBDirectory.setAllowAddIndexes(false);
```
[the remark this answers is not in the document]
ScottDugas: Actually, these should always have the same starting state, so instead I added a `reset()` method to `TestFDBDirectory` and `TestingCodec` and call it in a Setup method, much cleaner.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -527,7 +528,7 @@ public byte[] readStoredFields(String segmentName, int docId) throws IOException
         final byte[] key = storedFieldsSubspace.pack(Tuple.from(segmentName, docId));
         final byte[] rawBytes = context.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_GET_STORED_FIELDS,
                 context.instrument(LuceneEvents.Events.LUCENE_READ_STORED_FIELDS,
-                        context.ensureActive().get(key)));
+                        agilityContext.get(key)));
```
ohadzeliger, line 531: Why this change? Is this something that needs to be applied over all the transaction access in the directory?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
```diff
@@ -117,48 +120,39 @@ private static boolean isUsingFDBDirectory() {
         return System.getProperty("tests.directory", "random").equals(TestFDBDirectory.class.getName());
     }
 
+    @Override
```
ohadzeliger, line 123: Since these tests change some global settings, would it make sense to have `@AfterAl` added to the end as well?

ohadzeliger (commented):
Just a couple of questions

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: Had a discussion with @jjezra and it looks like this should always the case - maybe the original Context should be removed?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
ScottDugas: Good idea, but added an override to `tearDown` instead of an `@AfterAll`.

ohadzeliger (approved):

ohadzeliger (approved):

PR #2391, opened 2023-12-05
RESOLVES #2390: Add Lucene Partitioning Metadata

### LINKED ISSUE #2390: Add Lucene Partitioning Metadata
Certain datasets may grow into hundreds of millions of records. A single Lucene index would not support such a large number of documents. A new feature creates multiple physical Lucene indexes (partitions) and spreads the documents across them, based on a timestamp associated with each document.

This first phase of this feature introduces the metadata needed to support a partitioned Lucene index.

wshaib-apple:
This initial update adds support for partitioning a logical Lucene index into multiple physical indexes spanning a time range. Each physical Lucene index (hereafter referred to as `partition`) will store documents whose timestamps fall within the time range it covers. Partitions do *not* have overlapping time ranges. A document's timestamp, for the purposes of partitioning, is indicated by a record's field name in the index's options. The following options are added to the Lucene index:
- `textDocumentPartitionTimestamp`: the name of the field to use as timestamp for partitioning. If this option isn't specified, the index is not partitioned.
- `textDocumentPartitionHiWatermark`: high watermark size for a lucene partition, beyond which the partition would be split, or a new partition would be created. If not specified, a server default will be applied (TODO).
- `textDocumentPartitionLoWatermark`: low watermark size for a lucene partition, below which the partition would be a candidate for merge with another adjacent partition. If not specified, a server default will be applied (TODO).

The new partitions will be laid out in subspaces exactly like the current implementation, however they will be scoped as follows:
- `indexSubspace.subspace(Tuple.from(groupingKey, 1, partitionMinTimestamp))` where `partitionMinTimestamp ` is a long that has the minimum document timestamp stored in the partition.

Partition metadata will be stored in FDB as (k,v):
- k: `indexSubspace.subspace(Tuple.from(groupingKey, 0, partitionMinTimestamp)`
- v: a protobuf message containing `partitionId`, `from` timestamp, `to` timestamp, and document `count`

This update *does not* (yet) support
- splitting or merging partitions
- index (segment) merging across multiple partitions
- query continuation
- optimized query based on timestamp predicate (if provided)

This update sends all queries (including spellchecks) to the most recent partition.

**If a Lucene index is not partitioned, the existing behavior and db schema are not affected**

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
@@ -113,6 +113,7 @@ public enum LogMessageKeys {
     INDEX_MERGES_LIMIT,
     INDEX_MERGES_FOUND,
     INDEX_MERGES_TRIED,
+    INDEX_MERGE_LOCK,
     VALUE_KEY,
     PRIMARY_KEY,
     VALUE,
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
@@ -306,6 +306,10 @@ public enum LogMessageKeys {
     HASH_FUNCTION,
     HASH,
 
+    // Lucene
+    PARTITION_ID,
+    PARTITIONING_TIMESTAMP,
+
     // Record context properties
     PROPERTY_NAME,
     PROPERTY_TYPE,
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
@@ -213,6 +213,25 @@ public class IndexOptions {
      */
     public static final String RTREE_USE_NODE_SLOT_INDEX = "rtreeUseNodeSlotIndex";
 
+    /**
+     * Option to designate a record field as the timestamp by which the corresponding document would be partitioned in
+     * Lucene.
+     */
+    public static final String TEXT_DOCUMENT_PARTITION_TIMESTAMP = "textDocumentPartitionTimestamp";
+
+    /**
+     * Option to set high watermark size for a lucene partition, beyond which the partition would be split, or a new
+     * partition would be created.
+     */
+    public static final String TEXT_DOCUMENT_PARTITIION_HI_WATERMARK = "textDocumentParititionHiWatermark";
+
+    /**
+     * Option to set low watermark size for a lucene partition, below which the partition would be a candidate for
+     * merge with another adjacent partition.
+     */
+
+    public static final String TEXT_DOCUMENT_PARTITIION_LO_WATERMARK = "textDocumentParititionLoWatermark";
+
     private IndexOptions() {
     }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
@@ -223,14 +223,14 @@ public class IndexOptions {
      * Option to set high watermark size for a lucene partition, beyond which the partition would be split, or a new
      * partition would be created.
      */
-    public static final String TEXT_DOCUMENT_PARTITIION_HI_WATERMARK = "textDocumentParititionHiWatermark";
+    public static final String TEXT_DOCUMENT_PARTITIION_HI_WATERMARK = "textDocumentPartitionHiWatermark";
 
     /**
      * Option to set low watermark size for a lucene partition, below which the partition would be a candidate for
      * merge with another adjacent partition.
      */
 
-    public static final String TEXT_DOCUMENT_PARTITIION_LO_WATERMARK = "textDocumentParititionLoWatermark";
+    public static final String TEXT_DOCUMENT_PARTITIION_LO_WATERMARK = "textDocumentPartitionLoWatermark";
 
     private IndexOptions() {
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
@@ -213,25 +213,6 @@ public class IndexOptions {
      */
     public static final String RTREE_USE_NODE_SLOT_INDEX = "rtreeUseNodeSlotIndex";
 
-    /**
-     * Option to designate a record field as the timestamp by which the corresponding document would be partitioned in
-     * Lucene.
-     */
-    public static final String TEXT_DOCUMENT_PARTITION_TIMESTAMP = "textDocumentPartitionTimestamp";
-
-    /**
-     * Option to set high watermark size for a lucene partition, beyond which the partition would be split, or a new
-     * partition would be created.
-     */
-    public static final String TEXT_DOCUMENT_PARTITIION_HI_WATERMARK = "textDocumentPartitionHiWatermark";
-
-    /**
-     * Option to set low watermark size for a lucene partition, below which the partition would be a candidate for
-     * merge with another adjacent partition.
-     */
-
-    public static final String TEXT_DOCUMENT_PARTITIION_LO_WATERMARK = "textDocumentPartitionLoWatermark";
-
     private IndexOptions() {
     }
 }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -373,7 +373,6 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
      * @param newRecord record to be written
      * @param groupingKey grouping key
      * @return partition id
-     * @param <M> Message
      */
     private <M extends Message> Integer addWrittenDocumentToPartition(@Nullable FDBIndexableRecord<M> newRecord, @Nonnull Tuple groupingKey) {
         if (newRecord == null || !partitioner.isPartitioningEnabled()) {
@@ -397,7 +396,6 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
      * @param oldRecord record to be deleted
      * @param groupingKey grouping key
      * @return partition id (or null if partitioning isn't enabled)
-     * @param <M> Message
      */
     private <M extends Message> Integer removeDeletedDocumentFromPartition(@Nullable FDBIndexableRecord<M> oldRecord, @Nonnull Tuple groupingKey) {
         if (oldRecord == null || !partitioner.isPartitioningEnabled()) {
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -347,8 +347,9 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
                 // get partition timestamp value from record
                 newRecordTimestamp = getPartitioningTimestampValue(partitioner.getPartitionTimestampFieldName(), newRecord);
                 oldRecordTimestamp = getPartitioningTimestampValue(partitioner.getPartitionTimestampFieldName(), oldRecord);
+
                 if (newRecordTimestamp == null) {
-                    throw new RuntimeException("error getting partitioning timestamp: " + partitioner.getPartitionTimestampFieldName());
+                    throw new RecordCoreArgumentException("error getting partitioning timestamp: " + partitioner.getPartitionTimestampFieldName());
                 }
             }
 
@@ -396,7 +397,28 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
             return null;
         }
 
-        for (Map.Entry<Descriptors.FieldDescriptor, Object> field : record.getRecord().getAllFields().entrySet()) {
+        int dotIndex = fieldName.indexOf('.');
+        String containingTypeName = null;
+        String actualFieldName = fieldName;
+        if (dotIndex >= 0) {
+            // nested
+            containingTypeName = fieldName.substring(0, dotIndex);
+            actualFieldName = fieldName.substring(dotIndex + 1);
+        }
+        if (containingTypeName != null) {
+            for (Map.Entry<Descriptors.FieldDescriptor, Object> field : record.getRecord().getAllFields().entrySet()) {
+                if (containingTypeName.equalsIgnoreCase(field.getKey().getName()) && field.getValue() instanceof Message) {
+                    return getPartitioningTimestampValue(actualFieldName, ((Message) field.getValue()).getAllFields());
+                }
+            }
+        } else {
+            return getPartitioningTimestampValue(actualFieldName, record.getRecord().getAllFields());
+        }
+        return null;
+    }
+
+    private Long getPartitioningTimestampValue(@Nonnull String fieldName, Map<Descriptors.FieldDescriptor, Object> fields) {
+        for (Map.Entry<Descriptors.FieldDescriptor, Object> field : fields.entrySet()) {
             if (fieldName.equalsIgnoreCase(field.getKey().getName())) {
                 if (field.getValue() instanceof Long) {
                     return (Long)field.getValue();
@@ -406,7 +428,6 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
                 }
             }
         }
-
         return null;
     }
 
@@ -414,12 +435,12 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         FieldType ft = new FieldType();
 
         try {
-            ft.setIndexOptions(getIndexOptions((String) Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_INDEX_OPTIONS),
-                            LuceneFunctionNames.LuceneFieldIndexOptions.DOCS_AND_FREQS_AND_POSITIONS.name())));
+            ft.setIndexOptions(getIndexOptions((String)Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_INDEX_OPTIONS),
+                    LuceneFunctionNames.LuceneFieldIndexOptions.DOCS_AND_FREQS_AND_POSITIONS.name())));
             ft.setTokenized(true);
             ft.setStored(field.isStored());
-            ft.setStoreTermVectors((boolean) Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_WITH_TERM_VECTORS), false));
-            ft.setStoreTermVectorPositions((boolean) Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_WITH_TERM_VECTOR_POSITIONS), false));
+            ft.setStoreTermVectors((boolean)Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_WITH_TERM_VECTORS), false));
+            ft.setStoreTermVectorPositions((boolean)Objects.requireNonNullElse(field.getConfig(LuceneFunctionNames.LUCENE_FULL_TEXT_FIELD_WITH_TERM_VECTOR_POSITIONS), false));
             ft.setOmitNorms(true);
             ft.freeze();
         } catch (ClassCastException ex) {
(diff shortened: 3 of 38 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -183,23 +184,26 @@ public void invalidatePrefix(@Nonnull Tuple prefix) {
         }
     }
 
-    private FDBDirectoryWrapper getDirectoryWrapper(@Nullable Tuple groupingKey) {
+    private FDBDirectoryWrapper getDirectoryWrapper(@Nullable Tuple groupingKey, @Nullable Integer partitionId) {
         final Tuple mapKey = groupingKey == null ? TupleHelpers.EMPTY : groupingKey;
+        if (partitionId != null) {
+            mapKey.add(LucenePartitioner.PARTITION_DATA_SUBSPACE).add(partitionId);
```
wshaib-apple, line 190: ⬆️ bug: `add` returns a copy. fix being prepared.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java
```diff
@@ -213,6 +213,25 @@ public class IndexOptions {
...
+     */
+    public static final String TEXT_DOCUMENT_PARTITIION_HI_WATERMARK = "textDocumentPartitionHiWatermark";
+
+    /**
+     * Option to set low watermark size for a lucene partition, below which the partition would be a candidate for
+     * merge with another adjacent partition.
+     */
+
+    public static final String TEXT_DOCUMENT_PARTITIION_LO_WATERMARK = "textDocumentPartitionLoWatermark";
+
```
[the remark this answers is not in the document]
wshaib-apple: Yes, I had asked Scott about this, and he suggested the options approach. My understanding is there is a way to update the option value without rebuilding the index (?)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -506,6 +531,70 @@ private StoreTimer.Counter getCounter(@Nonnull final FDBRecordContext recordCont
...
+        metaDataBuilder.addIndex(joined, new Index("joinNestedConcat", concat(
+                field("complex").nest(function(LuceneFunctionNames.LUCENE_STORED, field("is_seen"))),
+                field("simple").nest(function(LuceneFunctionNames.LUCENE_TEXT, field("text"))
+                )
+        ), LuceneIndexTypes.LUCENE, ImmutableMap.of(IndexOptions.TEXT_DOCUMENT_PARTITION_TIMESTAMP, "complex.time")));
+        metaDataBuilder.addIndex("SimpleDocument", "simple$group", field("group"));
+    }
+
+    @Test
+    void partitionedJoinedIndexTest() {
```
ScottDugas, line 566: It would be good to also have a test that has multiple groups, and is partitioned.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -506,6 +531,70 @@ private StoreTimer.Counter getCounter(@Nonnull final FDBRecordContext recordCont
...
+        try (FDBRecordContext context = openContext()) {
+            openRecordStore(context, LuceneIndexTest::joinedPartitionedLuceneIndexMetadataHook);
+
+            TestRecordsTextProto.ComplexDocument cd = TestRecordsTextProto.ComplexDocument.newBuilder()
```
ScottDugas, line 570: It would probably be good to save a second record that doesn't match.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+    public static final int PARTITION_META_SUBSPACE = 0;
+    public static final int PARTITION_DATA_SUBSPACE = 1;
+    private final IndexMaintainerState state;
+    private final boolean partitioningEnabled;
+    private final String partitionTimestampFieldName;
+    private final Map<Tuple, List<LucenePartitionInfoProto.LucenePartitionInfo>> partitionMetaInfoCache = new HashMap<>();
+
+    public LucenePartitioner(@Nonnull IndexMaintainerState state) {
+        this.state = state;
+        partitionTimestampFieldName = state.index.getOption(com.apple.foundationdb.record.metadata.IndexOptions.TEXT_DOCUMENT_PARTITION_TIMESTAMP);
```
ScottDugas, line 64: Note: the record layer has a concept of index validation that can be run to ensure that the options and expressions make sense. At some point we should add validation that field exists, and is of the right type.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -332,40 +336,132 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexable
...
-        return AsyncUtil.DONE;
+    private Long getPartitioningTimestampValue(@Nonnull String fieldName, Map<Descriptors.FieldDescriptor, Object> fields) {
+        for (Map.Entry<Descriptors.FieldDescriptor, Object> field : fields.entrySet()) {
+            if (fieldName.equalsIgnoreCase(field.getKey().getName())) {
+                if (field.getValue() instanceof Long) {
+                    return (Long)field.getValue();
+                }
+                if (field.getValue() instanceof Double) {
+                    return ((Double)field.getValue()).longValue();
+                }
```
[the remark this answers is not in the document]
wshaib-apple: I added this because I was using the record definitions in the tests, where `ComplexDocument.time` is defined as `double`. I can make this only accept `long`, but I will have to tweak the test record definition (or create a new one, which may be simpler)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+                partitionId = partitions.get(partitions.size() - 1).getId();
+            }
+        }
+        return partitionId;
+    }
+
+    /**
+     * return the partition to which a document with a given timestamp should be assigned.
+     * <b>Note:</b> {@link #loadPartitioningMetadata(Set)} should be called
+     * before the first call to this method.
```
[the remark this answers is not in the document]
wshaib-apple: it is also not sitting right with me... one thought I had was adding an `initialized` attribute to the class, and then throwing an exception here if it's not set.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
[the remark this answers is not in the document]
ScottDugas: It should be fine to add another field to that message, or you could use `score`. Creating a new record type is also fine.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+                    .setId(0)
+                    .setCount(0)
+                    .setFrom(timestamp)
+                    .setTo(timestamp)
+                    .build());
+            return partitions.get(0);
+        }
+
+        return selectWritePartition(partitions, timestamp);
+        // TO DO: check and split the assigned partition if it's too big
```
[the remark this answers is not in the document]
wshaib-apple: Right. I didn't mean we would do the split here, but rather "queue" a split somewhere. Or did you mean that checking partitions for splitting happens independently and continuously via some groomer?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+     * remove a document from a partition metadata and save to db.
+     * Note that only the document count is changed (decremented). <code>from</code> and <code>to</code> are unchanged.
+     *
+     * @param grouping grouping key
+     * @param partition partition metadata object
+     */
+    public void removeFromAndSavePartitionMetadata(final Tuple grouping, final LucenePartitionInfoProto.LucenePartitionInfo partition) {
+        LucenePartitionInfoProto.LucenePartitionInfo.Builder builder = partition.toBuilder();
+        // note that the to/from of the partition do not get updated, since that would require us to know what the next potential boundary
+        // value(s) are.
```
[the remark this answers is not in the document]
wshaib-apple: This would only produce a change in the metadata if the removed document happens to have the max or the min timestamp of the partition. The range of the partition remains correct in all cases, which is why I think it may be ok to let it stay like this (at least for now).
Eventually, when we split partitions, the metadata would need to be updated, but that would be simpler.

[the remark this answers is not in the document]
ScottDugas: That makes sense.
It basically means that you may have partitions with ranges larger than the min/max of the actual data, but it will never be less, so you can still guarantee a strict ordering of the partitions.
And the splitting will need to update, but it will have to select the x oldest or newest documents anyways, at which point it can correctly update the metadata for the partition.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+    private LucenePartitionInfoProto.LucenePartitionInfo selectWritePartition(final List<LucenePartitionInfoProto.LucenePartitionInfo> partitions, final long timestamp) {
+
+        // candidate partitions must include at least one
+        Preconditions.checkNotNull(partitions);
+        Preconditions.checkArgument(!partitions.isEmpty());
+
+        // ----------------------------------------------------------
+        // The partitioning rules are as follows (per requirements doc):
+        //
+        // When adding an email
```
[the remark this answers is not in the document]
wshaib-apple: I copied this entire comment from the quip doc for reference while I wrote the implementation. I will remove it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,356 @@
...
+        //
+        // If its date is within a range of an existing partition it will get added to that partition.
+        // If it is between two partitions, it will get added to the smaller one, expanding that partitions date range
+        // If it is older than the oldest partition’s min date it will get added there.
+        // If it is newer than the newest partition’s max date it will get added there.
+        //
+        // If the partition that we just added to is larger than “max” size we will probabilistically split it, with a growing probability as the size increases.
```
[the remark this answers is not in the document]
wshaib-apple: theoretical in the context of mail, but this can happen in other contexts, or in the case of an improperly built test population tool etc. Not sure how to handle this, will think about it.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -362,10 +378,19 @@ protected void openRecordStore(FDBRecordContext context, FDBRecordStoreTestBase.
         RecordMetaDataBuilder metaDataBuilder = RecordMetaData.newBuilder().setRecords(TestRecordsTextProto.getDescriptor());
         metaDataBuilder.getRecordType(COMPLEX_DOC).setPrimaryKey(concatenateFields("group", "doc_id"));
         hook.apply(metaDataBuilder);
-        recordStore = getStoreBuilder(context, metaDataBuilder.getRecordMetaData())
-                .setSerializer(TextIndexTestUtils.COMPRESSING_SERIALIZER)
-                .createOrOpen();
-        setupPlanner(null);
+        recordStore = getStoreBuilder(context, metaDataBuilder.getRecordMetaData()).createOrOpen();
```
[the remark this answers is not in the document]
wshaib-apple: Not too sure about the specifics, but I had to get the join index type test working, and this made it work. I can re-add the `COMPRESSING_SERIALIZER` option to the record store I guess.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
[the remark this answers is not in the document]
ScottDugas: I think it would be via some sort of client-provided queue.
It may also make sense to combine this and the merge into a single `IndexOperation` to maintain the index, and the `LuceneIndexMaintainer` will determine whether anything needs to be moved between partitions, and whether merges need to happen.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,419 @@
...
+    /**
+     * helper - create a new partition metadata instance.
+     *
+     * @return partition metadata instance
+     */
+    @Nonnull
+    private LucenePartitionInfoProto.LucenePartitionInfo newPartitionMetadata(long timestamp) {
+        return LucenePartitionInfoProto.LucenePartitionInfo.newBuilder()
+                .setCount(0)
+                .setTo(ByteString.copyFrom(Tuple.from(0L).pack()))
```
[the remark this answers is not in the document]
wshaib-apple: It does get set when the record is saved. adding it also here will help only with readability.

ScottDugas (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/LogMessageKeys.java
```diff
@@ -306,6 +306,10 @@ public enum LogMessageKeys {
     HASH_FUNCTION,
     HASH,
 
+    // Lucene
```
ScottDugas, line 309: I missed it, but you should update the release notes to indicate that the associated issue is being resolved.
You just replace one of the commented out lines, like:
```
```

becomes
```
* **Feature** Add lucene partitioning metadata [(Issue #2390)](https://github.com/FoundationDB/fdb-record-layer/issues/2390)
```

And then the build will put it in the actual release notes when it creates a release.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -0,0 +1,420 @@
...
+    /**
+     * get the <code>long</code> timestamp value from a {@link FDBIndexableRecord}, given a field name.
+     *
+     * @param fieldName field name
+     * @param rec record
+     * @return long if field is found
+     * @throws RecordCoreException if no field of type <code>long</code> with given name is found
+     */
+    @Nonnull
+    private <M extends Message> Long getPartitioningTimestampValue(@Nonnull String fieldName, @Nonnull FDBIndexableRecord<M> rec) {
```
MMcM, line 293: It feels like perhaps this should build (and cache) a `KeyExpression` to do this work, rather than calling `Message.getAllFields` all the time. Or at least do the `getDescriptorForType` and `findFieldByName` and `getField` itself.

I don't have a problem with this being in a later pull request, though.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
```diff
@@ -80,6 +80,24 @@ public class LuceneIndexOptions {
...
+     * Option to designate a record field as the timestamp by which the corresponding document would be partitioned in
+     * Lucene.
+     */
+    public static final String INDEX_PARTITION_BY_TIMESTAMP = "partitionTimestamp";
+
+    /**
+     * Option to set high watermark size for a lucene partition, beyond which the partition would be split, or a new
+     * partition would be created.
+     */
+    public static final String INDEX_PARTITION_HI_WATERMARK = "partitionHiWatermark";
```
MMcM, line 93: I think `HIGH` might be preferable or (less so but more consistent) `LO` below.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -29,38 +29,80 @@
...
 import static com.apple.foundationdb.record.lucene.LuceneIndexOptions.INDEX_PARTITION_BY_TIMESTAMP;
+import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
 import static com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer.Waits.WAIT_LOAD_LUCENE_PARTITION_METADATA;
 
 /**
  * Manage partitioning info for a <b>logical</b>, partitioned lucene index, in which each partition is a separate physical lucene index.
  */
 @API(API.Status.EXPERIMENTAL)
 public class LucenePartitioner {
+    private static final ConcurrentHashMap<String, KeyExpression> partitioningKeyExpressionCache = new ConcurrentHashMap<>();
```
ScottDugas, line 60: Having a static, unbounded cache here could be problematic in the future, if there are a lot of different index definitions on the same machine. Probably fine for the experimental phase of the Lucene index.

It seems almost as if we want a different subclass of `Index` that gets created, and thus the clients caching of the `RecordMetaData`, but that would be well beyond the scope of this change.
I guess the other possibility would be to have the index defined with multiple `KeyExpression` or to have the root expression be split by grouping, and partitioning. But that also brings a large amount of complexity, so let's stick with the option, and parsing for now (and potentially forever).

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -29,38 +29,80 @@
...
+
+    /**
+     * make (and cache) a key expression from the partitioning field name.
+     *
+     * @param partitionTimestampFieldName partitioning field name
+     * @return key expression
+     */
+    @Nullable
+    private KeyExpression makePartitioningKeyExpression(@Nullable final String partitionTimestampFieldName) {
+        if (partitionTimestampFieldName == null) {
```
ScottDugas, line 86: Because of where this is called, I don't think this will ever be true.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -29,38 +29,80 @@
...
+            if (nameComponents.length == 1) {
+                // no nesting
+                return field(nameComponents[0]);
+            }
+            // nameComponents.length >= 2
+            List<KeyExpression> fields = Arrays.stream(nameComponents).map(Key.Expressions::field).collect(Collectors.toList());
+            for (int i = fields.size() - 1; i > 0; i--) {
+                fields.set(i - 1, ((FieldKeyExpression) fields.get(i - 1)).nest(fields.get(i)));
+            }
+            return fields.get(0);
```
ScottDugas, line 104: `Key.Expressions::field` returns a `FieldKeyExpression`, so if you change the type of fields to `List<FieldKeyExpression>` I think you can remove the cast.

ScottDugas (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: `.nest()` returns a `NestingKeyExpression` which is not a `FieldKeyExpression`, so the list needs to be of the base type.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: If the index isn't partition, this will be `null`

ScottDugas: Yes, but I think if the index isn't partitioned, this method isn't called.

wshaib-apple: I don't know, maybe you're seeing something I am not... :)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: 👍

MMcM (approved):

PR #2395, opened 2023-12-06
Resolves #2394: Lucene primary key segment index creates more segments than needed

### LINKED ISSUE #2394: Lucene primary key segment index creates more segments than needed
It should not be necessary to flush the writer and get a brand new reader, since `tryDeleteDocument` operates on docids and is already prepared to decline in edge cases.

MMcM:


MMcM:
`LuceneStoredFieldsTest .testDeleteAllDocuments` failure is due to the new delete approach managing to clean up all segments, rather than leaving one with all tombstoned documents. Probably the test fixture needs to assert on the number of _live_ documents.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1374,6 +1375,40 @@ void checkFieldInfoCountingAfterMerge() {
...
+            segmentCountBefore = getSegmentCount(index, Tuple.from(0));
+        }
+        try (FDBRecordContext context = openContext(contextProps)) {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            for (int i = 0; i < 10; i++) {
+                recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(i, numbersText(i), "", 0));
+            }
+            context.commit();
+        }
+        assertThat(timer.getCount(LuceneEvents.Events.LUCENE_DELETE_DOCUMENT_BY_QUERY), lessThan(5));
```
ScottDugas, line 1403: Debugging this test against main, it looks like `LUCENE_FIND_MERGES` is 14, would it make sense to assert about that counter too? Locally it appears to be 4, which still seems higher than I would expect, but making sure it's less than 9 seems good. Of course that wouldn't apply if we defer merges. Deferring merges would also effect the segment count assertions.

ScottDugas (changes requested):

ScottDugas (approved):

PR #2399, opened 2023-12-10
Add nested required fields in IndexKeyValueToPartialRecord

pengpeng-lu:
Hit a bug when a nested message has a required field


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -368,6 +368,12 @@ public void addRequiredMessageFields() {
                 if (fieldDescriptor.isRequired() && fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
                     nestedBuilders.putIfAbsent(fieldDescriptor.getName(), new Builder(fieldDescriptor.getMessageType(), true));
                 }
+                if (fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
+                    Builder builder = nestedBuilders.get(fieldDescriptor.getName());
```
alecgrieser, line 372: Would this loop be equivalent to just `nestedBuilders.values().forEach(Builder::addRequiredMessageFields)`? That may be a bit clearer (or maybe it's six of one, half dozen of the other) that effectively what's going on here is just recursively adding in builders for any required fields.

alecgrieser (changes requested):
Would it be straightforward to add a test for this to make sure we cover the nested case?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -368,6 +368,12 @@ public void addRequiredMessageFields() {
                 if (fieldDescriptor.isRequired() && fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
                     nestedBuilders.putIfAbsent(fieldDescriptor.getName(), new Builder(fieldDescriptor.getMessageType(), true));
                 }
+                if (fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
```
MMcM, line 371: I would either do as @alecgrieser suggests below and have a recursion pass over all child builders at the end rather than field-by-field, _or_ combine this `if` and the previous one into a single "nested message" one that (1) add builder if message field itself is required and (2) recurse if there is a child builder.

MMcM (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/IndexKeyValueToPartialRecordTest.java
```diff
@@ -104,4 +104,12 @@ void convertIndexToPartialRecordConcurrently() throws InterruptedException {
         service.shutdown();
         Assertions.assertTrue(completed);
     }
+
+    @Test
+    void testAddRequiredMessage() {
+        // has a required message field in a nested field
+        IndexKeyValueToPartialRecord.Builder builder = IndexKeyValueToPartialRecord.newBuilder(TestRequiredNestedRecordProto.MyNestedRecord.getDescriptor());
+        builder.addRequiredMessageFields();
+        Assertions.assertTrue(builder.isValid(true));
```
pengpeng-lu, line 113: build.isValid returns false before the change.

MMcM (approved):

alecgrieser (approved):

PR #2406, opened 2023-12-12
[DO NOT MERGE] log message change pr build test

lrjenkins:
nonsense change to test PR build


ammolitor:
it appears the filter works. this PR did not trigger a build.

ammolitor (changes requested):
this should remain a log.warn

PR #2413, opened 2023-12-14
Remove references of caller context to avoid future misuse

jjezra:
The purpose of this PR is to prevent developers for accidentally misusing the caller context of FDBDirectory.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
@@ -23,6 +23,7 @@ package com.apple.foundationdb.record.lucene;
 import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
 import com.apple.foundationdb.record.provider.foundationdb.KeyValueCursor;
 import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.ByteArrayUtil;
@@ -52,6 +53,7 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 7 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -76,15 +78,21 @@ public LucenePrimaryKeySegmentIndex(@Nonnull FDBDirectory directory, @Nonnull Su
      */
     @VisibleForTesting
     public List<List<Object>> readAllEntries() {
+        AtomicReference<List<List<Object>>> list = new AtomicReference<>();
+        directory.getAgilityContext().accept(aContext -> readAllEntries(aContext, list));
```
ScottDugas, line 82: Why use `accept` with the `AtomicReference`, rather than just calling `apply`?
I think you would have to move the `asyncToSync` up to this method.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -126,10 +134,17 @@ public DocumentIndexEntry(@Nonnull final Tuple primaryKey, @Nonnull final byte[]
      */
     @Nullable
     public DocumentIndexEntry findDocument(@Nonnull DirectoryReader directoryReader, @Nonnull Tuple primaryKey) {
+        final AtomicReference<DocumentIndexEntry> doc = new AtomicReference<>();
+        directory.getAgilityContext().accept(aContext -> findDocument(aContext, doc, directoryReader, primaryKey));
```
ScottDugas, line 138: Here also, if you pull the `asyncToSync` up to here, you can use `apply` and remove the `AtomicReference`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -65,6 +70,42 @@ default void clear(Range range) {
         accept(context -> context.ensureActive().clear(range));
     }
 
+    default CompletableFuture<List<KeyValue>> getRange(byte[] begin, byte[] end) {
+        return apply(context -> context.ensureActive().getRange(begin, end).asList());
+    }
+
+    @Nonnull
+    FDBRecordContext getCallerContext();
```
ScottDugas, line 78: Perhaps worth a javadoc on this method, specifically calling out that you shouldn't actually use it.
Or perhaps it would be better to just change all the default methods below to not be default, further encapsulating the caller context.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -918,8 +917,20 @@ public int getBlockSize() {
         return blockSize;
     }
 
-    public FDBRecordContext getContext() {
-        return context;
+    @VisibleForTesting
+    FDBRecordContext getCallerContext() {
+        // We cannot control the way the returned context is used. Limiting it to testing seems to be a good idea.
```
ScottDugas, line 922: It would probably be good to followup with removing this too, but that can probably wait.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
```diff
@@ -69,7 +69,7 @@ public void testGetIncrement() {
         assertEquals(1, directory.getIncrement());
         assertEquals(2, directory.getIncrement());
         assertCorrectMetricCount(LuceneEvents.Counts.LUCENE_GET_INCREMENT_CALLS, 2);
-        directory.getContext().commit();
+        directory.getCallerContext().commit();
```
ScottDugas, line 72: In the (potential) followup where `getCallerContext` is removed, you could change the `context` created in `setUp`, and use that in all of these usages.
Although, looking at this test, and some others, it would probably make sense to remove the fields for `directory` and just create the `context` & directory in a `try-with-resources` in the test (as is done in this test), and then reference the contexts that we are creating.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
jjezra: I've tried that first, and it broke.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
jjezra: Same here - apply caused cancellation exception.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: For now, at least, it is required for `FDBDirectory.getCallerContext`. But we can change it in the future. 
I'll add javadoc.

jjezra: ✅

ScottDugas (approved):

PR #2417, opened 2023-12-18
Resolve #2401: Make test less flaky

### LINKED ISSUE #2401: Stored fields assertDocCountPerSegment sensitivity
A change to document deletion broke `StoredFieldsTest` because a previously tombstone-only segment was now cleaned up earlier. The purpose of the assertion is to find stored field orphans; could that be done in a way that ignores such a difference?

ohadzeliger:
Lucene stored fields test was flaky when it was expecting documents in segments that were the results of a merge.
Added forced merge and `LUCENE_MERGE_SEGMENTS_PER_TIER` as well as additional fragmentation to the segments such that a merge is forced and the result is predictable.


ohadzeliger:
There has been 1 test that failed in the initial commit. Rerunning seems to have gone smoothly. Messaged @jjezra about it.

MMcM (approved):

PR #2419, opened 2023-12-19
Remove PERMUTED from the name of MIN and MAX permutable indexes

arnaud-lacurie:


MMcM (changes requested):
I am fine with changing `PERMUTED` to something more directly suggestive of the use case. But plain `MAX` and `MIN` are achievable with a regular index (just scan the endpoints limit 1 -- this works today). What this enhanced (it includes a full value B-tree) index adds is the ability to sort by the extremum.

alecgrieser:
There still can be benefits to using one of these indexes, though, even in the unpermuted case. In particular, if you want to leave some of the grouping columns unbound, like:

```
SELECT c, max(d) FROM type WHERE a = $aValue AND b = $bValue GROUP BY c ORDER BY c ASC
```

That can be done with an index on `field("d").groupBy(concatenateFields("a", "b", "c"))`with a permuted size of 0. Theoretically, you could do this with a regular index on `concatenateFields("a", "b", "c", "d")`, but you'd need to combine a skip scan to get each value `c` followed by a singular reverse scan to get `max(d)`. (FWIW, this could be done with a single skip scan if that order by was `ORDER BY c DESC`.)

Overall, I think naming the index `MIN`/`MAX` and then letting its permuted-ness come through in the index option makes sense, though my opinions are actually not that strong. We actually currently don't allow a permuted size of 0 (it fails index validation), but I think it actually would be useful to allow that.

PR #2420, opened 2023-12-19
Resolves #2418: Integrate the PERMUTED_MIN/_MAX index types into the query planners

### LINKED ISSUE #2418: Integrate the PERMUTED_MIN/_MAX index types into the query planners
The `PERMUTED_MIN` and `PERMUTED_MAX` index types can currently be defined in the meta-data, but they are not integrated into queries. This applies to both the Cascades planner, which can plan aggregate functions out of the box, as well as to some of the abstractions we have for evaluating aggregate functions or executing a covering aggregate scan that can be used without Cascades.

alecgrieser:
This does a few things:

1. It defines what index aggregate functions the `PERMUTED_MIN` and `PERMUTED_MAX` functions can evaluate. Effectively, this communicates with consumers that those indexes can be used to satisfy the `min` and `max` function. It also provides corresponding implementations of those functions using the index.
1. The `planCoveringIndex` method in the maintainer is updated to allow for scans over the index that cover the unpermuted grouping columns.
1. Match candidates for the index with the appropriate contents and the appropriate ordering are now generated for consumption by Cascades.
1. Ordering logic is updated so that the ordering keys are matchable for aggregate indexes. For most indexes, this follows the order of the key expression, though the `PERMUTED_MIN` and `PERMUTED_MAX` indexes have special handling for the permuted size.

This resolves #2418.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.query.plan.cascades;
 
 import com.apple.foundationdb.record.metadata.Index;
+import com.apple.foundationdb.record.metadata.IndexOptions;
 import com.apple.foundationdb.record.metadata.IndexTypes;
 import com.apple.foundationdb.record.metadata.RecordType;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
@@ -52,6 +53,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -79,6 +81,8 @@ public class AggregateIndexExpansionVisitor extends KeyExpressionExpansionVisito
     @Nonnull
     private final GroupingKeyExpression groupingKeyExpression;
 
+    private final int columnPermutations;
+
     /**
      * Constructs a new instance of {@link AggregateIndexExpansionVisitor}.
      *
@@ -91,6 +95,12 @@ public class AggregateIndexExpansionVisitor extends KeyExpressionExpansionVisito
         this.index = index;
         this.groupingKeyExpression = ((GroupingKeyExpression)index.getRootExpression());
         this.recordTypes = recordTypes;
+        @Nullable String permutationOption = index.getOption(IndexOptions.PERMUTED_SIZE_OPTION);
+        this.columnPermutations = permutationOption == null ? 0 : Integer.parseInt(permutationOption);
+    }
+
+    public boolean isPermuted() {
+        return IndexTypes.PERMUTED_MAX.equals(index.getType()) || IndexTypes.PERMUTED_MIN.equals(index.getType());
     }
 
     /**
@@ -136,7 +146,7 @@ public class AggregateIndexExpansionVisitor extends KeyExpressionExpansionVisito
         final var traversal = ExpressionRefTraversal.withRoot(maybeWithSort);
         return new AggregateIndexMatchCandidate(index,
                 traversal,
-                selectWhereQunAndPlaceholders.getRight().stream().map(Placeholder::getParameterAlias).collect(Collectors.toList()),
+                placeHolderAliases,
                 recordTypes,
                 baseQuantifier.getFlowedObjectType(),
                 groupByQun.getRangesOver().get().getResultValue(),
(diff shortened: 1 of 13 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -320,13 +324,57 @@ public static Optional<FieldPath> stripFieldPrefixMaybe(@Nonnull FieldPath field
...
+            return true;
+        }
+        if (t1.isNullable() != t2.isNullable()) {
+            return false;
+        }
+        if (t1.isArray()) {
+            if (!t2.isArray()) {
+                return false;
+            }
+            @Nullable Type t1Element = ((Type.Array)t1).getElementType();
```
normen662, line 347: This can only be `null` if the type is erased which should not happen as these types are resolved. I am assuming you are being defensive here and just return `false` if it happens. Maybe you can do a `Objects.requireNonNull()` to enforce it is not `null`. Or alternatively, maybe we can make the API better. There are other places where we have different flavors, life e.g. `getThing()` and `getThingMaybe()` where the latter returns an `Optional` while the former throws if it is empty.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -320,13 +324,57 @@ public static Optional<FieldPath> stripFieldPrefixMaybe(@Nonnull FieldPath field
...
-            if (fieldPathOrdinals.get(i) != potentialPrefixFieldOrdinals.get(i) || !fieldPathTypes.get(i).equals(potentialPrefixFieldTypes.get(i))) {
+            if (fieldPathOrdinals.get(i) != potentialPrefixFieldOrdinals.get(i) || !equalIgnoringFieldNames(fieldPathTypes.get(i), potentialPrefixFieldTypes.get(i), new HashSet<>())) {
                 return Optional.empty();
             }
         }
         return Optional.of(fieldPath.subList(potentialPrefixPath.size(), fieldPath.size()));
     }
 
+    @SuppressWarnings("PMD.CompareObjectsWithEquals") // Deliberate use of reference equality
+    private static boolean equalIgnoringFieldNames(@Nonnull Type t1, @Nonnull Type t2, @Nonnull Set<Pair<Type, Type>> explored) {
```
normen662, line 335: This method needs some comments and should adhere to other styling like `final` in the signature, etc.

Other than that I am wondering if this method can be simplified/restructured into the field path logic.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
```diff
@@ -246,9 +256,11 @@ private Pair<SelectExpression, List<CorrelationIdentifier>> constructSelectHavin
 
         final var placeholderAliases = ImmutableList.<CorrelationIdentifier>builder();
         final var selectHavingGraphExpansionBuilder = GraphExpansion.builder().addQuantifier(groupByQun);
+        final List<Value> groupingValues = groupingValueReference == null ? Collections.emptyList() : Values.deconstructRecord(groupingValueReference);
```
alecgrieser, line 259: During an in-person review of this code, this change came up. To be clear, the change here was to construct the list of grouping values before going through and assigning place holders. That way, the _size_ of the grouping values can be referenced below in the `isPermuted()`-guarded statement.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidate.java
```diff
@@ -261,6 +261,7 @@ default Set<String> getQueriedRecordTypeNames() {
     }
 
     @Nonnull
+    @SuppressWarnings("fallthrough")
```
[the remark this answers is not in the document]
alecgrieser: Removed. For reference, I added this while working on the PR when during an earlier version, the `PERMUTED_MIN` and `PERMUTED_MAX` cases added a value index expansion first and then fell through to the other aggregate index cases. But I've removed that fall through and now just specify both expansions in a case that is specific to the `PERMUTED_MIN` and `PERMUTED_MAX` index types because the other approach was a little too clever by half

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
alecgrieser: We discussed this during an in-person review, and we now think the right thing is to remove the type check entirely from `stripFieldPrefixMaybe`. After that, this method becomes unnecessary, so I've removed it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
alecgrieser: Now irrelevant as this method has been removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexMatchCandidate.java
```diff
@@ -202,9 +217,12 @@ public List<MatchedOrderingPart> computeMatchedOrderingParts(@Nonnull final Matc
             //
             // Compute a Value for this normalized key.
             //
+            final var value = deconstructedValue.get(permutedIndex).rebase(aliasMap);
+            /*
```
normen662, line 221: please take this comment out -- and maybe add a comment

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexMatchCandidate.java
alecgrieser: Done

normen662 (approved):

hatyo (approved):

Issue #763, opened 2019-10-25
Evaluating a query component for equality comparison with a enum constant fails


MMcM:
For example, remove the index definition from `FDBRecordStoreQueryTest.enumFields` and still get the same results (if not the same plan).

```
com.apple.foundationdb.record.RecordCoreException: Tried to compare non-comparable object class com.google.protobuf.Descriptors$EnumValueDescriptor
```

alecgrieser:
Great! I suppose it's also possible that your enum field filter is better executed (performance wise) with the index, too, though this bug should be fixed.

PR #770, opened 2019-11-04
Fixes #763: Evaluating a query component for equality comparison with a enum constant fails

alecgrieser:
The approach here was to compare enum fields based on their field number. This should give the same semantics for enum equality and inequality comparisons when no index is defined as is done when there is an index. There is some amount of type safety lost in the comparison, which is unfortunate.

One unfortunate thing about our API is that when you call `getField` on an enum field on a protobuf message, you get back an `EnumValueDescriptor`. However, our query validator throws an error unless you include a `ProtocolMessageEnum` of the right type. (And one can convert to an `EnumValueDescriptor` by calling `.toValueDescriptor`, but it is hard to convert back, hence why the enum number is used for comparison internally.) I kind of think that our value is more sensical, but it is not the type returned by `.getField` or allowed by `.setField` (though I think *is* allowed on generated messages), so ¯\\\_(ツ)_/¯.

While I was here, I added tests on `bool`s, `float`s, and `double`s. I wanted to add more tests surrounding weird NaN behavior, but unfortunately, it kept collapsing all NaN values (positive and negative and with weird mantissas) to a canonical NaN, so I could only make them so weird. That doesn't appear to be the contract according to the Javadocs, so something may be up.

This fixes #763.

PR #2428, opened 2023-12-28
resolves: #2443 serialized plans

normen662:
- added plan serialization/deserialization framework
- added `PlanSerializable` as the root interface to implement plan serialization
- added tons of protobuf message definitions
- added way of extending proto definitions in downstream projects
- removed usages of dangerous plan fragments such as protobuf descriptors or runtime-specific data structures
- rewrote some operators to make certain planner specific attributes nullable as those are not serialized
- `DynamicEnum` ad `EnumLite` extensions to allow for serialization of plans using enums as part of their plan fragments (e.g. comparison `field = RED`)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
```diff
@@ -0,0 +1,232 @@
...
+    private static final String PREFIX = "META-INF/services/";
+
+    private static final Map<Class<? extends Message>, Method> fromProtoMapMethodMap =
+            loadFromProtoMethodMap();
+
+    /**
+     * Parse a single line from the given configuration file, adding the
+     * name on the line to set of names if not already seen.
+     */
+    private static int parseLine(BufferedReader r, int lc, Set<String> namesAlreadySeen, Set<String> names)
```
alecgrieser, line 76: Why are we parsing the config files ourselves? Is this because the `ServiceLoader` returns a list of instantiated instances, and you need a collection of `Class<?>` objects instead? And the `ServiceLoader` can't load the instances because they need a default constructor to be service loadable, which this is not.

I don't love having custom parsing logic of these fields here (or the use of reflection, if I'm honest). I think the way we've typically done this before is by introducing a `Factory` class that can have default constructors, and then making the `Factory` the thing in the service loader.

For example, I think if this were the interface:

```java
public interface PlannerDeserializer<M extends Message, P extends PlannerSerializable> {
    Class<M> getProtoMessageClass();
    P fromProto(ProtoSerializationContext serializationContext, M protoMessage);
}
```

Then I think you could use a `ServiceLoader` to collect those factory implementations without the need for custom parsing logic here or reflection

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
```diff
@@ -0,0 +1,232 @@
...
+ * Class to facilitate serialization and deserialization of
+ * {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan}s.
+ */
+public class PlanSerialization {
+    private static final Logger logger = LoggerFactory.getLogger(PlanSerialization.class);
+
+    private static final String PREFIX = "META-INF/services/";
+
+    private static final Map<Class<? extends Message>, Method> fromProtoMapMethodMap =
+            loadFromProtoMethodMap();
```
alecgrieser, line 70: Instead of this being static, should there be some kind of configurable `PlannerDeserializerRegistry` that can be configured by the user like, for example, the `IndexMaintainerRegistry`? The benefit of that kind of approach is that it lets downstream adopters choose their own dependency injection strategy for any custom deserializers

alecgrieser (changes requested):
I haven't looked at too many of the individual implementations yet, but I had a few comments on the basic framework that I figured were good to get out before taking a deeper look

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/ProtoMessage.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.apple.foundationdb.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for classes that serialize as part of a query plan.
```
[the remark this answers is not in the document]
normen662: It has to be upstream, as the annotation processor needs it and that has to be available when for is compiled.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Column.java
```diff
@@ -89,4 +92,21 @@ public int planHash(@Nonnull final PlanHashMode hashMode) {
...
         }
     }
+
+    @Nonnull
+    @Override
+    public PColumn toProto(@Nonnull final PlanSerializationContext serializationContext) {
+        PColumn.Builder builder = PColumn.newBuilder();
+        builder.setField(field.toProto(serializationContext));
+        builder.setValue(value.toValueProto(serializationContext));
+        return builder.build();
```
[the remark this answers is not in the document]
normen662: agreed.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -0,0 +1,1501 @@
...
+  optional PValue value = 2;
+  optional PComparison comparison = 3;
+}
+
+//
+// Plans
+//
+message PPhysicalQuantifier {
+  optional string alias = 1;
+  repeated PPlanReference plan_references = 2;
```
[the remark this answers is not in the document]
normen662: it's for the case in the future when we may want to serialize variations.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -0,0 +1,1501 @@
...
+}
+
+//
+// Plan reference holder to assign ids to plans in order to support plan DAGs.
```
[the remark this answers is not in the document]
normen662: In general, we will encounter DAGs in the future. Mike even suggested adding support for cycles. The easiest way to create something like that is a self join over a view:

```
CREATE VIEW V AS SELECT * from T ORDER BY a WHERE ...
```

and then

```
SELECT * FROM V v1, V v2 WHERE ...
```

depending on the cost of computing `V` which is normally just macroed in, the planner may decide to create a common subexpression, materialize it and read it twice.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanSerializationContext.java
```diff
@@ -0,0 +1,94 @@
...
+        this(mode, IdentityBiMap.create());
+    }
+
+    public PlanSerializationContext(@Nonnull final PlanHashMode mode, @Nonnull final IdentityBiMap<RecordQueryPlan, Integer> knownPlansMap) {
+        this.mode = mode;
+        this.knownPlansMap = knownPlansMap;
+    }
+
+    @Nonnull
+    public PlanHashMode getMode() {
```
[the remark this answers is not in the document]
normen662: `PlanHashable` and `PlanSerialization` will go hand in hand. `PlanHashMode` is versioned so we it 'll be able to directly code migration logic into the deserializer.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -0,0 +1,1352 @@
...
+  optional bool children_map_is_null = 1;
+  repeated IntChildPair child_pair = 2;
+  optional PValue value = 3;
+}
+
+//
+// Values
+//
+message PValue {
+  extensions 5000 to max;
```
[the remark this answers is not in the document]
normen662: I think we should talk about this.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -0,0 +1,1352 @@
...
+      SUB_FI = 34;
+      SUB_FL = 35;
+      SUB_FF = 36;
+      SUB_FD = 37;
+      SUB_DI = 38;
+      SUB_DL = 39;
+      SUB_DF = 40;
+      SUB_DD = 41;
+
+      MUL_II = 42;  // my favorite
```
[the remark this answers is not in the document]
normen662: Your comment is actually geared towards how that `Value` is implemented. You can encapsulate the input arguments to bring them to the max type and then perform the operation or you can have a physical operator for every possible combination. We went with the latter.

This PR does not attempt to be optimal here, we try to just reflect the implementation in proto: If the java class follows a particular structure, the proto structure will attempt to mimic that as close as possible.

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/CheckProtoMessageAnnotationProcessor.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Override
+    public synchronized void init(@Nonnull final ProcessingEnvironment processingEnv) {
+        super.init(processingEnv);
+    }
+
+    @Nonnull
+    @Override
+    public SourceVersion getSupportedSourceVersion() {
+        return SourceVersion.latestSupported();
```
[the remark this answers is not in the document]
alecgrieser: I'm not thinking of anything specific. My main concern was just that I wanted to understand what `SourceVersion` is doing here and what implications it has for changing JVMs (or Java source versions when we want to start using Java 17 language features)

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
alecgrieser: Sure

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
alecgrieser: I think what this PR shows is that the approach of enumerating out the various types quickly falls to the tyranny of the fundamental counting principle. I've found this pretty annoying when trying to update Java code (if I'm perfectly honest), but I think this is worse because it it bakes this choice into the serialization format which makes it harder to change in the future. If we, for example, want to introduce a new "type" here for the left and right arguments, then we'll have to introduce them without changing the enum numbers of any of the existing types.

I think this is a place where it would be better for the serialization to diverge from the POJO structure (or for the POJO structure to similarly change)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
alecgrieser: Wouldn't using a `Factory` interface let us avoid that before Java 17 (and avoid reflection post-Java 17)? I think it may also let us avoid the annotation that was added in this PR, if I'm understanding the annotation's purpose correctly

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
alecgrieser: If that's not documented somewhere, it would probably be good to do so.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanSerializationContext.java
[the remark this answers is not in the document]
alecgrieser: So, if I'm understanding this correctly, the idea is to re-use the `PlanHashMode` version rather than to introduce a new `PlanSerialization` version?

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/ProtoMessage.java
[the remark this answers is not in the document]
alecgrieser: I think my preference would then be for a separate annotation processing package, rather than putting it in `fib-extensions`.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
normen662: I changed this as per discussion into a stub that can either take an extension or an `Any` or (if we wanted to) our own version of `Any`

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/CheckProtoMessageAnnotationProcessor.java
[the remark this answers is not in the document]
normen662: Every annotation processor has to declare the supported versions. This particular return is to most lenient.

fdb-extensions/src/main/java/com/apple/foundationdb/annotation/ProtoMessage.java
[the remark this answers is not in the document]
normen662: This is the same for another annotation processor we use. I will log an issue to make that its own package.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
normen662: I think I found a compromise which removes the hard-to-maintain switch statements by keeping some enum maps.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanSerializationContext.java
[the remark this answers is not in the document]
normen662: Yes and no -- it's up to us. I think we should have the versioning of both by in sync with one another.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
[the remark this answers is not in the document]
normen662: `ExpressionRef<>` and `GroupExpressionRef` are their counterparts. Will add a comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
normen662: So I played with your suggested approach quite a bit to get this as concise as possible. I had thought about that before I went the route you see in the PR but I ended up quite disliking it for the additional boilerplate it introduces. There are 200ish implementations that now have to define a static deserialization class. Quite tedious. I also started to try to get the serializer autogenerated through the `ProtoMessage` annotation but that didn't work as the generated code needed to create the `AutoService` annotation which for some reason didn't get picked up to do its magic in `META-INF`. I will try to get Mike's opinion on this. It looks as if we just need to wait until that bug in the JVM is fixed and I would be able to cut a lot of code.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -174,6 +191,8 @@ private static Comparable toComparable(@Nullable Object obj) {
         } else if (obj instanceof UUID) {
             UUID uuid = (UUID)obj;
             return new UnsignedUUID(uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+        } else if (obj instanceof Internal.EnumLite) {
+            return ((Internal.EnumLite)obj).getNumber();
```
MMcM, line 195: I think this fixes #763. Maybe we should include the `enumFieldsWithoutIndex` test from #770 to confirm that.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanSerializationContext.java
```diff
@@ -0,0 +1,113 @@
...
+import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan;
+import com.apple.foundationdb.record.query.plan.serialization.DefaultPlanSerializationRegistry;
+import com.apple.foundationdb.record.query.plan.serialization.PlanSerializationRegistry;
+import com.google.common.base.Verify;
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+
+/**
+ * TBD.
```
alecgrieser, line 35: At some point, we probably do want to expand on this. We should probably mark this class as `INTERNAL`, I think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpaceDirectory.java
```diff
@@ -838,6 +839,7 @@ public enum KeyType {
 
         // Function that tests a value to see if it is of this type
         @Nonnull
+        @SpotBugsSuppressWarnings("SE_BAD_FIELD")
```
alecgrieser, line 842: There is something definitely weird about this check failure. I'm pretty sure enums don't even serialize in a way that requires serializing the fields, so I think some wires got crossed. But I'm definitely fine with taking this suppression to avoid blocking this PR

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PushFilterThroughFetchRule.java
```diff
@@ -236,11 +235,6 @@ private QueryPredicate pushLeafPredicate(@Nonnull RecordQueryFetchFromPartialRec
                                              @Nonnull CorrelationIdentifier oldInnerAlias,
                                              @Nonnull CorrelationIdentifier newInnerAlias,
                                              @Nonnull final QueryPredicate leafPredicate) {
-        if (leafPredicate instanceof QueryComponentPredicate) {
-            // We cannot push these predicates. They always contain nesteds.
-            return null;
-        }
```
alecgrieser, line 242: Is this change somehow related to serialization?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
```diff
@@ -0,0 +1,240 @@
...
+        Verify.verify(protoEnumClass.isEnum());
+        final E1[] javaEnumConstants = domainEnumClass.getEnumConstants();
+        final E2[] protoEnumConstants = protoEnumClass.getEnumConstants();
+        Verify.verify(javaEnumConstants.length == protoEnumConstants.length);
+
+        final EnumBiMap<E1, E2> enumBiMap = EnumBiMap.create(domainEnumClass, protoEnumClass);
+        for (int i = 0; i < javaEnumConstants.length; i ++) {
+            final E1 e1 = javaEnumConstants[i];
+            final E2 e2 = protoEnumConstants[i];
+            Verify.verify(e1.ordinal() == e2.ordinal());
```
alecgrieser, line 234: Hm, do we care if the ordinals match? I can think of examples where we may have trouble keeping the ordinals aligned (e.g., if an enum value is added in the middle of a POJO definition because it groups well with existing values, but we want to put it at the end of the proto definition to simplify enum number assignment).

I'm okay with keeping this check here for now and then removing it later if/when we find it to be problematic

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
```diff
@@ -0,0 +1,240 @@
...
+import java.util.Objects;
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * Class to facilitate serialization and deserialization of
+ * {@link com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan}s.
+ */
+public class PlanSerialization {
```
alecgrieser, line 56: Probably want this to be marked as `INTERNAL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerializationRegistry.java
```diff
@@ -0,0 +1,42 @@
...
+import com.apple.foundationdb.record.PlanDeserializer;
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Plugin interface to be used by the plan serialization framework to dispatch serialization/deserialiation
+ * invocations properly.
+ */
+public interface PlanSerializationRegistry {
```
alecgrieser, line 32: This should probably be marked as `UNSTABLE` or `EXPERIMENTAL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/DefaultPlanSerializationRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String TYPE_URL_PREFIX = "c.a.fdb.types";
+
+    private final Map<Class<? extends Message>, PlanDeserializer<? extends Message, ?>> fromProtoClassDeserializerMap;
+    private final Map<String, Class<? extends Message>> fromProtoTypeUrlClassMap;
+
+    public DefaultPlanSerializationRegistry() {
+        final var methodMaps = loadFromProtoMethodMaps();
+        this.fromProtoClassDeserializerMap = methodMaps.getLeft();
+        this.fromProtoTypeUrlClassMap = methodMaps.getRight();
+    }
```
alecgrieser, line 51: It feels like this should be private to avoid people accidentally re-initializing this class (which it looks like we may be doing during tests)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -503,6 +512,59 @@ protected void setOptimizeForIndexFilters(boolean shouldOptimizeForIndexFilters)
...
+
+    /**
+     * Serialize the plan to bytes, parse those bytes, reconstruct, and compare the deserialized plan against the
+     * original plan.
+     * @param plan the original plan
+     * @return the deserialized and verified plan
+     */
+    @Nonnull
+    protected static RecordQueryPlan verifySerialization(@Nonnull final RecordQueryPlan plan) {
+        PlanSerializationContext serializationContext = new PlanSerializationContext(new DefaultPlanSerializationRegistry(),
```
alecgrieser, line 549:
```suggestion
        PlanSerializationContext serializationContext = new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE,
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -503,6 +512,59 @@ protected void setOptimizeForIndexFilters(boolean shouldOptimizeForIndexFilters)
...
+        final RecordQueryPlanProto.PRecordQueryPlan planProto = plan.toRecordQueryPlanProto(serializationContext);
+        final byte[] serializedPlan = planProto.toByteArray();
+        final RecordQueryPlanProto.PRecordQueryPlan parsedPlanProto;
+        try {
+            parsedPlanProto = RecordQueryPlanProto.PRecordQueryPlan.parseFrom(serializedPlan);
+        } catch (InvalidProtocolBufferException e) {
+            throw new RuntimeException(e);
+        }
+
+        serializationContext = new PlanSerializationContext(new DefaultPlanSerializationRegistry(), PlanHashable.CURRENT_FOR_CONTINUATION);
```
alecgrieser, line 560:
```suggestion
        serializationContext = new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE, PlanHashable.CURRENT_FOR_CONTINUATION);
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -503,6 +512,59 @@ protected void setOptimizeForIndexFilters(boolean shouldOptimizeForIndexFilters)
...
+        final RecordQueryPlan deserializedPlan =
+                RecordQueryPlan.fromRecordQueryPlanProto(serializationContext, parsedPlanProto);
+        Assertions.assertEquals(plan.planHash(PlanHashable.CURRENT_FOR_CONTINUATION), deserializedPlan.planHash(PlanHashable.CURRENT_FOR_CONTINUATION));
+        Assertions.assertTrue(plan.structuralEquals(deserializedPlan));
+        return deserializedPlan;
```
alecgrieser, line 565: One interesting quirk here is that this means that we are always running the deserialized plans from Cascades. Theoretically, this means that we could face a bug where, for instance, a plan is broken unless we first serialize and deserialize it (e.g., there's some state that gets set up correctly during plan deserialization but is in an uninitialized state before). I don't really have a _great_ solution for it, though. We could randomly return the serialized or deserialized plan, though that wouldn't be repeatable. We could modify `executeQuery` so that it executes the plan both before serialization and after serialization and validates the same results get back, or something, though that's also a bit iffy.

This shouldn't be a concern for now, because the tests were already validating that the non-serialized plan works, but it could be a problem in the future

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenePlanSerializationTest.java
```diff
@@ -0,0 +1,74 @@
...
+                        COMPLEX_MULTIPLE_TEXT_INDEXES_WITH_AUTO_COMPLETE_STORED_FIELDS);
+        PlanSerializationContext planSerializationContext = PlanSerializationContext.newForCurrentMode();
+        final RecordQueryPlanProto.PPlanReference proto = planSerializationContext.toPlanReferenceProto(plan);
+        final byte[] bytes = proto.toByteArray();
+        final RecordQueryPlanProto.PPlanReference parsedProto =
+                RecordQueryPlanProto.PPlanReference.parseFrom(bytes);
+        planSerializationContext = PlanSerializationContext.newForCurrentMode();
+        final RecordQueryPlan parsedPlan = planSerializationContext.fromPlanReferenceProto(parsedProto);
+        Verify.verify(parsedPlan instanceof LuceneIndexQueryPlan);
+        // TODO proper verification as we currently do not properly round-trip;
```
alecgrieser, line 72: This is currently not as much of a practical problem as it would otherwise be because we do Lucene planning with the old planner, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PushFilterThroughFetchRule.java
normen662: I removed support for this kind of predicate which was long overdue. Didn't need serialization for it so rater removed it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/DefaultPlanSerializationRegistry.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerializationRegistry.java
normen662: marked them as `EXPERIMENTAL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
normen662: done

alecgrieser (approved):

PR #2453, opened 2024-01-19
Resolves #2452: BooleanNormalizer does not handle NestedFields

### LINKED ISSUE #2452: BooleanNormalizer does not handle NestedFields
The `BooleanNormalizer` currently does not attempt to normalize `NestedFields` (except at leaf level). So, for example, if given the statement :

```java
Query.and(
    Query.field("a").equalsParameter("aParam"),
    Query.field("b").matches(
        Query.or(
            Query.field("c").equalsParameter("cParam"),
            Query.field("d").equalsParameter("dParam")
        )
    )
)
```

Then that can be transformed into DNF as:

```java
Query.or(
    Query.and(
        Query.field("a").equalsParameter("aParam"),
        Query.field("b").matches(Query.field("c").equalsParameter("cParam"))
    ),
    Query.and(
        Query.field("a").equalsParameter("aParam"),
        Query.field("b").matches(Query.field("d").equalsParameter("dParam"))
    )
)
```

But the `BooleanNormalizer` will currently ignore this transformation and will instead choose to leave the statement as is because it will never explore down the `NestedField` (e.g., the predicate on `Query.field("b"))`).

### LINKED ISSUE #2461: RecordQueryPlannerConfiguration should be proto serializable
It would be nice if `RecordQueryPlannerConfiguration` were proto serializable. This would allow the planner configuration to be included with continuations, which could then inform the planner configuration to use when planning a query. This would allow us to more safely roll out planner configuration changes, because when a configuration property is changed, instances would follow the configuration value from the original time the query was planned.

### LINKED ISSUE #2463: PlanOrderingKey.mergeComparisonKeys can incorrectly normalize ListKeyExpressions
The `ListKeyExpression` is like `ThenKeyExpression`, except that each element of a `ListKeyExpression` becomes a single column in the final expression. That means that when we normalize the expression for positions, we _don't_ expand children, which is how a `ThenKeyExpression` works.

The `PlanOrderingKey.mergeComparisonKeys` function can ignore that and produce a `PlanOrderingKey` which tries to normalize its child components even if one of them comes from a `ListKeyExpression`. We should preserve those single-column fields, I think. In practice, this can result in `SyntheticRecordType`s not correctly matching for ordered unions if one of their constituent types has a multi-column primary key.

alecgrieser:
This combines a few different planner changes related to supporting an `OR` query on an `UnnestedRecordType`.

* The `BooleanNormalizer` now can normalize `NestedField` expressions, allowing for more expressions to be placed into DNF, which in turn allows us to match more plans to ordered unions or intersections than before. I've added in the `FDBOrQueryToUnionTest` class examples of queries that previously would not have been placed into DNF form (at least by the old planner). The new planner appears to already be selecting the right plans in these cases. This resolves #2452.
* Because normalization can change plans, I added a configuration property to the `RecordQueryPlannerConfiguration` to determine whether this normalization should be applied or not. I was concerned a bit about the roll out here, so I added a proto serialization of the class that an adopter could attach to a continuation so that queries can be re-planned with the same continuation. This resolves #2461.
* The `PlanOrderingKey` may be tripped up if the same field appears in both the equality bound prefix and in the other set of keys. This could happen if an equality bound field was also in the record primary key, for example. This is related to #2462, though the PR as written does not handle if the same field appears twice in the _non_-equality bound section, so I'm not sure we can close that issue out (yet)
* Because `UnnestedRecordType`s use `ListKeyExpression`s, the construction of the final ordering key (from merging different components together) could be wrong if one of the constituents has a multi-column primary key. In addition to producing a final ordering that wasn't _quite_ right, it also could make matching the primary key incorrect during ordering. This fixes #2463 by (sometimes) choosing to combine the sub-expressions with a `ListKeyExpression`


fdb-extensions/src/test/java/com/apple/foundationdb/async/rtree/RTreeModificationTest.java
```diff
@@ -57,6 +57,7 @@
  * Tests testing insert/update/deletes of data into/in/from {@link RTree}s.
  */
 @Tag(Tags.RequiresFDB)
+@Tag(Tags.Slow)
```
alecgrieser, line 60: While running tests locally, I found it expedient to run tests with `SKIP_SLOW_TESTS=true ./gradlew test`. I've marked some tests as `Slow` if they exceeded the 2 second limit that we now apply

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -22,83 +22,59 @@
...
 import javax.annotation.Nullable;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * A set of configuration options for the {@link RecordQueryPlanner}.
  */
 @API(API.Status.MAINTAINED)
 public class RecordQueryPlannerConfiguration {
```
alecgrieser, line 44: This class has been rewritten so that it doesn't have storage of its own, it instead uses a backing proto message. The `Builder`, likewise, updates a proto format. This makes the `toProto` method easy, and it should be transparent to anyone not using the proto serialization.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/sorting/RecordQueryPlannerSortConfiguration.java
```diff
@@ -49,11 +59,15 @@ protected RecordQueryPlannerSortConfiguration() {
      * This method returns {@code true}.
      *
      * A subclass can override this method to use more complex criteria to decide.
-     * @param query the query that cannot be planned without a sort
      * @return whether to allow non-index sorting
      */
-    public boolean shouldAllowNonIndexSort(@Nonnull RecordQuery query) {
-        return true;
+    public boolean shouldAllowNonIndexSort() {
```
alecgrieser, line 64: I think the intention here was to allow for subclasses with more complex decision trees (based on the query). This becomes hard to serialize, so I think we want to instead encourage users to look at their query, and then based on that, decide whether to allow non-index sorts or not. I couldn't really find a great way of serializing this information otherwise.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/PlanOrderingKey.java
```diff
@@ -65,13 +68,24 @@ public class PlanOrderingKey {
     private final int prefixSize;
     private final int primaryKeyStart;
     private final int primaryKeyTail;
+    @Nonnull
+    private final Set<Integer> equalityBound;
```
MMcM, line 72: I believe these are _later duplicates_ of equality bound keys. Perhaps the name could reflect that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/planning/BooleanNormalizer.java
```diff
@@ -275,26 +319,37 @@ private List<List<QueryComponent>> orToDNF(@Nonnull final List<QueryComponent> c
...
+                toDNF(children.get(index), negate, parentPath).stream().flatMap(right -> crossProductSoFar.stream().map(left -> {
                     final List<QueryComponent> combined = new ArrayList<>(left);
                     combined.addAll(right);
                     return combined;
                 })).collect(Collectors.toList()));
     }
 
+    @Nonnull
+    private List<List<QueryComponent>> nestedFieldToDNF(@Nonnull final NestedField nestedField, final boolean negate, @Nonnull final Deque<String> parentPath) {
+        // And the parent field to the parent path. Pop from the stack after normalizing the children
```
MMcM, line 345:
```suggestion
        // Add the parent field to the parent path. Pop from the stack after normalizing the children
```

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/PlanOrderingKey.java
```diff
@@ -65,13 +68,25 @@ public class PlanOrderingKey {
...
         this.primaryKeyTail = primaryKeyTail;
+
+        ImmutableSet.Builder<Integer> equalityBoundBuilder = ImmutableSet.builder();
+        for (int pos = prefixSize; pos < keys.size(); pos++) {
+            final KeyExpression component = keys.get(pos);
+            if (keys.indexOf(component) < prefixSize) {
+                equalityBoundBuilder.add(pos);
+            }
+        }
+        this.equalityBoundInTail = equalityBoundBuilder.build();
```
pengpeng-lu, line 89: Could you add some comments here? Also, you mentioned that `though the PR as written does not handle if the same field appears twice in the non-equality bound section, so I'm not sure we can close that issue out (yet)` in the PR summary part. My understanding is that if in an unnested query, if we `order by columnA, columnB`, we could have the same prefix field appears twice in the non-equality bound section? why doesn't the current modification solve this case?

alecgrieser: > My understanding is that if in an unnested query, if we order by columnA, columnB, we could have the same prefix field appears twice in the non-equality bound section?

In theory, yes, this could happen. It would require an index which contained the same component multiple times. I was thinking that this could happen if one of the primary key components was also in the index. (That's generally normalized out, though I thought the `ListKeyExpression` used by synthetic types was complicating that.)

> why doesn't the current modification solve this case?

I'm not sure if your question is "what about this code makes it so that this case isn't handled?" or "why did you decide to write code that didn't handle this case?".

If the former, the problem is that if the _first_ occurrence of a key expression occurs in the non-equality bound tail, then when we go to match it because we'll calculate the index of the position based on its _first_ occurrence, which won't match the expected `nextNonPrefix` value. That's not a problem for equality bound keys, because we check if the key is equality bound first.

If the latter, the reason I didn't handle it was because I was working off of an example where (I thought) a component was equality bound, and so I wanted to make sure that that case works. (As it happens, I'm actually less sure that that was necessary, because the real problem was that the equality-bound component was being normalized in a bit of weird way.) Keeping track of the way the tail works is a bit tricky, and so I didn't want to do that if I didn't develop tests for it. I started going about fixing it, and I think it might actually not be too bad (because, I think, it may be sufficient to check only that the component's position is `< nextNonPrefix`), but I would prefer to get this in first, if possible.

MMcM (approved):

PR #2464, opened 2024-01-24
Reduce unnecessary memory allocations in Cascades

hatyo:
This is mainly refactoring to improve memory allocation in Cascades.

- `Value#getChildren()` is now amortized. Previously, many implementations were calculating the list of children each time the method is invoked, but this is not necessary considering the immutability of `Value`.
- `TreeLike#inPreOrder(Predicate<>)` is slightly incorrect since the predicate does not propagate in recursive call, but more importantly, it is inefficient in terms of memory consumption, considering that it creates a list for _each_ node of the tree each time it is invoked, the lists are accessed via a complex hierarchy of concatenated iterators. These small object allocations are unnecessary and could stress the garbage collection if it is done very frequently which is the case since this method is used indirectly e.g. to understand a `Predicate` structure during index matching.
In this PR a new implementation that is more efficient is introduced. It employs an algorithm that mixes level-based and preorder-based traversal. It uses a stack which holds a reference to the _entire_ list of children and an index to bookkeep the traversed node. It is exposed as a normal `Iterator<>`.
- With the new `Iterator`, the `TreeLike` now extends `Iterable<>` interface giving an instance of this iterator when `iterator()` method is called.
- Similar to `Collection.stream()`, a `stream()` method is added that returns a `Stream` object which uses the above `Iterator` implementation.
I noticed that in almost all callsites, `TreeLike#inPreOrder`'s allocated `Iterable` is _not_ used, i.e. it is mostly used for stream-like interface filtering, therefore, replacing these calls with the `stream()` call implemented above is probably more efficient as it does not allocate as many objects. It also streamlines the code better.
- a `height()` method is introduced in `TreeLike` that returns the max-height of the tree, the implementations of this method are amortized, it is mainly used to set the capacity of the pre-order iterator stack so we can avoid yet other potential reallocations needed for resizing.

I kept the `inPreOrder` overloads, and `filter` but marked them as deprecated in favour of using `stream()` (and `Stream#filter()`). But I am totally fine if we want to remove them immediately.

To compare the new implementation of pre-order with the older one, I implemented a simple JMH Benchmark in a separate draft PR #2467, please check it out if you want to get more performance details.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanHashable.java
```diff
@@ -244,6 +244,9 @@ static int objectPlanHash(@Nonnull final PlanHashMode mode, @Nullable final Obje
         if (mode.getKind() != PlanHashKind.LEGACY && obj instanceof Map) {
             return mapsPlanHash(mode, (Map<?, ?>)obj);
         }
+        if (obj instanceof PlanHashable) {
+            return ((PlanHashable)obj).planHash(mode);
+        }
```
hatyo, line 249: Changing the order here is necessary, because now a `Value` is actually an `Iterable`, therefore, it will match line 250.

If we do not change the order of lines, the following will happen:
1. a `Value` such as `RCV(x,y,z)` will match at line 250, calls into `iterablePlanHash`.
2. the method `iterablePlanHash` will iterate over the passed iterable object
3. the _first_ item in the `RCV(x,y,z)` is the `RCV` itself (because of pre-order and the recursive nature of the non-linear structure `TreeLike<>`.
4. it calls now in `objectPlanHash`
5. it goes back to 1 (!)

-> infinite recursion.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NotValue.java
```diff
@@ -99,7 +105,7 @@ public Optional<QueryPredicate> toQueryPredicate(@Nullable final TypeRepository
     @Nonnull
     @Override
     public Iterable<? extends Value> getChildren() {
-        return ImmutableList.of(child);
+        return childrenSupplier.get();
```
normen662, line 108: All `Value`s extend from `AbstractValue`. Can you put all suppliers in there, define an abstract `computeChildren()` and have that memoized.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -83,8 +87,11 @@ default Iterable<? extends T> inPreOrder() {
      * @param descentIntoPredicate a predicate that determines of the traversal enters the subtree underneath the
      *        current node
      * @return an {@link Iterable} of nodes
+     * @deprecated this is deprecated for performance reasons, to apply a filter, get a {@link TreeLike#iterator()} which
+     *             returns an {@link Iterator} with pre-order traversal.
      */
     @Nonnull
+    @Deprecated
```
normen662, line 94: Please have this use your much better iterator.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
```diff
@@ -77,4 +85,13 @@ private int computeSemanticHashCode() {
         return fold(Value::hashCodeWithoutChildren,
                 (hashCodeWithoutChildren, childrenHashCodes) -> Objects.hash(childrenHashCodes, hashCodeWithoutChildren));
     }
+
+    @Override
+    public int height() {
```
normen662, line 90: Why do we need a `height()`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AndOrValue.java
```diff
@@ -68,6 +70,8 @@ public class AndOrValue extends AbstractValue implements BooleanValue {
     @Nonnull
     private final Value rightChild;
     @Nonnull
+    private final Supplier<Iterable<? extends Value>> childrenSupplier = Suppliers.memoize(this::computeChildren);
```
normen662, line 73: Again, please condense into `AbstractValue`

normen662 (changes requested):

normen662:
Hey, I like the iterator a lot, it's much more performant. Also, quite a bit of resolve to properly profile it. I have not looked into the `replace()` changes yet as I hit submit too early. Will do that tomorrow.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanHashable.java
hatyo: I am assuming the plan hash has enough test coverage such that if the PRB is green (which is), then plan hash should not have been changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -35,13 +34,15 @@
...
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * Interface for tree-like structures and helpers for folds and traversals on these tree-like structures of node type
  * {@link T}.
  * @param <T> type parameter of the node
  */
-public interface TreeLike<T extends TreeLike<T>> {
+public interface TreeLike<T extends TreeLike<T>> extends Iterable<T> {
```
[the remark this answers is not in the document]
hatyo: ok, removed the extension of `Iterable`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/PlanHashable.java
hatyo: Restored the code now that `TreeLike` does not extend `Iterable`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/NotValue.java
hatyo: I think I wanted to avoid using a `Supplier` where it is superfluous to do so. But what you suggest is cleaner, so I refactored the code accordingly.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AndOrValue.java
hatyo: Done, I also refactored `ValueWithChild` implementations to avoid extra allocation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
hatyo: I mentioned why I need it in the PR message:
> a height() method is introduced in TreeLike that returns the max-height of the tree, the implementations of this method are amortized, it is mainly used to set the capacity of the pre-order iterator stack so we can avoid yet other potential reallocations needed for resizing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -69,13 +70,27 @@ public interface TreeLike<T extends TreeLike<T>> {
...
+     * Returns an iterator that traverse the nodes in pre-order.
+     * @return an iterator that traverse the nodes in pre-order.
+     */
+    @Nonnull
+    default Iterator<T> preOrderIterator() {
+        return PreOrderIterator.over(getThis());
+    }
+
+    @Nonnull
+    default Stream<T> stream() {
```
normen662, line 83: Please rename to `preOrderStream()` and then have all 50 callers use this instead of `Streams.stream(preOrderIterator())......`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -252,6 +255,8 @@ default Optional<T> replaceLeavesMaybe(@Nonnull final UnaryOperator<T> replaceOp
         });
     }
 
+    int height();
```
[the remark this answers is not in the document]
normen662: Can't we just implement `height()` as a static `TreeLike` method() that this method calls. That uses a `fold()` as there is nothing specific that `height()` needs from `Value` or `QueryPredicate` or such. If you then need to memoize that you can by defining a height() method in the particular `AbstractValue` or similar with a memoizing supplier.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
normen662: The max height is like `4` or `5`. I feel this `height()` method is too specific for a `TreeLike`. Please refer to my comment further up. We can compute the height as a static method in treelike that is not memoized and then define a `height()` on e.g. `AbstractValue` that memoizes and calls that `TreeLike.height()` static method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
```diff
@@ -77,4 +89,22 @@ private int computeSemanticHashCode() {
...
+        return heightSupplier.get();
+    }
+
+    private int computeHeight() {
+        return StreamSupport.stream(getChildren().spliterator(), false).mapToInt(TreeLike::height).max().orElse(0) + 1;
```
normen662, line 99: `Streams.stream(...)` please.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ExistsValue.java
```diff
@@ -54,32 +54,25 @@
...
+        super(child);
     }
 
     @Override
     @SuppressWarnings({"java:S2637", "ConstantConditions"}) // TODO the alternative component should not be null
     @SpotBugsSuppressWarnings("NP_NONNULL_PARAM_VIOLATION")
     public Optional<QueryPredicate> toQueryPredicate(@Nullable final TypeRepository typeRepository,
                                                      @Nonnull final CorrelationIdentifier innermostAlias) {
-        return Optional.of(new ExistsPredicate(child.getAlias()));
+        return Optional.of(new ExistsPredicate(((QuantifiedObjectValue)getChild()).getAlias()));
```
normen662, line 69: AbstractValueWithChild is defining a child but there is no proto message reflecting that. Thus every extending class needs to deal with the child in their own `toProto` and `fromProto` which is something we absolutely want to avoid as a 1-1 relationship between proto and pojo is necessary to not go insane.
So you can either define a proto message `PAbstractValueWithChild` and use the usual pattern with a  `super`  field for all extending classes, or you keep these super classes interfaces for now.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ExistsValue.java
hatyo: reverted interface-to-abstract-class changes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PreOrderIterator.java
```diff
@@ -0,0 +1,113 @@
...
+ * each stack frame the entire list of node's children is stored instead in addition to an index pointing out
+ * to the current element that is returned by the stream.
+ *
+ * @param <T> The type of the iterator element.
+ */
+@NotThreadSafe
+public final class PreOrderIterator<T extends TreeLike<T>> implements Iterator<T> {
+
+    @Nonnull
+    private final Deque<Pair<Iterable<? extends T>, Integer>> stack;
```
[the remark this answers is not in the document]
hatyo: Right, but I don't want to do it in this PR to limit its scope, there are 80+ locations where we still use `Pair` currently, we can do this in a separate cleanup PR. (I am also using `MutablePair` by the way, but we can also replace that in a cleanup PR). https://github.com/FoundationDB/fdb-record-layer/issues/2484

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
hatyo: Done, moved `height` to `TreeLike`, `AbstractValue` memoizes the calculation (which is delegated to the implementation in `TreeLike`. No need to override in children because the method in `TreeLike` is very quick as it uses the (now memoized) children's iterator to calculate the height.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
hatyo: Done. Method is refactored and sits now in `TreeLike`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
hatyo: Done. Check my reply above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
hatyo: Method is replaced with a normal `Stream#filter` as now `TreeLike` can return a `Stream` via `TreeLike#preOrderStream()` method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/AbstractQueryPredicate.java
```diff
@@ -56,6 +58,7 @@ protected AbstractQueryPredicate(final boolean isAtomic) {
         this.isAtomic = isAtomic;
         this.correlatedToSupplier = Suppliers.memoize(this::computeCorrelatedTo);
         this.semanticHashCodeSupplier = Suppliers.memoize(this::computeSemanticHashCode);
+        this.heightSupplier = Suppliers.memoize(QueryPredicate.super::height);
```
normen662, line 61: I like that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractValue.java
normen662: Me likey!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
```diff
@@ -83,6 +84,12 @@ public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
         return Set.of();
```
normen662, line 84: Can you please make this a `ImmutableSet.of()`. You didn't change it here, I just stumbled over it.

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
hatyo: ok, I will do as part of solving #2483

PR #2466, opened 2024-01-25
RESOLVES #2465: Support rearranging documents between Lucene partitions

### LINKED ISSUE #2465: Support rearranging documents between Lucene partitions

wshaib-apple:
- before index merge, partitions are examined for the count of documents they contain. If above a high watermark, the oldest N documents will be moved to an older partition. If no older partition exists, or if the N documents would make the older partition go over the high watermark, a new partition will be create and docs moved there.
- N is configurable via a property. It should be small enough so that the entire move can be performed in a single transaction. It is currently set to 10.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -273,7 +273,7 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
     }
 
     @SuppressWarnings({"PMD.CloseResource", "java:S2095"})
-    private void deleteDocument(Tuple groupingKey, Integer partitionId, Tuple primaryKey) throws IOException {
+    void deleteDocument(Tuple groupingKey, Integer partitionId, Tuple primaryKey) throws IOException {
         final long startTime = System.nanoTime();
         final IndexWriter indexWriter = directoryManager.getIndexWriter(groupingKey, partitionId, indexAnalyzerSelector.provideIndexAnalyzer(""));
         @Nullable final LucenePrimaryKeySegmentIndex segmentIndex = directoryManager.getDirectory(groupingKey, partitionId).getPrimaryKeySegmentIndex();
@@ -311,7 +311,8 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
 
     @Override
     public CompletableFuture<Void> mergeIndex() {
-        return directoryManager.mergeIndex(partitioner, indexAnalyzerSelector.provideIndexAnalyzer(""));
+        return partitioner.rebalancePartitions()
+                .thenCompose(ignored -> directoryManager.mergeIndex(partitioner, indexAnalyzerSelector.provideIndexAnalyzer("")));
     }
 
     @Nonnull
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
@@ -26,29 +26,49 @@ import com.apple.foundationdb.StreamingMode;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncIterable;
 import com.apple.foundationdb.async.AsyncUtil;
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.ExecuteProperties;
+import com.apple.foundationdb.record.KeyRange;
+import com.apple.foundationdb.record.PipelineOperation;
 import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
+import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.ScanProperties;
+import com.apple.foundationdb.record.cursors.ChainedCursor;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
+import com.apple.foundationdb.record.lucene.directory.FDBDirectoryManager;
 import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.metadata.RecordType;
 import com.apple.foundationdb.record.metadata.expressions.FieldKeyExpression;
+import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexableRecord;
+import com.apple.foundationdb.record.provider.foundationdb.FDBIndexedRecord;
 import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
+import com.apple.foundationdb.record.provider.foundationdb.IndexOrphanBehavior;
+import com.apple.foundationdb.record.query.expressions.Comparisons;
+import com.apple.foundationdb.record.query.plan.ScanComparisons;
+import com.apple.foundationdb.subspace.Subspace;
 import com.apple.foundationdb.tuple.Tuple;
+import com.google.common.base.Verify;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
-import static com.apple.foundationdb.record.lucene.LuceneIndexOptions.INDEX_PARTITION_BY_TIMESTAMP;
 import static com.apple.foundationdb.record.metadata.Key.Expressions.field;
 import static com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer.Waits.WAIT_LOAD_LUCENE_PARTITION_METADATA;
 
@@ -63,15 +83,20 @@ public class LucenePartitioner {
     private final IndexMaintainerState state;
     private final boolean partitioningEnabled;
     private final String partitionTimestampFieldName;
+    private final int indexPartitionHighWatermark;
     private final KeyExpression partitioningKeyExpression;
 
     public LucenePartitioner(@Nonnull IndexMaintainerState state) {
         this.state = state;
-        partitionTimestampFieldName = state.index.getOption(INDEX_PARTITION_BY_TIMESTAMP);
+        partitionTimestampFieldName = state.index.getOption(LuceneIndexOptions.INDEX_PARTITION_BY_TIMESTAMP);
         this.partitioningEnabled = partitionTimestampFieldName != null;
         if (partitioningEnabled && (partitionTimestampFieldName.isEmpty() || partitionTimestampFieldName.isBlank())) {
             throw new RecordCoreArgumentException("Invalid partition timestamp field name", LogMessageKeys.FIELD_NAME, partitionTimestampFieldName);
         }
+        String strIndexPartitionHighWatermark = state.index.getOption(LuceneIndexOptions.INDEX_PARTITION_HIGH_WATERMARK);
+        indexPartitionHighWatermark = strIndexPartitionHighWatermark == null ?
+                                          Objects.requireNonNull(state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_PARTITION_HIGH_WATERMARK)) :
+                                          Integer.parseInt(strIndexPartitionHighWatermark);
         this.partitioningKeyExpression = makePartitioningKeyExpression(partitionTimestampFieldName);
     }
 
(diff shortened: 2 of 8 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -458,4 +483,194 @@ public static long getFrom(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo
...
+     * partition needing re-balancing will be processed. If there are other partitions that need re-balancing,
+     * they will be processed during subsequent calls.
+     *
+     * @param groupingKey grouping key
+     * @return void future
+     */
+    @Nonnull
+    private CompletableFuture<Void> processPartitionRebalancing(@Nonnull final Tuple groupingKey) {
+
+        return getAllPartitionMetaInfo(groupingKey).thenCompose(partitionInfos -> {
```
wshaib-apple, line 540: Instead of having to get the metadata entries of all partitions (in order to determine the max partition id, and the document count of each partition), perhaps an optimization would be to create a "summary" entry that has the max id as well as a map of partition metadata key -> doc count.
Thoughts?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -581,28 +598,198 @@ void basicGroupedPartitionedTest() {
             validateSegmentAndIndexIntegrity(COMPLEX_PARTITIONED, partition1Subspace, context, "_0.cfs");
             validateSegmentAndIndexIntegrity(COMPLEX_PARTITIONED, partition2Subspace, context, "_0.cfs");
         }
+    }
+
+    @Test
+    void repartitionGroupedTest() throws ExecutionException, InterruptedException, InvalidProtocolBufferException {
```
ScottDugas, line 604: It seems like there should be a test without grouping

ScottDugas, line 604: We should probably add a test with a bunch of records with the same timestamp. Or note it as an outstanding issue.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -581,28 +598,198 @@ void basicGroupedPartitionedTest() {
...
+
+        final int totalDocCount = 20;
+        Consumer<FDBRecordContext> schemaSetup = context -> rebuildIndexMetaData(context, COMPLEX_DOC, COMPLEX_PARTITIONED);
+
+        // create/save documents
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            long start = Instant.now().toEpochMilli();
+            for (int i = 0; i < totalDocCount; i++) {
+                recordStore.saveRecord(createComplexDocument(1000L + i, ENGINEER_JOKE, 1, start + i * 100));
```
ScottDugas, line 618: It would probably be good to not have the order of the ids perfectly align with the timestamp.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -581,28 +598,198 @@ void basicGroupedPartitionedTest() {
...
+        final int totalDocCount = 20;
+        Consumer<FDBRecordContext> schemaSetup = context -> openRecordStore(context, LuceneIndexTest::joinedPartitionedLuceneIndexMetadataHook);
+
+        // create/save documents
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            long start = Instant.now().toEpochMilli();
+            for (int i = 0; i < totalDocCount; i++) {
+                TestRecordsTextProto.ComplexDocument cd = TestRecordsTextProto.ComplexDocument.newBuilder()
+                        .setGroup(42)
```
ScottDugas, line 707: We should also have a test with multiple groups, or have multiple groups here

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -581,28 +598,198 @@ void basicGroupedPartitionedTest() {
...
 
+        // initially, all documents are saved into one partition
+        List<LucenePartitionInfoProto.LucenePartitionInfo> partitionInfos = getPartitionMeta(JOINED_INDEX,
+                Tuple.from(42), contextProps, schemaSetup);
+        assertEquals(1, partitionInfos.size());
+        assertEquals(totalDocCount, partitionInfos.get(0).getCount());
+
+        // run re-partitioning
+        explicitMergeIndex(JOINED_INDEX, contextProps, schemaSetup);
+        partitionInfos = getPartitionMeta(JOINED_INDEX,
```
ScottDugas, line 730: It seems like this should also assert that the right documents end up in the right group

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -581,28 +598,198 @@ void basicGroupedPartitionedTest() {
...
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+
+            byte[] continuation = validateLastDocumentInPartition(null, 0, 7, Tuple.from(1L, 1012));
+            continuation = validateLastDocumentInPartition(continuation, 2, 5, Tuple.from(1L, 1006));
+            validateLastDocumentInPartition(continuation, 1, 5, Tuple.from(1L, 1000));
+        }
+    }
+
+    private byte[] validateLastDocumentInPartition(byte[] continuation, int expectedPartitionId, int amountToSkip, Tuple expectedDocId)
```
ScottDugas, line 668: I wonder if it would be worthwhile to have a search that explicitly fetches the `IndexWriter` from the group/partition and makes sure that searching for everything in there finds the set of documents that should be there.

ScottDugas (changes requested):
There's definitely more tests to be added, let's discuss offline exactly what we want to bring in here, or do as a followup.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: you mean `IndexReader`?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -659,39 +712,46 @@ void repartitionGroupedTest() throws ExecutionException, InterruptedException, I
...
+                return Tuple.fromBytes(primaryKey.binaryValue().bytes);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }).collect(Collectors.toSet()));
     }
 
     @Test
-    void repartitionSyntheticGroupedTest() {
+    void repartitionSyntheticGroupedTest() throws IOException {
```
ScottDugas, line 753: This test could also benefit from using `validateDocsInPartition`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -311,7 +311,8 @@ private void deleteDocument(Tuple groupingKey, Integer partitionId, Tuple primar
 
     @Override
     public CompletableFuture<Void> mergeIndex() {
-        return directoryManager.mergeIndex(partitioner, indexAnalyzerSelector.provideIndexAnalyzer(""));
+        return partitioner.rebalancePartitions()
```
ScottDugas, line 314: I didn't realize this in the first pass, but this doesn't play well with `AgilityContext`, namely we will rebalance the partition, and then, the `AgilityContext` will create a new transaction to do merges, and, if this partition ever commits it will fail.
Let's leave this as is, and followup with a better transaction story.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -458,4 +483,207 @@ public static long getFrom(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo
...
+        final CompletableFuture<? extends List<? extends FDBIndexableRecord<Message>>> fetchedRecordsFuture;
+        if (recordTypes.iterator().next().isSynthetic()) {
+            fetchedRecordsFuture = cursor.mapPipelined(indexEntry -> state.store.loadSyntheticRecord(indexEntry.getPrimaryKey()),
+                    state.store.getPipelineSize(PipelineOperation.INDEX_TO_RECORD)).asList();
+        } else {
+            fetchedRecordsFuture = state.store.fetchIndexRecords(cursor, IndexOrphanBehavior.SKIP).map(FDBIndexedRecord::getStoredRecord).asList();
+        }
+
+        return fetchedRecordsFuture.thenCompose(records -> {
+            if (records.size() > 0) {
```
MMcM, line 634: `> 1`, right? There needs to be one left and at least one to move.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -458,4 +483,207 @@ public static long getFrom(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo
...
+
+                // update current partition's meta
+                state.context.ensureActive().clear(partitionMetadataKeyFromTimestamp(groupingKey, getFrom(partitionInfo)));
+                LucenePartitionInfoProto.LucenePartitionInfo.Builder builder = partitionInfo.toBuilder()
+                        .setCount(partitionInfo.getCount() - records.size())
+                        .setFrom(ByteString.copyFrom(Tuple.from(newBoundaryTimestamp).pack()));
+                savePartitionMetadata(groupingKey, builder);
+
+                // value of the "destination" partition's `from` timestamp
+                final long overflowPartitionFromTimestamp = getPartitioningTimestampValue(records.get(0));
```
MMcM, line 662: What happens if there are ties? In particular, if some after the end of the cursor have the same timestamp? If this is a failure of the caller to obey the contract of the timestamp data can we at least detect that?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -458,4 +483,207 @@ public static long getFrom(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo
...
+                return findPartitionInfo(groupingKey, overflowPartitionFromTimestamp).thenCompose(previousPartition -> {
+                    if (previousPartition == null || previousPartition.getCount() + records.size() > indexPartitionHighWatermark || previousPartition.getId() == partitionInfo.getId()) {
+                        // create a new "overflow" partition
+                        savePartitionMetadata(groupingKey, newPartitionMetadata(overflowPartitionFromTimestamp, maxPartitionId + 1).toBuilder());
+                    }
+
+                    Iterator<? extends FDBIndexableRecord<Message>> recordIterator = records.iterator();
+                    return AsyncUtil.whileTrue(() -> indexMaintainer.update(null, recordIterator.next())
+                            .thenApply(ignored -> recordIterator.hasNext()));
```
MMcM, line 671: It's not a bug, but the layering here feels a little uncomfortable.

Partitions are below the index maintainer: it accepts / returns records and divides work to the appropriate ones.
But the re-partitioner is _above_ the index maintainer. This means (1) you ask the maintainer to delete / update the record and (2) you modify the partition metadata at just the right time and then re-enter records in the same way.
Before seeing this, I would have imagined a partition system below the index maintainer (and above Lucene & FDB) and the re-partitioner as a _sibling_ of the maintainer calling into partitions to remove & add.

I guess I am not actually asking to have anything changed, just capturing my reaction.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
ScottDugas: Yes, @wshaib-apple and I had discussed this, and it doesn't support that. it probably makes sense to at least throw an error at this point, and that should be fairly easy and worth bringing in now.
It would also be pretty easy to support any situation where you have fewer than `LuceneRecordContextProperties.LUCENE_REPARTITION_DOCUMENT_COUNT` documents with the same timestamp. 
To support the broader case my two thoughts (right now) would be to:
1. Add the primary key to the `from` so that you're guaranteed to have unique entries.
2. Allow the from of one partition to be the same as the to in the other and do something else to ensure that each document is only in one partition.

If you have more than a very small ratio of `LuceneIndexOptions.INDEX_PARTITION_HIGH_WATERMARK` to documents with the same timestamp, the whole feature kind of falls over, but I don't think we need to require the timestamp to be unique.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
ScottDugas: I had a similar feeling, somewhere else it has to get the index maintainer.
I think the answer is that we need to take some of the `IndexMaintainer` and push it down, so that the `LucenePartitioner` can call `update` on that.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ScottDugas: I still think it would be good to add `validateDocsInPartition` here, but it might make sense to first add a version that asserts the constraints of partitions, rather than having to explicitly declare which documents are in which partition.

ScottDugas (approved):

MMcM (approved):

PR #2468, opened 2024-01-26
Fixes #2462: PlanOrderingKey tripped up if the same expression appears multiple times

### LINKED ISSUE #2462: PlanOrderingKey tripped up if the same expression appears multiple times
If the `PlanOrderingKey` contains the same expression appears multiple times, it can hit trouble when it tries to check that key. For example, if the field is in both in the equality-bound prefix and appears again after, then there is no need to put it in the final `comparisonKey`, and the indexing into the `PlanOrderingKey` can be misaligned.

### LINKED ISSUE #2469: PlanOrderingKey contains value portion of key-with-value-expression in order
The current `PlanOrderingKey` implementation will include the value components of a `KeyWithValueExpression` within the professed ordering for a plan. This can result in plans that have an incorrect idea as to the order of returned results, which can lead to faulty unions and intersections. From what I can tell, this is not as much of a problem for a single index scan, as that may not need to look at the `PlanOrderingKey`.

alecgrieser:
This updates the handling for key expressions in `PlanOrderingKey` so that if the same expression appears multiple times, we only match it once.

This fixes #2462. That work was started in #2453, but it still had some edge cases where if none of the key expressions were in the equality prefix, the logic wasn't quite right.

I also added (in a separate commit) logic that fixes #2469. This is here mainly to avoid too many conflicts from updating the same bit of code, though it's conceptually separate.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/PlanOrderingKey.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/PlanOrderingKey.java
@@ -68,9 +68,9 @@ public class PlanOrderingKey {
     private final int prefixSize;
     private final int primaryKeyStart;
     private final int primaryKeyTail;
-    // Keep a set of positions that are equality bound but are not in the prefix. (All components in the prefix are equality bound.)
+    // Keep a set of positions that are duplicates of previous key components
     @Nonnull
-    private final Set<Integer> equalityBoundInTail;
+    private final Set<Integer> duplicatePositions;
 
     public PlanOrderingKey(@Nonnull List<KeyExpression> keys, int prefixSize,
                            int primaryKeyStart, int primaryKeyTail) {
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/PlanOrderingKey.java
```diff
@@ -283,10 +286,18 @@ private static KeyExpression orderingCompatiblePlanKey(@Nonnull PlanOrderingKey
...
+                    //    case, then we're trivially ordered by it
+                    // 2. This key component appears in the non-equality bound tail, and we've already
+                    //    matched this key.
+                    // In either case, the ordering key is already satisfied, so we can add it to the
+                    // final component list.
+                    //
+                    // Note that we can get an already bound equality predicate if when combining sub-plans,
+                    // we need to slot in all the results from one sub-plan before or after the elements of
+                    // another sub-plan in order to satisfy the plans overall ordering requirement.
                     components.add(component);
```
alecgrieser, line 301: This is the key, and I'm not sure if I've explained it correctly. The basic conceit here is that if the key component is less than `nextNonPrefix`, then either:

1. The key component is equality bound, and so we're the same as before. All results in the plan will have the same value for the expression, so we're "ordered" by it trivially.
2. The key component is not equality bound, but we're already ordered by it. That means that results for the plan is ordered in a way such that the plan and the candidate key are ordered in a way means that they are the same up to some prefix which must include this key component. So, we only need to validate that the _suffixes_ are compatibly ordered for a fixed value of the prefix (up to `nextNonPrefix`). For such a grouping, all results have the same value for this expression, so we are likewise correctly ordered.

I'm not sure on how to best explain this, though. Hopefully, that tracks, because it would be nice to validate that this fix is conceptually correct

MMcM (approved):

pengpeng-lu (approved):

PR #2470, opened 2024-01-30
Resolves #2491: Use multiple partitions to perform rebalancing

### LINKED ISSUE #2491: Use multiple transactions when rebalancing lucene partitions
We should use multiple partitions when rebalancing documents between lucene partitions.
This will allow us to repartition large backlogs and when there are many groups.

ScottDugas:
The main change here is that this uses multiple transactions to move documents between partitions. It also ensures that merge happens on all partitions.
This is a rather simplistic implementation, with a fixed property for how many to move per transaction.
In order to test all the permutations this adds some randomized tests that create a bunch of random data, in a single, most-recent partition and then runs the mergeIndex to move them between partitions.

It also does a few other things:
1. The search used in getOldestNDocuments was not right, and wasn't sorting properly
2. The JOINED index used in tests has the sort added
3. It changes the tuple at the beginnig by doing `groupKey.add()` rather than using a nested tuple.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
+++ b/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
@@ -1,0 +1,63 @@
+/*
+ * RandomizedTestUtils.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2024 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.test;
+
+import org.junit.jupiter.params.provider.Arguments;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/**
+ * A Utility class for adding randomness to tests.
+ */
+public final class RandomizedTestUtils {
+    private RandomizedTestUtils() {
+    }
+
+    /**
+     * Return a stream of randomly generated arguments based on the gradle properties.
+     * <p>
+     *     Tip: Use {@code Stream.concat(streamOfFixedArguments, RandomizedTestUtils.randomArguments(random -> { ... }} to have some fixed
+     *     arguments, and also add randomized arguments when including random tests.
+     * </p>
+     * @param randomArguments A mapper from a {@link Random} to the arguments to be provided to
+     * @return a stream of {@link Arguments} for a {@link org.junit.jupiter.params.ParameterizedTest}
+     */
+    public static Stream<Arguments> randomArguments(Function<Random, Arguments> randomArguments) {
+        if (includeRandomTests()) {
+            Random random = ThreadLocalRandom.current();
+            return IntStream.range(0, getIterations()).mapToObj(i -> randomArguments.apply(random));
+        } else {
+            return Stream.of();
+        }
+    }
+
+    private static int getIterations() {
+        return Integer.parseInt(System.getProperty("tests.iterations", "0"));
+    }
+
+    private static boolean includeRandomTests() {
+        return Boolean.parseBoolean(System.getProperty("tests.includeRandom", "false"));
+    }
+}
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -111,6 +111,7 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
     private final Executor executor;
     LuceneIndexKeySerializer keySerializer;
     private boolean serializerErrorLogged = false;
+    @Nonnull
     private final LucenePartitioner partitioner;
 
     public LuceneIndexMaintainer(@Nonnull final IndexMaintainerState state, @Nonnull Executor executor) {
@@ -321,6 +322,13 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
                 .thenCompose(ignored -> directoryManager.mergeIndex(partitioner, indexAnalyzerSelector.provideIndexAnalyzer("")));
     }
 
+    @Nonnull
+    @VisibleForTesting
+    public LucenePartitioner getPartitioner() {
+        return partitioner;
+    }
+
+    @SuppressWarnings("PMD.CloseResource") // the runner is closed in a whenComplete on the future returned
     public CompletableFuture<Void> rebalancePartitions() {
         if (!partitioner.isPartitioningEnabled()) {
             return CompletableFuture.completedFuture(null);
@@ -328,7 +336,10 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         final FDBRecordStore.Builder storeBuilder = state.store.asBuilder();
         FDBDatabaseRunner runner = state.context.newRunner();
         final Index index = state.index;
-        return rebalancePartitions(runner, storeBuilder, index);
+        return rebalancePartitions(runner, storeBuilder, index)
+                .whenComplete((result, error) -> {
+                    runner.close();
+                });
     }
 
     private static CompletableFuture<Void> rebalancePartitions(final FDBDatabaseRunner runner, final FDBRecordStore.Builder storeBuilder, final Index index) {
@@ -349,8 +360,6 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
                             });
                 });
             });
-        }).whenComplete((result, exception) -> {
-            runner.close();
         });
     }
 
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -34,6 +34,7 @@ import com.apple.foundationdb.record.RecordCursorContinuation;
 import com.apple.foundationdb.record.RecordCursorStartContinuation;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TupleRange;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectoryManager;
 import com.apple.foundationdb.record.lucene.idformat.LuceneIndexKeySerializer;
@@ -48,6 +49,7 @@ import com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexableRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
+import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainer;
 import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
 import com.apple.foundationdb.record.provider.foundationdb.IndexOperation;
 import com.apple.foundationdb.record.provider.foundationdb.IndexOperationResult;
@@ -328,6 +330,18 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         return partitioner;
     }
 
+    @Nonnull
+    private static LucenePartitioner getPartitioner(final Index index, final FDBRecordStore store) {
+        final IndexMaintainer indexMaintainer = store.getIndexMaintainer(index);
+        if (indexMaintainer instanceof LuceneIndexMaintainer) {
+            return ((LuceneIndexMaintainer)indexMaintainer).partitioner;
+        } else {
+            throw new RecordCoreException("Index being repartitioned is no longer a lucene index")
+                    .addLogInfo(LogMessageKeys.INDEX_NAME, index.getName(),
+                            LogMessageKeys.INDEX_TYPE, index.getType());
+        }
+    }
+
     @SuppressWarnings("PMD.CloseResource") // the runner is closed in a whenComplete on the future returned
     public CompletableFuture<Void> rebalancePartitions() {
         if (!partitioner.isPartitioningEnabled()) {
@@ -344,23 +358,20 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
 
     private static CompletableFuture<Void> rebalancePartitions(final FDBDatabaseRunner runner, final FDBRecordStore.Builder storeBuilder, final Index index) {
         AtomicReference<RecordCursorContinuation> continuation = new AtomicReference<>(RecordCursorStartContinuation.START);
-        return AsyncUtil.whileTrue(() -> {
-            return runner.runAsync(context -> {
-                return storeBuilder.setContext(context).openAsync().thenCompose(store -> {
-                    store.getIndexDeferredMaintenanceControl().setAutoMergeDuringCommit(false);
-                    return ((LuceneIndexMaintainer)store.getIndexMaintainer(index))
-                            .partitioner.rebalancePartitions(continuation.get())
-                            .thenApply(newContinuation -> {
-                                if (newContinuation.isEnd()) {
-                                    return false;
-                                } else {
-                                    continuation.set(newContinuation);
-                                    return true;
-                                }
-                            });
-                });
-            });
-        });
+        return AsyncUtil.whileTrue(
+                () -> runner.runAsync(
+                        context -> storeBuilder.setContext(context).openAsync().thenCompose(store -> {
+                            store.getIndexDeferredMaintenanceControl().setAutoMergeDuringCommit(false);
+                            return getPartitioner(index, store).rebalancePartitions(continuation.get())
+                                    .thenApply(newContinuation -> {
+                                        if (newContinuation.isEnd()) {
+                                            return false;
+                                        } else {
+                                            continuation.set(newContinuation);
+                                            return true;
+                                        }
+                                    });
+                        })));
     }
 
     @Nonnull
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -30,6 +30,8 @@ import com.apple.foundationdb.record.IsolationLevel;
 import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.RecordCursorContinuation;
+import com.apple.foundationdb.record.RecordCursorStartContinuation;
 import com.apple.foundationdb.record.ScanProperties;
 import com.apple.foundationdb.record.TupleRange;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
@@ -37,12 +39,15 @@ import com.apple.foundationdb.record.lucene.directory.FDBDirectoryManager;
 import com.apple.foundationdb.record.lucene.idformat.LuceneIndexKeySerializer;
 import com.apple.foundationdb.record.lucene.idformat.RecordCoreFormatException;
 import com.apple.foundationdb.record.lucene.search.BooleanPointsConfig;
+import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.IndexRecordFunction;
 import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexableRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecord;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
 import com.apple.foundationdb.record.provider.foundationdb.IndexMaintainerState;
 import com.apple.foundationdb.record.provider.foundationdb.IndexOperation;
 import com.apple.foundationdb.record.provider.foundationdb.IndexOperationResult;
@@ -85,6 +90,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
@@ -311,8 +317,36 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
 
     @Override
     public CompletableFuture<Void> mergeIndex() {
-        return partitioner.rebalancePartitions()
+        return rebalancePartitions()
                 .thenCompose(ignored -> directoryManager.mergeIndex(partitioner, indexAnalyzerSelector.provideIndexAnalyzer("")));
+    }
+
+    public CompletableFuture<Void> rebalancePartitions() {
+        if (!partitioner.isPartitioningEnabled()) {
+            return CompletableFuture.completedFuture(null);
+        }
+        final FDBRecordStore.Builder storeBuilder = state.store.asBuilder();
+        FDBDatabaseRunner runner = state.context.newRunner();
+        final Index index = state.index;
+        AtomicReference<RecordCursorContinuation> continuation = new AtomicReference<>(RecordCursorStartContinuation.START);
+        return AsyncUtil.whileTrue(() -> {
+            return runner.runAsync(context -> {
+                return storeBuilder.openAsync().thenCompose(store -> {
+                    return ((LuceneIndexMaintainer)store.getIndexMaintainer(index))
+                            .partitioner.rebalancePartitions(continuation.get())
+                            .thenApply(newContinuation -> {
+                                if (newContinuation.isEnd()) {
+                                    return false;
+                                } else {
+                                    continuation.set(newContinuation);
+                                    return true;
+                                }
+                            });
+                });
+            });
+        }).whenComplete((result, exception) -> {
+            runner.close();
+        });
     }
 
     @Nonnull
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
@@ -328,10 +328,15 @@ public class LuceneIndexMaintainer extends StandardIndexMaintainer {
         final FDBRecordStore.Builder storeBuilder = state.store.asBuilder();
         FDBDatabaseRunner runner = state.context.newRunner();
         final Index index = state.index;
+        return rebalancePartitions(runner, storeBuilder, index);
+    }
+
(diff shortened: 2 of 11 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -121,18 +129,58 @@ public CompletableFuture<Void> mergeIndex(LucenePartitioner partitioner, LuceneA
...
+                        2);
     }
 
-    private void mergeIndex(LuceneAnalyzerWrapper analyzerWrapper, Tuple groupingKey, @Nullable Integer partitionId) {
-        try {
-            getDirectoryWrapper(groupingKey, partitionId, true).mergeIndex(analyzerWrapper);
-        } catch (IOException e) {
-            throw new RecordCoreStorageException("Lucene mergeIndex failed", e);
+    private CompletableFuture<Void> mergeIndex(LuceneAnalyzerWrapper analyzerWrapper, Tuple groupingKey, @Nullable LucenePartitioner partitioner, final AgilityContext agileContext) {
+        if (!partitioner.isPartitioningEnabled()) {
```
wshaib-apple, line 137: `partitioner` can be `null`

ScottDugas: It actually can't, I'll update the annotations.

fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Return a stream of randomly generated arguments based on the gradle properties.
+     * <p>
+     *     Tip: Use {@code Stream.concat(streamOfFixedArguments, RandomizedTestUtils.randomArguments(random -> { ... }} to have some fixed
+     *     arguments, and also add randomized arguments when including random tests.
+     * </p>
+     * @param randomArguments A mapper from a {@link Random} to the arguments to be provided to
+     * @return a stream of {@link Arguments} for a {@link org.junit.jupiter.params.ParameterizedTest}
+     */
+    public static Stream<Arguments> randomArguments(Function<Random, Arguments> randomArguments) {
```
ScottDugas, line 47: I think this could eventually be pretty useful, but the only use case I have in this code is to have `random -> Arguments.of(random.nextLong())` where the long seeds another `Random`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -311,10 +318,51 @@ void deleteDocument(Tuple groupingKey, Integer partitionId, Tuple primaryKey) th
...
+                });
+    }
+
+    private static CompletableFuture<Void> rebalancePartitions(final FDBDatabaseRunner runner, final FDBRecordStore.Builder storeBuilder, final Index index) {
+        AtomicReference<RecordCursorContinuation> continuation = new AtomicReference<>(RecordCursorStartContinuation.START);
+        return AsyncUtil.whileTrue(() -> {
+            return runner.runAsync(context -> {
+                return storeBuilder.setContext(context).openAsync().thenCompose(store -> {
+                    store.getIndexDeferredMaintenanceControl().setAutoMergeDuringCommit(false);
+                    return ((LuceneIndexMaintainer)store.getIndexMaintainer(index))
```
ScottDugas, line 351: If the store changes between transactions, and the index no longer is a `LuceneIndexMantainer` this will fail with a `ClassCastException`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -295,14 +316,18 @@ byte[] partitionMetadataKeyFromTimestamp(@Nonnull Tuple groupKey, long timestamp
...
 
+    private static long getTimestampFromPartitionInfo(@Nonnull final LucenePartitionInfoProto.LucenePartitionInfoOrBuilder builder) {
+        return Tuple.fromBytes(builder.getFrom().toByteArray()).getLong(0);
+    }
+
     @Nonnull
     CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> findPartitionInfo(@Nonnull Tuple groupKey, long timestamp) {
-        Range range = new Range(state.indexSubspace.subspace(Tuple.from(groupKey, PARTITION_META_SUBSPACE)).pack(),
-                state.indexSubspace.subspace(Tuple.from(groupKey, PARTITION_META_SUBSPACE, timestamp)).pack());
+        Range range = new Range(state.indexSubspace.subspace(groupKey.add(PARTITION_META_SUBSPACE)).pack(),
```
ScottDugas, line 329: In testing I noticed that the keys were having nested tuples. This doesn't seem right to me, so I changed it to use `add` so that it is, e.g. `[g0, 1, 23498]` instead of `[[g0], 1, 23498]`. Since key expressions have a fixed number of columns, the nesting of the tuple was just using extra bytes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -414,11 +439,14 @@ private LucenePartitionInfoProto.LucenePartitionInfo newPartitionMetadata(long t
...
      *
      * @param groupKey group key
+     * @param context the context in which to execute; should generally be {@code state.context}
+     * @param indexSubspace the index subspace; should generally be {@code state.indexSubspace}
      * @return partition metadata future
      */
     @Nonnull
-    private CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> getNewestPartition(@Nonnull Tuple groupKey) {
-        return getEdgePartition(groupKey, true);
+    private static CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> getNewestPartition(
```
ScottDugas, line 447: Made this static because some code needs to access this from, potentially, other transactions.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -487,40 +519,62 @@ public static long getTo(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo p
...
         final int groupingCount = expression.getGroupingCount();
 
+        final ScanProperties scanProperties = ScanProperties.FORWARD_SCAN.with(
+                props -> props.clearState().setReturnedRowLimit(1));
+
+        final Range range = state.indexSubspace.range();
+        final KeyRange keyRange = new KeyRange(range.begin, range.end);
+        final Subspace subspace = state.indexSubspace;
         try (RecordCursor<Tuple> cursor = new ChainedCursor<>(
                 state.context,
```
ScottDugas, line 552: It only uses the context to get the limit manager. It would probably be good to have it take the limit manager, but I didn't want to put that in this PR.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -487,40 +519,62 @@ public static long getTo(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo p
...
                 lastKey -> FDBDirectoryManager.nextTuple(state.context, subspace, keyRange, lastKey, scanProperties, groupingCount),
                 Tuple::pack,
                 Tuple::fromBytes,
-                null,
+                start.toBytes(),
```
ScottDugas, line 556: It would probably be good if this could pass in `start` on it's own, but I didn't want to pull that into this PR.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -532,31 +586,42 @@ public CompletableFuture<Void> rebalancePartitions() {
...
 
             for (LucenePartitionInfoProto.LucenePartitionInfo partitionInfo : partitionInfos) {
                 if (partitionInfo.getCount() > indexPartitionHighWatermark) {
                     // process one partition
 
                     // get the N oldest documents in the partition (note N = (count of docs to move) + 1, since we need
                     // the (N+1)th doc's timestamp to update the partition's "from" field.
+                    final Integer repartitionDocumentCount = Math.min(
+                            Objects.requireNonNull(state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_REPARTITION_DOCUMENT_COUNT)),
+                            indexPartitionHighWatermark);
```
ScottDugas, line 611: This _shouldn't_ be relevant in production environments, where `repartitionDocumentCount` should be much less than the high watermark, but since one is from the context, and the other from the index, I figured it would be good to protect against it.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -579,7 +644,7 @@ public LuceneRecordCursor getOldestNDocuments(@Nonnull final LucenePartitionInfo
         LuceneScanParameters scan = new LuceneScanQueryParameters(
                 comparisons,
                 new LuceneQuerySearchClause(LuceneQueryType.QUERY, "*:*", false),
-                new Sort(new SortField(partitionTimestampFieldName, SortField.Type.LONG, false)),
+                new Sort(new SortField(partitionTimestampFieldName.replace('.', '_'), SortField.Type.LONG, false)),
```
ScottDugas, line 647: Without this, every document was missing the `partitionTimestampFieldName`, and thus the order they come back was seemingly random (maybe insertion order, or dependent on some merging).
For example, the tests use `complex.timestamp` which as a lucene field is `complex_timestamp`.
We may want to reconsider how we declare the partitioning timestamp.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -612,10 +678,10 @@ public LuceneRecordCursor getOldestNDocuments(@Nonnull final LucenePartitionInfo
      * @return void future
```
ScottDugas, line 678: TODO: I need to update the javadoc here. It's a future containing the number of documents that were moved.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -687,9 +760,28 @@ private CompletableFuture<Void> moveDocsFromPartition(@Nonnull final LuceneParti
...
-    private CompletableFuture<List<LucenePartitionInfoProto.LucenePartitionInfo>> getAllPartitionMetaInfo(@Nonnull final Tuple groupingKey) {
-        Range range = state.indexSubspace.subspace(Tuple.from(groupingKey, PARTITION_META_SUBSPACE)).range();
+    @VisibleForTesting
+    public CompletableFuture<List<LucenePartitionInfoProto.LucenePartitionInfo>> getAllPartitionMetaInfo(@Nonnull final Tuple groupingKey) {
+        Range range = state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE)).range();
         final AsyncIterable<KeyValue> rangeIterable = state.context.ensureActive().getRange(range, Integer.MAX_VALUE, true, StreamingMode.WANT_ALL);
         return AsyncUtil.collect(rangeIterable).thenApply(all -> all.stream().map(LucenePartitioner::partitionInfoFromKV).collect(Collectors.toList()));
     }
+
+    public static CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> getNextPartitionInfo(
```
ScottDugas, line 770: This is needed for `FDBDirectoryManager` to find the partitions to do merging.
I think this adds to the argument that the relationship between the layers here is messy and needs to be rethought.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -101,17 +109,17 @@ public CompletableFuture<Void> mergeIndex(LucenePartitioner partitioner, LuceneA
...
-            return AsyncUtil.DONE;
+            return mergeIndex(analyzerWrapper, TupleHelpers.EMPTY, partitioner, agilityContext);
         }
         GroupingKeyExpression expression = (GroupingKeyExpression) rootExpression;
         final int groupingCount = expression.getGroupingCount();
 
         final RecordCursor<Tuple> cursor = new ChainedCursor<>(
                 state.context,
-                lastKey -> nextTuple(state.context, subspace, keyRange, lastKey, scanProperties, groupingCount),
+                lastKey -> agilityContext.apply(context -> nextTuple(context, subspace, keyRange, lastKey, scanProperties, groupingCount)),
```
ScottDugas, line 122: If the merge of an individual group takes more than 5 seconds the chained cursor won't be able to get the next group, and will fail. This has it use the AgilityContext to get the next group.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -121,18 +129,59 @@ public CompletableFuture<Void> mergeIndex(LucenePartitioner partitioner, LuceneA
...
+            return AsyncUtil.whileTrue(() -> getNextPartitionInfo(groupingKey, agileContext, lastPartitionInfo)
+                    .thenApply(partitionId -> mergePartition(analyzerWrapper, groupingKey, agileContext, partitionId, lastPartitionInfo)));
+        }
+    }
+
+    @Nonnull
+    private Boolean mergePartition(final LuceneAnalyzerWrapper analyzerWrapper, final Tuple groupingKey,
+                                   final AgilityContext agileContext, final Integer partitionId,
+                                   final AtomicReference<LucenePartitionInfoProto.LucenePartitionInfo> lastPartitionInfo) {
+        if (lastPartitionInfo.get() == null) {
```
ScottDugas, line 157: I'm not sure why there is an `AtomicReference` used here; I'll see if I can clean that up.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -188,15 +237,38 @@ public void invalidatePrefix(@Nonnull Tuple prefix) {
...
+    private FDBDirectoryWrapper getDirectoryWrapper(@Nullable Tuple groupingKey, @Nullable Integer partitionId, final AgilityContext agilityContext) {
         Tuple mapKey = groupingKey == null ? TupleHelpers.EMPTY : groupingKey;
         if (partitionId != null) {
             mapKey = mapKey.add(LucenePartitioner.PARTITION_DATA_SUBSPACE).add(partitionId);
         }
-        return createdDirectories.computeIfAbsent(mapKey, key -> new FDBDirectoryWrapper(state, key, mergeDirectoryCount, useAgilityContext));
+        return createdDirectories.computeIfAbsent(mapKey, key -> new FDBDirectoryWrapper(state, key, mergeDirectoryCount, agilityContext));
+    }
+
+
```
ScottDugas, line 251: TODO: excessive blank line

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -188,15 +237,38 @@ public void invalidatePrefix(@Nonnull Tuple prefix) {
...
+    private AgilityContext getAgilityContext(boolean useAgilityContext) {
```
ScottDugas, line 252: This was moved from `FDBDirectoryWrapper`, I think unchanged.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -364,19 +377,33 @@ public class LuceneIndexTest extends FDBRecordStoreTestBase {
...
-            )), LuceneIndexTypes.LUCENE, ImmutableMap.of(
-                    INDEX_PARTITION_BY_TIMESTAMP, "complex.timestamp",
-                    INDEX_PARTITION_HIGH_WATERMARK, "10"));
+    @Nonnull
+    private static Index getJoinedIndex(final String partitionHighWatermark) {
+        return new Index("joinNestedConcat",
+                concat(
+                        field("complex").nest(function(LuceneFunctionNames.LUCENE_STORED, field("is_seen"))),
+                        field("simple").nest(function(LuceneFunctionNames.LUCENE_TEXT, field("text"))),
+                        field("complex").nest(function(LuceneFunctionNames.LUCENE_SORTED, field("timestamp")))
```
ScottDugas, line 388: The `timestamp` field was left out in both of these. I think this is necessary for partitioning to actually work, although I added the field before fixing the `getOldestNDocument` search.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -424,6 +451,7 @@ private static Index getMapOnValueIndexWithOption(@Nonnull String name, @Nonnull
             ImmutableMap.of(IndexOptions.TEXT_TOKENIZER_NAME_OPTION, AllSuffixesTextTokenizer.NAME,
                     LuceneIndexOptions.PRIMARY_KEY_SEGMENT_INDEX_ENABLED, "true"));
 
+
```
ScottDugas, line 454: TODO remove unnecessary blank line

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -684,26 +712,22 @@ void repartitionGroupedTest(Pair<Index, Tuple> indexAndGroupingKey) throws IOExc
         explicitMergeIndex(index, contextProps, schemaSetup);
         partitionInfos = getPartitionMeta(index,
                 groupingKey, contextProps, schemaSetup);
-        // now there should be 2 partitions: partition 0 with (totalDocCount - 6) docs, and partition 1 with 6 docs
```
ScottDugas, line 687: We now do all the repartitioning, not just one in a call.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -748,7 +771,223 @@ private void validateDocsInPartition(Index index, int partitionId, Tuple groupin
...
+                Stream.of(
+                        Arguments.of(13L),
+                        Arguments.of(-644766138635622644L),
+                        Arguments.of(-1089113174774589435L)),
+                RandomizedTestUtils.randomArguments(random -> Arguments.of(random.nextLong())));
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void randomizedRepartitionTest(long seed) throws IOException {
```
ScottDugas, line 902: I've run this a few times with 100 iterations with randomness, all successful.
I only added a couple here to be run regularly.

It might make sense to have this have more arguments (e.g. `isGrouped`/`isSynthetic`/`groupCount`) and then have a seed for all the other stuff so that we can more easily guarantee that the non-random test has various combinations. I may tackle that if I feel I have the cyles, or defer for later.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -748,7 +771,223 @@ private void validateDocsInPartition(Index index, int partitionId, Tuple groupin
...
+                schemaSetup.accept(context);
+                for (int j = 0; j < docCount; j++) {
+                    final int group = isGrouped ? random.nextInt(random.nextInt(10) + 1) : 0; // irrelevant if !isGrouped
+                    final int countInGroup = ids.computeIfAbsent(group, key -> new HashMap<>()).size();
+                    // we currently don't support multiple records with the same timestamp, specifically at the boundaries
+                    long timestamp = start + countInGroup + random.nextInt(20) - 5;
+                    final Set<Long> existingTimestamps = allExistingTimestamps.computeIfAbsent(group, key -> new HashSet<>());
+                    while (!existingTimestamps.add(timestamp)) {
+                        timestamp++;
+                    }
```
ScottDugas, line 954: Since we don't currently support having two records at the boundary with the same timestamp this makes that easier.
If we improve that functionality, we'll probably want to change this.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -824,7 +1076,6 @@ private void explicitMergeIndex(Index index, RecordLayerPropertyStorage contextP
                     .build()) {
                 indexBuilder.mergeIndex();
             }
-            commit(context);
```
ScottDugas, line 827: The commit is unnecessary, because the indexer doesn't actually use the index.
In theory this could create a store builder that doesn't have the context at all.
I wanted to remove it to ensure that nothing was using this context.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -857,8 +1103,7 @@ private static JoinedRecordTypeBuilder joinedMetadataHook(@Nonnull final RecordM
         final JoinedRecordTypeBuilder joined = metaDataBuilder.addJoinedRecordType("luceneJoinedPartitionedIdx");
         joined.addConstituent("complex", "ComplexDocument");
         joined.addConstituent("simple", "SimpleDocument");
-        joined.addJoin("simple", field("group"), "complex", field("group"));
-        metaDataBuilder.addIndex("SimpleDocument", "simple$group", field("group"));
```
ScottDugas, line 861: I think this was just unnecessary.
I commented it out to see why it was required, and nothing broke, so I just left it off.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -857,8 +1103,7 @@ private static JoinedRecordTypeBuilder joinedMetadataHook(@Nonnull final RecordM
...
+        joined.addJoin("simple", field("doc_id"), "complex", field("header").nest("header_id"));
```
ScottDugas, line 1106: Having the grouping field also be the join field made it nearly impossible to do multi-group tests that made sense, and I don't think there's actually a use case for that.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
```diff
@@ -0,0 +1,241 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * A utility for validating the consistency and contents of a lucene index.
```
ScottDugas, line 67: I would love to add more validations to this, but this seemed most of what was necessary right now.
It would probably be good to also validate that merges happened, but I don't know of a good way to do that at the moment, other than try to do a merge, and making sure the segments didn't change.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
```diff
@@ -0,0 +1,241 @@
...
+                                                                                Tuple groupingKey) {
+        try (FDBRecordContext context = contextProvider.get()) {
+            final FDBRecordStore recordStore = schemaSetup.apply(context);
+            LuceneIndexMaintainer indexMaintainer = (LuceneIndexMaintainer) recordStore.getIndexMaintainer(index);
+            return indexMaintainer.getPartitioner().getAllPartitionMetaInfo(groupingKey).join();
+        }
+    }
+
+
+    private void validateDocsInPartition(final FDBRecordStore recordStore, Index index, int partitionId, Tuple groupingKey,
```
ScottDugas, line 190: There is still `LuceneIndexTest.validateDocsInPartition` which I think is very similar; I should see if that can call this.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -295,14 +316,18 @@ byte[] partitionMetadataKeyFromTimestamp(@Nonnull Tuple groupKey, long timestamp
     void savePartitionMetadata(@Nonnull Tuple groupKey, @Nonnull final LucenePartitionInfoProto.LucenePartitionInfo.Builder builder) {
         LucenePartitionInfoProto.LucenePartitionInfo updatedPartition = builder.build();
         state.context.ensureActive().set(
-                partitionMetadataKeyFromTimestamp(groupKey, Tuple.fromBytes(builder.getFrom().toByteArray()).getLong(0)),
+                partitionMetadataKeyFromTimestamp(groupKey, getTimestampFromPartitionInfo(builder)),
                 updatedPartition.toByteArray());
     }
 
+    private static long getTimestampFromPartitionInfo(@Nonnull final LucenePartitionInfoProto.LucenePartitionInfoOrBuilder builder) {
```
wshaib-apple, line 323: there is already a static method in this class that does this (`getFrom`, taking a `LucenePartitionInfo` parameter). Maybe these can be combined

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -76,6 +89,8 @@
  */
 @API(API.Status.EXPERIMENTAL)
 public class LucenePartitioner {
+
```
wshaib-apple, line 92: Should we confirm that the provided partitioning field is also defined as "sortable" in the index?

ScottDugas: I think that's a good idea, but I think it would be better to do that as a followup because it's a bit tricky to confirm that.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ScottDugas: The changes in #2495 caused this test to break sometimes because there are more arrangements that can occur. I pushed a new change that is more precise in the assertions, and have run it 1000 times successfully.

wshaib-apple (dismissed):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * A broad validation of the lucene index, asserting consistency, and that various operations did what they were
+     * supposed to do.
+     * <p>
+     *     This has a lot of validation that could be added, and it would be good to be able to control whether it's
+     *     expected that `mergeIndex` had been run or not; right now it assumes it has been run.
+     * </p>
+     * @param index the index to validate
+     * @param expectedIds a map from group to primaryKey to timestamp
```
MMcM, line 88: Is there any reason for this (and the subroutines it calls) to use `id` instead of `primaryKey`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -501,40 +529,62 @@ public static long getTo(@Nonnull LucenePartitionInfoProto.LucenePartitionInfo p
     /**
      * Re-balance full partitions, if applicable.
      *
-     * @return void future
+     * @param start The continuation at which to resume rebalancing, as returned from a previous call to
+     * {@code rebalancePartitions}.
+     * @return a tuple of the next group that needs repartitioning
```
MMcM, line 534: I think this could be worded more like `start` is. That the continuation state is a tuple advancing by groups isn't, that I can see, part of the contract of this function. Which is roughly, keep calling this until you get an ended continuation. It could do part of a group or multiple groups or whatever it wants inside without changing that.

MMcM (changes requested):

MMcM (approved):

wshaib-apple (approved):

Issue #2360, opened 2023-11-14
Update to JDK 17


alecgrieser:
Given that JDK 11 is reaching the end of its (LTS) support, we should update the project to JDK 17, which is the next LTS version of Java. At the very least, that means updating our build chain so that it uses the newer JDK version. For compatibility reasons, we may need to preserve JDK 11 byte code compatibility.

alecgrieser:
Sonarcloud has since stopped supporting JDK 11, so we will not be able to get successful PRBs with sonar until we update our JDK version

PR #2471, opened 2024-01-31
Set release option during Java builds

alecgrieser:
This configures the Java builds to set the release option to fix the byte code to JDK 11. This will allow us to move our build instance forward to JDK 17 without needing to worry about breaking compatibility with any JDK 11 consumers of the project. This gets us ready for #2360

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -236,6 +236,9 @@ subprojects {
 
                 //enable incremental compilation
                 options.incremental = true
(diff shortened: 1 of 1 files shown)


ScottDugas (approved):

ScottDugas:
Also, I think you meant to say this gets us ready for #2360

Issue #2475, opened 2024-02-01
Kahn iterator is unstable, causing `Value` simplification engine to return wrong results sometimes


hatyo:
I came across this bug while implementing some tests for a new algorithm that maps parts of a `Value` to another `Value`. Part of that algorithm relies heavily on `Value.pullUp` logic which describes how a (sub)`Value` is referenced with a quantifier that sits on top. Some of the tests were failing sporadically, this is related to `Value.pullUp` returning the wrong results sometimes.

After some digging, it turned out that the root cause of the problem it related to a topological sort implementation that uses Kahn iterator, the iterator is actually unstable, and it could return the nodes with 0-indegree in a different order because it is putting these nodes in a `HashMap`.

When the nodes' class is using the default hash code implementation, which is the case for `AbstractRuleSet` class hierarchy, then the algorithm will return the 0-indegree nodes in different order each time, because the objects' hash code is different.

PR #2476, opened 2024-02-01
Fix topological sort instability

hatyo:
This fixes instability in Kahn iterator that causes it to return a different topological order when used to sort objects of a class with a default hash code implementation.

This addresses #2475


hatyo:
@normen662 build failed because of Sonarqube:

>   The version of Java (11.0.22) used to run this analysis is deprecated, and SonarCloud no longer supports it. Please upgrade to Java 17 or later.
>   You can find more information here: https://docs.sonarsource.com/sonarcloud/appendices/scanner-environment/

normen662 (approved):

PR #2486, opened 2024-02-05
Fixes #2485: FDBExceptions::wrapException incorrectly wraps InterruptedExceptions

### LINKED ISSUE #2485: FDBExceptions::wrapException incorrectly wraps InterruptedExceptions
There's a bug on this line: https://github.com/FoundationDB/fdb-record-layer/blob/77efbbb8f3f65e7113fb31af1ae41e1da60c01b7/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBExceptions.java#L198

There are two constructors of the `RecordCoreInterruptedException` class. One of them takes a message and an `InterruptedException` (as the cause). The other takes a message and a variadic number of `Object`s (that must have even length). Because `ex` has declared type `Throwable`, it chooses the second one, but it’s pretty clear that the intention was to chose the former. This results in the following error being thrown:

```
java.lang.IllegalArgumentException: Unbalanced key/value logging info
		at com.apple.foundationdb.util.LoggableKeysAndValuesImpl.addLogInfo(LoggableKeysAndValuesImpl.java:103)
		at com.apple.foundationdb.util.LoggableException.<init>(LoggableException.java:49)
		at com.apple.foundationdb.record.RecordCoreException.<init>(RecordCoreException.java:36)
		at com.apple.foundationdb.record.RecordCoreInterruptedException.<init>(RecordCoreInterruptedException.java:38)
		at com.apple.foundationdb.record.provider.foundationdb.FDBExceptions.wrapException(FDBExceptions.java:198)
```

alecgrieser:
This updates the logic in `FDBExceptions::wrapException` to avoid an error with the message "Unbalanced key/value logging info". This was ultimately caused by a problem where the wrong constructor was being selected when creating the `RecordCoreInterruptedException`, so this adds a cast so that the signature now aligns with the correct constructor.

This fixes #2485.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBExceptionsTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBExceptionsTest.java
@@ -21,6 +21,7 @@
 package com.apple.foundationdb.record.provider.foundationdb;
 
 import com.apple.foundationdb.record.RecordCoreArgumentException;
+import com.apple.foundationdb.record.RecordCoreInterruptedException;
 import com.apple.foundationdb.record.logging.CompletionExceptionLogHelper;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.util.LoggableKeysAndValues;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBExceptionsTest.java
```diff
@@ -86,12 +90,32 @@ public void loggableTimeoutException() {
...
+    void wrapCheckedException() {
         Exception cause = createCheckedException();
         final String methodName = "createCheckedException";
         testWrappedStackTrace(cause, methodName);
     }
 
+    @Test
+    void wrapInterruptedException() {
+        final InterruptedException err = createInterruptedException();
+        RuntimeException wrappedErr = FDBExceptions.wrapException(err);
```
alecgrieser, line 102: This line used to fail with an `IllegalArgumentException` before this change was made (the same `IllegalArgumentException` that is referenced in the issue)

hatyo (approved):

PR #2492, opened 2024-02-07
make simplifaction rules agnostic to order of execution

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/CompensateRecordConstructorRule.java
```diff
@@ -59,28 +59,33 @@ public CompensateRecordConstructorRule() {
...
-            // i.e. the columns flowing results of values we care about.
-            //
-            for (final var childValueEntry : childResultPair.getRight().entrySet()) {
-                final var argumentValue = childValueEntry.getKey();
-                final var argumentValueCompensation = childValueEntry.getValue();
-                final var field = column.getField();
-                mergedMatchedValuesMap.putIfAbsent(argumentValue,
-                        new FieldValueCompensation(FieldValue.FieldPath.ofSingle(field.getFieldNameOptional().orElse(null), field.getFieldType(), i), argumentValueCompensation));
+                //
+                // No we have a column and the result we computed for all columns that do have results associated with them,
```
hatyo, line 76:
```suggestion
                // Now we have a column and the result we computed for all columns that do have results associated with them,
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/CompensateRecordConstructorRule.java
```diff
@@ -59,28 +59,33 @@ public CompensateRecordConstructorRule() {
...
 
-        for (int i = 0; i < recordConstructorValue.getColumns().size(); ++i) {
-            final var column = recordConstructorValue.getColumns().get(i);
-            final var childResultPair = call.getResult(column.getValue());
-            if (childResultPair == null) {
-                continue;
-            }
+        final var recordConstructorValueResult = call.getResult(recordConstructorValue);
+        if (recordConstructorValueResult != null) {
+            resultingMatchedValuesMap.put(recordConstructorValue, recordConstructorValueResult.getValue().get(recordConstructorValue));
```
hatyo, line 66: Can we avoid the extra `Map` by doing something like:

```java
call.yield(recordConstructorValue, recordConstructorValueResult.getRight());
return;
```

normen662: There may be other values in that map that I don't want to return.

hatyo (approved):
LGTM.

PR #2495, opened 2024-02-07
RESOLVES #2494: Optimize insertion of older Lucene documents to avoid…

### LINKED ISSUE #2494: Optimize insertion of older Lucene documents to avoid unnecessary partition re-balancing

wshaib-apple:
… partition re-balancing

- when a document being inserted is assigned to a partition that's at or higher than its high watermark, and the document's timestamp is older than all documents in the partition, create a new partition for the inserted document. This avoids re-balancing of the full partition later.

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
@@ -843,12 +843,14 @@ public class LuceneIndexTest extends FDBRecordStoreTestBase {
             validateDocsInPartition(index, 0, groupingKey, totalDocCount, makeKeyTuples(docGroupFieldValue, 1000, 1009), "text:propose");
         }
 
-        // now add a document older than the oldest document in partition 0
-        // it should go into partition 1
+        // now add 5 documents older than the oldest document in partition 0
+        // they should go into partition 1
         try (FDBRecordContext context = openContext(contextProps)) {
             schemaSetup.accept(context);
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -814,6 +814,44 @@ void repartitionSyntheticGroupedTest() throws IOException {
...
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            validateDocsInPartition(index, 0, groupingKey, totalDocCount, makeKeyTuples(docGroupFieldValue, 1000, 1009), "text:propose");
+        }
+
+        // now add a document older than the oldest document in partition 0
+        // it should go into partition 1
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            recordStore.saveRecord(createComplexDocument(1000L + totalDocCount, ENGINEER_JOKE, docGroupFieldValue, start - 1));
```
ScottDugas, line 850: We should test adding more than one record to make sure that if you insert 20 records in reverse order you end up with 2 10-document partitions, and not 10 1-document partitions, and then a 10-document partition.

ScottDugas (changes requested):

MMcM (approved):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Done

ScottDugas (approved):

Issue #2408, opened 2023-12-13
Add planner configuration property to control union ordering key behavior


alecgrieser:
With #2336 (fixed by #2350), the comparison keys on ordered union plans can change. This can result in plans changing, potentially in ways that can invalidate the continuation. We should have a planner configuration property that allows users to opt in to the old behavior, even though (I think) the old plans should always be worse. This property can allow the user to control when they decide to roll this feature out.

PR #2496, opened 2024-02-07
Resolves #2493: PlanOrderingKey too strict when computing order for In union

### LINKED ISSUE #2493: PlanOrderingKey too strict when computing order for In union
With #2336, an issue was fixed where the planner would avoid planning an OR query as a union when there were additional keys in a plan's `PlanOrderingKey`. There's an analogous problem with `IN` queries. For example:

```java
RecordQuery.newBuilder()
     .setRecordType(type)
     .setFilter(Query.field("field_1").in(listParameter))
     .setSort(Key.Expressions.field("field_2"))
     .build()
```

This can be satisfied by any value index prefixed by `(field_1, field_2)`. However, because the `PlanOrderingKey` constraint is too strict, this will only be planned if the index is _exactly_ on `(field_1, field_2)`, for more-or-less the same reasons as #2336.

alecgrieser:
This takes basically the same approach that was taken for #2350, but this time applying it to `InUnion` plans rather than just regular union plans.

This change is guarded behind a planner configuration plan. This is necessary because the new algorithm can choose new ordering keys for the in-union, and so the user needs to be able to control its roll out. This is to mirror what we did with #2408.

This resolves #2493.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
@@ -20,6 +20,7 @@
 
 package com.apple.foundationdb.record.provider.foundationdb.query;
 
+import com.apple.foundationdb.record.Bindings;
 import com.apple.foundationdb.record.EvaluationContext;
 import com.apple.foundationdb.record.IndexScanType;
 import com.apple.foundationdb.record.PlanHashable;
@@ -37,6 +38,7 @@ import com.apple.foundationdb.record.metadata.expressions.KeyExpression.FanType;
 import com.apple.foundationdb.record.provider.foundationdb.FDBQueriedRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
 import com.apple.foundationdb.record.query.RecordQuery;
+import com.apple.foundationdb.record.query.expressions.AndComponent;
 import com.apple.foundationdb.record.query.expressions.Comparisons;
 import com.apple.foundationdb.record.query.expressions.Query;
 import com.apple.foundationdb.record.query.expressions.QueryComponent;
@@ -45,18 +47,22 @@ import com.apple.foundationdb.record.query.plan.RecordQueryPlannerConfiguration;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.BindingMatcher;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.PrimitiveMatchers;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.RecordQueryPlanMatchers;
-import com.apple.foundationdb.record.query.plan.cascades.matching.structure.ValueMatchers;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryIndexPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan;
+import com.apple.foundationdb.tuple.Tuple;
 import com.apple.test.BooleanSource;
 import com.apple.test.Tags;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.protobuf.Message;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
 
+import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -152,20 +168,20 @@ void testInQueryNoIndex() throws Exception {
...
-                record -> assertThat(record.getNumValue2(), anyOf(is(0), is(2))),
+                rec -> assertThat(rec.getNumValue2(), anyOf(is(0), is(2))),
                 context -> assertDiscardedAtMost(33, context)));
     }
 
     /**
      * Verify that an IN (with parameter) without an index is implemented as a filter on a scan.
      */
-    @Test
+    @DualPlannerTest
```
alecgrieser, line 184: This is the test class with the highest number of changes, so it's worth specifying what changed here:

1. I updated all of the tests to be `DualPlanner` tests so that we'd get coverage of both planners. For many tests, the plans produced are similar, though they may use different implementation classes (e.g., `InValues` joins instead of `InParameter` joins). All of the existing queries already worked with Cascades, so this just served to double check this
2. For queries that specified a sort, I parameterized the query by `reverse` and now assert that the plan's reverse-ness matches the parameter's. The reverse version of the plans have different hashes; for all of the existing plans, the hashes didn't change. I also modified the runtime checks to validate that these results are correctly sorted
3. The most important additional test is `testParameterizedInQuerySortedWithExtraUnorderedColumns` which checks that when there's an in-parameter predicate extra columns in the index, that we can still plan that as an in-union (when the config is set).
4. The new test `testListInQuerySortedWithExtraUnorderedColumns` is similar, but it uses a list value instead of a list parameter, and it can check both translating the IN into an in-union and translating the IN into an OR (and thus planning it as a regular union). As it happens, when the IN is translated into an OR, this already did the right thing as long as you set the config parameter added in #2408. That mode is more-or-less deprecated, though, so this test is intended a nice validation that things work together.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -113,25 +116,44 @@ protected void setupSimpleRecordStore(RecordMetaDataHook recordMetaDataHook,
...
-                                         TestHelpers.DangerousConsumer<TestRecords1Proto.MySimpleRecord.Builder> checkRecord)
+                                         TestHelpers.DangerousConsumer<TestRecords1Proto.MySimpleRecord> checkRecord)
             throws Exception {
         return querySimpleRecordStore(recordMetaDataHook, plan, contextSupplier, checkRecord, context -> { });
     }
 
     protected int querySimpleRecordStore(RecordMetaDataHook recordMetaDataHook, RecordQueryPlan plan,
                                          Supplier<EvaluationContext> contextSupplier,
-                                         TestHelpers.DangerousConsumer<TestRecords1Proto.MySimpleRecord.Builder> checkRecord,
+                                         TestHelpers.DangerousConsumer<TestRecords1Proto.MySimpleRecord> checkRecord,
```
alecgrieser, line 126: This change isn't strictly necessary for this PR. The main change is to use the immutable `MySimpleRecord` class for assertions instead of the mutable `MySimpleRecord.Builder`s. None of the assertions were mutating the records, but it felt strange that the API could support it, so I modified it to remove it. It does mean that we need to call `.build()` before calling the assertions, but that should be a relatively minor thing (to be clear: in test code)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -152,20 +168,20 @@ void testInQueryNoIndex() throws Exception {
...
                     predicatesFilterPlan(descendantPlans(scanPlan().where(scanComparisons(unbounded()))))
-                            .where(predicates(valuePredicate(ValueMatchers.fieldValueWithFieldNames("num_value_2"), new Comparisons.ListComparison(Comparisons.Type.IN, ImmutableList.of(0, 2))))));
+                            .where(predicates(valuePredicate(fieldValueWithFieldNames("num_value_2"), new Comparisons.ListComparison(Comparisons.Type.IN, ImmutableList.of(0, 2))))));
 
             assertEquals(738091077, plan.planHash(PlanHashable.CURRENT_LEGACY));
             assertEquals(-2062922437, plan.planHash(PlanHashable.CURRENT_FOR_CONTINUATION));
         }
         assertEquals(67, querySimpleRecordStore(NO_HOOK, plan, EvaluationContext::empty,
-                record -> assertThat(record.getNumValue2(), anyOf(is(0), is(2))),
+                rec -> assertThat(rec.getNumValue2(), anyOf(is(0), is(2))),
```
alecgrieser, line 177: While I was here, I've renamed `record` variables to `rec`. In JDK 17, record will be a keyword, though not strictly speaking a reserve word, so it will still be legal to have variables named `record`, though it is now frowned upon. Since I was already modifying a lot of this test class, I went ahead and cleaned this up, too

MMcM (approved):

MMcM (approved):

pengpeng-lu (approved):

PR #2497, opened 2024-02-08
union type fails if two types share fields with the same name but different types

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1262,87 @@ static Any any() {
         return Any.INSTANCE;
     }
 
+    class AnyRecord extends Record {
```
pengpeng-lu, line 1263: I'm hesitated about which extends which here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PrimaryKeyProperty.java
```diff
@@ -312,6 +313,9 @@ public Optional<List<Value>> visitUnorderedUnionPlan(@Nonnull final RecordQueryU
         @Nonnull
         @Override
         public Optional<List<Value>> visitScanPlan(@Nonnull final RecordQueryScanPlan scanPlan) {
+            if (Type.anyRecord().equals(scanPlan.getResultType().getInnerType())) {
+                return Optional.empty();
+            }
```
pengpeng-lu, line 318: without this "if" statement, we could try to translate primary key expression on AnyRecord and fail.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
```diff
@@ -1970,13 +1972,13 @@ void testOrQuerySplitContinuations(OrQueryParams orQueryParams) throws Exception
                 .build();
         orQueryParams.setPlannerConfiguration(this);
 
-        // Index(MySimpleRecord$num_value_3_indexed [[1],[1]]) ∪ Index(MySimpleRecord$num_value_3_indexed [[3],[3]]) ∪ Index(MySimpleRecord$num_value_3_indexed [[5],[5]])
+        // Fetch(Covering(Index(MySimpleRecord$num_value_3_indexed [[1],[1]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]) ∪[_.num_value_3_indexed, _.rec_no] Covering(Index(MySimpleRecord$num_value_3_indexed [[3],[3]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]) ∪[_.num_value_3_indexed, _.rec_no] Covering(Index(MySimpleRecord$num_value_3_indexed [[5],[5]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]))
         RecordQueryPlan plan = planQuery(query);
```
pengpeng-lu, line 1976: this is the result of `System.out.println(plan)`, and it looks like this before my change. The real difference introduced by my change is that before the change, it is `concat(primaryKey, field("num_value_3_indexed")` and now the order is changed to `concat(field("num_value_3_indexed", primaryKey)`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
```diff
@@ -412,8 +414,8 @@ void testOrQuery1(OrQueryParams orQueryParams) throws Exception {
         assertMatchesExactly(plan, planMatcher);
 
         if (planner instanceof CascadesPlanner) {
-            assertEquals(-1974122514, plan.planHash(PlanHashable.CURRENT_LEGACY));
-            assertEquals(1183017387, plan.planHash(PlanHashable.CURRENT_FOR_CONTINUATION));
+            assertEquals(-1974121674, plan.planHash(PlanHashable.CURRENT_LEGACY));
+            assertEquals(1183017507, plan.planHash(PlanHashable.CURRENT_FOR_CONTINUATION));
```
pengpeng-lu, line 418: planhash changes because the order of the concat index changes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PrimaryKeyProperty.java
hatyo: Could this cause the common primary key shared of two types to be incorrectly not returned?

hatyo (commented):
Left a few comments, and started an async discussion thread to get more information.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1259,99 @@ static Any any() {
...
+
+        @Nonnull
+        @Override
+        public PRecordType toProto(@Nonnull final PlanSerializationContext serializationContext) {
+            return PRecordType.newBuilder().build();
+        }
+
+        @Nonnull
+        @Override
+        public RecordQueryPlanProto.PType toTypeProto(@Nonnull final PlanSerializationContext serializationContext) {
```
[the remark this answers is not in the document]
normen662: Specifically,
- add `PAnyRecord` into `record_query_plan.proto`
- add a `Deserializer` as a nested class using the `@AutoService` (see other implementors in `Type`)
- change `toProto()` above to create a `PAnyRecord` instead of an empty `PRecordType`
- add a test

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1259,99 @@ static Any any() {
...
+
+        @Override
+        public String toString() {
+            return getTypeCode().toString();
+        }
+
+        @Nonnull
+        @Override
+        public PRecordType toProto(@Nonnull final PlanSerializationContext serializationContext) {
+            return PRecordType.newBuilder().build();
```
normen662, line 1340: This is not correct -- `PAnyRecord` should not be serialized into an empty regular `PRecordType`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
normen662: Why is there still a difference? I think this needs to be resolved before this can go in.

normen662 (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
normen662: Nevermind, this is just a comment. Please undo it as it does not reflect the current plan.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/Values.java
```diff
@@ -46,6 +46,9 @@ public class Values {
     @Nonnull
     public static List<Value> deconstructRecord(@Nonnull Value recordValue) {
         Verify.verify(recordValue.getResultType().isRecord());
+        if (Type.anyRecord().equals(recordValue.getResultType())) {
+            return List.of();
+        }
```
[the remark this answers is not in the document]
normen662: This is not correct. It should not return an empty list. In fact, there is no reasonable result it could return here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1260,128 @@ static Any any() {
         return Any.INSTANCE;
     }
 
+    /**
+     * Special {@link Type.Record} that is undefined.
```
normen662, line 1264: It's not a `Type.Record` anymore.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1260,128 @@ static Any any() {
...
+        public TypeCode getTypeCode() {
+            return TypeCode.RECORD;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean isNullable() {
+            return true;
```
normen662, line 1290: we should have a `isNullable` as an attribute in this class and return that

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1260,128 @@ static Any any() {
...
+        }
+
+        @Nonnull
+        @Override
+        public AnyRecord withNullability(final boolean newIsNullable) {
+            Verify.verify(newIsNullable);
```
normen662, line 1296: this should set the attribute `isNullable` and it should not rely on it being `true`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -1259,6 +1260,128 @@ static Any any() {
...
+        @Override
+        public RecordQueryPlanProto.PType toTypeProto(@Nonnull final PlanSerializationContext serializationContext) {
+            return RecordQueryPlanProto.PType.newBuilder().setAnyRecordType(toProto(serializationContext)).build();
+        }
+
+        @Nonnull
+        @SuppressWarnings("unused")
+        public static AnyRecord fromProto(@Nonnull final PlanSerializationContext serializationContext,
+                                    @Nonnull final PAnyRecordType anyTypeProto) {
+            return anyRecord();
```
normen662, line 1357: You need to serialize `isNullable` as well.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -69,6 +69,10 @@ message PType {
     // nothing
   }
 
+  message PAnyRecordType {
+      // nothing
```
normen662, line 73: I think you need to serialize

```
optional bool is_nullable = 1;
```

normen662 (changes requested):

normen662 (approved):

PR #2504, opened 2024-02-12
Resolves #2503: Use Locale.ROOT when parsing lucene queries

### LINKED ISSUE #2503: Lucene queries are parsed with the default locale
Lucene queries are parsed with the default locale of the machine. It should instead use a fixed locale, or be configurable via an index option.

This currently means that `FDBLuceneHighlightingTest#highlightedNumberRangeQuery` fails when run on a machine where the default locale doesn't have the same group and decimal separator characters as `en_US`, such as `de`.

ScottDugas:
To help protect us in the future, I also changed it so that if you run with tests.includeRandom it will run with a random locale.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
@@ -44,6 +44,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexExpressions.java
```diff
@@ -158,14 +159,14 @@ public boolean isSorted() {
         public PointsConfig getPointsConfig() {
             switch (type) {
                 case INT:
-                    return new PointsConfig(NumberFormat.getInstance(), Integer.class);
+                    return new PointsConfig(NumberFormat.getInstance(Locale.ROOT), Integer.class);
```
ohadzeliger, line 162: Question: Does that have any implication on the way numbers are formatted and parsed from a user perspective? What does ROOT locale do that is different than Locale.ENGLISH or Locale.US? 
Lucene support in general is restricted to English (see stop words for example). Does ROOT match that definition?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneLocaleTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @ParameterizedTest
+    @ValueSource(strings = {"default", "en_US", "de", "de_DE", "fr"})
+    void highlightedNumberRangeQuery(Locale locale) {
+        final Locale defaultLocale = Locale.getDefault();
+        try {
+            Locale.setDefault(locale);
+            try (FDBRecordContext context = openContext()) {
+                rebuildIndexMetaData(context, COMPLEX_DOC, TEXT_AND_STORED_COMPLEX);
+                recordStore.saveRecord(LuceneIndexTestUtils.createComplexDocument(1623L, "Hello record layer", "Hello record layer 2", 5, 1_234_097, false, 8.123));
```
ohadzeliger, line 75: May be interesting to add a record with a number that is outside the search range

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneLocaleTest.java
ScottDugas: Yeah, makes sense. I added two records, one where everything is greater, and one where everything is smaller.

ohadzeliger (approved):

PR #2507, opened 2024-02-13
Name the grouping keys in the underlying select expression of the aggregate index match candidate

### LINKED ISSUE #2506: Aggregate index match candidate does not name the grouping keys in the underlying select expression.
When constructing the aggregate index match candidate, we construct underlying `select` has a left-most `RCV` containing the grouping keys, there is a bug in the representation of the fields in that group is that their types are unnamed, this causes sometimes matching problems for certain types of queries.

hatyo:
When constructing the aggregate index match candidate, we construct underlying select has a left-most RCV containing the grouping keys, there is a bug in the representation of the fields in that group is that their types are unnamed, this causes sometimes matching problems for certain types of queries.

This PR fixes this behaviour and makes sure that the grouping keys are named.

This fixes #2506


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
```diff
@@ -191,8 +192,21 @@ private Pair<Quantifier, List<Placeholder>> constructSelectWhereAndPlaceholders(
         }
 
         // add an RCV column representing the grouping columns as the first result set column
+        // also, make sure to set the field type names correctly for each field value in the grouping keys RCV.
         final var groupingValue = RecordConstructorValue.ofColumns(
-                baseExpansion.getResultColumns().subList(0, groupingKeyExpression.getGroupingCount()));
+                baseExpansion.getResultColumns().subList(0, groupingKeyExpression.getGroupingCount())
+                        .stream()
+                        .map(column -> {
+                            if (column.getValue() instanceof FieldValue) {
```
normen662, line 200: I thought there was some helper that would allow you to inherit the last field name in the access path of a field value to become the name of the column if that's the input to an RCV(). There is definitely a method to inquire of a field value's last step; maybe that will help find the method I am thinking of.

normen662: If that logic does not already exist it would be useful to add it in RCV if possible so other code can benefit from that.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexExpansionVisitor.java
hatyo: I wasn't aware of that, checking.

hatyo: I did some investigation and tried to move this field name amendment logic to `RecordConstructorValue#ofColumns` to become the following:

```java
    @Nonnull
    private static Type.Record computeResultType(@Nonnull final Collection<Column<? extends Value>> columns) {
        final var fields = columns.stream()
                .map(column -> {
                    if (column.getValue() instanceof FieldValue) {
                        final var fieldValueNameMaybe = ((FieldValue)column.getValue()).getLastFieldName();
                        return fieldValueNameMaybe.map(fieldValueName -> column.getField().withName(fieldValueName)).orElse(column.getField());
                    }
                    return column.getField();
                })
        return Type.Record.fromFields(false, fields);
    }
```

This looks innocent, however doing so broke too many tests that effectively try to do something like:

```sql
select a, a, a, a, a from t;
```

Because the now-duplicate field names are rejected by `Type#normalizeFields` which [guards](https://github.com/FoundationDB/fdb-record-layer/blob/f2ac2b9b35f5f679307cfe2389f75552f294878e/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java#L1975) against this.

It is interesting to see _why_ this works, it is because the `ExpansionVisitor` takes care of [deduping](https://github.com/FoundationDB/fdb-record-layer/blob/f2ac2b9b35f5f679307cfe2389f75552f294878e/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GraphExpansion.java#L177) field names, before they are handed in to `Type#normalize` later on during plan generation.

hatyo: So, I think it is ok to leave it localised here?

normen662: yup

normen662 (approved):

Issue #2505, opened 2024-02-13
Disabled indexes should not participate in deleteRecordsWhere


alecgrieser:
The `RecordsWhereDeleter` currently accumulates a list of index maintainers, and then it validates the deletion predicate can be satisfied on the index. However, when an index is `DISABLED`, it isn't maintained, and so it doesn't need to be included in the calculation. We should be able to skip over that index both when checking the index and when issuing the deletes.

PR #2508, opened 2024-02-14
Fixes #2502: DeleteRecordsWhere should modify predicate for synthetic record types

### LINKED ISSUE #2502: DeleteRecordsWhere should modify predicate for synthetic record types
The `deleteRecordsWhere` method lets the user set a predicate, and if it can match that predicate to a prefix of each index (including the primary key index), then it will execute the delete as a series of range deletes. When working with synthetic record types, we can a hit a problem, though. If you want to delete all records with a certain field, say:

```java
recordStore.deleteRecordsWhere(Query.field("a").equalsValue("foo"))
```

Then that means something like "delete all stored records where the field `a` is equal to `"foo"`". For synthetic record types, that would mean something like "delete all entries where one of the constituents has an `a` field equal to `"foo"`".  Currently, we don't make that translation, so `deleteRecordsWhere` queries on synthetic types will fail.

alecgrieser:
When we delete all elements of a type, we need to modify the check so that it will remove all synthetic recorsd that reference the deleted base elements. This means modifying the predicate slightly. In particular, if there's a delete predicate like:

```java
Query.field("foo").equalsValue("bar")
```

And there's an unnested record type where the parent constituent is named `p`, then the correct predicate on the unnested record is:

```java
Query.field("p").matches(Query.field("foo").equalsValue("bar"))
```

This PR makes a few changes:

1. It makes sure that we gather up the indexes on synthetic types to check whether they are valid. This used to not work for cases where we were indexing only a single record type.
1. It validates that it can modify the predicate. For instance, if the index is on a single unnested record type, it can rewrite the predicate to be on the parent constituent. This check attempts to be a little conservative, completely skipping `JoinedRecordTypes`, which have more complicted semantics here

This fixes https://github.com/FoundationDB/fdb-record-layer/issues/2502.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -1639,8 +1639,22 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
 
     @Override
     public CompletableFuture<Void> deleteRecordsWhereAsync(@Nonnull QueryComponent component) {
+        if (recordStoreStateRef.get() == null) {
+            return preloadRecordStoreStateAsync().thenCompose(ignore -> deleteRecordsWhereAsync(component));
+        }
+
         preloadCache.invalidateAll();
-        return new RecordsWhereDeleter(component).run();
+        recordStoreStateRef.get().beginRead();
+        boolean async = false;
+        try {
+            CompletableFuture<Void> future = new RecordsWhereDeleter(component).run();
+            async = true;
+            return future.whenComplete((ignore, err) -> recordStoreStateRef.get().endRead());
+        } finally {
+            if (!async) {
+                recordStoreStateRef.get().endRead();
+            }
+        }
     }
 
     /**
@@ -1722,7 +1736,10 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                 typelessComponent = remainingComponent;
             }
 
-            indexMaintainers = allIndexes.stream().map(FDBRecordStore.this::getIndexMaintainer).collect(Collectors.toList());
+            indexMaintainers = allIndexes.stream()
+                    .filter(index -> !isIndexDisabled(index))
+                    .map(FDBRecordStore.this::getIndexMaintainer)
+                    .collect(Collectors.toList());
 
             evaluated = deleteRecordsWhereCheckRecordTypes();
             if (recordTypeKeyComparison == null) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -1854,12 +1854,17 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
                 if (syntheticRecordType instanceof UnnestedRecordType) {
                     UnnestedRecordType unnestedRecordType = (UnnestedRecordType) syntheticRecordType;
                     UnnestedRecordType.NestedConstituent parent = unnestedRecordType.getParentConstituent();
-                    if (recordType == null || recordType.equals(parent.getRecordType())) {
-                        if (constituentName != null && !constituentName.equals(parent.getName())) {
-                            return false;
-                        }
-                        constituentName = parent.getName();
+                    if (recordType != null && !recordType.equals(parent.getRecordType())) {
+                        // If the delete is limited to a single type, only process indexes on unnested records of
+                        // precisely that type. In theory, we could use the constituent name as a predicate here,
+                        // but if the different record types have different constituent names, then they can't
+                        // be used to form a prefix.
+                        return false;
                     }
+                    if (constituentName != null && !constituentName.equals(parent.getName())) {
+                        return false;
+                    }
+                    constituentName = parent.getName();
                 } else {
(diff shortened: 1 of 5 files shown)


alecgrieser:
I also added a fix for #2505 in this PR. It's not strictly speaking related, though I suspect we want both in relatively short order. I can separate out that commit into a separate PR, though it will conflict with this PR.

alecgrieser:
Added an additional test for `PERMUTED_MIN` and `PERMUTED_MAX` indexes. Technically unrelated, though there was a question as to whether that index type would have trouble with `deleteRecordsWhere`, so it seemed like a good idea to add another test

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
```diff
@@ -838,4 +911,378 @@ void indexDoubleNestedType() {
...
+            assertDeleteRecordsWhereFails(OUTER, null, "syntheticOther");
+        }
+    }
+
+    @Test
+    void deleteWhereWithDifferentTypedParents() {
+        // Create a multi-type index on two different unnested types. Each one has a parent constituent named and an inner constituent, so the index is well-defined.
+        // The first column in the index is middle.rec_no, which we also set as the first column in the combined primary key. This means that it should be legal to
+        // perform a delete where with a middle.rec_no predicate
+        final RecordMetaDataHook hook = addMultiTypeDoubleUnnestedIndex()
```
ScottDugas, line 1125: I don't have a good answer for this, and any answer would probably be a fair amount of work, but these tests are kind of hard to follow because it's hard to keep track of what things like `addMultiTypeDoubleUnnestedIndex` actually amount to.
I suspect the answer is to either:
1. Have some more larger, models, that have intuitive meanings that we could learn. 
2. Have a more fluent / pithy way of defining metadata so that a test could pretty easily have everything here.

Both of those are fairly sizable bodies of work that require some real thought, so I definitely don't think we should bring them into this PR, but something to keep in mind.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
```diff
@@ -838,4 +911,378 @@ void indexDoubleNestedType() {
             commit(context);
         }
     }
+
+    @Test
+    void deleteRecordsWhere() {
```
ScottDugas, line 916: Is there any need to add a test that has an `and` in the `deleteWhere` predicate? I don't see one, but I may have forgotten.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1763,29 +1794,106 @@ private void deleteRecordsWhereCheckIndexes() {
...
+                    for (SyntheticRecordType.Constituent constituent : syntheticRecordType.getConstituents()) {
+                        if (constituentName.equals(constituent.getName()) && !recordType.equals(constituent.getRecordType())) {
+                            return false;
+                        }
+                    }
                 }
             }
+            if (typelessComponent == null) {
+                return true;
+            }
```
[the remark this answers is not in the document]
alecgrieser: Yeah

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
alecgrieser: Yeah, there's a bit of a trade-off here for brevity and clarity. The way the tests are now, it's actually fairly easy to use the building blocks to create new tests, but only if you already know what the building blocks are. There may be ways with the macros (or macro-like code) to gain back a bit more clarity without sacrificing brevity, but I'm not sure I know how.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1763,29 +1794,123 @@ private void deleteRecordsWhereCheckIndexes() {
...
+                    return false;
+                }
+            }
+            if (constituentName == null) {
+                return false;
+            }
+            if (recordType != null) {
+                // In the first loop, we have validated that there is a single constituent that contains elements
+                // of the record type we're looking to delete. In this loop, make sure that there isn't a constituent
+                // with that name that is of a different type
```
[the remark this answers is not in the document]
alecgrieser: Hm, so, this loop is here because in an earlier version of the code, the first loop would just ignore any unnested records when `recordType != null` where the record type didn't match the desired type. So this loop was there to validate that none of those ignored constituents had the same name as the one we were going to add to the predicate, because that would have been incorrect.

Later, though, I realized that there was a possibility that someone might do something like:

1. Define two unnested record types on different stored types with different names for the parent constituent
2. Construct a multi-type index on those two types
3. Do a delete where on one of the stored types with no other predicates

In that case, there was a bug that would result in the entire index being cleared out. So I made the first loop also validate that the type of each parent matches `recordType` (if not null), so this loop is extraneous now. There was already a test, `deleteWhereOnMultiTypeFailsWithAmbiguousParent`, that validated this, and that test passes without this loop.

It's true, though, that we will have to be a little more careful with joined record types. I'll go ahead and remove this loop for now, though, as it's not really doing anything

MMcM (approved):

PR #2514, opened 2024-02-15
make all the pattern settings the same

ammolitor:
Trying this to see if tests get timestamps.


ammolitor:
This did not do what I hoped it would, closing.

PR #2510, opened 2024-02-14
Resolves #2509: Use metrics for rebalancing partitions

ScottDugas:
This downgrades the log to debug level, and adds metrics for timing, and document counts.

PR #2519, opened 2024-02-19
add SizeEvent to the storeTimer

g31pranjal:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBTransactionContext.java
```diff
@@ -198,6 +198,20 @@ public void record(@Nonnull StoreTimer.Event event, long timeDelta) {
         }
     }
 
+    /**
+     * Record an IO event with the size (in bytes) of data read/written.
```
ScottDugas, line 202: Again, I think the unit should be from the `SizeEvent`, not fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -376,22 +378,32 @@ public interface Wait extends StoreTimer.Event {
...
-        boolean isSize();
+        default boolean isSize() {
+            return false;
+        }
     }
 
     /**
-     * Contains the number of occurrences and cummulative time spent on an associated {@link StoreTimer.Event}.
+     * {@link Event}s that only count the number of occurrences of a read/write operation and also the size read/written.
+     * There is no meaningful time duration associated with these events.
```
ScottDugas, line 396: As implemented, isn't the time going to return whatever the size is?
That seems somewhat less than ideal to have it be ambiguous.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
```diff
@@ -268,4 +262,42 @@ public boolean isSize() {
...
+        LUCENE_WRITE("lucene index writes", true),
+        /** Write Stored Fields operations on the FDBDirectory. */
+        LUCENE_WRITE_STORED_FIELDS("lucene write stored fields"),
+
+        ;
+
+        private final String title;
+        private final boolean delayedUntilCommit;
+
+
```
ScottDugas, line 283:
```suggestion
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
```diff
@@ -268,4 +262,42 @@ public boolean isSize() {
...
+     */
+    public enum SizeEvents implements StoreTimer.SizeEvent {
+
+        /** writeFileReference operation in the FDBDirectory.*/
+        LUCENE_WRITE_FILE_REFERENCE("lucene write file references"),
+        /** writeData operation in FDBDirectory. */
+        LUCENE_WRITE("lucene index writes", true),
+        /** Write Stored Fields operations on the FDBDirectory. */
+        LUCENE_WRITE_STORED_FIELDS("lucene write stored fields"),
+
```
ScottDugas, line 277:
```suggestion
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
```diff
@@ -189,10 +189,6 @@ public String logKey() {
     public enum Counts implements StoreTimer.Count {
         /** Number of times the getIncrement() function is called in the FDBDirectory. */
         LUCENE_GET_INCREMENT_CALLS("lucene increments", false),
-        /** Number of writeFileReference calls in the FDBDirectory.*/
-        LUCENE_WRITE_FILE_REFERENCE_CALL("lucene write file references", false),
-        /** Total number of bytes that were attempted to be written (not necessarily committed) for file references in the FDBDirectory. */
-        LUCENE_WRITE_FILE_REFERENCE_SIZE("lucene write file reference size", true),
         /** Count of writeData calls in FDBDirectory. */
         LUCENE_WRITE_CALL("lucene index writes", false, null, true),
```
ScottDugas, line 193: Can `LUCENE_WRITE_CALL` be removed?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -339,7 +338,7 @@ Stream<Pair<Long, byte[]>> getAllFieldInfosStream() {
                 .map(keyValue -> Pair.of(fieldInfosSubspace.unpack(keyValue.getKey()).getLong(0), keyValue.getValue()));
     }
 
-    public static boolean isSegmentInfo(String name) {
+    public static boolean    isSegmentInfo(String name) {
```
ScottDugas, line 341:
```suggestion
    public static boolean isSegmentInfo(String name) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -376,22 +378,32 @@ public interface Wait extends StoreTimer.Event {
...
+     */
+    public interface SizeEvent extends StoreTimer.Event {
```
ScottDugas, line 398: There should be an issue associated with this change, and a release notes update.

ScottDugas, line 398: There should be some new tests in `FDBStoreTimerTest` that `SizeEvent`s do the right thing for:
1. `recordSize`
2. snapshot
3. diff
4. clear

ScottDugas: I created an issue: https://github.com/FoundationDB/fdb-record-layer/issues/2525

g31pranjal: thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreSubTimer.java
```diff
@@ -50,6 +50,14 @@ public void record(final StoreTimer.Event event, final long timeDifferenceNanos)
         super.record(event, timeDifferenceNanos);
     }
 
+    @Override
+    public void recordSize(SizeEvent event, long bytes) {
```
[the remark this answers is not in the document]
g31pranjal: Yeah, that make sense

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -147,6 +148,7 @@ private static void computeDifference(@Nonnull Map<Event, Counter> timerCounters
      * @return the counter for the event or {@code null} if not counter exists
      */
     @Nullable
+    @VisibleForTesting
```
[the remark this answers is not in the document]
g31pranjal: done

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
g31pranjal: done

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
g31pranjal: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
g31pranjal: added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
g31pranjal: Beneath the scene, yes. However, the converse is not true because for getting and recording the size, the consumer would require explicitly a SizeEvent. If it's really pressing, we can separate the concerns.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
g31pranjal: yes, left by mistake

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
ScottDugas: The release notes still need to be updated, and it would probably be good to have use the `Resolves` syntax in the commit message to auto-resolve the issue.
See: https://github.com/FoundationDB/fdb-record-layer/pull/2510/files#diff-53d9d97604b51916cd0e08d61a4e86d6357b9a672ee28c624df8aca4ec50647dL24 as an example

See also: https://github.com/FoundationDB/fdb-record-layer/blob/main/CONTRIBUTING.md#opening-a-pull-request

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
ScottDugas: I think there's not really a good way to adjust.
If it weren't for backwards compatibility, I would say that we should have `TimeEvent` and `SizeEvent` as siblings in a hierarchy. rather than having `SizeEvent` extending `Event` which implies time.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
g31pranjal: Sorry, did that now!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -436,17 +444,26 @@ public int getCount() {
          * @return the cumulative time spent on the associated event
          */
         public long getTimeNanos() {
-            return timeNanos.get();
+            return getCumulativeValue();
```
alecgrieser, line 447: We should probably note that this method only returns the time for time-based events. The backwards compatibility argument for leaving this in at all rather than splitting event into `TimeEvent` and `SizeEvent` makes sense to me, but we probably want to spell that out in documentation for clarity's sake.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -388,10 +388,18 @@ public interface Count extends StoreTimer.Event {
     }
 
     /**
-     * Contains the number of occurrences and cummulative time spent on an associated {@link StoreTimer.Event}.
+     * {@link Event}s that count the number of occurrences of an operation and also the size of the operation.
+     * There is no meaningful time duration associated with these events.
```
alecgrieser, line 392: I think there should probably be something here to explain why someone would want to use this over, say, a `Count`--particularly how this differs from `Count` events where `isSize()` is true.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
```diff
@@ -268,4 +256,40 @@ public boolean isSize() {
...
+     * Size Events.
+     */
+    public enum SizeEvents implements StoreTimer.SizeEvent {
+
+        /** writeFileReference operation in the FDBDirectory.*/
+        LUCENE_WRITE_FILE_REFERENCE("lucene write file references"),
+        /** writeData operation in FDBDirectory. */
+        LUCENE_WRITE("lucene index writes", true),
+        /** Write Stored Fields operations on the FDBDirectory. */
+        LUCENE_WRITE_STORED_FIELDS("lucene write stored fields"),
```
alecgrieser, line 270: In addition to these Lucene-related events, there are a few core Record Layer count events that I think it would make sense to update. For instance, we could probably remove/deprecate practically all of the events where there is both a `BLAH_BLAH` event and a `BLAH_BLAH_BYTES` event in favor of a single event. We don't necessarily need to do that in this PR, though

alecgrieser (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneEvents.java
g31pranjal: Yeah, I see that there are lot many of that sort. I am not looking to do it in this PR but there is definitely a possibility of follow-up work that cleans up "size" metrics.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -114,7 +114,7 @@ void testInsertDocuments(boolean useOptimizedStoredFieldsFormat) throws Exceptio
                     assertDocCountPerSegment(directory, List.of("_0"), List.of(3));
                 }
                 assertTrue(timer.getCounter(LuceneEvents.Waits.WAIT_LUCENE_GET_STORED_FIELDS).getCount() > 1);
-                assertTrue(timer.getCounter(LuceneEvents.Counts.LUCENE_WRITE_STORED_FIELDS).getCount() >= 3);
+                assertTrue(timer.getCounter(LuceneEvents.SizeEvents.LUCENE_WRITE_STORED_FIELDS).getCount() >= 3);
```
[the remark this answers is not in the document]
g31pranjal: 1. Yeah, I did realize that. However, I am leaving this as is given the context of this PR.
2. I am not sure about that. `LUCENE_WRITE_STORED_FIELDS` was used as a count and reported size to `LUCENE_WRITE_SIZE`. However, the size is not tested here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
g31pranjal: I am not quite sure why `isSize()` is used and was introduced. It seems to be there just to distinguish "true" Count events from "size" counts in `CountAggregate` event but that doesn't seem to be a sufficient use case.

g31pranjal: Added a brief description here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -389,7 +389,13 @@ public interface Count extends StoreTimer.Event {
 
     /**
      * {@link Event}s that count the number of occurrences of an operation and also the size of the operation.
-     * There is no meaningful time duration associated with these events.
+     * {@link SizeEvent} is similar in implementation to (Time){@link Event} except for the fact that the recorded value
+     * is not the time. It is also different from {@link Count} since along with the size of the event in concern, it
+     * also records the number of times the record has occurred.
```
alecgrieser, line 394:
```suggestion
     * also records the number of times the event has occurred.
```

In theory, you could re-word this as "the number of times the event has been recorded", I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -389,7 +389,13 @@ public interface Count extends StoreTimer.Event {
...
+     *
+     * This is in principle similar to having 2 {@link Count} event like, {@code COUNT_NUM} and {@code COUNT_SIZE} with
```
alecgrieser, line 396:
```suggestion
     * This is in principle similar to having two {@link Count} events like, {@code COUNT_NUM} and {@code COUNT_SIZE} with
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -630,7 +638,7 @@ public void increment(@Nonnull Count event, int amount) {
     /**
      * Get the total time spent for a given event.
      *
-     * @param event the event to get time information for
+     * @param event the event to get time information for.
```
alecgrieser, line 641:
```suggestion
     * @param event the event to get time information for
```

Parameters (and return values) don't get periods because they are sentence fragments

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/StoreTimer.java
```diff
@@ -389,7 +389,13 @@ public interface Count extends StoreTimer.Event {
...
     /**
      * {@link Event}s that count the number of occurrences of an operation and also the size of the operation.
-     * There is no meaningful time duration associated with these events.
+     * {@link SizeEvent} is similar in implementation to (Time){@link Event} except for the fact that the recorded value
+     * is not the time. It is also different from {@link Count} since along with the size of the event in concern, it
+     * also records the number of times the record has occurred.
+     *
+     * This is in principle similar to having 2 {@link Count} event like, {@code COUNT_NUM} and {@code COUNT_SIZE} with
+     * {@link Count#isSize()} is equal to true for the latter. However, in this case the consumer ends up registering 2
+     * events in case of an occurrence and also maintains how they are interpreted by viewers.
```
alecgrieser, line 398: I feel like I _kind of_ understand what this is trying to say, but it's a bit muddled. I think you're trying to say that by combining the count and size into a single event, it becomes easier for the user to manage because they only need to record a single occurrence to update both and means that the link between the total and count is more obvious when reading metrics. It may need some word smithing if that's what you mean

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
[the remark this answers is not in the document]
alecgrieser: > I am not sure about that. LUCENE_WRITE_STORED_FIELDS was used as a count and reported size to LUCENE_WRITE_SIZE. However, the size is not tested here.

My point was just that because you changed how these size events are recorded, we should probably have coverage of the new functionality within the test. If you feel like the other tests here are enough for that, then it's probably fine

alecgrieser (changes requested):

alecgrieser (dismissed):

ScottDugas (approved):

PR #2524, opened 2024-02-21
RESOLVES #2523 Harden Lucene continuations with respect to background…

### LINKED ISSUE #2523: Harden Lucene continuations with respect to background work
Ensure that continuations will work when background work to do merges, rearrange between partitions, and adding new documents is ongoing in the background.

wshaib-apple:
… work

Cursor continuations should be more resilient with respect to underlying cross-partition data changes.
 - When the search is sorted by the partition timestamp, the cursor will attempt to continue from the same timestamp value, even when it lies in a different partition on continuation.
 - When the search is unsorted, or sorted by a field other than the partition timestamp, the cursor will continue from the same relative position within the partition it stopped at. If that position is no longer in range of the partition, the cursor will jump to the next partition and continue from there. In this case, some records that we returned previously could be returned again.

Additionally, a small unrelated change was made to a couple of blocking methods in `LucenePartitioner`, converting them to async for better performance.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -54,6 +54,7 @@
       <option name="DOWHILE_BRACE_FORCE" value="3" />
       <option name="WHILE_BRACE_FORCE" value="3" />
       <option name="FOR_BRACE_FORCE" value="3" />
+      <option name="WRAP_LONG_LINES" value="true" />
     </codeStyleSettings>
     <codeStyleSettings language="XML">
       <indentOptions>
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -54,7 +54,6 @@
       <option name="DOWHILE_BRACE_FORCE" value="3" />
       <option name="WHILE_BRACE_FORCE" value="3" />
       <option name="FOR_BRACE_FORCE" value="3" />
-      <option name="WRAP_LONG_LINES" value="true" />
     </codeStyleSettings>
     <codeStyleSettings language="XML">
       <indentOptions>
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
@@ -72,7 +72,10 @@ class LuceneCursorContinuation implements RecordCursorContinuation {
         return false;
     }
 
-    public static LuceneCursorContinuation fromScoreDoc(ScoreDoc scoreDoc, @Nullable Integer partitionId, @Nullable Long partitionTimestamp) {
+    public static LuceneCursorContinuation fromScoreDoc(ScoreDoc scoreDoc,
+                                                        @Nullable Integer partitionId,
+                                                        @Nullable Long partitionTimestamp,
+                                                        @Nullable Long anchorTimestamp) {
         LuceneContinuationProto.LuceneIndexContinuation.Builder builder = LuceneContinuationProto.LuceneIndexContinuation.newBuilder()
                 .setDoc(scoreDoc.doc)
                 .setShard(scoreDoc.shardIndex)
@@ -83,6 +86,9 @@ class LuceneCursorContinuation implements RecordCursorContinuation {
         }
         if (partitionTimestamp != null) {
             builder.setPartitionTimestamp(partitionTimestamp);
+        }
+        if (anchorTimestamp != null) {
+            builder.setAnchorTimestamp(anchorTimestamp);
         }
         if (scoreDoc instanceof FieldDoc) {
             for (Object field : ((FieldDoc)scoreDoc).fields) {
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
@@ -74,8 +74,7 @@ class LuceneCursorContinuation implements RecordCursorContinuation {
 
     public static LuceneCursorContinuation fromScoreDoc(ScoreDoc scoreDoc,
                                                         @Nullable Integer partitionId,
-                                                        @Nullable Long partitionTimestamp,
-                                                        @Nullable Long anchorTimestamp) {
+                                                        @Nullable Long partitionTimestamp) {
         LuceneContinuationProto.LuceneIndexContinuation.Builder builder = LuceneContinuationProto.LuceneIndexContinuation.newBuilder()
                 .setDoc(scoreDoc.doc)
                 .setShard(scoreDoc.shardIndex)
@@ -86,9 +85,6 @@ class LuceneCursorContinuation implements RecordCursorContinuation {
         }
         if (partitionTimestamp != null) {
             builder.setPartitionTimestamp(partitionTimestamp);
-        }
-        if (anchorTimestamp != null) {
-            builder.setAnchorTimestamp(anchorTimestamp);
         }
         if (scoreDoc instanceof FieldDoc) {
             for (Object field : ((FieldDoc)scoreDoc).fields) {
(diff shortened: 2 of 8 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1158,18 +1162,208 @@ void partitionedJoinedIndexTest() {
...
+            // so now we should get the remaining docs
+            List<IndexEntry> entries = indexEntryCursor.asList().join();
+            final Set<Tuple> expectedKeys;
+            final RecordCursor.NoNextReason noNextReason;
+            final RecordCursorResult<IndexEntry> lastResult = indexEntryCursor.onNext().get();
+            final int expectedCount;
+            if (sortType == SortType.ASCENDING) {
+                expectedKeys = Set.copyOf(primaryKeys.subList(18, 25));
+                expectedCount = 7;
+                noNextReason = RecordCursor.NoNextReason.SOURCE_EXHAUSTED;
```
ScottDugas, line 1352: Due to the fact that you just start with `0`, this ends up continuing in the same partition id as it was before. It would probably be good if this continued in partition 3.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1158,18 +1162,208 @@ void partitionedJoinedIndexTest() {
...
+                // ungrouped, simple
+                Arguments.of(false, false, SortType.ASCENDING),
+                Arguments.of(false, false, SortType.DESCENDING),
+                Arguments.of(false, false, SortType.UNSORTED)
+        );
+    }
+
+    @ParameterizedTest(name = "isGrouped: {0}, isSynthetic: {1}, sort type: {2}")
+    @MethodSource
+    void continuationDuringRepartitioningTest(boolean isGrouped,
```
ScottDugas, line 1211: As a followup it would probably be good to also add a search to the randomized test. It won't have as strict assertions, but it could assert that the two sorted end up getting all the ids, and the unsorted doesn't throw an error. It could create a couple queries with random limits before partitioning, and then continue to get the remainder.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
```diff
@@ -84,6 +87,9 @@ public static LuceneCursorContinuation fromScoreDoc(ScoreDoc scoreDoc, @Nullable
         if (partitionTimestamp != null) {
             builder.setPartitionTimestamp(partitionTimestamp);
         }
+        if (anchorTimestamp != null) {
+            builder.setAnchorTimestamp(anchorTimestamp);
+        }
```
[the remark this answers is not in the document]
wshaib-apple: I added `anchorTimestamp` to the continuation to store the document's timestamp we stopped at when sorting by partition timestamp field (we need that to figure out which partition we should be in on continuing).
However, on second look, I think the sort field will be store in the `FieldDoc` which is already in the continuation, so no need to store the value explicitly. I have made the change to remove `anchorTimestamp`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -173,11 +180,21 @@ public class LuceneRecordCursor implements BaseCursor<IndexEntry> {
         this.timer = state.context.getTimer();
         this.query = query;
         this.sort = sort;
+        partitioner = new LucenePartitioner(state);
```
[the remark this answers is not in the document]
wshaib-apple: Fixed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -215,78 +238,145 @@ public CompletableFuture<RecordCursorResult<IndexEntry>> onNext() {
...
-                try {
-                    maybePerformScan();
-                    return lookupResults.onNext();
-                } catch (IOException ioException) {
-                    throw new RecordCoreException(ioException)
-                            .addLogInfo(LogMessageKeys.QUERY, query);
+        final CompletableFuture<Long> nextPartitionTimestamp;
+        if (sortedByPartitioningKey && !isReverseSort) {
+            // if we're in a partitioning-field-ascending-sort query, get the next more recent partition
+            nextPartitionTimestamp = partitioner.getPartitionMetaInfoById(partitionId, groupingKey).thenApply(curPartition -> LucenePartitioner.getTo(curPartition) + 1);
```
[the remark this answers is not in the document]
wshaib-apple: good catch. updated.

.idea/codeStyles/Project.xml
```diff
@@ -54,6 +54,7 @@
       <option name="DOWHILE_BRACE_FORCE" value="3" />
       <option name="WHILE_BRACE_FORCE" value="3" />
       <option name="FOR_BRACE_FORCE" value="3" />
+      <option name="WRAP_LONG_LINES" value="true" />
```
ScottDugas, line 57: I'm not sure that we want to take this in. 
```suggestion
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -854,4 +859,24 @@ public static CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> ge
...
+            @Nullable final Long currentPartitionFromTimestamp,
+            @Nonnull final Subspace indexSubspace) {
+        if (currentPartitionFromTimestamp == null) {
+            return getNewestPartition(groupingKey, context, indexSubspace);
+        } else {
+            final Range range = new TupleRange(
+                    partitionMetadataKeyTuple(groupingKey, currentPartitionFromTimestamp),
+                    partitionMetadataKeyTuple(groupingKey, Long.MAX_VALUE),
+                    EndpointType.TREE_START,
+                    EndpointType.RANGE_EXCLUSIVE)
```
ScottDugas, line 875: This looks wrong, I think you want
```suggestion
                    partitionMetadataKeyTuple(groupingKey, currentPartitionFromTimestamp),
                    partitionMetadataKeyTuple(groupingKey),
                    EndpointType.RANGE_EXCLUSIVE,
                    EndpointType.TREE_END)
```

1. what happens if someone has a timestamp of `Long.MAX_VALUE`?
2. I think the difference between `TREE_START` and `RANGE_EXCLUSIVE` is more semantics than actual behavior difference. But the key is what you want it to return if there's nothing newer. Do you return the current partition, or null.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -226,7 +228,6 @@ public class LuceneIndexTest extends FDBRecordStoreTestBase {
     private static Index complexPartitionedIndex(final Map<String, String> options) {
         return new Index("Complex$partitioned",
                 concat(function(LuceneFunctionNames.LUCENE_TEXT, field("text")),
-                        function(LuceneFunctionNames.LUCENE_STORED, field("timestamp")),
```
ScottDugas: This is a nice win that I did not foresee with the removal of the `anchorTimestamp`

ScottDugas (changes requested):

ScottDugas (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -108,6 +109,9 @@ public LucenePartitioner(@Nonnull IndexMaintainerState state) {
         if (partitioningEnabled && (partitionTimestampFieldName.isEmpty() || partitionTimestampFieldName.isBlank())) {
             throw new RecordCoreArgumentException("Invalid partition timestamp field name", LogMessageKeys.FIELD_NAME, partitionTimestampFieldName);
         }
+        // partition timestamp field name in lucene, when nested, has `_` in place of `.`
+        partitionTimestampFieldNameInLucene = partitionTimestampFieldName == null ? null : partitionTimestampFieldName.replace('.', '_');
+
```
MMcM, line 114: I wonder whether this could be done at a higher level so that here we only see a consistent naming scheme. Not needed for this PR, though.

MMcM (approved):

PR #2527, opened 2024-02-22
Resolve #2526: Scrubber: eliminate missing record exception for dangl…

### LINKED ISSUE #2526: Scrubber: eliminate exceptions for dangling index entries of synthetic indexes

jjezra:
…ing synthetic index entries.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
```diff
@@ -191,8 +191,11 @@ private CompletableFuture<FDBStoredRecord<Message>> deleteIndexIfDangling(FDBRec
...
         final IndexEntry indexEntry = indexEntryResult.get();
         if (indexContext.isSynthetic) {
-            return store.loadSyntheticRecord(indexEntry.getPrimaryKey()).thenApply(syntheticRecord -> {
-                if (syntheticRecord.getConstituents().isEmpty()) {
+            return store.loadSyntheticRecord(indexEntry.getPrimaryKey()).handle((syntheticRecord, ex) -> {
+                // The null hypothesis is that it is not dangling.
+                final boolean isDangling = ((findException(ex, RecordDoesNotExistException.class) != null) ||
+                                                  (syntheticRecord != null && syntheticRecord.getConstituents().isEmpty()));
+                if (isDangling) {
                     // None of the constituents of this synthetic type are present, so it must be dangling
```
alecgrieser, line 199: I believe this will identify anything where _any_ of the constituents is not found as dangling, which is different from the comment. It probably is the case that we want that behavior, I think.

It is a bit weird that this uses exception handling as control flow, though. If we wanted to fix that, I think we'd need a new variant of `loadSyntheticRecord` that takes an `IndexOrphanBehavior` argument, or something like that.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -411,9 +412,13 @@ void testDetectDanglingUnnestedIndex() {
...
                         .build())
                 .build()) {
-            assertThrows(RecordDoesNotExistException.class, indexScrubber::scrubDanglingIndexEntries);
+            indexScrubber.scrubDanglingIndexEntries();
             indexScrubber.scrubMissingIndexEntries();
         }
+        Assertions.assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
+        Assertions.assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_DANGLING_ENTRIES));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_MISSING_ENTRIES));
```
alecgrieser, line 421: It would probably be nice to extend this test so that we validate that it will delete the dangling entries if `setAllowRepair` is `true` and that that results in a clean index (similar to what the equivalent test for missing entries does).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -411,9 +412,13 @@ void testDetectDanglingUnnestedIndex() {
                         .setAllowRepair(false)
                         .build())
                 .build()) {
-            assertThrows(RecordDoesNotExistException.class, indexScrubber::scrubDanglingIndexEntries);
+            indexScrubber.scrubDanglingIndexEntries();
             indexScrubber.scrubMissingIndexEntries();
         }
+        Assertions.assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
```
alecgrieser, line 418: I think we can be more precise with this assertion. At the very least, this should be:

```
assertThat(timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED), greaterThan(0));
```

To make the assertion a little more readable, but we actually know the expected number of records to scan should be `5 * numRecords`, I think

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -568,9 +573,13 @@ void testDetectDanglingFromJoinedIndex() {
...
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
+        Assertions.assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_DANGLING_ENTRIES));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_MISSING_ENTRIES));
```
alecgrieser, line 582: In theory, we could do with an additional test here that validates the behavior if there's a join index and one (but not all) of the constituents is deleted. I believe, as written, it will identify those as dangling (and I think that's right)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingScrubDangling.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
jjezra: ✅

PR #2531, opened 2024-02-23
Fixes #2530: Synthetic records use wrong configuration if built during checkVersion

### LINKED ISSUE #2530: Synthetic records use wrong configuration if built during checkVersion
After #2513, when building an index on a synthetic record type during `checkVersion`, the `OnlineIndexer` is now being set with a configuration that does not remember that it is one synthetic types, and so the index can be built with incorrect entries. The cause for this is that:

* `RecordMetaData.getIndexesSince(int sinceVersion)` (which returns a map from indexes modified since `sinceVersion` to the set of types the index is defined on) now returns a collection of the underlying _stored_ record types, not the synthetic types. This allows us to more correctly make determinations about whether the index is on new types or not
* If we decide to rebuild an index, that collection of types is passed to the `OnlineIndexer` in `FDBRecordStore.rebuildIndex`
* The `OnlineIndexer` constructs `IndexingContexts` (which includes book-keeping about whether the index is on synthetic types), but it bases that determination off of the pre-set types if the record types are set

From what I can tell, this call in `FDBRecordStore.rebuildIndex` is the _only_ time we pre-set the record types within the Record Layer, though adopters are free to do so when they construct the `OnlineIndexer`.

alecgrieser:
This updates the logic used to build indexes during check version so that it no longer takes the record type list from `getIndexesSince`. This means that the online indexer recalculates the list of record types it needs, which allows it to properly note whether the index is on synthetic records or not. The new test used to fail before this change.

This fixes #2350.


jjezra (approved):
Nice!

PR #2533, opened 2024-02-23
Resolves ##2532; Add back deleteRecordsWhere support for join indexes

ScottDugas:
This adds back rudimentary support for deleteRecordsWhere on join indexes.
This is a rather limited approach to implementing deleteRecordsWhere, and there are tons of situations where it could execute deleteRecordsWhere but this doesn't support it.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -844,126 +840,6 @@ void repartitionAndMerge(Index index, Tuple groupingKey, int repartitionCount, i
         }
     }
 
-    static Stream<Arguments> randomizedRepartitionTest() {
```
ScottDugas, line 847: This test was moved to `LuceneIndexMaintenanceTest`. The primary reason I moved it was to decouple it from the metadata used in this class. The metadata in this class has a bunch of other, unused types, that don't have a `group` field, thus preventing `deleteRecordsWhere`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1839,7 +1850,54 @@ private boolean canDeleteWhereForIndexOnStoredTypes(@Nonnull final IndexMaintain
...
+            for (RecordType indexRecordType : recordTypesForIndex) {
+                // currently do not support multitype indexes with joined types
+                if (!(indexRecordType instanceof JoinedRecordType) || joinedRecordType != null) {
+                    // this should only be called when all of them are JoinedRecordTypes
+                    return false;
+                }
+                joinedRecordType = (JoinedRecordType)indexRecordType;
+            }
+
+            if (typelessComponent != null &&
```
alecgrieser, line 1872: Should there be a check that `recordTypeKeyComparison == null` here? I think it may technically be unnecessary, though I think we'd want slightly different logic for the single-type case. In particular:

1. We can clear out the index entirely if `typelessComponent` is null and there are no outer joins
2. We don't actually need to find a join that matches, but if there isn't one, we just need to match against the constituent of the relevant type
3. If we find a join on the relevant type that matches the delete-where criteria, we could theoretically translate that to a query on a different constituent (potentially applying some compensation to account for different field names)

I think what we have here is sort of a special-case of (3), in that we require that the non-participating types match the predicate even if that's not necessary.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBJoinIndexTest.java
```diff
@@ -0,0 +1,286 @@
...
+                        Key.Expressions.field("parent").nest(Key.Expressions.field("int_value")),
+                        Key.Expressions.field("child").nest(Key.Expressions.field("other_value")));
+            } else if (groupInWrongSpot) {
+                concat = Key.Expressions.concat(
+                        Key.Expressions.field("parent").nest(Key.Expressions.field("int_value")),
+                        Key.Expressions.field("parent").nest(Key.Expressions.field("group")),
+                        Key.Expressions.field("child").nest(Key.Expressions.field("other_value")));
+            } else {
+                concat = Key.Expressions.concat(
+                        Key.Expressions.field("parent").nest(Key.Expressions.field("group")),
```
alecgrieser, line 205: Is it worth adding a check that the first column could be `field("child").nest("group")` instead (that is, that using the first constituent isn't special)?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBJoinIndexTest.java
```diff
@@ -0,0 +1,286 @@
...
+ * Tests of various join index test behaviors.
+ */
+@Tag(Tags.RequiresFDB)
+public class FDBJoinIndexTest extends FDBRecordStoreTestBase {
+
+    private static final String JOIN_TYPE = "JoinChildren";
+    private static final String JOIN_INDEX = "AJoinIndex";
+
+    @Test
+    void deleteWhereValue() {
```
alecgrieser, line 68: I suppose in theory this could use a test for delete-where with a type filter predicate. If we ban those for joined index types then it would probably be sufficient just to validate that that throws an error.

I'm trying to think if there are other interesting cases that aren't covered under test here, and I can't really think of any other obvious ones.

alecgrieser (changes requested):

alecgrieser (approved):

PR #2540, opened 2024-02-27
Lucene filelock heartbeat

jjezra:


docs/ReleaseNotes.md
```diff
@@ -29,8 +29,10 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 * **Performance** Improvement 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Add an option for other non-time size events [(Issue #2525)](https://github.com/FoundationDB/fdb-record-layer/issues/2525)
+* **Feature** FDBDirectoryLockFactory: Implement heartbeat at ensureValid [(Issue #2539)](https://github.com/FoundationDB/fdb-record-layer/issues/2539)
+* **Feature** FDBDirectoryLockFactory: Implement heartbeat at ensureValid [(Issue #2539)](https://github.com/FoundationDB/fdb-record-layer/issues/2539)
+* **Feature** FDBDirectoryLockFactory: Implement heartbeat during ensureValid [(Issue #2539)](https://github.com/FoundationDB/fdb-record-layer/issues/2539)
```
ScottDugas, line 35:
```suggestion
* **Feature** FDBDirectoryLockFactory: Implement heartbeat during ensureValid [(Issue #2539)](https://github.com/FoundationDB/fdb-record-layer/issues/2539)
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -58,43 +58,38 @@ private static class FDBDirectoryLock extends Lock {
...
 
         @Override
         public void close() {
-            fileLockClear(timeStampMillis);
+            fileLockClear();
             closed = true;
         }
 
         @Override
         public void ensureValid() {
```
ScottDugas, line 83: Looking at the call hierarchy, this is only used by `LockValidatingDirectoryWrapper`
Are we always using that, such that we can trust that any transaction doing a write would call `ensureValid`, or is it possible that it wouldn't validate, at the extreme, until it completes the merge.

ScottDugas: It looks like `IndexWriter` always wraps the directory, so we should be ok, but it could become a problem if more parts of the codec are changed to bypass the file-api and interact directly with `FDBDirectory` directly.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -107,40 +102,58 @@ private static long fileLockValueToTimestamp(byte[] value) {
...
-                                                        .addLogInfo(LuceneLogMessageKeys.LOCK_EXISTING_TIMESTAMP, existingTimeStamp,
-                                                                LuceneLogMessageKeys.LOCK_DIRECTORY, this);
-                                            }
-                                            // Here: this lock is either too old, or in the future. Steal it
-                                            if (LOGGER.isWarnEnabled()) {
-                                                LOGGER.warn(KeyValueLogMessage.of("FileLock: Set: found old lock, discard it",
-                                                        LuceneLogMessageKeys.LOCK_EXISTING_TIMESTAMP, existingTimeStamp,
-                                                        LuceneLogMessageKeys.LOCK_DIRECTORY, this));
+        public void fileLockSet(boolean isHeartbeat) {
+            synchronized (fileLockSetLock) {
```
ScottDugas, line 106: This could cause issues with parallel threads being consumed waiting for the synchronized, but unable to actually start the asynchronous handling below, which could cause deadlocks in the reading, since you are synchronizing around the set. 
Can we pull the get out of the `synchronized` block.
You may have to store `this.timeStampMillis` in a temporary variable, and check that it didn't change while you did the read.

ScottDugas (changes requested):

docs/ReleaseNotes.md
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
jjezra: I agree. There is a strong assumption that `ensureValid` is called periodically. If we'll identify a problem, a plan 'B' might be to hook an `ensureValid` call to the agilityContext.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
jjezra: I'm not sure how it can be done without including the `get' in a synchronized block. 
For example: 
```
myTimeStamp = this.timeStampMillis;  <<== local timeStampMillis at time A
agilityContext.asyncToSync(...
                        agilityContext.apply(aContext ->
                                aContext.ensureActive().get(fileLockKey) <<== reads remote timeStampMillis at time B
```
Or 
```
agilityContext.asyncToSync(...
                        agilityContext.apply(aContext ->
                                aContext.ensureActive().get(fileLockKey) <<== reads remote timeStampMillis at time A
                                        .thenAccept(val -> {
                      myTimeStamp = this.timeStampMillis;  <<== local timeStampMillis at time B
```
Is there an atomic `get(key,  lambdaToBeExecutedAtTheTimeOfTheGet)` operation?

jjezra: I might be missing something, but isn't it is safer to raise a semaphore before starting to wait on a worker thread via `asyncToSync`?

ScottDugas: There is not an atomic `get(key, lambdaToBeExecutedAtTheTimeOfTheGet)`

The problem is when this code is running in a `ForkJoinPool` with a parallelism of, e.g. 16. If you have 16 of these happening at the same time, you will have 16 threads inside the synchronized, waiting on a join. This has consumed all of the parallelism for the pool. So now, when the `get` goes to complete, there won't be any threads left on the `ForkJoinPool` to finish the `get`, this means that none of the `synchronized` blocks will ever complete, and the pool will be locked. 

This was the root cause behind #2234 

I think you can move the `synchronized` block after the `get`.

```java
        public void fileLockSet(boolean isHeartbeat) {
            agilityContext.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FILE_LOCK_SET,
                    agilityContext.apply(aContext -> fileLockSet(isHeartbeat, aContext)
                    ));
        }

        private CompletableFuture<Void> fileLockSet(boolean isHeartbeat, FDBRecordContext aContext) {
            final long nowMillis = System.currentTimeMillis();
            return aContext.ensureActive().get(fileLockKey)
                    .thenAccept(val -> {
                        synchronized (fileLockSetLock) {
                            long existingTimeStamp = fileLockValueToTimestamp(val);
                            if (isHeartbeat) {
                                fileLockCheckHeartBeat(existingTimeStamp);
                            } else {
                                fileLockCheckNewLock(existingTimeStamp, nowMillis);
                            }
                            this.timeStampMillis = nowMillis;
                            byte[] value = fileLockValue(timeStampMillis);
                            aContext.ensureActive().set(fileLockKey, value);
                        }
                    });
        }
```

The fact that you are reading, and setting in the same `FDBRecordContext` protects you from the possibility that another transaction changed the timestamp.
I think you do need some handling to protect against the possibility of:
1. get -> t0
2. other -> updates to t1
3. existingTimeStamp now equals this.timeStampMillis which messes with the heartbeat code.

One simple solution would be to have the file lock contain a timestamp and a uuid. If the uuid changes you've lost the lock. Then the code in `fileLockSet` would validate that the uuid hasn't changed, and _only_ set a new timestamp if `nowMillis` is greater than `timeStampMillis` and the one that it read. Although, realistically, if it is the same uuid, and it goes down, it shouldn't be a problem.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -107,40 +104,61 @@ private static long fileLockValueToTimestamp(byte[] value) {
...
+            // Here: this lock is either too old, or in the future. Steal it
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn(KeyValueLogMessage.of("FileLock: Set: found old lock, discard it",
+                        LuceneLogMessageKeys.LOCK_EXISTING_TIMESTAMP, existingTimeStamp,
+                        LuceneLogMessageKeys.LOCK_DIRECTORY, this));
+            }
+        }
+
+
+        private void fileLockClear() {
```
ScottDugas, line 161: I just realized this doesn't work in the case where there is a heartbeat. Namely, it's possible that it does a heartbeat, and fails to commit. If it tries to clear the lock in a separate transaction, it won't know what the timestamp should be, since it already updated `this.timeStampMillis`. Switching it to use a uuid would allow `fileLockClear` to just check the `uuid`, and not the timestamp.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
jjezra: Implemented: 
* Moved locks (heartbeat and clear) to be within the get. 
* Use UUID to identify self lock.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -95,50 +95,58 @@ public void ensureValid() {
         }
 
 
-        private static byte[] fileLockValue(long timeStampMillis) {
-            return Tuple.from(timeStampMillis).pack();
+        private byte[] fileLockValue() {
+            return Tuple.from(selfStampUuid, timeStampMillis).pack();
```
ScottDugas, line 99: If you flipped the two around, so that `timeStampMillis` is first, I think this would be backwards compatible. Probably not super important given the current state of the lucene index, but something to keep in mind for future changes.

ScottDugas (approved):

PR #2542, opened 2024-02-28
RESOLVES #2541: Allow Non-Unique Lucene Partitioning Field Values

### LINKED ISSUE #2541: Allow Non-Unique Lucene Partitioning Field Values

wshaib-apple:
- partitioning field no longer has to be unique
- partitioning field no longer has to be of type `Long/long`
- document primary key is implicitly added to the partitioning field value when creating a partitioning key
- fixed a bug that didn't correctly serialize then deserialize a `BytesRef` to/from protobuf
- updated test to run with non-unique partitioning field values


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -261,36 +249,32 @@ public <M extends Message> CompletableFuture<Integer> addToAndSavePartitionMetad
...
-     * @param groupKey  grouping key
-     * @param timestamp document timestamp
+     * @param partitioningKey partitioning key
      * @return assigned partition id
      */
     @Nonnull
-    private CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull final Tuple groupKey, @Nonnull final Long timestamp) {
-        return getOrCreatePartitionInfo(groupKey, timestamp).thenApply(assignedPartition -> {
+    private CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull final Tuple groupingKey, @Nonnull final Tuple partitioningKey) {
+        return getOrCreatePartitionInfo(groupingKey, partitioningKey).thenApply(assignedPartition -> {
```
ScottDugas, line 265: Just to confirm, everywhere you use `partitioningKey` it is both the value, and the primary key, correct?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -341,37 +324,37 @@ private CompletableFuture<Integer> removeFromAndSavePartitionMetadata(@Nonnull f
...
      * @param builder builder instance
      */
-    void savePartitionMetadata(@Nonnull Tuple groupKey, @Nonnull final LucenePartitionInfoProto.LucenePartitionInfo.Builder builder) {
+    void savePartitionMetadata(@Nonnull Tuple groupingKey,
+                               @Nonnull Tuple partitioningKey,
+                               @Nonnull final LucenePartitionInfoProto.LucenePartitionInfo.Builder builder) {
         LucenePartitionInfoProto.LucenePartitionInfo updatedPartition = builder.build();
         state.context.ensureActive().set(
-                partitionMetadataKeyFromTimestamp(groupKey, getFrom(builder)),
+                partitionMetadataKeyFromPartitioningValue(groupingKey, partitioningKey),
```
ScottDugas, line 350: Why is this method now taking a `partitioningKey`, rather than getting it from the `builder`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -325,26 +328,26 @@ private CompletableFuture<Void> handleCrossPartitionDiscontinuity() {
...
+            // next, if the current partition's new boundary excludes the continuation's "search after" partitioning key,
+            // we need to change the current partition to the one that does contain this partitioning key.
+            if (searchAfterPartitioningKey != null && (searchAfterPartitioningKey.compareTo(currentFrom) < 0 || searchAfterPartitioningKey.compareTo(currentTo) > 0)) {
 
                 // need to preserve the current "searchAfter" doc across the partition change
                 dontResetSearchAfter = true;
 
                 // get the proper partition
-                getProperContinuationPartition = partitioner.findPartitionInfo(groupingKey, searchAfterTimestamp - 1).thenCompose(properPartitionInfo -> {
+                getProperContinuationPartition = partitioner.findPartitionInfo(groupingKey, searchAfterPartitioningKey).thenCompose(properPartitionInfo -> {
```
ScottDugas, line 345: Why was it `-1` before, but now it is equality?
Should it be using `LucenePartitioner.getNextOlderPartitionInfo` ?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -325,26 +328,26 @@ private CompletableFuture<Void> handleCrossPartitionDiscontinuity() {
...
+            partitionKey = currentFrom;
 
             final CompletableFuture<Void> getProperContinuationPartition;
 
-            // next, if the current partition's new boundary excludes the continuation's "search after" timestamp,
-            // we need to change the current partition to the one that does contain this timestamp.
-            if (searchAfterTimestamp != null && (searchAfterTimestamp < currentFrom || searchAfterTimestamp > currentTo)) {
+            // next, if the current partition's new boundary excludes the continuation's "search after" partitioning key,
+            // we need to change the current partition to the one that does contain this partitioning key.
+            if (searchAfterPartitioningKey != null && (searchAfterPartitioningKey.compareTo(currentFrom) < 0 || searchAfterPartitioningKey.compareTo(currentTo) > 0)) {
```
ScottDugas, line 339: Can/should this use `LucenePartitioner.isOlderThan` or `isNewerThan`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -376,20 +379,25 @@ private CompletableFuture<RecordCursorResult<IndexEntry>> switchToNextPartitionA
...
         return nextPartitionFuture.thenCompose(nextPartition -> {
             if (nextPartition != null && nextPartition.getId() != partitionId) {
                 // reset scan params/state
                 exhausted = false;
-                this.partitionTimestamp = LucenePartitioner.getFrom(nextPartition);
+                this.partitionKey = partitioner.getPartitionKey(nextPartition);
                 this.partitionId = nextPartition.getId();
                 if (!dontResetSearchAfter) {
                     searchAfter = null;
+                    searchAfterPartitioningKey = null;
```
ScottDugas, line 400: I think this is correct, was it just a bug before, or is something semantic changing in the surrounding code that requires this clearing to be explicit here?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
```diff
@@ -83,14 +85,18 @@ public static LuceneCursorContinuation fromScoreDoc(ScoreDoc scoreDoc,
...
             builder.setPartitionId(partitionId);
         }
-        if (partitionTimestamp != null) {
-            builder.setPartitionTimestamp(partitionTimestamp);
+        if (partitionKey != null) {
+            builder.setPartitionKey(ByteString.copyFrom(partitionKey.pack()));
+        }
+        if (searchAfterPrimaryKey != null) {
+            builder.setSearchAfterPrimaryKey(ByteString.copyFrom(searchAfterPrimaryKey.pack()));
         }
```
ScottDugas, line 93: I think you want to get this from the `scoreDoc`. It should always be the second field in the ordering.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -461,14 +474,28 @@ private void maybePerformScan() throws IOException {
...
+                        if (nextResult != null && nextResult.get() != null) {
+                            primaryKey = nextResult.get().getPrimaryKey();
+                        } else {
+                            primaryKey = null;
+                        }
+                        RecordCursorContinuation continuationFromDoc = LuceneCursorContinuation.fromScoreDoc(
+                                searchAfter,
+                                partitionId,
+                                partitionKey,
+                                primaryKey);
```
ScottDugas, line 487: This seems incorrect, we should be getting the primaryKey from the `searchAfter`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -461,14 +474,28 @@ private void maybePerformScan() throws IOException {
...
                         final Optional<NoNextReason> stoppedReason = limitManager.getStoppedReason();
                         if (stoppedReason.isEmpty()) {
                             throw new RecordCoreException("limit manager stopped LuceneRecordCursor but did not report a reason");
                         } else {
-                            nextResult = RecordCursorResult.withoutNextValue(LuceneCursorContinuation.fromScoreDoc(searchAfter, partitionId, partitionTimestamp), stoppedReason.get());
+                            nextResult = RecordCursorResult.withoutNextValue(LuceneCursorContinuation.fromScoreDoc(
+                                    searchAfter,
+                                    partitionId,
+                                    partitionKey,
+                                    null), stoppedReason.get());
```
ScottDugas, line 498: `null` seems incorrect here, we should get the primary key from the `searchAfter`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -174,8 +174,8 @@ private Boolean mergePartition(@Nonnull final LuceneAnalyzerWrapper analyzerWrap
 
     private CompletableFuture<Integer> getNextPartitionInfo(final Tuple groupingKey, final AgilityContext agileContext,
```
ScottDugas, line 175: Probably worth renaming this to `getNextOlderPartitionInfo` too. If that's a lot of lines changed though, better as a followup.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1190,18 +1189,20 @@ private enum SortType {
     static Stream<Arguments> continuationDuringRepartitioningTest() {
         return Stream.of(true, false)
                 .flatMap(grouped -> Stream.of(true, false)
-                        .flatMap(synthetic -> Arrays.stream(SortType.values())
-                                .map(sortType -> Arguments.of(grouped, synthetic, sortType))));
+                        .flatMap(uniqueTimestamps -> Stream.of(true, false)
+                                .flatMap(synthetic -> Arrays.stream(SortType.values())
+                                        .map(sortType -> Arguments.of(grouped, synthetic, uniqueTimestamps, sortType)))));
```
ScottDugas, line 1194: We should remove the code in https://github.com/FoundationDB/fdb-record-layer/blob/86e1659b20609765ac3394eb49699788845b9ed8/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java#L148-L151 that protects against duplicate timestamps.
This will probably require updating the sorting logic in https://github.com/FoundationDB/fdb-record-layer/blob/86e1659b20609765ac3394eb49699788845b9ed8/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java#L97 so that it expects the right thing when there are duplicate timestamps.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1257,8 +1258,16 @@ void continuationDuringRepartitioningTest(boolean isGrouped,
...
-        Sort sort = sortType == SortType.UNSORTED ? null :
-                    new Sort(new SortField(isSynthetic ? "complex_timestamp" : "timestamp", SortField.Type.LONG, sortType == SortType.DESCENDING));
+        final Sort sort;
+        if (sortType == SortType.UNSORTED) {
+            sort = null;
+        } else if (uniqueTimestamps) {
+            sort = new Sort(new SortField(isSynthetic ? "complex_timestamp" : "timestamp", SortField.Type.LONG, sortType == SortType.DESCENDING));
+        } else {
+            sort = new Sort(new SortField(isSynthetic ? "complex_timestamp" : "timestamp", SortField.Type.LONG, sortType == SortType.DESCENDING),
+                    new SortField(LuceneIndexMaintainer.PRIMARY_KEY_SEARCH_NAME, SortField.Type.STRING, sortType == SortType.DESCENDING));
```
ScottDugas, line 1268: I think you mentioned this offline, but can we make this implicit?
If we can, we should probably test the possibility that the user also specifies the primary key in the search.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1226,7 +1227,7 @@ void continuationDuringRepartitioningTest(boolean isGrouped,
                         .setDocId(1000L + i)
                         .setIsSeen(true)
                         .setText("A word about what I want to say")
-                        .setTimestamp(start + i * 100)
+                        .setTimestamp(uniqueTimestamps ? start + i * 100 : start)
```
ScottDugas, line 1230: We should also test with a few (e.g. 3) distinct timestamps.
And we should make the timestamps have a different ordering from primary key.
Perhaps also not having the ordering of timestamps align with the insert order.

This might start to warrant adding some randomized arguments to this test, if that's easy enough, so that we don't run all permutations every time. But I don't think that's worth it right now if it means spending a bunch of time trying to get the validation to handle all the possibilities.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneCursorContinuation.java
wshaib-apple: right

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: Correct

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: yes... I wrote this before adding those functions. Will update.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: `findPartitionInfo` implementation changed (it does range scan on the `partitionKey` instead of timestamp, inclusive)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Mixing the order will require a bit more sophisticated tracking of expectations. I'll probably defer this to a following update.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: the implicit logic would then be:
1. if `Sort` has `partitionFieldName` first and doesn't have `_s` second, insert `_s` into the sort fields immediate after the first (even if it means displacing potentially other fields to the right)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: I think the previous version should have also reset `searchAfterTimestamp`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: We won't need to pass it explicitly (will grab it from the `ScoreDoc`)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: We won't need to pass it explicitly (will grab it from the `ScoreDoc`)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -263,6 +271,19 @@ public class LuceneRecordCursor implements BaseCursor<IndexEntry> {
         closed = false;
     }
 
+    private void ensurePrimaryKeyIsInSort() {
+        SortField[] fields = sort.getSort();
+        if (fields.length < 2 || !LuceneIndexMaintainer.PRIMARY_KEY_SEARCH_NAME.equals(fields[1].getField())) {
+            SortField[] updatedFields = new SortField[fields.length + 1];
+            updatedFields[0] = fields[0];
+            updatedFields[1] = new SortField(LuceneIndexMaintainer.PRIMARY_KEY_SEARCH_NAME, SortField.Type.STRING, fields[0].getReverse());
```
MMcM, line 279: Why second and not last? Doesn't that break the contract of the sort, which so far has only been checked to have the partition key first. If the intention is just to break ties, it doesn't much matter where, I don't think.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
ScottDugas: I'm not sure I see how `findPartitonInfo` changed in a relevant way.
It used to add the `timestamp` to the metadata space, and now it adds the `timestamp,primaryKey`, but the scan itself is the same.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
ScottDugas: Could that cause it to flip-flop between partitions, if the second sort field is not the primary key? Could it cause data to be invisible?

Thinking about what it would look like in the pathological case, of having two partitions, entirely with the same partitioning field. I think some care is needed to make sure that you don't miss data.
Otherwise if you have
| primary key | value |
| -- | -- |
| a | 6 |
| b | 5 |
| c | 4 |
| d | 3 |

And that is in two partitions:

| primary key | value |
| -- | -- |
| c | 4 |
| d | 3 |

| primary key | value |
| -- | -- |
| a | 6 |
| b | 5 |

And you are searching descending by `value`. If you jump to the second partition, after continuing from the first, you need to make sure to clear out the `value` part of the continuation, but keep the partitioning field, otherwise it would end up skipping all of the second partition's value.

If it works as the last sort field, I think it would work in any position after the partitioning field.

This seems a bit tricky / complicated to get right; perhaps we could restrict in this version so that if the only sort provided is the partitioning field, we'll add the primary key, otherwise we will treat it as we would if it was not sorted by the partitioning field at all? Thoughts @MMcM

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
wshaib-apple: Yes it does break the contract of the sort. However the intent was to convert the partitioning key in the sort to the actual **cross partition** unique key for a deterministic/correct list of matches, not for breaking ties.
Adding the primary key anywhere other than second in this case will not guarantee properly sorted results across partitions. Alternatively, (and perhaps the correct behavior), we can implicitly add the primary key **only** if the sort has the partition field alone.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -272,14 +272,13 @@ public class LuceneRecordCursor implements BaseCursor<IndexEntry> {
     }
 
     private void ensurePrimaryKeyIsInSort() {
+        // precondition: sort is by partition key (see LucenePartitioner.isSortedByPartitionField())
+        // so, either partition field + primary key (explicitly) or just partition field.
```
ScottDugas, line 276: The fact that this method is so tightly coupled with `LucenePartitioner.isSortedByPartitionField()` isn't great, and it would be better if it was all contained in one place. Could `isSortedByPartitionField()` also do this logic and return a simple class with the two booleans in the `Pair`, and the updated `sort`

ScottDugas (approved):

ScottDugas (dismissed):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
MMcM: That makes sense. So perhaps it's best only thinking that it's sorted by partitioning key if it's that alone, plus an optional second primary key, which we otherwise add.

ScottDugas (dismissed):

ScottDugas (approved):

MMcM (approved):

PR #2544, opened 2024-02-29
Fixes #2543: Permuted min/max indexes on keys producing duplicates only index the first entry

### LINKED ISSUE #2543: Permuted min/max indexes on keys producing duplicates only index the first entry
There's a bug in the permuted min/max index maintainer where if you had an index expression with a `FanOut` within it, we're only indexing the first entry in the list. The problem is that the maintainer has logic like:

```java
protected CompletableFuture<Void> updateEntriesAsync(List<IndexEntry> entries) {
    for (IndexEntry entry : entries) {
         return updateOneKeyAsync(entry);
    }
    return AsyncUtil.DONE;
}
```

Because there's a `return` there, we always short-circuit after the first entry. We need to modify this to something like:

```java
protected CompletableFuture<Void> updateEntriesAsync(List<IndexEntry> entries) {
    final List<CompletableFuture<Void>> work = new ArrayList();
    for (IndexEntry entry : entries) {
         work.add(updateOneKeyAsync(entry));
    }
    return AsyncUtil.whenAll(work);
}
```

This will spawn futures for each entry and then wait on them all. One caveat here is that we need to be a little mindful of the fact that we can have multiple elements for the _same_ group, in which case some care has to be made to make sure that we only update the group with the most extreme value.

alecgrieser:
The core issue was that there was an early-out in the `PermutedMinMaxIndexMaintainer`'s implementation of `updateIndexKeys`. This updates the loop so that it no longer early outs, and thus all entries will get included in the index. Some care has been taken so that if the same record produces multiple entries in the same group (which can happen if the repeated field is in the grouping key and the same value appears multiple time in the repeated field OR if the repeated field is in the aggregate value), then only the most extreme value for the record is used to determine what new entries to add to the permuted space. All values still get added to the primary index space, though, so that part of the index should behave exactly like a normal value index on the whole key (i.e., ignoring the `GroupingKeyExpression`).

This fixes #2543.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
@@ -32,6 +32,7 @@ import com.apple.foundationdb.record.metadata.IndexAggregateFunction;
 import com.apple.foundationdb.record.metadata.IndexOptions;
 import com.apple.foundationdb.record.metadata.IndexTypes;
 import com.apple.foundationdb.record.metadata.Key;
+import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreTestBase;
@@ -53,6 +54,8 @@ import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
```diff
@@ -201,6 +210,443 @@ public void tie() {
...
+
+        try (FDBRecordContext context = openContext()) {
+            openSimpleRecordStore(context, hook);
+
+            saveRecord(1, "yes", 1, 0, 1, 2, 3, 4, 5);
+            saveRecord(2, "no", 1, 0, 1, 3, 5);
+
+            assertEquals(Collections.singletonList(
+                    Tuple.from(min ? 1 : 5, 1)
+            ), scanGroup(Tuple.from("yes"), false));
```
alecgrieser, line 488: This assertion fails if the fix is made to avoid an early return, but no effort is made to de-dupe entries across groups for the same record. In that case, what can happen is that we can get multiple values in the index corresponding to the same record. Which ones we get precisely depends somewhat on timing/concurrency problems.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/PermutedMinMaxIndexTest.java
```diff
@@ -201,6 +210,443 @@ public void tie() {
...
+                    Tuple.from(min ? 1 : 5, 1)
+            ), scanGroup(Tuple.from("no"), false));
+
+            assertEquals(Arrays.asList(
+                    Tuple.from(1, 1),
+                    Tuple.from(2, 1),
+                    Tuple.from(3, 1),
+                    Tuple.from(4, 1),
+                    Tuple.from(5, 1)
+            ), scanValue(Tuple.from("yes", 1), false));
```
alecgrieser, line 500: There's a version of this that I almost wrote where the ordinary tree index only gets one entry per group per record. Essentially, it would take what is in their name, but instead of calling `super.updateIndexKeys` with _all_ of the entries, it would instead call `updateOneKeyAsync` within a loop only with the most extreme entry per group. I added this check to assert that we're not doing that, because this would violate the way the planner currently treats the ordinary tree index (i.e., it expects it to behave just like a value index on the whole key)

MMcM (approved):

PR #2547, opened 2024-03-01
Support Bytes type for operands of RelOpValue (EQ, NEQ, NULL, NOT_NULL)

arnaud-lacurie:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/LiteralValue.java
```diff
@@ -107,7 +107,10 @@ public boolean equalsWithoutChildren(@Nonnull final Value other, @Nonnull final
         }
 
         final LiteralValue<?> that = (LiteralValue<?>)other;
-        return Objects.equals(value, that.value);
+        if (value == null && that.value == null) {
```
normen662, line 110: Nice! Trap avoided!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -734,7 +743,10 @@ private enum UnaryPhysicalOperator {
         IS_NOT_NULL_FI(Comparisons.Type.NOT_NULL, Type.TypeCode.FLOAT, Objects::nonNull),
         IS_NOT_NULL_DI(Comparisons.Type.NOT_NULL, Type.TypeCode.DOUBLE, Objects::nonNull),
         IS_NOT_NULL_SS(Comparisons.Type.NOT_NULL, Type.TypeCode.STRING, Objects::nonNull),
-        IS_NOT_NULL_BI(Comparisons.Type.NOT_NULL, Type.TypeCode.BOOLEAN, Objects::nonNull);
+        IS_NOT_NULL_BI(Comparisons.Type.NOT_NULL, Type.TypeCode.BOOLEAN, Objects::nonNull),
```
normen662, line 746: Not a comment to you but I have not clue what the last prefix represents shouldn't this always be just representing the input type like in your case? I mean what does the `I` in `IS_NOT_NULL_FI` stand for?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -648,7 +649,15 @@ private enum BinaryPhysicalOperator {
         GTE_US(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.STRING, (l, r) -> null),
         GTE_UV(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.VERSION, (l, r) -> null),
         GTE_VU(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.VERSION, Type.TypeCode.UNKNOWN, (l, r) -> null),
-        GTE_VV(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.VERSION, Type.TypeCode.VERSION, (l, r) -> ((FDBRecordVersion)l).compareTo((FDBRecordVersion) r) >= 0),;
+        GTE_VV(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.VERSION, Type.TypeCode.VERSION, (l, r) -> ((FDBRecordVersion)l).compareTo((FDBRecordVersion) r) >= 0),
```
normen662, line 652: So we explicitly disallow `ByteString` bytes here?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
arnaud-lacurie: I have no clue either. I also don't know why all the other ones have `I` as a suffix, but strings have `S`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
arnaud-lacurie: I guess so, although it's more that tests were passing well with this, so I would not introduce something else unless we think it's necessary.

normen662 (approved):

PR #2548, opened 2024-03-01
Make advanced constructor private in LiteralValue

arnaud-lacurie:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/LiteralValue.java
```diff
@@ -148,7 +147,7 @@ public RecordQueryPlanProto.PValue toValueProto(@Nonnull final PlanSerialization
     @Nonnull
     public static <T> LiteralValue<T> fromProto(@Nonnull final PlanSerializationContext serializationContext,
                                                 @Nonnull final PLiteralValue literalValueProto) {
-        return new LiteralValue<>(Type.fromTypeProto(serializationContext, literalValueProto.getResultType()),
+        return new LiteralValue<>(
```
normen662, line 150: This one must call the other constructor to make sure the result type cannot be resolved in any other fashion.

normen662 (changes requested):

PR #2550, opened 2024-03-05
remove atomic boolean from translateCorrelations logic

normen662:


hatyo (approved):
I like this version much better, thank you!

PR #2553, opened 2024-03-05
Fixes #2552: Nested fields can preserve field protobuf indexes incorrectly in index expansion

### LINKED ISSUE #2552: Nested fields can preserve field ordinals incorrectly in index expansion
When creating the matching candidates for nested expressions, the expansion visitor can end up assigning ordinal numbers to nested fields that then get carried up into combined expressions. This can lead to protobuf validation errors if those ordinal numbers are re-used elsewhere.

For example, if you had an index on:

```java
concat(field("foo"), field("bar", FanType.FanOut).nest(concatenateFields("baz", "quox"))
```

Then this gets translated into a graph where there's an explode expression on the "bar" field with the "baz" and "quox" fields being selected out. Those two fields get assigned ordinals 1 and 2. But then that gets combined into a larger select which assigns "foo" ordinal 1. The second two fields should get new ordinals, but they're currently keeping their old ordinal values, which means the result type of the select fails protobuf validation.

alecgrieser:
This updates the way that graph expansions get combined during index match candidate generation so that field protobuf indexes are not preserved. This updates the `normalizeFields` logic in `Type.Record` so that if any of the field indexes are unset or contain dupes, then we always set the field indexes (whereas before, we would preserve those indexes if they were there, even if we'd identified duplicate field indexes in the type).

This also introduces a new helper method to make it easier to write `Column`s without needing to repeat the type and uses that new abstraction in some existing code paths.

This fixes #2552.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Column.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Column.java
@@ -29,6 +29,7 @@ import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 
 import javax.annotation.Nonnull;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * A class to hold a value computing a column together with information about the field of a record it supplies
@@ -75,6 +76,10 @@ public class Column<V extends Value> implements PlanHashable, PlanSerializable {
 
     public static <V extends Value> Column<V> unnamedOf(@Nonnull final V value) {
         return new Column<>(Field.unnamedOf(value.getResultType()), value);
+    }
+
+    public static <V extends Value> Column<V> of(@Nonnull Optional<String> fieldNameOptional, @Nonnull final V value) {
+        return new Column<>(Field.of(value.getResultType(), fieldNameOptional), value);
     }
 
     public static <V extends Value> Column<V> of(@Nonnull final Field field, @Nonnull final V value) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
@@ -43,7 +43,6 @@ import javax.annotation.Nonnull;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.List;
-import java.util.Optional;
 
 /**
  * Expansion visitor that implements the shared logic between primary scan data access and value index access.
@@ -128,7 +127,7 @@ public class KeyExpressionExpansionVisitor implements KeyExpressionVisitor<Visit
                 // explode this field and prefixes of this field
                 final Quantifier.ForEach childBase = fieldKeyExpression.explodeField(baseQuantifier, fieldNamePrefix);
                 value = state.registerValue(childBase.getFlowedObjectValue());
-                column = Column.of(Optional.of(fieldName), value);
+                column = Column.unnamedOf(value);
                 final GraphExpansion childExpansion;
                 if (state.isKey()) {
                     childExpansion = GraphExpansion.ofResultColumnAndPlaceholder(column, value.asPlaceholder(newParameterAlias()));
@@ -146,7 +145,7 @@ public class KeyExpressionExpansionVisitor implements KeyExpressionVisitor<Visit
                         .builderWithInheritedPlaceholders().pullUpQuantifier(childQuantifier).build();
             case None:
                 value = state.registerValue(FieldValue.ofFieldNames(baseQuantifier.getFlowedObjectValue(), fieldNames));
-                column = Column.of(Optional.of(fieldName), value);
+                column = Column.unnamedOf(value);
                 if (state.isKey()) {
                     return GraphExpansion.ofResultColumnAndPlaceholder(column, value.asPlaceholder(newParameterAlias()));
                 }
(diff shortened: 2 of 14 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
```diff
@@ -121,16 +122,18 @@ public GraphExpansion visitExpression(@Nonnull FieldKeyExpression fieldKeyExpres
                 .add(fieldName)
                 .build();
         final Value value;
+        final Column<?> column;
```
alecgrieser, line 125: This updates the `Column` from an unnamed column to one that preserves the field name. This mainly makes it easier to track where the field comes from, though it's not strictly necessary for this PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -2130,6 +2130,7 @@ private static List<Field> normalizeFields(@Nullable final List<Field> fields) {
                 }
                 if (field.fieldIndexOptional.isPresent() && !fieldIndexesSeen.add(field.getFieldIndex())) {
                     override = true;
+                    break;
```
alecgrieser, line 2133: This is a small optimization so that once we find a duplicated field index, we stop looking. The `fieldIndexesSeen` set is not referenced after this loop.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Column.java
```diff
@@ -77,6 +78,10 @@ public static <V extends Value> Column<V> unnamedOf(@Nonnull final V value) {
         return new Column<>(Field.unnamedOf(value.getResultType()), value);
     }
 
+    public static <V extends Value> Column<V> of(@Nonnull Optional<String> fieldNameOptional, @Nonnull final V value) {
+        return new Column<>(Field.of(value.getResultType(), fieldNameOptional), value);
+    }
```
alecgrieser, line 83: This new helper method allows us clean up code, because there were a fair number of cases where we took the result type from the `value`, and then wanted to attach a field name. By line count, most of the changes in this PR are mechanical updates to use this new helper method where appropriate

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -2154,7 +2155,7 @@ private static List<Field> normalizeFields(@Nullable final List<Field> fields) {
                     fieldToBeAdded =
                             new Field(field.getFieldType(),
                                     Optional.of(explicitFieldName),
-                                    Optional.of(field.getFieldIndexOptional().orElse(i + 1)));
+                                    Optional.of(i + 1));
```
alecgrieser, line 2158: This is the only substantive code change. If we're in this part of the loop (as one of the comments states), the field names and protobuf indexes have been identified as fair game to change. We could update this to make a more concerted effort to change this index numbers more (e.g., by giving it a new index if it is one of the duplicate values), but it shouldn't matter

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBNestedRepeatedQueryTest.java
```diff
@@ -1511,6 +1522,93 @@ void sumWholeRecordValueWithDuplicatesByKey() {
         }, this::querySumIntValueForRecordByKey);
     }
 
+    @DualPlannerTest(planner = DualPlannerTest.Planner.CASCADES)
+    void sumByKeyAndOther() {
```
alecgrieser, line 1526: This test used to throw a protobuf validation error before this change. The index matchcandidate for the `sumValueByKeyAndOther` would create a result value that had two fields with protobuf index 1, hence the error.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
hatyo: That makes sense to me.

hatyo (changes requested):
Great work, and nice refactoring, the PR looks more or less good to me, except giving match candidates columns, if you could just revert that part which is orthogonal to your PR then we can bring it in.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
alecgrieser: Yeah, I can do that as a precaution

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Column.java
alecgrieser: The benefit of using an `Optional` here is that in a case where you're starting with an `Optional` field name, you can call this method instead of doing something like:

```java
Column<V> column = maybeFieldName
    .map(fieldName -> Column.of(fieldName, value))
    .orElseGet(() -> Column.unnamedOf(value))
```

Instead, you can do:

```java
Column<V> column = Column.of(maybeFieldName, value);
```

I think there was a version of this PR where I actually made use of this to do something like `Column.of(field.getFieldNameOptional(), value)` when copying over the value from one field into another. That has since been removed, so it's a little less obvious.

Note that we end up wrapping the field name in an optional when creating the `Field` anyway, so it's not like we can avoid an object creation by having a string version.

I could see the benefit to _also_ having a variation that takes a `String`, to avoid needing quite as many `Optional.of`s at call sites.

hatyo (approved):

PR #2561, opened 2024-03-06
Resolve #2559: IndexDeferredMaintenanceControl: Change the default to defer index operations

### LINKED ISSUE #2559: IndexDeferredMaintenanceControl: Change the default to defer index operations
It seems that merge might be called unintentionally during other operation. To prevent that, require explicit call to allow deferred operations before executing them. So the default will be to defer.

jjezra:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexDeferredMaintenanceControl.java
```diff
@@ -31,7 +31,7 @@
  */
 public class IndexDeferredMaintenanceControl {
     private Set<Index> mergeRequiredIndexes = null;
-    private boolean autoMergeDuringCommit = true;
+    private boolean autoMergeDuringCommit = false;
```
ScottDugas, line 34: We should followup to improve the ux, and ensure that users can't accidentally never do merges.

ScottDugas (approved):

ScottDugas (approved):

ScottDugas (approved):

PR #2562, opened 2024-03-07
Add metric for Lucene number of file refs and their size

### LINKED ISSUE #2566: Add metric for the size of list of files loaded by loadFileReferenceCacheForMemoization

g31pranjal:
Resolves #2566


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -553,6 +553,7 @@ private CompletableFuture<Void> loadFileReferenceCacheForMemoization() {
             list.forEach(kv -> {
                 String name = metaSubspace.unpack(kv.getKey()).getString(0);
                 final FDBLuceneFileReference fileReference = Objects.requireNonNull(FDBLuceneFileReference.parseFromBytes(LuceneSerializer.decode(kv.getValue())));
+                agilityContext.recordSize(LuceneEvents.SizeEvents.LUCENE_FILES_COUNT, kv.getValue().length);
```
ScottDugas, line 556: This doesn't look right, this would record the size of each of the file references, which is not what we want, we want to record the size of `list`

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
g31pranjal: changed that!

ScottDugas (approved):

PR #2571, opened 2024-03-08
Resolve #2570: Lucene: avoid concurrent merges

### LINKED ISSUE #2570: Lucene: avoid concurrent merges
Switch from ConcurrentMergeScheduler to SerialMergeScheduler

jjezra:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -111,7 +111,7 @@ public DirectoryReader getWriterReader(boolean flush) throws IOException {
         return writerReader;
     }
 
-    private static class FDBDirectoryMergeScheduler extends ConcurrentMergeScheduler {
+    private static class FDBDirectoryMergeScheduler extends SerialMergeScheduler {
```
ScottDugas, line 114: `SerialMergeScheduler` has a `synchronized` block around the `merge` method, to prevent multiple threads from doing a merge at the same time.
Do we really want that? Would it make more sense for multiple requests to do a `merge` to be able to execute the merge at the same time?
`ConcurrentMergeScheduler` also has a `synchronized` on its `merge` method, so we may need to synchronize around getting the next merge.

ScottDugas, line 114: Can we make this controlled either by a `RecordContextProperty` or index option?

ScottDugas, line 114: We should add a test that confirms the goals of switching to `SerialMergeScheduler`, and probably document why we want to switch.
Looking at logs/comments I could not figure out why we use the `ConcurrentMergeScheduler` to begin with, it appears to have part of the initial implementation commit ac71c207
Primarily, my understanding is that if you use `ConcurrentMergeScheduler` it will continue to do merges in the background, even after we have moved on. This may only happen in error scenarios. I came across it in https://github.com/FoundationDB/fdb-record-layer/pull/2569 but there is probably a more targeted test that we could write (that test is actually trying to catch a totall different bug).

ScottDugas: I can think of two potential ways to test this in a more targeted way:
1. Utilize the `FDBStoreTimer` and verify that after `explicitMerge` completes the timer doesn't change for some time
2. Fetch `Thread.getCurrenthread().getThreadGroup()` and validate that it is empty after `explicitMerge` completes. You may have to filter out threads like in `FDBThreadFilter`, but I suspect you could get a test that fails with the `ConcurrentMergeScheduler`. 

I think in order to get a useful test, you would need to have some more interesting merges; I think having a loop that saves 2 records/transaction with deferred merge, and then doing a merge with `LuceneRecordContextProperties.LUCENE_MERGE_SEGMENTS_PER_TIER` set to something low (like 3.0) should get enough merges that the `ConcurrentMergeScheduler` ends up having a lot of work.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
jjezra: ✅

jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -111,6 +112,81 @@ public DirectoryReader getWriterReader(boolean flush) throws IOException {
...
+                "squid:S2245", // ThreadLocalRandom not used in case where cryptographic security is needed
+                "squid:S3776", // Cognitive complexity is too high. Candidate for later refactoring
+        })
+        @Override
+        public synchronized void merge(final MergeSource mergeSource, final MergeTrigger trigger) throws IOException {
+            long startTime = System.nanoTime();
+            if (state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_MULTIPLE_MERGE_OPTIMIZATION_ENABLED) && trigger == MergeTrigger.FULL_FLUSH) {
+                if (ThreadLocalRandom.current().nextInt(mergeDirectoryCount) == 0) {
+                    if (mergeSource.hasPendingMerges()) {
+                        MergeUtils.logExecutingMerge(LOGGER, "Executing Merge based on probability", agilityContext, state.indexSubspace, key, trigger);
```
ScottDugas, line 147: Probably worth differentiating the logs here, from the FDBDirectoryMergeScheduler.
That might be easily done by moving this out so that it is not a nested class, and giving it its own logger.
It might also make sense to change the log of the non-default, but I don't think it's a big deal either way, and suggested edits are easy to apply.
```suggestion
                        MergeUtils.logExecutingMerge(LOGGER, "Executing Merge serially based on probability", agilityContext, state.indexSubspace, key, trigger);
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -111,6 +112,81 @@ public DirectoryReader getWriterReader(boolean flush) throws IOException {
...
+                    }
+                    serialMerge(mergeSource, trigger);
+                } else {
+                    skipMerge(mergeSource);
+                }
+            } else {
+                if (mergeSource.hasPendingMerges()) {
+                    MergeUtils.logExecutingMerge(LOGGER, "Executing Merge", agilityContext, state.indexSubspace, key, trigger);
```
ScottDugas, line 155:
```suggestion
                    MergeUtils.logExecutingMerge(LOGGER, "Executing Merge Serially", agilityContext, state.indexSubspace, key, trigger);
```

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
jjezra: I strongly agree, but made a slight variation - changing the messages of the twin class to "Executing Merge Concurrently" (as that is the less obvious one). I hope that it's OK with you.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
jjezra: ✅

ScottDugas (approved):

PR #2579, opened 2024-03-13
Fixes #2573: Delete PrimaryKeySegmentIndex entries when files are deleted

### LINKED ISSUE #2573: PrimaryKeySegmentIndex can be corrupted with AgilityContext
When run with an agile context, PrimaryKeySegmentIndex may delete entries for segments that don't end up getting merged, forcing future updates to use a query to do updates, defeating the purpose of the PrimaryKeySegmentIndex.
This is most likely to happen if there is some sort of error, such as a timeout in a transaction after the PrimaryKeySegmentIndex gets merged.

ScottDugas:
Since I new this would involve a lot of extracting of interfaces, and moving code I thought it might be easier to go commit-by-commit, and have some pretty simple commits for that kind of stuff.
Some of the merges a little later on didn't go super smoothly, so it might be better to diff the 3 versions of:
`LucenePrimaryKeySegmentIndex` (`v1`, `v2` and the interface).

I was unsure whether to take Lucene's approach, and copy things to avoid breaking things, or extract base classes. I made two different decisions here. Looking back it may make sense to have `LucenePrimaryKeySegmentIndexV1` extend from `LucenePrimaryKeySegmentIndexV2` based on implementations (perhaps better names would make that make more sense). However, given the corruption, and the current state of lucene, I expect V1 to be removed shortly, so it probably doesn't matter.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndexTest.java
```diff
@@ -0,0 +1,210 @@
...
+import static com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.createSimpleDocument;
+import static com.apple.foundationdb.record.provider.foundationdb.indexes.TextIndexTestUtils.SIMPLE_DOC;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Lower level test of {@link LucenePrimaryKeySegmentIndex}.
+ */
+public class LucenePrimaryKeySegmentIndexTest extends FDBRecordStoreTestBase {
```
ohadzeliger, line 56: Maybe more tests:
- No documents (empty segment)
- Documents without stored fields
- Segment with all documents deleted

ohadzeliger: - Document modified (e.g. moved to a different segment but keeps PK)?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -518,6 +521,20 @@ public byte[] readStoredFields(String segmentName, int docId) throws IOException
         return rawBytes;
     }
 
+    @Nonnull
+    public List<KeyValue> readAllStoredFields(String segmentName) {
+        final Range range = storedFieldsSubspace.range(Tuple.from(segmentName));
+        final List<KeyValue> list = asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_GET_STORED_FIELDS,
+                agilityContext.getRange(range.begin, range.end));
+        if (list == null) {
```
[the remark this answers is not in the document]
ScottDugas: I think the contract (although not explicitly called out) is that `agilityContext.getRange` will return a `CompletableFuture` of a non-null list, but there is no `NonnullCompletableFuture`, so it gives a warning.
So, I think this should _never_ happen.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -700,6 +700,11 @@ private boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference>
         return true;
     }
 
+    public boolean usesOptimizedStoredFields() {
+        return getBooleanIndexOption(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED, false)
```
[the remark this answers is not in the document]
ScottDugas: Yes, I think we want to continue experimenting with the performance of these things, and then remove everything that didn't pan out. Maintaining dozens of unused options will be a maintenance pain going forward.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndexTest.java
ScottDugas: I think:
- "Documents without stored fields" should not be possible because we always include the primary key

The others should all be covered by `LuceneStoredFieldsTest`, although some of those may need to add more calls to the primary key validation.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -65,36 +58,35 @@
...
-
     @Override
     protected RecordLayerPropertyStorage.Builder addDefaultProps(final RecordLayerPropertyStorage.Builder props) {
         return super.addDefaultProps(props)
                 .addProp(LuceneRecordContextProperties.LUCENE_INDEX_COMPRESSION_ENABLED, true);
     }
 
+    enum StoredFieldsType {
+        File(options -> {
+            options.remove(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED);
```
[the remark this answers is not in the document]
ScottDugas: Added a `Default` option, and changed all of these to explicitly set the options to `false`.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -109,7 +101,7 @@ void testInsertDocuments(boolean useOptimizedStoredFieldsFormat) throws Exceptio
             queryAndAssertFields(query, "text", Map.of(
                     1623L, "Document 1",
                     1624L, "Document 2"));
-            if (useOptimizedStoredFieldsFormat) {
+            if (type != StoredFieldsType.File) {
```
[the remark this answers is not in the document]
ScottDugas: Yes, but went with `usesOptimizedStoredFields`. This was necessary for adding a `Default` `StoredFieldsType`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -379,18 +419,36 @@ private Tuple toPrimaryKey(final FDBQueriedRecord<Message> record) {
...
+            assertEquals(expectedDocsPerSegment.get(i), keyValues.size(), expectedSegmentNames.get(i));
         }
     }
 
     private void rebuildIndexMetaData(final FDBRecordContext context, final String document, final Index index) {
         Pair<FDBRecordStore, QueryPlanner> pair = LuceneIndexTestUtils.rebuildIndexMetaData(context, path, document, index, useCascadesPlanner);
         this.recordStore = pair.getLeft();
         this.planner = pair.getRight();
-        recordStore.getIndexDeferredMaintenanceControl().setAutoMergeDuringCommit(true);
+        recordStore.getIndexDeferredMaintenanceControl().setAutoMergeDuringCommit(false);
```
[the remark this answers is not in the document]
ScottDugas: Yes, this is to align with our intended usage.
It also makes it more predictable. There are tests that explicitly call merge, but the merge was sometimes happening during commit.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneStoredFieldsTest.java
```diff
@@ -331,7 +333,7 @@ void testDeleteAllDocuments(StoredFieldsType type) throws Exception {
             LuceneIndexTestUtils.mergeSegments(recordStore, index);
         }
 
-        if (type != StoredFieldsType.File) {
+        if (type.usesOptimizedStoredFields) {
             try (FDBRecordContext context = openContext(contextProps)) {
                 rebuildIndexMetaData(context, SIMPLE_DOC, index);
                 try (FDBDirectory directory = new FDBDirectory(recordStore.indexSubspace(index), context, index.getOptions())) {
```
ohadzeliger, line 339: Could have used the new method to assert no documents in any segment

ohadzeliger (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndexV2.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+/**
+ * Maintain a B-tree index of primary key to segment and doc id.
+ * This allows for efficient deleting of a document given that key, such as when doing index maintenance from an update.
+ */
+public class LucenePrimaryKeySegmentIndexV2 implements LucenePrimaryKeySegmentIndex {
```
ScottDugas, line 54: I spent some time debating, whether to follow lucene`s example and copy the classes so as not to disturb the previous version, or extract a base class for shared logic.
I went with copying here, and extending for the stored fields. I'm still not sure how I feel about it, but I expect that `LucenePrimaryKeySegmentIndexV1` will be removed shortly, given the corruption issues that the tests find.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -343,6 +343,8 @@ public CompletableFuture<Void> mergeIndex() {
                 .thenCompose(ignored -> {
                     IndexDeferredMaintenanceControl maintenanceControl = state.store.getIndexDeferredMaintenanceControl();
                     maintenanceControl.setLastStep(IndexDeferredMaintenanceControl.LastStep.MERGE);
+                    // This agilityContext will be used to determine/iterate grouping keys and partitions. The time gap between calls might
```
ScottDugas: I thought I did this correctly, but I later discovered I did not, so I checked out the main for a few relevant classes, and took a new approach to managing the agility context.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndexV2.java
MMcM: This probably does make removing it easier (without leaving a stray base class) sooner.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -516,6 +523,20 @@ public byte[] readStoredFields(String segmentName, int docId) throws IOException
         return rawBytes;
     }
 
+    @Nonnull
+    public List<KeyValue> readAllStoredFields(String segmentName) {
```
MMcM, line 527: I half-expected this to be `Cursor`-based, rather than synchronously making a list of key values, from which a list of primary key byte arrays. Maybe there is no legitimate need for such an optimization. How many documents in a segment do we expect?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexOptions.java
```diff
@@ -97,6 +97,14 @@ public class LuceneIndexOptions {
      * merge with another adjacent partition.
      */
     public static final String INDEX_PARTITION_LOW_WATERMARK = "partitionLowWatermark";
+    /**
+     * Whether a separate B-tree index gives a direct mapping of live documents to segment by record primary key, that
+     * utilizes an optimized stored fields to manage merges.
+     * Boolean string ({@code true} or {@code false}).
+     * This is incompatible with {@link #PRIMARY_KEY_SEGMENT_INDEX_ENABLED} and {@link #OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED};
+     * if either of those are enabled, this cannot.
```
MMcM, line 105: What I found confusing here is that it seems to imply that using V2 means you have to forego optimized stored fields. But what is really going on is that with V2 you have no _choice_ about the matter. Maybe that just needs to be captured in this comment.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexValidator.java
```diff
@@ -56,6 +57,22 @@ public void validate(@Nonnull MetaDataValidator metaDataValidator) {
...
+    }
+
+    private static void validatePrimaryKeyOptions(final Map<String, String> options) {
+        if (Boolean.parseBoolean(options.get(LuceneIndexOptions.PRIMARY_KEY_SEGMENT_INDEX_V2_ENABLED))) {
+            if (Boolean.parseBoolean(options.get(LuceneIndexOptions.PRIMARY_KEY_SEGMENT_INDEX_ENABLED))) {
+                throw new MetaDataException(
+                        "Index cannot enable both " + LuceneIndexOptions.PRIMARY_KEY_SEGMENT_INDEX_V2_ENABLED + " and "
+                                + LuceneIndexOptions.PRIMARY_KEY_SEGMENT_INDEX_ENABLED);
+            }
+            if (Boolean.parseBoolean(options.get(LuceneIndexOptions.OPTIMIZED_STORED_FIELDS_FORMAT_ENABLED))) {
```
MMcM, line 70: This means that if I had `optimizedStoredFieldsFormatEnabled=false` in the index options, it would be fine. But something like that suggests a (confused) attempt to disable `usesOptimizedStoredFields`, which it won't.

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePrimaryKeySegmentIndex.java
```diff
@@ -313,31 +312,6 @@ public Collection<Accountable> getChildResources() {
...
-    /**
-     * Reconcile primary key index entries for a segment.
-     * Delete those for a segment that is about to be deleted or add for a segment that has been filled without going through the wrapper.
-     * The primary keys are gotten from the stored fields, so this must complete and not yet have been removed.
-     * @param reader a reader for the target segment
-     * @param add {@code true} to add entries; {@code false} to delete them
-     * @throws IOException if the stored fields gets an error
-     */
-    @SuppressWarnings("PMD.CloseResource")
-    public void addOrDeleteDocumentPrimaryKeys(@Nonnull SegmentReader reader, boolean add) throws IOException {
```
MMcM: This is approximately the code that would be needed if one wanted unoptimized storage and deferred segment index entry deletes, right? But we have no plans to support that combination.

PR #2583, opened 2024-03-14
Resolves #2582: Randomize index build tests

### LINKED ISSUE #2582: Randomize index build tests
Most of our index building tests already use a random number generator to create data. It should be fairly straightforward to use the new randomized test framework to run those tests with additional random seeds.

alecgrieser:
The index build tests already largely randomly generated their data. This updates their source of seeds so that they take a random seed as a parameter, and it configures them to run with additional random seeds if randomized tests are enabled.

This resolves #2582.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
+++ b/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
@@ -33,6 +33,8 @@ import java.util.stream.Stream;
  * A Utility class for adding randomness to tests.
  */
 public final class RandomizedTestUtils {
+    private static final long FIXED_SEED = 0xf17ed5eedL;
+
     private RandomizedTestUtils() {
     }
 
@@ -62,12 +64,19 @@ public final class RandomizedTestUtils {
      * the seed can be included in the test display name, which means that a failing seed can be
      * recorded and the test re-run with that seed.
      *
+     * <p>
+     * To ensure tests are consistent when random tests are not included, users are encouraged
+     * to provide a set of random seeds that are always included in the returned stream via
+     * the {@code staticSeeds} parameter. However, if this argument is left empty, this will
+     * still always include at least one fixed seed, {@value FIXED_SEED}.
+     * </p>
+     *
      * @param staticSeeds a set of seeds to always include in the returned random seeds
      * @return a stream of random {@code long}s to initialize {@link Random}s
      */
     @Nonnull
     public static Stream<Long> randomSeeds(long... staticSeeds) {
-        LongStream longStream = LongStream.of(staticSeeds);
+        LongStream longStream = staticSeeds.length == 0 ? LongStream.of(FIXED_SEED) : LongStream.of(staticSeeds);
         if (includeRandomTests()) {
             Random random = ThreadLocalRandom.current();
             longStream = LongStream.concat(longStream,
--- a/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
+++ b/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
@@ -22,10 +22,11 @@ package com.apple.test;
 
 import org.junit.jupiter.params.provider.Arguments;
 
+import javax.annotation.Nonnull;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
-import java.util.stream.IntStream;
+import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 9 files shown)


alecgrieser:
Note that this means that we're running the indexing tests three times now during prb (once each with some pre-baked seeds), but the duration of the build seems to be roughly in line with the rest of the recent PR runs

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildGroupedCountIndexTest.java
```diff
@@ -76,7 +76,9 @@ private static Stream<String> sourceIndexNames() {
 
     static Stream<Arguments> sourceIndexes() {
         return sourceIndexNames()
-                .map(Arguments::of);
+                .flatMap(sourceIndexName ->
+                        OnlineIndexerBuildIndexTest.randomSeeds().map(seed ->
+                                Arguments.of(sourceIndexName, seed)));
```
jjezra, line 81: Would this function be more readable if named `randomSourceIndexesAndSeeds`? 
(following `randomSeeds`'s presedent)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildSumIndexTest.java
```diff
@@ -183,34 +183,36 @@ private void sumRebuildFiltered(@Nonnull List<TestRecords1Proto.MySimpleRecord>
 
     static Stream<Arguments> sourceIndexes() {
         return Stream.of(null, REC_NO_INDEX, NUM_VALUE_2_INDEX)
-                .map(Arguments::of);
+                .flatMap(indexName -> OnlineIndexerBuildIndexTest.randomSeeds()
+                        .map(seed -> Arguments.of(indexName, seed)));
     }
 
     static Stream<Arguments> sourceIndexesAndFiltered() {
```
jjezra, line 190: `randomSourceIndexesAndFiltersAndSeeds`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildUnnestedIndexTest.java
```diff
@@ -158,6 +159,12 @@ static OnlineIndexerTestUnnestedRecordHandler instance() {
         }
     }
 
+    private static Stream<Arguments> overlapArgs() {
```
jjezra, line 162: `randomOverlapAndSeeds`?
(or `overlalpSeedsArgs`, or something consistent with the other arg streamers names)

jjezra (changes requested):
It seems to be a great improvement of the tests.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildGroupedCountIndexTest.java
alecgrieser: I went with `souceIndexesAndRandomSeeds` given that (at least for now) the "random" only applies to the seed. We could definitely consider something like a "random arg", where we pick a random source index and seed. It's possible we could simplify these tests a bit and instead of having one test for each configuration, have one test that supplies a random configuration. But for now, all I did was make the seeds random.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildSumIndexTest.java
alecgrieser: Now `sourceIndexesAndRandomSeeds`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildUnnestedIndexTest.java
alecgrieser: Sure, made more consistent: `overlapAndRandomSeeds`. I kind of like having "args" here, but I figured if we did that, we should make all of them have "args"

jjezra (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildGroupedCountIndexTest.java
```diff
@@ -74,9 +74,11 @@ private static Stream<String> sourceIndexNames() {
         return Stream.of(null, "MySimpleRecord$primary_key", NUM_VALUE_2_INDEX.getName(), STR_VALUE_INDEX.getName());
     }
 
-    static Stream<Arguments> sourceIndexes() {
+    @Nonnull
+    static Stream<Arguments> sourceIndexesAndRandomSeeds() {
         return sourceIndexNames()
-                .map(Arguments::of);
+                .flatMap(sourceIndexName -> OnlineIndexerBuildIndexTest.randomSeeds()
```
ScottDugas, line 80: Note: this will effectively cause it to have 5x the number of iterations. I don't think how `iterations` is flushed out is fully concluded, but is that what is desired here? If so, perhaps worth commenting that we want to run for each sourceIndexName with `iterations` random seeds.

fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java
```diff
@@ -47,12 +48,34 @@ private RandomizedTestUtils() {
...
+     * This can be supplied to a {@link org.junit.jupiter.params.ParameterizedTest} to run a test
+     * under different scenarios. This is preferred over, say, a {@code Stream<Random>} because
+     * the seed can be included in the test display name, which means that a failing seed can be
+     * recorded and the test re-run with that seed.
+     *
+     * @param staticSeeds a set of seeds to always include in the returned random seeds
+     * @return a stream of random {@code long}s to initialize {@link Random}s
+     */
+    @Nonnull
+    public static Stream<Long> randomSeeds(long... staticSeeds) {
```
ScottDugas, line 69: Given the test here, perhaps it would make sense for this to include one global seed if none are provided, or always. That seems like it would make it easier to use, and harder to accidentally have your test not run with any seeds.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerBuildGroupedCountIndexTest.java
ScottDugas: I think this differs from the usage in `LuceneIndexMaintainerTest`, but I'm not really sure which is better. That one is annoying in that you have to make up random numbers for the seeds when creating specific arguments. I think if this were aligned, it would be something like:
```
Stream.concat(
    sourceIndexNames.map(source -> Arguments.of(source, 234098),
    RandomizedTestUtils.randomArguments(random -> Arguments.of(randomSource(random), random.nextLong()));
```

but I don't know which is better.
Perhaps better to get a few more tests utilizing the framework and then coalesce, and cleanup the ones that don't align.

jjezra (approved):

ScottDugas (approved):

PR #2589, opened 2024-03-15
Fixes #2574: Close underlying contexts from AgileContext on error

### LINKED ISSUE #2574: If an error happens during merge with an agile context, it may leak contexts
There is no logic to close the underlying contexts when a merge fails while using an agile context.
When this support is added, we should ensure that any future access to the `AgileContext` fails with a clear error.

jjezra:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -299,30 +311,60 @@ public void set(byte[] key, byte[] value) {
...
+        @Override
+        public void abortAndReset() {
+            // Here: the lock status is undefined. The main goal of this function is to revive this object for post failure cleanups
+            lock.tryUnlockWrite();
+            boolean releasedLock = lock.tryUnlockRead();
+            for (int maxTries = 20; releasedLock && maxTries > 0; maxTries --) {
+                releasedLock = lock.tryUnlockRead();
+            }
+            committingNow = true;
+            currentContext = null;
```
ScottDugas, line 342: This should call `currentContext.close()` if it is not `null`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -65,6 +66,14 @@ static AgilityContext agile(FDBRecordContext callerContext, final long timeQuota
 
     void flush();
 
+    // This can be called to declare the object as 'closed' after flush. Future "flush" will succeed,
+    // but any attempt to perform a transaction read/write will cause an exception.
+    void flushAndClose();
+
+    // This will abort the existing agile context (if agile) and reset the locks. The only reason to
+    // call this function is after an transaction exception, buy a wrapper function, and to allow post failure cleanups.
```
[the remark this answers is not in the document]
ScottDugas: It might be worth calling out that this does not close the `AgilityContext`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -106,11 +106,16 @@ public CompletableFuture<Void> mergeIndex(@Nonnull LucenePartitioner partitioner
...
+        // This agilityContext will be used to determine/iterate grouping keys and partitions. The time gap between calls might
+        // be too long for a non-agile context.
         final AgilityContext agilityContext = getAgilityContext(true);
+
         if (! (rootExpression instanceof GroupingKeyExpression)) {
-            // TO DO: https://github.com/FoundationDB/fdb-record-layer/issues/2414
-            return mergeIndex(analyzerWrapper, TupleHelpers.EMPTY, partitioner, agilityContext);
+            // Here: empty grouping keys tuple
+            return mergeIndex(analyzerWrapper, TupleHelpers.EMPTY, partitioner, agilityContext)
+                    .whenComplete((vignore, eignore) -> agilityContext.flushAndClose());
```
ScottDugas, line 116: I don't think we want to call `flushAndClose()` here if there is an error, only if it is successful. If there is an error, I think we want `abortAndReset`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -122,51 +127,52 @@ public CompletableFuture<Void> mergeIndex(@Nonnull LucenePartitioner partitioner
...
                 ScanProperties.FORWARD_SCAN);
 
-        // TO DO: https://github.com/FoundationDB/fdb-record-layer/issues/2414
         return cursor
                 .map(tuple ->
                         Tuple.fromItems(tuple.getItems().subList(0, groupingCount)))
                 .forEachAsync(groupingKey -> mergeIndex(analyzerWrapper, groupingKey, partitioner, agilityContext),
-                        2);
+                        2)
+                .whenComplete((vignore, eignore) -> agilityContext.flushAndClose());
```
ScottDugas, line 135: I don't think we want to call `flushAndClose()` here if there is an error, only if it is successful. If there is an error, I think we want `abortAndReset`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -122,51 +127,52 @@ public CompletableFuture<Void> mergeIndex(@Nonnull LucenePartitioner partitioner
...
                                                @Nonnull LucenePartitioner partitioner, final AgilityContext agileContext) {
         if (!partitioner.isPartitioningEnabled()) {
-            try {
-                getDirectoryWrapper(groupingKey, null, agileContext).mergeIndex(analyzerWrapper);
-                return AsyncUtil.DONE;
-            } catch (IOException e) {
-                throw new RecordCoreStorageException("Lucene mergeIndex failed", e)
-                        .addLogInfo(LuceneLogMessageKeys.GROUP, groupingKey);
-            }
+            mergeIndexNow(analyzerWrapper, groupingKey, null);
```
ScottDugas, line 141: Do we want this to be using the same `agileContext` as the one used for getting the grouping keys?
In the old code, it reused the `AgileContext`, but with the new code, it creates a separate `AgileContext` for each merge.
It seems as though this could be creating an excessive number of transactions in the case where most groups/partitions don't need to do a merge. If you can open 10 indexes in one transaction, then the old code could go through 10 partitions in a single context, but this will now create at least 11 (one for grouping keys & partition metadata and one for each partition)

The upside to the approach here is that if you have a merge that actually does work on a single partition, you can be sure that work is committed, and decrease the chances of being half-way through a merge. Also, it decreases the chance that you will do one merge, and fail on the second merge, and lose all the work from the first, because writing the `segments_N` ended up in the second transaction. And, if you have partitions, each partition is probably getting close enough to the limit that opening more than a couple probably won't work anyways.


I'm not sure that I have a clear answer as to which is a better approach for efficient resource utilization, so going with this change is probably reasonable, as it is a simple solution to the leaking of contexts.
I just want to make sure that we are making a conscious decision here.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -65,6 +66,14 @@ static AgilityContext agile(FDBRecordContext callerContext, final long timeQuota
 
     void flush();
 
+    // This can be called to declare the object as 'closed' after flush. Future "flush" will succeed,
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -65,6 +66,14 @@ static AgilityContext agile(FDBRecordContext callerContext, final long timeQuota
...
+    // but any attempt to perform a transaction read/write will cause an exception.
+    void flushAndClose();
+
+    // This will abort the existing agile context (if agile) and reset the locks. The only reason to
```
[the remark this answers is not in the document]
jjezra: Added clarification.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -299,30 +311,60 @@ public void set(byte[] key, byte[] value) {
...
+
+        @Override
+        public void abortAndReset() {
+            // Here: the lock status is undefined. The main goal of this function is to revive this object for post failure cleanups
+            lock.tryUnlockWrite();
+            boolean releasedLock = lock.tryUnlockRead();
+            for (int maxTries = 20; releasedLock && maxTries > 0; maxTries --) {
+                releasedLock = lock.tryUnlockRead();
+            }
+            committingNow = true;
```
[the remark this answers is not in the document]
jjezra: Oops. Leftover from a previous version. 
✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -299,30 +311,60 @@ public void set(byte[] key, byte[] value) {
...
+        public void flushAndClose() {
+            closed = true;
+            commitNow();
+            logSelf("flushAndClose agility context");
+        }
+
+        @Override
+        public void abortAndReset() {
+            // Here: the lock status is undefined. The main goal of this function is to revive this object for post failure cleanups
+            lock.tryUnlockWrite();
```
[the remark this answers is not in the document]
jjezra: The assumption is that it is not called in the normal path and while other threads are using it, or the abort is likely to create an issue. But it synchronize on commitLockSync  seems harmless. 
✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: I wasn't sure about it. Can it be closed after commit failure? 
(meanwhile, ✅ )

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
ScottDugas: Yes, you are *always* supposed close contexts. They are tied to native state in the fdb client, that needs to get cleaned up. The close does not commit anything.

ScottDugas (approved):

PR #2595, opened 2024-03-19
cov with constant-alias

normen662:


fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -367,7 +367,7 @@ message PConditionSelectorValue {
 
 message PConstantObjectValue {
   optional string alias = 1;
-  optional int32 ordinal = 2;
+  optional string constant_id= 2;
```
normen662, line 370: This is ok, as no one is using this yet.

hatyo (approved):
LGTM.

PR #2597, opened 2024-03-20
WIP - Remodel aggregate plans

hatyo:


.idea/qaplug_profiles.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
normen662, line 1: please remove

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementStreamingAggregationRule.java
```diff
@@ -91,7 +89,7 @@ public void onMatch(@Nonnull final CascadesRuleCall call) {
 
         for (final var planPartition : planPartitions) {
             final var providedOrdering = planPartition.getAttributeValue(OrderingProperty.ORDERING);
-            if (requiredOrderingKeyValues == null || providedOrdering.satisfiesGroupingValues(requiredOrderingKeyValues)) {
+            if (providedOrdering.satisfiesGroupingValues(requiredOrderingKeyValues)) {
```
normen662, line 92: This may or may not work.

PR #2604, opened 2024-03-20
Resolevs #2601: Tests should run in unique key space paths 

### LINKED ISSUE #2601: Tests should run in unique key space paths
There's currently a bottle neck preventing test parallelization stemming from the fact that each test uses the same key space path. If we did a bit of refactoring to allocate unique key space paths for each test, then we could start thinking about parallelizing tests.

alecgrieser:
This introduces a new class, `TestKeySpacePathManager`, which does two things:

1. It allocates new key space paths for tests. Each newly allocated key space path contains a UUID that should make it unique for each test.
1. When the manager gets closed, it deletes the data from any key spaces that it has allocated.

This then provides a `TestKeySpacePathManagerExtension` which wraps the new class and makes sure that the path manager gets closed in an `AfterEach` callback. This also required reworking the way that the `FDBDatabase` gets initialized from the factory to use a testing extension, which in turn makes it harder to do something like incidentally use an `FDBDatabase` with the wrong API version set.

This path manager class is meant to be generic enough that if we needed to, we could theoretically allocate one keyspace for a single test _class_ and then use this same abstraction to handle its lifecycle, though that is not something that this attempts to handle. That's why the only extension that was added has a lifecycle that conforms with each test _method_ getting its own key space path. However, the path manager abstraction is supposed to be generic enough that we could theoretically extend it so that it manages a key space path which is shared among all tests in a given class.

This also does something similar with a `TestDatabaseExtension` and a `TestSubspaceExtension` for use within the `fdb-extensions` tests. Those could theoretically be spruced up a little, but they operate under similar principles.

This resolves #2601.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseFactoryImpl.java
```diff
@@ -93,6 +94,16 @@ public class FDBDatabaseFactoryImpl extends FDBDatabaseFactory {
     @Nonnull
     private Supplier<Boolean> transactionIsTracedSupplier = LOGGER::isTraceEnabled;
 
+    @API(API.Status.INTERNAL)
+    @VisibleForTesting
+    public static FDBDatabaseFactoryImpl testInstance(@Nonnull FDB initedFDB, @Nonnull APIVersion apiVersion) {
```
alecgrieser, line 99: This allows every test to have its own copy of the `FDBDatabaseFactory`. This means we will be able to run several concurrent tests that all access member variables of the `FDBDatabaseFactory` (which is normally a singleton). It also ensures that each test gets its own instance of an `FDBDatabase`, which means each one will independently track contexts to see if any are left un-opened by the test

fdb-extensions/src/test/java/com/apple/foundationdb/async/RangeSetTest.java
```diff
@@ -59,7 +60,12 @@
  * Tests for {@link RangeSet}.
  */
 @Tag(Tags.RequiresFDB)
-public class RangeSetTest extends FDBTestBase {
+@Execution(ExecutionMode.CONCURRENT)
```
alecgrieser, line 63: Note that when a test class is marked `CONCURRENT`, all of its test methods will be executed in parallel, but _only_if the system property, `junit.jupiter.execution.parallel.enabled` is set to `true`. This PR keeps the default as `false`, so the tests still execute in the same thread (for now).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBExceptionsTest.java
```diff
@@ -59,7 +66,13 @@
  * Tests for {@link FDBExceptions}.
  */
 @Tag(Tags.RequiresFDB)
+@Execution(ExecutionMode.CONCURRENT)
 class FDBExceptionsTest {
+    @Nonnull
+    public static final String COMPLETION_EXCEPTION_HELPER_LOCK = "CompletionExceptionLogHelper";
```
alecgrieser, line 72: There are several tests in this class that modify a static variable. They use the resource lock feature (with this lock) to ensure that those tests in particular do not run concurrently.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTestBase.java
```diff
@@ -146,34 +175,14 @@ protected RecordLayerPropertyStorage.Builder addDefaultProps(RecordLayerProperty
...
-        });
-    }
-
-    @AfterEach
-    public void checkForOpenContexts() {
-        int count = fdb.warnAndCloseOldTrackedOpenContexts(0);
-        assertEquals(0, count, "should not have left any contexts open");
+    @Nonnull
+    public static Index globalCountIndex() {
+        return new Index(COUNT_INDEX_NAME, new GroupingKeyExpression(EmptyKeyExpression.EMPTY, 0), IndexTypes.COUNT);
```
alecgrieser, line 180: The static `Index` objects can cause trouble because they are actually semi-stateful. In particular, the `RecordMetaDataBuilder` can update the last modified and and added versions (which is why the `clearAndInitialize` method used to reset them to zero with every test). Now, we just create a new instance rather than using static ones.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTestBase.java
```diff
@@ -146,34 +175,14 @@ protected RecordLayerPropertyStorage.Builder addDefaultProps(RecordLayerProperty
...
-        fdb.run(timer, null, context -> {
-            path.deleteAllData(context);
-            return null;
-        });
-    }
-
-    @AfterEach
-    public void checkForOpenContexts() {
-        int count = fdb.warnAndCloseOldTrackedOpenContexts(0);
-        assertEquals(0, count, "should not have left any contexts open");
```
alecgrieser, line 172: This `checkForOpenContexts` check has been moved to the `FDBDatabaseExtensions`'s `afterEach` callback.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreTestBase.java
```diff
@@ -220,8 +229,15 @@ public void commit(FDBRecordContext context) {
...
                 msg.addKeysAndValues(timer.getKeysAndValues());
+                msg.addKeyAndValue(LogMessageKeys.TRANSACTION_ID, context.getTransactionId());
                 logger.info(msg.toString());
             }
+        } catch (FDBExceptions.FDBStoreTransactionConflictException conflictException) {
+            List<Range> conflictRanges = context.getNotCommittedConflictingKeys();
+            if (conflictRanges != null && !conflictRanges.isEmpty()) {
+                conflictException.addLogInfo("conflict_ranges", conflictRanges);
+            }
+            throw conflictException;
```
alecgrieser, line 240: This is somewhat unrelated, but the idea is to hopefully use the `conflictingKeys` feature to help find problems if there are conflicts caused by running tests in parallel. This will ensure the `conflict_ranges` get put into the exception, which will then be logged (see the `ExceptionLoggingDetailsExtension`)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockTest.java
```diff
@@ -35,7 +43,21 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @Tag(Tags.RequiresFDB)
-class FDBDirectoryLockTest extends FDBDirectoryBaseTest {
+class FDBDirectoryLockTest {
```
alecgrieser, line 46: I ran this test in a mode that tracked all contexts, and it found some unclosed exceptions. This is currently not failing, because the opened contexts aren't being tracked (or I think that's the issue). The agility context was creating contexts it didn't open, though, at least on the branch I had. It's possible this has been fixed on a newer branch.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryBaseTest.java
```diff
@@ -48,20 +57,18 @@ public abstract class FDBDirectoryBaseTest {
...
+        fdb = dbExtension.getDatabase();
+        path = pathManager.createPath(TestKeySpace.RAW_DATA);
         FDBRecordContext context = fdb.openContext(getContextConfig());
+        subspace = fdb.run(path::toSubspace);
         directory = new FDBDirectory(subspace, context, null);
     }
 
+    @AfterEach
+    public void tearDown() {
+        directory.getCallerContext().close();
```
alecgrieser, line 69: This was exposed by the new `FDBDatabaseExtension`'s check for unopened contexts (when I ran the test in a mode that tracked all open contexts)

ScottDugas: It creates the `context` in `setUp`, so changing that to a field instead of a local variable seems better than using `getCallerContext()` here.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockTest.java
ScottDugas: AgilityContext should be properly closing all contexts, at least if it is closed properly, which this class might not do.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
```diff
@@ -161,7 +161,10 @@ public void testListAll() {
         assertCorrectMetricCount(LuceneEvents.Events.LUCENE_LIST_ALL, 2);
         // Assert that the cache is loaded only once even though directory::listAll is called twice
         assertCorrectMetricCount(LuceneEvents.Events.LUCENE_LOAD_FILE_CACHE, 1);
-        directory.getCallerContext().ensureActive().cancel();
+
+        FDBRecordContext originalContext = directory.getCallerContext();
+        originalContext.ensureActive().cancel();
+        originalContext.close();
```
ScottDugas, line 167: The close here should be unnecessary right, as it is closed in the `tearDown` of the base class?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
```diff
@@ -52,18 +50,9 @@
 // sonarcloud doesn't seem to be able to detect the junit4 style of just having the method start with "test"
 @SuppressWarnings("java:S2187")
 public class LuceneOptimizedStoredFieldsFormatTest extends BaseStoredFieldsFormatTestCase {
+    private static final FDBDatabaseExtension dbExtension = new FDBDatabaseExtension();
```
ScottDugas, line 53: I don't see any reason why this isn't using `BaseIndexFileFormatTestCaseUtils` for setup/teardown

gradle/scripts/log4j-test.properties
```diff
@@ -24,7 +24,7 @@ appenders = console
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d [%level] %logger{1.} - %m%n%ex{full}
+appender.console.layout.pattern = %d [%level] %logger{1.} - %m %X%n
```
ScottDugas, line 27: Don't we still want `%ex{full}`, at some future point we may want to play with the filters you can have here....

gradle/testing.gradle
```diff
@@ -99,6 +99,7 @@ def configureTestTask = { propertyPrefix, task ->
         tmpdir.mkdirs()
     }
     task.systemProperties['java.io.tmpdir'] = tmpdir.absolutePath
+    task.systemProperties['junit.jupiter.extensions.autodetection.enabled'] = 'true'
```
ScottDugas, line 102: If I remember correctly this will pick up any extensions from any dependency, right?

gradle/testing.gradle
```diff
@@ -217,6 +217,12 @@ tasks.withType(Test).configureEach { task ->
             systemProperties['tests.iterations'] = project.getProperty('tests.iterations')
         }
         task.testFramework.options.excludeTags.add('Performance')
+
+        if (task.name == 'test') {
```
ScottDugas, line 221: I don't think we want to limit this, I think we want all test tasks (at least for now), to set these properties.

gradle/scripts/log4j-test.properties
ScottDugas: See reference: https://logging.apache.org/log4j/2.x/manual/layouts.html
The %X adds the MDC

ScottDugas (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/test/FDBDatabaseExtension.java
```diff
@@ -0,0 +1,156 @@
...
+ * and that the {@link FDBDatabase} object used during a test does not leak between test runs. This registers
+ * call backs that run before and after tests, so it is suggested that users use the {@link org.junit.jupiter.api.extension.RegisterExtension}
+ * annotation to ensure that those callbacks run. Like so:
+ *
+ * <pre>{@code
+ *     @RegisterExtension
+ *     static final FDBDatabaseExtension = new FDBDatabaseExtension();
+ * }</pre>
+ */
+public class FDBDatabaseExtension implements BeforeEachCallback, AfterEachCallback {
```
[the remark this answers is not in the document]
alecgrieser: The one in `fdb-extensions` doesn't rely on any Record Layer abstractions. This one, in addition to initializing FDB, also creates a new factory and database object for each test, which allows each test to modify what would normally be global state. In theory, we could re-use `TestDatabaseExtension` here to initialize FDB, but then we would never exercise the `FDBDatabaseFactory`'s `initFDB` method in tests, which seems like it would be a bit of a miss.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -851,7 +852,7 @@ void threadedLuceneScanDoesntBreakPlannerAndSearch(@Nonnull Integer value) throw
             }
         } finally {
             // Restore the old executor so as not to disrupt other tests
-            FDBDatabaseFactory.instance().setExecutor(oldExecutor);
+            factory.setExecutor(oldExecutor);
```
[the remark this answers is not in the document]
alecgrieser: Yeah

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1346,8 +1348,8 @@ void testQueryWithManyDocuments() {
             }
         } finally {
```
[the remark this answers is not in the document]
alecgrieser: Yeah

gradle/testing.gradle
alecgrieser: I believe that would be true, if a dependency has included a test extension in a services file in their jar. I think that would only happen if we relied on a `test` configuration for any of our dependencies. I guess it is somewhat notable for downstream projects that if they relied on our test configuration and had this property set, then they'd get our automatic extensions registered, for better or worse. I'm not too concerned about the exception details one--that seems pretty harmless, in that all it does is generate an extra (useful!) log. The test MDC one I could see being slightly more controversial. If we cared to, we could remove one or both of those and add them to all of our test classes.

gradle/testing.gradle
alecgrieser: Hm, maybe we should revisit this when we set `paralle.enabled` to `true`. In the future, I think we want the `test` task to be executed in parallel, but not the `destructiveTest` or `performanceTest`. In theory, we could actually do `destructiveTest` with a `ResourcLock` to ensure that two tests that both wipe FDB don't execute in parallel.

gradle/scripts/log4j-test.properties
alecgrieser: Added back

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryBaseTest.java
alecgrieser: Sure

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
alecgrieser: Yeah, I suppose so

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
alecgrieser: Yeah, good point. I've consolidated that handling into `BaseIndexFileFormatTestCaseUtils`. There's something a little mismatched, in that the extension is static but is reset with every test. Theoretically, we'd want it to create a new one each time, but I'm not sure that is worth it given that I don't think we'll be parallelizing the JUnit4 tests.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/FDBThreadFilter.java
```diff
@@ -24,12 +24,13 @@
...
  * The randomized testing framework that Lucene uses has checks for leaked threads, but FDB depends on the
- * network thread, and common thread pool, so we filter those out.
+ * network thread, a common thread pool, and the {@link com.apple.foundationdb.test.TestExecutors#defaultThreadPool()}, so we filter those out.
  */
 public class FDBThreadFilter implements ThreadFilter {
     @Override
     public boolean reject(final Thread t) {
         return t.getName().equals("fdb-network-thread") ||
-               t.getName().startsWith("ForkJoinPool.commonPool");
+               t.getName().startsWith("ForkJoinPool.commonPool") ||
```
ScottDugas, line 33: The `ForkJoinPool.commonPool` might not be needed anymore.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedStoredFieldsFormatTest.java
```diff
@@ -73,17 +70,16 @@ private static boolean isUsingFDBDirectory() {
     public void setUp() throws Exception {
         super.setUp();
         TestingCodec.reset();
-        // We have to manually call the extension call backs because this is a JUnit4 class
-        dbExtension.beforeEach(null);
         TestFDBDirectory.reset();
+        BaseIndexFileFormatTestCaseUtils.resetStaticConfigs();
```
ScottDugas, line 74: Calling `resetStaticConfigs` means that this doesn't need to explicitly call `TestingCodec.reset()` or `TestFDBDirectory.reset()`. Same goes for the `tearDown()`

gradle/testing.gradle
ScottDugas: I suspect that a `ResourceLock` would effectively remove any parallelism anyways. And I think a `ResourceLock` only applies if they are in the same JVM.
Any test in `destructiveTest` would be expected to wipe the database, and thus they all would have the same resource lock, and if not, they can be in the regular `test`.

ScottDugas (approved):

PR #2607, opened 2024-03-21
allow nested/unnested results for group bys; group bys determine the …

normen662:
…result creation of streaming aggregates

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
@@ -34,14 +34,15 @@ import com.apple.foundationdb.record.query.plan.cascades.PartialMatch;
 import com.apple.foundationdb.record.query.plan.cascades.PredicateMap;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
 import com.apple.foundationdb.record.query.plan.cascades.RequestedOrdering;
-import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
 import com.apple.foundationdb.record.query.plan.cascades.explain.Attribute;
 import com.apple.foundationdb.record.query.plan.cascades.explain.InternalPlannerGraphRewritable;
 import com.apple.foundationdb.record.query.plan.cascades.explain.PlannerGraph;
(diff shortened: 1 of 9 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -81,16 +86,19 @@ public class GroupByExpression implements RelationalExpressionWithChildren, Inte
...
      * @param groupingValue The grouping {@code Value} used to determine individual groups, can be {@code null} indicating no grouping.
+     * @param aggregateValue The aggregation {@code Value} applied to each group.
+     * @param resultValueFunction a bi-function that allows us to create the actual result value of this expression
      * @param inner The underlying source of tuples to be grouped.
      */
-    public GroupByExpression(@Nonnull final AggregateValue aggregateValue,
-                             @Nullable final FieldValue groupingValue,
+    public GroupByExpression(@Nullable final Value groupingValue,
+                             @Nonnull final AggregateValue aggregateValue,
+                             @Nonnull final BiFunction<Value /* groupingValue */, Value, Value> resultValueFunction,
```
hatyo, line 96: Can this influence the computation of requested ordering? (`computeRequestedOrdering()`).
If the grouping key is something like `A, B, C`, but we return the result the following way: `B, C, A`, this is obviously inconsistent with the intended ordering of `(A, B, C)`.
What's your thoughts on this? e.g. limit the `BiFunction` s.t. it must return an order-compatible value?

normen662: There are three pieces of logic that have to do with ordering:
1. `GroupByExpression.computeRequestedOrdering()` which is before and after this PR somewhat broken as it pushes exactly one grouping part down.
2. `OrderingProperty` which seems to do it right no matter what as it looks at grouping values (and in the future aggregate values).
3. `ImplementStreamingAggregationRule` which seems to also do it correctly no matter what.

normen662: I think the reason why the result value does not matter much in `computeRequestedOrdering()` is because it expresses the ordering on top of `inner` not on top of the result of the `GroupBy`.

hatyo (approved):
LGTM.

PR #2621, opened 2024-03-25
RESOLVES #2616: Select Partition based on query filter

### LINKED ISSUE #2616: Select Partition based on query filter

wshaib-apple:
This update optimizes a given search query that contains a `partition field` predicate, to take advantage of the partition metadata and therefore to direct the searcher to start in the right _partition_ where applicable.
In order for a `partition field` predicate to be considered, the following conditions **must** be true:
1. the predicate's comparison operator is **not** `NOT_EQUALS`
2. Either the predicate is the only one in the filter, **OR** it is the only `partition field` predicate in a conjunction filter

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
@@ -22,7 +22,6 @@ package com.apple.foundationdb.record.lucene;
 
 import com.apple.foundationdb.KeyValue;
 import com.apple.foundationdb.Range;
-import com.apple.foundationdb.ReadTransaction;
 import com.apple.foundationdb.StreamingMode;
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncIterable;
@@ -87,11 +86,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
-
-import static com.apple.foundationdb.record.query.expressions.Comparisons.Type.GREATER_THAN;
-import static com.apple.foundationdb.record.query.expressions.Comparisons.Type.GREATER_THAN_OR_EQUALS;
-import static com.apple.foundationdb.record.query.expressions.Comparisons.Type.LESS_THAN;
-import static com.apple.foundationdb.record.query.expressions.Comparisons.Type.LESS_THAN_OR_EQUALS;
 
 /**
  * Manage partitioning info for a <b>logical</b>, partitioned lucene index, in which each partition is a separate physical lucene index.
@@ -167,7 +161,7 @@ public class LucenePartitioner {
     @Nullable
     public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
         if (isPartitioningEnabled()) {
-            LucenePartitionInfoProto.LucenePartitionInfo partitionInfo = selectQueryPartition(groupKey, null);
+            LucenePartitionInfoProto.LucenePartitionInfo partitionInfo = selectQueryPartition(groupKey, null).startPartition;
             if (partitionInfo != null) {
                 return partitionInfo.getId();
             }
@@ -176,10 +170,25 @@ public class LucenePartitioner {
     }
 
     /**
-     * return the partition ID on which to run a query, given a grouping key.
+     * Synchronous counterpart of {@link #selectQueryPartitionAsync(Tuple, LuceneScanQuery)}.
+     *
+     * @param groupKey group key
+     * @param luceneScanQuery query
+     * @return partition query hint, or <code>null</code> if partitioning isn't enabled or
+     * no partitioning metadata exist for the given query
+     */
+    public PartitionedQueryHint selectQueryPartition(@Nonnull Tuple groupKey, @Nullable LuceneScanQuery luceneScanQuery) {
+        return state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, selectQueryPartitionAsync(groupKey, luceneScanQuery));
+    }
+
+    /**
+     * Return the partition ID on which to run a query, given a grouping key.
+     * the most recent partition is returned, unless the query is sorted by the
+     * partitioning field in ascending order, in which case the oldest partition
+     * is returned.
      *
      * If no partitioning field predicate can be used to determine a particular
-     * starting partition, he most recent partition is returned, unless the query
+     * starting partition, the most recent partition is returned, unless the query
      * is sorted by the partitioning field in ascending order, in which case the
      * oldest partition is returned.
      *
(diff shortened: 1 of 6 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1352,6 +1359,180 @@ void simpleCrossPartitionQuery() {
...
+                    timeTooOld, "timeTooOld",
+                    timeTooNew, "timeTooNew");
+
+            //
+            // p0: from: t0+1200 to: t0+1300
+            // p1: from: t0+1301 to: t1+1400
+            // p2: from: t1+1410 to: t2+1500
+            // p3: from: t2+1510 to: t2+1600
+            //
+            createPartitionMetadata(index, groupKey, 0, time0, time0, Tuple.from(1, 1200), Tuple.from(1, 1300));
```
ScottDugas, line 1449:
```suggestion
            // create partition metadata directly rather than depending on repartitioning to ensure that the partitions
            // are what we want
            createPartitionMetadata(index, groupKey, 0, time0, time0, Tuple.from(1, 1200), Tuple.from(1, 1300));
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1352,6 +1359,180 @@ void simpleCrossPartitionQuery() {
...
+        Pair<Index, Consumer<FDBRecordContext>> indexConsumerPair = setupIndex(options, true, false);
+        final Index index = indexConsumerPair.getLeft();
+        Consumer<FDBRecordContext> schemaSetup = indexConsumerPair.getRight();
+
+        final RecordLayerPropertyStorage contextProps = RecordLayerPropertyStorage.newBuilder()
+                .addProp(LuceneRecordContextProperties.LUCENE_REPARTITION_DOCUMENT_COUNT, 8)
+                .build();
+        final String luceneSearch = "text:about";
+
+        class LuceneQueryBuilder {
```
ScottDugas, line 1382: Are you declaring a class inside a method? I didn't think java even allowed that, and it feels off.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1352,6 +1359,180 @@ void simpleCrossPartitionQuery() {
         }
     }
 
+    static Stream<Arguments> findStartingPartitionTest() {
+        return Stream.of(true, false).map(Arguments::of);
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void findStartingPartitionTest(boolean isSynthetic) {
```
ScottDugas, line 1368: It would be good to have some more functional tests that create records with duplicates, and non duplicates across many times, and then iterating over the entire time range to validate.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+                (comparisonType == LESS_THAN || comparisonType == LESS_THAN_OR_EQUALS) && isAscending) {
+            throw new RecordCoreException("Internal, unexpected comparison type and sort order combination");
+        }
+
+        final Range range;
+        final int limit;
+        final boolean reverse;
+        switch (comparisonType) {
+            case GREATER_THAN_OR_EQUALS:
+                range = new Range(state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE).addAll(partitionField)).pack(),
```
ScottDugas, line 261: This definitely seems more complicated than it needs to be, especially the fact that it fetches partition metadata, and then has to fetch more, or the fact that EQUALS has to iterate over all of them. Lets discuss offline after I run through some math

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+     * a specific starting partition, that partition will be computed and returned.
      *
      * @param groupKey group key
-     * @param sort sort
+     * @param luceneScanQuery query
      * @return partition, or <code>null</code> if partitioning isn't enabled or no partitioning metadata exist
      */
     @Nullable
-    public LucenePartitionInfoProto.LucenePartitionInfo selectQueryPartition(@Nonnull Tuple groupKey, @Nullable Sort sort) {
+    public LucenePartitionInfoProto.LucenePartitionInfo selectQueryPartition(@Nonnull Tuple groupKey, @Nullable LuceneScanQuery luceneScanQuery) {
```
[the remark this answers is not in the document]
wshaib-apple: I can convert this method as suggested, and then wrap it with another shim that calls it with `asyncToSync`, since ultimately this is called by `LuceneIndexMaintainer.scan()` which is synchronous.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
                 return state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, getOldestPartition(groupKey));
             }
+
+            // return oldest partition, unless the requested field predicate is even older: in that case return
+            // the newest (which guarantees no matches, and the cursor won't continue on to other partitions)
+            if ((comparisonType == LESS_THAN || comparisonType == LESS_THAN_OR_EQUALS) && isAscending) {
+                LucenePartitionInfoProto.LucenePartitionInfo oldest = state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, getOldestPartition(groupKey));
+                if (isPrefixOlderThanPartition(Tuple.from(partitionFieldPredicate.getComparand()), Objects.requireNonNull(oldest))) {
+                    return state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, getNewestPartition(groupKey, state.context, state.indexSubspace));
+                }
```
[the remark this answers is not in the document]
wshaib-apple: Right, good point.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+            // the oldest (which guarantees no matches, and the cursor won't continue on to other partitions)
+            if ((comparisonType == GREATER_THAN || comparisonType == GREATER_THAN_OR_EQUALS) && !isAscending) {
+                LucenePartitionInfoProto.LucenePartitionInfo newest = state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, getNewestPartition(groupKey, state.context, state.indexSubspace));
+                if (isNewerThan(Tuple.from(partitionFieldPredicate.getComparand()), Objects.requireNonNull(newest))) {
+                    return state.context.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, getOldestPartition(groupKey));
+                }
+                return newest;
+            }
+
+            // return comparand's "starting" partition
```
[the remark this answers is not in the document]
wshaib-apple: Thought it would be readable this way...

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+                                                                                                 boolean isAscending,
+                                                                                                 Comparisons.Type comparisonType) {
+        // we're here either because:
+        //  (EQUALS) or
+        //  (LESS_THAN OR LESS_THAN_OR_EQUALS) AND DESCENDING or
+        //  (GREATER_THAN_OR_EQUALS OR GREATER_THAN) AND ASCENDING
+        if ((comparisonType == GREATER_THAN || comparisonType == GREATER_THAN_OR_EQUALS) && !isAscending ||
+                (comparisonType == LESS_THAN || comparisonType == LESS_THAN_OR_EQUALS) && isAscending) {
+            throw new RecordCoreException("Internal, unexpected comparison type and sort order combination");
+        }
```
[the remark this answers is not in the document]
wshaib-apple: yes

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: basically you need to iterate in the case of EQUALS for the extreme case where you have one or more partitions where **all** the docs have the same partition field value. So you iterate until you hit the first partition that has non-equal docs too. You don't know ahead of time how many such partitions exist (super contrived in reality, but possible)

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+     * @param luceneScanQuery lucene query
+     * @return <code>null</code> if zero or more than one partitioning field predicate exist in the query's top-level
+     * clauses, otherwise the predicate is returned
+     */
+    @Nullable
+    private LuceneComparisonQuery checkQueryForPartitionFieldPredicate(final @Nonnull LuceneScanQuery luceneScanQuery) {
+        Query query = luceneScanQuery.getQuery();
+        List<LuceneComparisonQuery> partitionFieldPredicates = new ArrayList<>();
+        if (query instanceof BooleanQuery) {
+            List<BooleanClause> clauses = ((BooleanQuery) query).clauses();
```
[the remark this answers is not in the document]
wshaib-apple: what about `Occur.SHOULD`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+            // we only care about "top level" clauses, and won't descend
+            for (BooleanClause clause : clauses) {
+                Query clauseQuery = clause.getQuery();
+                if (clauseQuery instanceof LuceneComparisonQuery &&
+                        ((LuceneComparisonQuery) clauseQuery).getFieldName().equals(partitionFieldNameInLucene)) {
+                    partitionFieldPredicates.add((LuceneComparisonQuery)clauseQuery);
+                }
+            }
+        }
+        return partitionFieldPredicates.size() == 1 ? partitionFieldPredicates.get(0) : null;
```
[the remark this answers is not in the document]
wshaib-apple: because if we have more than one such clause, (e.g. `partitionField < 12 OR partitionField < 9 OR partition > 12 AND ...`) then it can complicated.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneQueryFieldComparisonClause.java
```diff
@@ -389,22 +389,46 @@ public BoundQuery bind(@Nonnull FDBRecordStoreBase<?> store, @Nonnull Index inde
...
+                    return toBoundQuery(new LuceneComparisonQuery(LongPoint.newExactQuery(appliedFieldName, (Long)comparand),
+                            appliedFieldName,
+                            comparison.getType(),
+                            comparand));
                 case NOT_EQUALS: {
                     long value = (Long)comparand;
                     BooleanQuery.Builder builder = new BooleanQuery.Builder();
-                    builder.add(LongPoint.newRangeQuery(appliedFieldName, Long.MIN_VALUE, value - 1), BooleanClause.Occur.SHOULD);
-                    builder.add(LongPoint.newRangeQuery(appliedFieldName, value + 1, Long.MAX_VALUE), BooleanClause.Occur.SHOULD);
+                    builder.add(
```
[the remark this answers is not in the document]
wshaib-apple: how so?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Right. That's what I was looking into

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Yeah :) it self contains that logic within the test method. If it's an eyesore I can refactor it.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -4204,7 +4399,7 @@ void testAutoCompleteSearchForPhraseWithoutFreqsAndPositions(Map<String, Index>
         }
     }
 
-    private static final List<String> spellcheckWords = List.of("hello", "monitor", "keyboard", "mouse", "trackpad", "cable", "help", "elmo", "elbow", "helps", "helm", "helms", "gulps");
+    private static final List<String> spellcheckWords = java.util.List.of("hello", "monitor", "keyboard", "mouse", "trackpad", "cable", "help", "elmo", "elbow", "helps", "helm", "helms", "gulps");
```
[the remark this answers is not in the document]
wshaib-apple: yeah not sure what happened there. I guess it must be IntelliJ "organizing" imports or something odd like that.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -167,29 +177,201 @@ public Integer selectQueryPartitionId(@Nonnull Tuple groupKey) {
...
+                range = new Range(state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE)).pack(),
+                        ByteArrayUtil.strinc(state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE).addAll(partitionField)).pack()));
+                limit = 1;
+                reverse = true;
+                break;
+            case LESS_THAN:
+            case EQUALS:
+                range = new Range(state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE)).pack(),
+                        ByteArrayUtil.strinc(state.indexSubspace.subspace(groupingKey.add(PARTITION_META_SUBSPACE).addAll(partitionField)).pack()));
+                limit = ReadTransaction.ROW_LIMIT_UNLIMITED;
```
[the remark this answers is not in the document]
wshaib-apple: see above

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Update: done, and merged in a separate PR

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1509,6 +1509,144 @@ void simpleCrossPartitionQuery() {
...
+                    timeTooNew, "timeTooNew");
+
+            //
+            // p0: from: t0+1200 to: t0+1300
+            // p1: from: t0+1301 to: t1+1400
+            // p2: from: t1+1410 to: t2+1500
+            // p3: from: t2+1510 to: t2+1600
+            // create partition metadata directly rather than depending on repartitioning to ensure that the partitions
+            createPartitionMetadata(index, groupKey, 0, time0, time0, Tuple.from(1, 1200), Tuple.from(1, 1300));
+            createPartitionMetadata(index, groupKey, 1, time0, time1, Tuple.from(1, 1301), Tuple.from(1, 1400));
```
ScottDugas, line 1564: It would be good to have a test where there is a gap between the `to` of one partition and the `from` of the next partition

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1509,6 +1509,144 @@ void simpleCrossPartitionQuery() {
...
+                    Type.GREATER_THAN, Map.of(SortType.ASCENDING, -1, SortType.DESCENDING, -1, SortType.UNSORTED, -1),
+                    Type.GREATER_THAN_OR_EQUALS, Map.of(SortType.ASCENDING, -1, SortType.DESCENDING, -1, SortType.UNSORTED, -1),
+                    Type.LESS_THAN, Map.of(SortType.ASCENDING, 0, SortType.DESCENDING, 3, SortType.UNSORTED, 3),
+                    Type.LESS_THAN_OR_EQUALS, Map.of(SortType.ASCENDING, 0, SortType.DESCENDING, 3, SortType.UNSORTED, 3),
+                    Type.EQUALS, Map.of(SortType.ASCENDING, -1, SortType.DESCENDING, -1, SortType.UNSORTED, -1));
+            startingPartitionExpectation.put(timeTooNew, expectationsForTimeTooNew);
+
+            // check all combinations against expectations
+            for (Long predicateComparand : List.of(time0, time1, time2, time1_2, timeTooOld, timeTooNew)) {
+                for (Comparisons.Type comparisonType : List.of(Type.GREATER_THAN, Type.GREATER_THAN_OR_EQUALS, Type.LESS_THAN, Type.LESS_THAN_OR_EQUALS, Type.EQUALS)) {
```
ScottDugas, line 1638: Would it also be good to test the starting partition when there is no comparison.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1509,6 +1509,144 @@ void simpleCrossPartitionQuery() {
...
+                    for (SortType sortType : EnumSet.allOf(SortType.class)) {
+                        LOGGER.debug("comparison: {} sort: {} time: {}", comparisonType, sortType, timesForLogging.get(predicateComparand));
+                        LuceneScanQuery luceneScanQuery = buildLuceneScanQuery(index, isSynthetic, comparisonType, sortType, predicateComparand, luceneSearch);
```
ScottDugas, line 1641: `buildLuceneScanQuery` uses pretty low-level constructs. Given the amount of logic around determining whether a predicate is appropriate, it seems like we should be generating the predicates from `RecordLayerQuery` objects.

ScottDugas, line 1641: It would be good to also have some tests of the logic that checks whether a predicate is suitable.
Perhaps it would be worth making that logic package-level, and then having tests that plan a query and assert that it either has a predicate or doesn't have a predicate, rather than asserting about which partition is chosen.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
ScottDugas: Yes, I mean `RecordQuery`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1341,42 +1345,133 @@ class GroupSpec {
...
+        final RecordLayerPropertyStorage contextProps = RecordLayerPropertyStorage.newBuilder()
+                .addProp(LuceneRecordContextProperties.LUCENE_REPARTITION_DOCUMENT_COUNT, 8)
+                .build();
+        final String luceneSearch = isSynthetic ? "simple_text:about" : "text:about";
+
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            LucenePartitioner partitioner = getIndexMaintainer(index).getPartitioner();
+
+            for (Comparisons.Type comparisonType : List.of(Type.NOT_EQUALS, Type.LESS_THAN, Type.LESS_THAN_OR_EQUALS, Type.GREATER_THAN, Type.LESS_THAN_OR_EQUALS, Type.EQUALS)) {
```
ScottDugas, line 1434: This is good, but it seems like we should have some more negative tests, such as using `or`, or comparing a different field, or putting it in the lucene string.
(probably if it is in the lucene string we should comment that parsing the string and supporting this is probably possible, but not a priority right now.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1454,6 +1457,145 @@ void partitionFieldPredicateDetectionTest(boolean isSynthetic) {
                     assertEquals(expectedLuceneComparisonQuery, luceneComparisonQuery);
                 }
             }
+
+            // test some "negative" use cases, asserting we won't mistakenly detect a partition field predicate and use
```
ScottDugas, line 1461: It would probably be nice to have this as a separate test.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1454,6 +1457,145 @@ void partitionFieldPredicateDetectionTest(boolean isSynthetic) {
...
+                            // planning raises exception
+                            false,
+                            // planning produces non-lucene plan
+                            true,
+                            // synthetic planning raises exception
+                            true,
+                            // syntehtic planning produces non-lucene plan
+                            false,
+                            // partition field predicate detected and will be used to optimize search
+                            false),
```
ScottDugas, line 1490: I think this is always false, and thus can be removed.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1454,6 +1457,145 @@ void partitionFieldPredicateDetectionTest(boolean isSynthetic) {
...
+                    // full disjunction
+                    Query.or(groupPredicate, textSearchPredicate, partitionFieldPredicate), new QueryPlanningExpectation(
+                            // planning raises exception
+                            false,
+                            // planning produces non-lucene plan
+                            true,
+                            // synthetic planning raises exception
+                            true,
+                            // syntehtic planning produces non-lucene plan
+                            false,
```
ScottDugas, line 1488: Given the comments, perhaps it would be better to have an `enum` with `NonLucenePlan` and `Exception` values, and then have the expectation take one for synthetic, and one for non-synthetic.

ScottDugas: It might also be ok to just silently ignore all of these. As far as this test is concerned it just cares that it doesn't ever pick the partitioning key.
If you do that, it would probably be good to have happy paths using the same query components, so something like:
```
Map<QueryComponent, Boolean> filterToOutcomeMap = Map.of(
    Query.or(groupPredicate, textSearchPredicate, partitionFieldPredicate), false,
    ....
    Query.and(groupPredicate, textSearchPredicate, partitionFieldPredicate), true))
```

And then you know that all the query construction and execution is good, and it is isolating to the fact that you are using `or`, etc.

ScottDugas (changes requested):
Not super important requests, but you have conflicts on this class anyway.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: It makes things a bit more complex to try to jam all of these situations into a single flow; for example, some filters will cause exceptions to be thrown. Also, the "positive" cases try all different combinations of operators etc, which the negative cases don't have to. I will keep them separate, and will split into two tests as suggested in a previous comment.

ScottDugas (approved):

PR #2624, opened 2024-03-27
Integrate Value Matching in Cascades (Step 1)

hatyo:
This is the first step of integrating the value mapping API introduced in #2498. A new method `translateValue` is added to `QueryPredicate` interface allowing us to perform translation of value-based predicates. This is required for a subsequent integration that pulls the (residual) predicates using the same value-mapping API for later applying them as a compensation. This PR does not affect the planner performance at all as it doesn't influence its matching logic.


hatyo:
There is still a bug that's causing the pull up of predicates not to work as expected. It is, for the most part, related to how we model the aggregate plans / match candidates. I will exclude the calculation of max-match-map and predicate pull up from this PR until we fix this modelling problem.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
```diff
@@ -680,7 +681,7 @@ default AliasMap bindIdentities(@Nonnull final RelationalExpression otherExpress
      */
     @Nonnull
     default Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candidateExpression,
-                                           @Nonnull final AliasMap aliasMap,
+                                           @Nonnull final AliasMap bindingAliasMap,
```
normen662, line 684: I thought we were going to add an optional of the `translationMap` that is constructed of the composition of the max match maps from below, or directly if leaves.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
```diff
@@ -694,19 +695,22 @@ default Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candi
      * @param aliasMap a map of alias defining the equivalence between aliases and therefore quantifiers
      * @param partialMatchMap a map from quantifier to a {@link PartialMatch} that pulled up along that quantifier
      *        from one of the expressions below that quantifier
+     * @param maxMatchMap the max match map.
```
normen662, line 698: Is this the max match map on this level? How is this constructed? It seems that every caller only ever calls this with `Optional.empty()` though I may be mistaken on this one...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
```diff
@@ -408,9 +408,9 @@ public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candid
                     .stream()
                     .allMatch(QueryPredicate::isTautology);
             if (allNonFiltering) {
-                return MatchInfo.tryMerge(partialMatchMap, mergedParameterBindingMap, PredicateMap.empty(), remainingValueComputationOptional)
-                        .map(ImmutableList::of)
-                        .orElse(ImmutableList.of());
+                return MatchInfo.tryMerge(partialMatchMap, ImmutableMap.of(), PredicateMap.empty(), PredicateMap.empty(), Optional.empty(), Optional.empty())
```
normen662, line 411: Why is the `remainingValueComputaionOptional` gone?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
```diff
@@ -243,7 +244,8 @@ public Optional<PredicateMultiMap.PredicateMapping> impliesCandidatePredicate(@N
                 //
                 return Optional.of(PredicateMultiMap.PredicateMapping.orTermMapping(this,
                         new ConstantPredicate(true),
-                        getDefaultCompensatePredicateFunction()));
+                        getDefaultCompensatePredicateFunction(),
```
normen662, line 247: I don't think these are right. Don't you need to use the proper translation map (that gets derived from the underlying  M3 maps)? Same with the order predicate kinds...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/PredicateWithValueAndRanges.java
```diff
@@ -331,7 +344,8 @@ public Optional<PredicateMapping> impliesCandidatePredicate(@NonNull final Alias
         if (semanticEquals(candidatePredicate, aliasMap)) {
```
normen662, line 344: Shouldn't this be a `semanticEquals()` on a predicate that is _translated_? Or is the `this` already translated?

normen662 (changes requested):

.idea/misc.xml
```diff
@@ -51,4 +51,7 @@
   <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="true" project-jdk-name="11" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/out" />
   </component>
+  <component name="SuppressionsComponent">
```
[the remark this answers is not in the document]
hatyo: done.

build.gradle
```diff
@@ -289,3 +289,8 @@ if (!ext.fdbEnvironment.isEmpty()) {
         }
     }
 }
+
+tasks.withType(Checkstyle) {
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -2557,6 +2595,17 @@ public Comparison translateCorrelations(@Nonnull final TranslationMap translatio
             }
         }
 
+        @Nonnull
+        @Override
+        @SuppressWarnings("PMD.CompareObjectsWithEquals")
```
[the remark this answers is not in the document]
hatyo: Which one exactly? `MultiColumnComparison` has its own `withValue`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchInfo.java
```diff
@@ -69,24 +70,37 @@ public class MatchInfo {
...
+    @Nonnull
+    private final PredicateMap accumulatedPredicateMap;
+
     @Nonnull
     private final List<MatchedOrderingPart> matchedOrderingParts;
 
     @Nonnull
     private final Optional<Value> remainingComputationValueOptional;
 
+    @Nonnull
```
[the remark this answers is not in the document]
hatyo: Sure thing, it would be nice to have documentation for the other members as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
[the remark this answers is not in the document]
normen662: Oh I see.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
hatyo: Yes that was the plan but I couldn't do it because of the ambiguity bug, so, not in this PR

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
hatyo: You're right, this was done before I reverted it due to a bug that prevents calculation of M3 correctly. I left the signature like this though so I can expand it in a subsequent PR with the proper calculation of the M3.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
hatyo: That is definitely a mistake. Reverted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
hatyo: I set the calculation of the translated predicate to `Optional.empty()`.

I think you're right, however the translation maps are providing means to translate a value between a query result value and a candidate result value, however, for predicates, I think we should only pull up the predicates along the candidate's expressions. In other words, one of them is horizontal, while the other is vertical if that makes sense.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/PredicateWithValueAndRanges.java
hatyo: I think the semantic equality is correctly (as it is using the proper alias mapping), it is checking whether the `this` (query) predicate is semantically equal to the candidate predicate it proceeds to return a mapping. I think this had been always the only check needed for establishing a predicate mapping for non-trivial predicates (tautology/contradiction) before the introduction of `PredicateWithValuesAndRanges`.

Nevertheless, I think the translation here is incorrect, because we will never apply the query predicate as a residual since it is semantically equal to the candidate predicate. In other words, the mapping should not contain a translated predicate here if I am not mistaken.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/QueryPredicate.java
```diff
@@ -174,13 +174,12 @@ default Optional<PredicateMapping> impliesCandidatePredicate(@NonNull final Alia
         if (candidatePredicate.isTautology()) {
             return Optional.of(PredicateMapping.regularMapping(this,
                     candidatePredicate,
-                    getDefaultCompensatePredicateFunction()));
+                    getDefaultCompensatePredicateFunction(),
+                    Optional.of(this.toResidualPredicate().translateValue(value -> value.translateCorrelations(TranslationMap.rebaseWithAliasMap(aliasMap))))));
```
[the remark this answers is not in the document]
hatyo: Replaced with `Optional.empty()` until we a better understanding about what we want to do here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -1464,6 +1485,12 @@ public Comparison translateCorrelations(@Nonnull final TranslationMap translatio
             return new ValueComparison(type, comparandValue.translateCorrelations(translationMap), parameterRelationshipGraph);
         }
 
+        @Nonnull
+        @Override
+        public Comparison translateValue(@Nonnull UnaryOperator<Value> valueTranslator) {
+            return withValue(valueTranslator.apply(comparandValue));
+        }
+
```
normen662, line 1493: Can you make this implementation the `default` in `Comparison` and then remove ll the overrides in all the overriding comparisons?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -2557,6 +2595,17 @@ public Comparison translateCorrelations(@Nonnull final TranslationMap translatio
             }
         }
 
+        @Nonnull
+        @Override
+        @SuppressWarnings("PMD.CompareObjectsWithEquals")
+        public Comparison translateValue(@Nonnull final UnaryOperator<Value> valueTranslator) {
```
normen662, line 2601: I think this can then be done by the default, no?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -2762,6 +2811,28 @@ public Comparison withType(@Nonnull final Type newType) {
...
+            if (newComparison == originalComparison) {
+                return this;
+            }
+            return from(function, newComparison);
+        }
+
+        @Nonnull
+        @Override
+        @SuppressWarnings("PMD.CompareObjectsWithEquals")
+        public Comparison translateValue(@Nonnull final UnaryOperator<Value> valueTranslator) {
```
normen662, line 2828: Same here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
```diff
@@ -291,9 +293,10 @@ private Optional<PredicateMultiMap.PredicateMapping> impliesWithValuesAndRanges(
                              Objects.requireNonNull(foldNullable(Function.identity(),
                                      (queryPredicate, childFunctions) -> queryPredicate.injectCompensationFunctionMaybe(partialMatch,
                                              boundParameterPrefixMap,
-                                             ImmutableList.copyOf(childFunctions)))))));
+                                             ImmutableList.copyOf(childFunctions))))),
+                    Optional.of(this.toResidualPredicate().translateValue(value -> value.translateCorrelations(TranslationMap.rebaseWithAliasMap(aliasMap))))));
```
normen662, line 297: Can you make this `Optional.empty()` as well?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/PredicateWithValueAndRanges.java
```diff
@@ -234,6 +235,14 @@ public static PredicateWithValueAndRanges ofRanges(@Nonnull final Value value, @
         return new PredicateWithValueAndRanges(value, ranges);
     }
 
+    @Nonnull
+    @Override
+    public PredicateWithValueAndRanges translateValue(@Nonnull final UnaryOperator<Value> translator) {
+        final var newValue = Verify.verifyNotNull(this.getValue().replace(translator));
```
normen662, line 241: Can you change this to just:

```
Verify.verifyNotNull(translator.apply(this.getValue());
```

This is more powerful as you can put the `replace()` call into the lambda you hand in and you can truly just swizzle the old `Value` tree for a newly constructed one. It's also the way this is done in `Comparison`.

Can you do the same for all the other implementations?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
normen662: I am not sure I follow your comment fully, however, I think it doesn't matter for this PR.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
hatyo: I am afraid that is not possible. `comparandValue` is special to `ValueComparison`. the two remaining overrides in `MultiColumnComparison` and `InvertedFunctionComaprison` are different than this one, so there is no value of moving it up, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
hatyo: No, `inner` is special to `MultiColumnComparison`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
hatyo: ditto.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/OrPredicate.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
normen662: Right, but the special stuff happens in `withValue()` and its overrides so you just need to call that one.

normen662 (approved):

PR #2627, opened 2024-03-29
Resolve #2626: Agility Context: check/commit context before the user …

### LINKED ISSUE #2626: Agility Context: check/commit context before the user operation
To reduce the probability of timeout/transaction too old, add a `commitIfNeeded()` call in the beginning of the accept/apply operations.

jjezra:
…operation

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
@@ -316,6 +316,7 @@ public interface AgilityContext {
         @Override
         public <R> CompletableFuture<R> apply(Function<FDBRecordContext, CompletableFuture<R>> function) {
             ensureOpen();
+            commitIfNeeded();
             final long stamp = lock.readLock();
             boolean successfulCreate = false;
             try {
@@ -363,6 +364,7 @@ public interface AgilityContext {
         @Override
         public void accept(final Consumer<FDBRecordContext> function) {
             ensureOpen();
+            commitIfNeeded();
             final long stamp = lock.readLock();
             try {
                 createIfNeeded();
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
@@ -313,29 +313,55 @@ class AgilityContextTest extends FDBRecordStoreTestBase {
                         );
                         break;
                     case Apply:
-                        final CompletionException completionException = Assertions.assertThrows(CompletionException.class, () ->
-                                agilityContext.apply(innerContext -> innerContext.ensureActive()
-                                        .get(key).thenApply(oldVal -> {
-                                            if (oldVal == null) {
+                        if (i == 0) {
+                            final CompletionException completionException = Assertions.assertThrows(CompletionException.class, () -> {
+                                try {
+                                    agilityContext.apply(innerContext -> innerContext.ensureActive()
+                                            .get(key).thenApply(oldVal -> {
+                                                if (oldVal == null) {
+                                                    innerContext.ensureActive().set(key, val);
+                                                } else {
+                                                    final Tuple oldTuple = Tuple.fromBytes(oldVal);
+                                                    innerContext.ensureActive().set(key,
+                                                            TupleHelpers.subTuple(oldTuple, 0, 2)
+                                                                    .add(oldTuple.getLong(2) + 1)
+                                                                    .pack());
+                                                }
+                                                throw new FailException();
+                                            })).join();
+                                } catch (Exception ex) {
+                                    agilityContext.abortAndClose();
+                                    throw ex;
+                                }
+                            });
+                            assertThat(completionException.getCause(), Matchers.instanceOf(FailException.class));
+                        } else {
+                            Assertions.assertThrows(RecordCoreStorageException.class, () ->
+                                    agilityContext.apply(innerContext -> innerContext.ensureActive()
+                                            .get(key).thenApply(oldVal -> {
                                                 innerContext.ensureActive().set(key, val);
-                                            } else {
-                                                final Tuple oldTuple = Tuple.fromBytes(oldVal);
-                                                innerContext.ensureActive().set(key,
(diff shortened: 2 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -299,6 +299,7 @@ private void reportFdbException(Throwable ex) {
         @Override
         public <R> CompletableFuture<R> apply(Function<FDBRecordContext, CompletableFuture<R>> function) {
             ensureOpen();
+            commitIfNeeded();
```
ScottDugas, line 302: This needs to be before the `ensureOpen()`. `commitIfNeeded()` results in us not having a context.
This is probably an indication that we could use a couple more tests, specifically where it calls `accept`/`apply` and then sleeps longer than the timeout, and then calls `accept`/`apply` again.
I think if you take advantage of versionstamp you could assert that they actually happen in separate transactions.
I think something like this should fail without your change and pass after your change is in:
```java
    @Test
    void name() {
        try (FDBRecordContext factoryContext = openContext()) {
            final AgilityContext agilityContext = AgilityContext.agile(factoryContext, 2, 10000);
            AtomicReference<FDBRecordContext> firstOperation = new AtomicReference<>();
            AtomicReference<FDBRecordContext> secondOperation = new AtomicReference<>();
            agilityContext.accept(context -> {
                context.ensureActive().set(key, value);
                firstOperation.set(context);
            });
            Thread.sleep(5);
            agilityContext.accept(context -> {
                context.ensureActive().set(key, value);
                secondOperation.set(context);
            });
            agilityContext.flush();
            MatcherAssert.assertThat(secondOperation.get().getCommittedVersion(), Matchers.greaterThan(firstOperation.get().getCommittedVersion()));
        }
    }
```
You could parameterize that test to run both with `accept` and `apply` like some of the other tests in `AgilityContextTest`

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: IIUIC, this test may cause a "transaction too old error". As the first call will not commit and the second, even if trying to commit immediately, may already fail. Maybe we could have had: 
```
....
            Thread.sleep(4200);
            agilityContext.accept(context -> {
                context.ensureActive().set(key, value);
                secondOperation.set(context);
            });
```
But it still might be a very flaky test.

ScottDugas: I don't think you should have that long of a sleep. By looking at the committed versions, as I described, you are ensuring that `AgilityContext` is checking its time limit, and committing them separately.
I guess you could also just do a reference equality check to assert that the two operations get different contexts, that might be better, as it depends on less fdb semantics.

jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
```diff
@@ -494,6 +501,56 @@ void testCloseOnCommitFailure() {
...
+        try (FDBRecordContext userContext = openContext()) {
+            key = this.path.toSubspace(userContext).pack(Tuple.from(prefix, "a").pack());
+            final AgilityContext agilityContext = AgilityContext.agile(userContext, 2, 10000);
+            AtomicReference<FDBRecordContext> firstOperation = new AtomicReference<>();
+            agilityContext.accept(context -> {
+                context.ensureActive().set(key, Tuple.from(1).pack());
+                firstOperation.set(context);
+            });
+            Thread.sleep(5);
+            // Here: after this operation, the first auto-context should be committed
```
ScottDugas, line 516: Is it worth asserting that the first context has not committed at this point, to prevent the possibility that the `accept` above takes 2ms?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
```diff
@@ -494,6 +501,56 @@ void testCloseOnCommitFailure() {
         }
     }
 
+    @Test
+    void testAutoCommitVersionStampOuterSleep() throws InterruptedException {
```
ScottDugas, line 505: Both of these run against `accept`, it would also be good to run them against `apply`.
Probably no need to run them for any other part of the api, since those all call `accept`/`apply`.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
```diff
@@ -342,14 +344,19 @@ void testAgilityContextOneLongWriteFail(boolean useProp, Method method, LimitTyp
...
+                try (FDBRecordContext validationContext = openContext(insertProps)) {
+                    for (int i = 0; i < loopCount; i++) {
+                        byte[] key = subspace.pack(Tuple.from(2023, i));
+                        final byte[] value = validationContext.ensureActive().get(key).join();
+                        assertNull(value);
+                    }
                 }
+            } else {
+                // Here: since quota is also checked before the operation, it may be committed before throwing the exception
+                assertThat(timer.getCount(limitType.timerEvent), Matchers.greaterThan(0));
```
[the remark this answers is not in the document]
jjezra: Had to change the test a bit. 
✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
jjezra: ✅

ScottDugas (approved):

PR #2628, opened 2024-03-29
Add tests of lucene validity with many documents

ScottDugas:
These tests are mostly slow, and won't run as part of PRB, but add coverage that is otherwise hard to get since lucene is in charge of the merging logic


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -151,6 +153,113 @@ void randomizedRepartitionTest(boolean isGrouped,
...
+                      long seed) throws IOException {
+        manyDocument(isGrouped, isSynthetic, primaryKeySegmentIndexEnabled, partitionHighWatermark,
+                repartitionCount, loopCount, 10, seed);
+    }
+
+
+    static Stream<Arguments> manyDocumentsArguments() {
+        return Stream.concat(
+                Stream.concat(
+                        Stream.of(Arguments.of(true,  true,  true,  10, 2, 50, 3, -644766138635622644L)),
```
ScottDugas, line 189: This takes about 12 seconds in PRB.

jjezra (approved):

PR #2635, opened 2024-04-03
Lucene lock flush

jjezra:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -317,6 +329,21 @@ public <R> CompletableFuture<R> apply(Function<FDBRecordContext, CompletableFutu
             });
         }
 
+        @Override
+        @SuppressWarnings("PMD.CloseResource") // closed in a future
+        public <R> CompletableFuture<R> applyInRecoveryPath(Function<FDBRecordContext, CompletableFuture<R>> function) {
+            // Create a new, dedicated context. Apply, flush, and close it.
+            FDBRecordContextConfig contextConfig = contextConfigBuilder.build();
+            final FDBRecordContext recoveryContext = database.openContext(contextConfig);
+            return function.apply(recoveryContext)
```
ScottDugas, line 338: if `function` throws an exception, you won't ever close the `recoveryContext`.
The easiest way I've found to do this is something like:
```
boolean successful = false;
try {
    future = function.apply(recoveryContext).whenComplete( /* ... as you have it */);
    successful = true;
} finally {
    if (!successful) {
    recoveryContext.close();
}
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -358,20 +385,22 @@ public void flushAndClose() {
...
             // Here: the lock status is undefined. The main goal of this function is to revive this object for post failure cleanups
             synchronized (commitLockSync) {
+                closed = true;
+                committingNow = true; // avoid future commits
+                currentWriteSize = 0;
+                if (currentContext != null) {
+                    currentContext.close();
+                    currentContext = null;
+                }
+                // release the locks, in case another thread is waiting on them
```
ScottDugas, line 398: This will just mean that the other operations will fail fast, right?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -920,13 +920,12 @@ private void clearLockIfLocked() {
     @Override
     public void close() {
         try {
-            clearLockIfLocked();
+            clearLockIfLocked(false);
```
ScottDugas, line 923: Don't we close the directory after failures? Will this have the right behavior?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -317,6 +329,21 @@ public <R> CompletableFuture<R> apply(Function<FDBRecordContext, CompletableFutu
             });
         }
 
+        @Override
+        @SuppressWarnings("PMD.CloseResource") // closed in a future
+        public <R> CompletableFuture<R> applyInRecoveryPath(Function<FDBRecordContext, CompletableFuture<R>> function) {
```
ScottDugas, line 334: We should add tests that this does what we want, namely, if we have a bunch of threads using the same `AgilityContext`, with a low time limit, it should commit frequently, and once we call `abortAndClose` nothing else should be committed, and everything should be aborted.
And we then call `applyInRecoverContext` and that works. I think if you have an `AtomicBoolean` that is written from the other threads, and gets changed to `true` right after closing, and then assert that `true` is never written.
It's not perfect, but better than nothing (although perhaps not better than something you can think of)

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: ✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: New read/writes will fail in `ensureOpen()`. If there is any commitNow() waiting for a write lock, it will become a no-op. 
Added an `ensureOpen()` to `createIfNeeded()` in case there is a user operation waiting for a read lock. 
✅

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
jjezra: If failed while a lock exists, `clearLockIfLocked(false)` will fail with "Already Closed" error, but the second attempt with true should succeed. 
Note that `abortAndClose()` can be called multiple times.

ScottDugas (approved):

PR #2639, opened 2024-04-04
FIXES #2638: Review Lucene's use of executors

### LINKED ISSUE #2638: Review Lucene's use of executors
audit Lucene's usage of `supplyAsync` or `runAsync`, as they use the default executor, but they should be using a special executor

wshaib-apple:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -503,14 +503,14 @@ private static IndexOptions getIndexOptions(@Nonnull String value) {
     @Override
     public RecordCursor<IndexEntry> scanUniquenessViolations(@Nonnull TupleRange range, @Nullable byte[] continuation, @Nonnull ScanProperties scanProperties) {
         LOG.trace("scanUniquenessViolations");
-        return RecordCursor.empty();
+        return RecordCursor.empty(state.context.getExecutor());
```
ScottDugas, line 506: I suspect it's always the same, but it should use the executor provided to the maintainer, rather than the one on the context.
```suggestion
        return RecordCursor.empty(executor);
```

ScottDugas (changes requested):

MMcM (dismissed):

ScottDugas (approved):

ScottDugas (dismissed):

ScottDugas (approved):

PR #2646, opened 2024-04-05
Resolve #2645: Logging for Lucent file locks

### LINKED ISSUE #2645: Improve Lucent file lock logging
The current "create Lucene file lock" is made before the lock is taken, meaning that if the locking operation fails, it creates the impression that the lock was takes (and is never cleared). In order to debug locking issues, we need to know which locks are taken, which failed to be taken and which ones were cleared.

ohadzeliger:


ohadzeliger:
Release notes mixup

PR #2651, opened 2024-04-09
Fixes #2650: Equality index scan can match additional suffix …

### LINKED ISSUE #2650: Equality index scan can match additional suffix starting with zero byte
The target key, used as a prefix, will end with `\x00`; the embedded zero will encode as `\x00\xff` and so match the prefix.

MMcM:
… starting with zero byte.

Stop an inclusive end with `key+FF`, not `strinc(key)`.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -124,7 +125,9 @@ private CompletableFuture<Void> buildMultiTargetIndex(@Nonnull SubspaceProvider
         } else {
             final Range range = tupleRange.toRange();
             rangeStart = range.begin;
-            rangeEnd = range.end;
+            // tupleRange has an inclusive high endpoint, so end isn't a valid tuple.
+            // But buildRangeOnly needs to convert missing Ranges back to TupleRanges, so round up.
+            rangeEnd = ByteArrayUtil.strinc(range.end);
```
MMcM, line 130: Indexing by record deals mostly in tuple ranges or ranges of tuples. First because `scanRecords` needs them and second because it uses record primary keys as built-range boundaries.

An alternative approach would be something like the following.
* Pull the code out of `buildRangeOnly` that recovers a `TupleRange` from a `Range` into a static `TupleRange.fromRange`.
* Make it handle the `INCLUSIVE` case by recognizing the (illegal) `FF` byte at the end of the high bytes.
* Make `postIterateRangeOnly` take the whole `TupleRange` and handle that high endpoint case.

All of which seemed unnecessarily complicated if this record type key prefix case is the only inclusive range in this code.

Is there still a problem case?

alecgrieser: Is `ByteArrayUtil.strinc(range.end)` guaranteed to return a valid Tuple? Does this only work if the last element of the tuple is an integer (and incrementing the string doesn't cause it to roll over to the next byte)? For example, this works for a tuple ending in `3L` (`strinc(Tuple(3L)) = strinc(b'\x15\x03') = b'\x15\x04' = Tuple(4L)`) but not `255L` (`strinc(Tuple(255L)) = strinc(b'\x15\xff') = b'\x16'`, which is not a Tuple) or like "hello" (`strinc(Tuple("hello")) = strinc(b'\x02hello\x00) = b'\x02hello\x01'`, also not a Tuple).

FWIW, I don't think this code makes the situation any worse, as it's effectively going to revert the value of `rangeEnd` to its value before this change.

I think having a `TupleRange.fromRange` that can produce a range of tuples from a range with non-tuple elements could make sense. There's part of me that wonders if we instead shouldn't have a `scanRecordsInternal` variant that takes a normal FDB range for occasions like these where the range of records scanned doesn't necessarily align with the rest of the API. The `TupleRange` API is already missing some things, like an endpoint type that means "exclusive of this key but not exclusive of all values prefixed by value" (except continuations, which _kind of_ mean that).

alecgrieser: cc @jjezra

alecgrieser (dismissed):
LGTM, barring some copy-paste bugs in the test. I think your point about index ranges merits some additional discussion, though I believe the change here mostly just preserves _status quo ante bellum_, so at least it doesn't make things worse

alecgrieser (approved):

Issue #2484, opened 2024-02-05
Remove usages of `org.apache.commons.lang3.tuple.Pair` et al


hatyo:
Part of cutting the dependency on `org.apache.commons.lang3` requires finding / implementing alternatives to:

- `org.apache.commons.lang3.tuple.Pair`.
- `org.apache.commons.lang3.tuple.ImmutablePair`
- `org.apache.commons.lang3.tuple.MutablePair`

PR #2652, opened 2024-04-10
Begins #2484: Remove usages of `org.apache.commons.lang3.tuple.Pair` et al

alecgrieser:
This begins #2484 by introducing new `Pair` implementations to serve as alternatives to the one from Apache comments. It then uses the classes in a few places that do not have visible consequences. For example, we have a few places where we expose the old `Pair` class in our API (usually as a return value, but sometimes as a parameter). To fully remove the class, we'll need to replace those, but for now, this just adjusts usages that are entirely internal (which means mostly test classes for now).


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/pair/ImmutablePair.java
```diff
@@ -0,0 +1,107 @@
...
+        return Objects.equals(getLeft(), other.getLeft()) && Objects.equals(getRight(), other.getRight());
+    }
+
+    @Override
+    public int hashCode() {
+        if (hashCode == 0) {
+            synchronized (this) {
+                if (hashCode == 0) {
+                    hashCode = Objects.hash(left, right);
+                }
```
hatyo, line 83: I do like this, do you think this is more efficient than using a lambda which lazily calculate (and cache) the hash code?
At first glance it seems like it is faster, but I am not sure considering that `serialized` could have an internal bound lock.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/pair/NonnullPair.java
```diff
@@ -0,0 +1,91 @@
...
+ * A {@link Pair} implementation where both elements are not null. This utility allows
+ * the caller to omit null checks when retrieving elements from the pair.
+ *
+ * @param <L> the type of the left element
+ * @param <R> the type of the right element
+ */
+@API(API.Status.INTERNAL)
+public class NonnullPair<L, R> implements Pair<L, R> {
+    @Nonnull
+    private final Pair<L, R> pair;
```
hatyo, line 40: Can we omit this and just store left and right? We could avoid an extra allocation.

But If you intend to give the caller the ability to use an _existing_ `Pair` object as non-null, maybe you could make `NonnullPair` an interface, and provide a utility method in `Pair` that allows the user to treat _current_ `Pair` object as non-null:

```java
final var p = Pair.of(3, 4);
final var nnp = p.asNonnullable();
```

Or at least provide another constructor that accepts passing a `Pair` argument allowing it to strengthen its contract (making it non null).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/pair/ComparablePair.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * An implementation of the {@link Pair} that adds comparison support to the {@link Pair}. This
+ * allows this kind of pair to be used within ordered collections. The comparison performs lexicographic comparison
+ * on the elements, that is, the left elements are compared first and if equal, comparison proceeds to the right element.
+ *
+ * @param <L> the type of the left element in the pair
+ * @param <R> the type of the right element in the pair
+ */
+@API(API.Status.INTERNAL)
+public class ComparablePair<L extends Comparable<? super L>, R extends Comparable<? super R>> implements Pair<L, R>, Comparable<ComparablePair<L, R>> {
```
hatyo, line 38: This is great!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/pair/ImmutablePair.java
alecgrieser: One thing to note is that the most common technique that we use for memoizing things like this is using `Suppliers.memoize`, which has an internal double-checked lock very similar to this: https://github.com/google/guava/blob/858caf425cdf3183776b61b437ba7a4129b6539b/guava/src/com/google/common/base/Suppliers.java#L184

So, we wouldn't save the lock by switching to a lambda. We may save an allocation, though hotspot could be inlining some of that, so it's a bit hard to say. One approach that we could try here would be a non-locking double checked lock, that is, replacing this with something like:

```java
if (hashCode == 0) {
    hashCode = Objects.hash(left, right);
}
return hashCode;
```

As long as the `hashCode` variable is `volatile` and as long as the hashes of left and right always return the same value (which they should), I think this would work, and we'd avoid the lock at the cost of sometimes having multiple threads calculating the hash for the same `Pair`.

I'm not sure I have that much of a strong preference between this approach and using a (constant) lambda with `Suppliers.memoize`. I sort of think it's actually easier to see what's going on, but I grant that others may like delegating the lock to a library function.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/util/pair/NonnullPair.java
alecgrieser: The main reason not to store `left` and `right` separately is that we'd need to basically duplicate the code in `ImmutablePair.java`. In theory, if we had templating, we could let the compiler do that, but not in vanilla Java. I'm not _too_ concerned with the extra allocation, because Hotspot may be able to do that inlining for us, and if it can't, I think this is a place where it would be better to see if there are performance problems first and then optimize.

The other reason concerns `MutablePair`s: there's not (currently) a `MutablePair` implementation, though we do use `MutablePair`s once in the codebase. (I may actually remove that usage in a future refactoring, in which case this becomes a bit moot.) But if you wanted a non-null AND mutable pair, then you could do this by wrapping a `MutablePair` in a `NonnullPair`. As you say, we'd also need an API for wrapping a `Pair`--and I think either of the ones you've suggeseted would do that.

I do think that those are API improvements we could consider. So far, in converting code, I haven't really needed them, but I can see their utility. My inclination is to add this as a follow up later if we need it.

PR #2655, opened 2024-04-11
Have indexing by record scan by key ranges instead of by tuple ranges

MMcM:
This avoids having to recover a `TupleRange` from an unbuilt `Range`.

See https://github.com/FoundationDB/fdb-record-layer/pull/2651#discussion_r1559872782


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -172,12 +170,13 @@ private CompletableFuture<Boolean> buildRangeOnly(@Nonnull FDBRecordStore store,
             if (range == null) {
                 return AsyncUtil.READY_FALSE; // no more missing ranges - all done
             }
-            final Tuple rangeStart = RangeSet.isFirstKey(range.begin) ? null : Tuple.fromBytes(range.begin);
-            final Tuple rangeEnd = RangeSet.isFinalKey(range.end) ? null : Tuple.fromBytes(range.end);
-            final TupleRange tupleRange = TupleRange.between(rangeStart, rangeEnd);
+            final byte[] keyPrefix = store.recordsSubspace().pack();
+            final byte[] rangeStart = ByteArrayUtil.join(keyPrefix, range.begin);
+            final byte[] rangeEnd = ByteArrayUtil.join(keyPrefix, range.end);
+            final KeyRange keyRange = new KeyRange(rangeStart, rangeEnd);
```
alecgrieser, line 176: I'm trying to wrap my head around the `KeyRange`. Looking at `TupleRange::toRange` (which is still used to translate a `KeyRange` into an FDB `Range`), a `KeyRange` where the low endpoint is `RANGE_INCLUSIVE` and the end endpoint is always `RANGE_EXCLUSIVE` will always behave just like a `Range` over those same two keys. But it also behaves just like a `TupleRange` with packed elements, so for instance, something like `new KeyRange(bytes, EndpointType.RANGE_INCLUSIVE, bytes, EndpointType.RANGE_INCLUSIVE)` represents all keys "strictly" prefixed by the range (that is, excluding keys prefixed by `bytes + '\xff'`).

That feels fair enough.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -125,9 +125,7 @@ private CompletableFuture<Void> buildMultiTargetIndex(@Nonnull SubspaceProvider
         } else {
             final Range range = tupleRange.toRange();
             rangeStart = range.begin;
-            // tupleRange has an inclusive high endpoint, so end isn't a valid tuple.
-            // But buildRangeOnly needs to convert missing Ranges back to TupleRanges, so round up.
-            rangeEnd = ByteArrayUtil.strinc(range.end);
+            rangeEnd = range.end;
```
alecgrieser, line 128: Setting up this particular test may be a pain, so it may not be worth it, but I think if we wanted to validate that this fixed anything, we could add a test that used non-integer record type keys. Or a test that used a record type key with what would have been problematic value for its record type key, like 255

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -866,6 +867,20 @@ RecordCursor<FDBStoredRecord<M>> scanRecords(@Nullable Tuple low, @Nullable Tupl
...
+     * Scan the records in the database in a key range.
+     *
+     * @param range key range
+     * @param continuation any continuation from a previous scan
+     * @param scanProperties skip, limit and other scan properties
+     *
+     * @return a cursor that will scan everything in the range, picking up at continuation, and honoring the given scan properties
+     */
+    @Nonnull
+    RecordCursor<FDBStoredRecord<M>> scanRecordsKeyRange(@Nonnull KeyRange range,
```
alecgrieser, line 880: Do you think this API should be `INTERNAL`? Most of the use cases I can think of are probably internal, though there may be some use case for APIs that returns a `KeyRange` that the user can hand back, or something.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1192,32 +1194,56 @@ public RecordCursor<FDBStoredRecord<Message>> scanRecords(@Nullable final Tuple
...
+                                                                                         @Nonnull ScanProperties scanProperties) {
+        return scanTypedRecordsInternal(typedSerializer,
+                                        builder -> builder.setRange(range),
+                                        continuation, scanProperties);
+    }
+
+    @Nonnull
+    @SuppressWarnings("PMD.CloseResource")
+    private <M extends Message> RecordCursor<FDBStoredRecord<M>> scanTypedRecordsInternal(@Nonnull RecordSerializer<M> typedSerializer,
+                                                                                          @Nonnull Consumer<KeyValueCursor.Builder> setRange,
```
alecgrieser, line 1226: minor: I was trying to think if there was a better way to structure this than a `Consumer`. The answer may be "no". The best alternatives I could think of would be to either have this take a `KeyValueCursor.Builder` than callers would pre-fill with the range already set, or something more exotic like adding a `.setRange(KeyValueCursor.Builder)` to the `TupleRange` and `KeyValueRange` abstractions, or making `TupleRange` and `KeyValueRange` implement some common interface that could set the low and high endpoints on a `KeyValueCursor.Builder`.

PR #2658, opened 2024-04-12
Continues #2484: Remove usages of `org.apache.commons.lang3.tuple.Pair` et al

alecgrieser:
This builds on top of https://github.com/FoundationDB/fdb-record-layer/pull/2652. It changes the internal uses of `Pair` to the new class, but it leaves external API callers untouched. As such, it doesn't call out the changes in the release notes, as they shouldn't be notable for adopters. The basic approach is that:

* All `Cascades` classes are treated as internal, so I didn't worry about using `Pair` there. Likewise for a few remaining tests
* Most other cases, I was able to validate that `Pair` was only referenced in `private` members or variables
* I was able to remove `Pair`s entirely from a few places. For instance, when finding missing index entries, we were already wrapping any that we found in an `InvalidIndexEntry` abstraction that behaved a lot like a `Pair`. We now perform the wrapping before checking if the index entries exist, and then we don't need to wrap them twice (saving an object creation along the way).
* The case of `PreOrderPruningIterator` is particularly notable, in that it replaces a `Pair<Iterable, Integer>` where the integer represented the position within an `Iterable` with an `Iterator`. For some iterable types, that is basically what the `Iterator` is, though for some others, the `Iterator` should be more efficient.
* The uses in `Bindings` and `DynamicMessageSerializer` may be a bit more controversial. There, I changed the signature from `Pair` to `Map.Entry` in some methods that are generally internal, but that some users _may_ have referenced. But because it's replaced with `Map.Entry`, users can accommodate the new API without having to explicitly reference the new `Pair` classes.

This continues the work on #2484.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GraphExpansion.java
```diff
@@ -206,7 +206,7 @@ public Sealed seal() {
             final var resultPlaceHolders = Lists.newArrayList(deDupPlaceholders);
             final var localPlaceHolderPairs =
                     IntStream.range(0, deDupPlaceholders.size())
-                            .mapToObj(i -> Pair.of(deDupPlaceholders.get(i), i))
+                            .mapToObj(i -> NonnullPair.of(deDupPlaceholders.get(i), i))
```
alecgrieser, line 209: Only used within local variables within this method

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PreOrderPruningIterator.java
```diff
@@ -48,13 +46,11 @@
 public final class PreOrderPruningIterator<T extends TreeLike<T>> implements Iterator<T> {
 
     @Nonnull
-    private final Deque<Pair<Iterable<? extends T>, Integer>> stack;
+    private final Deque<Iterator<? extends  T>> stack;
```
alecgrieser, line 49: The changes here was to convert this from a `Deque` of `Pair<Iterable, Integer>`, where the integer represented the current position within the `Iterable`, to a `Deque` of `Iterators`. Previously, in each iteration, we'd advance the `Integer` in the top `Pair` until we'd exhausted the tree. Now, we advance the `Iterator`.

For array-based `Iterable`s, the `Iterator` is more-or-less a `Pair<Iterable, Integer>`, and advancing the iterator works by getting the value in the array by position and then incrementing the value, so the implementations are basically the same. For linked list-based `Iterable`s, the `Iterator` effectively maintains a pointer to the correct node within the list, so this implementation would be a performance improvement (from quadratic to linear performance per node child size).

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBIndexOutputTest.java
```diff
@@ -93,11 +102,7 @@ public void testCopyBytesReadAheadPipelineShortBuffer() throws Exception {
         IndexInput blocks = directory.openChecksumInput(FILE_NAME, IOContext.READONCE);
         output = new FDBIndexOutput(FILE_NAME_TWO, directory);
         output.setExpectedBytes((PrefetchableBufferedChecksumIndexInput) blocks, 2 * FDBDirectory.DEFAULT_BLOCK_SIZE);
-        assertEquals("[(1,0), (1,1)]", directoryCacheToString());
-    }
-
-    private String directoryCacheToString() {
-        return directory.getBlockCache().asMap().keySet().stream().sorted().collect(Collectors.toList()).toString();
+        assertEquals(List.of(Pair.of(1L, 0), Pair.of(1L, 1)), directoryCacheKeys());
```
alecgrieser, line 105: Replaced the string-based assert with object comparison. The `Pair` does have a `toString` method that we could switch to, but this seemed like a more sensible assert anyway.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateComputationRule.java
```diff
@@ -23,7 +23,7 @@
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.BindingMatcher;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
-import org.apache.commons.lang3.tuple.Pair;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 26: This is used as part of the type specification of the `RESULT` arg of an `AbstractQueryPredicateRule`. Given the way these rules are basically internal (and experimental), I just made them use the new `Pair`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/DefaultQueryPredicateRuleSet.java
```diff
@@ -27,10 +27,10 @@
 import com.apple.foundationdb.record.query.plan.cascades.predicates.NotPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.OrPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 30: Fall out from `QueryPredicateComputationRule`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateComputationRuleCall.java
```diff
@@ -27,7 +27,7 @@
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 import com.apple.foundationdb.record.query.plan.cascades.values.simplification.AbstractRule;
-import org.apache.commons.lang3.tuple.Pair;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 30: Fall out from `QueryPredicateComputationRule`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateComputationRuleSet.java
```diff
@@ -22,8 +22,8 @@
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 25: Fall out from `QueryPredicateComputationRule`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateWithCnfRuleSet.java
```diff
@@ -25,10 +25,10 @@
 import com.apple.foundationdb.record.query.plan.QueryPlanConstraint;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
 import com.apple.foundationdb.record.query.plan.planning.BooleanPredicateNormalizer;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 28: Fall out from `QueryPredicateComputationRuleCall`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateWithDnfRuleSet.java
```diff
@@ -25,10 +25,10 @@
 import com.apple.foundationdb.record.query.plan.QueryPlanConstraint;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
 import com.apple.foundationdb.record.query.plan.planning.BooleanPredicateNormalizer;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 28: Fall out from `QueryPredicateComputationRule`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/OrderingProperty.java
```diff
@@ -76,6 +76,7 @@
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUpdatePlan;
 import com.apple.foundationdb.record.query.plan.sorting.RecordQueryDamPlan;
 import com.apple.foundationdb.record.query.plan.sorting.RecordQuerySortPlan;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 79: Used with local variables only

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementDistinctUnionRule.java
```diff
@@ -39,11 +39,11 @@
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryUnionPlan;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 42: Used with local variables only

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectDataAccessRule.java
```diff
@@ -26,19 +26,19 @@
...
 import com.apple.foundationdb.record.query.plan.cascades.LinkedIdentitySet;
 import com.apple.foundationdb.record.query.plan.cascades.MatchPartition;
 import com.apple.foundationdb.record.query.plan.cascades.PartialMatch;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifiers;
+import com.apple.foundationdb.record.query.plan.cascades.Reference;
 import com.apple.foundationdb.record.query.plan.cascades.RequestedOrderingConstraint;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.SelectExpression;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.BindingMatcher;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 38: Used with local variables only

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
```diff
@@ -142,12 +142,12 @@ private static BiMap<PhysicalOperator, PPhysicalOperator> getProtoEnumBiMap() {
         }
     }
 
-    private static final Map<Pair<Type.TypeCode, Type.TypeCode>, PhysicalOperator> PROMOTION_MAP;
+    private static final Map<NonnullPair<Type.TypeCode, Type.TypeCode>, PhysicalOperator> PROMOTION_MAP;
```
alecgrieser, line 145: Note only `private` access

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -89,11 +87,11 @@ public abstract class RelOpValue extends AbstractValue implements BooleanValue {
     private final Iterable<? extends Value> children;
 
     @Nonnull
-    private static final Supplier<Map<Pair<Comparisons.Type, Type.TypeCode>, UnaryPhysicalOperator>> unaryOperatorMapSupplier =
+    private static final Supplier<Map<UnaryOperatorSignature, UnaryPhysicalOperator>> unaryOperatorMapSupplier =
```
alecgrieser, line 90: Only `private` access, but I went with a real class here mainly to make the `Triple` replacement work in an analogous way

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/VariadicFunctionValue.java
```diff
@@ -73,7 +74,7 @@ public class VariadicFunctionValue extends AbstractValue {
     private final List<Value> children;
 
     @Nonnull
-    private static final Supplier<Map<Pair<ComparisonFunction, TypeCode>, PhysicalOperator>> operatorMapSupplier =
+    private static final Supplier<Map<NonnullPair<ComparisonFunction, TypeCode>, PhysicalOperator>> operatorMapSupplier =
```
alecgrieser, line 77: Only `private` access

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/WindowedValue.java
```diff
@@ -88,15 +88,15 @@ protected Iterable<? extends Value> computeChildren() {
     }
 
     @Nonnull
-    protected Pair<List<Value>, List<Value>> splitNewChildren(@Nonnull final Iterable<? extends Value> newChildren) {
+    protected NonnullPair<List<Value>, List<Value>> splitNewChildren(@Nonnull final Iterable<? extends Value> newChildren) {
```
alecgrieser, line 91: Only `protected` access

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
```diff
@@ -34,11 +34,11 @@
 import com.apple.foundationdb.record.query.plan.cascades.predicates.simplification.QueryPredicateComputationRuleCall;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.simplification.QueryPredicateComputationRuleSet;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 37: This does expose the `Pair` in a nominally public API. This should only be used internally, given that it's pretty deep in the planner, though the class isn't marked as such

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationRule.java
```diff
@@ -23,7 +23,7 @@
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.BindingMatcher;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
-import org.apache.commons.lang3.tuple.Pair;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 26: Used in the type parameter definition. Given that this is a planner rule, it seemed like it was okay to use the internal `Pair` class.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationRuleCall.java
```diff
@@ -25,7 +25,7 @@
 import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.PlannerBindings;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
-import org.apache.commons.lang3.tuple.Pair;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 28: Fall out of `ValueComputationRule`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationRuleSet.java
```diff
@@ -22,8 +22,8 @@
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 25: Fall out of `ValueComputationRuleCall`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/BlockingInAsyncDetectionTest.java
```diff
@@ -140,8 +137,6 @@ private void callAsyncBlocking(FDBDatabase database, boolean shouldTimeOut) {
             } else {
                 throw e;
             }
-        } finally {
-            database.setAsyncToSyncTimeout(existingTimeouts);
```
alecgrieser, line 144: No longer necessary because each test gets its own `database`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreIndexTest.java
```diff
@@ -67,6 +67,7 @@
 import com.apple.foundationdb.record.provider.foundationdb.indexes.InvalidIndexEntry;
 import com.apple.foundationdb.record.query.IndexQueryabilityFilter;
 import com.apple.foundationdb.record.query.expressions.Query;
+import com.apple.foundationdb.record.util.pair.NonnullPair;
```
alecgrieser, line 70: Only used locally; also a test class.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneAnalyzerRegistryImpl.java
```diff
@@ -25,7 +25,7 @@
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.MetaDataException;
 import com.apple.foundationdb.record.util.ServiceLoaderProvider;
-import org.apache.commons.lang3.tuple.Pair;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 28: Return type of a `private` method

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -45,6 +45,8 @@
 import com.apple.foundationdb.record.lucene.codec.PrefetchableBufferedChecksumIndexInput;
 import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.util.pair.ComparablePair;
+import com.apple.foundationdb.record.util.pair.NonnullPair;
```
alecgrieser, line 49: Only used by `private` and package private methods

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectorySharedCache.java
```diff
@@ -21,10 +21,10 @@
 package com.apple.foundationdb.record.lucene.directory;
 
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 24: Only used by `private` members

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FieldInfosStorage.java
```diff
@@ -24,10 +24,10 @@
 import com.apple.foundationdb.record.lucene.LuceneFieldInfosProto;
 import com.apple.foundationdb.record.lucene.LuceneLogMessageKeys;
 import com.apple.foundationdb.record.lucene.codec.LazyOpener;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 27: Fall out from `FDBDirectory`. Not part of any public API here

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -52,6 +52,7 @@
 import com.apple.foundationdb.record.query.plan.PlannableIndexTypes;
 import com.apple.foundationdb.record.query.plan.cascades.CascadesPlanner;
 import com.apple.foundationdb.record.query.plan.plans.RecordQueryPlan;
+import com.apple.foundationdb.record.util.pair.Pair;
```
alecgrieser, line 55: Only used for local variables. Also a test class

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PreOrderPruningIterator.java
hatyo: That's a good improvement. By the way, at least in Cascades, we almost always `ImmutableList`, which is backed by an array, so the cost of `Iterables.get` is constant. Nevertheless, I think using an `iterator()` is more elegant that manually keeping track of the pointer.

hatyo (approved):
LGTM.

PR #2660, opened 2024-04-15
Query plan with IN as Union fails to display plan graph in CloudKit console

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/Bindings.java
```diff
@@ -70,9 +70,20 @@ public String bindingName(@Nonnull String suffix) {
             return value + suffix;
         }
 
-        public String identifier(@Nonnull String bindingName) {
-            Verify.verify(bindingName.startsWith(value));
-            return bindingName.substring(value.length());
+        public String identifier(@Nonnull final String bindingName) {
```
normen662, line 73: So we want the old identifier method to also stay there since we want to allow callers to explicitly state the want to resolve an alias to an e.g. `CORRELATION`. The old `identifier()` should be renamed to let's say `typedIdentifier()` and it should stay an instance method. This new `identifier()` method should become static. All callers should call `typedIdentifier()`, except the one you have where we don't know what it is.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryInUnionPlan.java
```diff
@@ -205,7 +205,7 @@ public Set<CorrelationIdentifier> getCorrelatedTo() {
 
         final var inAliases = getInSources()
                 .stream()
-                .map(inSource -> CorrelationIdentifier.of(Bindings.Internal.CORRELATION.identifier(inSource.getBindingName())))
+                .map(inSource -> CorrelationIdentifier.of(Bindings.Internal.identifier(inSource.getBindingName())))
```
pengpeng-lu, line 208: only this one is using the new identifier

PR #2661, opened 2024-04-16
WIP - Replace child QOV with its Value.

hatyo:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/OrderingProperty.java
```diff
@@ -234,9 +236,25 @@ public Ordering visitIntersectionOnKeyExpressionPlan(@Nonnull final RecordQueryI
...
+            final var childCorrelation = childQuantifier.getAlias();
+
+            // replace occurrences of child QOV with the child value itself.
+            final var childValue = RecordConstructorValue.ofUnnamed(childQuantifier.getFlowedValues().stream().map(
+                    flowedValue -> flowedValue.replaceLeavesMaybe(value -> {
+                        if (value instanceof QuantifiedObjectValue && ((QuantifiedObjectValue)value).getAlias().equals(childCorrelation)) {
+                            return QuantifiedObjectValue.of(Quantifier.current(), value.getResultType());
+                        }
+                        return value;
+                    }).orElseThrow()).collect(ImmutableList.toImmutableList()));
```
hatyo, line 249: This is in order to avoid infinite replacement. An example is we want to replace a leaf QOV that is `Q8` with something like: `RCV(FV(Q8, 'I'))`, however this is an infinite replacement because the LHS appears in the RHS. To avoid this, we first replace `Q8` with `_current`, _then_ do the intended replacement. In other words:
Instead of `Q8 ---> RCV(FV(Q8, 'I'))` we do it in two steps:
`Q8 ---> _current => RCV(FV(_current, 'I'))`
`Q8 ---> RCV(FV(_current, 'I'))`
To make this work, we change the equivalence map in the pull up from `Q8 -> _current` to `_current -> _current`.

hatyo:
Superseded by a better fix from @normen662 https://github.com/FoundationDB/fdb-record-layer/pull/2662

PR #2669, opened 2024-04-19
FIXES #2668: Graceful Degradation breaks when index is being built

### LINKED ISSUE #2668: Graceful Degradation breaks when index is being built
`LucenePartitioner` assumes that the old record will always be indexed, but if the index is in `WriteOnly` mode that might not be the case, thus updating records while the index is being built may fail because they don’t have partition metadata yet.

wshaib-apple:
- During index build, it is possible for operations like record inserts/updates to fail or cause incorrect partition metadata to be stored. This update eliminates the failure cause in such scenarios, and reduces the probability of drifting document count in partition metadata. A subsequent update should audit and reconcile the doc count for the unlikely cases of mismatch.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -92,6 +94,13 @@
  */
 @API(API.Status.EXPERIMENTAL)
 public class LucenePartitioner {
+    /**
+     * partition error reasons.
+     */
+    public enum PartitionFailureReason {
```
wshaib-apple, line 100: TODO evaluate if still needed. This was for when there were potentially multiple situations to report/handle.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -145,6 +153,115 @@ void luceneOnlineIndexingTestSimple() {
...
+
+        final LuceneIndexTestValidator luceneIndexTestValidator = new LuceneIndexTestValidator(() -> openContext(contextProps), context -> {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            return recordStore;
+        });
+
+        final List<LucenePartitionInfoProto.LucenePartitionInfo> partitionMeta = luceneIndexTestValidator.getPartitionMeta(index, Tuple.from(1L));
+        assertTrue(partitionMeta.isEmpty());
+
+        // update a record; no partition information exist yet
```
ScottDugas, line 206: You also need to test updating a record that has already been inserted into the index, i.e. updating 1623L a second time.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -145,6 +153,115 @@ void luceneOnlineIndexingTestSimple() {
...
+        // this will succeed
+        try (final FDBRecordContext context = openContext(contextProps)) {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            assertTrue(recordStore.isIndexWriteOnly(index));
+            recordStore.saveRecord(createComplexDocument(1623L, ENGINEER_JOKE, 1, startTime));
```
ScottDugas, line 211: It would be good to replace the document ids with variables

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -145,6 +153,115 @@ void luceneOnlineIndexingTestSimple() {
...
+        // do not index any records and leave the index writeOnly
+        try (final FDBRecordContext context = openContext(contextProps)) {
+            rebuildIndexMetaData(context, COMPLEX_DOC, index);
+            final RuntimeException stopBuildException = new RuntimeException("stop build");
+            try (OnlineIndexer indexBuilder = OnlineIndexer.newBuilder()
+                    .setRecordStore(recordStore)
+                    .setIndex(index)
+                    .setInitialLimit(1)
+                    .setConfigLoader(config -> {
+                        throw stopBuildException;
```
ScottDugas, line 189: Another test that would probably be good to add, and be a bit more functional, would be to basically have a config loader that opens a context, and updates all the records in the index.
If you have it update them with some sort of counter, you can then search at the end once the index is readable and validate that they all have the right value. 
I think you'll also need to call `setMaxLimit(1)` to ensure that you modify the records between each record being updated.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -145,6 +153,115 @@ void luceneOnlineIndexingTestSimple() {
...
+                Map.of(
+                        Tuple.from(1), Map.of(
+                                Tuple.from(1, 899L), Tuple.from(startTime + 3_000),
+                                Tuple.from(1, 2222L), Tuple.from(startTime + 2_000),
+                                Tuple.from(1, 1547L), Tuple.from(startTime + 1_000),
+                                Tuple.from(1, 1623L), Tuple.from(startTime)
+                        )
+                ),
+                2,
+                "*:*");
```
ScottDugas, line 262: This test won't catch if either flow results in not-updating the index, it just makes sure that all the values are there. It would be good to do searches that verify that the index has the updated values.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -589,15 +557,53 @@ private CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> getOrCre
...
     @Nonnull
-    private CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> getPartitionInfoOrFail(@Nonnull Tuple groupingKey, @Nonnull Tuple partitioningKey) {
-        return assignPartitionInternal(groupingKey, partitioningKey, false);
+    <M extends Message> CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> tryGetPartitionInfo(
+            @Nonnull FDBIndexableRecord<M> record,
+            @Nonnull Tuple groupingKey) {
+        if (!isPartitioningEnabled()) {
+            return CompletableFuture.completedFuture(null);
+        }
+        return assignPartitionInternal(groupingKey, toPartitionKey(record), false)
```
ScottDugas, line 574: Instead of catching an exception here, `assignPartitionInternal` should take an `enum` for whether it should create, or we should have different methods composed so that this returns `null`.

ScottDugas (changes requested):

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -456,25 +465,63 @@ public <M extends Message> CompletableFuture<Void> update(@Nullable FDBIndexable
...
             }
-            return null;
-        })).collect(Collectors.toList())).thenCompose(ignored ->
+        }).collect(Collectors.toList())).thenCompose(ignored ->
                 // update new
-                AsyncUtil.whenAll(newRecordFields.entrySet().stream().map(entry -> partitioner.addToAndSavePartitionMetadata(Objects.requireNonNull(newRecord), entry.getKey()).thenApply(partitionId -> {
+                AsyncUtil.whenAll(newRecordFields.entrySet().stream().map(entry -> {
                     try {
-                        writeDocument(entry.getValue(), entry.getKey(), partitionId, newRecord.getPrimaryKey());
+                        return tryDelete(Objects.requireNonNull(newRecord), entry.getKey(), true).thenCompose(countDeleted ->
```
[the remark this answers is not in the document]
wshaib-apple: Yes, that's the scenario for `tryDelete()` . Let me know if you have a preference (splitting into two methods vs. some documentation) and I will adjust.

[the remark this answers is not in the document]
ScottDugas: I think two methods would be beneficial, but probably some documentation would be good either way.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1723,12 +1723,15 @@ void simpleCrossPartitionQuery() {
     }
 
     static Stream<Arguments> findStartingPartitionTest() {
-        return Stream.of(true, false).map(Arguments::of);
+        Random r = ThreadLocalRandom.current();
```
ScottDugas, line 1726: It would be better to use https://github.com/FoundationDB/fdb-record-layer/blob/f86e1dc34ae52faf04aa7cf7572bd26e49a3b4c5/fdb-extensions/src/test/java/com/apple/test/RandomizedTestUtils.java#L50
That limits injecting randomness to when it is requested, and not during PRB. This can reduce flakiness to PRB.

ScottDugas (approved):

PR #2683, opened 2024-04-25
Resolves #2682: Support protocol version 7.3 in ClientLogEvents

### LINKED ISSUE #2682: Support protocol version 7.3 in ClientLogEvents
The `ClientLogEvents` class supports reading client latency events from the FDB system keyspace. See: https://apple.github.io/foundationdb/transaction-profiler-analyzer.html There's also a Python script packaged with FDB that does this: https://github.com/apple/foundationdb/blob/main/contrib/transaction_profiling_analyzer/transaction_profiling_analyzer.py

The FDB protocol version is encoded into the keys, and FDB can change the format with different protocol versions. The 7.3 protocol version did not make any changes since 7.2, so we should be able to just add the new protocol version to the allow list.

alecgrieser:
The client log event format has not been updated since 7.2, so we are safe to update the log event known protocol versions to include FDB 7.3. There's not a great way of testing this without doing a more thorough job of updating our infrastructure to use a 7.3 server version and then run a few one off tests we have that try and parse these values. However, the Python script this is based on has been updated for 7.3 and it didn't need to make any changes from 7.2: https://github.com/apple/foundationdb/commit/321b77d8099f6ab3ec7c277139c0158abd823124

This resolves #2682.


ScottDugas:
Filed this for the test: https://github.com/FoundationDB/fdb-record-layer/issues/2684

ScottDugas (approved):

PR #2686, opened 2024-04-25
RESOLVES #2685: Add logging information to the runner used when repar…

### LINKED ISSUE #2685: Add logging information to the runner used when repartitioning

wshaib-apple:
…titioning

(Add logging information to the runner used when repartitioning)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -428,7 +429,7 @@ private static CompletableFuture<Void> rebalancePartitions(final FDBDatabaseRunn
                                                         continuation.set(newContinuation);
                                                         return true;
                                                     }
-                                                })))), state.context.getExecutor());
+                                                }))), repartitioningLogMessages.logMessages), state.context.getExecutor());
```
jjezra, line 432: I'm not sure which version of `runAsync` will be triggered by adding the `repartitioningLogMessages.logMessages)`. Would it be one that expects a lambda that provides an additional log items (as an even list of objects)?

jjezra (approved):

ScottDugas (approved):

PR #2688, opened 2024-04-26
Remove `MatchOrCompensateQuantifiedObjectValueRule` rule.

hatyo:
This is to avoid issues related to planning `order by` clause (in)correctly at the moment. Once we fix this issue, we will reactivate that rule.


normen662 (dismissed):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueSimplificationTest.java
```diff
@@ -367,42 +365,6 @@ void testDeriveOrderingValues() {
         Assertions.assertEquals(expectedResult, orderingValues);
     }
 
-    @Test
```
normen662, line 370: Can you leave the test in here and just remove the `@Test` and add a TODO to put it back in?

normen662 (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueSimplificationTest.java
normen662: Or leave this test in here but use a specifically constructed rules set including that rule just for this test case.

normen662 (approved):

PR #2690, opened 2024-04-29
Resolves #2689: Decompressing deserializer does not validate length of decompressed data

### LINKED ISSUE #2689: Decompressing deserializer does not validate length of decompressed data
For compressed records, we have a format that looks like:

```
(flags byte) + (version byte) + (decompressed length 4 bytes) + (compressed blob)
```

The `compressed blob` is ultimately generated by running zlib to compress the data, though that is wrapped using the `java.util.zip` API. We use the decompressed length so that when we decompress, the basic steps are:

1. Parse the 4 byte length
2. Allocate a byte array of the appropriate length
3. De-compress the blob into the pre-allocated array

This allows us to avoid needing to allocate multiple arrays for the size, but it also means that if the data somehow get corrupted in a way that makes the size incorrect (e.g., the decompressed length gets corrupted or the Huffman tree in the compressed blob is modified), then we can end up passing along incomplete/incorrect data to the underlying deserializer.

We should check that the decompressed size exactly matches the expected length. This means adding checks that:

1. We fully consume the `compressed blob`. The default behavior of the inflater here is to consume as much of the blob as possible until it has filled up the array. So, if we fill up the byte array without having consumed the entire blob, then we know the decompressed record exceeds its expected size
2. We fully fill the data array. If the underlying compressed data are too small, we can end up fully consuming the compressed blob but still leave extra bytes left over and the end of the data array.

alecgrieser:
This adds a check to the `decompress` method to validate that the decompressed data coming from a transformed record serializer exactly matches the size encoded in the serialized data frame. This ensures that if there is some kind of error that results in the compressed blob either deserializing into something bigger or smaller than the original size, we notice and get a error right away (rather than, say, a protobuf deserialization error).

This resolves #2689.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -33,6 +33,7 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** The `TransformedRecordSerializer` now validates that compressed data decompresses to the expected size [(Issue #2689)](https://github.com/FoundationDB/fdb-record-layer/issues/2689)
+* **Feature** The `TransformedRecordSerializer` now includes checksums in compressed data that is validated during decompression [(Issue #2689)](https://github.com/FoundationDB/fdb-record-layer/issues/2691)
 * **Feature** Updated the set of known protocol versions for client log event parsing to include FDB 7.3 [(Issue #2682)](https://github.com/FoundationDB/fdb-record-layer/issues/2682)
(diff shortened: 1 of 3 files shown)


MMcM (approved):

docs/ReleaseNotes.md
```diff
@@ -31,6 +31,9 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 * **Feature** Feature 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Add logging information to the runner used when repartitioning [(Issue #2685)](https://github.com/FoundationDB/fdb-record-layer/issues/2685)
 * **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Feature** Feature 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
```
MMcM, line 34: I think this is harmless, in that the build script will clean it up.

MMcM (approved):

PR #2694, opened 2024-04-30
Lucene lock: treat close as a potential recovery path

jjezra:
Do not assume that the agility context is still active.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
@@ -178,7 +178,7 @@ public final class FDBDirectoryLockFactory extends LockFactory {
                                         aContext.ensureActive().clear(fileLockKey);
                                         logSelf(isRecovery ? "FileLock: Cleared in Recovery path" : "FileLock: Cleared");
                                     } else if (! isRecovery) {
-                                        throw new AlreadyClosedException("FileLock: Expected lock during close.This=" + this + " existingUuid=" + existingUuid); // The string append methods should handle null arguments.
+                                        throw new AlreadyClosedException("FileLock: Expected to be locked during close.This=" + this + " existingUuid=" + existingUuid); // The string append methods should handle null arguments.
                                     }
                                 }
                             });
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
@@ -82,18 +82,6 @@ public final class FDBDirectoryLockFactory extends LockFactory {
         }
 
         @Override
-        public void close() {
-            fileLockClear();
-            flushAndClose();
-        }
-
-        private void flushAndClose() {
-            // always flush before declaring this object as closed. Else agilityContext may fail to commit, while this lock may skip retry
-            agilityContext.flush();
-            closed = true;
-        }
-
-        @Override
         public void ensureValid() {
             // ... and implement heartbeat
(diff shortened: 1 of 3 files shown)


jjezra:
Fixed conflict

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -222,7 +192,26 @@ protected void fileLockClearIfLocked() {
                 agilityContext.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FILE_LOCK_CLEAR,
                         agilityContext.apply(fileLockFunc));
             }
-            flushAndClose();
+            // always flush before declaring this object as closed. Else agilityContext may fail to commit, while this lock may skip retry
+            agilityContext.flush();
+            closed = true;
+        }
+
+        protected void fileLockClearIfLocked() {
```
ScottDugas, line 200: Can you add a test to `FDBDirectoryLockTest` or higher that fails without this change, and succeeds with it.
Based on the description, I think you should be able to get a better reproduction of the problem by adding to `FDBDirectoryTest` and creating an `FDBDirectory` with an agile context, and simulating what you think is happening. Potentially closing the `AgileContext` itself rather than having a timeout occur.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockTest.java
```diff
@@ -87,4 +89,29 @@ void testFileLock(boolean useAgile) throws IOException {
...
+    @ParameterizedTest
+    @CsvSource({"true,true", "true,false", "false,true", "false,false"})
+    void testFileLockClose(boolean useAgile, boolean abortAgilityContext) throws IOException {
+        try (FDBRecordContext context = fdb.openContext()) {
+            AgilityContext agilityContext =
+                    useAgile ?
+                    AgilityContext.agile(context, 1000, 100_0000) :
+                    AgilityContext.nonAgile(context);
+
+            FDBDirectory directory = new FDBDirectory(subspace, null, null, null, true, agilityContext);
```
ScottDugas, line 102: You should also close the directory after closing the lock.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockTest.java
```diff
@@ -87,4 +89,30 @@ void testFileLock(boolean useAgile) throws IOException {
...
+            if (abortAgilityContext) {
+                agilityContext.abortAndClose();
+                assertTrue(agilityContext.isClosed());
+            } else {
+                assertFalse(agilityContext.isClosed());
+            }
+            lock1.close();
+            directory.close();
+            agilityContext.flushAndClose();
+            context.commit();
```
ScottDugas, line 115: It would probably be good to validate that you can obtain a new lock, but I don't think this PR changes the risk here; this test covers that, as-is.

ScottDugas (approved):

PR #2697, opened 2024-04-30
 Lucene: merges should run in default transaction priority

jjezra:
Since user operations are waiting for merge to finish over the “file lock” semaphore, the merge operation should not be running in BATCH priority but DEFAULT.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
@@ -112,4 +112,8 @@ public final class LuceneRecordContextProperties {
      * assume that merge will not be performed during IO but in a background process.
      */
     public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_CONCURRENT_MERGE_SCHEDULER = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useConcurrentMergeScheduler", false);
+    /**
+     * Use "default transaction priority" during merge. The default of this prop is "true" because merge over multiple transactions may be holding back user operations.
+     */
+    public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_DEFAULT_PRIORITY_DURING_MERGE = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useDefaultPriorityDuringMerge", false);
 }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
@@ -113,7 +113,7 @@ public final class LuceneRecordContextProperties {
      */
     public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_CONCURRENT_MERGE_SCHEDULER = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useConcurrentMergeScheduler", false);
     /**
-     * Use "default transaction priority" during merge. The default of this prop is "true" because merge over multiple transactions may be holding back user operations.
+     * Use "default transaction priority" during merge. The default of this prop is {@code true} because merge over multiple transactions may be preventing user operations.
      */
-    public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_DEFAULT_PRIORITY_DURING_MERGE = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useDefaultPriorityDuringMerge", false);
+    public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_DEFAULT_PRIORITY_DURING_MERGE = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useDefaultPriorityDuringMerge", true);
 }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
@@ -51,8 +51,12 @@ public interface AgilityContext {
         return new NonAgile(callerContext);
     }
 
+    static AgilityContext agile(FDBRecordContext callerContext, @Nullable FDBRecordContextConfig.Builder contextBuilder, final long timeQuotaMillis, final long sizeQuotaBytes) {
+        return new Agile(callerContext, contextBuilder, timeQuotaMillis, sizeQuotaBytes);
+    }
+
     static AgilityContext agile(FDBRecordContext callerContext, final long timeQuotaMillis, final long sizeQuotaBytes) {
-        return new Agile(callerContext, timeQuotaMillis, sizeQuotaBytes);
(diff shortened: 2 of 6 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -194,9 +198,9 @@ class Agile implements AgilityContext {
         private long prevCommitCheckTime;
         private boolean closed = false;
 
-        protected Agile(FDBRecordContext callerContext, final long timeQuotaMillis, final long sizeQuotaBytes) {
+        protected Agile(FDBRecordContext callerContext, @Nullable FDBRecordContextConfig.Builder contextBuilder, final long timeQuotaMillis, final long sizeQuotaBytes) {
             this.callerContext = callerContext;
```
ScottDugas, line 202: Why are we still getting a `callerContext` if we are getting a `contextBuilder`?
It looks like the `callerContext` provides other things that you can't get from the `FDBRecordContextConfig`... Probably would be good to figure out how to clean that up so that this no longer references the `callerContext` after creation, that seems like a risk, but I don't think we want it in this PR.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
```diff
@@ -704,6 +706,26 @@ void testAutoCommitVersionStampInnerSleepUseApply() {
         }
     }
 
+
+    @Test
+    void luceneTransactionPriorityVerificationTest() {
```
ScottDugas, line 711: It seems like it would be good to parameterize this test with the `userPriority` being set to `DEFAULT` and `BATCH` and the `agilePriority` being set to inherity or use `BATCH`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -319,7 +321,14 @@ private AgilityContext getAgilityContext(boolean useAgilityContext) {
             sizeQuotaBytes =  Objects.requireNonNullElse(state.context.getPropertyStorage().getPropertyValue(LuceneRecordContextProperties.LUCENE_AGILE_COMMIT_SIZE_QUOTA), 900_000);
             deferredControl.setSizeQuotaBytes(sizeQuotaBytes);
         }
-        return AgilityContext.agile(state.context, timeQuotaMillis, sizeQuotaBytes);
```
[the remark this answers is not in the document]
ScottDugas: Wouldn't it be as simple as adding a second boolean to this method?
Having the batch priority for the other operations would help us avoid starting a merge if the system is under load.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -112,4 +112,8 @@ public final class LuceneRecordContextProperties {
      * assume that merge will not be performed during IO but in a background process.
      */
     public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_CONCURRENT_MERGE_SCHEDULER = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useConcurrentMergeScheduler", false);
+    /**
+     * Use "default transaction priority" during merge. The default of this prop is "true" because merge over multiple transactions may be holding back user operations.
+     */
+    public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_DEFAULT_PRIORITY_DURING_MERGE = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useDefaultPriorityDuringMerge", false);
```
ScottDugas, line 118: The comment says the default is `true`, which I think is right.
```suggestion
    public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_DEFAULT_PRIORITY_DURING_MERGE = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useDefaultPriorityDuringMerge", true);
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
```diff
@@ -112,4 +112,8 @@ public final class LuceneRecordContextProperties {
      * assume that merge will not be performed during IO but in a background process.
      */
     public static final RecordLayerPropertyKey<Boolean> LUCENE_USE_CONCURRENT_MERGE_SCHEDULER = RecordLayerPropertyKey.booleanPropertyKey("com.apple.foundationdb.record.lucene.useConcurrentMergeScheduler", false);
+    /**
+     * Use "default transaction priority" during merge. The default of this prop is "true" because merge over multiple transactions may be holding back user operations.
```
ScottDugas, line 116:
```suggestion
     * Use "default transaction priority" during merge. The default of this prop is {@code true} because merge over multiple transactions may be preventing user operations.
```

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
jjezra: We need the `callerContext` for stats and `getCallerContext()`. The rational is that the agility context vanishes within the operation, but some operations need the one that is visible to the caller.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
jjezra: I've simplified a bit. Hope it's good enough.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/AgilityContextTest.java
```diff
@@ -707,12 +707,13 @@ void testAutoCommitVersionStampInnerSleepUseApply() {
...
 
 
-    @Test
-    void luceneTransactionPriorityVerificationTest() {
+    @ParameterizedTest
+    @BooleanSource
+    void luceneTransactionPriorityVerificationTest(boolean usePriorityBatch) {
         // Assert the expected priorities for user context and agility context
         try (FDBRecordContext userContext = openContext()) {
             final FDBTransactionPriority userPriority = userContext.getPriority();
```
ScottDugas, line 715: Isn't `userPriority` always going to be `DEFAULT`?
I don't think there's much risk of messing this up, but the goal is that when you create an `AgileContext` with a callerContext that has `BATCH` priority you can upgrade to `DEFAULT`. That scenario is not covered here.

ScottDugas (approved):

PR #2698, opened 2024-05-01
Fix random arguments of LuceneIndexTest.findStartingPartitionTest

ScottDugas:


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1723,9 +1723,9 @@ void simpleCrossPartitionQuery() {
     }
 
     static Stream<Arguments> findStartingPartitionTest() {
-        return Stream.of(true, false).flatMap(isSynthetic ->
-                Stream.concat(Stream.of(1714058544895L), RandomizedTestUtils.randomArguments(random -> Arguments.of(random.nextLong())))
```
ScottDugas, line 1727: I'm pretty sure this wasn't a compile time error, because `Arguments.of` takes `Object`s, so it was creating a `Stream<Object>` instead of a `Stream<Long>`

wshaib-apple (approved):

PR #2706, opened 2024-05-08
Fixes #2705: Permuted index can mis-assign fields in IndexKeyValueToPartialRecord

### LINKED ISSUE #2705: Permuted index can mis-assign fields in IndexKeyValueToPartialRecord
There appears to be a bug in the way that the `IndexKeyValueToPartialRecord` is constructed for permuted min and max indexes, at least when using the Cascades planner. In particular, there are cases where if you have an index, we on say:

```java
field("x").groupBy(concatenateFields("a", "b", "c")) // Permuted size = 2
```

So that we get index entries like:

```
a, min/max(x), b, c
```

Then the aggregate index plan constructed by the cascades planner looks like:

```
AggregateIndexScan(Index(indexName <,> BY_GROUP) -> [_0: [a: KEY[0], b: KEY[3], c: KEY[4]], _1: [_0: KEY[1]])
```

Which is correct for `a` and for the min/max of `x`, but it's off by one for `b` and `c`. The correct mapping should be:

```
AggregateIndexScan(Index(indexName <,>) -> [_0: [a: KEY[0], b: KEY[2], c: KEY[3]], _1: [_0: KEY[1]])
```

This appears to only happen if the permuted size is >= 2. It looks like most of our tests with the Cascades planner only have a permuted size of 1, which is how it is not being surfaced by the current tests. Also, there may be issues with the old planner (it's a bit unclear), but we definitely see this problem with Cascades.

alecgrieser:
The calculation for the assigning index entries to fields used by the permuted min and max index wasn't quite right. There's a function that takes the fields ordered by their original place in the index (per-permutation) and then constructs a field extractor, meaning it needs to find the position in the entry accounting for the permutation. This worked by:

1. For grouping columns that precede the permuted aggregate, it preserved the index from the original expression. This is correct, as they are unpermuted.
1. For the permuted aggregate, it skips over `permutedCount` columns. This is correct, because the aggregate is moved from the end `permutedColumns` to the left.
1. For grouping columns that follow the permuted aggregate, it was _adding_ the `permutedCount`. This is _incorrect_, as those columns are moved `groupedCount` columns to the right, where `groupedCount` is equal to the number of aggregated columns. (Note that this value is actually independent from `permutedCount`.) So this should add `groupedCount` instead of `permutedCount`

In the existing tests that used this part of the planner, we only ever had one (1) permuted column and one (1) aggregate column, so the error wasn't hit by our existing tests. I've added a test where `groupedCount` > `permutedCount`, and sure enough, this exposed the issue.

This fixes #2705.


pengpeng-lu (approved):
👍 I really like your summary in this PR. It's a very clear explanation! Could you add it to the code comment before you merge the PR? Thanks!

MMcM (approved):

alecgrieser:
Opened up #2711 to commit the summary comment from this PR

PR #2711, opened 2024-05-09
Commit comment from #2706: Permuted index can mis-assign fields in IndexKeyValueToPartialRecord

alecgrieser:
This commits a version of the summary comment from #2706 into the code, as requested during code review: https://github.com/FoundationDB/fdb-record-layer/pull/2706#pullrequestreview-2046477132

There are some slight differences, in that the original summary column attempted to explain what the code was trying to do and how it was failing, whereas the comment comitted here only contains information about what the code _currently_ does (which, is also what it's supposed to do, or at least that's the idea).


pengpeng-lu (approved):
thanks for doing this! This is much clearer now.

PR #2730, opened 2024-05-18
Resolves #2747: Validate filelock on every transaction to reduce ttl

### LINKED ISSUE #2747: Validate filelock on every transaction to reduce ttl
Validate filelock on every transaction to reduce ttl

Right now the ttl is 10 minutes, making the consequences of an abandoned lock pretty impactful. If we change the FileLock to validate on every transaction we can change the ttl to be just long enough to get it to commit the next transaction (e.g. 20 seconds).

This could potentially result in additional writes/commits to the system if a lot of the transactions in the AgilityContext during a merge are not doing writes otherwise.

todor-m:
Resolves #2747 by validating and updating the filelock on every transaction commit. This allows a much shorter timeout of around 20 seconds instead of 10 minutes.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
@@ -250,7 +250,7 @@ public interface AgilityContext {
                     ensureOpen();
                     final FDBRecordContextConfig contextConfig = contextConfigBuilder.build();
                     currentContext = database.openContext(contextConfig);
-                    addCommitCheck(currentContext, commitCheck);
+                    addCommitCheckToContext(currentContext, commitCheck);
                     creationTime = now();
                     prevCommitCheckTime = creationTime;
                     currentWriteSize = 0;
@@ -258,7 +258,7 @@ public interface AgilityContext {
             }
         }
 
-        private static void addCommitCheck(final FDBRecordContext commitCheckContext, @Nullable final Function<FDBRecordContext, CompletableFuture<Void>> commitCheck) {
+        private static void addCommitCheckToContext(final FDBRecordContext commitCheckContext, @Nullable final Function<FDBRecordContext, CompletableFuture<Void>> commitCheck) {
             if (commitCheck != null) {
                 commitCheckContext.addCommitCheck(() -> commitCheck.apply(commitCheckContext));
             }
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
@@ -171,6 +171,14 @@ public interface AgilityContext {
     }
 
     /**
+     * The {@code commitCheck} callback is expected to use the argument context but should never call other
+     * agility context functions as it may cause a deadlock.
+     *
+     * @param commitCheck callback
+     */
+    void setCommitCheck(Function<FDBRecordContext, CompletableFuture<Void>> commitCheck);
+
+    /**
      * A floating window (agile) context - create sub contexts and commit them as they reach their time/size quota.
      */
     class Agile implements AgilityContext {
@@ -197,6 +205,7 @@ public interface AgilityContext {
         private boolean committingNow = false;
         private long prevCommitCheckTime;
         private boolean closed = false;
+        private Function<FDBRecordContext, CompletableFuture<Void>> commitCheck;
 
         protected Agile(FDBRecordContext callerContext, @Nullable FDBRecordContextConfig.Builder contextBuilder, final long timeQuotaMillis, final long sizeQuotaBytes) {
             this.callerContext = callerContext;
@@ -220,6 +229,11 @@ public interface AgilityContext {
         }
 
         @Override
+        public void setCommitCheck(final Function<FDBRecordContext, CompletableFuture<Void>> commitCheck) {
+            this.commitCheck = commitCheck;
+        }
+
+        @Override
         @Nonnull
         public FDBRecordContext getCallerContext() {
             return callerContext;
@@ -234,12 +248,19 @@ public interface AgilityContext {
             synchronized (createLockSync) {
                 if (currentContext == null) {
                     ensureOpen();
-                    FDBRecordContextConfig contextConfig = contextConfigBuilder.build();
+                    final FDBRecordContextConfig contextConfig = contextConfigBuilder.build();
                     currentContext = database.openContext(contextConfig);
+                    addCommitCheck(currentContext, commitCheck);
                     creationTime = now();
                     prevCommitCheckTime = creationTime;
                     currentWriteSize = 0;
                 }
+            }
+        }
+
+        private static void addCommitCheck(final FDBRecordContext commitCheckContext, @Nullable final Function<FDBRecordContext, CompletableFuture<Void>> commitCheck) {
+            if (commitCheck != null) {
+                commitCheckContext.addCommitCheck(() -> commitCheck.apply(commitCheckContext));
             }
         }
 
@@ -285,9 +306,8 @@ public interface AgilityContext {
                     committingNow = true;
                     final long stamp = lock.writeLock();
 
-                    try {
-                        currentContext.commit();
-                        currentContext.close();
+                    try (FDBRecordContext commitContext = currentContext) {
+                        commitContext.commit();
                     } catch (RuntimeException ex) {
                         closed = true;
                         reportFdbException(ex);
@@ -498,6 +518,11 @@ public interface AgilityContext {
         @Override
         public boolean isClosed() {
             return closed;
+        }
+
+        @Override
+        public void setCommitCheck(final Function<FDBRecordContext, CompletableFuture<Void>> commitCheck) {
+            callerContext.addCommitCheck(() -> commitCheck.apply(callerContext));
         }
     }
 
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
@@ -20,6 +20,7 @@
 
 package com.apple.foundationdb.record.lucene.directory;
 
+import com.apple.foundationdb.async.AsyncUtil;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.lucene.LuceneEvents;
@@ -92,7 +93,12 @@ public final class FDBDirectoryLockFactory extends LockFactory {
             logSelf("FileLock: Attempt to create a file Lock");
             fileLockSet(false);
             agilityContext.flush();
+            agilityContext.setCommitCheck(this::ensureValidIfNotClosed);
             logSelf("FileLock: Successfully created a file lock");
+        }
+
+        private CompletableFuture<Void> ensureValidIfNotClosed(final FDBRecordContext context) {
+            return closed ? AsyncUtil.DONE : fileLockSet(true, context);
         }
 
         @Override
@@ -206,9 +212,14 @@ public final class FDBDirectoryLockFactory extends LockFactory {
                 agilityContext.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FILE_LOCK_CLEAR,
                         agilityContext.apply(fileLockFunc));
             }
-            // always flush before declaring this object as closed. Else agilityContext may fail to commit, while this lock may skip retry
-            agilityContext.flush();
-            closed = true;
+            boolean flushed = false;
+            try {
+                closed = true; // prevent lock stamp update
+                agilityContext.flush();
+                flushed = true;
+            } finally {
+                closed = flushed; // allow close retry
+            }
         }
 
         protected void fileLockClearIfLocked() {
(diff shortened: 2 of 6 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -197,6 +205,16 @@ class Agile implements AgilityContext {
         private boolean committingNow = false;
         private long prevCommitCheckTime;
         private boolean closed = false;
+        private volatile Function<FDBRecordContext, CompletableFuture<Boolean>> contextPreconditions = context -> AsyncUtil.READY_TRUE;
```
ScottDugas, line 208: I only just started learning about `VarHandle`, but according to the JEP, the volatile may be unnecessary: https://openjdk.org/jeps/193

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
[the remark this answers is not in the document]
todor-m: > I think when we close the FileLock, we want to remove it as a pre-condition

This answers the most important question I had. 
If we expect only few file locks per context I can just make it always return `true` when the lock is closed. `IdentityHashMap` is not thread safe and a concurrent set will be too much for this use case. What about `CopyOnWriteList`? It will be fast enough for the few entries and relatively cheap.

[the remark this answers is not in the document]
jjezra: Isn't that an overkill? The agility context and the lock (via directory) are bound together, so we expect only one of each. We can have a single function in agility context, set by the lock's constructor, and called by AC before every commit if non-null. I see no reason for it to be protected from multi threads as it will be set only once, and there is no other thread until the lock is achieved. 
The callback can be as simple as the lock's `ensureValid` that returns quietly if already closed. 
Something like: 
````
FDBDirectoryLock: 
   FDBDirectoryLock: 
       agilityContext.setPreCommit(this::ensureValidIfNotClosed)

AgilityContext
    setPreCommit(callback): 
       this.callback = callback 

    commitNow():  
        ....
       try {
       // already protected by semaphores     
       if (this.callback != null): 
           this.callback.accept() 
       currentContext.commit();
       currentContext.close();
      .... 
````

Comment (todor-m) [reply fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java:208]: I have used only `Atomic*FieldUpdater` so far too. 

> A field need not be marked as volatile for an associated VarHandle to perform volatile access. In effect, the volatile modifier, if present, is ignored.

I am not sure how to interpret this. The `VarHandle` access doesn't need it but what about the plain read? My guess is that the `volatile` is not needed too. The old JVMs were adding `mfence` instruction only after a write. Most likely the `VarHanlde` codegen does something similar even when working on a plain field. Ignoring the `volatile` probably means that they put the memory fence anyway. But that may be a different type of memory fence (or the ARM acquire/release). 
This is why I left the `volatile`. It is either ignored or it should be there for the plain read.

File: fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java [hunk 2]
```diff
@@ -92,9 +92,21 @@ private FDBDirectoryLock(final AgilityContext agilityContext, final String lockN
...
         }
 
+        private void ensureValidIfNotClosed(final FDBRecordContext context) {
+            if (closed) {
+                return;
+            }
+            final long now = System.currentTimeMillis();
+            if (now > timeStampMillis + timeWindowMilliseconds) {
+                throw new AlreadyClosedException("Lock is too old. This=" + this + " now=" + now);
+            }
```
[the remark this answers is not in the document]
jjezra: `fileLockSet(true)` should do it, I think. It will also handle the `asyncToSync`.

jjezra (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -234,15 +244,35 @@ private void createIfNeeded() {
...
+                try {
+                    preCommit.accept(newContext);
+                    valid = true;
+                } finally {
+                    if (!valid) {
+                        newContext.close();
+                    }
+                }
+            }
+            return newContext;
```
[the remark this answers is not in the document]
todor-m: It is closed only when a validation exception is being thrown. There is no return statement in this case.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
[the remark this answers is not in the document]
todor-m: `fileLockSet(true)` calls back to `apply` which will try the `readLock` while holding the `writeLock`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -499,6 +510,11 @@ public void abortAndClose() {
         public boolean isClosed() {
             return closed;
         }
+
+        @Override
+        public void setPreCommit(final Consumer<FDBRecordContext> preCommit) {
+            // This is a no-op
```
ScottDugas, line 516: Shouldn't this just call `preCommit.accept(currentContext)` immediately?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -286,6 +294,9 @@ public void commitNow() {
                     final long stamp = lock.writeLock();
 
                     try {
+                        if (preCommit != null) {
+                            preCommit.accept(currentContext);
+                        }
```
ScottDugas, line 299: I see this comment from @jjezra 
> Wouldn't the right place to implement this is in commitNow()? We wish to written timestamp to be as new as possible, as it only becomes effective after transaction commit.

I think that's probably right, but it does mean that we could do a bunch of IO just to find out that we lost the FileLock. Probably best to assume that is rare, and leave the code as you have it.

Also, note that `FDBRecordContext` has a concept of commit checks, so you could call `currentContext.addCommitCheck` when the context is opened, and it would check by the time it is committed.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -286,6 +294,9 @@ public void commitNow() {
...
                         currentContext.commit();
                         currentContext.close();
```
ScottDugas, line 301: We don't close the context if the `preCommit` fails (or if the commit fails), we should ensure that we always call `close` by putting this in the finally.

ScottDugas (changes requested):
Looks about right, but needs tests and an update to the release notes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -181,6 +187,20 @@ private void fileLockCheckNewLock(byte[] val, long nowMillis) {
...
+            return aContext.ensureActive().get(fileLockKey)
+                    .thenAccept(val -> {
+                        synchronized (fileLockSetLock) {
+                            if (val != null) {
+                                this.timeStampMillis = nowMillis;
+                                final byte[] value = fileLockValue();
+                                aContext.ensureActive().set(fileLockKey, value);
+                            }
+                        }
+                    });
```
jjezra, line 201: IIUIC, at this path the assumption is that we own the lock since `closed == false`. Hence we should: 
1. require a non null
2. assert that the lock matches `selfStampUuid`. 


One easy way to do it would be to return `fileLockSet(true, aContext)`, I think.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -170,6 +170,8 @@ default <T> T getPropertyValue(@Nonnull RecordLayerPropertyKey<T> propertyKey) {
         return getCallerContext().getPropertyStorage().getPropertyValue(propertyKey);
     }
 
+    void setPreCommit(Function<FDBRecordContext, CompletableFuture<Void>> preCommit);
```
jjezra, line 173: Would you add a JavaDoc comment explaining that the `preCommit` callback is expected to use the argument context but should never call other agility context functions as it may cause a deadlock?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -499,6 +513,11 @@ public void abortAndClose() {
         public boolean isClosed() {
             return closed;
         }
+
+        @Override
+        public void setPreCommit(final Function<FDBRecordContext, CompletableFuture<Void>> preCommit) {
+            callerContext.addCommitCheck(() -> preCommit.apply(callerContext));
```
jjezra, line 519: Letting the non-agile implement it this way is a nice abstraction . I can't think of any issue expected in our code, but also not much gain - as the user context will commit after directory close. 
@ScottDugas, do you see any issue with it?

jjezra (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
todor-m: `fileLockSet` throws exceptions when `val` is `null`

todor-m: The close path `fileLockClearFlushAndClose` clears the lock first then does flush/commit, then assigns close = true. val is always null because of this.
https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java#L207-L211

jjezra: Can we change the refereed code to something like: 
```
try {
            closed = true; // prevent lock stamp update
            agilityContext.flush();
} catch (ex) {
            closed = false;  // allow close retry
            throw ex;
}
```
?

todor-m: Take a look now

jjezra: Nice! 
So now can we call `fileLockSet(true, aContext)` that will require valid uuid if not closed?

todor-m: It makes it the possible now (look at the latest change) but I couldn't find any tests making this call.

jjezra (dismissed):
Nice!

jjezra:
One comment, though - it might be a good idea to change the default time out to 20 seconds: 
https://github.com/FoundationDB/fdb-record-layer/blob/b48cbe32668c16458d3c85f2139ed809d0adaabc/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordContextProperties.java#L109

This will save us another PR.

jjezra (dismissed):

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -170,6 +170,14 @@ default <T> T getPropertyValue(@Nonnull RecordLayerPropertyKey<T> propertyKey) {
         return getCallerContext().getPropertyStorage().getPropertyValue(propertyKey);
     }
 
+    /**
+     * The {@code commitCheck} callback is expected to use the argument context but should never call other
+     * agility context functions as it may cause a deadlock.
+     *
+     * @param commitCheck callback
+     */
+    void setCommitCheck(Function<FDBRecordContext, CompletableFuture<Void>> commitCheck);
```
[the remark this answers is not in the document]
todor-m: Added a test `FDBDirectoryLockTest#testFileLockCallback`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneLockFailureTest.java
```diff
@@ -148,7 +152,12 @@ void testDeleteAll(boolean partitioned) throws IOException {
             grabLockExternally(partitioned, context, 2, 0);
             // Delete all just deletes the index, not through Lucene
             recordStore.deleteAllRecords();
-            context.commit();
+            try {
```
ScottDugas, line 155: I think we actually want this test to behave more like `testDeleteWhere` in that it should get the lock, then commit, then do `deleteAllRecords` in a separate context. I can followup with a quick PR to change that.

ScottDugas (approved):

PR #2732, opened 2024-05-20
Resolve #2731: Reproduce abandoned FileLock due to conflict issue

### LINKED ISSUE #2731: Reproduce abandoned FileLock due to conflict issue
Recreate the scenario of: 
- Lucene file lock is successfully acquired. 
- Another transaction fails with conflict while attempting to release the file lock. 
With a recent code, the lock is released during the directory's close.

jjezra:
Attempt to cause a conflict and abandon a file lock. 
The test is failing for 3.4.485.0

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
@@ -20,6 +20,7 @@
 
 package com.apple.foundationdb.record.lucene;
 
+import com.apple.foundationdb.Transaction;
 import com.apple.foundationdb.record.EvaluationContext;
 import com.apple.foundationdb.record.ExecuteProperties;
 import com.apple.foundationdb.record.IndexEntry;
@@ -36,6 +37,7 @@ import com.apple.foundationdb.record.TestRecordsTextProto;
 import com.apple.foundationdb.record.TestRecordsTextProto.ComplexDocument;
 import com.apple.foundationdb.record.lucene.LuceneIndexTestUtils.IndexedType;
 import com.apple.foundationdb.record.lucene.codec.LuceneOptimizedPostingsFormat;
+import com.apple.foundationdb.record.lucene.directory.AgilityContext;
 import com.apple.foundationdb.record.lucene.directory.FDBDirectory;
 import com.apple.foundationdb.record.lucene.directory.FDBLuceneFileReference;
 import com.apple.foundationdb.record.lucene.ngram.NgramAnalyzer;
@@ -97,9 +99,11 @@ import com.google.protobuf.Message;
 import org.apache.lucene.document.LongPoint;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.search.FieldDoc;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
+import org.apache.lucene.store.Lock;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Tag;
@@ -135,6 +139,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -1344,5 +1351,118 @@ void luceneOnlineIndexingTestTerribleRebalance2ndChance() {
...
+                }).join();
+                tr.set(conflictKey, Tuple.from(100, 20).pack());
+                say("main thread set agile:    " + Arrays.toString(conflictKey));
+            });
+
+
+            // Cause conflict, and wait for it
+            s1.release();
+            acquire(s2);
+            s2.release();
```
ScottDugas, line 1417: Because of the semaphores, I don't think you don't actually have multiple threads (which makes the test more deterministic, which is great), but I think it also means you could move the code for `conflictThread` here and remove all the semaphores.

Although you are adding a nested context, so perhaps putting it in a `Runnable` (or helper function) outside the `try (context = fdb.openContext)` above would help reduce the possibility of accidentally using the wrong context. The same probably goes for the `agile.accept`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -1344,5 +1351,118 @@ void luceneOnlineIndexingTestTerribleRebalance2ndChance() {
...
+
+    private void say(String msg) {
+        LOGGER.debug("::::::" + msg);
+    }
+
+    private void acquire(Semaphore semaphore) {
+        try {
+            semaphore.acquire();
+        } catch (InterruptedException ex) {
+            throw new RuntimeException(ex);
```
ScottDugas, line 1464: Not super important because it's a test, but generally, you should always re-interrupt the current thread if you catch `InterruptedException`, and it's good to be consistently in that habit.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -1344,5 +1351,118 @@ void luceneOnlineIndexingTestTerribleRebalance2ndChance() {
             assertTrue(Terriblerebalnce2ndChanceIndexMaintainer.isSecondPass);
         }
     }
+
+
+    @Test
+    void testConflictWithLock() throws IOException, InterruptedException {
```
ScottDugas, line 1357: This test has a bunch of moving pieces, for posterity it would probably be good to comment exactly what is trying to happen here, something like:
> If the file lock gets a conflict while closing the lock, it should ensure that the lock is released either during it's close, or the directory's close so that future operations can continue.

ScottDugas: Probably also worth adding some similar context to the Issue #2731 which currently has no description.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -1344,5 +1351,118 @@ void luceneOnlineIndexingTestTerribleRebalance2ndChance() {
             assertTrue(Terriblerebalnce2ndChanceIndexMaintainer.isSecondPass);
         }
     }
+
+
+    @Test
```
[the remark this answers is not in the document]
jjezra: Moved to `LuceneIndexTest`, as `recordStore` is needed.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
jjezra: ✅

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -6063,4 +6068,92 @@ private void assertIndexEntryPrimaryKeyTuples(Set<Tuple> primaryKeys, RecordCurs
...
+            store.saveRecord(doc3);
+            context.commit();
+        }
+    }
+
+    private void testConflictWithLockAgileSet(AgilityContext agile1, Tuple conflictTuple) {
+        agile1.accept(aContext -> {
+            byte [] conflictKey = path.toSubspace(aContext).pack(conflictTuple);
+            final Transaction tr = aContext.ensureActive();
+            tr.get(conflictKey).thenApply(val -> null).join();
```
ScottDugas, line 6143: Why have `.thenApply(val -> null)` here and below?

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
jjezra: Probably a leftover. Removed. 
✅

ScottDugas (approved):

PR #2740, opened 2024-05-28
RESOLVES #2739: Merge partitions when they become too small

### LINKED ISSUE #2739: Merge partitions when they become too small
Merge partitions when they become too small

wshaib-apple:
- when a partition falls below the low watermark, and if one or both of its neighboring partitions have capacity to absorb, it will be merged into them.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
@@ -63,7 +63,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.Query;
@@ -96,6 +95,7 @@ public class LucenePartitioner {
     private static final FDBStoreTimer.Waits WAIT_LOAD_LUCENE_PARTITION_METADATA = FDBStoreTimer.Waits.WAIT_LOAD_LUCENE_PARTITION_METADATA;
     private static final Logger LOGGER = LoggerFactory.getLogger(LucenePartitioner.class);
     private static final int DEFAULT_PARTITION_HIGH_WATERMARK = 400_000;
+    private static final int DEFAULT_PARTITION_LOW_WATERMARK = 1_000;
     private static final ConcurrentHashMap<String, KeyExpression> partitioningKeyExpressionCache = new ConcurrentHashMap<>();
     public static final int PARTITION_META_SUBSPACE = 0;
     public static final int PARTITION_DATA_SUBSPACE = 1;
@@ -103,6 +103,7 @@ public class LucenePartitioner {
     private final boolean partitioningEnabled;
     private final String partitionFieldNameInLucene;
     private final int indexPartitionHighWatermark;
+    private final int indexPartitionLowWatermark;
     private final KeyExpression partitioningKeyExpression;
 
     public LucenePartitioner(@Nonnull IndexMaintainerState state) {
@@ -119,7 +120,16 @@ public class LucenePartitioner {
         indexPartitionHighWatermark = strIndexPartitionHighWatermark == null ?
                                       DEFAULT_PARTITION_HIGH_WATERMARK :
                                       Integer.parseInt(strIndexPartitionHighWatermark);
+
+        String strIndexPartitionLowWatermark = state.index.getOption(LuceneIndexOptions.INDEX_PARTITION_LOW_WATERMARK);
+        indexPartitionLowWatermark = strIndexPartitionLowWatermark == null ?
+                                      DEFAULT_PARTITION_LOW_WATERMARK :
+                                      Integer.parseInt(strIndexPartitionLowWatermark);
         this.partitioningKeyExpression = makePartitioningKeyExpression(partitionFieldName);
+
+        if (indexPartitionHighWatermark < indexPartitionLowWatermark) {
+            throw new RecordCoreArgumentException("High watermark must be greater than low watermark");
+        }
     }
 
     /**
@@ -444,15 +454,18 @@ public class LucenePartitioner {
      *
      * @param newRecord record to be written
      * @param groupingKey grouping key
+     * @param assignedPartitionIdHint assigned partition override, if not null
      * @param <M> message
      * @return partition id or <code>null</code> if partitioning isn't enabled on index
      */
     @Nonnull
-    public <M extends Message> CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull FDBIndexableRecord<M> newRecord, @Nonnull Tuple groupingKey) {
+    public <M extends Message> CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull FDBIndexableRecord<M> newRecord,
+                                                                                        @Nonnull Tuple groupingKey,
+                                                                                        @Nullable Integer assignedPartitionIdHint) {
         if (!isPartitioningEnabled()) {
             return CompletableFuture.completedFuture(null);
         }
-        return addToAndSavePartitionMetadata(groupingKey, toPartitionKey(newRecord));
+        return addToAndSavePartitionMetadata(groupingKey, toPartitionKey(newRecord), assignedPartitionIdHint);
     }
 
     /**
@@ -460,12 +473,23 @@ public class LucenePartitioner {
      * The <code>count</code> will be incremented, and the <code>from</code> or <code>to</code> boundary values will
      * be adjusted if applicable.
      *
+     * @param groupingKey grouping key
      * @param partitioningKey partitioning key
+     * @param assignedPartitionIdHint assigned partition override, if not null
      * @return assigned partition id
      */
     @Nonnull
-    private CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull final Tuple groupingKey, @Nonnull final Tuple partitioningKey) {
-        return getOrCreatePartitionInfo(groupingKey, partitioningKey).thenApply(assignedPartition -> {
+    private CompletableFuture<Integer> addToAndSavePartitionMetadata(@Nonnull final Tuple groupingKey,
+                                                                     @Nonnull final Tuple partitioningKey,
+                                                                     @Nullable final Integer assignedPartitionIdHint) {
+
+        final CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> assignmentFuture;
+        if (assignedPartitionIdHint != null) {
+            assignmentFuture = getPartitionMetaInfoById(assignedPartitionIdHint, groupingKey);
+        } else {
+            assignmentFuture = getOrCreatePartitionInfo(groupingKey, partitioningKey);
+        }
+        return assignmentFuture.thenApply(assignedPartition -> {
             // assignedPartition is not null, since a new one is created by the previous call if none exist
             LucenePartitionInfoProto.LucenePartitionInfo.Builder builder = Objects.requireNonNull(assignedPartition).toBuilder();
             builder.setCount(assignedPartition.getCount() + 1);
@@ -741,8 +765,8 @@ public class LucenePartitioner {
         final KeyExpression rootExpression = state.index.getRootExpression();
 
         if (! (rootExpression instanceof GroupingKeyExpression)) {
-            return processPartitionRebalancing(Tuple.from(), documentCount, logMessages).thenApply(result -> {
-                if (result.getLeft() > 0) {
+            return processPartitionRebalancing(Tuple.from(), documentCount, logMessages).thenApply(movedCount -> {
+                if (movedCount > 0) {
                     // we did something, repeat
                     return RecordCursorStartContinuation.START;
                 } else {
@@ -772,8 +796,8 @@ public class LucenePartitioner {
                 if (cursorResult.hasNext()) {
                     final Tuple groupingKey = Tuple.fromItems(cursorResult.get().getItems().subList(0, groupingCount));
                     return processPartitionRebalancing(groupingKey, documentCount, logMessages)
-                            .thenCompose(repartitionResult -> {
-                                if (repartitionResult.getLeft() > 0) {
+                            .thenCompose(movedCount -> {
+                                if (movedCount > 0) {
                                     // we did something, stop so we can create a new transaction
                                     return AsyncUtil.READY_FALSE;
                                 } else {
@@ -790,6 +814,15 @@ public class LucenePartitioner {
         }
     }
 
+    private static class RepartitioningContext {
+        boolean emptyingPartition = false;
+        LuceneRecordCursor cursor;
+        int countToMove;
+        boolean removingOldest = true;
+        int maxPartitionId;
+        boolean newBoundaryRecordPresent = false;
+    }
+
     /**
      * Re-balance the first partition in a given grouping key by moving documents out of it.
      *
@@ -800,18 +833,20 @@ public class LucenePartitioner {
      * @param groupingKey grouping key
      * @param repartitionDocumentCount max number of documents to move in each transaction
      * @param logMessages {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner} additional log messages
-     * @return {@code true} future if there is more repartitioning to be done in this group
+     * @return future count of documents rebalanced in this group. If zero, no more re-balancing needed
      */
+    @SuppressWarnings("PMD.AvoidBranchingStatementAsLastInLoop")
     @Nonnull
-    public CompletableFuture<Pair<Integer, Integer>> processPartitionRebalancing(@Nonnull final Tuple groupingKey,
-                                                                                 int repartitionDocumentCount,
-                                                                                 RepartitioningLogMessages logMessages) {
+    public CompletableFuture<Integer> processPartitionRebalancing(@Nonnull final Tuple groupingKey,
+                                                                  int repartitionDocumentCount,
+                                                                  RepartitioningLogMessages logMessages) {
         if (repartitionDocumentCount <= 0) {
             throw new IllegalArgumentException("number of documents to move can't be zero");
         }
         return getAllPartitionMetaInfo(groupingKey).thenCompose(partitionInfos -> {
             // need to track the next partition id to use when creating a new one
-            int maxPartitionId = partitionInfos.stream().map(LucenePartitionInfoProto.LucenePartitionInfo::getId).max(Integer::compare).orElse(0);
+            RepartitioningContext repartitioningContext = new RepartitioningContext();
+            repartitioningContext.maxPartitionId = partitionInfos.stream().map(LucenePartitionInfoProto.LucenePartitionInfo::getId).max(Integer::compare).orElse(0);
             if (LOGGER.isTraceEnabled()) {
                 LOGGER.trace(partitionInfos.stream()
                         .sorted(Comparator.comparing(pi -> Tuple.fromBytes(pi.getFrom().toByteArray())))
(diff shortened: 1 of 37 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -444,28 +454,42 @@ public String getPartitionFieldNameInLucene() {
      *
      * @param newRecord record to be written
      * @param groupingKey grouping key
+     * @param assignedPartitionIdHint assigned partition override, if not null
```
ScottDugas, line 457: `hint` seems misleading, it looks like this is used no matter what, replacing any attempt to lookup what it should be.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
...
+
+        // assert correct partition setup after repartitioning
+        assertEquals(expectedPartitionCounts.length, partitionInfos.size());
+        assertArrayEquals(expectedPartitionCounts, partitionInfos.stream().mapToInt(LucenePartitionInfoProto.LucenePartitionInfo::getCount).toArray());
+
+        try (FDBRecordContext context = openContext(contextProps)) {
+            schemaSetup.accept(context);
+            // finally, assert all docs are accounted for
+            assertIndexEntryPrimaryKeyTuples(createdKeys.get(Tuple.from(1L)).keySet(),
+                    recordStore.scanIndex(COMPLEX_PARTITIONED, groupedTextSearch(COMPLEX_PARTITIONED, luceneSearch, 1), null, ExecuteProperties.newBuilder().build().asScanProperties(false)));
```
ScottDugas, line 1800: This doesn't seem to validate that the right documents are in the right partitions, it seems important to run `LuceneIndexTestValidator.validate` on this, to ensure that the `from`/`to` all make sense for the documents contained, and that they are still ordered correctly.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -958,25 +1076,42 @@ private CompletableFuture<Integer> moveDocsFromPartition(@Nonnull final LucenePa
...
                 // value of the "destination" partition's `from` value
                 final Tuple overflowPartitioningKey = toPartitionKey(records.get(0));
-                return findPartitionInfo(groupingKey, overflowPartitioningKey).thenCompose(previousPartition -> {
-                    if (previousPartition == null || previousPartition.getCount() + records.size() > indexPartitionHighWatermark || previousPartition.getId() == partitionInfo.getId()) {
+                final CompletableFuture<LucenePartitionInfoProto.LucenePartitionInfo> futureDestination;
+                if (repartitioningContext.removingOldest) {
+                    futureDestination = findPartitionInfo(groupingKey, overflowPartitioningKey);
+                } else {
+                    futureDestination = getNextNewerPartitionInfo(state.context, groupingKey, overflowPartitioningKey, state.indexSubspace);
+                }
```
ScottDugas, line 1104: Well above this we get all the partition info and put it in a list, would it make sense to sort that list, and use that instead of using these methods?
The downside is that you would need two versions of the code for `findPartitionInfo` and `getNextNewerPartitionInfo` depending on whether you're hitting FDB or a list, but it might be a little bit cleaner/faster.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -958,25 +1076,42 @@ private CompletableFuture<Integer> moveDocsFromPartition(@Nonnull final LucenePa
...
+                return futureDestination.thenCompose(destinationPartition -> {
+                    if (destinationPartition == null || destinationPartition.getCount() + records.size() > indexPartitionHighWatermark || destinationPartition.getId() == partitionInfo.getId()) {
                         // create a new "overflow" partition
```
ScottDugas, line 1107: Is it possible for it to hit this code when shrinking a partition?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -800,57 +833,121 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
...
 
+    /**
+     * convenience function that proxies into {@link #moveDocsFromPartition(LucenePartitionInfoProto.LucenePartitionInfo, Tuple, RepartitioningContext)}
+     * after outputting some logs.
+     *
+     * @param partitionInfo partition info
+     * @param groupingKey grouping key
+     * @param repartitioningContext repartition parameters
+     * @param logMessages log messages
+     * @return future count of documents mvoed
```
ScottDugas, line 931:
```suggestion
     * @return future count of documents moved
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -800,57 +833,121 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
...
+                    }
+                } else {
+                    // partition is above high watermark
+                    repartitioningContext.countToMove = 1 + Math.min(repartitionDocumentCount, indexPartitionHighWatermark);
+                    // sanity: in case someone passes a high enough repartitionDocumentCount such that removing
+                    // this many docs would flip the partition from being above the high watermark to being below
+                    // the low watermark
+                    if (currentPartitionCount - repartitioningContext.countToMove < indexPartitionLowWatermark) {
+                        repartitioningContext.countToMove = indexPartitionHighWatermark - indexPartitionLowWatermark;
+                    }
```
ScottDugas, line 912: I guess it's probably a bug if this happens, so maybe optimizing it is irrelevant, but it seems to me that you would want to move `(indexPartitionHighWatermark + indexPartitionLowWatermark) / 2`, rather than bringing it all the way to the low watermark.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -790,6 +814,15 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
         }
     }
 
+    private static class RepartitioningContext {
+        boolean emptyingPartition = false;
+        LuceneRecordCursor cursor;
+        int countToMove;
+        boolean removingOldest = true;
+        int maxPartitionId;
+        boolean newBoundaryRecordPresent = false;
```
ScottDugas, line 823: What is this one for?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -800,57 +833,121 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
...
+                        repartitioningContext.countToMove = Math.min(previousCapacity, repartitionDocumentCount);
+                        if (currentPartitionCount > repartitioningContext.countToMove) {
+                            repartitioningContext.countToMove++;
+                            repartitioningContext.newBoundaryRecordPresent = true;
+                        }
+                        repartitioningContext.emptyingPartition = false;
+                        repartitioningContext.cursor = getOldestNDocuments(partitionInfo, groupingKey, repartitioningContext.countToMove);
+                    } else {
+                        // partition is under low-watermark, but no capacity around it to absorb it
+                        continue;
```
ScottDugas, line 902: does `repartitioningCount` need to get reset if this happens? 
I think the flow would have to be something like:
1. see a partition that you should empty
2. all of the if/else-if above return false so it gets to here
3. next partition should not be emptied.

The only other field that is set is the `countToMove` which always gets set.
It does seem a bit brittle though to have it living between partitions, even if resetting `emptyingPartition` would be sufficient for right now.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -800,57 +833,121 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
...
+
+                final int currentPartitionCount = partitionInfo.getCount();
+
+                if (currentPartitionCount >= indexPartitionLowWatermark && currentPartitionCount <= indexPartitionHighWatermark) {
+                    // repartitioning not needed
+                    continue;
+                }
+
+                if (currentPartitionCount < indexPartitionLowWatermark) {
+                    // partitionInfos are ordered next -> previous
```
ScottDugas, line 872: `next` and `previous` I think are new terms only used here, perhaps replacing their usage with `newer` and `older` respectively.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -800,57 +833,121 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
...
+                    LucenePartitionInfoProto.LucenePartitionInfo next = i == 0 ? null : partitionInfos.get(i - 1);
+                    LucenePartitionInfoProto.LucenePartitionInfo previous = i == partitionInfos.size() - 1 ? null : partitionInfos.get(i + 1);
+                    repartitioningContext.countToMove = Math.min(currentPartitionCount, repartitionDocumentCount);
+                    if (currentPartitionCount > repartitionDocumentCount) {
+                        repartitioningContext.countToMove++;
+                    } else {
+                        repartitioningContext.emptyingPartition = true;
                     }
```
ScottDugas, line 880: It _might_ be clearer to remove the `Math.min` and just set `countToMove` in both blocks
```suggestion
                    if (currentPartitionCount > repartitionDocumentCount) {
                        repartitioningContext.countToMove = repartitionDocumentCount + 1;
                    } else {
                        repartitioningContext.countToMove = currentPartitionCount;
                        repartitioningContext.emptyingPartition = true;
                    }
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
...
+                Arguments.of(2, 4, 3, new int[] {1, 1}, new int[] {2}),
+                // consolidate a partition into its previous neighbor
+                Arguments.of(2, 4, 3, new int[] {2, 2, 1}, new int[] {2, 3}),
+                // consolidate a partition falling between two partitions with capacity into its previous neighbor
+                Arguments.of(3, 7, 3, new int[] {5, 2, 5}, new int[] {7, 5}),
+                // consolidate a partition falling between two partitions which individually don't have enough
+                // capacity, but together they do
+                Arguments.of(3, 7, 3, new int[] {6, 2, 6}, new int[] {7, 7}),
+                // cannot consolidate a partition that has no neighbors with capacity
+                Arguments.of(3, 7, 3, new int[] {6, 2, 7}, new int[] {6, 2, 7}),
```
ScottDugas, line 1733: It seems like you should have the same test starting with `{7, 2, 6}`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
...
+                // cannot consolidate partitions that have no neighbors with capacity
+                Arguments.of(3, 7, 3, new int[] {6, 2, 7, 3}, new int[] {6, 2, 7, 3}),
+                // consolidate one partition that has a neighbor with capacity, while another
+                // that doesn't, won't be consolidated
+                Arguments.of(3, 7, 3, new int[] {6, 2, 6, 1}, new int[] {6, 2, 7}),
+                // splitting one partition, removes the need to consolidate a previous low partition
+                Arguments.of(3, 7, 3, new int[] {6, 1, 8}, new int[] {6, 4, 5}),
+                // consolidating two neighboring partitions into their left and right neighbors
+                Arguments.of(3, 7, 3, new int[] {6, 1, 1, 6}, new int[] {7, 7}),
```
ScottDugas, line 1742: What about
```
Arguments.of(4, 7, 3, new int[] {6, 3, 4}, new int[] { ? ? }),
```

And should that behave differently if there's a whole bunch of newer partitions

```
Arguments.of(4, 7, 3, new int[] {6, 3, 4, 5, 5, 5, 5}, new int[] { ? ? }),
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -958,25 +1076,42 @@ private CompletableFuture<Integer> moveDocsFromPartition(@Nonnull final LucenePa
...
+                        destinationPartition = newPartitionMetadata(overflowPartitioningKey,  repartitioningContext.maxPartitionId + 1);
+                        savePartitionMetadata(groupingKey, destinationPartition.toBuilder());
                     }
 
                     Iterator<? extends FDBIndexableRecord<Message>> recordIterator = records.iterator();
-                    return AsyncUtil.whileTrue(() -> indexMaintainer.update(null, recordIterator.next())
+                    final int destinationPartitionId = destinationPartition.getId();
+                    return AsyncUtil.whileTrue(() -> indexMaintainer.update(null, recordIterator.next(), destinationPartitionId)
                             .thenApply(ignored -> recordIterator.hasNext()), state.context.getExecutor());
                 }).thenApply(ignored -> records.size());
```
ScottDugas, line 1116: I don't think we close the cursor used to get the newest or oldest documents.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -790,6 +814,15 @@ public CompletableFuture<RecordCursorContinuation> rebalancePartitions(RecordCur
         }
     }
 
+    private static class RepartitioningContext {
+        boolean emptyingPartition = false;
+        LuceneRecordCursor cursor;
```
ScottDugas, line 819: Would it make sense to have this have the target partition instead, and call `getOldestestNDocuments` or `getNewestNDocuments` based on `removingOldest`, closer to when the cursor is being used.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
         return hits.stream().map(Map.Entry::getKey).collect(Collectors.toList());
     }
 
+    static Stream<Arguments> simplePartitionConsolidationTest() {
+        return Stream.of(
```
ScottDugas, line 1722: Thinking about the testability, and complexity here, I wonder if it would make sense, and not be a big change, to have something like a `LuceneRepartitionPlanner` that takes a `List<LucenePartitionInfoProto.LucenePartitionInfo>` and returns a plan, which is something like:
> Move N newest documents from partition 3 to partition 4 and you are emptying partition 3.

Or

> Move M oldest documents from partition 2 to partition 1.

And then `LucenePartitioner` takes that plan and executes it in a single transaction. I'm not sure who is in charge of whether there is more to do.
That would probably allow decoupling the tests so you can have a lot of unit tests of the `RepartitionPlanner`, and fewer tests that actually involve the whole machinery and saving actual records to the database.

ScottDugas (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
...
+
+        List<LucenePartitionInfoProto.LucenePartitionInfo> partitionInfos = getPartitionMeta(index,
+                Tuple.from(1), contextProps, schemaSetup);
+        partitionInfos.sort(Comparator.comparing(LucenePartitionInfoProto.LucenePartitionInfo::getId));
+
+        // assert sane initial setup
+        assertEquals(initialPartitionCounts.length, partitionInfos.size());
+        assertArrayEquals(initialPartitionCounts, partitionInfos.stream().mapToInt(LucenePartitionInfoProto.LucenePartitionInfo::getCount).toArray());
+
+        explicitMergeIndex(index, contextProps, schemaSetup);
```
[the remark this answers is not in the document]
wshaib-apple: I think I ensured no infinite loop since we would never move enough docs to cause a destination partition to go above its high watermark...

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: see above

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: you mean `(indexPartitionHighWatermark - indexPartitionLowWatermark) / 2` ?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
wshaib-apple: in addition,

1. we would need to pass the partition info list down to this method (in the `repartitioningContext`?
2. make sure while implementing the new versions of `findPartitionInfo` and `getNextNewerPartitionInfo` to be aware whether the current partition has been cleared in the DB (a few lines above here).

is the gain in speed is worth this added complexity?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
```diff
@@ -1707,6 +1718,89 @@ private List<Tuple> queryLocal(Map<Tuple, Long> dataset, Comparisons.Type compar
...
+                Arguments.of(3, 7, 3, new int[] {6, 2, 7, 3}, new int[] {6, 2, 7, 3}),
+                // consolidate one partition that has a neighbor with capacity, while another
+                // that doesn't, won't be consolidated
+                Arguments.of(3, 7, 3, new int[] {6, 2, 6, 1}, new int[] {6, 2, 7}),
+                // splitting one partition, removes the need to consolidate a previous low partition
+                Arguments.of(3, 7, 3, new int[] {6, 1, 8}, new int[] {6, 4, 5}),
+                // consolidating two neighboring partitions into their left and right neighbors
+                Arguments.of(3, 7, 3, new int[] {6, 1, 1, 6}, new int[] {7, 7}),
+                // multiple-stage split
+                Arguments.of(3, 7, 3, new int[] {15}, new int[] {6, 6, 3})
```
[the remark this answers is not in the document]
wshaib-apple: > Relatedly all the tests have a `repartitionCount>=lowWatermark`, which I don't think is really what you would want....

true, they are, but that actually is a more stringent test. having `repartitionCount < lowWatermark` may not uncover some edge cases. I can make it pass both kinds of values.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTest.java
wshaib-apple: Arguments.of(5, 10, 1, new int[] {7, 4, 10}, new int[] {7, 4, 10}),

PR #2744, opened 2024-05-28
Support Bitmap function in queries

### LINKED ISSUE #2904: Support BitMap functions in Cascade planner
support queries like:
```
select bitmap_bucket_offset, bitmap_construct_agg(bitmap_bit_position(val)) as bitmap
from T
GROUP BY bitmap_bucket_offset(val) as bitmap_bucket_offset
```
bit_bucket_size = 10000 by default. So if val = 1, 2, 3, ... 10010, the result is:

offset: 0, bitmap = [b'00001110, 0, ... (1249 bytes of 0)]
offset: 10000, bitmap = [0, b'00000100 , 0, .. ]

pengpeng-lu:
This fixes #2904


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/IndexKeyValueToPartialRecord.java
```diff
@@ -275,7 +276,11 @@ public boolean copy(@Nonnull Descriptors.Descriptor recordDescriptor, @Nonnull M
             }
             switch (fieldDescriptor.getType()) {
                 case INT32:
-                    value = ((Long)value).intValue();
+                    if (value.getClass().isArray()) {
```
normen662, line 279: I think we want to return the bitmap as a bite array, i.e. in the way this is delivered from the index entry.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java
```diff
@@ -400,7 +412,11 @@ public enum PhysicalOperator {
         MOD_DI(LogicalOperator.MOD, TypeCode.DOUBLE, TypeCode.INT, TypeCode.DOUBLE, (l, r) -> (double)l % (int)r),
         MOD_DL(LogicalOperator.MOD, TypeCode.DOUBLE, TypeCode.LONG, TypeCode.DOUBLE, (l, r) -> (double)l % (long)r),
         MOD_DF(LogicalOperator.MOD, TypeCode.DOUBLE, TypeCode.FLOAT, TypeCode.DOUBLE, (l, r) -> (double)l % (float)r),
-        MOD_DD(LogicalOperator.MOD, TypeCode.DOUBLE, TypeCode.DOUBLE, TypeCode.DOUBLE, (l, r) -> (double)l % (double)r);
+        MOD_DD(LogicalOperator.MOD, TypeCode.DOUBLE, TypeCode.DOUBLE, TypeCode.DOUBLE, (l, r) -> (double)l % (double)r),
+
+        BITBUCKET_L(LogicalOperator.BIT_BUCKET, TypeCode.LONG, TypeCode.LONG, TypeCode.LONG, (l, r) -> Math.multiplyExact(Math.floorDiv((long)l, (long)r), (long)r)),
```
normen662, line 417: The second param should be an integer.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -207,6 +208,75 @@ private static Map<Pair<LogicalOperator, TypeCode>, PhysicalOperator> computeOpe
         return mapBuilder.build();
     }
 
+    /**
+     * BitMap aggregation {@code Value}.
+     */
+    public static class BitMap extends NumericAggregationValue implements StreamableAggregateValue, IndexableAggregateValue {
```
normen662, line 214: Naming convention `...Value`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -590,8 +672,34 @@ public enum PhysicalOperator {
...
         MAX_F(LogicalOperator.MAX, TypeCode.FLOAT, TypeCode.FLOAT, Objects::requireNonNull, (s, v) -> Math.max((float)s, (float)v), identity()),
-        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity());
-
+        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity()),
+        /*
+        pair.key -> last aggregation result, initial = 0
+        pair.value -> input value to be aggregated
+        so final result = pair.key | (1 << pair.value)
+        */
+        BITMAP_LL(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.LONG,
```
normen662, line 681: To mirror the index part, this should return a byte array. `BYTES`. The intermediate type should be a `long` or `int` respectively.

normen662: Please pay attention to endianness.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
pengpeng-lu: I see other classes are named as `Sum`, `Avg`... maybe `BitMap` is better here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -722,7 +834,11 @@ public NumericAccumulator(@Nonnull final PhysicalOperator physicalOperator) {
 
         @Override
         public void accumulate(@Nullable final Object currentObject) {
-            this.state = physicalOperator.evalPartialToPartial(state, currentObject);
+            if (state == null) {
+                this.state = physicalOperator.evalInitialToPartial(currentObject);
+            } else {
+                this.state = physicalOperator.evalPartialToPartial(state, currentObject);
+            }
```
pengpeng-lu, line 841: modify this function, because the old one seems to neglect the `evalInitialToPartial` scenario.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,8 +672,37 @@ public enum PhysicalOperator {
...
+                s -> {
+                    BitSet bs = new BitSet();
+                    bs.set(((Long)s).intValue());
+                    return bs.toLongArray()[0];
+                },
+                (s, v) -> {
+                    // convert s (long to bitset), set v, then convert it back to long
+                    BitSet bs = BitSet.valueOf(new long[] {(long)s});
+                    bs.set(((Long)v).intValue());
+                    return bs.toLongArray()[0];
```
hatyo, line 691: I am not sure how this is supposed to work, it seems we're limiting the bit set's size to 64 bits only, am I correct?

Can you please illustrate via an example how the aggregation should work?

I imagine if this is used in the context of a streaming aggregate, we would have to group the rows according to some bucket size, for example:

```sql
SELECT BITMAP(C)
FROM T
GROUP BY BIT_BUCKET(C, 1000)
```

Let's say we have the following rows:
```sql
C
==
0
1
4
64
65
1001
1002
---
```

I expect to get two rows back representing two groups:
```sql
RESULT
======
[1 1 0 0 1 ...... 0] [1 1 0 .... 0] <-- each representing a single byte, first row contains two longs, sort of VLE.
[0 1 1 0 .......  0]                <-- second group, mapping bits starting at 1000-inclusive.
```

If that's the case, and unless I am missing something, these aggregate definitions (initial, partial, and final) are incorrect because they do not take into consideration the actual value ranges and only assume 64-bit wide range

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,8 +672,37 @@ public enum PhysicalOperator {
...
+        /*
+        pair.key -> last aggregation result, initial = 0
+        pair.value -> input value to be aggregated
+        so final result = pair.key | (1 << pair.value)
+        */
+        BITMAP_LL(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
+                s -> {
+                    BitSet bs = new BitSet();
+                    bs.set(((Long)s).intValue());
+                    return bs.toLongArray()[0];
```
hatyo, line 685: I don't think this is correct, please see above.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -100,6 +101,17 @@ void testSumOnlyNulls() {
...
 
+    @Test
+    void testBitMap() {
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), new Object[]{0L, 1L, 2L, 0L}, Arrays.asList(0L, 1L, 2L)); // 111
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), longs, Arrays.asList(longs)); // 1111110
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), longsWithNulls, List.of(1L, 2L, 4L, 5L, 6L)); // 1110110
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), longsOnlyNull, null);
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_II, ofScalar(1)), ints, Arrays.asList(1L, 2L, 3L, 4L, 5L, 6L)); // 1111110
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_II, ofScalar(1)), intsWithNulls, Arrays.asList(1L, 2L, 4L, 5L, 6L)); // 1110110
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_II, ofScalar(1)), intsOnlyNull, null);
```
hatyo, line 112: Please add tests for values larger than 63.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -202,6 +208,29 @@ void testTupleSumCount() {
         });
     }
 
+    @Nullable
+    public static List<Long> collectOnBits(@Nullable byte[] bitmap) {
+        if (bitmap == null) {
+            return null;
+        }
+        int start = 0;
```
hatyo, line 216: is this necessary?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -237,33 +253,234 @@ private Reference constructGroupByPlan(final boolean withPredicateInSelectWhere,
...
+                    byte[] value = ((ByteString) queriedMessage.getField(recDescriptor.findFieldByName("bitmap_field"))).toByteArray();
+                    bitMapGroupByStrValue.put(Pair.of(strValue, bitBucketValue), AggregateValueTest.collectOnBits(value));
+                }).join();
+            }
+        }
+        assertEquals(expectedResult, bitMapGroupByStrValue);
+    }
+
+    @Nonnull
+    private Reference constructBitMapGroupByPlan(int bucketSize, boolean includeBitBucketFunc) {
```
hatyo, line 353: IMG_B4D3C3C7B7AC-1 2

I had a quick look at the plan and it does not look correct to me, namely:
1. The grouping keys group in `select-where` should be gone, in https://github.com/FoundationDB/fdb-record-layer/pull/2762 I introduced a new simplified version of aggregate plans that removes the grouping key from the `select-where`, it moves it up to the `group-by` expression instead.
2. The grouping keys does not include the `bit_bucket` function.

hatyo (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -453,10 +446,10 @@ protected RecordMetaDataHook setupHookAndAddData(final boolean addIndex, final b
                     metaDataBuilder.addIndex("MySimpleRecord", "MySimpleRecord$num_value_2", field("num_value_2"));
                 }
                 if (addAggregateIndex) {
-                    metaDataBuilder.addIndex("MySimpleRecord", new Index("AggIndex", field("num_value_3_indexed").groupBy(field("num_value_2")), IndexTypes.SUM, Map.of("bitmapValueEntrySize", String.valueOf(bucketSize))));
+                    metaDataBuilder.addIndex("MySimpleRecord", new Index("AggIndex", field("num_value_3_indexed").groupBy(field("num_value_2")), IndexTypes.SUM));
                 }
                 if (addBitMapIndex) {
-                    metaDataBuilder.addIndex("MySimpleRecord", new Index("BitMapIndex", field("num_value_2").groupBy(field("str_value_indexed"), field("num_value_2")), IndexTypes.BITMAP_VALUE));
+                    metaDataBuilder.addIndex("MySimpleRecord", new Index("BitMapIndex", modExpression(field("num_value_2"), bucketSize).groupBy(field("str_value_indexed"), bitBucketExpression(field("num_value_2"), bucketSize)), IndexTypes.BITMAP_VALUE, Map.of("bitmapValueEntrySize", String.valueOf(bucketSize))));
```
hatyo, line 452: `IndexOptions.BITMAP_VALUE_ENTRY_SIZE_OPTION`

hatyo, line 452: `"bitmapValueEntrySize" -> IndexOptions.BITMAP_VALUE_ENTRY_SIZE_OPTION`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -340,21 +349,17 @@ void testBitMapWithBitMapIndex() {
...
                     String strValue = (String) queriedMessage.getField(recDescriptor.findFieldByName("str_value_indexed"));
-                    Integer bitBucketValue = (int) queriedMessage.getField(recDescriptor.findFieldByName("bit_bucket_num_value2"));
+                    Integer bitBucketValue = (int) queriedMessage.getField(recDescriptor.findFieldByName("bit_bucket_num_value_2"));
                     byte[] value = ((ByteString) queriedMessage.getField(recDescriptor.findFieldByName("bitmap_field"))).toByteArray();
                     bitMapGroupByStrValue.put(Pair.of(strValue, bitBucketValue), AggregateValueTest.collectOnBits(value));
                 }).join();
             }
         }
-        assertEquals(expectedResult, bitMapGroupByStrValue);
+        // assertEquals(expectedResult, bitMapGroupByStrValue);
```
hatyo, line 358: could you please restore the assertion?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -271,7 +276,7 @@ void testBitMapWithStreamAggregation() {
                                         indexPlan()
 
                                 )).where(aggregations(recordConstructorValue(exactly(bitmapAggregationValue(anyValue()))))
-                                .and(groupings(ValueMatchers.fieldValueWithFieldNames("select_grouping_cols"))))));
+                                .and(groupings(ValueMatchers.anyValue())))));
```
hatyo, line 279: you can still verify that the grouping columns are `str_value_indexed`.

hatyo (changes requested):
Nice work, I had a look at the creation of the aggregate bitmap index and how it is matching correctly in your test.

Please consider adding tests for the following cases:
- the aggregate index is _not_ considered if the bitmap size and/or the bucket size is different from the query's.
- the streaming aggregate and the aggregate index return exactly the same results for the same query.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +674,49 @@ public enum PhysicalOperator {
...
+        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity()),
+        BITMAP_LL(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
+                s -> {
+                    int MAX_ENTRY_SIZE = 250_000;
+                    int pos = (int)s / 8;
+                    if (pos >= MAX_ENTRY_SIZE) {
+                        throw new RecordCoreArgumentException("entry size option is too large")
+                                .addLogInfo("entrySize", pos, "maxEntrySize", MAX_ENTRY_SIZE);
+                    }
+                    byte[] s1 = new byte[MAX_ENTRY_SIZE];
```
pengpeng-lu, line 686: The problem here is that I haven't found a way to pass in the "bucket size".

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -237,33 +257,220 @@ private Reference constructGroupByPlan(final boolean withPredicateInSelectWhere,
         }
     }
 
-    protected void setupHookAndAddData(final boolean addIndex, final boolean addAggregateIndex) {
+    @DualPlannerTest(planner = DualPlannerTest.Planner.CASCADES)
+    void testBitMapWithStreamAggregation() {
```
pengpeng-lu, line 261: for this, could only work with queries like `select a, bitmap(b % 100) from Foo group by a`. it doesn't work with `group by bit_bucket(b, 100)`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,8 +672,37 @@ public enum PhysicalOperator {
...
+        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity()),
+        /*
+        pair.key -> last aggregation result, initial = 0
+        pair.value -> input value to be aggregated
+        so final result = pair.key | (1 << pair.value)
+        */
+        BITMAP_LL(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
+                s -> {
+                    BitSet bs = new BitSet();
+                    bs.set(((Long)s).intValue());
```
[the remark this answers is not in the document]
pengpeng-lu: bitset can hold bit arrays longer than 64. 64 is its "unit", and they can hold multiple units.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
pengpeng-lu: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -500,7 +500,7 @@ protected RecordMetaDataHook setupHookAndAddData(final boolean addIndex, final b
 
     @Nonnull
     private static FunctionKeyExpression bitBucketExpression(@Nonnull KeyExpression fieldExpression, int bucketSize) {
-        return Key.Expressions.function("bit_bucket", concat(fieldExpression, Key.Expressions.value(bucketSize)));
+        return Key.Expressions.function("bitmap_bucket_offset", concat(fieldExpression, Key.Expressions.value(bucketSize)));
```
hatyo, line 503: This is currently failing because of an unintended behaviour in `TreeLike#replaceLeavesMaybe` which is addressed in this PR #2885.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -178,6 +194,23 @@ private Object[] pairsForAvg(Object[] objects) {
...
+                        return null;
+                    }
+                    BitSet result = new BitSet();
+                    if (object instanceof Integer) {
+                        result.set((int)object);
+                    } else {
+                        result.set(((Long)object).intValue());
+                    }
+                    return result;
+                }) // left for the sum, right for the count
```
hatyo, line 210: Is this comment correct? It looks like this is translating an array of numbers to an array of bit sets where these numbers are set?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -178,6 +194,23 @@ private Object[] pairsForAvg(Object[] objects) {
                 .toArray();
     }
 
+    private Object[] bitsetForBitMap(Object[] objects) {
+        return Arrays.stream(objects)
+                .map(object -> {
+                    if (object == null) {
+                        return null;
+                    }
+                    BitSet result = new BitSet();
```
hatyo, line 203:
```suggestion
                    final var result = new BitSet();
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -202,6 +235,28 @@ void testTupleSumCount() {
         });
     }
 
+    @Nullable
+    public static List<Long> collectOnBits(@Nullable byte[] bitmap) {
+        if (bitmap == null) {
+            return null;
+        }
+        List<Long> result = new ArrayList<>();
```
hatyo, line 243:
```suggestion
        final List<Long> result = new ArrayList<>();
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -100,6 +103,19 @@ void testSumOnlyNulls() {
         accumulateAndAssert(new NumericAggregationValue.Sum(PhysicalOperator.SUM_D, ofScalar(1D)), doublesOnlyNull, (Object)null);
     }
 
+    @Test
+    void testBitMap() {
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), bitsetForBitMap(new Object[]{0L, 1L, 2L, 0L}), Arrays.asList(0L, 1L, 2L)); // 111
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), bitsetForBitMap(new Object[]{0L, 1L, 2L, 0L, 64L, 65L, 66L}), Arrays.asList(0L, 1L, 2L, 64L, 65L, 66L)); // 111
+        accumulateAndAssertByteArray(new NumericAggregationValue.BitMap(PhysicalOperator.BITMAP_LL, ofScalar(1)), bitsetForBitMap(new Object[]{0L, 1L, 2L, 0L, 64L, 65L, 66L, 10000L, 10001L, 10100L}), Arrays.asList(0L, 1L, 2L, 64L, 65L, 66L, 10000L, 10001L, 10100L)); // 111
```
hatyo, line 110: Does this test the endianness as well? could you please make sure this is covered as well? It looks to me, in the assertion logic, you're iterating over the bytes and returning a list of set bits as long, that seems to glance over endianness's requirement.

Bottom line is, we must make sure the order of returned bytes of the streaming aggregate version of `bitmap` is aligned with what the indexed version of it returns.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/ArithmeticValueTest.java
```diff
@@ -110,13 +135,26 @@ public Stream<? extends Arguments> provideArguments(final ExtensionContext conte
...
+                    Arguments.of(List.of(LONG_10000, INT_10000), new ArithmeticValue.BitMapBitPositionFn(), 0L, false),
+                    Arguments.of(List.of(LONG_minus_1, INT_10000), new ArithmeticValue.BitMapBitPositionFn(), 9999L, false),
 
                     Arguments.of(List.of(INT_1, LONG_1), new ArithmeticValue.AddFn(), 2L, false),
                     Arguments.of(List.of(INT_1, LONG_2), new ArithmeticValue.SubFn(), -1L, false),
                     Arguments.of(List.of(INT_1, LONG_1), new ArithmeticValue.MulFn(), 1L, false),
                     Arguments.of(List.of(INT_1, LONG_2), new ArithmeticValue.DivFn(), 0L, false),
                     Arguments.of(List.of(INT_1, LONG_2), new ArithmeticValue.ModFn(), 1L, false),
 
+
```
hatyo, line 157: undo.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/ArithmeticValueTest.java
```diff
@@ -76,12 +84,29 @@ public Stream<? extends Arguments> provideArguments(final ExtensionContext conte
...
+                    Arguments.of(List.of(INT_minus_1, INT_10000), new ArithmeticValue.BitMapBucketOffsetFn(), -10000, false),
+                    Arguments.of(List.of(INT_5, INT_2), new ArithmeticValue.BitMapBucketNumberFn(), 2, false),
+                    Arguments.of(List.of(INT_1, INT_2), new ArithmeticValue.BitMapBucketNumberFn(), 0, false),
+                    Arguments.of(List.of(INT_1, INT_10000), new ArithmeticValue.BitMapBucketNumberFn(), 0, false),
+                    Arguments.of(List.of(INT_10000, INT_10000), new ArithmeticValue.BitMapBucketNumberFn(), 1, false),
+                    Arguments.of(List.of(INT_minus_1, INT_10000), new ArithmeticValue.BitMapBucketNumberFn(), -1, false),
+                    Arguments.of(List.of(INT_1, INT_2), new ArithmeticValue.BitMapBitPositionFn(), 1, false),
+                    Arguments.of(List.of(INT_1, INT_10000), new ArithmeticValue.BitMapBitPositionFn(), 1, false),
+                    Arguments.of(List.of(INT_10000, INT_10000), new ArithmeticValue.BitMapBitPositionFn(), 0, false),
+                    Arguments.of(List.of(INT_minus_1, INT_10000), new ArithmeticValue.BitMapBitPositionFn(), 9999, false),
```
hatyo, line 100: We should probably add some tests for `null`s as well.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
hatyo: I don't think I understand this comment.

hatyo (commented):
Nice work, few more comments and we can bring it in 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +674,33 @@ public enum PhysicalOperator {
         MAX_I(LogicalOperator.MAX, TypeCode.INT, TypeCode.INT, Objects::requireNonNull, (s, v) -> Math.max((int)s, (int)v), identity()),
         MAX_L(LogicalOperator.MAX, TypeCode.LONG, TypeCode.LONG, Objects::requireNonNull, (s, v) -> Math.max((long)s, (long)v), identity()),
         MAX_F(LogicalOperator.MAX, TypeCode.FLOAT, TypeCode.FLOAT, Objects::requireNonNull, (s, v) -> Math.max((float)s, (float)v), identity()),
-        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity());
+        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity()),
+        BITMAP_LL(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
+                s -> {
+                    BitSet sset = new BitSet();
```
hatyo, line 680: I think we have a discrepancy between this and what the index maintainer returns in case of using an indexed version of the bitmap.

If the bit bucket size is for example 1000, and we have only the first bit set:
- the indexed bitmap version will return 1000 bits with the first bit set.
- the streaming aggregate version will return only 1 bit.

I do not think this is ok. OTOH, I am not exactly sure how to fix it, is it e.g. possible to pre-allocate the bitset with a specific size?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -208,6 +210,75 @@ private static Map<Pair<LogicalOperator, TypeCode>, PhysicalOperator> computeOpe
         return mapBuilder.build();
     }
 
+    /**
+     * BitMap aggregation {@code Value}.
+     */
+    public static class BitMap extends NumericAggregationValue implements StreamableAggregateValue, IndexableAggregateValue {
```
hatyo, line 216: Can you please rename this to the more standard-ish `BitmapConstructAgg[Value]`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -210,24 +212,51 @@ default <V extends TreeLike<V>> Optional<V> mapMaybe(@Nonnull final BiFunction<T
         return Optional.ofNullable(foldNullable(Function.identity(), foldFunction));
     }
 
+    /**
```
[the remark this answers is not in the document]
pengpeng-lu: rebased main

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
pengpeng-lu: yeah this was a copy paste error. I removed the comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
pengpeng-lu: My proposal is, for the result byte array:
```
if res.length > BitmapValueIndexMaintainer.MAX_ENTRY_SIZE {
    throw ex;
} else if (res.length > BitmapValueIndexMaintainer.DEFAULT_ENTRY_SIZE {
    return res;
} else {
    return Arrays.copyOf(res, DEFAULT_ENTRY_SIZE);
}
```
wdyt?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
pengpeng-lu: I think the `collectOnBits` includes the endianness, if the endianness is incorrect, it wouldn't match the expected result.
```
for (int i = 0; i < bitmap.length; i++) {
            if (bitmap[i] != 0) {
                for (int j = 0; j < 8; j++) {
                    if ((bitmap[i] & (1 << j)) != 0) {
                        result.add(i * 8L + j);
                    }
                }
            }
        }
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java
```diff
@@ -138,7 +138,7 @@ public ArithmeticValue withChildren(final Iterable<? extends Value> newChildren)
     public int hashCodeWithoutChildren() {
         return PlanHashable.objectsPlanHash(PlanHashable.CURRENT_FOR_CONTINUATION, BASE_HASH, operator);
     }
-
```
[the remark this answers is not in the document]
hatyo: please revert white space changes here and below.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ValueMatchers.java
```diff
@@ -120,6 +120,14 @@ public static <V extends Value> BindingMatcher<NumericAggregationValue.Sum> sumA
                 downstream);
     }
 
+    @Nonnull
+    public static <V extends Value> BindingMatcher<NumericAggregationValue.BitMapConstructAgg> bitmapAggregationValue(@Nonnull final BindingMatcher<V> downstream) {
```
hatyo, line 124:
```suggestion
    public static <V extends Value> BindingMatcher<NumericAggregationValue.BitMapConstructAgg> bitMapConstructAggValue(@Nonnull final BindingMatcher<V> downstream) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
hatyo: I don't think there is necessarily a good solution here, unless we control the DDL/DQL completely, because it is possible to create an aggregate bitmap index using a custom bucket size with:

https://github.com/FoundationDB/fdb-record-layer/blob/fb3c634e4c1bb8101873f1e9ecb2d8fa7aa16538/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/IndexOptions.java#L157

therefore, it is impossible to align these, so your solution is what I would do as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/serialization/PlanSerialization.java
```diff
@@ -239,6 +239,9 @@ public static <E1 extends Enum<E1>, E2 extends Enum<E2> & ProtocolMessageEnum> B
             final E1 e1 = javaEnumConstants[i];
             final E2 e2 = protoEnumConstants[i];
             Verify.verify(e1.ordinal() == e2.ordinal());
+            if (!e1.name().equals(e2.name())) {
+                System.out.println("e1:" + e1.name() + "e2:" + e2.name());
+            }
```
hatyo, line 244: remove.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
```diff
@@ -178,6 +196,23 @@ private Object[] pairsForAvg(Object[] objects) {
                 .toArray();
     }
 
+    private Object[] bitsetForBitMap(Object[] objects) {
```
hatyo, line 199:
```suggestion
    @Nonnull
    private Object[] bitsetForBitMap(@Nonnull Object[] objects) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
```diff
@@ -417,7 +417,7 @@ protected <T> List<T> fetchResultValues(FDBRecordContext context, RecordQueryPla
     }
 
     protected <T> List<T> fetchResultValues(FDBRecordContext context, RecordQueryPlan plan, Function<Message, T> rowHandler,
-                                        TestHelpers.DangerousConsumer<FDBRecordContext> checkDiscarded, ExecuteProperties executeProperties) throws Exception {
+                                            TestHelpers.DangerousConsumer<FDBRecordContext> checkDiscarded, ExecuteProperties executeProperties) throws Exception {
```
hatyo, line 420: revert.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
```diff
@@ -163,7 +184,7 @@ public void testIndexPlanningWithPredicateInSelectWhereAndSelectHavingMatchesAgg
                 Optional.empty(),
                 IndexQueryabilityFilter.TRUE,
                 EvaluationContext.empty()).getPlan();
-
+
```
hatyo, line 187: please revert.

hatyo (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
pengpeng-lu: So I think we actually want the white space here. The original code looks like this: https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java#L419 the indentation doesn't look correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +677,53 @@ public enum PhysicalOperator {
...
+                s -> {
+                    int fixedResultArraySize = BitmapValueIndexMaintainer.DEFAULT_ENTRY_SIZE / 8;
+                    byte[] res = ((BitSet)s).toByteArray();
+                    if (res.length > BitmapValueIndexMaintainer.MAX_ENTRY_SIZE / 8) {
+                        throw new RecordCoreArgumentException("entry size option is too large")
+                                .addLogInfo("entrySize", res.length * 8, "maxEntrySize", BitmapValueIndexMaintainer.MAX_ENTRY_SIZE);
+                    } else if (res.length > fixedResultArraySize) {
+                        return res;
+                    } else {
+                        return Arrays.copyOf(res, fixedResultArraySize);
```
hatyo, line 701:
```suggestion
                        return res;
```

reason: I guess this can be inefficient, because it will create a `10000` bitmap for _each_ element in the group, OTOH I am not 100% sure this works, so you might want to debug it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +677,53 @@ public enum PhysicalOperator {
...
```
hatyo, line 724: This, OTOH, should stay as written.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBRecordStoreQueryTestBase.java
hatyo: Got it, thank you.

hatyo (approved):
LGTM.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -526,7 +611,8 @@ private enum LogicalOperator {
         SUM,
         AVG,
         MIN,
-        MAX
+        MAX,
+        BITMAP
```
hatyo, line 615:
```suggestion
        BITMAP_CONSTRUCT_AGG
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +677,53 @@ public enum PhysicalOperator {
...
+                    if (res.length > BitmapValueIndexMaintainer.MAX_ENTRY_SIZE / 8) {
+                        throw new RecordCoreArgumentException("entry size option is too large")
+                                .addLogInfo("entrySize", res.length * 8, "maxEntrySize", BitmapValueIndexMaintainer.MAX_ENTRY_SIZE);
+                    } else if (res.length > fixedResultArraySize) {
+                        return res;
+                    } else {
+                        return Arrays.copyOf(res, fixedResultArraySize);
+                    }
+                }),
+        BITMAP_I(LogicalOperator.BITMAP, TypeCode.INT, TypeCode.BYTES,
```
hatyo, line 704:
```suggestion
        BITMAP_CONSTRUCT_AGG_I(LogicalOperator.BITMAP, TypeCode.INT, TypeCode.BYTES,
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -591,7 +677,53 @@ public enum PhysicalOperator {
         MAX_I(LogicalOperator.MAX, TypeCode.INT, TypeCode.INT, Objects::requireNonNull, (s, v) -> Math.max((int)s, (int)v), identity()),
         MAX_L(LogicalOperator.MAX, TypeCode.LONG, TypeCode.LONG, Objects::requireNonNull, (s, v) -> Math.max((long)s, (long)v), identity()),
         MAX_F(LogicalOperator.MAX, TypeCode.FLOAT, TypeCode.FLOAT, Objects::requireNonNull, (s, v) -> Math.max((float)s, (float)v), identity()),
-        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity());
+        MAX_D(LogicalOperator.MAX, TypeCode.DOUBLE, TypeCode.DOUBLE, Objects::requireNonNull, (s, v) -> Math.max((double)s, (double)v), identity()),
+        BITMAP_L(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
```
hatyo, line 681:
```suggestion
        BITMAP_CONSTRUCT_AGG_L(LogicalOperator.BITMAP, TypeCode.LONG, TypeCode.BYTES,
```

hatyo (changes requested):

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -498,10 +506,16 @@ message PNumericAggregationValue {
       MAX_L = 14;
       MAX_F = 15;
       MAX_D = 16;
+
+      BITMAP_L = 17;
```
hatyo, line 510:
```suggestion
      BITMAP_CONSTRUCT_AGG_L = 17;
```

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -498,10 +506,16 @@ message PNumericAggregationValue {
...
+      BITMAP_I = 18;
```
hatyo, line 511:
```suggestion
      BITMAP_CONSTRUCT_AGG_I = 18;
```

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -498,10 +506,16 @@ message PNumericAggregationValue {
...
   }
   optional PPhysicalOperator operator = 1;
   optional PValue child = 2;
 
+  message PBitMap {
```
hatyo, line 516: PBitMapConstructAgg

hatyo, line 516:
```suggestion
  message PBitMapConstructAgg {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/BitmapAggregateIndexExpansionVisitor.java
```diff
@@ -0,0 +1,126 @@
...
+        final var pulledUpGroupingValuesMap = selectQunValue.pullUp(groupingValues, aliasMap, ImmutableSet.of(), selectWhereQun.getAlias());
+        final var explicitPulledUpGroupingValues = groupingValues.stream().map(groupingValue -> {
+            if (!pulledUpGroupingValuesMap.containsKey(groupingValue)) {
+                throw new RecordCoreException("could not pull grouping value " + groupingValue)
+                        .addLogInfo(LogMessageKeys.VALUE, groupingValue);
+            }
+            return pulledUpGroupingValuesMap.get(groupingValue);
+        }).collect(ImmutableList.toImmutableList());
+
+        final var pulledUpGroupingValues = ImmutableList.<Value>builder().addAll(explicitPulledUpGroupingValues).add(implicitGroupingValue).build();
```
hatyo, line 120:
```suggestion
        final var pulledUpGroupingValues = ImmutableList.<Value>builder().add(implicitGroupingValue).addAll(explicitPulledUpGroupingValues).build();
```

hatyo (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/BitmapAggregateIndexExpansionVisitor.java
```diff
@@ -0,0 +1,126 @@
...
+                    groupedValue.getCorrelatedTo()));
+            final var result = selectWhereQun.getRangesOver().get().getResultValue()
+                    .pullUp(List.of(groupedValue), aliasMap, ImmutableSet.of(), selectWhereQun.getAlias());
+            if (!result.containsKey(groupedValue)) {
+                throw new RecordCoreException("could not pull grouped value " + groupedValue)
+                        .addLogInfo(LogMessageKeys.VALUE, groupedValue);
+            }
+            argument = result.get(groupedValue);
+        } else {
+            throw new RecordCoreException("unable to plan group by with non-field value")
```
normen662, line 85: If the previous exception is `UnsupportedOperationException` I would say this one should be as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexMatchCandidate.java
```diff
@@ -364,10 +364,16 @@ public List<RecordType> getQueriedRecordTypes() {
         return recordTypes;
     }
 
+    protected int getGroupingCount() {
+        return IndexTypes.BITMAP_VALUE.equals(index.getType())
```
normen662, line 368: please introduce something like:

```
final int keyExpressionGroupingCount = ((GroupingKeyExpression)index.getRootExpression()).getGroupingCount();
```

as a local variable and then use it in the `?`.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -241,6 +241,7 @@ message PValue {
     PToOrderedBytesValue to_ordered_bytes_value = 42;
     PFromOrderedBytesValue from_ordered_bytes_value = 43;
     PCollateValue collate_value = 44;
+    PNumericAggregationValue.PBitMapConstructAgg numeric_aggregation_value_bitmap = 45;
```
normen662, line 244:
```suggestion
    PNumericAggregationValue.PBitMapConstructAgg numeric_aggregation_value_bitmap_construct_agg = 45;
```

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -522,10 +530,16 @@ message PNumericAggregationValue {
...
       MAX_F = 15;
       MAX_D = 16;
+
+      BITMAP_CONSTRUCT_AGG_L = 17;
+      BITMAP_CONSTRUCT_AGG_I = 18;
   }
   optional PPhysicalOperator operator = 1;
   optional PValue child = 2;
 
+  message PBitMapConstructAgg {
```
normen662, line 540:
```suggestion
  message PBitmapConstructAgg {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java
```diff
@@ -290,6 +290,28 @@ public ModFn() {
         }
     }
 
+    /**
+     * The {@code bitmap_bit_position} function.
+     */
+    @AutoService(BuiltInFunction.class)
+    public static class BitMapBitPositionFn extends BuiltInFunction<Value> {
```
normen662, line 297:
```suggestion
    public static class BitmapBitPositionFn extends BuiltInFunction<Value> {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java
```diff
@@ -290,6 +290,28 @@ public ModFn() {
...
+            super("bitmap_bit_position",
+                    ImmutableList.of(Type.any(), Type.any()), ArithmeticValue::encapsulateInternal);
+        }
+    }
+
+    /**
+     * The {@code bitmap_bucket_number} function.
+     */
+    @AutoService(BuiltInFunction.class)
+    public static class BitMapBucketNumberFn extends BuiltInFunction<Value> {
```
normen662, line 308: bitmap is a word -- please do a global find/replace with your changes.

normen662 (changes requested):
Just cosmetics.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
normen662: You are right. However, please don't uncase the `m` in `bitmap`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexMatchCandidate.java
pengpeng-lu: 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java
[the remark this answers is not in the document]
pengpeng-lu: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
pengpeng-lu: discussed offline and modified the code.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
pengpeng-lu: has pulled the fix and the problem was fixed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/AggregateValueTest.java
pengpeng-lu: so the old planner was using this function in test, this could guarantee that we behave exactly like the older planner.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
pengpeng-lu: The bits in the bitmap are little-endian. But the byte order isn't reversed. Which means 1,2,3 is [0x0E, ...(124 zeros)].

hatyo (approved):
LGTM.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
normen662: This is not resolved properly.

normen662: This is not resolved.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
pengpeng-lu: changed to `BitmapConstructAgg`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/GroupByTest.java
pengpeng-lu: I changed it to `.and(groupings(recordConstructorValue(exactly(anyValue(), anyValue())))))));` but replacing `anyValue()` with `ValueMatchers.findFieldValueWithNames("str_value_indexed")` doesn't work as I expected it should be.

normen662: Fixed.

normen662 (approved):

Issue #2745, opened 2024-05-29
change modeling of `OrderingPart` to use a sort order instead of a boolean for reversedness


normen662:
This is related to #1318. It would be useful to state that certain ordering parts are fixed, i.e. bound to one value vie an e.g. equality predicate. In those cases forward and backward become meaningless, they would both fit.

This issue addresses replacing the `isReverse` with a new enum `SortOrder` which can be `ASCENDING`, `DESCENDING`, and `FIXED`.

There are some places in the codebase where we reason about `Ordering` and in particular `OrderingPart`s where we rely on the equality of `OrderingPart`s. An ordering part `(q.a, ASC)` is equal to `(q.a, ASC)` only, in particular it is not equal to `(q.a, DESC)`, `(q.a, FIXED)`.

 Unfortunately, a `RequestedOrdering` which is used to enumerate satisfying orderings does not want equality but just compatibility, i.e. a requested ordering of `(q.a, ASC)` or `(q.a, DESC)` are both compatible with `(q.a, FIXED)` even though they are not equal. All of this is part of `Ordering.enumerateSatisfyingOrderings()` and `Ordering.enumerateSatisfyingOrderingComparisonKeyValues()`.

Similarly, merging two `Ordering`s has a similar problem. An ordering `(q.a, ASC)` and an ordering `(q.a, FIXED)` should be merged to `(q.a, ASC)`. The code that needs to be adapted is in `Ordering.mergePartialOrderOfOrderings()` which (again) only considers equality instead of compatibility.

alecgrieser:
As a note, having a separate `FIXED` ordering may help with #2723, though that issue is exacerbated by some other problems with how values are updated during `pullUp` that we probably want to fix, so maybe we deal with that separately.

PR #2748, opened 2024-06-05
Resolves #2127, #2745: Sort order in ordering parts

### LINKED ISSUE #2127: Planner can choose suboptimal plans for reverse sorts with some values bound in equality predicates
The planner currently doesn't match reverse sorts correctly if there are some values bound with equality predicates. For example, if there is a query like:

```java
RecordQuery.newBuilder()
    .setFilter(Query.field("a").equalsValue("foo"))
    .setSort(Key.Expressions.concatenateFields("a", "b"), /* reverse = */ true)
    .build()
```

Then if there's an index `ab` on `[a, b]`, the query should be able to execute a reverse scan on that one index, with the `a` field bound to `foo`. Something like:

```
Index(ab [[foo],[foo]] REVERSE)
```

However, the Cascades planner instead chooses a different plan:

```
Index(ab <,> REVERSE) | [a EQUALS "foo"]
```

That is, it scans the entire `ab` index and then executes the `a` filter as a residual predicate.

Note that if the sort order is not reversed, then the better plan is selected. The error appears to be in the way `RemoveSortRule` determines if the ordering has been satisfied. It will eventually produce the current ordering, which includes logic to try and handle that equality-bound predicates can appear in any place and it's okay, from an ordering perspective:

https://github.com/FoundationDB/fdb-record-layer/blob/380534fa0995790763cfefb612dd6f6e2eb995b1/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java#L384-L396

But it uses the default constructor for an `OrderingPart`, which assumes an ascending order. This then causes problems when trying to match the ordering parts, which will be descending for reverse sorts. (This is why it works for forward order, but not for reverse order.)

normen662:
- changes how an `Ordering` is modeled.  The partially ordered set that backs an `Ordering` is now based on `Value`s instead of `OrderingPart`s. The equality-bound key map is changed to a binding map (set multimap) that maps from values to `Binding`s.  A `Binding` can be `ASCENDING`, `DESCENDING` or `FIXED` (in which case the value is fixed to a constant value).
- orderings differiantiate sort orders by use case: `Ordering` computes the ordering of a plan or a thing that provides order (this the sort order it computes uses `ProvidedSortOrder`; matching yields `MatchedSortOrder`; and `RequestedOrdering` uses `RequestedSortOrder`
- set operations: separation of the different stages of computing the ordering of a union/intersection/in-union/in-join plan
- changes in `AbstractDataAccessRule` that prepare for inverted index keys
- logic dealing orderings use a lot of sanity checks that flushed out many problems
- this PR contains many fixes for related or not so related but exposed problems


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
```diff
@@ -25,66 +25,142 @@
...
+ *         {@link Value}s that do not contribute to the ordering of the records in the result set of an operation
+ *         are not contained in the partially-ordered set nor in its dependency map.
+ *     </li>
+ * </ul>
+ * Using a partially-ordered set, we can enumerate all traditional orderings by enumerating all topologically-sound
+ * permutations of that partially-ordered set.
+ * <br>
+ * In addition to jst using a {@link PartiallyOrderedSet} of {@link Value}s, we also keep a multimap for the domain of
+ * those {@link Value}s to {@link Binding}s. A {@link Binding} indicates if a {@link Value} is ascending, descending,
+ * of if it is bound to one or multiple special {@link Value} via a {@link Comparison}.
```
MMcM, line 111:
```suggestion
 * or if it is bound to one or multiple special {@link Value} via a {@link Comparison}.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
```diff
@@ -93,50 +169,70 @@ public class Ordering {
...
     @Nonnull
-    private final PartiallyOrderedSet<OrderingPart> orderingSet;
+    private final PartiallyOrderedSet<Value> orderingSet;
 
+    /**
+     * Indicator if the records flowed are to be considered distinct, thus producing a strict order. This field
+     * should get deprecated as it is not correct to assign this property to all enumerated orderings. For instance,
+     * an ordering that produces {@code a, b, c} is also ordered by {@code a, b}. While the former one might be strict,
+     * the latter one may not. I think for now we should just interpret this indicator in a way that only enumerated
+     * orderings that contain all values of the ordering are strict if this indicator is {@code true}.
```
MMcM, line 188: Do we ever have cause to enumerate a _longer_ orderings? With `a, b, c` distinct, `a, b, c, d` is also available.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/OrderingPart.java
```diff
@@ -76,35 +91,353 @@ public int hashCode() {
...
     }
+
+    /**
+     * A common interface all sort orders have to implement. Although, all sort orders offer enums for ascending and
+     * descending order and are thus somewhat overlapping in nature, different use cases may warrant subtle differences.
+     * For instance, a {@link ProvidedSortOrder} uses a {@link ProvidedSortOrder#FIXED} to indicate that an item is
+     * a fixed value (effectively a constant), however, a {@link RequestedSortOrder} does not support to request for
+     * fixed values. Conversely, a {@link RequestedSortOrder} can indicate {@link RequestedSortOrder#ANY} which
+     * indicates a sort order that is either ascending or descending (but not nothing). In an earlier iteration of this
+     * logic all sort orders where in one enum which led to problems as unexpected sort orders enum values were passed
```
MMcM, line 122:
```suggestion
     * logic all sort orders were in one enum which led to problems as unexpected sort orders enum values were passed
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/OrderingPart.java
```diff
@@ -76,35 +91,353 @@ public int hashCode() {
...
+     * to logic that was not able to digest it. Making it separate classes based on use case allows us to have the Java
+     * compiler to ensure that only meaningful sort orders are processed by the right logic.
```
MMcM, line 124:
```suggestion
     * compiler ensure that only meaningful sort orders are processed by the right logic.
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
```diff
@@ -79,7 +79,11 @@ public PlanContext getPlanContext() {
 
     @Override
     public boolean isSane() {
-        return true;
+        //
+        // Report insanity here which then causes all sanity checks to be run which may be CPU-intensive. Deactivate
+        // this behavior by returning true if performance is measured.
```
MMcM, line 84: Should this have some kind of environmental control then?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
normen662: No, we do not consider that. The reason is that strictly ordering should be ordered as a separate property but is not (yet)! In the future we can utilize that to also enumerate what you are suggesting. Also, I think only the next key can be added.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
normen662: I wanted to have this controlled vi a java property. Not done yet.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/AggregateIndexMatchCandidate.java
```diff
@@ -220,10 +226,13 @@ public List<MatchedOrderingPart> computeMatchedOrderingParts(@Nonnull final Matc
 
             // Grab the value for this sortParameterID from the selectHaving result columns
             final var value = deconstructedValue.get(permutedIndex).rebase(aliasMap);
-            builder.add(
-                    MatchedOrderingPart.of(value,
-                            comparisonRange == null ? ComparisonRange.Type.EMPTY : comparisonRange.getRangeType(),
-                            isReverse));
+
+            if (!normalizedValues.contains(value)) {
+                normalizedValues.add(value);
```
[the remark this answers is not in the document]
normen662: Changed for all match candidates.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
```diff
@@ -25,66 +25,142 @@
...
+ *     </li>
+ *     <li>
+ *          A result set satisfying an ordering of {@code a, b, c} where {@code a} is bound to a e.g. value {@code 3}
+ *          is automatically also ordered by {@code b, a, c} and {@code b, c, a} (in addition to the orderings above).
+ *          In a sense the ordering part for {@code a} can be freely moved withing the ordering (as long as {@code b}
+ *          comes before {@code c}.
+ *     </li>
+ *     <li>
+ *          A result set satisfying an ordering of {@code a, b, c} where {@code d := 2 * b} is automatically also
+ *          ordered by {@code a, d, c} (in addition to the orderings above).
```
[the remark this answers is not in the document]
normen662: Right. Yes!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
```diff
@@ -25,66 +25,142 @@
...
+ *     </li>
+ *     <li>
+ *         {@link Value}s that do not contribute to the ordering of the records in the result set of an operation
+ *         are not contained in the partially-ordered set nor in its dependency map.
+ *     </li>
+ * </ul>
+ * Using a partially-ordered set, we can enumerate all traditional orderings by enumerating all topologically-sound
+ * permutations of that partially-ordered set.
+ * <br>
+ * In addition to jst using a {@link PartiallyOrderedSet} of {@link Value}s, we also keep a multimap for the domain of
```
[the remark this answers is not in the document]
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/OrderingPart.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/OrderingPart.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Ordering.java
normen662: done.

MMcM (approved):

PR #2751, opened 2024-06-06
Fixes #2750: Some calls to toUpperCase and toLowerCase are missing fixed Locale

### LINKED ISSUE #2750: Some calls to toUpperCase and toLowerCase are missing fixed Locale
There are a few instances where we call `.toUpperCase` or `.toLowerCase` on a string, but we're missing the locale specifier. Amongst other things, this can result in errors if running on a system configured with the Turkish locale, which has different handling for `"I".toLowerCase()` and `"i".toUpperCase()` from the root locale. This can lead to errors for example when trying to encapsulate the "multiply" function, ultimately stemming from this line:

https://github.com/FoundationDB/fdb-record-layer/blob/9d65c0aa81100543cfc5215482767a747e84594b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArithmeticValue.java#L208

The `Enum` can return an empty optional if given the string "multiply".

alecgrieser:
This updates a few calls to `toUpperCase()` and `toLowerCase()` to specify `Locale.ROOT` instead of `Locale.getDefault()`. This fixes an issue where different handling of `"i".toUpperCase()` or `"I".toLowerCase()` were resulting in the `ArithmeticValue` to not successfully encapsulate multiplication. For the record, it turns out we're already using a PMD rule that forces the user to specify a `Locale`, but it only checks that a locale was specified, not that the `Locale` is a specific value. Also, the PMD checks don't run on test code, which is why those classes in test logic were not specifying a locale at all.

This fixes #2750.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -17,7 +17,7 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** The `ArithmeticValue` and `NumericAggregationValue` classes now uses a fixed `Locale` with `toUpperCase` when encapsulating a specified function by name to avoid mismatches when running in the Turkish locale [(Issue #2750)](https://github.com/FoundationDB/fdb-record-layer/issues/2750)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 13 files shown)


docs/ReleaseNotes.md
```diff
@@ -17,7 +17,7 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 // begin next release
 ### NEXT_RELEASE
 
-* **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** The `ArithmeticValue` and `NumericAggregationValue` classes now uses a fixed `Locale` with `toUpperCase` when encapsulating a specified function by name to avoid mismatches when running in the Turkish locale [(Issue #2750)](https://github.com/FoundationDB/fdb-record-layer/issues/2750)
```
alecgrieser, line 20: These changes are highlighted because it's the only part that should be API-visible (in that if someone was running with a Turkish locale and specified the function name `"multıply"`, they'll need to change it to `"multiply"`, but it seems unlikely that someone would intentionally do the former). The rest of the changes should only affect logging or tests

PR #2770, opened 2024-06-21
Resolve #2765: Index Scrubbing: support cleanup when dangling index entries are expected

### LINKED ISSUE #2765: Index Scrubbing: support cleanup when dangling index entries are expected
On some databases, as it seems, dangling index entries are part of the normal flow. To support routine cleanup of these entries, we should add: 
1. `OnlineIndexScrubber` API argument to start from scratch (to avoid continuation of previous sessions) 
2. `OnlineIndexScrubber` API argument to limit the number of deletes per time unit (to avoid FDB flooding)

With the existing way to avoid warning, the index scrubbing can be used as a vacuum method to remove expected dangling index entries.

jjezra:
Support users who wish to use a "lazy delete" - which clears the record in foreground, then let a background index scrubbing task clear the dangling index entries. This requires: 
1. Set full scan (new API) 
2. Set deletes/second limit (new API - required to avoid flooding the DB) 
3. Avoid warning (existing API)


ScottDugas:
`:fdb-record-layer-core:pmdMain` failed

PR #2771, opened 2024-06-23
Add SQL-like ordering directions

MMcM:
Lacks Cascades support.


fdb-extensions/src/test/java/com/apple/foundationdb/tuple/TupleOrderingTest.java
```diff
@@ -0,0 +1,102 @@
...
+                    () -> ByteArrayUtil2.toHexString(uninverted) + " != " + ByteArrayUtil2.toHexString(expected) + " for " + ByteArrayUtil2.toHexString(inverted));
+        }
+    }
+
+    private static final Object[] INVERT_OBJS = {
+        null,
+        new byte[] { }, new byte[] { 0x01 }, new byte[] { 0x01, 0x02 }, new byte[] { (byte)0xFF },
+        "", "a", "ab", "abc", "abcd", "abcde", "abcdef", "abcdefg", "ac",
+        Long.MIN_VALUE, -2, -1, 0, 1, 2, Long.MAX_VALUE,
+        -1e10, -2.0, -1.0, 0.0, 2.0, 1e10,
```
alecgrieser, line 80: In theory, this could include `-Infinity`, `-0.0`, `Infinity`, and a few positive and negative NaN values, just to cover bases. None of those should be particularly difficult cases for this algorithm, though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * {@code ORDER_xxx} function.
+ *
+ */
+@API(API.Status.EXPERIMENTAL)
+public class OrderFunctionKeyExpression extends FunctionKeyExpression implements QueryableKeyExpression {
```
alecgrieser, line 46: I think this could implement `InvertibleFunctionKeyExpression`. At the moment, we don't too too much with that, but I think that will be the building block needed to do things like extract the original values during a covering index scan. I suppose this could be done in a follow up PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
```diff
@@ -0,0 +1,114 @@
...
+    public int getMaxArguments() {
+        return 1;
+    }
+
+    @Nonnull
+    @Override
+    public <M extends Message> List<Key.Evaluated> evaluateFunction(@Nullable FDBRecord<M> record,
+                                                                    @Nullable Message message,
+                                                                    @Nonnull Key.Evaluated arguments) {
+        return Collections.singletonList(Key.Evaluated.scalar(ByteString.copyFrom(TupleOrdering.pack(arguments.toTuple(), direction))));
```
alecgrieser, line 78: This should probably be `ZeroCopyByteString.wrap` instead of `ByteString.copyFrom`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
```diff
@@ -0,0 +1,114 @@
...
+    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias,
+                         @Nonnull final List<String> fieldNamePrefix) {
+        // TODO need inner Value for match and Ordering info from direction.
+        throw new UnsupportedOperationException();
+    }
+
+    @Nullable
+    @Override
+    public Function<Object, Object> getComparandConversionFunction() {
+        return o -> ByteString.copyFrom(TupleOrdering.pack(Tuple.from(o), direction));
```
alecgrieser, line 102: This should also be `ZeroCopyByteString.wrap` instead of `ByteString.copyFrom`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/OrderQueryKeyExpression.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    @Nullable
+    public QueryKeyExpressionWithComparison adjustComparison(@Nonnull Comparisons.Comparison comparison) {
+        final boolean inverted = ((OrderFunctionKeyExpression)keyExpression).getDirection().isInverted();
+        Comparisons.Type type = comparison.getType();
+        switch (type) {
+            case EQUALS:
+            case NOT_EQUALS:
+                break;
+            case LESS_THAN:
```
alecgrieser, line 54: Does this also need to do something to handle nulls? For example, if the query comparison is `< "foo"`, I believe this gets translated to a query from `> null, < "foo"`, which would may need to be adjusted if the ordering is contraflows nulls

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrderingQueryTest.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    @Test
+    void testFilteredAsc() throws Exception {
+        sortAndFilter("order_asc_nulls_first", true, 4, 5);
+    }
+
+    @Test
+    void testFilteredDesc() throws Exception {
+        sortAndFilter("order_desc_nulls_last", true, 5, 4);
```
alecgrieser, line 188: It might be good to check `order_desc_nulls_first` with this predicate to make sure it doesn't return the `null` `str_value_indexed`. Likewise with `testFilteredAsc`, it would be good to check it with `order_asc_nulls_last`

alecgrieser (changes requested):

fdb-extensions/src/main/java/com/apple/foundationdb/tuple/TupleOrdering.java
```diff
@@ -0,0 +1,224 @@
...
+                inverted[out++] = (byte)((bits >> (nbits - 7)) & 0x7F);
+                nbits -= 7;
+            }
+        }
+        if (nbits == 0) {
+            inverted[out++] = (byte)0x80;
+        } else {
+            final int npad = 7 - nbits;
+            inverted[out++] = (byte)(((bits << npad) | ((1 << npad) - 1)) & 0x7F);
+            inverted[out++] = (byte)(0x80 | npad);
```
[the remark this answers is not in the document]
MMcM: I suppose there is no reason not to move it closer to make the significant bits more obvious.

fdb-extensions/src/test/java/com/apple/foundationdb/tuple/TupleOrderingTest.java
MMcM: Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
MMcM: Implemented.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
MMcM: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/OrderFunctionKeyExpression.java
MMcM: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/OrderQueryKeyExpression.java
MMcM: Ah, right, it does sometimes need a second comparison.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrderingQueryTest.java
MMcM: I just went ahead and made it test all cases.

alecgrieser (approved):

normen662 (approved):

PR #2777, opened 2024-06-25
rdar://129580618 ([SQL] Support Tuple IN LIST<Tuple>)

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
```diff
@@ -213,7 +213,6 @@ public InFn() {
         private static Value encapsulateInternal(@Nonnull final List<? extends Typed> arguments) {
             final Typed arg0 = arguments.get(0);
             final Type res0 = arg0.getResultType();
-            SemanticException.check(res0.isPrimitive(), SemanticException.ErrorCode.COMPARAND_TO_COMPARISON_IS_OF_COMPLEX_TYPE);
```
normen662, line 216: The following code never took a record type into account. Can you provide a test case that tests that you cannot do an e.g. 
```
(a, b) in (1, 2, 3)
(a, b) in ((1, 2, 3), (4, 5, 6))
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -242,8 +242,61 @@ private static Boolean compareEquals(Object value, Object comparand) {
...
+                return compareMessageEquals((Message)value, (Message)comparand);
+            } else {
+                return toClassWithRealEquals(value).equals(toClassWithRealEquals(comparand));
+            }
+        }
+    }
+
+    // return true if m1 and m2 are equal, i.e. if for all field index i, m1.getFieldByIndex(i) = m2.getFieldByIndex(i), m1 and m2 are equal
+    // m1 and m2 must only have primitive types (not repeated, map, or message type fields)
+    private static boolean compareMessageEquals(@Nonnull Message m1, @Nonnull Message m2) {
```
normen662, line 258: This code should bo to `MessageHelpers`. I am actually surprised that such a  method does not exist yet. It should also operate on ordinal positioning inside of the proto message as field indexes do not matter here.

normen662 (changes requested):
General comment. I would like to see some positive and negative test cases. Please also address the specific comments as well. How did you not have to change the in-rules?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/InOpValue.java
normen662, line 216: The following is not dealt with correctly: Let's say you have a table with longs `t.x, t.y, t,z)`. The encapsulation for this expression should promote the parts individually, i.e. the `(x.t, x.y, x.z) IN (1, 2, 3)` needs to be encapsulated into
`(x.t, x.y, x.z) IN (PROMOTE 1 TO LONG, PROMOTE 2 TO LONG, PROMOTE 2 TO LONG)`. Similarly, the opposite needs to happen in the inverse case. The code here was written for primitive types and I am almost sure it cannot handle the complex case.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -364,7 +364,7 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
             return fromProtoTypeToArray(descriptor, protoType, typeCode, false);
         } else {
             if (typeCode.isPrimitive()) {
-                return primitiveType(typeCode, isNullable);
+                return primitiveType(typeCode, false);
```
pengpeng-lu, line 367: at [Line#2081](https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java#L2081), `fromProtoType(...., isNullable = !fieldDescriptor.isRequired()`, conflicts with other places where primitive types are NOT nullable (isNullable = false), so I changed the logic here to make all primitiveTypes not nullable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -590,6 +596,9 @@ static Type fromObject(@Nullable final Object object) {
             }
             return new Type.Array(Type.fromListObject((List<?>)object));
         }
+        if (object instanceof DynamicMessage) {
+            return Record.fromDescriptor(((DynamicMessage) object).getDescriptorForType());
+        }
```
pengpeng-lu, line 601: This is change from Pranjal's PR: https://github.com/FoundationDB/fdb-record-layer/pull/2807/files#diff-0b60f460d35324d34656a59f5a7cc9b5660d4ee0f5a5c675946ba69f5ba07009

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -598,7 +607,7 @@ static Type fromObject(@Nullable final Object object) {
             return Type.any();
         }
         if (typeCode.isPrimitive()) {
-            return Type.primitiveType(typeCode, false);
+            return Type.primitiveType(typeCode, isNullable);
```
pengpeng-lu, line 610: before the change, any Type created by Type.fromObject, will have all primitive types isNullable = false. Add this method to allow callers to set isNullable.

pengpeng-lu: We have seen `promote(x as string)` before, it was because, initially x is set to `string, isNullable = false` by this `fromObject` method, and then it was promoted to `string, isNullable = true`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
```diff
@@ -419,6 +420,15 @@ public static boolean isPromotionNeeded(@Nonnull final Type inType, @Nonnull fin
...
         }
+        if (inType.isRecord() && inType.isRecord()) {
+            final List<Type> inTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
+            final List<Type> promoteToTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
+            SemanticException.check(inTypeElements.size() == promoteToTypeElements.size(), SemanticException.ErrorCode.INCOMPATIBLE_TYPE);
+            for (int i = 0; i < inTypeElements.size(); i++) {
+                SemanticException.check(!isPromotionNeeded(inTypeElements.get(i), promoteToTypeElements.get(i)), SemanticException.ErrorCode.INCOMPATIBLE_TYPE);
+            }
+            return false;
+        }
```
pengpeng-lu, line 431: change from Pranjal's PR

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -220,6 +220,8 @@ public static Object toClassWithRealEquals(@Nullable Object obj) {
             return obj;
         } else if (obj instanceof List) {
             return obj;
+        } else if (obj instanceof Message) {
```
normen662, line 223: This should not be in here as the class does not really have an `equals()` that we like. I changed the callers on my local build to special-case `Message` to always call the `MessageHelper` method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/ScanComparisons.java
```diff
@@ -330,6 +332,13 @@ public static Object toTupleItem(@Nullable Object item) {
             return ((Internal.EnumLite) item).getNumber();
         } else if (item instanceof FDBRecordVersion) {
             return ((FDBRecordVersion) item).toVersionstamp();
+        } else if (item instanceof DynamicMessage) {
```
normen662, line 335: This seems brittle as a nested message will fail. We need at least a check in here to make sure all the constituent parts are primitive. Also can you make sure that iterating using the keyset of the field descriptors will indeed give you back the field descriptors in definition-order?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
normen662: Why is that in here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -174,6 +174,30 @@ public static Descriptors.FieldDescriptor findFieldDescriptorOnMessageByOrdinal(
...
+    public static boolean compareMessageEquals(@Nonnull MessageOrBuilder m1, @Nonnull MessageOrBuilder m2) {
+        if (m1.getDescriptorForType().getFields().size() != m2.getDescriptorForType().getFields().size()) {
+            return false;
+        }
+        // compare FieldDescriptors one by one
+        for (int i = 1; i <= m1.getDescriptorForType().getFields().size(); i++) {
+            Descriptors.FieldDescriptor f1 = findFieldDescriptorOnMessage(m1, i);
+            Descriptors.FieldDescriptor f2 = findFieldDescriptorOnMessage(m2, i);
+            // do not support repeated or nested fields in the message
+            if (f1.isRepeated() || f1.isMapField() || f1.getJavaType() == Descriptors.FieldDescriptor.JavaType.MESSAGE
```
normen662, line 186: This should not blindly return `false` if you don't like what the caller gave you. This should do all these cases as `Verify.verify(...)`, i.e. this should fail! There should be a check during the construction of the QGM that makes sure that the tuple is not deeply nested, contains arrays, etc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/InComparisonToExplodeRule.java
```diff
@@ -148,24 +158,45 @@ public void onMatch(@Nonnull final CascadesRuleCall call) {
                 final var comparison = valuePredicate.getComparison();
                 Verify.verify(comparison.getType() == Comparisons.Type.IN);
                 final ExplodeExpression explodeExpression;
+
                 if (comparison instanceof Comparisons.ValueComparison) {
                     final var comparisonValue = (Comparisons.ValueComparison)comparison;
                     Verify.verify(comparisonValue.getComparandValue().getResultType().isArray());
-                    explodeExpression = new ExplodeExpression(comparisonValue.getComparandValue());
+                    Type arrayElementType = ((Type.Array) comparisonValue.getComparandValue().getResultType()).getElementType();
```
normen662, line 165: There is a ton of overlapping code here. Please refactor.

normen662 (changes requested):
I left comments. Please do write some tests. So this feature can be tested independently.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -242,7 +245,14 @@ private static Boolean compareEquals(Object value, Object comparand) {
         if (value == null || comparand == null) {
             return null;
         } else {
-            return toClassWithRealEquals(value).equals(toClassWithRealEquals(comparand));
+            if (value instanceof Message) {
```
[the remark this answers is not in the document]
MMcM: Having two protobuf messages be equal that happen to be shaped the same but come from different descriptors sounds suspicious in general. At the least, it will probably hide bugs in metadata maintenance. Is there no way to restrict it to whatever special case this is arising from?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
pengpeng-lu: I'll remove it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
pengpeng-lu: in `InOpValue`, check that messages are not deep nested, or have repeated fields. and here fail if see that.

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -174,6 +174,34 @@ public static Descriptors.FieldDescriptor findFieldDescriptorOnMessageByOrdinal(
...
     }
 
+    public static boolean compareMessageEquals(@Nonnull Object o1, @Nonnull Object o2) {
+        if (!(o1 instanceof Message) || !(o2 instanceof Message)) {
+            return false;
+        }
+        MessageOrBuilder m1 = (MessageOrBuilder) o1;
+        MessageOrBuilder m2 = (MessageOrBuilder) o2;
+
+        if (m1.getDescriptorForType().getFields().size() != m2.getDescriptorForType().getFields().size()) {
```
MMcM, line 184: If `m1.getDescriptorForType()` and `m2.getDescriptorForType()` are the same, can we just return `equals`?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
normen662: That is true.

MMcM (approved):

normen662 (dismissed):

MMcM (approved):

PR #2788, opened 2024-06-28
Resolves #2787: Index scrubber could enforce early transaction commit

### LINKED ISSUE #2787: Index scrubber could enforce early transaction commit
The indexing scrubber currently doesn't allow the user to specify that it should enforce early committing a transaction early. The index builder has a configuration option`.setTransactionTimeLimitMillis`, which lets the user specify a number of milliseconds after which it will commit a transaction. This is to avoid hitting the 5 second transaction time limit, and it defaults to 4 seconds. The index scrubber should be able to do the same thing.

alecgrieser:
The main substantive change here is that the `OnlineIndexScrubber.Builder` now allows the user to set a time limit and a transaction time limit. This is then honored by logic in `IndexinbBase`: https://github.com/FoundationDB/fdb-record-layer/blob/695c4053527faf59bcb604abd3fd7bf57fddb11f/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java#L935-L944

This should result in operations committing early rather than hitting `transaction_too_old` exceptions.

This also refactors the way that configuration is specified for the online indexer and online index scrubber so that they share a base-builder implementation. That base builder contains the de-duplicated code. In theory, we could have done this by updating the `OnlineIndexScrubber.Builder` with two new methods setting the time limits, but this seemed like a good change to make to avoid having as much config drift in the future.

This resolves #2787.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -905,18 +366,16 @@ public Builder setScrubbingPolicy(@Nonnull final ScrubbingPolicy.Builder scrubbi
          */
         public OnlineIndexScrubber build() {
             validate();
-            OnlineIndexer.Config conf = new OnlineIndexer.Config(limit, limit, maxRetries, recordsPerSecond,
-                    progressLogIntervalMillis, increaseLimitAfter, maxWriteLimitBytes,
-                    OnlineIndexer.Config.UNLIMITED_TIME, OnlineIndexer.Config.UNLIMITED_TIME);
```
alecgrieser, line 910: Note that in the old code, it always set `UNLIMITED_TIME` for both the overall time limit and the per-transaction time limit. The new default is that the build gets `UNLIMITED_TIME` as before but each transaction only gets 4 seconds, just like with index builds.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1013,6 +1009,15 @@ protected Builder() {
 
             }
 
+            /**
```
jjezra, line 1012: Given the shared external builder, shouldn't we move the shared config class to an external file?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -273,90 +267,24 @@ public ScrubbingPolicy build() {
...
 
-        protected int limit = 2000;
-        protected int maxWriteLimitBytes = OnlineIndexer.DEFAULT_WRITE_LIMIT_BYTES;
-        protected int maxRetries = OnlineIndexer.DEFAULT_MAX_RETRIES;
-        protected int recordsPerSecond = OnlineIndexer.DEFAULT_RECORDS_PER_SECOND;
-        private long progressLogIntervalMillis = OnlineIndexer.DEFAULT_PROGRESS_LOG_INTERVAL;
-        // Maybe the performance impact of this is low enough to be always enabled?
-        private boolean trackProgress = true;
-        private int increaseLimitAfter = OnlineIndexer.DO_NOT_RE_INCREASE_LIMIT;
         private long leaseLengthMillis = OnlineIndexer.DEFAULT_LEASE_LENGTH_MILLIS;
```
jjezra, line 280: The `leaseLengthMillis` seems to be belonged in the config instead of `IndexingCommon` (and be handled in the common builder). 
(It is used only once in the code, I think)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexBaseBuilder.java
```diff
@@ -0,0 +1,724 @@
...
+
+/**
+ * Base class that the {@link OnlineIndexer.Builder} and the {@link OnlineIndexScrubber.Builder} can both inherit
+ * from. Building and scrubbing indexes contains a fair amount of common configuration. This consolidates the
+ * common configuration. Note that each method returns the type specified as the concrete type of the resulting
+ * builder, so each implementation should be able to use this class fluently without casting.
+ *
+ * @param <B> concrete type of this builder
+ */
+public abstract class OnlineIndexBaseBuilder<B extends OnlineIndexBaseBuilder<B>> {
```
jjezra, line 46: Can the class be renamed to ` OnlineIndexingBaseBuilder`? 
(I am trying to keep a clear distinction between "index" and "indexing" by avoiding the term "to index" (not always successfully)..)

jjezra (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexBaseBuilder.java
alecgrieser: That actually was my initial name. I can change if it you want, but the idea was that this is the basis for both the `OnlineIndexer.Builder` and the `OnlineIndexScrubber.Builder` classes, so I wanted to encapsulate this works for both "online index" operations. I could go with `OnlineIndexOperationBaseBuilder`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: We could. That would be an API breaking change, in that the fully qualified information for this class would change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
alecgrieser: Maybe? The problem you run into is trying to in structuring a change like this is that the lease length millis is ignored for the online indexer _only_ if it sets the "used synchonrized session parameter", whereas the scrubber always uses a syncrhonized session. The fact that this differs between the two implementations makes it a little hard to merge those configurations together. If we wanted to do this, I feel like we should also merge their behavior a bit more.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: I've moved it out into its own class with 34c1f70. There would need to be a reaction for any downstream project that was setting these values, though not an overly complicated one

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: I agree. If a non-synchronized indexing session can be useful in some cases, so should be a non-synchronized index scrubbing.

jjezra (approved):

PR #2789, opened 2024-06-28
improvements to semantic equals

normen662:
This PR adds/changes the following:
- new `BooleanWithConstraint` to capture a boolean data type that also records a `QueryPlanConstraint` if `true`. Such a boolean is either `false`, or `true` under the assumption that the constraint holds. The constraint may be `QueryPlanConstraint.tautology()` if it is not actually constrained.
- new `ValueEquivalence` which defines e.g. aliases to be considered equal or value trees to be considered equal. This functions as a generalization of `AliasMap`s.
- new `UsesValueEquivalence` interface which can be implemented (currently implemented in `Comparisons.Comparison`, `Value`, `RangeConstraints`, and `QueryPredicate`) to indicate that the class provides `BooleanWIthConstraint semanticEquals(Object, ValueEquivalence)`
- new `semanticEquals(Object, ValueEquivalence)` method call tree (including `equalsWithoutChildren`, `equalsForChildren`) which work as before but instead of returning a `boolean` now return a `BooleanWithConstraint`

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UsesValueEquivalence.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UsesValueEquivalence.java
@@ -1,0 +1,54 @@
+/*
+ * AliasMap.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2020 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import com.apple.foundationdb.record.query.plan.QueryPlanConstraint;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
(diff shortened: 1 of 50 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FirstOrDefaultValue.java
```diff
@@ -110,17 +110,6 @@ public <M extends Message> Object eval(@Nonnull final FDBRecordStoreBase<M> stor
         return childrenObjects.get(0);
     }
 
-    @Override
```
normen662, line 113: This was a bug where this method `equalsWithoutChildren()` actually looked at its children thus doubling the effort.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UsesValueEquivalence.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * AliasMap.java
```
MMcM, line 2:
```suggestion
 * UsesValueEquivalence.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueEquivalence.java
```diff
@@ -0,0 +1,315 @@
+/*
+ * AliasMap.java
```
MMcM, line 2:
```suggestion
 * ValueEquivalence.java
```

MMcM (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UsesValueEquivalence.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ValueEquivalence.java
normen662: Done.

MMcM (approved):

PR #2796, opened 2024-07-02
Add more complete direction control to LogicalSortExpression.

MMcM:
Extends #2748 and #2771.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
```diff
@@ -129,6 +129,11 @@ public int hashCode() {
         return Objects.hash(getOrderingParts(), getDistinctness());
     }
 
+    @Override
+    public String toString() {
```
normen662, line 133: Thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
```diff
@@ -48,19 +51,31 @@
...
     public LogicalSortExpression(@Nonnull List<Value> sortValues,
                                  final boolean reverse,
                                  @Nonnull final Quantifier inner) {
-        this.sortValues = ImmutableList.copyOf(sortValues);
-        this.reverse = reverse;
-        this.inner = inner;
+        this(buildOrdering(sortValues, reverse), inner);
+    }
+
+    private static RequestedOrdering buildOrdering(@Nonnull List<Value> sortValues, boolean reverse) {
```
normen662, line 70: `@Nonnull` missing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
```diff
@@ -109,6 +119,22 @@ public Value getResultValue() {
         return inner.getFlowedObjectValue();
     }
 
+    @Override
+    public boolean semanticEquals(@Nullable final Object other, @Nonnull final AliasMap aliasMap) {
```
normen662, line 123: So `semanticEquals()` in `RelationalExpression` deals with the entire DAG of expressions. You probably just want to override `equalsWithoutChildren()`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
```diff
@@ -122,8 +148,7 @@ public boolean equalsWithoutChildren(@Nonnull RelationalExpression otherExpressi
         }
 
         final LogicalSortExpression other = (LogicalSortExpression) otherExpression;
-
-        return reverse == other.reverse && sortValues.equals(other.sortValues);
+        return ordering.equals(other.ordering);
```
normen662, line 151: I think with this here you can just delete the `semanticEquals()` override.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
```diff
@@ -109,6 +119,22 @@ public Value getResultValue() {
...
+        if (!RelationalExpressionWithChildren.super.semanticEquals(other, aliasMap)) {
+            return false;
+        }
+        LogicalSortExpression otherSort = (LogicalSortExpression)other;
+        // TODO: need semanticEquals on the Values?
+        return this.ordering.equals(otherSort.ordering);
+    }
+
+    @Override
+    public int semanticHashCode() {
```
normen662, line 133: You probably don't want to override this one, but `hashCodeWithoutChildren()`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
```diff
@@ -139,13 +164,13 @@ public int hashCode() {
 
     @Override
     public int hashCodeWithoutChildren() {
-        return Objects.hash(getSortValues(), isReverse());
+        return ordering.hashCode();
```
normen662, line 167: I think with this here you can just delete the `semanticHashCode()` override.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -457,10 +457,9 @@ private static Optional<ScanDirection> satisfiesRequestedOrdering(@Nonnull final
                     if (requestedSortOrder != OrderingPart.RequestedSortOrder.ANY) {
                         final var matchedSortOrder = orderingPart.getSortOrder();
 
-                        if ((matchedSortOrder == OrderingPart.MatchedSortOrder.ASCENDING &&
-                                     (requestedSortOrder == OrderingPart.RequestedSortOrder.ASCENDING)) ||
-                                (matchedSortOrder == OrderingPart.MatchedSortOrder.DESCENDING &&
-                                         (requestedSortOrder == OrderingPart.RequestedSortOrder.DESCENDING))) {
+                        // TODO: Will matchedSortOrder have counterflow nulls?
```
normen662, line 460: Probably yes! So this looks like that if we have requested `DESCENDING` but with the incompatible `NULLS` we cannot satisfy the requested ordering in a way and return `Optional.empty()`. I think that is fine for now if that's the intention.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PushRequestedOrderingThroughSortRule.java
```diff
@@ -60,27 +61,28 @@ public PushRequestedOrderingThroughSortRule() {
 
     @Override
     public void onMatch(@Nonnull final CascadesRuleCall call) {
-        final var bindings = call.getBindings();
+        final PlannerBindings bindings = call.getBindings();
```
normen662, line 64: `var` vs explicit type. Did you find the `var`s to be hard to read here? Genuinely curious about when we should adopt which style.

normen662 (changes requested):

normen662:
Can you check that the right compatibilities are enforced (as well as the wrong ones don't match) in `Ordering.combineBindingsForIntersection()` and `Ordering.combineBindingsForUnion()`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
MMcM: Right. There isn't a case where the match works _except_ for null ordering; that's a mismatch.

But I wasn't sure the match sort order wouldn't be called on someplace to give additional information about its derivation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PushRequestedOrderingThroughSortRule.java
MMcM: I can't bring myself to put generic bounds on the RHS ctor call, so there need to be explicit types for those at least. Which leads to mixed style, which I guess we have to agree is mostly okay.

When refactoring and changing the types of things that interact with generics, like `collection.contains`, I found seeing the explicit types helpful, which is why I changed more than the minimum number here. Although the IDE mostly isn't confused, probably.

In brand new code, I think I only _really_ prefer `var` as a standin for type names that are (1) obvious (I know that's subjective) and (2) very long (ditto).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
MMcM: Added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
MMcM: Removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
MMcM: Removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalSortExpression.java
MMcM: Done

MMcM:
> Can you check that the right compatibilities are enforced (as well as the wrong ones don't match) in `Ordering.combineBindingsForIntersection()` and `Ordering.combineBindingsForUnion()`?

I believe that the new counterflow nulls enum members are still `isDirectional`, so that
```
        if (leftSortOrder.isDirectional() && rightSortOrder.isDirectional()) {
            if (leftSortOrder != rightSortOrder) {
```
takes the conservative approach. Also the directional + `FIXED` cases, where it takes the non-fixed side, seem still correct.

normen662 (approved):

PR #2797, opened 2024-07-02
provide switch to enable or disable sanity checks in the plan debugger

normen662:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
```diff
@@ -269,6 +270,16 @@ private <T> Optional<T> getSilently(@Nonnull final String actionName, @Nonnull f
         }
     }
 
+    @Nonnull
+    public static DebuggerWithSymbolTables sane() {
+        return new DebuggerWithSymbolTables(true);
+    }
+
+    @Nonnull
+    public static DebuggerWithSymbolTables withSanityChecks() {
```
alecgrieser, line 279: These names are a little confusing to me. I take it the idea is that the `sane()` one assumes that all sanity checks would pass, but it may be a little clearer if these were just `withSanityChecks` and `withoutSanityChecks`.

alecgrieser (approved):
LGTM. I guess I found one of the names confusing, but I'm okay with it either way

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
normen662: Done.

alecgrieser (approved):

PR #2802, opened 2024-07-03
Resolves #2801: Covering optimization with invertible index fields

### LINKED ISSUE #2801: Covering optimization with invertible index fields
An `InvertibleFunctionKeyExpression` should be in `AvailableFields` in its original (inverted) form and inversion should be supported by `IndexKeyValueToPartialRecordTest`.

MMcM:
This is complete for the heuristic planner. For Cascades, it adds the necessary partial record support, but more work in `tryFetchCoveringIndexScan` is needed, depending on how precisely function index keys are represented in `Value`s.


normen662 (approved):

alecgrieser (approved):

alecgrieser:
I guess there's a question as to whether we should enhance `FieldData` to allow for more general computation. I think given that we want to replace it with Cascades primitives anyway (which should already be able to handle that sort of them), having the extra field on the copier is fine for the old planner

PR #2805, opened 2024-07-04
Resolves #2804: implement sieve to quickly eliminate intersection combinations

### LINKED ISSUE #2804: improve performance for index anding in `AbstractDataAccessRule`
I am investigating a case where the data access rule needs to enumerate candidates for index anding (intersection of index scans) for dozens of indexes. In my case the number of indexes is 12 and the number of combinations that need to be considered for intersections is around 4k. Since we compute the (among other things) the common ordering repeatedly we end up spending more than a second on this one rule.

Observations:
1. A necessary condition for the intersection to be plan-able is for the common ordering to exist. In fact, this condition is the most filtering condition when planning the intersection of index scans.
2. The intersection merge of n orderings is commutative and associative.
3. If there is a set of two orderings among the n orderings for a combination of single access orderings that does not have a common intersection ordering, the entire intersection ordering also does not exist and the intersection of these n single accesses cannot be planned.
4. If all combinations of size `k <= n - 1` do not have a common intersection ordering, then there cannot be a combination of size `k + 1`. Reasoning: Under the assumption that there is such a common intersection ordering of size `k + 1`, there must be a common ordering of size k that then can be intersection merged with the `(k + 1)`th ordering. That contradicts the assumption.

To address this issue:
1. define a data structure that records all two-element sets of orderings of single accesses that do not have a common intersection ordering,
2. define an algorithm that allows us to quickly establish whether a set of n orderings contains a two-element set of orderings that is known to not have a common intersection ordering
3. early out if we were unable to find combinations of size k that have a common intersection ordering

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -227,31 +232,187 @@ protected Set<? extends RelationalExpression> dataAccessForMatchPartition(@Nonnu
...
+                    final var membership = new BitSet(bestMaximumCoverageMatches.size());
+                    for (final var singleMatchedAccessVectored : kPartition) {
+                        membership.set(singleMatchedAccessVectored.getPosition());
+                    }
+
+                    for (int position = membership.nextClearBit(0);
+                             position >= 0 && position < bestMaximumCoverageMatches.size();
+                             position = membership.nextClearBit(position + 1)) {
+                        clearRowAndColumnAtPosition(checkBitMatrix, position);
+                    }
```
hatyo, line 329: I believe we can make this slightly more efficient by getting rid of `membership` bitset:

```java
                     // setAll(checkBitMatrix); // all bits to 1, this is not needed
                     // final var membership = new BitSet(bestMaximumCoverageMatches.size());
                     for (final var singleMatchedAccessVectored : kPartition) {
                         setRowAndColumnAtPosition(checkBitMatrix, position); // method that sets the row and column according to the position.
                     }
```

This works the inverse way starting off with `checkBitMatrix` set to zero, then setting respective `kPartition` positions to one. But this has the same worst-case cost as your approach, because setting 0-bits for `K:[3,N]` is _almost_ equal to setting 1-bits for `K:[3,N]`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -227,31 +232,187 @@ protected Set<? extends RelationalExpression> dataAccessForMatchPartition(@Nonnu
...
+                    hasCommonOrderingForK = true;
+
+                    resultSet.addAll(intersectionResult.getExpressions());
+                }
+
+                if (!hasCommonOrderingForK) {
+                    // if we weren't able to find combinations of size k that created intersections, there won't be any
+                    // of size k + 1
+                    break;
+                }
```
hatyo, line 355: This is a nice optimisation.
I wonder if we can integrate the "forbidden combinations" concept in the chooseK iterator logic itself, I would argue it would make it even more efficient (and perhaps elegant) considering the following interface:

```java
chooseK(N,K, sieveMatrix)
```

Then, there will be a pruning logic, similar to the one we have for calculating partial orders, that avoids generating combinations (of size K, K+1, K+2,.... N) that contain elements whose coordinates in the sieve matrix is 1.

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: So effectively we push the `for` loop into the `chooseK()`?

alecgrieser: Yeah, I'm not sure how you would implement such a function without effectively replicating the logic in this method, if I'm understanding the suggestion

hatyo (approved):
LGTM

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
hatyo: After some discussion, this is not possible because we need to unset the bits of the non-participating indexes. Thanks @normen662 for the clarification.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
hatyo: > So effectively we push the for loop into the chooseK()?

In a sense, yes.

> Yeah, I'm not sure how you would implement such a function without effectively replicating the logic in this method, if I'm understanding the suggestion

The logic will mostly stay the same, but I think we can leverage the pruning of K-sets to prune K+1 in a more efficient, dynamic programming type of, way, instead of recalculating it over and over again as we do here.

We can look at this later though.

PR #2807, opened 2024-07-05
Fixes #2806: add Support for nested DynamicMessage in ConstantObjectValue

### LINKED ISSUE #2806: add Support for nested `DynamicMessage` in `ConstantObjectValue`
The `ConstantObjectValue` has a very naive support to deal with `DynamicMessage` wherein the object is not checked if there is a need for promotions. However, this fails if the object is an array of `DynamicMessage`. 

We need a to have a way to support arbitrary nesting of `DynamicMessage` but not generally support promotions in them. For instance, we can check if the shape of `DynamicMessage` is _identical_ and fail if not without even trying to calculate promotions.

g31pranjal:
Fixes #2806

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
@@ -422,7 +422,7 @@ public class PromoteValue extends AbstractValue implements ValueWithChild, Value
         }
         if (inType.isRecord() && promoteToType.isRecord()) {
             final List<Type> inTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
-            final List<Type> promoteToTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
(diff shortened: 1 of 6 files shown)


normen662 (changes requested):
Can you add a some tests please?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
```diff
@@ -422,7 +422,7 @@ public static boolean isPromotionNeeded(@Nonnull final Type inType, @Nonnull fin
         }
         if (inType.isRecord() && promoteToType.isRecord()) {
             final List<Type> inTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
-            final List<Type> promoteToTypeElements = Objects.requireNonNull(((Type.Record) inType).getElementTypes());
+            final List<Type> promoteToTypeElements = Objects.requireNonNull(((Type.Record) promoteToType).getElementTypes());
```
normen662, line 425: Yikes!

normen662 (approved):
Thank you for taking the time and actually writing some meaningful tests.

PR #2808, opened 2024-07-06
Implement mixed sort ordering index scans for Cascades.

MMcM:
This is done without introducing new Values to represent the order inverting key function. Instead, the index's original key expression is consulted at the right time to refine the field value.


MMcM:
Closed in favor of #2858.

PR #2811, opened 2024-07-08
add rule to match complex values against a qov

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/PullUpCompensation.java
```diff
@@ -0,0 +1,43 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
+import com.apple.foundationdb.record.query.plan.cascades.values.Value;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Functional interface to perform compensation for partially matched {@link Value}-trees.
+ */
+@FunctionalInterface
+public interface PullUpCompensation {
```
alecgrieser, line 32: I like this! I think it's a lot more readable now that the lambda type has a name

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationTest.java
```diff
@@ -117,6 +120,60 @@ void testPullUpFieldValueThroughStreamingAggregation() {
...
+        final var new_x_b = FieldValue.ofFieldNames(upperCurrentValue, ImmutableList.of("x", "xb"));
+
+        final var expectedMap = ImmutableMap.of(
+                record_type, new RecordTypeValue(QuantifiedObjectValue.of(UPPER_ALIAS, new Type.AnyRecord(true))),
+                _a_ab__plus__x_xb, (Value)new ArithmeticValue.AddFn().encapsulate(ImmutableList.of(new_a_b, new_x_b)));
+        Assertions.assertEquals(expectedMap, resultsMap);
+    }
+
+    @Test
+    void testPullUpValue2() {
```
alecgrieser, line 151: Was one of these tests supposed to look at `recordType(FieldValue(blah))`? I'm pretty sure that that kind of transformation works, but I don't see a test for it.

alecgrieser (approved):

PR #2815, opened 2024-07-10
Combine #2674 and #2812

alecgrieser:
This combines the two PRs, #2674 and #2812, and ensures that #2812 accomplishes its goal, namely allowing us to match a literal value in the index with a constant object value in the query. It should only match if the underlying constant object value's value actually matches the value in the index.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/LongArithmethicFunctionKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/LongArithmethicFunctionKeyExpression.java
@@ -20,6 +20,7 @@
 
 package com.apple.foundationdb.record.metadata.expressions;
 
+import com.apple.foundationdb.record.FunctionNames;
 import com.apple.foundationdb.record.ObjectPlanHash;
 import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
@@ -51,6 +52,8 @@ import java.util.function.UnaryOperator;
 public class LongArithmethicFunctionKeyExpression extends FunctionKeyExpression implements QueryableKeyExpression {
     private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Long-Arithmetic-Key-Expression-Function");
 
+    @Nonnull
+    private final String valueFunctionName;
     private final int minArguments;
     private final int maxArguments;
 
@@ -59,8 +62,15 @@ public class LongArithmethicFunctionKeyExpression extends FunctionKeyExpression 
     @Nullable
     private final BinaryOperator<Long> binaryOperator;
 
-    private LongArithmethicFunctionKeyExpression(@Nonnull String name, @Nonnull KeyExpression arguments, int minArguments, int maxArguments, @Nullable UnaryOperator<Long> unaryOperator, @Nullable BinaryOperator<Long> binaryOperator) {
+    private LongArithmethicFunctionKeyExpression(@Nonnull String name,
+                                                 @Nonnull KeyExpression arguments,
+                                                 @Nonnull String valueFunctionName,
+                                                 int minArguments,
+                                                 int maxArguments,
+                                                 @Nullable UnaryOperator<Long> unaryOperator,
+                                                 @Nullable BinaryOperator<Long> binaryOperator) {
         super(name, arguments);
+        this.valueFunctionName = valueFunctionName;
         this.minArguments = minArguments;
         this.maxArguments = maxArguments;
         this.unaryOperator = unaryOperator;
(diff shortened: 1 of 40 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/LongArithmethicFunctionKeyExpression.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @Nonnull
+    @Override
+    public <S extends KeyExpressionVisitor.State, R> R expand(@Nonnull final KeyExpressionVisitor<S, R> visitor) {
+        return visitor.visitExpression(this);
+    }
+
+    @Nonnull
+    @Override
+    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull final Type baseType, @Nonnull final List<String> fieldNamePrefix) {
```
normen662, line 132: no final promotion?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/LongArithmethicFunctionKeyExpression.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    /**
+     * Factory for constructing built-in arithmetic functions that operate on {@code long}s.
+     */
+    @AutoService(FunctionKeyExpression.Factory.class)
+    public static class LongArithmethicFunctionKeyExpressionFactory implements FunctionKeyExpression.Factory {
+        @Nonnull
+        private static final List<FunctionKeyExpression.Builder> BUILDERS = ImmutableList.<FunctionKeyExpression.Builder>builder()
+                .add(Builder.binaryFunction(FunctionNames.ADD, (l, r) -> l + r))
```
normen662, line 263: So no `mathExact()` here? I mean it's fine, just want to make sure it is what you intended.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/RecordTypeKeyExpression.java
```diff
@@ -118,7 +118,7 @@ public RecordMetaDataProto.KeyExpression toKeyExpression() {
 
     @Nonnull
     @Override
-    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull final List<String> fieldNamePrefix) {
+    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull Type baseType, @Nonnull final List<String> fieldNamePrefix) {
```
normen662, line 121:
```suggestion
    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull final Type baseType, @Nonnull final List<String> fieldNamePrefix) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/VersionKeyExpression.java
```diff
@@ -107,7 +108,7 @@ public RecordMetaDataProto.KeyExpression toKeyExpression() {
 
     @Nonnull
     @Override
-    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull final List<String> fieldNamePrefix) {
+    public Value toValue(@Nonnull final CorrelationIdentifier baseAlias, @Nonnull final Type baseType, @Nonnull final List<String> fieldNamePrefix) {
```
normen662, line 111: if in the caller you always have the `baseQuantifier` you might as well hand in that one as it holds both the type and the alias. I think I started with just the alias but then later on changed quite a few places to take a quantifier but apparently not all.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryKeyExpressionWithOneOfComparison.java
```diff
@@ -90,7 +90,7 @@ public Comparisons.Comparison getComparison() {
     public GraphExpansion expand(@Nonnull final Quantifier.ForEach baseQuantifier,
                                  @Nonnull final Supplier<Quantifier.ForEach> outerQuantifierSupplier,
                                  @Nonnull final List<String> fieldNamePrefix) {
-        return GraphExpansion.ofPredicate(keyExpression.toValue(baseQuantifier.getAlias(), fieldNamePrefix).withComparison(comparison));
+        return GraphExpansion.ofPredicate(keyExpression.toValue(baseQuantifier.getAlias(), baseQuantifier.getFlowedObjectType(), fieldNamePrefix).withComparison(comparison));
```
normen662, line 93: connected to my earlier comment: just hand in the quantifier

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/VersionKeyExpression.java
alecgrieser: I took a look at this, and there is one case where we actually don't have the base quantifier, namely `ScalarTranslationVisitor`'s visit method for `KeyExpressionWithValue`:

https://github.com/FoundationDB/fdb-record-layer/blob/c90ef5e336e384d94b35e351d72cf7b8033bad8c/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ScalarTranslationVisitor.java#L138-L141

The `ScalarTranslationVisitor` only has type and alias information. I did briefly take a look at if I could modify it to use a quantifier, but it ended up being a little bit more tricky than I'd hoped. It usually uses the `Quantifier.current()` alias as its basis rather than using a real quantifier.

I added a default variant to this `toValue` method that takes a quantifier while delegating to the other method, and I updated the call sites that I could to use it.

normen662: okay bummer

normen662 (dismissed):

normen662 (approved):

PR #2817, opened 2024-07-11
performance improvements for in-union

normen662:


alecgrieser (approved):
This seems fine to me, given that the tests pass. It's not entirely obvious to me what is changing, or why it would make a difference

PR #2822, opened 2024-07-12
Fixes #2813: Fail fast when Batch GRV Rate Limit Exceeded

### LINKED ISSUE #2813: Lucene tests sometimes stall until they timeout
Specifically some people have noticed that `LuceneIndexTest.fullDeleteHelper` times out after 5 minutes rather than completing, but it's normal runtime is less than 500ms.

Similarly the newly added  `LuceneIndexMaintenanceTest.concurrentStoreTest` can experience a similar issue.

ScottDugas:
Loading the metadata version key does not seem to throw Batch GRV rate limit exceeded, while other operations on the same transaciton do. Instead it stalls forever, and the tests end up timing out at the junit level.
This change introduces a new combine that will fail when the loading the record store state fails, instead of waiting forever for the metadata version to be loaded.

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
@@ -946,6 +946,38 @@ public class MoreAsyncUtil {
     }
 
     /**
+     * Combine the results of two futures, but fail fast if either future fails.
+     * <p>
+     *     This has the same behavior as {@link CompletableFuture#thenCombine}, except, if either future fails, it won't
+     *     wait for the other one to complete before completing the result with the failure.
+     * </p>
+     * @param future1 one future
+     * @param future2 another future
+     * @param combiner a function to combine the results of both {@code future1} and {@code future2} into a single result.
+     *
+     * @param <T> the result type for {@code future1}
+     * @param <U> the result type for {@code future2}
+     * @param <R> the result type for the returned future
+     *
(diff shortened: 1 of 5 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
```diff
@@ -945,6 +945,38 @@ private static RuntimeException getRuntimeException(@Nonnull Throwable exception
...
+     * @param <R> the result type for the returned future
+     *
+     * @return a future that fails with one of the exceptions from {@code future1} or {@code future2} if either of those
+     * failed, or the result of applying {@code combiner} to their results if both succeeded.
+     */
+    public static <T, U, R> CompletableFuture<R> combineAndFailFast(CompletableFuture<T> future1,
+                                                                    CompletableFuture<U> future2,
+                                                                    BiFunction<T, U, R> combiner) {
+        return CompletableFuture.anyOf(future1, future2).thenCompose(vignore -> {
+            // We know at least one of them completed
```
alecgrieser, line 969: We actually know that one of them completed _successfully_, because `thenCompose` only runs on successful completion of a future. I believe that the rest of the logic in the function should work, in that if one future completes successfully in more-or-less the same time as the other completes exceptionally, then we could go down one of the two cases below.

The strongest case for this is probably that if one of the futures completes unsuccessfully while the other hangs, then the `.anyOf` will complete with the exception from the one that has completed, so the end result will be an exceptional future.

alecgrieser (approved):

PR #2824, opened 2024-07-15
`Value` subsumption should default to semantic equality

### LINKED ISSUE #2825: `Value` subsumption should default to semantic equality
Currently, only `FieldValue` is performing this delegation, but it is necessary to do it for all other `Value` types as well.

This is necessary in particular for `ConstantObjectValue` for it to be able to match a `LiteralValue` correctly (under the assumption that both are referring to the _same_ literal).

hatyo:
Currently, only `FieldValue` is performing this delegation, but it is necessary to do it for all other `Value` types as well.

This is necessary in particular for `ConstantObjectValue` for it to be able to match a `LiteralValue` correctly (under the assumption that both are referring to the _same_ literal).

This fixes #2825.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
```diff
@@ -35,6 +35,7 @@
 import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.cascades.Formatter;
 import com.apple.foundationdb.record.query.plan.cascades.SemanticException;
+import com.apple.foundationdb.record.query.plan.cascades.ValueEquivalence;
```
normen662, line 38: Seems like an unused import.

normen662 (changes requested):

normen662 (approved):

normen662 (approved):

hatyo:
in favour of https://github.com/FoundationDB/fdb-record-layer/pull/2838

PR #2828, opened 2024-07-17
add feature to record planning events and to check them in a later run

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ObjectPlanHash.java
```diff
@@ -48,6 +48,11 @@ public int queryHash(@Nonnull final QueryHashable.QueryHashKind hashKind) {
         return hashCode;
     }
 
+    @Override
+    public int hashCode() {
```
MMcM, line 52: I suppose it should have `equals` for completeness, too.

MMcM (changes requested):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
```diff
@@ -233,10 +252,68 @@ public void onDone() {
...
+                        "There are more prerecorded events, there are only " + state.getEvents().size() + " actual events.");
+            }
         }
         reset();
     }
 
+    private static void writeEventsDelimitedToFile(final List<PEvent> eventProtos) {
+        try {
+            final var tempFile = File.createTempFile("events-", ".bin",
+                    new File(System.getProperty("java.io.tmpdir")));
```
MMcM, line 273: Is this not what the two-arg version of `File.createTempFile` does?

MMcM (dismissed):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/debug/DebuggerWithSymbolTables.java
normen662: changed that call.

MMcM (approved):

PR #2830, opened 2024-07-17
Prune self-dependencies when calculating transitive closure.

hatyo:


hatyo:
Root cause is fixed in https://github.com/FoundationDB/fdb-record-layer/pull/2838

PR #2262, opened 2023-08-17
Resolve #2259: Deprecate special IndexingByRecords functions

jjezra:

PR #2834, opened 2024-07-18
Resolves #2484: Removes Apache commons usage

### LINKED ISSUE #2484: Remove usages of `org.apache.commons.lang3.tuple.Pair` et al

alecgrieser:
This PR removes the Apache commons dependency from all of the non-test artifacts. It builds upon previous work that mostly removed our usage of the `Pair` class from that library, and then also introduces alternatives for a few methods that we ended up relying on.

There are a few places where the API had to change because we were exposing `Pair` all the way up to the API level. They were:
* _Setting the async to sync timeout_. Now, instead of taking a `Pair<Long, TimeUnit>`, it takes a `Duration`, which is a kind of unit-associated time in much the same way that the pair was trying to represent.
* _Returning either a result value or an error_. A new `Result` class modeled on the `Result` monad present in several functional programming languages. In some ways, it's a kind of specialized `Pair`, but with more documentation as to what its constituents mean.
* _Splitting key spaces into ranges of `Tuple`s_. One method on the online indexer used `Pair<Tuple, Tuple>` instead of `TupleRange`. The first `Tuple` was always inclusive and the end `Tuple` was always exclusive, so this is equivalent to `TupleRange.between(startTuple, endTuple)`, so that's what replaced it.
* _Key space resolution could return un-resolved elements from the original `Tuple`_. This has been upgraded to its own named struct.
* _Time window leaderboard returned both the rank and the associated entry_. This has also gotten its own named struct.

These should be relatively straightforward things for users to respond to. The upshot for adopters is that they are free now to drop the Apache commons dependency (or safely pick their own version).


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PatternForLikeValue.java
```diff
@@ -83,14 +100,18 @@ public <M extends Message> String eval(@Nonnull final FDBRecordStoreBase<M> stor
...
+            replaceMap = REPLACE_MAP;
+        } else {
             SemanticException.check(escapeChar.length() == 1, SemanticException.ErrorCode.ESCAPE_CHAR_OF_LIKE_OPERATOR_IS_NOT_SINGLE_CHAR);
-            search = ObjectArrays.concat(new String[] {escapeChar + "_", escapeChar + "%"}, SEARCH, String.class);
-            replace = ObjectArrays.concat(new String[] {"_", "%"}, REPLACE, String.class);
+            replaceMap = ImmutableMap.<String, String>builderWithExpectedSize(REPLACE_MAP.size() + 2)
+                    .put(escapeChar + "_", "_")
+                    .put(escapeChar + "%", "%")
+                    .putAll(REPLACE_MAP)
+                    .build();
```
alecgrieser, line 112: This preserves the original iteration order, if we want to use a `replaceEach` variant that takes that into account. As it is, it would only be relevant if the escape character were somehow one of the elements in the static `REPLACE_MAP` (most probably `\\`), which would actually be equivalent to the "choose the longer search string if two at an index" rule used by the new `replaceEach`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -726,11 +725,11 @@ protected void buildIndexSingleTarget() {
...
      * @param minSplit not split if it cannot be split into at least <code>minSplit</code> ranges
      * @param maxSplit the maximum number of splits generated
-     * @return a list of split primary key ranges (the low endpoint is inclusive and the high endpoint is exclusive)
+     * @return a list of split primary key ranges
      */
     @API(API.Status.EXPERIMENTAL)
     @Nonnull
-    public List<Pair<Tuple, Tuple>> splitIndexBuildRange(int minSplit, int maxSplit) {
+    public List<TupleRange> splitIndexBuildRange(int minSplit, int maxSplit) {
         return getIndexerByRecordsOrThrow().splitIndexBuildRange(minSplit, maxSplit);
```
jjezra, line 733: Since the API is changing in this PR, it might be a good idea to preemptively mark this function as deprecated. 
Letting users adjust their code for this change without knowing that this function is a candidate for removal (https://github.com/FoundationDB/fdb-record-layer/pull/2262) seems a little unfair.

jjezra (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
alecgrieser: I've gone ahead and marked these methods as deprecated. I'll have to admit that #2262 sort of slipped my mind. It may be that we should just remove these methods now, if we're happy with mutual indexing, though I'm actually a bit out of touch with where we were with that project and with #2262 in particular

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/combinatorics/ChooseK.java
```diff
@@ -79,59 +116,57 @@ protected List<T> computeNext() {
...
+                    // offset to move on to the next element.
                     //
-                    MutablePair<Integer, Integer> currentPair = state.get(currentLevel);
-                    if (currentPair.left == -1) {
-                        currentPair.left = 1;
-                        currentPair.right = elements.size() - currentOffset + 1;
-                        lastOffset = 0;
+                    LevelState currentState = state.get(currentLevel);
+                    if (currentState.isBound()) {
+                        unbind(currentLevel); // Mark this level as unbound and clear out greater levels
```
jjezra, line 148: Nice!

jjezra (approved):

alecgrieser:
One comment from @normen662 (offline) was that we wanted to make sure that `ChooseK` didn't change its order. For that reason, I've updated the `ChooseKTest`s to assert on the order of returned combinations (by making the test assert on the contents of a `List` rather than a `Set`). I've validated that these tests passed on the current `main` without the other changes

normen662 (dismissed):

.idea/compiler.xml
```diff
@@ -61,19 +61,19 @@
       <profile name="Gradle Imported" enabled="true">
         <outputRelativeToContentRoot value="true" />
         <processorPath useClasspath="false">
-          <entry name="$PROJECT_DIR$/fdb-extensions/.out/libs/fdb-extensions-3.4-SNAPSHOT.jar" />
+          <entry name="$PROJECT_DIR$/fdb-extensions/.out/libs/fdb-extensions-3.5-SNAPSHOT.jar" />
```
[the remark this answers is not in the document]
alecgrieser: This is intentional. This file is used to bootstrap a developer who freshly checks out the repo, and it should be updated to reflect new dependencies. That being said, I'm not actually sure if this is the right approach for this problem (and we have done this alternate ways in the past)

fdb-extensions/src/main/java/com/apple/foundationdb/util/StringUtils.java
```diff
@@ -0,0 +1,249 @@
+/*
+ * StringUtil.java
```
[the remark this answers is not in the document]
alecgrieser: Fixed

normen662 (approved):

PR #2844, opened 2024-07-24
Add some sleep between tests in LuceneIndexMaintenanceTest

ScottDugas:
The idea here being that the tests cause enough load on the fdb cluster that it can't handle anything else. Adding a 1 second sleep in between may allow it to recover and proceed to the next set of tests in a healthy state.


ohadzeliger (approved):

ohadzeliger:
This is probably OK for quick check if that resolves the issue, but if it does, we should probably fix the test to reduce the load it places on the system...

PR #2845, opened 2024-07-24
Add commas to test durations

ScottDugas:
Not having them makes it a bit harder to read.
I thought about using `Duration.ofMillis` but the `toString` on that is like PT0.234 or PT1.4098 which I, personally find hard to parse visually.

### DIFF
--- a/gradle/testing.gradle
+++ b/gradle/testing.gradle
@@ -149,7 +149,7 @@ tasks.withType(Test) { theTask ->
         println "${Instant.now()} ${getFullDisplayName(descriptor)} STARTED"
     } 
     afterTest { descriptor, result ->
-        def duration = String.format("%,d", result.endTime - result.startTime)
+        def duration = String.format(Locale.ROOT, "%,d", result.endTime - result.startTime)
         println "${Instant.now()} ${getFullDisplayName(descriptor)} ${result.resultType} (${duration}ms)"
         println()
     }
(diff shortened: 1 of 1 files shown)


gradle/testing.gradle
```diff
@@ -149,7 +149,7 @@ tasks.withType(Test) { theTask ->
         println "${Instant.now()} ${getFullDisplayName(descriptor)} STARTED"
     } 
     afterTest { descriptor, result ->
-        def duration = result.endTime - result.startTime
+        def duration = String.format("%,d", result.endTime - result.startTime)
```
alecgrieser, line 152: Should this set a locale, especially given that nightlies run with a random locale?

```suggestion
        def duration = String.format(Locale.ROOT, "%,d", result.endTime - result.startTime)
```

Under the current logic, this uses the system locale to determine whether to use a comma or a period as the thousands separator, as well as for number formatting, which could make it harder for someone reading this to debug an issue. On the other hand, using the user's locale is probably preferred for local testing. If we have a signal for seeing if we're in CI, we could do something like:

```groovy
        def locale = inCI() ? Locale.ROOT : Locale.getDefault()
        def duration = String.format(locale, "%,d", result.endTime - result.startTime)
```

ScottDugas: I think you're right, that using `Locale.ROOT` makes the most sense.
Although, we could definitely add a way to detect if in CI, I'm not sure I like it.

ScottDugas: I also tested locally with the random locales, and this does not pick up the randomly chosen locale.
Specifically, I saw:
- `Running tests with locale: zgh_MA`

And without the `Locale.ROOT` it still resulted in a `,` for the grouping separator.

alecgrieser: Hm, that is interesting.

alecgrieser (approved):

PR #2848, opened 2024-07-26
Prep #2360: Ensure code compiles on JDK 17

alecgrieser:
This makes updates to the code base in order to prepare for updating to JDK 17. Contained within this is:

1. A bunch of Javadoc warnings became errors. For the most part, these were warnings about methods or classes that were entirely missing. I began updating some of those, but ultimately, there's a bit too much, so I added a flag to loosen up the Javadoc requirements.
1. SpotBugs needed to be updated in order to handle class files generated at the newer JDK. Some of the rules changed, and in particular, the newer version is more persnickety about retunring mutable values, and it doesn't like exceptions being thrown in the constructor. Some of those will be harder to address than others, so I've excluded some of the new rules and fixed up a few other things that were caught by spotbugs.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/KeyValueLogMessage.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/KeyValueLogMessage.java
@@ -37,40 +37,50 @@ import java.util.TreeMap;
  */
 @API(API.Status.MAINTAINED)
 public class KeyValueLogMessage {
+    @Nonnull
     private final long timestamp;
+    @Nonnull
     private final String staticMessage;
 
+    @Nonnull
     private final Map<String, String> keyValueMap;
 
     public static String of(@Nonnull final String staticMessage, @Nullable final Object... keysAndValues) {
-        return new KeyValueLogMessage(staticMessage, keysAndValues).toString();
+        return new KeyValueLogMessage(staticMessage, mapFromKeysAndValues(keysAndValues)).toString();
     }
 
     public static KeyValueLogMessage build(@Nonnull final String staticMessage, @Nullable final Object... keysAndValues) {
-        return new KeyValueLogMessage(staticMessage, keysAndValues);
+        return new KeyValueLogMessage(staticMessage, mapFromKeysAndValues(keysAndValues));
     }
 
-    private KeyValueLogMessage(@Nonnull final String staticMessage, @Nullable final Object[] keysAndValues) {
-        this.staticMessage = staticMessage;
-        this.timestamp = System.currentTimeMillis();
-        keyValueMap = new TreeMap<>();
+    private static Map<String, String> mapFromKeysAndValues(@Nullable Object[] keysAndValues) {
+        Map<String, String> keyValueMap = new TreeMap<>();
         if (keysAndValues != null) {
             for (int i = 0; i < (keysAndValues.length / 2) * 2; i += 2) {
-                addKeyValueImpl(keysAndValues[i + 0], keysAndValues[i + 1]);
+                addKeyValueImpl(keyValueMap, keysAndValues[i], keysAndValues[i + 1]);
             }
             if (keysAndValues.length % 2 == 1) {
                 throw new IllegalArgumentException("keys and values don't match");
             }
         }
+        return keyValueMap;
     }
 
-    private void addKeyValueImpl(@Nullable final Object key, @Nullable Object value) {
+    private KeyValueLogMessage(@Nonnull final String staticMessage, @Nonnull Map<String, String> keyValueMap) {
+        this.staticMessage = staticMessage;
+        this.timestamp = System.currentTimeMillis();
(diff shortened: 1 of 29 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/logging/KeyValueLogMessage.java
```diff
@@ -37,40 +37,50 @@
...
-    private void addKeyValueImpl(@Nullable final Object key, @Nullable Object value) {
+    private KeyValueLogMessage(@Nonnull final String staticMessage, @Nonnull Map<String, String> keyValueMap) {
+        this.staticMessage = staticMessage;
+        this.timestamp = System.currentTimeMillis();
+        this.keyValueMap = keyValueMap;
+    }
+
+    private static void addKeyValueImpl(@Nonnull Map<String, String> keyValueMap, @Nullable final Object key, @Nullable Object value) {
         if (key == null) {
             throw new IllegalArgumentException("null key passed to KeyValueLogMessage"); // better to throw an exception rather than silently swallow the key
```
alecgrieser, line 77: This exception has been removed from the constructor and put into a helper method that is used by the static initializer. This ensures that the `finalizer` never sees an incomplete object, which is what SpotBugs was concerned about. There are enough classes with this problem that I just suppressed the error. For some of them, we could consider doing this kind of change (or marking the class as `final`), but the story gets more complicated with others.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
```diff
@@ -220,8 +220,7 @@ private void setFragmentationData(@Nonnull FDBRecordStore store) {
         if (fragmentBoundaries == null || fragmentBoundaries.isEmpty()) {
             fragmentBoundaries = getPrimaryKeyBoundaries(store);
         }
-        // Can't use ThreadLocalRandom.current() without triggering a sonarcloud error
-        final SecureRandom rn = new SecureRandom();
+        final Random rn = ThreadLocalRandom.current();
```
alecgrieser, line 223: SpotBugs is complaining about creating an RNG for a single use here, which is why I'm changing this to use the thread local random. I looked at the original PR that tried this: #1939. The PR no longer says what the error was, though maybe @jjezra remembers. I think using the thread local random should be fine. It's possible that sonarcloud doesn't like that the RNG is being handed to the `getPrimeStep` method (after which there's not a guarantee that the RNG isn't accessed from another thread, at least not without examining the method). If that's the case, we could rework this by having `getPrimeStep` call `ThreadLocalRandom.current()` itself, or by suopressing the sonar error.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -926,7 +926,7 @@ private <T> void addAnonymousCommitHookToMap(@Nonnull Map<String, T> map, @Nonnu
             String name;
             // Yes, a collision is exceedingly unlikely, but...
             do {
-                name = INTERNAL_COMMIT_HOOK_PREFIX + "anon-" + (new Random()).nextInt(Integer.MAX_VALUE);
+                name = INTERNAL_COMMIT_HOOK_PREFIX + "anon-" + ThreadLocalRandom.current().nextInt(Integer.MAX_VALUE);
```
alecgrieser, line 929: This is to fix a new SpotBugs issue. It doesn't like a new random being created each time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
```diff
@@ -59,7 +59,7 @@ public class IndexingCommon {
     @Nonnull private final AtomicLong totalRecordsScanned;
     private final boolean trackProgress;
 
-    @Nonnull public OnlineIndexOperationConfig config; // this item may be modified on the fly
+    @Nonnull OnlineIndexOperationConfig config; // this item may be modified on the fly
```
alecgrieser, line 62: SpotBugs didn't like the member being `public`

alecgrieser:
@jjezra I'm asking for your review particularly for the change to use `ThreadLocalRandom` if you remember what sonarcloud didn't like about it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMutuallyByRecords.java
jjezra: If `sonarCloud` doesn't complain anymore, it should probably be fine. IIRC, that was the only reason for the `SecureRandom` usage.

jjezra (approved):

PR #2855, opened 2024-07-30
Resolve #2854: Fix flaky test - testMutualIndexingWeirdBoundaries

### LINKED ISSUE #2854: Fix flaky test - testMutualIndexingWeirdBoundaries
As it seems, mutual indexing test threads may be verifying uniform index states while another thread had already finished indexing and changing the states to readable.

jjezra:
As it seems, mutual indexing test threads may be verifying uniform index states while another
  thread had already finished indexing and changing the states to readable.


hatyo (approved):
LGTM!

PR #2869, opened 2024-08-07
resolves #2846

### LINKED ISSUE #2846: DimensionsKeyExpression not handled enough by QueryToKeyMatcher
Given an index like
```
covering([[p1, p2, d1, d2, f1, f2] dimensions(2,2), v1] split 6)
```
with dimensions beneath key-with-value, `canDeleteWhere` for an equality comparison on `p1` fails.

It handles such a case at top-level, but not inside like that. It needs to be more like grouping, for the prefix key, which is what is matchable.

The specific error gets to `unexpected` because the key class isn't in `KNOWN_KEY_EXPRESSIONS`.

Also, the `toString` has `dimensions(2,2)2` for some reason.

g31pranjal:
resolves #2846

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/DimensionsKeyExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/DimensionsKeyExpression.java
@@ -55,7 +55,7 @@ public class DimensionsKeyExpression extends BaseKeyExpression implements KeyExp
     private final int prefixSize;
     private final int dimensionsSize;
 
-    public DimensionsKeyExpression(@Nonnull final KeyExpression wholeKey,
+    private DimensionsKeyExpression(@Nonnull final KeyExpression wholeKey,
                                    final int prefixSize,
                                    final int dimensionsSize) {
         this.wholeKey = wholeKey;
@@ -63,7 +63,7 @@ public class DimensionsKeyExpression extends BaseKeyExpression implements KeyExp
         this.dimensionsSize = dimensionsSize;
     }
 
-    public DimensionsKeyExpression(@Nonnull RecordMetaDataProto.Dimensions dimensions) throws DeserializationException {
+    DimensionsKeyExpression(@Nonnull final RecordMetaDataProto.Dimensions dimensions) throws DeserializationException {
         this(KeyExpression.fromProto(dimensions.getWholeKey()), dimensions.getPrefixSize(), dimensions.getDimensionsSize());
     }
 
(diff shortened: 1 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/expressions/DimensionsKeyExpression.java
```diff
@@ -55,15 +55,15 @@ public class DimensionsKeyExpression extends BaseKeyExpression implements KeyExp
...
+    private DimensionsKeyExpression(@Nonnull final KeyExpression wholeKey,
                                    final int prefixSize,
                                    final int dimensionsSize) {
         this.wholeKey = wholeKey;
         this.prefixSize = prefixSize;
         this.dimensionsSize = dimensionsSize;
     }
 
-    public DimensionsKeyExpression(@Nonnull RecordMetaDataProto.Dimensions dimensions) throws DeserializationException {
+    DimensionsKeyExpression(@Nonnull RecordMetaDataProto.Dimensions dimensions) throws DeserializationException {
```
g31pranjal, line 66: final

fdb-record-layer-core/src/test/proto/test_records_multidimensional.proto
```diff
@@ -29,7 +29,10 @@ import "record_metadata_options.proto";
 option (schema).store_record_versions = true;
 
 message MyMultidimensionalRecord {
-  optional string rec_domain = 1;
+  message domain_info {
+    optional string rec_domain = 1;
+  }
```
g31pranjal, line 34: This change would have a test failing without the fix to `deleteWhere`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -2986,9 +2986,9 @@ protected boolean planNestedFieldOrComponentChild(@Nonnull KeyExpression child,
             @Nullable ScoredMatch scoredMatch = maybeSortedMatch.apply(null);
             if (scoredMatch != null) {
                 ComparisonRanges nextComparisonRanges = scoredMatch.getComparisonRanges();
-                if (!comparisons.isEqualities() && nextComparisonRanges.getEqualitiesSize() > 0) {
-                    throw new Query.InvalidExpressionException(
-                            "Two nested fields in the same and clause, combine them into one");
+                if (comparisons.getUncommittedComparisonRanges().stream().anyMatch(range -> range.getRangeType() != ComparisonRange.Type.EQUALITY)
```
normen662, line 2989: maybe make an `isEqualities(form, to)` instead of using this stream expression twice (here and in `ComparisonRanges`)

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
g31pranjal: done

MMcM (dismissed):

normen662 (approved):

PR #2873, opened 2024-08-08
Expose `Tuple`-based `MIN_EVER` and `MAX_EVER` indexes to Cascades

### LINKED ISSUE #2874: Expose `Tuple`-based `MIN_EVER` and `MAX_EVER` indexes to Cascades
At the moment, Cascades is able to deal with `LONG`-based versions of `MIN_EVER` and `MAX_EVER` indexes. There is a variation of these indexes that internally works with FDB `Tuple`. That variation however, is not exposed to Cascades and therefore is not considered when planning a query.

Since these two variations (`LONG`-based and `TUPLE`-based) mainly differ in their storage structure, we could expose their respective match candidates exactly the same way allowing the planner to pick the right one based on the requested input/output of the index aggregate function.

hatyo:
At the moment, Cascades is able to deal with `LONG`-based versions of `MIN_EVER` and `MAX_EVER` indexes. There is a variation of these indexes that internally works with FDB `Tuple`. That variation however, is not exposed to Cascades and therefore is not considered when planning a query.

Since these two variations (`LONG`-based and `TUPLE`-based) mainly differ in their storage structure, we could expose their respective match candidates exactly the same way allowing the planner to pick the right one based on the requested input/output of the index aggregate function.

This fixes #2874 .

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/IndexOnlyAggregateValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/IndexOnlyAggregateValue.java
@@ -31,14 +31,13 @@ import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.metadata.IndexTypes;
 import com.apple.foundationdb.record.planprotos.PIndexOnlyAggregateValue;
 import com.apple.foundationdb.record.planprotos.PIndexOnlyAggregateValue.PPhysicalOperator;
-import com.apple.foundationdb.record.planprotos.PMaxEverLongValue;
-import com.apple.foundationdb.record.planprotos.PMinEverLongValue;
+import com.apple.foundationdb.record.planprotos.PMaxEverValue;
+import com.apple.foundationdb.record.planprotos.PMinEverValue;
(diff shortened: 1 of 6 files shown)


normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/IndexOnlyAggregateValue.java
```diff
@@ -190,162 +189,160 @@ public boolean equals(final Object other) {
...
         }
 
         @Nonnull
         private static AggregateValue encapsulate(@Nonnull final List<? extends Typed> arguments) {
             Verify.verify(arguments.size() == 1);
             final Typed arg0 = arguments.get(0);
-            final Type type0 = arg0.getResultType();
-            SemanticException.check(type0.isNumeric(), SemanticException.ErrorCode.UNKNOWN, "only numeric types allowed in " + IndexTypes.MIN_EVER_LONG + " aggregation operation");
-            return new MinEverLongValue(PhysicalOperator.MIN_EVER_LONG, (Value)arg0);
+            return new MinEverValue(PhysicalOperator.MIN_EVER_LONG, (Value)arg0);
```
alecgrieser, line 222: Should the physical operator also be renamed?

PR #2882, opened 2024-08-15
Fixes #2881: VerifyException in Compensation if unmatched quantifiers is only non-empty thing requiring compensation

### LINKED ISSUE #2881: VerifyException in Compensation if unmatched quantifiers is only non-empty thing requiring compensation
There's a `Verify` in `Compensation.ofChildCompensationAndPredicateMap` here:

https://github.com/FoundationDB/fdb-record-layer/blob/ee43d69b4baf6334a4c337ea60ea4286a60e7462/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java#L338

Note that it validates that either the `predicateCompensationMap` or the `remainingComputationOptional` is set. However, it's called from `SelectExpression.compensate`, and the call is subject to this check:

https://github.com/FoundationDB/fdb-record-layer/blob/ee43d69b4baf6334a4c337ea60ea4286a60e7462/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java#L746-L747

Note that those are similar, except if there are any `unmatchedQuantifiers` (and no other need to compensate), then the check within `SelectExpression.compensate` can pass and lead to us calling `Compensation.ofChildCompensationAndPredicateMap` in a way that triggers the validate expression.

I believe that `Compensation` can already handle this kind of thing, as the unmatched quantifiers map was added way back in #1461. However, we apparently don't have a test case for this. I'm working on trying to get a test-case to use here.

alecgrieser:
This updates the `VerifyException` in `Compensation` so that it does not reject inputs that `SelectExpression.compensate` expects to be able to hand to it.

This fixes #2881.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
@@ -335,7 +335,7 @@ public interface Compensation {
                                                            @Nonnull final Set<Quantifier> unmatchedQuantifiers,
                                                            @Nonnull final Set<CorrelationIdentifier> compensatedAliases,
                                                            @Nonnull final Optional<Value> remainingComputationOptional) {
-        Verify.verify(!predicateCompensationMap.isEmpty() || remainingComputationOptional.isPresent());
+        Verify.verify(!predicateCompensationMap.isEmpty() || !unmatchedQuantifiers.isEmpty() || remainingComputationOptional.isPresent());
         return new ForMatch(isImpossible, childCompensation, predicateCompensationMap, matchedQuantifiers, unmatchedQuantifiers, compensatedAliases, remainingComputationOptional);
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
@@ -709,7 +709,7 @@ public interface Compensation {
                                                                           @Nonnull final Set<? extends Quantifier> unmatchedQuantifiers,
                                                                           @Nonnull final Set<CorrelationIdentifier> compensatedAliases,
                                                                           @Nonnull final Optional<Value> remainingComputationValueOptional) {
-            Verify.verify(!predicateCompensationMap.isEmpty() || remainingComputationValueOptional.isPresent());
+            Verify.verify(!predicateCompensationMap.isEmpty() || !unmatchedQuantifiers.isEmpty() || remainingComputationValueOptional.isPresent());
             return new ForMatch(isImpossible, childCompensation, predicateCompensationMap, matchedQuantifiers, unmatchedQuantifiers, compensatedAliases, remainingComputationValueOptional);
         }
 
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -709,7 +709,7 @@ public WithSelectCompensation derivedWithPredicateCompensationMap(final boolean
                                                                           @Nonnull final Set<? extends Quantifier> unmatchedQuantifiers,
                                                                           @Nonnull final Set<CorrelationIdentifier> compensatedAliases,
                                                                           @Nonnull final Optional<Value> remainingComputationValueOptional) {
-            Verify.verify(!predicateCompensationMap.isEmpty() || remainingComputationValueOptional.isPresent());
```
hatyo, line 712: For this particular case, the creation of the `Compensation` in `SelectExpression#compensate` is relevant for signalling the need of `Exists` to inject itself as a compensating `QueryPredicate` in the containing query fragment as in here: 

https://github.com/FoundationDB/fdb-record-layer/blob/c504a045f318da8bef5c611d9fa62579610c9a4f/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/ExistsPredicate.java#L212-L215

Moreover, a `Compensation` is something that is supposed to filter more rows and/or change the shape of the results, I think creating a compensation without a predicate nor a remaining computation is malformed, for more information, looks at the `apply` logic of `ForMatch`:

https://github.com/FoundationDB/fdb-record-layer/blob/c504a045f318da8bef5c611d9fa62579610c9a4f/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java#L726-L735

we return immediately if we do not find a predicate in the compensation predicate map.

and therefore, if in your case, when we notice an unmatched quantifier only, we should probably create an impossible compensation instead, which, in addition to what I mentioned above, more efficient, considering that it is only used for triggering the injection of `Exists` itself as a compensation.

hatyo: My suggestion is to change this block:

https://github.com/FoundationDB/fdb-record-layer/blob/c504a045f318da8bef5c611d9fa62579610c9a4f/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java#L746-L751

to become something like this:

```java
final var isCompensationNeeded =
        !predicateCompensationMap.isEmpty() || matchInfo.getRemainingComputationValueOptional().isPresent();

if (!isCompensationNeeded) {
    if (!unmatchedQuantifiers.isEmpty()) {
        return Compensation.impossibleCompensation();
    }
    return Compensation.noCompensation();
}
```

hatyo (changes requested):

normen662:
This is fixed in #2888. The test case from this PR is used verbatim in #2888. The problem that is exposed by this problem goes further than just the check. It is possible that a particular compensation is required but we never use the necessary quantifiers during the pull-up logic in `Compensation`. @hatyo and I debugged this today and we are pretty sure to have gotten to the bottom of this one.

Issue #2884, opened 2024-08-16
Infinite recursion in `TreeLike#replaceLeavesMaybe`.


hatyo:
This method enables replacing the leaves of a `TreeLike` via a replacement operator, currently, any newly added leaves are passed again to the replacement operator, if the customer is not careful, this could lead to infinite recursion.

This is what's happening when trying to translate `Value` that looks like `_.P3` where QOV `_` is correlated to say `A` with a translation map that maps alias `A` to `_.P4` where `_` continues to be corrected to `A` (the representation is somewhat confusing, considering that `_` is actually a leaf in `_.P3` `Value`-tree), this causes infinite replacement recursion, effectively because the replacement rule reproduces the LHS on the RHS `[X] -> [Y -> X]`

PR #2885, opened 2024-08-16
Add `TreeLike#replaceLeavesMaybe` overload that does not pass new leaves to customer

hatyo:
This addresses #2884.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
@@ -29,7 +29,6 @@ import com.google.common.collect.Streams;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -210,24 +212,51 @@ default <V extends TreeLike<V>> Optional<V> mapMaybe(@Nonnull final BiFunction<T
...
+     * methods providing generic tree folds and maps, this method provides an easy way to mutate the leaves of the
+     * tree-like rooted at {@code this}, newly added leaves are passed again to the {@code replaceOperator}.
+     * @param replaceOperator an operator that translates a tree-like of {@code T} into another tree-like of type
+     *        {@code T}.
+     * @return an {@link Optional} of a tree-like object that is result of the tree-map operation if the fold of the
+     *         tree rooted at {@code this} exists, {@code Optional.empty()} otherwise
+     */
+    @Nonnull
+    default Optional<T> replaceLeavesMaybe(@Nonnull final UnaryOperator<T> replaceOperator) {
+        return replaceLeavesMaybe(replaceOperator, true);
```
normen662, line 226: Shouldn't this be false? I felt better if the default would be false.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
hatyo: done, plus some minor cleanup.

arnaud-lacurie (approved):

PR #2887, opened 2024-08-16
Resolves #2886: Add timings to "Repartitioned records" log

### LINKED ISSUE #2886: Log detailed metrics during Lucene repartitioning
The FDBStoreTimer used when repartitioning Lucene indexes is used for many transactions by the IndexingMerger. This can make it hard to figure out why a repartitioning action is slow. We should add some crude metrics to the `Repartitioned Records` debug log that happens at the end.

ScottDugas:
This adds a breakdown of the time, plus a metrics diff for the store timer.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
@@ -48,6 +48,8 @@ import com.apple.foundationdb.record.metadata.RecordType;
 import com.apple.foundationdb.record.metadata.expressions.FieldKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.GroupingKeyExpression;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
+import com.apple.foundationdb.record.provider.common.StoreTimer;
+import com.apple.foundationdb.record.provider.common.StoreTimerSnapshot;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexableRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBIndexedRecord;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
```diff
@@ -1070,15 +1088,41 @@ private CompletableFuture<Integer> moveDocsFromPartition(@Nonnull final LuceneRe
...
+                            logMessage.addKeyAndValue("searchMicros", TimeUnit.NANOSECONDS.toMicros(timings.searchNanos - timings.initializationNanos));
+                            logMessage.addKeyAndValue("clearInfoMicros", TimeUnit.NANOSECONDS.toMicros(timings.clearInfoNanos - timings.searchNanos));
+                            if (timings.emptyingNanos > 0) {
+                                logMessage.addKeyAndValue("emptyingMicros", TimeUnit.NANOSECONDS.toMicros(timings.emptyingNanos - timings.clearInfoNanos));
+                            }
+                            if (timings.deleteNanos > 0) {
+                                logMessage.addKeyAndValue("deleteMicros", TimeUnit.NANOSECONDS.toMicros(timings.deleteNanos - timings.clearInfoNanos));
+                            }
+                            if (timings.metadataUpdateNanos > 0) {
+                                logMessage.addKeyAndValue("metadataUpdateMicros", TimeUnit.NANOSECONDS.toMicros(timings.metadataUpdateNanos - timings.deleteNanos));
```
ohadzeliger, line 1114: What if `timings.deleteNanos` == 0?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LucenePartitioner.java
ScottDugas: Those are set in the same `if` block, so they will either both be set, or both unset.

ohadzeliger (approved):

PR #2889, opened 2024-08-21
make ConstantObjectValue implement CreatesDynamicTypesValue

normen662:


johscheuer (approved):
LGTM 👍

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ConstantObjectValue.java
```diff
@@ -49,7 +49,7 @@
 /**
  * Represents a constant value that references a constant in __CONST__ binding of {@link EvaluationContext}.
  */
-public class ConstantObjectValue extends AbstractValue implements LeafValue, Value.RangeMatchableValue {
+public class ConstantObjectValue extends AbstractValue implements LeafValue, Value.RangeMatchableValue, CreatesDynamicTypesValue {
```
hatyo, line 52: This is because a `ConstantObjectValue` can wrap a non-primitive type whose `Type` is something that some visitors should know about for collection and bookkeeping, correct?

normen662: yes

hatyo (dismissed):

hatyo (approved):

PR #2895, opened 2024-08-23
Fixes #2894: Make luceneOnlineIndexingTestWithAllRecordUpdates not flaky

### LINKED ISSUE #2894: LuceneOnlineIndexingTest.luceneOnlineIndexingTestWithAllRecordUpdates is flaky
Sometimes `LuceneOnlineIndexingTest.luceneOnlineIndexingTestWithAllRecordUpdates` fails with:
```
LuceneOnlineIndexingTest > luceneOnlineIndexingTestWithAllRecordUpdates() FAILED
    org.opentest4j.AssertionFailedError: expected: <6> but was: <0>
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
        at app//org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
        at app//com.apple.foundationdb.record.lucene.LuceneOnlineIndexingTest.luceneOnlineIndexingTestWithAllRecordUpdates(LuceneOnlineIndexingTest.java:419)
```

ScottDugas:
This does two things:
1. It changes it to use a random seed so that the randomness is only picked up if run with `-Ptests.includeRandom`
2. It fixes the test so that it will always generate 5 unique documents.
The underlying issue is that sometimes it would pick two documents with the same docId, and but it indirectly expected there to be 5 created.

Because this felt heavyweight I also added a new `@RandomSeedSource` that can be used with `@ParameterizedTest` to more easily provide random seeds for tests. There were other usages of `RandomizedTestUtils.randomSeeds`, but those used the same set of seeds for many tests, so I thought they were best left as-is.


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneOnlineIndexingTest.java
```diff
@@ -359,15 +365,17 @@ void luceneOnlineIndexingTestWithAllRecordUpdates() {
                     .setRecordStore(recordStore)
                     .setIndex(index)
                     .setInitialLimit(1)
+                    .setLimit(1)
```
ScottDugas, line 368: This is probably unnecessary given the current implementation.
`setLimit` sets the max number of records it will scan per transaction. Right now it doesn't increase the limit for the number of documents that are being indexed, but if it did index more than one record per transaction, it would not call the config loader after each transaction, and the assertion would break.
Thus, I figured it was worth leaving this change in.

arnaud-lacurie (approved):

PR #2899, opened 2024-08-28
Implement Value for the collate_{jre,icu} functions

MMcM:
Rebase #2835 on top of #2858.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
@@ -893,15 +893,15 @@ public class Comparisons {
     /**
      * A comparison with a constant value.
      */
-    public static class SimpleComparison implements Comparison {
+    public abstract static class SimpleComparisonBase implements Comparison {
         private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Simple-Comparison");
 
         @Nonnull
-        private final Type type;
+        protected final Type type;
         @Nonnull
         protected final Object comparand;
 
-        public SimpleComparison(@Nonnull Type type, @Nonnull Object comparand) {
+        protected SimpleComparisonBase(@Nonnull Type type, @Nonnull Object comparand) {
             this.type = type;
             this.comparand = comparand;
         }
@@ -969,15 +969,6 @@ public class Comparisons {
             return type;
         }
 
-        @Nonnull
-        @Override
-        public Comparison withType(@Nonnull final Type newType) {
-            if (type == newType) {
-                return this;
-            }
-            return new SimpleComparison(newType, comparand);
-        }
-
         @Nullable
         @Override
         public Value getValue() {
@@ -1015,7 +1006,7 @@ public class Comparisons {
             if (o == null || getClass() != o.getClass()) {
                 return false;
             }
-            SimpleComparison that = (SimpleComparison) o;
+            SimpleComparisonBase that = (SimpleComparisonBase) o;
             return type == that.type &&
                     Objects.equals(toClassWithRealEquals(comparand), toClassWithRealEquals(that.comparand));
         }
@@ -1053,6 +1044,24 @@ public class Comparisons {
         @Override
         public Comparison translateCorrelations(@Nonnull final TranslationMap translationMap) {
             return this;
+        }
+    }
+
+    /**
+     * A comparison with a constant value.
+     */
+    public static class SimpleComparison extends SimpleComparisonBase {
+        public SimpleComparison(@Nonnull Type type, @Nonnull Object comparand) {
+            super(type, comparand);
+        }
+
+        @Nonnull
+        @Override
+        public Comparison withType(@Nonnull final Type newType) {
+            if (type == newType) {
+                return this;
+            }
+            return new SimpleComparison(newType, comparand);
         }
 
         @Nonnull
@@ -1133,11 +1142,11 @@ public class Comparisons {
     /**
      * A comparison with a bound parameter, as opposed to a literal constant in the query.
      */
-    public static class ParameterComparison implements ComparisonWithParameter {
+    public abstract static class ParameterComparisonBase implements ComparisonWithParameter {
         private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Parameter-Comparison");
 
         @Nonnull
-        private final Type type;
+        protected final Type type;
         @Nonnull
         protected final String parameter;
         @Nullable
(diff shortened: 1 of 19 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryKeyExpression.java
```diff
@@ -326,24 +332,86 @@ public int planHash(@Nonnull final PlanHashMode mode) {
         public int queryHash(@Nonnull final QueryHashKind hashKind) {
             return HashUtils.queryHash(hashKind, SIMPLE_COMPARISON_BASE_HASH, super.queryHash(hashKind), getKeyExpression());
         }
+
+        @Nonnull
+        @Override
+        public Comparisons.Comparison withType(@Nonnull final Comparisons.Type newType) {
```
normen662, line 338: This comment applies to the planHash here which seems to use `SIMPLE_COMPARISON_BASE_HASH` but should define its own I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryKeyExpression.java
```diff
@@ -258,24 +265,23 @@ private QueryKeyExpressionWithComparison nullComparison(@Nonnull Comparisons.Typ
...
-            return new QueryKeyExpressionWithComparison(keyExpression, new ConversionParameterComparison(type, param, conversion));
+        if (keyExpression.getComparandConversionFunction() != null) {
+            return new QueryKeyExpressionWithComparison(keyExpression, new ConversionParameterComparison(type, param, keyExpression));
         } else {
             return new QueryKeyExpressionWithComparison(keyExpression, new Comparisons.ParameterComparison(type, param));
         }
     }
 
-    private final class ConversionSimpleComparison extends Comparisons.SimpleComparison {
+    private static final class ConversionSimpleComparison extends Comparisons.SimpleComparisonBase {
```
normen662, line 275: Should this maybe moved to `Comparisons`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -654,6 +654,14 @@ private enum BinaryPhysicalOperator {
         NEQ_BYU(Comparisons.Type.NOT_EQUALS, Type.TypeCode.BYTES, Type.TypeCode.UNKNOWN, (l, r) -> null),
         NEQ_BYBY(Comparisons.Type.NOT_EQUALS, Type.TypeCode.BYTES, Type.TypeCode.BYTES, (l, r) -> Comparisons.evalComparison(Comparisons.Type.NOT_EQUALS, l, r)),
         NEQ_UBY(Comparisons.Type.NOT_EQUALS, Type.TypeCode.UNKNOWN, Type.TypeCode.BYTES, (l, r) -> null),
+        LT_BYU(Comparisons.Type.LESS_THAN, Type.TypeCode.BYTES, Type.TypeCode.UNKNOWN, (l, r) -> null),
```
normen662, line 657: I know this is just following the same patterns but we shouldn't have implicit promotions for anything with `Type.TypeCode.UNKNOWN` in it. That's just plain wrong. `UNKNOWN` is reserved for stuff created by the heuristic planner, but that code path should never really use this `Value`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBCollateQueryTest.java
```diff
@@ -194,7 +194,24 @@ public void rangeScan() throws Exception {
         assertThat(plan, indexScan("collated_name"));
     }
 
-    @Test
+    @Disabled // TODO need an issue to document this: We need to be able to reapply a predicate for these cases
```
normen662, line 197:
```suggestion
    @Disabled // TODO https://github.com/FoundationDB/fdb-record-layer/issues/2911
```

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
MMcM: So those should all be removed? Will something break?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CollateValue.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+    }
+
+    @Nonnull
+    private static Value encapsulate(@Nonnull final TextCollatorRegistry collatorRegistry,
+                                     @Nonnull final List<? extends Typed> arguments) {
+        final int nargs = arguments.size();
+        Verify.verify(nargs >= 1 && nargs <= 3);
+        final Typed stringArg = arguments.get(0);
+        SemanticException.check(stringArg.getResultType().isPrimitive(), SemanticException.ErrorCode.ARGUMENT_TO_ARITHMETIC_OPERATOR_IS_OF_COMPLEX_TYPE);
```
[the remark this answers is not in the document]
MMcM: Added

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -893,15 +893,15 @@ public static String toPrintable(@Nullable Object value) {
     /**
      * A comparison with a constant value.
      */
-    public static class SimpleComparison implements Comparison {
+    public abstract static class SimpleComparisonBase implements Comparison {
         private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Simple-Comparison");
```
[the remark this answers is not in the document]
MMcM: Yes. I think the concrete classes will have the same state and the same hash.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryKeyExpression.java
MMcM: It's `QueryKeyExpressions. SIMPLE_COMPARISON_BASE_HASH`, not `SimpleComparisonBase. BASE_HASH`, so it already does define its own. They had to be in the parent class because these classes didn't used to be `static` and Java limitations. But now that they are `static`, they can be (renamed and) moved down.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryKeyExpression.java
MMcM: I'm reluctant because of the `QueryKeyExpression` field, which would then sort-of create a circular dependency of concepts.

normen662: Okay.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
normen662: They should all be removed, but it is possible that something breaks downstream. Can you log an issue?

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
MMcM: https://github.com/FoundationDB/fdb-record-layer/issues/2915

PR #2908, opened 2024-09-05
Drop readable unique pending

ScottDugas:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintainer.java
```diff
@@ -591,6 +591,11 @@ public RecordCursor<IndexEntry> scanUniquenessViolations(@Nonnull TupleRange ran
         return RecordCursor.empty(executor);
     }
 
+    @Override
+    public CompletableFuture<Void> clearUniquenessViolations() {
+        throw new RecordCoreException("Lucene index does not support uniqueness constraint");
```
jjezra, line 596:
```suggestion
      if (state.index.isUnique()) {
            throw new RecordCoreException("Lucene index does not support uniqueness constraint");
      }
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMergeTest.java
```diff
@@ -252,6 +252,11 @@ public RecordCursor<IndexEntry> scanUniquenessViolations(@Nonnull final TupleRan
                     throw new UnsupportedOperationException();
                 }
 
+                @Override
+                public CompletableFuture<Void> clearUniquenessViolations() {
+                    throw new UnsupportedOperationException();
```
jjezra, line 257: Would this index ever become readable?

jjezra (commented):
I am reviewing this PR while in Draft mode (vacation it is), my apologizes for any premature nit-picking.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreUniqueIndexTest.java
```diff
@@ -303,74 +338,458 @@ public void changeIndexAtFixedSubspaceKey() throws Exception {
...
+
+        @Nonnull
+        @Override
+        public RecordCursor<IndexEntry> scanUniquenessViolations(@Nonnull final TupleRange range, @Nullable final byte[] continuation, @Nonnull final ScanProperties scanProperties) {
+            return underlying.scanUniquenessViolations(range, continuation, scanProperties);
+        }
+
+        @Override
+        public CompletableFuture<Void> clearUniquenessViolations() {
+            return super.clearUniquenessViolations();
```
alecgrieser, line 721: It might be worth highlighting that this method is the only one that _doesn't_ delegate to the value index maintainer, resulting in `clearUniquenessViolations` being a no-op. I think it's pretty clear form the index name that that's the intention, but highlighting that (somewhat subtle) difference is probably good for future maintainers trying to understand how it achieves that goal.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreUniqueIndexTest.java
```diff
@@ -303,74 +338,458 @@ public void changeIndexAtFixedSubspaceKey() throws Exception {
...
+     * {@link IndexMaintainer#clearUniquenessViolations}, the old flow of increasing the version when changing the state
+     * still works.
+     * @param allowReadableUniquePending whether to allow {@link IndexState#READABLE_UNIQUE_PENDING}
+     * @throws Exception if there is an issue
+     */
+    @ParameterizedTest(name = "bumpVersionWhenChangingToNonUnique(allowReadableUniquePending={0})")
+    @BooleanSource
+    void bumpVersionWhenChangingToNonUnique(boolean allowReadableUniquePending) throws Exception {
+        final DropUniquenessConstraint dropUniquenessConstraint = new DropUniquenessConstraint(
+                allowReadableUniquePending, NO_UNIQUE_CLEAR_INDEX_TYPE, true);
```
alecgrieser, line 414: Is it worth adding a test of what happens with the `NO_UNIQUE_CLEAR_INDEX_TYPE` type and we don't bump the version? I think I'd expect it to act like the index doesn't have any uniqueness constraints, but we'd still be able to find some by scanning the index violations subspace

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1520,16 +1520,20 @@ public void addIndexUniquenessCommitCheck(@Nonnull Index index, @Nonnull Complet
     @Nonnull
     @VisibleForTesting
     CompletableFuture<Void> whenAllIndexUniquenessCommitChecks(@Nonnull Index index) {
+        return AsyncUtil.whenAll(checkAllIndexUniquenessChecks(commitCheck -> commitCheck.getIndex().equals(index)));
```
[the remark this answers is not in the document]
ScottDugas: Yeah, I filed another issue for that: #2919 and will fix separately.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreUniqueIndexTest.java
ScottDugas: Good call. I added a javadoc. And since this is now using the default implementation, instead of defining its own method to throw, I renamed it.

ScottDugas (commented):
I addressed all the comment except the one that I put in the separate PR: [2920: Fixes #2919: scope the IndexUniquenessCommitCheck by subpsace](https://github.com/FoundationDB/fdb-record-layer/pull/2920)

Once #2920 is merged, I will merge it into this, and adjust as appropriate.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
```diff
@@ -775,6 +775,34 @@ public synchronized List<CommitCheckAsync> getCommitChecks(@Nonnull Predicate<Co
...
+            @Nonnull Predicate<Throwable> shouldSwallow) {
+        return commitChecks.entrySet().stream()
+                .filter(entry -> filter.apply(entry.getValue()))
+                .map(entry -> MoreAsyncUtil.swallowException(entry.getValue().checkAsync(), shouldSwallow)
+                        .whenComplete((result, err) -> removeCommitCheck(entry.getKey())))
+                .collect(Collectors.toList());
+    }
+
+    private synchronized void removeCommitCheck(String key) {
+        commitChecks.remove(key);
```
alecgrieser, line 803: I'd be a little concerned about a `ConcurrentModificationException`s here, if `removeCommitCheck` attempts to remove an element while the `stream` is still going over it. The alternative is probably to do something like (1) collect the filtered commit checks into a list and then (2) remove them only once the iteration is done.

alecgrieser (dismissed):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordContext.java
ScottDugas: `removeCommitCheck` and `removeCommitChecks` are both `synchronized`, so it won't be able to remove the check while it is being iterated over, unless I'm missing something

alecgrieser (approved):

PR #2918, opened 2024-09-18
Resolve #2917: Allow converting a single/multi target indexing session to mutual one

### LINKED ISSUE #2917: Allow converting a single/multi target indexing session to mutual one
Right now, if an indexing session is very slow, the only way to convert it to mutual (and speed it up) is to disable the index and lose any progress that was already done.

jjezra:
This change should allow speeding up an indexing session with insufficient performance by converting it to mutual session without loosing all the progress that was already made.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -1118,7 +1118,6 @@ public class OnlineIndexer implements AutoCloseable {
         private final DesiredAction ifMismatchPrevious;
         private final DesiredAction ifReadable;
         private final boolean allowUniquePendingState;
-        private final boolean allowTakeoverContinue;
         private final Set<TakeoverTypes> allowedTakeoverSet;
         private final long checkIndexingMethodFrequencyMilliseconds;
         private final boolean mutualIndexing;
@@ -1167,8 +1166,7 @@ public class OnlineIndexer implements AutoCloseable {
          * @param ifMismatchPrevious desired action if the index is partly built, but by a different method then currently requested
          * @param ifReadable desired action if the existing index state is READABLE (i.e. already built)
          * @param allowUniquePendingState if false, forbid {@link IndexState#READABLE_UNIQUE_PENDING} state
-         * @param allowTakeoverContinue if true and possible, allow indexing continuation of a different indexing method
-         * @param allowedTakeoverSet if non-null, allow a subset of indexing type conversion (non-null overrides allowTakeoverContinue)
+         * @param allowedTakeoverSet a subset of {@link TakeoverTypes}, with allowed indexing type conversion
          * @param mutualIndexing if true, use mutual indexing (i.e., index in a way that allows other processes to cooperatively build the index)
          * @param mutualIndexingBoundaries if present, use this predefined list of ranges. Else, split ranges by shards
          * @param allowUnblock if true, allow unblocking
@@ -1179,7 +1177,7 @@ public class OnlineIndexer implements AutoCloseable {
         @SuppressWarnings("squid:S00107") // too many parameters
         private IndexingPolicy(@Nullable String sourceIndex, @Nullable Object sourceIndexSubspaceKey, boolean forbidRecordScan,
                                DesiredAction ifDisabled, DesiredAction ifWriteOnly, DesiredAction ifMismatchPrevious, DesiredAction ifReadable,
-                               boolean allowUniquePendingState, boolean allowTakeoverContinue, Set<TakeoverTypes> allowedTakeoverSet,
+                               boolean allowUniquePendingState, Set<TakeoverTypes> allowedTakeoverSet,
                                long checkIndexingMethodFrequencyMilliseconds,
                                boolean mutualIndexing, List<Tuple> mutualIndexingBoundaries,
                                boolean allowUnblock, String allowUnblockId,
@@ -1193,7 +1191,6 @@ public class OnlineIndexer implements AutoCloseable {
             this.ifMismatchPrevious = ifMismatchPrevious;
             this.ifReadable = ifReadable;
             this.allowUniquePendingState = allowUniquePendingState;
-            this.allowTakeoverContinue = allowTakeoverContinue;
             this.allowedTakeoverSet = allowedTakeoverSet;
             this.checkIndexingMethodFrequencyMilliseconds = checkIndexingMethodFrequencyMilliseconds;
             this.mutualIndexing = mutualIndexing;
@@ -1266,7 +1263,6 @@ public class OnlineIndexer implements AutoCloseable {
                     .setIfMismatchPrevious(ifMismatchPrevious)
                     .setIfReadable(ifReadable)
                     .allowUniquePendingState(allowUniquePendingState)
-                    .allowTakeoverContinue(allowTakeoverContinue)
                     .allowTakeoverContinue(allowedTakeoverSet)
                     .checkIndexingStampFrequencyMilliseconds(checkIndexingMethodFrequencyMilliseconds)
                     .setMutualIndexing(mutualIndexing)
@@ -1320,7 +1316,7 @@ public class OnlineIndexer implements AutoCloseable {
                 case WRITE_ONLY:    return getIfWriteOnly();
                 case READABLE:      return getIfReadable();
                 case READABLE_UNIQUE_PENDING: return DesiredAction.MARK_READABLE;
-                default: throw new RecordCoreException("bad index state: " + state.toString());
+                default: throw new RecordCoreException("bad index state: " + state);
             }
         }
 
@@ -1336,13 +1332,17 @@ public class OnlineIndexer implements AutoCloseable {
         /**
          * If true, allow - in some specific cases - to continue building an index that was partly built by a different indexing method.
          * (See {@link Builder#allowTakeoverContinue(Collection)} and {@link Builder#allowTakeoverContinue(boolean)}).
-         * @param newMethod the new (attempting to continue) indexing method
-         * @param oldMethod the old (previously used) indexing method
+         * @param newStamp the new (attempting to continue) indexing type stamp
+         * @param oldStamp the old (previously used) indexing type stamp
          * @return true if allowed
          */
-        public boolean shouldAllowTypeConversionContinue(IndexBuildProto.IndexBuildIndexingStamp.Method newMethod,
-                                                         IndexBuildProto.IndexBuildIndexingStamp.Method oldMethod) {
+        public boolean shouldAllowTypeConversionContinue(IndexBuildProto.IndexBuildIndexingStamp newStamp,
+                                                         IndexBuildProto.IndexBuildIndexingStamp oldStamp) {
+            final IndexBuildProto.IndexBuildIndexingStamp.Method newMethod = newStamp.getMethod();
+            final IndexBuildProto.IndexBuildIndexingStamp.Method oldMethod = oldStamp.getMethod();
             if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                // Here: if allowed, an index that was partly built as a set will be continued as a single. Note that other target
+                // indexes from the original set may have already continued individually.
                 if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
                     return isTypeConversionAllowed(TakeoverTypes.MULTI_TARGET_TO_SINGLE);
                 }
@@ -1351,20 +1351,29 @@ public class OnlineIndexer implements AutoCloseable {
                 }
             }
             if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
-                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS ||
-                        oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                if (!isTypeConversionAllowed(TakeoverTypes.BY_RECORDS_TO_MUTUAL)) {
+                    // This conversion is not allowed, check no further
+                    return false;
+                }
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
+                    // This conversion should be allowed if:
+                    // a) The new mutual indexing set is identical to the old one
+                    // b) The mutual indexing continues only one target index (splitting it to one at a tiem)
+                    return newStamp.getTargetIndexCount() == 1 ||
+                            (newStamp.getTargetIndexCount() == oldStamp.getTargetIndexCount() &&
+                                    new HashSet<>(newStamp.getTargetIndexList()).containsAll(oldStamp.getTargetIndexList()));
+                }
+
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
                     // Note that single-target-mutual is a private case of multi-target-mutual
-                    return isTypeConversionAllowed(TakeoverTypes.BY_RECORDS_TO_MUTUAL);
+                    return newStamp.getTargetIndexCount() == 1; // No real need to check the stamp's index name
                 }
             }
             return false;
         }
 
         private boolean isTypeConversionAllowed(TakeoverTypes takeoverType) {
-            return
-                    allowedTakeoverSet == null ?
-                    allowTakeoverContinue :
-                    allowedTakeoverSet.contains(takeoverType);
+            return allowedTakeoverSet != null && allowedTakeoverSet.contains(takeoverType);
         }
 
 
@@ -1426,8 +1435,7 @@ public class OnlineIndexer implements AutoCloseable {
             private DesiredAction ifWriteOnly = DesiredAction.CONTINUE;
             private DesiredAction ifMismatchPrevious = DesiredAction.CONTINUE;
             private DesiredAction ifReadable = DesiredAction.CONTINUE;
-            private boolean doAllowUniqueuPendingState = false;
-            private boolean doAllowTakeoverContinue = false;
+            private boolean doAllowUniquePendingState = false;
             private Set<TakeoverTypes> allowedTakeoverSet = null;
             private long checkIndexingStampFrequency = 60_000;
             private boolean useMutualIndexing = false;
@@ -1563,7 +1571,7 @@ public class OnlineIndexer implements AutoCloseable {
              * @return this builder
              */
             public Builder allowUniquePendingState(boolean allow) {
-                this.doAllowUniqueuPendingState = allow;
+                this.doAllowUniquePendingState = allow;
                 return this;
             }
 
@@ -1584,13 +1592,13 @@ public class OnlineIndexer implements AutoCloseable {
              * <li>"Single index by records" may continue a multi index session.</li>
              * <li>"Single index by records" may continue a mutually built session.</li>
              *  </ul>
-             *  This function may be too general; {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
+             *  This function will allow/disallow the all possible conversion types. {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
              *  Note - if {@link #allowTakeoverContinue(Collection)} is called with a non-null argument, it will override this function's argument.
              * @param allow if true, allow takeover.
              * @return this builder
              */
             public Builder allowTakeoverContinue(boolean allow) {
-                this.doAllowTakeoverContinue = allow;
+                this.allowedTakeoverSet = allow ? EnumSet.allOf(TakeoverTypes.class) : EnumSet.noneOf(TakeoverTypes.class);
                 return this;
             }
 
@@ -1731,7 +1739,7 @@ public class OnlineIndexer implements AutoCloseable {
                 }
                 return new IndexingPolicy(sourceIndex, sourceIndexSubspaceKey, forbidRecordScan,
                         ifDisabled, ifWriteOnly, ifMismatchPrevious, ifReadable,
-                        doAllowUniqueuPendingState, doAllowTakeoverContinue, allowedTakeoverSet,
+                        doAllowUniquePendingState, allowedTakeoverSet,
                         checkIndexingStampFrequency,
                         useMutualIndexing, useMutualIndexingBoundaries, allowUnblock, allowUnblockId,
                         initialMergesCountLimit, reverseScanOrder);
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -1150,7 +1150,11 @@ public class OnlineIndexer implements AutoCloseable {
              * Allow a single target indexing continuation when the index was partly built as mutual indexing (either single or multi target).
              */
             MUTUAL_TO_SINGLE,
-            // TODO:  TO_MUTUAL - either single or multi target to mutual indexing
+            /**
+             * Allow converting (single or multi target) by-records to mutual.
+             * Note that single-target-mutual is a private case of multi-target-mutual. The non-mutual distinction is used for backward compatibility.
+             */
+            BY_RECORDS_TO_MUTUAL,
         }
 
         /**
@@ -1338,22 +1342,29 @@ public class OnlineIndexer implements AutoCloseable {
          */
         public boolean shouldAllowTypeConversionContinue(IndexBuildProto.IndexBuildIndexingStamp.Method newMethod,
                                                          IndexBuildProto.IndexBuildIndexingStamp.Method oldMethod) {
-            if (allowedTakeoverSet != null) {
-                if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
-                    if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
-                        return allowedTakeoverSet.contains(TakeoverTypes.MULTI_TARGET_TO_SINGLE);
-                    }
-                    if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
-                        return allowedTakeoverSet.contains(TakeoverTypes.MUTUAL_TO_SINGLE);
-                    }
+            if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
+                    return isTypeConversionAllowed(TakeoverTypes.MULTI_TARGET_TO_SINGLE);
                 }
-                return false;
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
+                    return isTypeConversionAllowed(TakeoverTypes.MUTUAL_TO_SINGLE);
+                }
             }
-            return allowTakeoverContinue &&
-                   // Takeover is allowed only in certain cases
-                   (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS &&
-                    (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS ||
-                     oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS));
+            if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS ||
+                        oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
+                    // Note that single-target-mutual is a private case of multi-target-mutual
+                    return isTypeConversionAllowed(TakeoverTypes.BY_RECORDS_TO_MUTUAL);
+                }
+            }
+            return false;
+        }
+
+        private boolean isTypeConversionAllowed(TakeoverTypes takeoverType) {
+            return
+                    allowedTakeoverSet == null ?
+                    allowTakeoverContinue :
+                    allowedTakeoverSet.contains(takeoverType);
         }
 
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
@@ -1592,7 +1592,7 @@ public class OnlineIndexer implements AutoCloseable {
              * <li>"Single index by records" may continue a multi index session.</li>
              * <li>"Single index by records" may continue a mutually built session.</li>
              *  </ul>
-             *  This function will allow/disallow the all possible conversion types. {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
+             *  This function will allow/disallow all possible conversion types. {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
              *  Note - if {@link #allowTakeoverContinue(Collection)} is called with a non-null argument, it will override this function's argument.
              * @param allow if true, allow takeover.
              * @return this builder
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1338,22 +1342,29 @@ public boolean shouldAllowUniquePendingState(FDBRecordStore store) {
...
+                    return isTypeConversionAllowed(TakeoverTypes.BY_RECORDS_TO_MUTUAL);
+                }
+            }
+            return false;
+        }
+
+        private boolean isTypeConversionAllowed(TakeoverTypes takeoverType) {
+            return
+                    allowedTakeoverSet == null ?
+                    allowTakeoverContinue :
```
ScottDugas, line 1366: Perhaps it would make sense to push this into the builder, and have:
```
            public Builder allowTakeoverContinue(boolean allow) {
                this.allowedTakeoverSet = allow ? EnumSet.allOf(TakeoverTypes.class) : EnumSet.noneOf(TakeoverTypes.class);
                return this;
            }
```
and change the default value of `this.allowedTakeoverSet` to `EnumSet.noneOf(TakeoverTypes.class)` instead of `null`

Then this method just becomes `allowedTakeoverSet.contains(takeoverType)`, and may make sense to inline.

Perhaps better as a followup PR though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1338,22 +1342,29 @@ public boolean shouldAllowUniquePendingState(FDBRecordStore store) {
...
                                                          IndexBuildProto.IndexBuildIndexingStamp.Method oldMethod) {
-            if (allowedTakeoverSet != null) {
-                if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
-                    if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
-                        return allowedTakeoverSet.contains(TakeoverTypes.MULTI_TARGET_TO_SINGLE);
-                    }
-                    if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
-                        return allowedTakeoverSet.contains(TakeoverTypes.MUTUAL_TO_SINGLE);
-                    }
+            if (newMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
```
ScottDugas, line 1345: So the total set of methods is:
```
  enum Method {
    INVALID      = 0;  // let zero be invalid
    BY_RECORDS   = 1;  // indexing by records - scanning all records in a record store
    BY_INDEX     = 2;  // indexing by a source index. When source index points to all possible records of the target index.
    SCRUB_REPAIR = 3;  // while readable, verify index validity. Repair if needed.
    MULTI_TARGET_BY_RECORDS = 4; // scan records, build multiple target indexes
    MUTUAL_BY_RECORDS = 5; // scan records, build multiple target indexes, while allowing multiple indexer processes
    NONE = 6; // return this stamp in query when null - never to be written in the DB
  };

```

And it would never make sense to go to, or from `INVALID` or `SCRUB_REPAIR`.
`BY_INDEX` to any of the others I think is not possible because the ranges would be wrong.
`NONE` to anything is implicitly supported (to the extent that `NONE` is a real type)

And the only other one that is not covered after this change `BY_RECORDS` to `MULTI_TARGET_BY_RECORDS`, although you could cause that to happen by first switching to `MUTUAL_BY_RECORDS`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1338,22 +1342,29 @@ public boolean shouldAllowUniquePendingState(FDBRecordStore store) {
...
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MULTI_TARGET_BY_RECORDS) {
+                    return isTypeConversionAllowed(TakeoverTypes.MULTI_TARGET_TO_SINGLE);
+                }
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.MUTUAL_BY_RECORDS) {
+                    return isTypeConversionAllowed(TakeoverTypes.MUTUAL_TO_SINGLE);
```
ScottDugas, line 1350: I know this existed before, but the comment below makes me wonder what happens if you have a `MUTUAL_BY_RECORDS` that has multiple targets? Does it get split up?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -948,4 +948,61 @@ void testForbidConversionOfActiveMultiTarget() throws InterruptedException {
...
+                indexBuilder.buildIndex();
+            }
+        });
+        t1.start();
+        startBuildingSemaphore.acquire();
+        startBuildingSemaphore.release();
+        // Try as mutual and fail
+        try (OnlineIndexer indexBuilder = newIndexerBuilder(indexes)
+                .setIndexingPolicy(OnlineIndexer.IndexingPolicy.newBuilder()
+                        .setMutualIndexing(true))
```
ScottDugas, line 998: The key thing here is that it doesn't configure the policy to allow `BY_RECORDS_TO_MUTUAL`, right? Otherwise this indexer would keep building, and the other would fail.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -403,11 +403,15 @@ void testMutualIndexingCrashAndRefuseContinueNonMutually() {
         assertAllValidated(indexes);
     }
 
-    OnlineIndexer.IndexingPolicy.Builder mutualTakeOverIndexingPolicy(boolean explicit) {
+    OnlineIndexer.IndexingPolicy.Builder mutualTakeOverIndexingPolicy(boolean explicit, boolean mutual) {
```
ScottDugas, line 406: Nit:
```suggestion
    OnlineIndexer.IndexingPolicy.Builder mutualTakeOverIndexingPolicy(boolean explicit, boolean toMutual) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
         assertAllValidated(indexes);
     }
 
+    @ParameterizedTest
```
ScottDugas, line 458: It seems like there should be a test where there are two indexers,
1. building `MULTI_TARGET_BY_RECORDS` and has `MUTUAL_TO_SINGLE` allowed
2. building `MUTUAL_BY_RECORDS` and has `BY_RECORDS_TO_MUTUAL` allowed

I believe the two will both do indexing, with the state constantly swapping.
If that is the case, probably worth calling out in the documentation, because you basically have no isolation between the two, and would presumably get a lot of conflicts if you had the two running.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -403,11 +403,15 @@ void testMutualIndexingCrashAndRefuseContinueNonMutually() {
         assertAllValidated(indexes);
     }
 
-    OnlineIndexer.IndexingPolicy.Builder mutualTakeOverIndexingPolicy(boolean explicit) {
+    OnlineIndexer.IndexingPolicy.Builder mutualTakeOverIndexingPolicy(boolean explicit, boolean mutual) {
         final OnlineIndexer.IndexingPolicy.Builder builder = OnlineIndexer.IndexingPolicy.newBuilder();
+        final List<OnlineIndexer.IndexingPolicy.TakeoverTypes> conversionSet =
```
ScottDugas, line 408: I think it should be a separate PR, if at all, but it kind of seems like there should be a separate test class for takeover, rather than putting the tests in classes focused on the source or target.

That might also make it easier, or more obvious where some of this code could be de-duplicated and it could be easier to cover all the permutations that are allowed or not-allowed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+                    throw new RecordCoreException(testThrowMsg);
+                })
+                .build()) {
+            RecordCoreException e = assertThrows(RecordCoreException.class, indexBuilder::buildIndex);
+            assertTrue(e.getMessage().contains(testThrowMsg));
+        }
+
+        // Continue as mutual
+        final FDBStoreTimer timer = new FDBStoreTimer();
+        try (OnlineIndexer indexBuilder = newIndexerBuilder(indexes, timer)
```
ScottDugas, line 489: Since the goal of this change is to avoid losing the work done by the first indexer, it seems like this test should assert that the mutual indexing has to index some records, but not all the records.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+        assertAllValidated(indexes);
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testSingleTargetContinueAsMutual(boolean explicit) {
+        // Start building with a single target, stop, continue as mutual
+        // Make sure that the mutual indexing is unblocked
+        // Note that a single target session produces a different indexing type stamp than a multi target session
+        final Index index = new Index("indexA", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS);
```
ScottDugas, line 507: It might be worth extracting a helper method since, I think the only difference in the test code between this and `testMultiTargetContinueAsMutual` is the `indexes` list.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+                .setConfigLoader(old -> {
+                    throw new RecordCoreException(testThrowMsg);
+                })
+                .build()) {
+            RecordCoreException e = assertThrows(RecordCoreException.class, indexBuilder::buildIndex);
+            assertTrue(e.getMessage().contains(testThrowMsg));
+        }
+
+        // Continue as mutual
+        try (OnlineIndexer indexBuilder = newIndexerBuilder(index)
```
ScottDugas, line 530: Similarly, this should assert about the number of records scanned.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+            indexBuilder.buildIndex(true);
+        }
+
+        // validate
+        assertReadable(indexes);
+        assertAllValidated(indexes);
+    }
+
+    @ParameterizedTest
+    @BooleanSource
```
ScottDugas, line 543: This doesn't use the parameter
```suggestion
    @Test
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+        indexes.add(new Index("indexA", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS));
+
+        int numRecords = 32;
+        populateData(numRecords);
+
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        // Build as mutual, crash half way
```
ScottDugas, line 559:
```suggestion
        // Build as multi-target, crash half way
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+        List<Index> indexes = List.of(index);
+
+        int numRecords = 32;
+        populateData(numRecords);
+
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        // Build as mutual, crash half way
```
ScottDugas, line 517:
```suggestion
        // Build as single-target, crash half way
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+        indexes.add(new Index("indexA", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS));
+
+        int numRecords = 32;
+        populateData(numRecords);
+
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        // Build as mutual, crash half way
```
ScottDugas, line 475:
```suggestion
        // Build as multi-target, crash half way
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
         assertAllValidated(indexes);
     }
 
+    @ParameterizedTest
+    @BooleanSource
+    void testMultiTargetContinueAsMutual(boolean explicit) {
```
ScottDugas, line 460: It seems like there should also be tests of multi to-and-from mutual where they have different, potentially overlapping index sets.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
ScottDugas, line 458: It seems like there should be a test exemplifying the behavior around SynchronizedSession, and how the mutual indexing doesn't respect it.
I believe if you use SynchronizedSession, and try to takeover from MULTI to SINGLE, while the former is still active due to the SynchronizedSession, but mutual will not do that, it will takeover even if the MULTI just committed a second ago. 

Perhaps this should be broken out into a separate body of work to improve the SynchronizedSession relationship with MUTUAL, but it might be pretty easy to add those tests, and then reference an issue to improve the situation.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -948,4 +948,61 @@ void testForbidConversionOfActiveMultiTarget() throws InterruptedException {
         // happy indexes assertion
         assertReadable(indexes);
     }
+
+    @Test
+    void testForbidConversionOfActiveMultiTargetToMutual() throws InterruptedException {
```
ScottDugas, line 953: Does the "Active" matter, wouldn't this code forbid the conversion regardless.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +455,132 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = allIndexesHook(indexes);
+        openSimpleMetaData(hook);
+        disableAll(indexes);
+
+        // Build as mutual, crash half way
+        final String testThrowMsg = "Intentionally crash during test";
+        try (OnlineIndexer indexBuilder = newIndexerBuilder(indexes)
+                .setLimit(10)
+                .setConfigLoader(old -> {
+                    throw new RecordCoreException(testThrowMsg);
```
ScottDugas, line 564: This fails and then forbids converting to MUTUAL, should there be another test where it pauses, and continues, similar to: `testForbidConversionOfActiveMultiTargetToMutual` above.
Should there be a test similar to this one but forbidding mutual_to_multi?

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: ✅ 

(It was an easy change)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: As you've said, switching from `BY_INDEX` to any by-records method should never be allowed. 
Switching from `BY_RECORDS` to `MULTI_TARGET_BY_RECORDS` could have been allowed only if no records were scanned. 
It can make sense to verify that when switching from `BY_RECORDS` or `MULTI_TARGET_BY_RECORDS` to `MUTUAL_BY_RECORDS`, the very same target index set is being used. I'll try to add it.

jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: Yes. converting to `BY_RECORDS` will split any multi target. The original motivation was a fallback. 
I will add a comment to clarify.

jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: Yes. The test verifies that: 
1. The conversion is forbidden. 
2. The conversion attempt does not affect the original multi-target session, and that the multi-target session is completed successfully.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: (Did you mean the first item to be 
1. building `BY_RECORDS` and has `MUTUAL_TO_SINGLE` allowed
?)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: (Done in the helper function)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: Not really. The mutual does not use the sync lock.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: Do you mean build as mutual, pause, fail to convert to multi, then continue the mutual? 
If so: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1351,20 +1351,29 @@ public boolean shouldAllowTypeConversionContinue(IndexBuildProto.IndexBuildIndex
...
+                    // b) The mutual indexing continues only one target index (splitting it to one at a tiem)
+                    return newStamp.getTargetIndexCount() == 1 ||
+                            (newStamp.getTargetIndexCount() == oldStamp.getTargetIndexCount() &&
+                                    new HashSet<>(newStamp.getTargetIndexList()).containsAll(oldStamp.getTargetIndexList()));
+                }
+
+                if (oldMethod == IndexBuildProto.IndexBuildIndexingStamp.Method.BY_RECORDS) {
                     // Note that single-target-mutual is a private case of multi-target-mutual
-                    return isTypeConversionAllowed(TakeoverTypes.BY_RECORDS_TO_MUTUAL);
+                    return newStamp.getTargetIndexCount() == 1; // No real need to check the stamp's index name
```
ScottDugas, line 1369: What happens if the indexes are different?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
```diff
@@ -1584,13 +1592,13 @@ public Builder allowTakeoverContinue() {
              * <li>"Single index by records" may continue a multi index session.</li>
              * <li>"Single index by records" may continue a mutually built session.</li>
              *  </ul>
-             *  This function may be too general; {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
+             *  This function will allow/disallow the all possible conversion types. {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
```
ScottDugas, line 1595:
```suggestion
             *  This function will allow/disallow all possible conversion types. {@link #allowTakeoverContinue(Collection)} supports allowing a subset of these conversions.
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -910,19 +911,24 @@ void testForbidConversionOfActiveMultiTarget() throws InterruptedException {
...
-                        } catch (InterruptedException e) {
-                            throw new RuntimeException(e);
+                        if (passed.get()) {
+                            try {
+                                startBuildingSemaphore.release();
+                                pauseMutualBuildSemaphore.acquire(); // pause to try building indexes
+                            } catch (InterruptedException e) {
+                                throw new RuntimeException(e);
+                            }
+                            pauseMutualBuildSemaphore.release();
```
ScottDugas, line 928: Should this be in a `finally`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
```diff
@@ -970,19 +976,24 @@ void testForbidConversionOfActiveMultiTargetToMutual() throws InterruptedExcepti
...
```
ScottDugas, line 993: should this be in a `finally`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -556,29 +585,256 @@ void testMultiTargetForbidContinueAsMutual() {
...
+        // Take over as mutual
+        try (OnlineIndexer indexBuilder = newIndexerBuilder()
+                .setTargetIndexes(indexes)
+                .setIndexingPolicy(mutualTakeOverIndexingPolicy(false, true)
+                        .setMutualIndexing())
+                .build()) {
+            indexBuilder.buildIndex();
         }
+
+        // let the other thread finish, verify that it did was not completed
```
ScottDugas, line 769:
```suggestion
        // let the other thread finish, verify that it was not completed
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -440,7 +448,8 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
             try (OnlineIndexer indexBuilder = newIndexerBuilder()
                     .setIndex(index)
                     .setTimer(timer)
-                    .setIndexingPolicy(mutualTakeOverIndexingPolicy(explicit))
+                    .setIndexingPolicy(mutualTakeOverIndexingPolicy(explicit, false)
+                            .allowTakeoverContinue(true))
```
ScottDugas, line 452: doesn't `allowTakeover` here override the usage of `explicit` in the line above?
Wouldn't this be the same as:
```
                    .setIndexingPolicy(mutualTakeOverIndexingPolicy(false, false))
```

But I think the original version was intentional, to make sure that both `allowTakeoverContinue(true)` and `allowTakeoverContinue(List.of(...))` work the same.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
```diff
@@ -451,6 +460,383 @@ void testMutualIndexingCrashAndAllowContinueNonMutually(boolean explicit) {
...
+                    }
+                    passed.set(true);
+                    return old;
+                })
+                .build()) {
+            RecordCoreException e = assertThrows(RecordCoreException.class, indexBuilder::buildIndex);
+            assertTrue(e.getMessage().contains(testThrowMsg));
+        }
+
+        // Continue as mutual, by a different sublists of the indexes. All should fail
```
ScottDugas, line 605: Should there be a similar test for going `mutual` to `multi-target`?

ScottDugas (changes requested):
Mostly just typos, but one change looks suspicious.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: The old stamp does not have a target index (sort of backward compatibility, maybe it should be changed in the future). 
The new stamp is generated by the mutual indexer. If we check the stamp's validity here, we should check it everywhere.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexer.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMultiTargetTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexerMutualTest.java
jjezra: It is a leftover from a code change that was undone (but not fully undone - sorry). I removed line 452. 
✅

ScottDugas (approved):

ScottDugas (approved):

PR #664, opened 2019-07-22
Resolves #663: Enable the capability to add logging details to mapped exceptions

dsimmen:

PR #2924, opened 2024-09-19
Resolves #2923: Create a new version of AsyncToSync

### LINKED ISSUE #2923: Lucene to use a custom version of asyncToSync that does not inject exceptions
Since Lucene logic realizes Futures and uses the results for further logic, it needs to reason about the exceptions being thrown by the lower levels for these futures. The context's AsyncToSync implementation allows exceptions to be remapped which may hinder the ability to catch specific exceptions.

ohadzeliger:
The existing implementation of context AsyncToSync allows exceptions to be mapped to other exceptions. In order to be able to reason about the errors throws by futures when relalized by AsyncToSync this PR creates a new version of AsyncToSync, to be used by Lucene code, so that exception mapping is no longer taking place.
This is the first step in handling Lucene exceptions, the others being mapping to -and from- IOExceptions on Lucene code boundaries.

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
@@ -1,0 +1,162 @@
+/*
+ * LuceneConcurrency.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2024 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.lucene;
+
+import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.async.MoreAsyncUtil;
+import com.apple.foundationdb.record.RecordCoreException;
+import com.apple.foundationdb.record.logging.LogMessageKeys;
+import com.apple.foundationdb.record.provider.common.StoreTimer;
+import com.apple.foundationdb.record.provider.foundationdb.FDBDatabase;
+import com.apple.foundationdb.record.provider.foundationdb.FDBExceptions;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer;
+import org.apache.commons.lang3.tuple.Pair;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.apple.foundationdb.record.lucene.LuceneRecordContextProperties.LUCENE_USE_LEGACY_ASYNC_TO_SYNC;
+
+/**
+ * Utility class for methods related to synchronizing Futures.
+ */
+public class LuceneConcurrency {
+
+    /**
+     * An implementation of {@code asyncToSync} that is isolated from external exception injections.
+     * This implementation does NOT perform exception mapping, nor does it check whether the calls have
+     * "async" in the stack trace (as the original {@link FDBRecordContext#asyncToSync} does).
+     * This method is meant to be used internally in places where obtaining and using the result pf asynchronous
+     * operation is required.
+     * This method uses the {@link FDBDatabase#getAsyncToSyncTimeout} to find the period to use for the timeout.
+     * This method will throw the runtime exception that was thrown by the Future's realization in case such error
+     * occurred.
+     * This method will use the legacy  {@link FDBRecordContext#asyncToSync} if the LUCENE_USE_LEGACY_ASYNC_TO_SYNC
+     * property is set to TRUE (the default)
+     *
+     * @param event the timer event to use for recording the waits
+     * @param async the future to wait on
+     * @param recordContext the context to use for callback, recording the event and getting the timeout
+     *
+     * @return the result of the future's operation
+     */
+    @Nullable
+    @API(API.Status.INTERNAL)
+    public static <T> T asyncToSync(@Nonnull StoreTimer.Wait event, @Nonnull CompletableFuture<T> async, @Nonnull FDBRecordContext recordContext) {
+        if (recordContext.getPropertyStorage().getPropertyValue(LUCENE_USE_LEGACY_ASYNC_TO_SYNC)) {
+            return recordContext.asyncToSync(event, async);
+        }
+
+        if (recordContext.hasHookForAsyncToSync() && !MoreAsyncUtil.isCompletedNormally(async)) {
+            recordContext.getHookForAsyncToSync().accept(event);
+        }
+        if (async.isDone()) {
+            try {
+                return async.get();
+            } catch (ExecutionException ex) {
+                throw FDBExceptions.wrapException(ex);
+            } catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
+                // TODO: The original mapping for InterruptedException would be INTERNAL_ERROR, now it will be RecordCoreInterruptedException
+                throw FDBExceptions.wrapException(ex);
+            }
+        } else {
+            final Pair<Long, TimeUnit> asyncToSyncTimeout = recordContext.getDatabase().getAsyncToSyncTimeout(event);
+            Timeout timeout = (asyncToSyncTimeout == null) ? null : Timeout.of(asyncToSyncTimeout.getLeft(), asyncToSyncTimeout.getRight());
+            final FDBStoreTimer timer = recordContext.getTimer();
+            final long startTime = System.nanoTime();
+            try {
+                if (timeout != null) {
+                    return async.get(timeout.getDuration(), timeout.getTimeUnit());
+                } else {
+                    return async.get();
+                }
+            } catch (TimeoutException ex) {
+                if (timer != null) {
+                    timer.recordTimeout(event, startTime);
+                    throw new AsyncToSyncTimeoutException(ex.getMessage(), ex, LogMessageKeys.TIME_LIMIT.toString(), timeout.getDuration(), LogMessageKeys.TIME_UNIT.toString(), timeout.getTimeUnit());
+                }
+                throw new AsyncToSyncTimeoutException(ex.getMessage(), ex);
+            } catch (ExecutionException ex) {
+                throw FDBExceptions.wrapException(ex);
+            } catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
+                // TODO: The original mapping for InterruptedException would be INTERNAL_ERROR, now it will be RecordCoreInterruptedException
+                throw FDBExceptions.wrapException(ex);
+            } finally {
+                if (timer != null) {
+                    timer.recordSinceNanoTime(event, startTime);
+                }
+            }
+        }
+    }
+
+    private LuceneConcurrency() {
+    }
+
+    /**
+     * A representation of a timeout value: A duration and a time unit.
+     */
+    public static class Timeout {
+        private final long duration;
+        @Nonnull
+        private final TimeUnit timeUnit;
+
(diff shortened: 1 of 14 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
```diff
@@ -0,0 +1,162 @@
...
+            try {
+                if (timeout != null) {
+                    return async.get(timeout.getDuration(), timeout.getTimeUnit());
+                } else {
+                    return async.get();
+                }
+            } catch (TimeoutException ex) {
+                if (timer != null) {
+                    timer.recordTimeout(event, startTime);
+                    throw new AsyncToSyncTimeoutException(ex.getMessage(), ex, LogMessageKeys.TIME_LIMIT.toString(), timeout.getDuration(), LogMessageKeys.TIME_UNIT.toString(), timeout.getTimeUnit());
```
ScottDugas, line 101: Why create a new `AsyncToSyncTimeoutException` instead of throwing `LoggableTimeoutException`?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
```diff
@@ -0,0 +1,162 @@
...
+                }
+                throw new AsyncToSyncTimeoutException(ex.getMessage(), ex);
```
ScottDugas, line 103: I see that the `FDBDatabase.asyncToSync` does this too, but why does this only add the extra details if the timer is non-null?
The original PR (https://github.com/FoundationDB/fdb-record-layer/pull/664/files#diff-a2b46d3c7810aff21a344c3904a14d949eda54d0be86d34915f26ea01ce612c6R916) has no context

Either way, it seems best to minimize the changes here, and we can change that at a later time.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
```diff
@@ -0,0 +1,162 @@
...
+        private final TimeUnit timeUnit;
+
+        public Timeout(final long duration, @Nonnull final TimeUnit timeUnit) {
+            this.duration = duration;
+            this.timeUnit = timeUnit;
+        }
+
+        public static Timeout of(final long duration, @Nonnull final TimeUnit timeUnit) {
+            return new Timeout(duration, timeUnit);
+        }
```
ScottDugas, line 135: Why have a public constructor _and_ factory method that are identical

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ScottDugas, line 101: Nit: For readability to ensure that there is a correct set of pairs, it would be good to wrap this line
```suggestion
                    throw new AsyncToSyncTimeoutException(ex.getMessage(), ex,
                            LogMessageKeys.TIME_LIMIT.toString(), timeout.getDuration(),
                            LogMessageKeys.TIME_UNIT.toString(), timeout.getTimeUnit());
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.apple.foundationdb.record.lucene.LuceneRecordContextProperties.LUCENE_USE_LEGACY_ASYNC_TO_SYNC;
```
ScottDugas, line 41: I think it would be better to not statically import this. It's only used in one place, and it would be clearer as `LuceneRecordContextProperties.LUCENE_USE_LEGACY_ASYNC_TO_SYNC`

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -373,7 +378,7 @@ public <R> CompletableFuture<R> applyInRecoveryPath(Function<FDBRecordContext, C
                 future = function.apply(recoveryContext)
                         .whenComplete((result, ex) -> {
                             if (ex == null) {
-                                recoveryContext.commit();
+                                commit(recoveryContext);
```
ScottDugas, line 381: After this PR, at some point....
This was in existing code, but this probably shouldn't be calling `asyncToSync`, and instead should be doing
```
function.apply(recoveryContext)
    .thenCompose(recoveryContext.commitAsync)
    .whenComplete(() -> recoveryContext.close());
```

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
```diff
@@ -178,6 +179,10 @@ default <T> T getPropertyValue(@Nonnull RecordLayerPropertyKey<T> propertyKey) {
      */
     void setCommitCheck(Function<FDBRecordContext, CompletableFuture<Void>> commitCheck);
 
+    default void commit(@Nonnull FDBRecordContext context) {
+        LuceneConcurrency.asyncToSync(FDBStoreTimer.Waits.WAIT_COMMIT, context.commitAsync(), context);
+    }
+
```
ScottDugas, line 185: After this PR, at some point....
I _feel_ like instead of calling `asyncToSync` `AgilityContext` should just have asynchronous methods, and callers can use `asyncToSync`.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -405,7 +406,7 @@ void flakyMerge(boolean isGrouped,
                     assertFalse(requireFailure && i < 15, i + " merge should have failed");
                     success = true;
                 } catch (RecordCoreException e) {
-                    final LoggableTimeoutException timeoutException = findTimeoutException(e);
+                    final LoggableKeysAndValues<? extends Exception> timeoutException = findTimeoutException(e);
```
ScottDugas, line 409: If the new `asyncToSync` is changed to throw the same exception, this should be changed back.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -959,15 +959,18 @@ private void createComplexRecords(int count,
...
 
-    private static LoggableTimeoutException findTimeoutException(final RecordCoreException e) {
+    private static LoggableKeysAndValues<? extends Exception> findTimeoutException(final RecordCoreException e) {
         Map<Throwable, String> visited = new IdentityHashMap<>();
         ArrayDeque<Throwable> toVisit = new ArrayDeque<>();
         toVisit.push(e);
         while (!toVisit.isEmpty()) {
             Throwable cause = toVisit.removeFirst();
             if (!visited.containsKey(cause)) {
+                // TODO: This should eventually be removed since LoggableTimeoutException would not be thrown
```
ScottDugas, line 969: Why not remove now?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -405,7 +406,7 @@ void flakyMerge(boolean isGrouped,
                     assertFalse(requireFailure && i < 15, i + " merge should have failed");
```
ScottDugas, line 406: Have you run this and `LuceneIndexTest` with `LUCENE_USE_LEGACY_ASYNC_TO_SYNC` defaulting to `false`, so that all the code uses the new path?

ScottDugas (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: `LoggableTimeoutException` is a Checked Exception, and I needed a subclass of `RecordCoreException` to throw from the `asyncToSync`. The original code in `FDBException.wrapExceptions` would have wrapped this in a `RecordCoreException`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: Yeah, that was the thinking on my end too. I'll collect the followups for future implementation.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/AgilityContext.java
ohadzeliger: Added to future issues.
Together with `LucenePartitioner`:
```
public PartitionedQueryHint selectQueryPartition(@Nonnull Tuple groupKey, @Nullable LuceneScanQuery luceneScanQuery) {
        return LuceneConcurrency.asyncToSync(WAIT_LOAD_LUCENE_PARTITION_METADATA, selectQueryPartitionAsync(groupKey, luceneScanQuery), state.context);
}
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ohadzeliger: This would still be thrown when the property is set to use the legacy asyncToSync. Changed the comment.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ohadzeliger: Initially planned to add tests with the second PR, but this is a good idea. Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private LuceneConcurrency() {
+    }
+
+    /**
+     * A representation of a timeout value: A duration and a time unit.
+     */
+    public static class Timeout {
```
MMcM, line 123: I am not exactly opposed to the creation of this new class, but I see that the only(?) usage constructs one from separate values retrieved via an API that returns `Pair` and then deconstructs it just a few lines later.

Would it make more sense to push this into `getAsyncToSyncTimeout` in place of the `Pair`? I guess that's an incompatible change.

Would it make sense to also include the `Event`, so that it's more scoped to the abstracts immediately here?

MMcM (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: The original idea behind this class was the ability to add a version of `asyncToSync` that could take the timeout value without consulting the `FDBDatabase`, but that did not pan out because it is incompatible with existing tests and would not end up being useful. Then I left the class in place as it offered some indication of the constituents of the timeout (as opposed to `Pair.getRight` and `Pair.getLeft`), but if that is causing confusion, I will remove it.

ScottDugas (approved):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneConcurrency.java
ohadzeliger: Done (Removed `Timeout` class)

MMcM (approved):

ScottDugas (approved):

PR #2928, opened 2024-09-24
Resolve #2927: Avoid occasional "runner has been closed" error during indexing

### LINKED ISSUE #2927: Avoid occasional "runner has been closed" error during indexing
As it seems: 
Synchronized runner is created by runner.startSynchronizedSessionAsync, while the calling runner sets itself as the underlying runner (which is the main runner for the indexing session). 
 Let it be  mainRunner and SyncRunner (where the latest points to `mainRunner).
runWithSynchronizedRunnerAndEndSession is called with the  syncRunner, saves it in indexingCommon, the uses MoreAsyncUtil.composeWhenComplete to wait for the code to finish, then remove the syncRunner from indexingCommon. 
When the indexer is closed, mainRunner is closed and if a syncRunner is not null, it closes too. 
Typically the indexer is closed only after MoreAsyncUtil.composeWhenComplete gets to clear the syncRunner from common. But sometimes (seen it in exception path), the indexer is closed first and then: 
Closes mainRunner
Closes syncRunner, which attempts to close the underlying runner, which is the mainRunner that cannot be re-closed.

jjezra:
Synchronized runner is created by `runner.startSynchronizedSessionAsync`, with the calling runner setting itself as the underlying runner (which is the main runner for the indexing session).
 Let the two runners be `mainRunner` and `SyncRunner` (where the latest points to `mainRunner`).

`runWithSynchronizedRunnerAndEndSession` is called with the `syncRunner`, and saves it in `indexingCommon`, then uses `MoreAsyncUtil.composeWhenComplete` to wait for the indexing to finish, then removes the `syncRunner` from `indexingCommon`.
When the indexer is closed, `mainRunner` is closed and if the `syncRunner` is not null, it closes too.
Typically the indexer is closed only after `MoreAsyncUtil.composeWhenComplete` removes the `syncRunner` from common. But sometimes (as seen in exception path), the indexer is closed first and then:
  Closes the `mainRunner`
  Closes the `syncRunner`, which attempts to close the underlying runner, which is the `mainRunner`, which cannot be re-closed.
  
 The quick solution would be simply to avoid closing the `syncRunner`. In the longer term, it might be a good idea to use another mechanism for online indexing.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
@@ -287,8 +287,5 @@ public class IndexingCommon {
 
     public void close() {
         runner.close();
-        if (synchronizedSessionRunner != null) {
-            synchronizedSessionRunner.close();
-        }
     }
 }
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -21,7 +21,7 @@ Starting with version [3.4.455.0](#344550), the semantics of `UnnestedRecordType
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
```diff
@@ -287,8 +287,5 @@ public boolean loadConfig() {
 
     public void close() {
         runner.close();
-        if (synchronizedSessionRunner != null) {
```
ScottDugas, line 290: Another option would be to _only_ close the `synchronizedSessionRunner` if it is non-null, and only close the `runner` if `synchronizedSessionRunner` is null.

jjezra: Since we do not control the timing of the other thread, checking if the `syncRunner` is null may cause some weird race conditions. So I thought that it would be safer to only close to `mainRunner`, and only in this path. 
Although it it does is to close the underlying runner, it does feel weird that the `syncRunner` is never being closed. I hope to replace this mechanism ASAP.

PR #2933, opened 2024-09-30
resolved #2940: do not plan redundant/useless intersections

### LINKED ISSUE #2940: avoid creating useless/redundant intersections when planning data access
#### index definitions: 
```
index I1: a, id
index I2: b, id
```
and query
```
SELECT * FROM T WHERE a = 5 and b = 'foo'
```
Observation: Intersection of `I1 ∩ I2` is useful producing an intersection order on `id` with `a`, `b` equality-bound

If we add
```
index I3: a, b, id
```
We should consider it useless to intersect `I1`, `I2` or `I1 ∩ I2` with it as all of these subsets already constrain `a` and `b` to be equally bound.
 
We should not create those and similar intersections.

normen662:
Do not plan intersections during data access that do not have a chance of contributing an improving factor:
1. Do not plan an intersection of n accesses if the intersection's equality-bound keys and ordering are the same as any intersection of (n-1) accesses, that is the smaller intersection has the same cardinality and order as the current intersection that is being planned.
2. Do not plan an intersection if there is some other intersection of a true superset of accesses that is not redundant as to point 1.
3. Do not plan an intersection if one of the accesses participating in the intersection has a max cardinality of `0` or `1`.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
@@ -66,15 +66,19 @@ public class RequestedOrdering {
      */
     @Nonnull
     private final List<RequestedOrderingPart> orderingParts;
-
+    @Nonnull
     private final Distinctness distinctness;
+    private final boolean isExhaustive;
 
     @Nonnull
     private final Supplier<Map<Value, RequestedSortOrder>> valueRequestedSortOrderMapSupplier;
 
-    private RequestedOrdering(@Nonnull final List<RequestedOrderingPart> orderingParts, final Distinctness distinctness) {
+    private RequestedOrdering(@Nonnull final List<RequestedOrderingPart> orderingParts,
+                              @Nonnull final Distinctness distinctness,
+                              @Nonnull final boolean isExhaustive) {
         this.orderingParts = ImmutableList.copyOf(orderingParts);
         this.distinctness = distinctness;
+        this.isExhaustive = isExhaustive;
         this.valueRequestedSortOrderMapSupplier = Suppliers.memoize(this::computeValueSortOrderMap);
     }
 
@@ -84,6 +88,10 @@ public class RequestedOrdering {
 
     public boolean isDistinct() {
         return distinctness == Distinctness.DISTINCT;
+    }
+
+    public boolean isExhaustive() {
+        return isExhaustive;
     }
 
     /**
@@ -198,7 +206,7 @@ public class RequestedOrdering {
             final var rebasedOrderingValue = orderingValue.rebase(translationMap);
             pushedDownOrderingPartsBuilder.add(new RequestedOrderingPart(rebasedOrderingValue, orderingPart.getSortOrder()));
         }
-        return new RequestedOrdering(pushedDownOrderingPartsBuilder.build(), Distinctness.PRESERVE_DISTINCTNESS);
+        return new RequestedOrdering(pushedDownOrderingPartsBuilder.build(), Distinctness.PRESERVE_DISTINCTNESS, isExhaustive());
     }
 
     @Nonnull
@@ -219,7 +227,15 @@ public class RequestedOrdering {
         if (this.distinctness == distinctness) {
             return this;
         }
-        return new RequestedOrdering(orderingParts, distinctness);
+        return new RequestedOrdering(getOrderingParts(), distinctness, isExhaustive());
+    }
+
+    @Nonnull
+    public RequestedOrdering exhaustive() {
+        if (this.isExhaustive()) {
+            return this;
+        }
+        return new RequestedOrdering(getOrderingParts(), getDistinctness(), true);
     }
 
     /**
@@ -228,12 +244,13 @@ public class RequestedOrdering {
      */
     @Nonnull
     public static RequestedOrdering preserve() {
-        return new RequestedOrdering(ImmutableList.of(), Distinctness.PRESERVE_DISTINCTNESS);
+        return new RequestedOrdering(ImmutableList.of(), Distinctness.PRESERVE_DISTINCTNESS, false);
     }
 
     @Nonnull
     public static RequestedOrdering ofParts(@Nonnull final List<RequestedOrderingPart> requestedOrderingParts,
                                             @Nonnull final Distinctness distinctness,
+                                            final boolean isExhaustive,
                                             @Nonnull final Set<CorrelationIdentifier> constantAliases) {
         final var primitiveRequestedOrderingParts = ImmutableList.<RequestedOrderingPart>builder();
         for (final var requestedOrderingPart : requestedOrderingParts) {
@@ -248,17 +265,18 @@ public class RequestedOrdering {
                 primitiveRequestedOrderingParts.add(simplifiedRequestedOrderingPart);
             }
         }
-        return ofPrimitiveParts(primitiveRequestedOrderingParts.build(), distinctness);
+        return ofPrimitiveParts(primitiveRequestedOrderingParts.build(), distinctness, isExhaustive);
     }
 
     @Nonnull
     public static RequestedOrdering ofPrimitiveParts(@Nonnull final List<RequestedOrderingPart> requestedOrderingParts,
-                                                     @Nonnull final Distinctness distinctness) {
+                                                     @Nonnull final Distinctness distinctness,
+                                                     final boolean isExhaustive) {
         Debugger.sanityCheck(() -> Verify.verify(
                 requestedOrderingParts.stream()
                         .allMatch(requestedOrderingPart -> requestedOrderingPart.getValue()
                                 .getResultType().isPrimitive())));
-        return new RequestedOrdering(requestedOrderingParts, distinctness);
+        return new RequestedOrdering(requestedOrderingParts, distinctness, isExhaustive);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrderingConstraint.java
```diff
@@ -34,14 +35,40 @@ public class RequestedOrderingConstraint implements PlannerConstraint<Set<Reques
 
     @Nonnull
     @Override
-    public Optional<Set<RequestedOrdering>> combine(@Nonnull final Set<RequestedOrdering> currentConstraint, @Nonnull final Set<RequestedOrdering> newConstraint) {
-        if (currentConstraint.containsAll(newConstraint)) {
+    public Optional<Set<RequestedOrdering>> combine(@Nonnull final Set<RequestedOrdering> currentConstraint,
```
normen662, line 38: This is now done in a better way as requested orders are hopefully combined in a better way leading to less re-exploration.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementInJoinRule.java
```diff
@@ -140,48 +147,45 @@ public void onMatch(@Nonnull final CascadesRuleCall call) {
             final var providedOrdering = planPartition.getAttributeValue(OrderingProperty.ORDERING);
 
             for (final RequestedOrdering requestedOrdering : requestedOrderings) {
-                final ImmutableList<InSource> sources =
-                        getInSourcesForRequestedOrdering(explodeAliasToQuantifierMap, explodeAliases,
+                final var sourcesStream =
```
normen662, line 150: The changes in this file are related to actually creating all possible orderings if the requested ordering is exhaustive. Orderings can be manipulated by pre-sorting the `InSource` of an `in-join` plan. By creating ascending and descending orders we create new possibilities if the result of this `in-join` partakes in a distinct set operation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementInJoinRule.java
```diff
@@ -216,130 +220,210 @@ private ImmutableList<InSource> getInSourcesForRequestedOrdering(@Nonnull final
...
+            // TODO consider creating the non-counterflow orders as well
+            return ImmutableList.of(ProvidedSortOrder.ASCENDING, ProvidedSortOrder.DESCENDING);
+        } else {
+            return ImmutableList.of(ProvidedSortOrder.ASCENDING);
+        }
+    }
+
+    @Nullable
+    @SuppressWarnings("unchecked")
+    private static InSource computeInSource(@Nonnull final Value explodeValue,
```
normen662, line 401: This method was added since previously we had very similar code in two different places almost being just copy/pasta of one another.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
```diff
@@ -86,6 +90,10 @@ public boolean isDistinct() {
         return distinctness == Distinctness.DISTINCT;
     }
 
+    public boolean isExhaustive() {
```
MMcM, line 93: At least one of these methods or the class itself needs a comment on what `exhaustive` means in this context.

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
normen662: I have it on the attribute declaration, will add that here as well.

normen662: Done.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrderingConstraint.java
```diff
@@ -34,14 +35,40 @@ public class RequestedOrderingConstraint implements PlannerConstraint<Set<Reques
...
+                    continue;
+                }
+                final var newOrderingParts = newRequestedOrdering.getOrderingParts();
+                final var currentOrderingParts = currentRequestedOrdering.getOrderingParts();
+
+                if (!currentRequestedOrdering.isExhaustive() && newRequestedOrdering.isExhaustive()) {
+                    continue;
+                }
+                if (currentRequestedOrdering.isExhaustive() && newOrderingParts.size() >= currentOrderingParts.size()) {
+                    if (newOrderingParts.subList(0, currentOrderingParts.size()).equals(currentOrderingParts)) {
```
hatyo, line 54: It would probably make sense to wrap this in a method that returns `true` if a `RequestedOrdering` prefixes / subsumes another `RequestedOrdering` to make the intention clearer.

By the way, it looks like `combine` is not symmetric, right?
for example:
`combine( (Set.of('a', 'b'), exhaustive=true), (Set.of('a', 'b', 'c'), exhaustive=true)) != combine( (Set.of('a', 'b', 'c'), exhaustive=true), (Set.of('a', 'b'), exhaustive = true))` since the first `combine` would return an empty set, while the second `combine` would return `Set.of('a', 'b'), Set.of('a', 'b', 'c'), exhaustive = true` if I am not mistaken, is this expected?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
hatyo: So this enables the property to signal whether the enforcer is interested in either getting the best plan that satisfies that particular property, or enumerate all candidate plans that satisfy it instead. Correct?

Can we think of this as a parameterized branch-and-bound? in other words, can we think of the `exhaustiveness` attribute as a way to apply some sort of costing in-place? If so, would it make sense to think of a more generalized form of costing instead of carrying it out through this boolean attribute?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -218,13 +223,62 @@ protected Set<? extends RelationalExpression> dataAccessForMatchPartition(@Nonnu
...
+        // Alternative view: Mathematically, we consider the power set of all accesses when we plan the intersections
+        // of accesses. Each item in that set is a partition. There is a partial order PO defined on the power set that
+        // states that two elements a, b: aPOb if the set of accesses in b includes the accesses in a.
+        // Some partitions cannot be used outright as they e.g. would not even range over compatibly-ordered data
+        // streams. These partitions are evicted from the power set. All partitions containing those partitions are
+        // immediately evicted as well. Apart from those partitions, there are other albeit useless/redundant elements
+        // (partitions) in the power; we evict those and the ones containing these useless partitions in the same
+        // fashion (as they are also useless/redundant).
+        // What we end up with is a subset of the power set of all accesses that are possible to intersect,
+        // that are not considered useless/redundant and that are local maximums with respect to PO.
```
hatyo, line 269: I think what you're describing is very similar to a Hasse diagram of power set, if we think of this hypercube as a directed graph where direction indicates set containment, then the algorithm works by identifying a node that can be deleted from the graph, and iteratively removing all of the nodes along all paths originating from that node.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
normen662: I'd say the exhaustiveness is part of the requirement, it's not a modifier of the requirement. As much as the ordering requirement is already facilitating branch-and-bound (which is currently exhaustive), this is now doing away with that requirement if we can and just requesting the best plan.

There are other things you may want to request (wrt to orderings) that will be more complicated than exhaustiveness. For instance, for joins with a requested ordering of `a, b, c, d` you should create plans on the outer with strict orders for `a`, `a, b`, `a, b, c`, and a regular (non-strict) order of `a, b, c, d`. That's a lot of requested orders. Alternatively you could also say, give me all orders where the order is a prefix of `a, b, c, d` and strict. My point here is that the requested requirement (requested ordering here) is ours to formulate and depends very much on the operators that are in play, etc. This will become much more elaborate over time to capture all the needs we will have from the sub graph.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrderingConstraint.java
normen662: Yes, it is expected but I think it is symmetric. If you already have a request for exhaustively producing all possible orders for `a, b`, you don't have to honor anything with a prefix of `a, b, c`. If on the other hand you already have `a, b, c` and then get `a, b` you just continue with `a, b`. In the first case you return empty since you don't want the set of required orders to change (but mind you `a, b` is already in there) while in the second case you want to promote the set of requested orders to something more encompassing: you go from `a, b, c` to `a, b`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: That's absolutely correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrderingConstraint.java
normen662: New issue to be opened.

PR #2937, opened 2024-10-04
Fixes #2936: Missing subspace provider in FDBRecordStore logs

### LINKED ISSUE #2936: Missing subspace provider in FDBRecordStore logs
There are a couple of places in `FDBRecordStore` where we're logging information about something that happened in the store, but we're not including the subspace provider so we're missing out on which record store the log is about. For example, the "index state change" message, indicating an index was either enabled or disabled, is missing those log keys.

alecgrieser:
There were a handful of places in the `FDBRecordStore` class that were were logging but didn't include the subspace provider as context. This means that we couldn't tell what store the logs were trying to log about, which made it hard to correlate with other operations to the store. This resolves that by adding in the missing information to the missing logs.

This fixes #2936.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -1985,8 +1985,10 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         private CompletableFuture<Void> run() {
             if (evaluated == null) {
                 // no record types
-                LOGGER.warn(KeyValueLogMessage.of("Tried to delete prefix with no record types",
-                        subspaceProvider.logKey(), subspaceProvider.toString(context)));
+                if (LOGGER.isWarnEnabled()) {
+                    LOGGER.warn(KeyValueLogMessage.of("Tried to delete prefix with no record types",
+                            subspaceProvider.logKey(), subspaceProvider.toString(context)));
+                }
                 return AsyncUtil.DONE;
             }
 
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4226,7 +4231,6 @@ public CompletableFuture<Void> rebuildIndex(@Nonnull final Index index, @Nonnull
                                 LogMessageKeys.INDEX_NAME, index.getName(),
                                 LogMessageKeys.INDEX_VERSION, index.getLastModifiedVersion(),
                                 LogMessageKeys.REASON, reason.name(),
-                                subspaceProvider.logKey(), subspaceProvider.toString(context),
```
alecgrieser, line 4229: This got moved into `logExceptionAsWarn`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -1128,9 +1128,14 @@ default RecordCursor<FDBIndexedRecord<M>> scanIndexRecords(@Nonnull final Index
                 } catch (UnsupportedRemoteFetchIndexException ex) {
                     // In this case (e.g. the index maintainer does not support remote fetch), log as info
                     if (LOGGER.isInfoEnabled()) {
-                        LOGGER.info(KeyValueLogMessage.of("scanIndexRecords: Remote fetch unsupported, continuing with Index scan",
+                        KeyValueLogMessage msg = KeyValueLogMessage.build("scanIndexRecords: Remote fetch unsupported, continuing with Index scan",
                                 LogMessageKeys.MESSAGE, ex.getMessage(),
-                                LogMessageKeys.INDEX_NAME, index.getName()));
+                                LogMessageKeys.INDEX_NAME, index.getName());
+                        final SubspaceProvider subspaceProvider = getSubspaceProvider();
+                        if (subspaceProvider != null) {
```
alecgrieser, line 1135: `getSubspaceProvider` is `@Nullable` on the `FDBRecordStoreBase` interface. However, none of the implementations return `null`, so it seems like something we could change. If we did, we could simplify this to avoid the null check.

MMcM (approved):

PR #2942, opened 2024-10-10
Fixes #2941: In-union plans don't consider sargability of in-value predicates

### LINKED ISSUE #2941: In-union plans don't consider sargability of in-value predicates
The old planner has logic to enable it to only use an in-join plan if the list in the in is pushed into an index predicate. However, the logic for in-unions goes around that re-planning work, so we can end up with in-union plans which end up performing the same scans over-and-over again needlessly. For example, a query like:

```java
RecordQuery.newBuilder()
   .setRecordType("myType")
   .setFilter(Query.and(
       Query.field("x").in(xList),
       Query.field("y").in(yList)
   ))
   .setSort(field("z"))
   .build();
```

If there's a value index on `concatenateFields("x", "z")`, then this might get planned as something like:

```
∪(__in_x__0 IN $xList, __in_y__1 IN $yList) Index(xzIndex [EQUALS $__in_x__0]) | [y EQUALS $__in_y__1]
```

Note that both the `xList` and the `yList` are providing values to the in-union, but only the `xList` value is being used to influence the index scan, with the the `categoryList` value being used in a residual filter. That means that for every element of `xList`, we execute the same scan over the index and pick out a different `xList` element, which is inefficient. We should instead be able to generate a plan like:

```
∪(__in_x__0 IN $xList) Index(xzIndex [EQUALS $__in_x__0]) | [y IN $yList]
```

Now every leg of the union is over a different value from `xList`, and all values passing the `yList` predicate are returned from each leg. This is already solved, albeit incompletely, for in-joins by checking if the extracted in values are used in sargable predicates, and then replanning without including those non-sargable in clauses. We should be able to do the same thing here.

alecgrieser:
This updates the in-union planning in the old planner so that it behaves like in-join planning. That is to say, it attempts to use a base plan that only has extracted IN-predicates that are pushed into a scan. Non-sargable predicates are passed as is to the filter plan. For backwards compatibility reasons, this is hidden behind a planner configuration property.

This fixes https://github.com/FoundationDB/fdb-record-layer/issues/2941.


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -2553,4 +2558,123 @@ void testInQueryWithNonSargableReplan(final boolean shouldReplan) throws Excepti
...
+        RecordQuery query = RecordQuery.newBuilder()
+                .setRecordType("MySimpleRecord")
+                .setFilter(Query.and(inFilter1, inFilter2))
+                .setSort(field("num_value_3_indexed"))
+                .build();
+
+        RecordQueryPlan plan = planQuery(query);
+
+        if (useCascadesPlanner) {
+            // This is a suboptimal plan!
```
alecgrieser, line 2631: Note that the Cascades planner doesn't quite get this right. (There are a few other plans in this fixture that have similar issues.) I left them as is for now, but we could tackle it. I think the issue is that the cost model looks at the number of non-sargable in comparisons when evaluating plans, but it can be tricked if the offending plan is nested.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
```diff
@@ -1164,6 +1171,7 @@ RecordQueryPlannerConfiguration.Builder configure(RecordQueryPlannerConfiguratio
                 case AS_UNION:
                     config.setAttemptFailedInJoinAsOr(true);
                     config.setAttemptFailedInJoinAsUnionMaxSize(1000);
+                    // config.setMaxNumReplansForInUnion(1);
```
normen662, line 1174: Please remove!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
normen662: That is unfortunately correct. Can you open an issue please?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -547,10 +547,22 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+        @Nonnull
+        private final InExtractor inExtractor;
+
+        PlanWithInExtractor(@Nonnull ScoredPlan plan, @Nonnull InExtractor inExtractor) {
+            this.plan = plan;
+            this.inExtractor = inExtractor;
+        }
+    }
+
+    private PlanWithInExtractor planFilterForInWithReplans(@Nonnull PlanContext planContext, @Nonnull InExtractor inExtractor, boolean needOrdering, int maxNumReplansInConfig) {
```
normen662, line 562: Shouldn't that be `planFilterWithInWithReplans` :-) according to the existing naming scheme?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -547,10 +547,22 @@ private ScoredPlan planFilter(@Nonnull PlanContext planContext, @Nonnull QueryCo
...
+        int maxNumReplans = Math.max(maxNumReplansInConfig, 0);
+        boolean allowNonSargedInBindings = maxNumReplansInConfig < 0;
+
```
normen662, line 565: Similarly, can we rename `planFilterForInJoin` to something like `planFilterForIn` or something more fundamental like `planBaseAccessPath` as that method is the lowest of the filter descenders IIRC.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
```diff
@@ -598,11 +610,21 @@ private ScoredPlan planFilterWithInJoin(@Nonnull PlanContext planContext, @Nonnu
         }
 
         Verify.verifyNotNull(bestPlan);
-        
-        final RecordQueryPlan wrapped = inExtractor.wrap(planContext.rankComparisons.wrap(bestPlan.getPlan(), bestPlan.includedRankComparisons, metaData));
-        final ScoredPlan scoredPlan = new ScoredPlan(bestPlan.score, wrapped);
+        return new PlanWithInExtractor(bestPlan, inExtractor);
```
normen662, line 613: So `bestPlan` is in fact the best plan _underneath_ the future in plan (inJoin or inUnion) together with the `inExtractor` that was used to get to the best plan. Is the right?

normen662: Maybe give this method a javadoc to reflect that.

normen662 (changes requested):
It's great that the change is not super complicated. I think all is structured well and can go in after some polishing. I trust the tests are sufficient (I only glanced at them). Can you please open tickets whenever you feel that Cascades screws up a particular query?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: I went with `planExtractedInsFilter` because I think the most relevant thing for the user is to note that the `filter` being planned is different from the original query filter. It's not quite `planBaseAccessPath` because it applies some compensation on top of base scans (e.g., it creates intersections and it applies filters)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlanner.java
alecgrieser: Yeah, that's right. I've added Javadoc on these two methods to hopefully make that clearer.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
alecgrieser: Done

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBInQueryTest.java
alecgrieser: Filed #2950

normen662 (dismissed):

normen662 (approved):

PR #2944, opened 2024-10-13
Resolves #1887: Split synthetic record planner test

### LINKED ISSUE #1887: Split out SyntheticRecordPlannerTest into multiple classes
The `SyntheticRecordPlannerTest` has grown a bit large. We should consider refactoring this class into multiple smaller test fixtures.

gsegatti:
Under #1887 it is proposed we refactor the `SyntheticRecordPlannerTest` into smaller bits. I've attempted such  under this PR, by:

- Creating a `AbstractSyntheticRecordPlannerTest` holding common methods and class attributes over the current set of tests, which is then extended across the newer and smaller text classes.
- Grouping common-ish tests under smaller classes. Now the `oneToOne()`, `manyToOne()`, [...], tests are grouped under `SyntheticRecordPlannerJoinsTest`. `joinIndex()` and `buildJoinIndex()` under `SyntheticRecordPlannerJoinIndexTest` and so forth.

That's it, no new tests added nor removed, simply moved functions to different files. This is my first PR here, so there may still be a few things misplaced. Please feel free to point them out if you notice anything!


MMcM (approved):
Confirmed that this just moves methods around and ends up with something reasonably sized.

PR #2945, opened 2024-10-14
Fixes #548: Fix concurrency issues with MapPipelinedCursor.close

### LINKED ISSUE #548: testOrQuery5WithLimits intermittently fails with a NoSuchElementException

ScottDugas:
The underlying issue is that multiple threads are interacting with the (non-thread-safe) queue at the same time, in complex ways. We could add `synchronized` to a `close` and `onNext`, (or some part of onNext), but instead I changed the cleanup around so that it only interacts with the pipeline in a single chain of futures.
The other change is the introduction of the `nextFuture` field, which removes the code that assumes that the pipeline will be non-empty, this seems like a good change to have, even if the other change makes it unnecessary.

I tried to do some trickery around how it interacts with the queue to make sure it handles an empty queue correctly, but since `ArrayDeque` is not thread safe, that resulted in `close` getting into an infinite loop with an-emptyable queue. I left the first attempt there as the first commit in this PR, although I think none of the code change remains.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java
@@ -32,6 +32,7 @@ import javax.annotation.Nullable;
 import java.util.ArrayDeque;
 import java.util.Iterator;
 import java.util.Queue;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Function;
@@ -56,6 +57,7 @@ public class MapPipelinedCursor<T, V> implements RecordCursor<V> {
     @Nonnull
     private final Queue<CompletableFuture<RecordCursorResult<V>>> pipeline;
     private boolean innerExhausted = false;
+    private volatile boolean closing = false;
 
     @Nullable
     private CompletableFuture<RecordCursorResult<T>> waitInnerFuture = null;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1268,4 +1278,58 @@ void asStreamTest() {
 
     }
 
+    @RepeatedTest(100)
```
ScottDugas, line 1281: Opening the draft PR with this repetition, but I'm going to remove it before merging, as it seems excessive.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java
```diff
@@ -90,15 +93,17 @@ public CompletableFuture<RecordCursorResult<V>> onNext() {
 
     @Override
     public void close() {
-        while (!pipeline.isEmpty()) {
```
alecgrieser, line 93: Should we do the same thing to the `FlatMapPipelinedCursor`, which I believe can hit similar issues?

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java
alecgrieser: I filed a new issue about this: #2953

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/MapPipelinedCursor.java
```diff
@@ -77,8 +81,7 @@ public CompletableFuture<RecordCursorResult<V>> onNext() {
             return CompletableFuture.completedFuture(nextResult);
         }
         return AsyncUtil.whileTrue(this::tryToFillPipeline, getExecutor())
-                // pipeline will necessarily contain something if we stopped looping, so pipeline.remove() is nonnull
-                .thenCompose(vignore -> pipeline.peek()) // future should already be (nearly) ready if we stopped looping
+                .thenCompose(vignore -> pipeline.peek())
                 .thenApply(result -> {
                     if (result.hasNext()) {
```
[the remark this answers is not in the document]
ScottDugas: Added back a comment: https://github.com/FoundationDB/fdb-record-layer/pull/2955
I believe it was removed when I introduced `nextFuture`, but that didn't work, so I switched back to interacting with `pipeline`, and didn't add a comment back. I tried to thoroughly cover why the pipeline will not be empty at either of those situations.

PR #2952, opened 2024-10-16
Resolve #2951: Add Lucene lock duration to store timer

### LINKED ISSUE #2951: Add Lucene lock duration
In order to assess the impact of Lucene lock on other operations, we would need the lock duration to be added to the store timer. This would simplify the process of obtaining the overall time of the lock as we won't need to correlate the creation/release time by lock ID.

ohadzeliger:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryLockFactory.java
```diff
@@ -232,6 +235,7 @@ private void fileLockClearFlushAndClose(boolean isRecovery) {
                 agilityContext.asyncToSync(LuceneEvents.Waits.WAIT_LUCENE_FILE_LOCK_CLEAR,
                         agilityContext.apply(fileLockFunc));
             }
+            agilityContext.recordEvent(LuceneEvents.Events.LUCENE_FILE_LOCK_DURATION, System.nanoTime() - lockStartTime);
```
ScottDugas, line 238: If we fail to `flush` and close in a recovery path this would get emitted twice. Perhaps we should only do this if `agilityContext.flush()` succeeds?

ohadzeliger: Discussed with @jjezra and it looks like if we recorded the event once, the `close` state will change to `TRUE` and the next call to `fileLockClearIfLocked` will be noop. We will only enter `fileLockClearFlushAndClose` again in case the `close` failed (and the event was not recorded).

ScottDugas (approved):

PR #2954, opened 2024-10-18
Fix #2953: FlatMapPipelinedCursor can fail with NoSuchElementException or NullPointerException if concurrently closed

### LINKED ISSUE #2953: FlatMapPipelinedCursor can fail with NoSuchElementException or NullPointerException if concurrently closed
This is an analogue to #548, but for `FlatMapPipelinedCursors`. We don't have an illustrative flaky test like we did for that issue, but in much the same way as with #548, if someone is pulling from a cursor and then calls `close` in another thread (which can happen during certain pipelined operations), then they can end up getting one of these internal errors.

There is a slight nuance here that is easier swept under the rug with the `MapPipelinedCursor`. Namely, that first cursor doesn't have work it needs to clean up in quite the same way (there can be futures in the pipeline, and it tries to call `cancel` on them, but cancelling a future does very little in reality). With a `FlatMapPipelinedCursor`, because there are inner cursors hanging around, it's a bit more important that we actually close all the outstanding work. That might mean we need to have a `synchronized` block around closing/accessing the pipeline, or something along those lines.

These two classes (`MapPipelinedCursor` and `FlatMapPipelinedCursor`) are pretty similar but _just_ different enough that they share essentially no code other than the default `RecordCursor` methods.

alecgrieser:
This updates the `FlatMapPipelinedCursor` to control all of its accesses of the (non-thread safe) pipeline via synchronized methods. It furthermore tries
to ensure that any cursors that get created are closed. The use of `synchronized` methods here, unlike in the `MapPipelinedCursro`, is necessitated by the
possibility that the user stops iterating on a cursor and calls `close()`. If nothing calls `.onNext()` on that closed cursor, then nothing will end up closing
the generated cursors in the alternative architecture.

This fixes https://github.com/FoundationDB/fdb-record-layer/issues/2953.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FlatMapPipelinedCursor.java
```diff
@@ -56,6 +57,13 @@
 @API(API.Status.MAINTAINED)
 @SuppressWarnings("PMD.CloseResource")
 public class FlatMapPipelinedCursor<T, V> implements RecordCursor<V> {
+    private static final CompletableFuture<Boolean> CANCELLED;
```
ScottDugas, line 60: We have this in two places now. Perhaps worth creating `MoreAsyncutil.ALREADY_CANCELLED`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1277,28 +1277,46 @@ void asStreamTest() {
...
+    private RecordCursor<Integer> mapPipelinedCursorToClose(int iteration, CompletableFuture<Void> signal) {
+        return RecordCursor.fromList(EXECUTOR, IntStream.range(0, iteration % 199).boxed().collect(Collectors.toList()))
+                .mapPipelined(val -> signal.thenApplyAsync(ignore -> val + 349, EXECUTOR), iteration % 19 + 2);
+    }
+
+    @Nonnull
+    private RecordCursor<String> flatMapPipelinedCursorToClose(int iteration, CompletableFuture<Void> signal) {
+        return RecordCursor.flatMapPipelined(
+                outerContinuation -> RecordCursor.fromList(EXECUTOR, IntStream.range(0, iteration % 199).boxed().collect(Collectors.toList()), outerContinuation),
+                (outerValue, innerContinuation) -> new FutureCursor<>(EXECUTOR, signal.thenApplyAsync(ignore -> String.valueOf(outerValue), EXECUTOR)),
```
ScottDugas, line 1290: Are we missing some coverage here, because this is just a `FutureCursor`, and not a cursor that has multiple elements?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/RecordCursorTest.java
```diff
@@ -1277,28 +1277,46 @@ void asStreamTest() {
...
-                if (e instanceof CompletionException && e.getCause() != null) {
+                if (e instanceof ExecutionException && e.getCause() != null) {
                     errToCount = e.getCause();
                     errToCount.addSuppressed(e);
                 } else {
                     errToCount = e;
                 }
+                if (errToCount instanceof CancellationException) {
+                    continue;
+                }
```
ScottDugas, line 1319: `MapPipelinedCursor`, I think, never throws `CancellationException` in this test.
Did you see that?
The code has paths where it could, so if it is not, that's probably a shortcoming of the test.

ScottDugas (approved):

ScottDugas:
I created a followup in response to my own comments: https://github.com/FoundationDB/fdb-record-layer/pull/2956

PR #2955, opened 2024-10-18
Add back better version of comment

ScottDugas:
As noted in: https://github.com/FoundationDB/fdb-record-layer/pull/2945#discussion_r1805666165 The comment was removed, but explaining why this is safe is still good.
I believe it was removed during a partial solution, and didn't get added back when the code started touching the pipeline again.


ScottDugas:
Comment was improved as part of #2954

PR #2956, opened 2024-10-18
Some followup from PR #2954

ScottDugas:
- [Add MoreAsyncUtil.ALREADY_CANCELLED](https://github.com/FoundationDB/fdb-record-layer/commit/aaf4ebd4796ff08a6795b9fc87f749a2e1fc49cd)
MapPipelineCursor and FlatMapPipelinedCursor had a version of this,
so reducing duplication.
- [Test concurrent closing with a non-singleton inner cursor](https://github.com/FoundationDB/fdb-record-layer/commit/d489748fcf00bfcf7352620f0a372a1688d89437)

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
@@ -59,6 +59,8 @@ import static com.apple.foundationdb.async.AsyncUtil.whileTrue;
 @API(API.Status.UNSTABLE)
 public class MoreAsyncUtil {
 
+    public static final CompletableFuture<Boolean> ALREADY_CANCELLED;
+
     private static final Supplier<ScheduledThreadPoolExecutor> scheduledExecutorSupplier = Suppliers.memoize(() -> {
         ThreadFactory threadFactory = new ThreadFactoryBuilder()
                 .setDaemon(true)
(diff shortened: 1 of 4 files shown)


fdb-extensions/src/main/java/com/apple/foundationdb/async/MoreAsyncUtil.java
```diff
@@ -59,9 +59,14 @@ public class MoreAsyncUtil {
     private static ScheduledThreadPoolExecutor scheduledThreadPoolExecutor
             = new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().setDaemon(true).build());
 
+    public static final CompletableFuture<Boolean> ALREADY_CANCELLED;
```
alecgrieser, line 62: I think the thing that feels weird about this (when it's in the context of `MoreAsyncUtil`) is that you could imagine wanting to have different generic values of future, sort of like `Optional.empty()`. In theory, you can do it by making `ALREADY_CANCELLED` `private` and of type `CompletableFuture<?>`, and then introducing a method to cast it to the right generic type:

```java
    public static <T> CompletableFuture<T> alreadyCancelled() {
        return (CompletableFuture<T>)ALREADY_CANCELLED;
    }
```

It does rely on type erasure and also the fact that we know that the cancelled future will never actually return a `T`, but it does make it a bit more versatile.

ScottDugas: That is fair. I don't like the idea of using type erasure to do this.
Instead I created that method, and then `FlatMapPipelinedCursor` and `MapPipelinedCursor` put the `Boolean` version in a constant.

alecgrieser (approved):

PR #2957, opened 2024-10-18
Upgrade junit to 5.11.2

ScottDugas:


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/highlight/LuceneHighlighterTest.java
```diff
@@ -113,36 +113,51 @@ public static Stream<Arguments> analyzers() throws IOException {
 
         UAX29URLEmailAnalyzer noStopWords = new UAX29URLEmailAnalyzer(CharArraySet.EMPTY_SET);
         noStopWords.setMaxTokenLength(30);
-        final Analyzer standardAnalyzer = LuceneAnalyzerWrapper.getStandardAnalyzerWrapper().getAnalyzer();
```
ScottDugas, line 116: There are two commits to cover this change (maybe easier to review).
With junit 11:
1. New autoCloseArguments attribute in @ParameterizedTest to close java.lang.AutoCloseable arguments after each invocation of the parameterized test method. This attribute defaults to true.

Since `Analyzer` is  `AutoCloseable`, I need to make sure that it generates new instances for each `Arguments`, including in `SpecialCharacterAnalyzerCombinations.

2. Changing it to use the new `argumentsSet` to provide a name, instead of the extra argument.

arnaud-lacurie (approved):

PR #2958, opened 2024-10-23
WIP - recursive CTE

hatyo:


fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -1724,3 +1724,13 @@ message PRecordQueryUnionPlanBase {
...
   optional PRecordQueryAbstractDataModificationPlan super = 1;
 }
+
+//
+// PTypedQueryResult
+//
+message PQueryResult {
+  oneof datum {
+    PComparableObject primitive = 1;
+    bytes complex = 2;
```
hatyo, line 1734: only the `datum` is serialized. it can be:
- either a primitive type, for that I am reusing `PComparableObject` and its SerDe APIs, instead of more or less duplicating it.
- or a complex type, i.e. a record represented as a `Message`, that is serialized as-is.
Maybe it is better to introduce a different, more elaborate, PB message for the serialization of `datum` that does not reuse `PComparableObject`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryResult.java
```diff
@@ -44,28 +54,31 @@ public class QueryResult {
...
+    //       some extra information that are exclusive to the `queriedRecord` such as the record type and the record
+    //       version, and the size, however this information can be carried out through modelling the plan operators
+    //       differently such that they include any relevant information (such as record version) in the flown result
+    //       (i.e. the `datum`) itself.
     @Nullable
     private final FDBQueriedRecord<?> queriedRecord;
 
-    @Nullable
-    private final IndexEntry indexEntry;
-
```
hatyo, line 52: After analyzing the usages of this field I went ahead and removed it because it was either set to null, or set to the _same_ `IndexEntry` field of the `queriedRecord`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/QueryResult.java
```diff
@@ -44,28 +54,31 @@ public class QueryResult {
...
+    // TODO: this can be removed because it is redundant. The primary key information is encoded
+    //       inside the `datum` object, and can be retrieved by a `Value` that knows how to reconstruct
+    //       the primary key from it correctly.
     @Nullable
     private final Tuple primaryKey;
 
+    // transient field that amortizes the calculation of the serialized form of this immutable object.
     @Nullable
-    private final RecordType recordType;
```
hatyo, line 57: removed for the same reasoning about removing `IndexEntry` field above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/InMemoryQueueCursor.java
```diff
@@ -0,0 +1,207 @@
...
+    public void add(@Nonnull QueryResult... elements) {
+        Arrays.stream(elements).forEach(this::add);
+    }
+
+    @Nonnull
+    @Override
+    public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
+        if (!elements.isEmpty()) {
+            return CompletableFuture.completedFuture(RecordCursorResult.withNextValue(elements.poll(),
+                    new Continuation(ImmutableList.copyOf(elements))));
```
hatyo, line 100: Copying the elements here:
- it makes the `Continuation` state independent of the `Cursor`'s (with each invocation of `onNext` the cursor removes the head of the queue).
- it is more expensive, however it is a "true" snapshot, i.e. if the user adds more elements to the cursor after the continuation is created, these elements will not be part of the continuation, this seems different from the semantics of other continuations that store a mere offset for example, I am not sure if this is ok.

Having said that, should we change the design and make the cursor grow the added items without removing anything from it, this way, the continuation can be a mere index of the current head and we do not have to copy the elements.

MMcM: One of the things that needs to be considered is what to do about changes to the database between getting a continuation and using. it. For example, if the recursion is walking down a tree, what do we do if a row in the ancestor path gets deleted in the meantime? Do we continue to return some of its former descendants?

PR #2969, opened 2024-11-07
Fixes #2968: Sort FieldInfos attributes before serializing

### LINKED ISSUE #2968: Lucene indexes may end up with a bunch of FieldInfos
The goal of `LuceneOptimizedFieldInfosFormat` is to only have one `FieldInfos` on disk, but it appears that for larger record stores it can end up with multiple `FieldInfos`

ScottDugas:
This way when we compare the FieldInfo in the global FieldInfos it will come back the same, even if the Map has a different ordering

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedFieldInfosFormatTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedFieldInfosFormatTest.java
@@ -22,6 +22,8 @@ package com.apple.foundationdb.record.lucene.codec;
 
 
 import com.apple.foundationdb.record.TestHelpers;
+import com.apple.foundationdb.record.lucene.LuceneEvents;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/codec/LuceneOptimizedFieldInfosFormatTest.java
```diff
@@ -241,7 +244,7 @@ void fewerFieldsInNextSegment2(boolean oneTransaction) throws Exception {
     void mixedFieldNumbers(boolean oneTransaction) throws Exception {
         testWithTwoSegments(oneTransaction,
                 singleFieldInfos("foo", 0),
-                singleFieldInfos("foo", 1));
+                singleFieldInfos("foo", 1), 1);
```
ScottDugas, line 247: This is somewhat concerning to me, I feel like this should result in two separate FieldInfos.

ohadzeliger (approved):

PR #2972, opened 2024-11-08
Fixes #2971: Test failure in MoreAsyncUtilTest.getWithDeadlineRunsOnExecutor

### LINKED ISSUE #2971: Test failure in MoreAsyncUtilTest.getWithDeadlineRunsOnExecutor
There was a test failure encountered during nightly tests:

```
MoreAsyncUtilTest > getWithDeadlineRunsOnExecutor() FAILED
    org.opentest4j.AssertionFailedError: expected: <test-deadline-exceeded-thread-0> but was: <Test worker>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//com.apple.foundationdb.async.MoreAsyncUtilTest.getWithDeadlineRunsOnExecutor(MoreAsyncUtilTest.java:136)
```

That test asserts that the right executor is used when running a call back. However, there are conditions where the callback can actually run on the current thread, and that's apparently what was hit during this test run.

alecgrieser:
There was a race condition in a test assertion, and it was actually possible for a test callback to land on the same thread as the test worker if setting up the future chain took longer than the deadline in the test. That should be fairly unlikely, but it could happen, and so this expands the assertion to include that case.

This fixes #2971.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/async/MoreAsyncUtilTest.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/async/MoreAsyncUtilTest.java
@@ -39,6 +39,10 @@ import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.either;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.startsWith;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
@@ -100,7 +104,8 @@ public class MoreAsyncUtilTest {
         String callbackThreadName = MoreAsyncUtil.delayedFuture(5, TimeUnit.MILLISECONDS)
                 .thenApply(ignore -> Thread.currentThread().getName())
                 .get();
-        assertTrue(callbackThreadName.startsWith("fdb-scheduled"), () -> "Callback executed on thread: " + callbackThreadName + ". Should have been executed on delayed executor");
+        assertThat("Callback should have been executed on thread started by default scheduled executor",
+                callbackThreadName, startsWith("fdb-scheduled"));
 
         ScheduledExecutorService scheduledExecutor = new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder()
                 .setDaemon(true)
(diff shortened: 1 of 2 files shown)


fdb-extensions/src/test/java/com/apple/foundationdb/async/MoreAsyncUtilTest.java
```diff
@@ -133,7 +138,12 @@ public void getWithDeadlineRunsOnExecutor() throws ExecutionException, Interrupt
                         return Thread.currentThread().getName();
                     })
                     .get();
-            assertEquals("test-deadline-exceeded-thread-0", callbackThreadName);
+
+            // Most of the time, the callback should come fom the scheduledExecutor. However, if there's some hiccup and actually
+            // setting up the future chain takes longer than the deadline time, then it's possible for the callback to complete on
+            // the test worker thread
+            assertThat("Callback should have been executed on thread managed by scheduled executor or by calling thread",
```
ScottDugas, line 145: This reduces the thoroughness of this test.
Basically, if it reliably took 5ms to setup the future chain (I've seen two test runs fail this way), it would no longer know whether it actually used the provided `scheduledExecutor`.
Would it make sense to increase the deadline to like 15ms.
Would it help to have the `scheduledExecutor` run something before this to make sure the thread pool is fully primed?

alecgrieser: Yeah, I'd thought about how it reduces the thoroughness. I wasn't super concerned about it, because locally, it seems to pretty reliably land on the scheduled executor, so it's not like that isn't getting coverage. I did think about just increasing the deadline, but I didn't know what a reasonable value would be for that. (I thought 5 ms was going to be enough, but alas, that was incorrect.) Part of that is I'm not actually sure what it's waiting on. I don't _think_ it's actually the `scheduledExecutor` in this case, because that future is the one that's completing (too) quickly, and it's some part of `CompletableFuture.anyOf`.

I guess we could make this a repeated test, as it seems unlikely that running this test, say, 10 times, none of them would wind up executing on the scheduled executor.

The main thing I was worried about was that somehow, the `getWithDeadline` would wind up executing its callback on some other scheduled executor pool (like the default one), and this test does give us some coverage of that.

PR #2983, opened 2024-11-19
Resolves #2982: Improvements to "context not closed" logs

### LINKED ISSUE #2982: Improvements to "context not closed" logs
There are some potential improvements that it would be nice to make to our logs for "context not closed" logs to increase usability.

First, if the context has an MDC context, it would be nice to include that data in the log message rather than the MDC in the calling thread. The MDC can contain information that lets us look at the creation of the transaction rather than the closing event, which is relatively less interesting.

Second, we should note what cluster was being used, as there can be times where cluster related issues can lead to at the very least the appearance of an unclosed transaction.

Third, the `age_seconds` log message key contains the amount of time since the transaction was supposed to be closed, but it would probably be more useful to include the time since the transaction was opened.

alecgrieser:
This updates the "context not closed" information to do the following:

* Logs contain the MDC context associated with the transaction, if set. Otherwise, it continues to use the calling thread's MDC.
* The cluster file is included in the information so that adopters can look for patterns related to the calling transaction.
* The "age seconds" is now time since the transaction was opened, not its time since it could have been closed.

This resolves #2982.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
@@ -47,13 +47,13 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
-import java.util.NoSuchElementException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabase.java
```diff
@@ -1193,31 +1193,44 @@ public <T> T get(CompletableFuture<T> future) throws InterruptedException, Execu
...
-            return 0;
-        }
-        try {
-            if (trackedOpenContexts.firstKey() > nanoTime) {
-                return 0;
-            }
-        } catch (NoSuchElementException ex) {
+        long cutoffTime = System.nanoTime() - TimeUnit.SECONDS.toNanos(minAgeSeconds);
+        @Nullable Map.Entry<Long, FDBRecordContext> firstEntry = trackedOpenContexts.firstEntry();
+        if (firstEntry == null || firstEntry.getKey() > cutoffTime) {
```
alecgrieser, line 1198: This requires a bit of explanation: `SortedMap.firstKey` returns the lowest key or throws `NoSuchElementException` if it's empty. That's why the old code needed to catch `NoSuchElementException`. However, `NavigableMap.firstEntry` returns `null` if the map is empty, so the new code just does a null check, allowing us to avoid catching here. This is a somewhat muddled API, so I assume different methods were created at different times with somewhat confusing semantics

MMcM (approved):

PR #2987, opened 2024-11-26
Fixes #2986: TimeWindowLeaderboardIndex throws NPE on null ranks

### LINKED ISSUE #2986: TimeWindowLeaderboardIndex throws NPE on null ranks
The `TimeWindowLeaderboardIndexMaintainer` currently throws a null pointer exception on this line:
https://github.com/FoundationDB/fdb-record-layer/blob/ddb0edf7e4e919bdfab214433750ae3c9ed00e61/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/TimeWindowLeaderboardIndexMaintainer.java#L622

In particular, `rankForScore` returns a `Long`, which is then cast into a `long` when the `TimeWindowRankAndEntry` is constructed. If the `Long` is `null`, then the implicit unboxing throws an NPE.

alecgrieser:
This modifies the `TimeWindowRankAndEntry` struct to allow the returned `rank` to be empty. This represents the case where a record was not found in one of the leaderboards and so a `null` rank is returned. Note that this returns us to the state of the world prior to #2834, which is what introduced the struct (with its non-null `rank` field). Prior to that PR, the rank could be `null`, and so any code that worked prior to that commit (which is also the most recent change as of writing) should be able to be trivially adapted to this new version.

This fixes #2986.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/LeaderboardIndexTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/LeaderboardIndexTest.java
@@ -62,6 +62,7 @@ import com.apple.foundationdb.record.test.FDBDatabaseExtension;
 import com.apple.foundationdb.record.test.TestKeySpace;
 import com.apple.foundationdb.record.test.TestKeySpacePathManagerExtension;
 import com.apple.foundationdb.tuple.Tuple;
+import com.apple.test.BooleanSource;
 import com.apple.test.Tags;
 import com.google.common.primitives.Longs;
 import com.google.protobuf.Message;
@@ -70,6 +71,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
 
 import javax.annotation.Nullable;
 import java.math.BigInteger;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/LeaderboardIndexTest.java
```diff
@@ -527,6 +530,52 @@ public void basicUngroupedNestedReverse() {
         }
     }
 
+    @ParameterizedTest(name = "emptyUngroupedNested[highScoreFirst={0}]")
+    @BooleanSource
+    public void emptyUngroupedNested(boolean highScoreFirst) {
```
alecgrieser, line 535: This test actually passes both before and after this change, but having coverage of an empty index seemed like a good idea.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/leaderboard/LeaderboardIndexTest.java
```diff
@@ -527,6 +530,52 @@ public void basicUngroupedNestedReverse() {
...
+
+            assertNull(leaderboards.evaluateQueryFunction(leaderboards.queryRank(), rec));
+            assertNull(leaderboards.evaluateQueryFunction(leaderboards.queryTimeWindowRank(TimeWindowLeaderboard.ALL_TIME_LEADERBOARD_TYPE, -1), rec));
+            assertNull(leaderboards.evaluateQueryFunction(leaderboards.queryTimeWindowRank(TEN_UNITS, 10102), rec));
+        }
+    }
+
+    @ParameterizedTest(name = "missingFromUngrouped[highScoreFirst={0}]")
+    @BooleanSource
+    public void missingFromUngrouped(boolean highScoreFirst) {
```
alecgrieser, line 563: Both this test and the one below would fail with an NPE prior to the fix in this PR

MMcM (approved):

Issue #2990, opened 2024-12-04
LucenePartitioner is not thread safe if multiple independent records are updated concurrently


ScottDugas:
If multiple different records in the same partition are added concurrently the counts can get skewed, and in theory, so can the endpoints, although that's less likely.

Issue #2989, opened 2024-12-04
Synchronization blocks in FDBDirectoryWrapper can cause a deadlock


ScottDugas:
If multiple concurrent writes to different records come in (which is normally allowed), and the lucene IndexWriter hadn't already been initialized, the synchronized block can use up all the threads in the ForkJoinPool and stall.

ScottDugas:
Related: https://github.com/FoundationDB/fdb-record-layer/issues/3146 the synchronized block can run into issues because it's not actually sufficient.

PR #2992, opened 2024-12-05
Resolve #2991: tests for Lucene "all-match" scans

ohadzeliger:

PR #2996, opened 2024-12-05
Add a test that does concurrent lucene updates

ScottDugas:
This adds a new test that does concurrent updates to different records in lucene.
The test references two specific issues:
- #2989
- #2990 

And configures itself so as to avoid hitting those issues, but as those are fixed, we can expand the test to include coverage for those issues.

### DIFF
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
@@ -784,6 +784,15 @@ public class LuceneIndexMaintenanceTest extends FDBRecordStoreConcurrentTestBase
         }
     }
 
+    /**
+     * Test that updating records in the same transaction does not cause issues with the executor, and doesn't result in
+     * a corrupted index.
+     * <p>
+     *     See issues: <a href="https://github.com/FoundationDB/fdb-record-layer/issues/2989">#2989</a> and
+     *     <a href="https://github.com/FoundationDB/fdb-record-layer/issues/2990">#2990</a>.
+     * </p>
+     * @throws IOException if there's an issue with Lucene
+     */
     @Test
     void concurrentUpdate() throws IOException {
         // Once the two issues noted below are fixed, we should make this parameterized, and run with additional random
--- a/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
+++ b/fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
@@ -75,6 +75,7 @@ import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
@@ -86,6 +87,8 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
```diff
@@ -128,71 +126,102 @@ void validate(Index index, final Map<Tuple, Map<Tuple, Tuple>> expectedDocumentI
...
                     .collect(Collectors.toList());
-            List<LucenePartitionInfoProto.LucenePartitionInfo> partitionInfos = getPartitionMeta(index, groupingKey);
-            partitionInfos.sort(Comparator.comparing(info -> Tuple.fromBytes(info.getFrom().toByteArray())));
-            final String allCounts = partitionInfos.stream()
-                    .map(info -> Tuple.fromBytes(info.getFrom().toByteArray()).toString() + info.getCount())
-                    .collect(Collectors.joining(",", "[", "]"));
-            Set<Integer> usedPartitionIds = new HashSet<>();
-            Tuple lastToTuple = null;
-            int visitedCount = 0;
+            if (partitionHighWatermark > 0) {
```
ScottDugas, line 133: Tip: This code is much easier to review if you hide whitespace changes, because the bulk of `validateUnpartitionedGroup` is the same, just indented.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -781,6 +784,85 @@ public void ensureValid() throws IOException {
...
+                .build();
+        for (int i = 0; i < loopCount; i++) {
+            LOGGER.info(KeyValueLogMessage.of("concurrentUpdate loop",
+                    "iteration", i,
+                    "groupCount", dataModel.groupingKeyToPrimaryKeyToPartitionKey.size(),
+                    "docCount", dataModel.groupingKeyToPrimaryKeyToPartitionKey.values().stream().mapToInt(Map::size).sum(),
+                    "docMinPerGroup", dataModel.groupingKeyToPrimaryKeyToPartitionKey.values().stream().mapToInt(Map::size).min(),
+                    "docMaxPerGroup", dataModel.groupingKeyToPrimaryKeyToPartitionKey.values().stream().mapToInt(Map::size).max()));
+
+            long start = 234098;
```
ohadzeliger, line 835: From a previous PR - shouldn't we change the start between iterations?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -781,6 +784,85 @@ public void ensureValid() throws IOException {
...
 
+    @Test
+    void concurrentUpdate() throws IOException {
+        // Once the two issues noted below are fixed, we should make this parameterized, and run with additional random
+        // configurations.
+        AtomicInteger threadCounter = new AtomicInteger();
+        // Synchronization blocks in FDBDirectoryWrapper can cause a deadlock
+        // see https://github.com/FoundationDB/fdb-record-layer/issues/2989
+        // So set the pool large enough to overcome that
+        this.dbExtension.getDatabaseFactory().setExecutor(new ForkJoinPool(30,
```
ohadzeliger, line 795: Does that have to be cleaned at the end of the test or is the dbFactory getting a new executor on every test?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestDataModel.java
```diff
@@ -64,6 +66,7 @@ public class LuceneIndexTestDataModel {
      * </p>
      */
     final Map<Tuple, Map<Tuple, Tuple>> groupingKeyToPrimaryKeyToPartitionKey;
+    Map<Tuple, Function<Message, Message>> updateableRecords;
```
ohadzeliger, line 69: This seems very specific and limited - won't it make sense to store the records as "saved records" and handle the update-specific logic in the test class rather than at the model?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -781,6 +784,85 @@ public void ensureValid() throws IOException {
         }
     }
 
+    @Test
+    void concurrentUpdate() throws IOException {
```
ohadzeliger, line 788: Just to make sure I understand: This test ensures that concurrent updates in the same transaction do not crash, hang or corrupt the index, right?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ScottDugas: I assume you're referring to this: https://github.com/FoundationDB/fdb-record-layer/pull/2992#discussion_r1878883222
I think I reasoned to myself there, that `start` was added to make the "timestamp" look more like a timestamp, and not just 0, 1, 2, ...
We auto increment based on the count in the group (plus some randomness), so this really could be 0.
I think it would be best to push `start` into a field in the data model, but perhaps as a later PR (I have some additional tests I've started adding on top of this)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestDataModel.java
ScottDugas: I may be hyper-focused on the maintenance tests which have less interesting record shapes, but I feel like, if we want to pull this out, I would also want to abstract away the types a little bit, so that the code above doesn't have to worry about whether it is synthetic or not.
Or perhaps, that the `Function` here should be replaced with an object, that has a variety of operations.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ScottDugas: It does not. The https://github.com/FoundationDB/fdb-record-layer/blob/5f21535c68bf397bf7636335b00cfccbc72878fa/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/test/FDBDatabaseExtension.java#L154 clears out the database factory after each test, so that each test has it's own factory, and thus they can test things like this without affecting others, either serially or in parallel.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ScottDugas: I will add that as a comment.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestDataModel.java
ohadzeliger: For the sake of making progress, then, would you like to check this in as-is and postpone these improvements to a later pr?

ScottDugas: I changed this to be an object in https://github.com/FoundationDB/fdb-record-layer/pull/3003 but I think there is definitely room to grow, and polish there.

PR #3001, opened 2024-12-12
Remove static LuceneIndexTestDataModel methods

ScottDugas:
There was only one usage of this that wasn't already creating a `LuceneIndexTestDataModel`, this changes `concurrentStoreTest` so that it can just use the model and then removes the static methods.
Removing the static methods will allow us to more easily maintain state, without having to add more and more parameters to the `saveRecord` method.

Resolves: #3002


ohadzeliger (changes requested):
Looks good- just one small comment

PR #3016, opened 2025-01-07
Move Stable API to Unstable

arnaud-lacurie:


fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -38,7 +40,8 @@
 import java.util.Map;
 import java.util.Objects;
 
-public final class Options {
+@API(API.Status.EXPERIMENTAL)
+public class Options {
```
alecgrieser, line 44: I'm not necessarily opposed to removing the `final` here, but it does seem like a bit of a separate change. I'm not sure the `final`ity of our classes was particularly thoughtfully considered, but final-by-default is probably closer to what we want than the other way around

fdb-relational-api/fdb-relational-api.gradle
```diff
@@ -53,11 +53,15 @@ buildHost=${InetAddress.getLocalHost().getCanonicalHostName()}
 }
 
 dependencies {
+    api project(':fdb-extensions')
+    annotationProcessor project(':fdb-extensions')
```
alecgrieser, line 57: There's a bit of a problem here, in that `fdb-extensions` has the FDB client as a transitive dependency, which we don't want to take on in the `fdb-relational-api` project (as it's used by the client subproject). One possible solution would be to split out the stability annotations out of `fdb-extensions` into their own standalone sub-project (`fdb-api-annotations` or something) that all of the other projects can then add as a dependency. Or we hold off on adding API stability annotations in `fdb-relational-api`, `fdb-relational-grpc`, and `fdb-relational-jdbc`, which I think are the three subprojects that we don't want to have a dependency on `fdb-java`.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/SqlInsertTest.java
```diff
@@ -39,6 +41,7 @@
 import java.sql.DriverManager;
 import java.util.Map;
 
+@API(API.Status.EXPERIMENTAL)
 public class SqlInsertTest {
```
alecgrieser, line 45: I don't think we really need API annotations on test classes. It's probably not a big deal if they're there, just unnecessary

alecgrieser (changes requested):
I think the big thing here is that the current PR adds `fdb-java` as a transitive dependency to a few subprojects that we don't want to have it (namely `fdb-relational-api` and its descendants)

alecgrieser:
I'm going to merge this and then tackle moving the annotations out and also removing API stability annotations in tests into a separate PR

alecgrieser (approved):

PR #1869, opened 2022-10-13
Update documentation for 3.3 version

alecgrieser:
There are a few places in our docs where we reference the current version. This updates those, as #1855 updated the minor version to 3.3 from 3.2

PR #3021, opened 2025-01-09
Add release notes and bump version to 4.0

arnaud-lacurie:


alecgrieser (commented):
There are references to the old version number (3.5) in `docs/Versioning.md` that we usually update when we make changes to the minor (or major) version. See: #1869 for how that's been done in the past

alecgrieser (approved):

PR #3030, opened 2025-01-10
Server continuations

ohadzeliger:
Add support for continuations for the client-server (grpc) relational layer.
Client can get continuations from a result set and add them to the queries to continue the query.
Continuations are only available after exhausting all the rows in the result set (on or after the last row).


fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockContinuation.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public boolean atEnd() {
+        return atEnd;
+    }
+
+    @Override
+    public byte[] serialize() {
+        return toString().getBytes(StandardCharsets.UTF_8);
```
ScottDugas, line 71: It seems weird to have a non-reversible `serialize()`, but I'm assuming you don't want to implement a new protobuf to support serializing the mock...

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+        this.columnTypes = columnTypes;
+    }
+
+    @Override
+    public String getTypeName() throws SQLException {
+        return typeName;
+    }
+
+    @Override
+    public StructMetaData getStructMetaData(int oneBasedColumn) throws SQLException {
```
ScottDugas, line 48: Should this throw an exception instead of returning `null`?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+        return null;
+    }
+
+    @Override
+    public ArrayMetaData getArrayMetaData(int oneBasedColumn) throws SQLException {
+        return null;
```
ScottDugas, line 54: Should this throw an exception instead of returning `null`?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public int getColumnCount() throws SQLException {
+        return columnTypes.size();
+    }
+
+    @Override
+    public String getColumnName(int column) throws SQLException {
+        return "";
```
ScottDugas, line 64: Should this throw an exception instead of returning `""`?

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/result_set.proto
```diff
@@ -41,6 +42,10 @@ message ResultSet {
   // Each element in `rows` is a row whose format is defined by metadata. The ith element
   // in each row matches the ith field in ResultSetMetadata.
   repeated Struct row = 2;
+
+  // Optional continuation that can continue the query from the point the current results ended.
+  // Not all queries support continuations. Null (or non existent) continuations means "no continuation".
```
ScottDugas, line 47: What queries don't support continuations?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ProtobufConversionTest.java
```diff
@@ -63,4 +72,37 @@ public void testToColumnIntegerBiFunction() {
...
+
+        Assertions.assertEquals(2, converted.getRowCount());
+        assertRow(List.of(1, 2, 3), converted.getRow(0));
+        assertRow(List.of(4, 5, 6), converted.getRow(1));
+        assertContinuation(continuation.serialize(), RpcContinuationReason.TRANSACTION_LIMIT_REACHED, converted.getContinuation());
+    }
+
+    private void assertContinuation(byte[] state, RpcContinuationReason reason, RpcContinuation actual) {
+        if (state == null) {
+            Assertions.assertNull(actual.getInternalState().toByteArray());
```
ScottDugas, line 94: You don't actually utilize this path in this class

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -225,6 +224,9 @@ void checkResultInternal(@Nonnull Object actual, @Nonnull String queryDescriptio
                 var success = isExact ? getVal().equals(actualPlan) : actualPlan.contains((String) getVal());
                 if (success) {
                     logger.debug("✅️ plan match!");
+                    while (resultSet.next()) {
```
ScottDugas, line 227: Why do you need to exhaust the resultSet here?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.stream.Stream;
+
+public class ResultSetContinuationTest {
+
+    public static Stream<Arguments> continuationSource() {
+        return Stream.of(
+                Arguments.of(MockContinuation.BEGIN),
+                Arguments.of(MockContinuation.END),
+                Arguments.of(new MockContinuation(Continuation.Reason.TRANSACTION_LIMIT_REACHED, new byte[]{1}, false, false)),
+                Arguments.of(new MockContinuation(Continuation.Reason.CURSOR_AFTER_LAST, new byte[]{1}, false, true)),
```
ScottDugas, line 46: I feel like theses should be different
```suggestion
                Arguments.of(new MockContinuation(Continuation.Reason.TRANSACTION_LIMIT_REACHED, new byte[]{1}, false, false)),
                Arguments.of(new MockContinuation(Continuation.Reason.CURSOR_AFTER_LAST, new byte[]{2}, false, true)),
```

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
```diff
@@ -0,0 +1,134 @@
...
+                Arguments.of(new MockContinuation(Continuation.Reason.QUERY_EXECUTION_LIMIT_REACHED, new byte[0], false, true)));
```
ScottDugas, line 47: Why does this have an empty state?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
```diff
@@ -0,0 +1,134 @@
...
+                null,
+                TestUtils.row(1, 2, 3), TestUtils.row(4, 5, 6));
+        Assertions.assertThrows(NullPointerException.class, () -> TypeConversion.toProtobuf(resultSet));
+    }
+
+    private void assertContinuation(Continuation expected, Continuation actual) {
+        Assertions.assertEquals(expected.atBeginning(), actual.atBeginning());
+        Assertions.assertEquals(expected.atEnd(), actual.atEnd());
+        Assertions.assertEquals(expected.getReason(), actual.getReason());
+        // This, again, assumes that the entire inner continuation is serialized as the state
```
ScottDugas, line 115: Why not call `actual.serialize()`?

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
```diff
@@ -0,0 +1,134 @@
...
+    void testMidContinuationFails() throws Exception {
+        RelationalResultSet resultSet = TestUtils.resultSet(
+                "TestType",
+                List.of(Types.INTEGER, Types.INTEGER, Types.INTEGER),
+                MockContinuation.BEGIN,
+                TestUtils.row(1, 2, 3), TestUtils.row(4, 5, 6));
+        ResultSet rsProto = TypeConversion.toProtobuf(resultSet);
+
+        try (RelationalResultSetFacade deserializedResultSet = new RelationalResultSetFacade(rsProto)) {
+            Assertions.assertThrows(SQLException.class, () -> deserializedResultSet.getContinuation());
```
ScottDugas, line 97: This is only testing that it fails before getting any results. It seems like that is a good thing to test, but also that we should test after getting one.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalRpcContinuation.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public RelationalRpcContinuation(@Nonnull RpcContinuation proto) {
+        this.proto = proto;
+    }
+
+    @Override
+    public byte[] serialize() {
+        // Serialize the INTERNAL STATE. This is done since on the client side, when the query is modified to include an
+        // encoded version of the continuation as a string, we have to use the INTERNAL flavor of the continuation to have
+        // it properly deserialized by the server. The server only accepts the internal continuation as part of the query.
```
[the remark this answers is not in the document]
ScottDugas: ok, IIUC the continuation is combining two parts of `RecordCursorResult`, namely `NoNextReason` and `RecordCursorContinuation`. The `NoNextReason` is not relevant when actually continuing, so we are serializing the `executionState` which is really the `RecordCursorContinuation`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -154,7 +153,7 @@ private static String appendWithContinuationIfPresent(@Nonnull String queryStrin
         if (!currentQuery.isEmpty() && currentQuery.charAt(currentQuery.length() - 1) == ';') {
             currentQuery = currentQuery.substring(0, currentQuery.length() - 1);
         }
-        if (continuation instanceof ContinuationImpl) {
+        if (continuation != null) {
```
ScottDugas, line 156: Should we be adding any sort of assertions to the yaml test framework about the reason for the continuation?
Should we assert that we never get back the beginning continuation?
I don't think the yaml tests have a good way to cause `TRANSACTION_LIMIT_REACHED`, but the others should be testable in an integrated way, which seems worthwhile.

ScottDugas (changes requested):

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalRpcContinuation.java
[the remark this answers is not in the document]
ohadzeliger: The `ContinuationImpl` is the "server-side" continuation - this is the one that wraps around the record layer cursor state (among other things). The Relational layer expects the serialization of this continuation to be part of the query - the query parser decodes it and then deserializes it into a `ContinuatonImpl`.
The client-server continuation `RelationalRpcContinuation` wraps around `ContinuationImpl` and adds the other state pieces required by the API (`reason`, `atBeginning`, `atEnd`). The `ContinuationImpl` is then stored as `InternalState`. When the query is constructed on the client side with the continuation, we *have* to serialize the `ContinuationImpl` into the query since the query parser does not know how to deserialize anything else.
This is quite hacky but I did not see how to achieve the goal without breaking architectural dependency or rewriting the query parser.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalRpcContinuation.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * RelationalRpcContinuation.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2021-2024 Apple Inc. and the FoundationDB project authors
```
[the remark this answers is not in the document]
ohadzeliger: Done for all new files

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/AbstractMockResultSet.java
```diff
@@ -0,0 +1,193 @@
...
+
+    protected abstract MockResultSetRow advanceRow() throws RelationalException;
+
+    @Override
+    public boolean next() throws SQLException {
+        try {
+            currentRow = advanceRow();
+            return currentRow != null;
+        } catch (RelationalException ve) {
+            throw ve.toSqlException();
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockContinuation.java
ohadzeliger: Yes, and this is helpful for debugging and it also makes it obvious in case we try to deserialize it accidentally. We might be forced to change it if the code requires it to be deserialized but so far we don't.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
ohadzeliger: Done.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
ohadzeliger: Done.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/MockResultSetMetadata.java
ohadzeliger: Done.

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/result_set.proto
ohadzeliger: That wasn’t accurate. The field is optional since some result sets do not support continuations - clarified.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ProtobufConversionTest.java
ohadzeliger: Right - that remained from an earlier version of the test. Changed.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: `resultSet` is actually a reference to `actual` parameter, and by the call stack, (see `QueryExecutor:98`):
```
                    config.checkResult(queryResult, this.toString());
                    if (queryResult instanceof RelationalResultSet) {
                        continuationAfter = ((RelationalResultSet) queryResult).getContinuation();
                    }
```

will fail otherwise.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: Sure, Done.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: Empty state is a valid state (as well as null (see BEGIN/END), so I thought this is a good test case.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: Actual in this case is the RPC continuation (that wraps around the inner, server-side, continuation). 
I can pull that out of the assert method to the test method, if that helps.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ScottDugas: Yes, but why would there be more results? Shouldn't an explain always return one?

ScottDugas (changes requested):

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/result_set.proto
```diff
@@ -44,7 +44,7 @@ message ResultSet {
   repeated Struct row = 2;
 
   // Optional continuation that can continue the query from the point the current results ended.
-  // Not all queries support continuations. Null (or non existent) continuations means "no continuation".
+  // A result set is non-null for the RelationalResultSet API but is not always supported (exception thrown when not).
```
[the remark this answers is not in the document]
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Good catch. There was a bug in the resultSet's `hasNext()` that prevented from returning the continuation when we were on the last row. This is fixed and now covered by the improved `testMidContinuationFails` test.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: In this case the continuation is always `RelationalRpcContinuation` and so they are always the same. Done.

fdb-relational-grpc/src/test/java/com/apple/foundationdb/relational/jdbc/ResultSetContinuationTest.java
ohadzeliger: Yes, that's a good enhancement. Done. 
This was also used to cover the off-by-one bug in the result set facade that only allowed continuations after the last row.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Created a ticket

fdb-relational-grpc/fdb-relational-grpc.gradle
```diff
@@ -65,13 +65,11 @@ dependencies {
...
-    runtimeOnly("org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}")
-    implementation("com.google.protobuf:protobuf-java:${protobufVersion}")
-    testImplementation("org.junit.jupiter:junit-jupiter-api:${junitVersion}")
-    testImplementation("org.junit.jupiter:junit-jupiter-params:${junitVersion}")
-    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${junitVersion}")
+    runtimeOnly(libs.log4j.slf4jBinding)
+
+    testImplementation(libs.bundles.test.impl)
+    testRuntimeOnly(libs.bundles.test.runtime)
+    testImplementation(libs.grpc.testing)
```
ScottDugas, line 72: Huh, I'm surprised you don't need junit in here, but it ran the tests, so I guess it's good.

ScottDugas (approved):

PR #3031, opened 2025-01-10
Ability to prevent yamsql tests from running against old versions.

ScottDugas:
With the new mixed-mode tests being run as part of PRB, we'll need a way to say that a test should not run against old versions. This adds support for disabling an entire yaml file. Future steps would be to allow disabling a specific query, providing a specific supported version, and also having the build update the `!current_version` to the version being released.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
@@ -24,11 +24,10 @@ import com.apple.foundationdb.relational.api.Options;
 import com.apple.foundationdb.relational.api.RelationalConnection;
 import com.apple.foundationdb.relational.api.RelationalPreparedStatement;
 import com.apple.foundationdb.relational.api.RelationalStatement;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.net.URI;
 import java.sql.Array;
 import java.sql.DatabaseMetaData;
@@ -36,7 +35,12 @@ import java.sql.SQLException;
 import java.sql.Struct;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * A connection factory that creates a connection that can be used with multiple servers.
@@ -47,6 +51,7 @@ import java.util.stream.Collectors;
 public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFactory {
     // The fixed index of the default connection
     public static final int DEFAULT_CONNECTION = 0;
+    private final Set<String> versionsUnderTest;
 
     /**
      * Server selection policy.
@@ -64,23 +69,32 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
     private final List<YamlRunner.YamlConnectionFactory> alternateFactories;
 
     public MultiServerConnectionFactory(@Nonnull final YamlRunner.YamlConnectionFactory defaultFactory,
-                                        @Nullable final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
+                                        @Nonnull final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
         this(ConnectionSelectionPolicy.DEFAULT, 0, defaultFactory, alternateFactories);
     }
 
     public MultiServerConnectionFactory(@Nonnull final ConnectionSelectionPolicy connectionSelectionPolicy,
                                         final int initialConnection,
                                         @Nonnull final YamlRunner.YamlConnectionFactory defaultFactory,
-                                        @Nullable final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
+                                        @Nonnull final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
         this.connectionSelectionPolicy = connectionSelectionPolicy;
         this.initialConnection = initialConnection;
         this.defaultFactory = defaultFactory;
         this.alternateFactories = alternateFactories;
+        this.versionsUnderTest =
+                Stream.concat(Stream.of(defaultFactory), alternateFactories.stream())
+                        .flatMap(yamlConnectionFactory -> yamlConnectionFactory.getVersionsUnderTest().stream())
+                        .collect(Collectors.toSet());
     }
 
     @Override
     public RelationalConnection getNewConnection(@Nonnull URI connectPath) throws SQLException {
         return new MultiServerRelationalConnection(connectionSelectionPolicy, initialConnection, defaultFactory.getNewConnection(connectPath), alternateConnections(connectPath));
+    }
+
+    @Override
+    public Set<String> getVersionsUnderTest() {
+        return versionsUnderTest;
     }
 
     @Nullable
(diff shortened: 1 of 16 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
```diff
@@ -64,25 +69,34 @@ public enum ConnectionSelectionPolicy { DEFAULT, ALTERNATE }
     private final List<YamlRunner.YamlConnectionFactory> alternateFactories;
 
     public MultiServerConnectionFactory(@Nonnull final YamlRunner.YamlConnectionFactory defaultFactory,
-                                        @Nullable final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
+                                        @Nonnull final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
```
ohadzeliger, line 72: Do we have to change the parameter to @Nonnull? I guess we can still manage with only "current version" connection, which would translate to "current version"?
I think that even with this change, if we have two "embedded connection" factories, the result would still have an empty version set.

ohadzeliger, line 72: making alternateFactories Nullable would probably mean doing the same at line 48

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlRunner.java
```diff
@@ -65,7 +67,24 @@ public final class YamlRunner {
     private final YamlExecutionContext executionContext;
 
     public interface YamlConnectionFactory {
+        /**
+         * Convert a connection uri into an actual connection.
+         * @param connectPath the path to connect to
+         * @return A new {@link RelationalConnection} for the given path appropriate for this test class
```
ohadzeliger, line 73: Add "...and an empty set if we only support the current version?"?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Block that configures aspects of the test that do not require a connection yet.
+ * <p>
+ *     This supports the following sub-commands:
+ *     <ul>
+ *         <li>{@code supported_version}: if this is set, it will disable the test when running against a server
```
ohadzeliger, line 38:
```suggestion
 *         <li>{@code supported_version}: if this is set, it will enable the test only when running against a server
```

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
```diff
@@ -0,0 +1,103 @@
...
+            final Set<String> versionsUnderTest = executionContext.getConnectionFactory().getVersionsUnderTest();
+            // IntelliJ, at least, doesn't display the reason, so log it
+            logger.info(
+                    "Skipping test that only works against the current version, when we're running with these versions: {}",
+                    versionsUnderTest);
+            Assumptions.assumeTrue(versionsUnderTest.isEmpty(),
+                    () -> "Test only works against the current version, but we are running with these versions: " +
+                    versionsUnderTest);
+        } else {
+            throw new RuntimeException("Unsupported supported_version: " + supportedVersion);
```
ohadzeliger, line 69: Will support for concrete version (the one that will be created by the build script once a release is cut) be supported later?

ohadzeliger, line 69: Will that allow an empty "OPTIONS" block? Should we allow that?

ohadzeliger (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
ScottDugas: Line 48 is in the middle of a comment, so I don't think that's what you meant?
Did you mean line 68?

ScottDugas: I guess I should have followed through with changing downstream, but AFAICT, there is no difference between passing in `null` and `List.of()`, and if we change this to `@Nonnull`, we can remove a bunch of nullability checks below.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlRunner.java
ScottDugas: I'm assuming you meant to put this comment in response to the javadoc on `getVersionsUnderTest` (line 85)

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
ScottDugas: I'd prefer to continue to use `disable`, to align with the junit annotations, but reworded it, because it should be disabled when running older versions, not newer versions.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
ScottDugas: Yes, Added that to the PR description.

ScottDugas: Currently we will not support an options block that doesn't have a `supported_version`; I don't see much reason to do so (empty, or with other random stuff).

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
ohadzeliger: Yeah, empty list will work too.

ohadzeliger: Yes. line numbers changed with the repo move

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
ohadzeliger: Cool.
So, if you were to add `supported_version: !current_version` to the YAML, it will prevent the test from running and because it will not be changed to a specific version number it will keep skipping that test, right?

ohadzeliger (dismissed):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
ScottDugas: Correct, and once we start updating it in the build, anything that wasn't manually changed to a version will get that version.

ohadzeliger (approved):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/FileOptions.java
```diff
@@ -0,0 +1,103 @@
...
+ *         <p>
+ *             <b>Example:</b>
+ *             <pre>{@code
+ *                 ---
+ *                 options:
+ *                     supported_version: !current_version
+ *             }</pre>
+ *     </ul>
+ */
+public class FileOptions {
```
g31pranjal, line 51: Not that it is necessary, do you think there is an advantage in ensuring that the FileOptions is strictly the first block in a YAMSQL file (if one is present)? However, that does not matter currently, as it would simply fail the assumption if conditions are not met at parse time but it would still be good to have such a syntax rule as a way of enforcing developers to defines options the first thing in the file (for better readability?). 

We may also want to avoid having multiple fileOption blocks in the file as a way to not allow having multiple option values.

ScottDugas: I created a followup issue to add those restrictions: #3045

PR #3040, opened 2025-01-14
Support UDF in plan generator

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/ScalarValuedFunction.java
```diff
@@ -0,0 +1,67 @@
...
+import com.apple.foundationdb.record.RecordMetaDataProto;
+import com.apple.foundationdb.record.query.plan.cascades.MacroFunction;
+import com.apple.foundationdb.record.query.plan.serialization.DefaultPlanSerializationRegistry;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Defines a scalar User-defined-function.
+ */
+public class ScalarValuedFunction {
```
normen662, line 34: Why don't we just make a trait-like interface called `SerializableFunction` that makes this proto-serializable? `MacroFunction` can then implement `SerializableFunction` Then we don't have to wrap the `MacroFunction` here. It would also allow us to
support other functions in the future that are not `MacroFunction`. It's pretty straightforward for `toProto(...)`, slightly less so on the way back. Please look how the plan serialization achieves the same thing by using:
1. `oneof` in the proto declarations
2. `PlanDeserializer` using `AutoService` to get the  dispatch of the from proto part right

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Key.java
```diff
@@ -392,21 +392,21 @@ public static class Evaluated {
          * Values used in index keys in place of missing fields.
          */
         public enum NullStandin {
-            NULL(RecordMetaDataProto.Field.NullInterpretation.NOT_UNIQUE), // Missing field here skips uniqueness checks.
-            NULL_UNIQUE(RecordMetaDataProto.Field.NullInterpretation.UNIQUE), // Missing field here like ordinary value, but null, for uniqueness.
-            NOT_NULL(RecordMetaDataProto.Field.NullInterpretation.NOT_NULL); // Missing field has type's ordinary default value.
+            NULL(RecordKeyExpressionProto.Field.NullInterpretation.NOT_UNIQUE), // Missing field here skips uniqueness checks.
```
normen662, line 395: static imports may make this more readable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -131,6 +136,7 @@ protected RecordMetaData(@Nonnull Descriptors.FileDescriptor recordsDescriptor,
                              @Nonnull Map<String, Index> indexes,
                              @Nonnull Map<String, Index> universalIndexes,
                              @Nonnull List<FormerIndex> formerIndexes,
+                             @Nonnull Set<ScalarValuedFunction> scalarValuedFunctions,
```
normen662, line 139: There are a few different comments further down in this PR around this. Let me just point out here that this structure is persisted and long-lived. In fact, modifying its contents and layout may be impossible or very hard once this is used. One thing I would like to avoid is to get ourselves into a dead end here.
We most likely want to support other functions-like objects than macros. For instance, we would like to use full SQL-bodied functions or table functions. This attribute here should potentially be able to contain all of these. Using my naming from a files down, this should be something named `SerializableFunction` (because that's all you need here, i.e. a name, a bunch of parameter declarations with their types and a return type). Under the hood it can be any of the particular flavors I mentioned (of which we only have a simple `MacroFunction` as of yet).

normen662 (changes requested):
I am putting @alecgrieser on this review as well to make sure the proto layout and dependencies are what we want. I left a bunch of comments. There is some more work required but it's on the right track. I would recommend not ever removing `-Werror` (not even in your local build). It is useful. Also, please make sure the PR runs through the PRB.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MacroFunction.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return builder
+                .setFunctionName(functionName)
+                .setBody(bodyValue.toValueProto(serializationContext))
+                .build();
+    }
+
+
+    @Nonnull
+    public PValue toValueProto(@Nonnull final PlanSerializationContext serializationContext) {
```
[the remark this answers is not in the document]
pengpeng-lu: removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MacroFunction.java
```diff
@@ -0,0 +1,89 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.values.QuantifiedObjectValue;
+import com.apple.foundationdb.record.query.plan.cascades.values.Value;
+import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Main interface for defining a user-defined function that can be evaluated against a number of arguments.
```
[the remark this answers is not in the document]
pengpeng-lu: done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/MacroFunctionTest.java
```diff
@@ -0,0 +1,78 @@
...
+                Column.of(fields.get(0), new LiteralValue<>(fields.get(0).getFieldType(), "Rose")),
+                Column.of(fields.get(1), new LiteralValue<>(fields.get(1).getFieldType(), 1L))
+        ));
+
+        Assertions.assertEquals(FieldValue.ofFieldName(testValue1, "name"), macroFunction.encapsulate(List.of(testValue1)));
+    }
+
+    @Test
+    void testAdd() {
+        List<Type.Record.Field> fields = List.of(
```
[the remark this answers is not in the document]
pengpeng-lu: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Function.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Main interface for defining a function that can be evaluated against a number of arguments.
+ * @param <T> The resulting type of the function.
+ */
+@SuppressWarnings("PMD.AbstractClassWithoutAbstractMethod")
+public abstract class Function<T extends Typed> {
```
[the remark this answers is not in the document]
pengpeng-lu: 👍 makes sense. changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaData.java
```diff
@@ -693,6 +705,7 @@ public RecordMetaDataProto.MetaData toProto(@Nullable Descriptors.FileDescriptor
         }
 
         // Add in the final options.
+        builder.addAllScalarValuedFunction(scalarValuedFunctions.stream().map(ScalarValuedFunction::toProto).collect(Collectors.toList()));
```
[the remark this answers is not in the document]
pengpeng-lu: removed the confusing comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
```diff
@@ -238,14 +238,14 @@ public GraphExpansion visitExpression(@Nonnull final NestingKeyExpression nestin
                         .add(parent.getFieldName())
                         .build();
                 if (NullableArrayTypeUtils.isArrayWrapper(nestingKeyExpression)) {
-                    final RecordMetaDataProto.KeyExpression childProto = nestingKeyExpression.getChild().toKeyExpression();
+                    final RecordKeyExpressionProto.KeyExpression childProto = nestingKeyExpression.getChild().toKeyExpression();
```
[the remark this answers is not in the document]
pengpeng-lu: unfortunately this file also has `metadata.expressions.KeyExpression`, so we can't remove the `RecordKeyExpressionProto` here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/Key.java
pengpeng-lu: unfortunately this file also has metadata.expressions.KeyExpression.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -228,9 +228,10 @@ private void loadProtoExceptRecords(@Nonnull RecordMetaDataProto.MetaData metaDa
                 typeBuilder.setRecordTypeKey(LiteralKeyExpression.fromProtoValue(typeProto.getExplicitKey()));
             }
         }
+        PlanSerializationContext serializationContext = new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE,
+                PlanHashable.CURRENT_FOR_CONTINUATION);
         for (RecordMetaDataProto.PUserDefinedFunction function: metaDataProto.getUserDefinedFunctionsList()) {
-            UserDefinedFunction func = (UserDefinedFunction)PlanSerialization.dispatchFromProtoContainer(new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE,
-                    PlanHashable.CURRENT_FOR_CONTINUATION), function);
+            UserDefinedFunction func = (UserDefinedFunction)PlanSerialization.dispatchFromProtoContainer(serializationContext, function);
```
pengpeng-lu: serializationContext stores record types that it has "seen" before, for record type that we've "seen", it'll be serialized simply with a `reference_id`, so we need the "seen" types to deserialize the proto as well.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/UserDefinedFunctionCatalog.java
```diff
@@ -53,12 +54,12 @@ void registerFunction(@Nonnull final String functionName,
...
 
     public boolean containsFunction(@Nonnull final String name) {
-        return functionsMap.containsKey(name);
+        return isCaseSensitive ? functionsMap.containsKey(name) : functionsMap.containsKey(name.toUpperCase(Locale.ROOT));
     }
 
     @Nonnull
     public Optional<? extends CatalogedFunction> lookup(@Nonnull final String functionName, Expressions arguments) {
-        final var functionSupplier = functionsMap.get(functionName);
+        final var functionSupplier = isCaseSensitive ? functionsMap.get(functionName) : functionsMap.get(functionName.toUpperCase(Locale.ROOT));
```
pengpeng-lu: `TemporaryFunctionTest.createTemporaryFunctionCaseSensitivityOption` failed so I changed these 2 lines, confused why it worked before?

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
```diff
@@ -882,6 +882,117 @@ void aliasingTableToResolveAmbiguityWorks() throws Exception {
         }
     }
 
+    @Test
```
pengpeng-lu: working on moving these tests to yaml test.

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -177,16 +177,12 @@ dropTempFunction
     : DROP TEMPORARY FUNCTION (IF EXISTS)? schemaQualifiedRoutineName=fullId
     ;
 
-createFunction
```
pengpeng-lu, line 184: `createFunction` was never used, so deleted it, as well as the method in the Visitors classes.

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -1115,6 +1116,11 @@ scalarFunctionName
     | JAVA_CALL
     ;
 
+macroFunctionName
```
hatyo: Please use `fullId` instead.

fdb-relational-core/src/main/antlr/RelationalParser.g4
hatyo: ok.

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -269,6 +265,10 @@ routineBody
     // | externalBodyReferences TODO
     ;
 
+macroFunctionBody
```
hatyo: I do not think we need an extra rule here. Can you make it part of the `routineBody`:

```antlrv4
routineBody
    : AS queryTerm         #statementBody
    | AS uid               #marcoBody
    | sqlReturnStatement   #expressionBody
    // | externalBodyReferences TODO
    ;
```

I think this is probably more streamlined with the rest of the invoked routines parsing rules (and less verbose).

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -927,6 +927,7 @@ functionCall
     : aggregateWindowedFunction                                     #aggregateFunctionCall // done (supported)
     | specificFunction                                              #specificFunctionCall //
     | scalarFunctionName '(' functionArgs? ')'                      #scalarFunctionCall // done (unsupported)
+    | macroFunctionName '(' functionArgs? ')'                       #macroFunctionCall
```
hatyo: future-thinking here, can you name this: `userDefinedScalarFunction` instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -228,10 +228,11 @@ private void loadProtoExceptRecords(@Nonnull RecordMetaDataProto.MetaData metaDa
                 typeBuilder.setRecordTypeKey(LiteralKeyExpression.fromProtoValue(typeProto.getExplicitKey()));
             }
         }
+        PlanSerializationContext serializationContext = new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE,
```
hatyo: I think you moved this outside so the creation of the `serializationContext` is done exactly once, instead of `N` times if we have `N` UDFs in the `metadataProto`, right?

The idea is great, but unforuntately, it means we're creating the context even if there are no functions, which seems wasteful, can we find a better solution in the middle that does something like:

```java
          PlanSerializationContext serializationContext = null;
          for (RecordMetaDataProto.PUserDefinedFunction function: metaDataProto.getUserDefinedFunctionsList()) {
              if (serializationContext == null){
serializationContext =new

PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE, PlanHashable.CURRENT_FOR_CONTINUATION);
              }
            // ... 
```

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
```diff
@@ -41,21 +41,23 @@ public class RecordLayerInvokedRoutine implements InvokedRoutine {
...
-    private final Function<Boolean, CompiledSqlFunction> compilableSqlFunctionSupplier;
+    private final Function<Boolean, UserDefinedFunction> userDefinedRoutine;
 
     public RecordLayerInvokedRoutine(@Nonnull final String description,
                                      @Nonnull final String normalizedDescription,
                                      @Nonnull final String name,
                                      boolean isTemporary,
-                                     @Nonnull final Function<Boolean, CompiledSqlFunction> compilableSqlFunctionSupplier) {
+                                     boolean isCompiledSql,
+                                     @Nonnull final Function<Boolean, UserDefinedFunction> userDefinedRoutine) {
```
hatyo: do you think we can use inheritance (with distinct builders) instead to make the distinction between sql functions and macro functions?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
```diff
@@ -171,10 +179,9 @@ public Builder setTemporary(boolean isTemporary) {
         @Nonnull
         public RecordLayerInvokedRoutine build() {
             Assert.notNullUnchecked(name);
-            Assert.notNullUnchecked(description);
-            Assert.notNullUnchecked(compilableSqlFunctionSupplier);
-            return new RecordLayerInvokedRoutine(description, normalizedDescription, name, isTemporary,
-                    compilableSqlFunctionSupplier);
+            // Assert.notNullUnchecked(description);
```
hatyo: Maybe using different builder classes can help here.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
```diff
@@ -171,10 +179,9 @@ public Builder setTemporary(boolean isTemporary) {
...
+            Assert.thatUnchecked(userDefinedFunctionSupplier != null);
```
hatyo: You can use `Assert.notNullUnchecked` instead

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
```diff
@@ -346,12 +355,41 @@ private RecordLayerInvokedRoutine getInvokedRoutineMetadata(@Nonnull final Parse
...
                 .setDescription(functionDefinition)
-                .withCompilableRoutine(ignored -> compiledSqlFunction)
+                .withUserDefinedRoutine(ignored -> compiledSqlFunction, true)
                 .setNormalizedDescription(getDelegate().getPlanGenerationContext().getCanonicalQueryString())
                 .setTemporary(isTemporary)
                 .build();
     }
 
+    @Nonnull
+    private RecordLayerInvokedRoutine getMacroFunctionMetadata(@Nonnull final ParserRuleContext functionCtx,
```
hatyo: this code is copied verbatim from above, please remove it and try to customize the function to return what you need instead.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
```diff
@@ -368,19 +406,50 @@ public ProceduralPlan visitDropTempFunction(@Nonnull RelationalParser.DropTempFu
...
     @Override
     public CompiledSqlFunction visitTempSqlInvokedFunction(@Nonnull RelationalParser.TempSqlInvokedFunctionContext ctx) {
         return visitSqlInvokedFunction(ctx.functionSpecification(), ctx.routineBody(), true);
     }
 
+    @Nonnull
+    private MacroFunction visitMacroFunction(@Nonnull final RelationalParser.FunctionSpecificationContext functionSpecCtx,
+                                                        @Nonnull final RelationalParser.MacroFunctionBodyContext bodyCtx) {
+        // get the function name.
+        final var functionName = visitFullId(functionSpecCtx.schemaQualifiedRoutineName).toString();
```
hatyo: same here. We can sync offline to discuss the options.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizer.java
```diff
@@ -626,7 +627,7 @@ public static NormalizationResult normalizeAst(@Nonnull final SchemaTemplate sch
             // immediate materialization of temporary function, this is required to collect any auxiliary literals discovered
             // during plan generation of the temporary function. The literals and combined with query literals and provided
             // for the execution of a (cached) physical plan.
-            final var compiledFunction = recordLayerRoutine.getCompilableSqlFunctionSupplier().apply(caseSensitive);
+            final var compiledFunction = (CompiledSqlFunction)recordLayerRoutine.getUserDefinedFunctionSupplier().apply(caseSensitive);
```
hatyo, line 632: nothing is preventing us from creating a temporary marco function, but if the user does so, I suppose the cast above will throw?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/SemanticAnalyzer.java
```diff
@@ -502,6 +503,45 @@ public Optional<Expression> lookupNestedField(@Nonnull Identifier requestedIdent
         return Optional.of(nestedAttribute);
     }
 
+    @Nonnull
+    public Optional<Value> lookupNestedField(@Nonnull Identifier requestedIdentifier,
```
hatyo: Is this trying to resolve a field? if so, why not use the method above with a synthetic `logicaloperator`?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/Identifier.java
```diff
@@ -128,6 +128,13 @@ public boolean prefixedWith(@Nonnull Identifier identifier) {
         return true;
     }
 
+    @Nonnull
+    public List<String> removePrefix(@Nonnull Identifier prefix) {
```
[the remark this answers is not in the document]
hatyo: This can be removed once the already existing resolver is used to resolve the nested field which is the result of the macro function.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizerTests.java
```diff
@@ -407,7 +407,7 @@ private static RecordLayerSchemaTemplate schemaTemplateWithFunction(@Nonnull fin
                         .setDescription(functionDdl)
                         .setNormalizedDescription(canonicalFunctionDdl)
                         // invoking the compiled routine should only happen during plan generation.
-                        .withCompilableRoutine(ignored -> new CompiledSqlFunction("", ImmutableList.of(), ImmutableList.of(),
+                        .withUserDefinedRoutine(ignored -> new CompiledSqlFunction("", ImmutableList.of(), ImmutableList.of(),
```
hatyo: I think you can add some tests for adding temporary macro functions.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
hatyo: would be nice to make the YAML tests part of this PR as well :-)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
pengpeng-lu: good idea, changed.

fdb-relational-core/src/main/antlr/RelationalParser.g4
pengpeng-lu: so the reason that I didn't use `fullId` here is to avoid ambiguity, for example, I'll hit exception like `bitmap_bit_position(uid)` is ambiguous, because `fullId` contains `simpleId`, which contains `functionNameBase`, which contains `bitmap_bit_position`, while only using ID and DOUBLE_QUOTE_ID here rules out `functionNameBase`.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
pengpeng-lu: modified.

fdb-relational-core/src/main/antlr/RelationalParser.g4
pengpeng-lu: sure, I also renamed the `MacroFunction` class to `UserDefinedScalarFunction`.

fdb-relational-core/src/main/antlr/RelationalParser.g4
pengpeng-lu: agree, changed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
pengpeng-lu: merge them into 1 method, actually merge it into `routineBody` kinda automatically solve the problem :) Before I didn't like the two similar methods here, but was stuck with different input types.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
pengpeng-lu: removed this method :)

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
pengpeng-lu: yes we can, the downside of distinct builders is that in places where we call `RecordLayerInvokedRoutine.Builder` today, we'll need to do if/else to create different builders. So now I put it in one class, the downside is that we'll need to either pass a boolean into the class to differentiate (my old implementation), or we can pass in the serializable function directly (this is my most recent implementation). Which one do you prefer?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/SemanticAnalyzer.java
pengpeng-lu: merge it with the existing `lookupNestedField` method.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
pengpeng-lu: yes, moved to yaml :)

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizerTests.java
pengpeng-lu: added.

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -44,6 +44,12 @@ public void groupByTests(YamlTest.Runner runner) throws Exception {
         runner.runYamsql("groupby-tests.yamsql");
     }
 
+    @TestTemplate
+    @MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_METRICS)
```
hatyo: please remove.

yaml-tests/src/test/resources/user-defined-scalar-function-tests.yamsql
```diff
@@ -0,0 +1,58 @@
...
+    - query: INSERT INTO NESTED
+        VALUES (1, (200, 1), ((5, 11), (10, 100))),
+               (2, (201, 2), ((5, 12), (10, 100))),
+               (3, (202, 3), ((5, 13), (10, 100))),
+               (4, (203, 4), ((5, 14), (10, 100))),
+               (5, (329, 5), ((5, 15), (10, 140))),
+               (6, (330, 6), ((5, 16), (10, 140)))
+---
+test_block:
+  name: group-by-tests
```
hatyo: rename the block to something like `function-tests`

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
hatyo: should we keep the tests or remove them now that we have equivalent tests in YAML.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/MacroFunctionTest.java
```diff
@@ -34,7 +34,7 @@
 import java.util.Optional;
 
 /**
- * Tests of {@link MacroFunction}.
+ * Tests of {@link UserDefinedScalarFunction}.
  */
 public class MacroFunctionTest {
```
hatyo: Can you rename this to `UserDefinedScalarFunctionTest`?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompilableSqlFunction.java
```diff
@@ -61,7 +60,7 @@
  * function plan as a leg of a binary join, where
  */
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
-public class CompiledSqlFunction extends UserDefinedFunction implements  WithPlanGenerationSideEffects {
+public class CompilableSqlFunction extends com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction implements WithPlanGenerationSideEffects {
```
hatyo: import instead of the fully-qualified class name.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/UserDefinedFunctionCatalog.java
hatyo: Can you please revert this line? it is actually incorrect to perform any normalization here. I think the tested you referred to should pass, but if it is not, we'll have to think about the root cause of the bug and solve it there.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/UserDefinedFunctionCatalog.java
```diff
@@ -53,12 +54,12 @@ void registerFunction(@Nonnull final String functionName,
     }
 
     public boolean containsFunction(@Nonnull final String name) {
-        return functionsMap.containsKey(name);
+        return isCaseSensitive ? functionsMap.containsKey(name) : functionsMap.containsKey(name.toUpperCase(Locale.ROOT));
```
hatyo: please revert.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RecordMetadataDeserializer.java
```diff
@@ -165,7 +169,19 @@ private RecordLayerInvokedRoutine.Builder generateInvokedRoutineBuilder(@Nonnull
                 .setName(name)
                 .setDescription(body)
                 .setTemporary(false)
-                .withCompilableRoutine(getSqlFunctionCompiler(name, metadata, body));
+                .withUserDefinedRoutine(ignored -> getSqlFunctionCompiler(name, metadata, body).apply(ignored))
```
hatyo: why not just call `getSqlFunctionCompiler(name, metadata, body)`?

yaml-tests/src/test/resources/user-defined-scalar-function-tests.yamsql
```diff
@@ -0,0 +1,58 @@
...
+schema_template:
+    create type as struct st1(y bigint, z bigint)
+    create type as struct st2(w bigint, x bigint)
+    create type as struct st3(u st2, v st1)
+    create type as struct st4(s bigint, t bigint)
+    create table nested(id bigint, q st4, r st3, primary key(q.s, r.u.w))
+    create index i2 as select r.v.z from nested order by r.v.z
+    CREATE FUNCTION st2(IN x TYPE ST3) RETURNS ST2 AS x.u
+    CREATE FUNCTION z(IN x TYPE ST3) RETURNS bigint AS x.v.z
+    CREATE FUNCTION self(IN x bigint) RETURNS bigint AS x;
```
hatyo: can you please add more tests? I would like to see whether it is possible to hit an issue related to reusing correlation identifiers. Here is a list of tests we can add:

- scalar function being part of compilable user defined table function.
- invoke the scalar function twice in SQL query `select myFunc(x), myFunc(x) from t`.
- invoke complex user defined table function that has multiple invocations of user defined scalar function multiple times.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RoutineParser.java
```diff
@@ -39,11 +38,11 @@
 public interface RoutineParser {
 
     @Nonnull
-    UserDefinedFunction parse(@Nonnull String routineString, boolean isCaseSensitive);
+    com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction parse(@Nonnull String routineString, boolean isCaseSensitive);
```
hatyo: please avoid long qualified names and do imports instead.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RoutineParser.java
```diff
@@ -39,11 +38,11 @@
...
 
     @Nonnull
-    UserDefinedFunction parseTemporaryFunction(@Nonnull String functionName, @Nonnull String routineString,
-                                               @Nonnull PreparedParams preparedParams, boolean isCaseSensitive);
+    com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction parseTemporaryFunction(@Nonnull String functionName, @Nonnull String routineString,
```
hatyo: ditto.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
```diff
@@ -401,39 +403,57 @@ private CompiledSqlFunction visitSqlInvokedFunction(@Nonnull final RelationalPar
...
-        // 2. add a fake quantifier with the function parameters (if any) to resolve their references in the function body
-        //    during its plan generation.
-        final var fragment = getDelegate().pushPlanFragment();
-
-        parametersCorrelation.ifPresent(quantifier -> fragment.addOperator(LogicalOperator.newUnnamedOperator(
-                Expressions.fromQuantifier(quantifier), quantifier)));
-        if (isTemporary) {
-            body = Assert.castUnchecked(visit(bodyCtx), LogicalOperator.class);
+        if (isScalar) {
+            // user defined scalar function
```
hatyo: please remove the comment, it is kinda obvious ;-)

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
```diff
@@ -401,39 +403,57 @@ private CompiledSqlFunction visitSqlInvokedFunction(@Nonnull final RelationalPar
...
+            final var semanticAnalyzer = getDelegate().getSemanticAnalyzer();
+            // get parameter names and corresponding QuantifiedObjectValue
+            List<Identifier> paramNameIdList = new ArrayList<>();
+            List<QuantifiedObjectValue> paramValueList = new ArrayList<>();
+            for (RelationalParser.SqlParameterDeclarationContext sqlParameterDeclarationContext: functionSpecCtx.sqlParameterDeclarationList().sqlParameterDeclarations().sqlParameterDeclaration()) {
```
hatyo: please use `visitSqlParameterDeclarationList`, it already returns `Expressions` that have the underlying `Value` parameters.

Also, please wrap it with a `DisabledLiteralProcessing` closure, we do NOT want to have the parameter constants (if any) mess with the query semantics, see below how it is done.

You can verify later that you have a single `Expression` passed as argument, and use it in the `SemanticAnalyzer` lookup call later.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/TypedVisitor.java
```diff
@@ -183,13 +183,10 @@ public interface TypedVisitor extends RelationalParserVisitor<Object> {
...
     @Override
-    CompiledSqlFunction visitCreateFunction(RelationalParser.CreateFunctionContext ctx);
+    CompilableSqlFunction visitTempSqlInvokedFunction(RelationalParser.TempSqlInvokedFunctionContext ctx);
 
     @Override
-    CompiledSqlFunction visitTempSqlInvokedFunction(RelationalParser.TempSqlInvokedFunctionContext ctx);
-
-    @Override
-    CompiledSqlFunction visitSqlInvokedFunction(RelationalParser.SqlInvokedFunctionContext ctx);
+    com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction visitSqlInvokedFunction(RelationalParser.SqlInvokedFunctionContext ctx);
```
hatyo: please avoid using long fully-qualified class names and use `import` instead if possible.

yaml-tests/src/test/java/YamlIntegrationTests.java
pengpeng-lu: removed.

yaml-tests/src/test/resources/user-defined-scalar-function-tests.yamsql
pengpeng-lu: renamed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/MacroFunctionTest.java
pengpeng-lu: renamed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompilableSqlFunction.java
pengpeng-lu: changed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RoutineParser.java
pengpeng-lu: changed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RoutineParser.java
pengpeng-lu: 👍

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/TypedVisitor.java
pengpeng-lu: 👍 probably IntelliJ did this when I renamed the class...

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DdlVisitor.java
pengpeng-lu: sure :)

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/serde/RecordMetadataDeserializer.java
pengpeng-lu: good point, changed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/UserDefinedFunctionCatalog.java
pengpeng-lu: reverted.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/UserDefinedFunctionCatalog.java
pengpeng-lu: 👍 The test passed after reverting this line.

pengpeng-lu: so we'll need to handle the case sensitivity somewhere, for example right now if I create a function with lower case, it'll be stored as upper case, but when I call the function with lower case, it'll only look for lower case, so it'll not be able to find the function. In the `SqlFunctionCatalogImpl.lookUpBuiltInFunction`, we do `final var functionValidator = builtInSynonyms.get(name.toLowerCase(Locale.ROOT));`. so that's why I think we should put something similar in the UserDefinedFunctionCatalog.lookup`. wdyt?

hatyo: If you revert all the changes in this file, and run the test `TemporaryFunctionTest.createTemporaryFunctionCaseSensitivityOption` it should pass if you're working on top of #3511.

Now, I see that the YAML test is failing, this is due to the way you're parsing the function name without consulting the case-sensitivity flag in the plan generator.

```java
    @Nonnull
    @Override
    public Expression visitUserDefinedScalarFunctionCall(@Nonnull RelationalParser.UserDefinedScalarFunctionCallContext ctx) {
        final var functionName = ctx.userDefinedScalarFunctionName().getText(); <---------
        Expressions arguments = visitFunctionArgs(ctx.functionArgs());
        return getDelegate().resolveFunction(functionName, arguments.asList().toArray(new Expression[0]));
    }
```

This is not correct, instead you should parse the function name with case-sensitivity checks, this is for example done in `visitUid`:

https://github.com/FoundationDB/fdb-record-layer/blob/6c0597e3ed753ef9cd8be124e456126d098c44bb/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/IdentifierVisitor.java#L73-L81

So, you can for example change the parser rules and make `userDefinedScalarFunctionName` delegate to `uid`:

```
userDefinedScalarFunctionName
    : uid
    ;
```

Then, in `visitUserDefinedScalarFunctionCall` make it delegate to to `visitUid`:

```java
    @Nonnull
    @Override
    public Expression visitUserDefinedScalarFunctionCall(@Nonnull RelationalParser.UserDefinedScalarFunctionCallContext ctx) {
        final var functionName = visitUid(ctx.userDefinedScalarFunctionName().uid());
        Expressions arguments = visitFunctionArgs(ctx.functionArgs());
        return getDelegate().resolveFunction(functionName.getName(), arguments.asList().toArray(new Expression[0]));
    }
```

That way, parsing the function name respects whatever case-sensitivity option set in the SQL session. The YAML test should pass.

(I haven't checked whether this is causing any ambiguity problems in the parser though, so you might want to double-check before refactoring according to the above).

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DelegatingVisitor.java
```diff
@@ -878,6 +873,12 @@ public Identifier visitFullId(@Nonnull RelationalParser.FullIdContext ctx) {
         return getDelegate().visitFullId(ctx);
     }
 
+    @Nonnull
+    @Override
+    public Identifier visitUserDefinedScalarFunctionStatementBody(@Nonnull RelationalParser.UserDefinedScalarFunctionStatementBodyContext ctx) {
+        return (Identifier)getDelegate().visitUserDefinedScalarFunctionStatementBody(ctx);
```
hatyo: The cast is incorrect, Please add the following signature:

```java
    @Nonnull
    @Override
    Identifier visitUserDefinedScalarFunctionStatementBody(@Nonnull RelationalParser.UserDefinedScalarFunctionStatementBodyContext ctx);
```

to `TypedVisitor` interface, also add `@Nonnull` to any missing implementation.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DelegatingVisitor.java
```diff
@@ -259,17 +259,12 @@ public ProceduralPlan visitDropTempFunction(final RelationalParser.DropTempFunct
...
-
-    @Override
-    public CompiledSqlFunction visitTempSqlInvokedFunction(final RelationalParser.TempSqlInvokedFunctionContext ctx) {
+    public CompilableSqlFunction visitTempSqlInvokedFunction(final RelationalParser.TempSqlInvokedFunctionContext ctx) {
         return getDelegate().visitTempSqlInvokedFunction(ctx);
     }
 
     @Override
-    public CompiledSqlFunction visitSqlInvokedFunction(@Nonnull RelationalParser.SqlInvokedFunctionContext ctx) {
+    public com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction visitSqlInvokedFunction(@Nonnull RelationalParser.SqlInvokedFunctionContext ctx) {
```
hatyo: Please do not use fully-qualified names.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/DelegatingVisitor.java
```diff
@@ -1208,6 +1209,18 @@ public Object visitIfNotExists(@Nonnull RelationalParser.IfNotExistsContext ctx)
...
 
+    @Nonnull
+    @Override
+    public Object visitUserDefinedScalarFunctionName(@Nonnull RelationalParser.UserDefinedScalarFunctionNameContext ctx) {
+        return getDelegate().visitUserDefinedScalarFunctionName(ctx);
+    }
+
+    @Nonnull
+    @Override
+    public Object visitUserDefinedScalarFunctionCall(@Nonnull RelationalParser.UserDefinedScalarFunctionCallContext ctx) {
```
hatyo: shouldn't this always return an `Expression`?

pengpeng-lu: yes, changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedScalarFunction.java
```diff
@@ -1,5 +1,5 @@
 /*
- * MacroFunction.java
+ * UserDefinedScalarFunction.java
```
normen662: This is a scalar function:
```
CREATE FUNCTION FOO(x INT) AS
   SELECT COUNT(a)
   FROM T
   WHERE x = 10
```

but it's not what this PR is about. I don't like the new name and vote for a name that reflects that the function is
1. scalar
2. it also expands in place
3. makes it clear it's not the kind of function I listed above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedMacroFunction.java
```diff
@@ -38,15 +38,15 @@
 import java.util.stream.Collectors;
 
 /**
- * MacroFunction that expands a body (referring to parameters) into a {@link Value} (through encapsulation) call site.
+ * UserDefinedScalarFunction that expands a body (referring to parameters) into a {@link Value} (through encapsulation) call site.
```
normen662: There are user-defined scalar functions that are not covered by this kind of function, i.e:
```
CREATE FUNCTION foo(p int) AS
  SELECT count(a)
  FROM T
  GROUP BY x
  WHERE x = p
```
These are regular SQL-bodied functions that do provably (or only at-runtime provably) return exactly one value.

I liked the name `MacroFunction` as it encapsulated the drop in/replace logic of these mini-functions. Can we rename it back? Or something else that clearly describes what this thing does?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedScalarFunction.java
```diff
@@ -100,18 +100,18 @@ public static MacroFunction fromProto(@Nonnull final PlanSerializationContext se
      * Deserializer.
      */
     @AutoService(PlanDeserializer.class)
-    public static class Deserializer implements PlanDeserializer<PMacroFunctionValue, MacroFunction> {
+    public static class Deserializer implements PlanDeserializer<PUserDefinedScalarFunctionValue, UserDefinedScalarFunction> {
```
normen662: Can we please refrain from using plan serializer logic here. Can you refactor this logic to express the dynamic dispatch in a common way that plan serialization and this logic + @hatyo 's view all use the same underlying logic?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -228,10 +228,14 @@ private void loadProtoExceptRecords(@Nonnull RecordMetaDataProto.MetaData metaDa
                 typeBuilder.setRecordTypeKey(LiteralKeyExpression.fromProtoValue(typeProto.getExplicitKey()));
             }
         }
+        PlanSerializationContext serializationContext = null;
         for (RecordMetaDataProto.PUserDefinedFunction function: metaDataProto.getUserDefinedFunctionsList()) {
+            if (serializationContext == null) {
```
normen662: This is not needed in a refactored world. Strictly speaking, the plan serialization context is necessary state such that repeated sub plans/values/types do not get serialized repeatedly. It a little bit like dictionary encoding. I think this is not needed here.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -250,7 +250,7 @@ message PValue {
     PCollateValue collate_value = 44;
     PNumericAggregationValue.PBitmapConstructAgg numeric_aggregation_value_bitmap_construct_agg = 45;
     PQuantifiedRecordValue quantified_record_value = 46;
-    PMacroFunctionValue macro_function_value = 47;
+    PUserDefinedScalarFunctionValue macro_function_value = 47;
```
normen662: `PUserDefinedScalarFunctionValue`s counterpart is not a `Value` so this should be defined here. Also the name should be `PUserDefinedScalarFunction` because that's its counterpart. This file should not be touched by this PR at all I think, it  should all go into metadata.

fdb-record-layer-core/src/main/proto/record_metadata.proto
```diff
@@ -200,7 +200,7 @@ message MetaData {
 
 message PUserDefinedFunction {
   oneof specific_function {
-    com.apple.foundationdb.record.planprotos.PMacroFunctionValue macro_function = 1;
+    com.apple.foundationdb.record.planprotos.PUserDefinedScalarFunctionValue user_defined_scalar_function = 1;
```
normen662: Please move that message definition to this file, rename it to `PUserDefinedScalarFunction`, see below.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedScalarFunctionTest.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * MacroFunctionTest.java
```
normen662:
```suggestion
 * UserDefinedScalaFunctionTest.java
```

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedScalarFunction.java
pengpeng-lu: I don't quite understand here, isn't `RawSqlFunction` using the same logic? let's sync offline.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/UserDefinedMacroFunction.java
```diff
@@ -20,33 +20,33 @@
...
+import com.apple.foundationdb.record.query.plan.serialization.DefaultPlanSerializationRegistry;
 
 import javax.annotation.Nonnull;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 /**
- * MacroFunction that expands a body (referring to parameters) into a {@link Value} (through encapsulation) call site.
+ * UserDefinedScalarFunction that expands a body (referring to parameters) into a {@link Value} (through encapsulation) call site.
```
normen662:
```suggestion
 * UserDefinedMacroFunction that expands a body (referring to parameters) into a {@link Value} (through encapsulation) call site.
```

PR #3046, opened 2025-01-15
Resolves #3045: Require FileOptions to be first block of yamsql

### LINKED ISSUE #3045: Only support FileOptions at start of yamsql file
To avoid having conflicting options, and to simplify, enforce what would probably be a convention anyways, and restrict FileOptions to be the first block of the file.

ScottDugas:
Followup from: #3031

When #3049 goes in we should add the following test to `SupportedVersionsTest` (or vice-versa):
```
# This test should fail because the options are too late
---
schema_template:
    create table t1(id bigint, col1 bigint, primary key(id))
---
options:
    supported_version: 3.0.17.0
---
test_block:
  tests:
    -
      - query: SELECT * FROM t1;
      - result: []
 ```

### CODE REVIEW

**Comment by ScottDugas** [comment]
```
OnlineIndexerBuildRankIndexTest$Unsafe > oneHundredElementsParallelOverlapRank(long) > [3] 251681697 FAILED
    java.util.concurrent.CompletionException: com.apple.foundationdb.synchronizedsession.SynchronizedSessionLockedException: Failed to initialize the session because of an existing session in progress
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
```

g31pranjal (approved):
LGTM! 
Are we still interested in the test that you mentioned in the PR description?

g31pranjal (approved):
LGTM!

PR #3047, opened 2025-01-15
Enable multi version tests

ohadzeliger:
Enable yaml tests that were disabled before, now that the gaps have been filled.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/RunExternalServerExtension.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/RunExternalServerExtension.java
@@ -101,11 +101,7 @@ public class RunExternalServerExtension implements BeforeAllCallback, AfterAllCa
         processBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
         processBuilder.redirectError(ProcessBuilder.Redirect.INHERIT);
 
-        this.serverProcess = processBuilder.start();
-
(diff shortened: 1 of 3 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/RunExternalServerExtension.java
```diff
@@ -87,7 +86,6 @@ public String getVersion() {
 
     @Override
     public void beforeAll(ExtensionContext context) throws Exception {
-        Assumptions.abort(); // Will be able to re-enable when we have a published external server to use here
```
ScottDugas, line 90: Given the issues with explain changing between versions, I'm not sure that we want to enable this as part of prb yet.

ScottDugas (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/RunExternalServerExtension.java
ohadzeliger: Do you mean in general, or just for the tests that contain explain assertions?

ScottDugas: The issue was resolved by https://github.com/FoundationDB/fdb-record-layer/pull/3055

ScottDugas (approved):

PR #3048, opened 2025-01-15
Remove string format

ScottDugas:


ScottDugas:
Replaced with: https://github.com/FoundationDB/fdb-record-layer/pull/3117

PR #3049, opened 2025-01-15
yamsql: Add supported_version to blocks and queries, and support for actual versions

ScottDugas:
This combines a few items that build on each other:
1. the supported_version option now supports old versions, not just `!current_version`
2. you can add `supported_version` to a single query
3. you can add `supported_version` to a test_block

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
@@ -33,19 +33,18 @@ import com.apple.foundationdb.relational.util.Environment;
 import com.apple.foundationdb.relational.yamltests.CustomYamlConstructor;
 import com.apple.foundationdb.relational.yamltests.Matchers;
 import com.apple.foundationdb.relational.yamltests.YamlExecutionContext;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Assertions;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 
 /**
  * A {@link QueryCommand} is one of the possible {@link Command} supported in the YAML testing framework that is used
@@ -114,6 +113,9 @@ public final class QueryCommand extends Command {
         }
         this.queryInterpreter = interpreter;
         this.queryConfigs = configs;
+        Assert.thatUnchecked(queryConfigs.stream()
+                .filter(config -> config instanceof QueryConfig.SkipConfig).count() < 2,
+                "Too many skips in a Query " + lineNumber);
     }
 
     public void execute(@Nonnull final RelationalConnection connection, boolean checkCache, @Nonnull QueryExecutor executor) {
@@ -144,6 +146,13 @@ public final class QueryCommand extends Command {
         Continuation continuation = null;
         int maxRows = 0;
         boolean exhausted = false;
+        final List<String> skips = queryConfigs.stream().filter(config -> config instanceof QueryConfig.SkipConfig)
+                .map(config -> ((QueryConfig.SkipConfig)config).getMessage())
+                .collect(Collectors.toList());
+        if (!skips.isEmpty()) {
+            logger.info(skips.stream().collect(Collectors.joining("\n")));
+            return;
+        }
         var queryConfigsIterator = queryConfigs.listIterator();
         while (queryConfigsIterator.hasNext()) {
             var queryConfig = queryConfigsIterator.next();
@@ -162,7 +171,7 @@ public final class QueryCommand extends Command {
                 // results
                 int finalMaxRows1 = maxRows;
                 runWithDebugger(() -> executor.execute(connection, null, queryConfig, checkCache, finalMaxRows1));
-            } else {
+            } else if (!QueryConfig.QUERY_CONFIG_SUPPORTED_VERSION.equals(queryConfig.getConfigName())) {
                 if (QueryConfig.QUERY_CONFIG_ERROR.equals(queryConfig.getConfigName())) {
                     Assert.that(!queryConfigsIterator.hasNext(), "ERROR config should be the last config specified.");
                 }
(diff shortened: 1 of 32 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -114,6 +113,9 @@ private QueryCommand(int lineNumber, @Nonnull QueryInterpreter interpreter, @Non
         }
         this.queryInterpreter = interpreter;
         this.queryConfigs = configs;
+        Assert.thatUnchecked(queryConfigs.stream()
+                .filter(config -> config instanceof QueryConfig.SkipConfig).count() < 2,
+                "Too many skips in a Query " + lineNumber);
```
g31pranjal, line 118: I might not understand it fully, but it looks like multiple things are happening at different places that are aimed towards validating if the `QueryCommand` should run or not. 

There is a `yamsql` "compile-time" check of the supported version (and logging) happening in `parse(...)`, and there is also a runtime check and logging happening during `execute(...)`. We can maybe avoid duplication by having a `NoOp` command, which is similar in essence to the `NoOpBlock` and `NoOpConfig`.

g31pranjal: There might also be some value in ensuring that the `supportedVersion` is the first config in the series of configs. Else, it gives a notion of configs that defined before the `supportedVersion` would run for all the versions while the ones defined after are version protected.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -144,6 +146,13 @@ private void executeInternal(@Nonnull final RelationalConnection connection, boo
         Continuation continuation = null;
         int maxRows = 0;
         boolean exhausted = false;
+        final List<String> skips = queryConfigs.stream().filter(config -> config instanceof QueryConfig.SkipConfig)
+                .map(config -> ((QueryConfig.SkipConfig)config).getMessage())
+                .collect(Collectors.toList());
+        if (!skips.isEmpty()) {
+            logger.info(skips.stream().collect(Collectors.joining("\n")));
+            return;
+        }
```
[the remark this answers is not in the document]
g31pranjal: Based on one of my other comments, if we are introducing a `NoOpCommand`, this thing can be completely removed from the `execute` code path.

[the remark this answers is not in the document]
g31pranjal: Also realized that the QueryCommand that has a `skipMessage` are being filtered out and not actually being executed. So, this code should not be reached anyways. We might as well replace it with an `Assert` on non-existence of `SkipConfig`

g31pranjal (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/SkippedCommand.java
```diff
@@ -0,0 +1,60 @@
...
+import com.apple.foundationdb.relational.yamltests.YamlExecutionContext;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.annotation.Nonnull;
+import java.sql.SQLException;
+
+/**
+ * A command that should be skipped by the block.
+ */
```
g31pranjal, line 35: Probably add a line here that this Command is never executed. It is simply a placeholder that denotes that the command should not be included in execution.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -342,6 +345,19 @@ void checkResultInternal(@Nonnull Object actual, @Nonnull String queryDescriptio
...
 
+    private static QueryConfig getSupportedVersionConfig(final Object value, final int lineNumber, final YamlExecutionContext executionContext) {
+        final SupportedVersionCheck check = SupportedVersionCheck.parse(value, executionContext);
+        if (!check.isSupported()) {
+            return new SkipConfig(QUERY_CONFIG_SUPPORTED_VERSION, value, lineNumber, executionContext, check.getMessage());
+        }
+        return new QueryConfig(QUERY_CONFIG_SUPPORTED_VERSION, value, lineNumber, executionContext) {
+            @Override
+            void checkResultInternal(@Nonnull final Object actual, @Nonnull final String queryDescription) throws SQLException {
+                // Nothing to do, this query is supported
```
g31pranjal, line 356: This should probably be through a `RuntimeException` - which could ensure that this config is never run for a query. It's mostly precautionary.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -374,4 +392,22 @@ public static QueryConfig parse(@Nonnull Object object, @Nonnull YamlExecutionCo
...
+        private final String message;
+
+        public SkipConfig(final String configMap, final Object value, final int lineNumber, final YamlExecutionContext executionContext, final String message) {
+            super(configMap, value, lineNumber, executionContext);
+            this.message = message;
+        }
+
+        @Override
+        void checkResultInternal(@Nonnull final Object actual, @Nonnull final String queryDescription) throws SQLException {
+            logger.info("Line: " + getLineNumber() + " " + message);
```
g31pranjal, line 406: Same.

This should probably be through a `RuntimeException` - which will ensure that this config is never run for a query. It's mostly precautionary.

g31pranjal (changes requested):
I have a few minor comments that is pretty straightforward to address.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/SkippedCommand.java
ohadzeliger: Done.

yaml-tests/src/test/java/SupportedVersionTest.java
```diff
@@ -0,0 +1,96 @@
...
+    YamlRunner.YamlConnectionFactory createConnectionFactory() {
+        return new YamlRunner.YamlConnectionFactory() {
+            @Override
+            public RelationalConnection getNewConnection(@Nonnull URI connectPath) throws SQLException {
+                return DriverManager.getConnection(connectPath.toString()).unwrap(RelationalConnection.class);
+            }
+
+            @Override
+            public Set<String> getVersionsUnderTest() {
+                return Set.of("3.0.18.0");
```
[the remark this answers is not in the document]
ohadzeliger: Created a ticket

g31pranjal (approved):

PR #3055, opened 2025-01-16
Skip explain when running in mixed mode

ohadzeliger:


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -146,6 +146,15 @@ void checkErrorInternal(@Nonnull SQLException e, @Nonnull String queryDescriptio
 
     }
 
+    /**
+     * Allow the config to control whether it should be executed or not.
+     * The executor will skip the execution and validation if the config returns false.
+     * @return true if the config should go on for execution, false if not.
+     */
+    boolean shouldExecute() {
```
ohadzeliger, line 154: Note that this is similar (and can be converted to) the `SkipConfig` from #3049, once both Prs are merged in.

g31pranjal: Using `SkipConfig` is probably a good idea for this use-case as well. That way, we can get rid of `shouldExecute` altogether.  

We can probably  wait for #3049 to get in and then rebase this one on top of that.

g31pranjal: On the second though, I believe `SkipConfig` has slightly different semantics. If there is a `SkipConfig` we do not execute any of the other configs that are available for the query. More like a "break" statement. 

However, in this case, we need some `NoOpConfig` that is there but just not executed. It could be the same anonymous specialization of `QueryConfig` that we have for the case when a version is supported in #3049

g31pranjal (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Agree. I can add that instead of the additional method to the interface if you think that's more appropriate. The one downside I can see for it is that changing the config type we will lose the content, but for now I think we can live t=with that.

ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -361,6 +361,23 @@ void checkResultInternal(@Nonnull final Object actual, @Nonnull final String que
...
+     * @param lineNumber the line number in the test file
+     * @param executionContext the execution context for the test
+     * @return an instance of a NoOp config
+     */
+    public static QueryConfig getNoOpConfig(int lineNumber, @Nonnull YamlExecutionContext executionContext) {
+        return new QueryConfig(QUERY_CONFIG_NO_OP, null, lineNumber, executionContext) {
+            @SuppressWarnings("PMD.CloseResource") // lifetime of autocloseable persists beyond method
+            @Override
+            void checkResultInternal(@Nonnull Object actual, @Nonnull String queryDescription) throws SQLException {
+            }
```
g31pranjal, line 377: minor: Add assertion about it not executing

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -361,6 +361,23 @@ void checkResultInternal(@Nonnull final Object actual, @Nonnull final String que
         };
```
g31pranjal, line 361: mildly related: `getSupportedVersionConfig` can now be refactored to avoid anonymous class definition:  
```
    private static QueryConfig getSupportedVersionConfig(final Object value, final int lineNumber, final YamlExecutionContext executionContext) {
        final SupportedVersionCheck check = SupportedVersionCheck.parse(value, executionContext);
        if (!check.isSupported()) {
            return new SkipConfig(QUERY_CONFIG_SUPPORTED_VERSION, value, lineNumber, executionContext, check.getMessage());
        }
        return getNoOpConfig(lineNumber, executionContext);
   }
```

g31pranjal (approved):
minor comments. LGTM!

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Sure, will do in a subsequent PR.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
ohadzeliger: Sure, will do in a subsequent PR.

ohadzeliger:
Will make the two requested changes in a subsequent PR since we'd like this change to be merged - it will allow us to enable mixed-mode testing!

PR #3060, opened 2025-01-23
Break out a helper class from GenerateVisitor annotation processor.

MMcM:
This allows service loading all annotation processors and even determining that this one does not apply, without having javapoet in the classpath.


alecgrieser (approved):
This looks good to me. The underlying error is a bit esoteric, but I think the idea that we should be able to construct the `Processor` to at least determine which annotations it processes without requiring external dependencies seems pretty defensible. We may also consider taking this on to the 4.0.559 branch

alecgrieser (approved):

PR #3064, opened 2025-01-27
make explain return planner metrics as separate column

normen662:
```
 RelationalStructMetaData(
  FieldDescription.primitive("TASK_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("TASK_TOTAL_TIME_NS", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("TRANSFORM_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("TRANSFORM_TIME_NS", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("TRANSFORM_YIELD_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("INSERT_TIME_NS", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("INSERT_NEW_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls),
  FieldDescription.primitive("INSERT_REUSED_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls), ...)
```

Most of the PR is just making sure that the `Debugger` stats structures become accessible to the code that prepares and packages the result that is sent to the client. For that purpose I extracted `Stats` from the respective `Debugger` implementations and make it is own first-class citizen. Similar goes for `StatsMaps`. One thing I wanted to make sure is that once the `StatsMaps` are accessed outside of the installed `Debugger`, the `StatsMaps` become immutable. For that purpose we make a copy, as usual, however, we only want to make that copy when the client of the planner (i.e. QueryPlan) needs the stats (when responding to an `EXPLAIN`).

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
@@ -318,6 +318,9 @@ public class CascadesPlanner implements QueryPlanner {
                 .put(QueryPlanInfoKeys.TOTAL_TASK_COUNT, taskCount)
                 .put(QueryPlanInfoKeys.MAX_TASK_QUEUE_SIZE, maxQueueSize)
                 .put(QueryPlanInfoKeys.CONSTRAINTS, constraints)
+                .put(QueryPlanInfoKeys.STATS_MAPS,
+                        Debugger.getDebuggerMaybe().flatMap(Debugger::getStatsMaps)
+                                .orElse(null))
                 .build();
         return new QueryPlanResult(plan, info);
     }
@@ -354,7 +357,13 @@ public class CascadesPlanner implements QueryPlanner {
                     evaluationContext);
             final var plan = resultOrFail();
             final var constraints = QueryPlanConstraint.collectConstraints(plan);
-            return new QueryPlanResult(plan, QueryPlanInfo.newBuilder().put(QueryPlanInfoKeys.CONSTRAINTS, constraints).build());
+            return new QueryPlanResult(plan,
+                    QueryPlanInfo.newBuilder()
+                            .put(QueryPlanInfoKeys.CONSTRAINTS, constraints)
+                            .put(QueryPlanInfoKeys.STATS_MAPS,
+                                    Debugger.getDebuggerMaybe()
+                                            .flatMap(Debugger::getStatsMaps).orElse(null))
+                            .build());
         } finally {
             Debugger.withDebugger(Debugger::onDone);
         }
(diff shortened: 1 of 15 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
```diff
@@ -318,6 +318,9 @@ public QueryPlanResult planQuery(@Nonnull final RecordQuery query, @Nonnull Para
                 .put(QueryPlanInfoKeys.TOTAL_TASK_COUNT, taskCount)
                 .put(QueryPlanInfoKeys.MAX_TASK_QUEUE_SIZE, maxQueueSize)
                 .put(QueryPlanInfoKeys.CONSTRAINTS, constraints)
+                .put(QueryPlanInfoKeys.STATS_MAPS,
+                        Debugger.getDebuggerMaybe().flatMap(Debugger::getStatsMaps)
```
alecgrieser, line 322: I take it that the query plan result won't be returned if the `STATS_MAP` is empty if the debugger isn't present? Maybe that's fine, as this information is a little internal. It does potentially limit the ability to get these stats through the relational server, but maybe we don't care about that (at least for the moment)

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -273,12 +282,23 @@ private RelationalResultSet executeExplain(@Nonnull ContinuationImpl parsedConti
                     FieldDescription.primitive("VERSION", Types.INTEGER, DatabaseMetaData.columnNoNulls),
                     FieldDescription.primitive("PLAN_HASH_MODE", Types.VARCHAR, DatabaseMetaData.columnNoNulls)
             );
+            StructMetaData plannerMetricsMetadata = new RelationalStructMetaData(
+                    FieldDescription.primitive("TASK_COUNT", Types.BIGINT, DatabaseMetaData.columnNoNulls),
```
alecgrieser, line 286: Do you know if we have coverage of what happens if there isn't a debugger set (e.g., making sure that we don't return a `null` column here)? This looks right to me, but I just wanted to see if it's covered.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/ExplainTests.java
```diff
@@ -67,8 +67,8 @@ public ExplainTests() {
 
     @Test
     void explainResultSetMetadataTest() throws Exception {
-        final var expectedLabels = List.of("PLAN", "PLAN_HASH", "PLAN_DOT", "PLAN_GML", "PLAN_CONTINUATION");
-        final var expectedTypes = List.of(Types.VARCHAR, Types.INTEGER, Types.VARCHAR, Types.VARCHAR, Types.STRUCT);
+        final var expectedLabels = List.of("PLAN", "PLAN_HASH", "PLAN_DOT", "PLAN_GML", "PLAN_CONTINUATION", "PLANNER_METRICS");
+        final var expectedTypes = List.of(Types.VARCHAR, Types.INTEGER, Types.VARCHAR, Types.VARCHAR, Types.STRUCT, Types.STRUCT);
```
alecgrieser, line 71: I'm not sure it needs to be here, but it would be nice to have a test somewhere that asserted on the values of the planner stats struct, even if it was just making sure the values weren't null or zero. Getting a more sophisticated test set of assertions is probably for another time.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -234,6 +235,12 @@ void checkResultInternal(@Nonnull Object actual, @Nonnull String queryDescriptio
                         reportTestFailure(diffMessage);
                     }
                 }
+
+                final var plannerMetrics = resultSet.getStruct(6);
+                if (plannerMetrics != null) {
```
alecgrieser, line 240: Should the `plannerMetrics` ever be `null`? I'd think that it should always be non-null except maybe for mixed mode tests where the column wouldn't be defined (and I guess must not be running here because it's happening in the `explain` step)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
normen662: That's right.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
normen662: There are plenty of tests where we don't have a `Debugger` installed (all that's not EXPLAIN) and this is `null`

alecgrieser (approved):

PR #3066, opened 2025-01-27
Auto update supported_version in yamsql files

ScottDugas:
This adds the necessary tooling to update the `.yamsql` files, replacing `!current_version` with the version being released. I validated that it does the replacement correctly, locally. In order to support having a comment about `!current_version` in `showcasing-tests.yamsql`, I have it leave it alone if it is immediately surrounded by quotes. This seemed like a good compromise that avoided me having to get too much into yaml parsing, although something about whether the whole line is a comment could also work.

To be extra safe, the commit script validates that it's not trying to update anything other than `.yamsql` files.

I thought about adding the commit code to the gradle build also, but I couldn't figure out a way to have the dependencies be enforced so that committing always ran after all the subprojects updated their `.yamsql` files, so I went with a separate  python script, similar to how we update the release notes. I don't think it would be hard to do the commit from gradle directly.

### DIFF
--- a/build/commit_yamsql_updates.py
+++ b/build/commit_yamsql_updates.py
@@ -5,7 +5,7 @@
 #
 # This source file is part of the FoundationDB open source project
 #
-# Copyright 2015-2018 Apple Inc. and the FoundationDB project authors
+# Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/build/commit_yamsql_updates.py
+++ b/build/commit_yamsql_updates.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+#
+# commit_yamsql_updates.py
+#
+# This source file is part of the FoundationDB open source project
+#
+# Copyright 2015-2018 Apple Inc. and the FoundationDB project authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


alecgrieser (approved):

yaml-tests/src/test/resources/showcasing-tests.yamsql
```diff
@@ -39,7 +39,8 @@
 options:
     # supported_version specifies what versions this works with
     # Any version greater than the provided version is expected to work
-    # When developing a new feature use !current_version to specify that it only works with the current version
+    # When developing a new feature use "!current_version" (but without the quotes) to specify that it only works with
+    # the current version
```
ohadzeliger, line 43: Maybe add "'!current_version` will be replaced by the build system once a new release is being built"?
Also maybe "tests will be skipped when a version of the server exists that is not supported"?

build/commit_yamsql_updates.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+#
+# commit_yamsql_updates.py
+#
+# This source file is part of the FoundationDB open source project
+#
+# Copyright 2015-2018 Apple Inc. and the FoundationDB project authors
```
ohadzeliger, line 8:
```suggestion
# Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```

build/commit_yamsql_updates.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# Utility script for committing the updates to yamsql as part of release.
+#
+# As part of a release we should update the *.yamsql files, replacing !current_version
+# with the version being released. This is done via a gradle task: updateYamsql
+# This script will check that there are no extraneous changes staged, and commit
+# the updates to the yamsql files.
+#
+
+import argparse
```
ohadzeliger, line 32: Is this meant to be run by CI?

ohadzeliger (changes requested):

build/commit_yamsql_updates.py
ohadzeliger: Yes - saw the changes in another PR.

ohadzeliger (approved):

PR #3071, opened 2025-01-28
Add some GitHub workflow actions to begin CI/CD transition there

MMcM:
See https://github.com/MMcM/fdb-record-layer/actions/workflows/pull_request.yml for examples of running a pull request against a branch that has these actions (that is, against this branch).


arnaud-lacurie (dismissed):

arnaud-lacurie (dismissed):

arnaud-lacurie (approved):

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,57 @@
...
+          fdb-record-layer-lucene/.out/test-results/**/TEST-*.xml
+          fdb-record-layer-jmh/.out/test-results/**/TEST-*.xml
+          examples/.out/test-results/**/TEST-*.xml
+          fdb-relational-api/.out/test-results/**/TEST-*.xml
+          fdb-relational-core/.out/test-results/**/TEST-*.xml
+          fdb-relational-cli/.out/test-results/**/TEST-*.xml
+          fdb-relational-grpc/.out/test-results/**/TEST-*.xml
+          fdb-relational-jdbc/.out/test-results/**/TEST-*.xml
+          fdb-relational-server/.out/test-results/**/TEST-*.xml
+          yaml-tests/.out/test-results/**/TEST-*.xml
```
alecgrieser, line 57: Is there a way to glob all of these without having to specify the sub-projects? Something like `**/.out/test-results/**/TEST-*.xml`? If we had more subprojects, it would be nice not to have to futz with these files quite so much

alecgrieser (approved):

.github/workflows/nightly.yml
MMcM: I think the dot-filename `.out` confuses `glob`; at least none of the simple things I tried worked.

arnaud-lacurie: We could change `.out` to `build` in a subsequent PR if that helps.

PR #3072, opened 2025-01-29
WIP - Start calculating inner's first item concurrently.

hatyo:
- this could improve the performance of the pipeline now that it schedule up to N-calculations of the inner's first cursor item, where N is the size of the pipeline.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FlatMapPipelinedCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FlatMapPipelinedCursor.java
@@ -267,6 +267,7 @@ public class FlatMapPipelinedCursor<T, V> implements RecordCursor<V> {
         if (closed) {
             pipelineQueueEntry.close();
         }
+        pipelineQueueEntry.getNextInnerPipelineFuture();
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/FlatMapPipelinedCursor.java
```diff
@@ -267,6 +267,7 @@ private synchronized void addEntryToPipeline(PipelineQueueEntry pipelineQueueEnt
         if (closed) {
             pipelineQueueEntry.close();
         }
+        pipelineQueueEntry.getNextInnerPipelineFuture();
```
alecgrieser, line 270: Do we need both this change and the change in the `PipelineQueueEntry` constructor? If I'm reasoning about this correctly, it seems like we only need one of the two (that this is extraneous if the next future is fired off during the constructor, and the constructor call is extraneous if the future is created here)

hatyo: Correct, this was an oversight, let me remove that.

alecgrieser (approved):

PR #3074, opened 2025-01-29
Resolves #3073: Add enum column support to relational server

### LINKED ISSUE #3073: Add support for enum columns to the relational server
We should add support for columns that are enums to the server. These have `Types.OTHER`, and contain a string.

ScottDugas:


fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalResultSetFacade.java
```diff
@@ -320,6 +320,11 @@ public Object getObject(int oneBasedColumn) throws SQLException {
                 int index = PositionalIndex.toProtobuf(oneBasedColumn);
                 Column column = this.delegate.getRow(rowIndex).getColumns().getColumn(index);
                 return column == null || !column.hasBinary() ? null : column.getBinary().toByteArray();
+            case Types.OTHER:
+                // Probably an enum, it's not clear exactly how we should handle this, but we currently only have one
+                // thing which appears as OTHER
+                o = getString(oneBasedColumn);
```
g31pranjal, line 326: I think this is mostly in line as to what we have in other parts of the system (in DirectAccessAPI etc.) but we should formally decide as to how we want to treat our ENUMs. I think most of the DBs actually map it to either a STRING or  INTEGER externally and do an implicit cast internally, but they loose some bits of type info in the way to do so.

g31pranjal (approved):
This was much needed since a long time! Thanks for tackling this.
Also, sorry for the delay

g31pranjal (approved):

PR #3075, opened 2025-01-29
Set maxrows to 1

ohadzeliger:


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -170,6 +190,62 @@ private Continuation executeContinuation(@Nonnull RelationalConnection connectio
...
+                RelationalResultSetMetaData metadata = resultSet.getMetaData(); // The first metadata will be used for all
+                boolean hasResult = resultSet.next(); // Initialize result set value retrieval. Has only one row.
+                // Edge case: when there are no results at all, we don't want to store the result set at all,
+                // to create an empty aggregated result set
+                if (hasResult) {
+                    results.add(resultSet);
+                }
+                // Have continuations - keep running the query
+                Continuation continuation = resultSet.getContinuation();
+                while (!continuation.atEnd()) {
```
ScottDugas, line 225: It seems like something here should assert that the number returned by each individual result is `<= FORCED_MAX_ROWS`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -170,6 +190,62 @@ private Continuation executeContinuation(@Nonnull RelationalConnection connectio
...
+        return s;
+    }
+
+    private Object executeStatementAndForceContinuations(@Nonnull Statement s, @Nullable String queryString,
+                                                         final RelationalConnection connection, @Nullable Integer maxRows) throws SQLException {
+        // Check if we need to force continuations
+        if ((maxRows == null) && forceContinuations) {
+            s.setMaxRows(FORCED_MAX_ROWS);
+            Object result = executeStatement(s, queryString);
+            if (result instanceof RelationalResultSet) {
```
ScottDugas, line 212: This method is rather complex, perhaps a method for the contents of this `if` would be helpful.

yaml-tests/src/test/java/MultiServerJDBCYamlTests.java
```diff
@@ -69,6 +73,165 @@ public MultiServerInitialConnectionExternalTests() {
...
+            super(1);
+        }
+
+        @Override
+        public Map<String, Object> getAdditionalOptions() {
+            return Map.of(OPTION_FORCE_CONTINUATIONS, true);
+        }
+
+        @Override
+        @Disabled("Infinite continuation loop")
```
ScottDugas, line 91: Instead of just comments, should these reference issues?

yaml-tests/src/test/java/MultiServerJDBCYamlTests.java
```diff
@@ -30,8 +31,11 @@
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
+import static com.apple.foundationdb.relational.yamltests.YamlExecutionContext.OPTION_FORCE_CONTINUATIONS;
```
ScottDugas, line 37: Given this is only used once, I _feel_ like it would be better to not have a static import for it.
But I also suggested removing this in favor of an object you can call methods on.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AggregateResultSet.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.apple.foundationdb.relational.api.Continuation;
+import com.apple.foundationdb.relational.api.RelationalResultSet;
+import com.apple.foundationdb.relational.api.RelationalResultSetMetaData;
+
+import javax.annotation.Nonnull;
+import java.sql.SQLException;
+import java.util.Iterator;
+
+import static com.apple.foundationdb.relational.api.exceptions.ErrorCode.UNSUPPORTED_OPERATION;
```
ScottDugas, line 31: Since this is only used once, I don't think having a static import improves readability.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AbstractAggregateResultSet.java
```diff
@@ -0,0 +1,202 @@
...
+import com.apple.foundationdb.relational.api.RelationalResultSetMetaData;
+import com.apple.foundationdb.relational.api.RelationalStruct;
+import com.apple.foundationdb.relational.api.exceptions.ErrorCode;
+
+import java.sql.SQLException;
+
+/**
+ * A result set implementation that aggregates many result sets into one, where each internal result set acts as a row
+ * in the aggregate.
+ * TODO: This is a copy of the AbstractMockRssultSet. We should clear that tech debt by integrating the types
```
ScottDugas, line 34:
```suggestion
 * TODO: This is a copy of the AbstractMockResultSet. We should clear that tech debt by integrating the types
```

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AbstractAggregateResultSet.java
```diff
@@ -0,0 +1,202 @@
...
+        return metadata;
+    }
+
+    @Override
+    public boolean isClosed() throws SQLException {
+        return isClosed;
+    }
+
+    private void checkCurrentRow() throws SQLException {
+        if (currentRow == null) {
```
ScottDugas, line 198: Should this check `isClosed` at all?

ScottDugas (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
```diff
@@ -36,11 +36,13 @@
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
 
 @SuppressWarnings({"PMD.GuardLogStatement"}) // It already is, but PMD is confused and reporting error in unrelated locations.
 public final class YamlExecutionContext {
+    public static final String OPTION_FORCE_CONTINUATIONS = "optionForceContinuations";
```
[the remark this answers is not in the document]
ohadzeliger: There is a pattern that we can use to enforce types by creating a generic-types enum constants. Let me see if it creates a better experience.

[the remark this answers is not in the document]
ohadzeliger: I'll get that done in a followup in an attempt to get this PR in sooner to get coverage increase

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -65,14 +71,36 @@ public class QueryExecutor {
...
+     */
+    QueryExecutor(@Nonnull String query, int lineNumber, @Nullable List<Parameter> parameters, boolean forceContinuations) {
         this.lineNumber = lineNumber;
         this.query = query;
         this.parameters = parameters;
+        this.forceContinuations = forceContinuations;
     }
 
+    /**
+     * Execute tehe query.
```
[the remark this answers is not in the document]
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -65,14 +71,36 @@ public class QueryExecutor {
...
     QueryExecutor(@Nonnull String query, int lineNumber, @Nullable List<Parameter> parameters) {
+        this(query, lineNumber, parameters, false);
+    }
+
+    /**
+     * Constructor.
+     * @param query the query to execute
+     * @param lineNumber the line number from the test script
+     * @param parameters the parameters to bind
+     * @param forceContinuations Whether to force continuations in case the query does not explicitly specify a limit.
```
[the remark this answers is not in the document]
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -170,6 +190,62 @@ private Continuation executeContinuation(@Nonnull RelationalConnection connectio
         return continuationAfter;
     }
 
+    private RelationalPreparedStatement prepareContinuationStatement(@Nonnull RelationalConnection connection,
+                                                                     @Nonnull Continuation continuation,
+                                                                     @Nullable Integer maxRows) throws SQLException {
+        var s = connection.prepareStatement("EXECUTE CONTINUATION ?;");
+        s.setMaxRows((maxRows != null) ? maxRows : DEFAULT_MAX_ROWS);
```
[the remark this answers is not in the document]
ohadzeliger: Done. This makes more sense and actually gets us closer to what production code may do

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -170,6 +190,62 @@ private Continuation executeContinuation(@Nonnull RelationalConnection connectio
...
+                }
+                // Use first metadata for the aggregated result set as they are all the same
+                // Use last continuation
+                return new AggregateResultSet(metadata, continuation, results.iterator());
+            } else {
+                // non-result set - just return
+                return result;
+            }
+        } else {
+            s.setMaxRows((maxRows != null) ? maxRows : DEFAULT_MAX_ROWS);
```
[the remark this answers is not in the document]
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ohadzeliger: That is a good addition. Unfortunately, we can't do It at the result level (here) because of the statefulness of the result set interface. I tried to do that within the aggregate result set but since we can't control who calls `next()` on the individual rows, we can't assert on their status. Ideas welcome.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ohadzeliger: Done.

yaml-tests/src/test/java/MultiServerJDBCYamlTests.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AggregateResultSet.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -203,46 +204,53 @@ private RelationalPreparedStatement prepareContinuationStatement(@Nonnull Relati
...
+        if (result instanceof RelationalResultSet) {
+            @SuppressWarnings("PMD.CloseResource")
+            RelationalResultSet resultSet = (RelationalResultSet)result;
+            List<RelationalResultSet> results = new ArrayList<>();
+            RelationalResultSetMetaData metadata = resultSet.getMetaData(); // The first metadata will be used for all
+
+            boolean hasResult = resultSet.next(); // Initialize result set value retrieval. Has only one row.
+            // Edge case: when there are no results at all, we don't want to store the result set at all,
+            // to create an empty aggregated result set
+            if (hasResult) {
```
ScottDugas, line 233: Should this return the `resultSet` if `!hasResult` ? Rather than returning an empty `AggregateResultSet`?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -203,46 +204,53 @@ private RelationalPreparedStatement prepareContinuationStatement(@Nonnull Relati
...
+            }
+            // Have continuations - keep running the query
+            Continuation continuation = resultSet.getContinuation();
+            while (!continuation.atEnd()) {
+                try (var s2 = prepareContinuationStatement(connection, continuation, FORCED_MAX_ROWS)) {
+                    resultSet = (RelationalResultSet)executeStatement(s2, null);
+                    resultSet.next(); // Initialize result set value retrieval. Has only one row.
+                    continuation = resultSet.getContinuation();
+                    if (!continuation.atEnd()) {
+                        results.add(resultSet);
```
ScottDugas, line 244: IIUC if the total rows is an even multiple of the maxRows, the last resultSet is always an empty resultSet, and that's why this works.
But if that weren't the case, this would always be missing the last value.

Is it possible to have a ResultSet with a single row, and `getContinuation()` returns an end continuation? Such that we would drop that last row? We could have a bug where when maxrows=1 we always return the last row twice, which would be hidden by this flow.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AggregateResultSet.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    protected boolean hasNext() {
+        return rowIterator.hasNext();
+    }
+
+    @Override
+    protected RelationalResultSet advanceRow() throws SQLException {
+        if (!hasNext()) {
+            return null;
+        }
```
ScottDugas, line 58: I believe this is already assuming that the underlying result sets only have one row, so if you store the last row in a field, and right here do:
```
Assertions.assertFalse(lastRow.next());
```
That would assert that every underlynig resultSet only has one row.

ScottDugas (changes requested):

yaml-tests/src/test/java/MultiServerJDBCYamlTests.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ohadzeliger: Sure, that would work.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ohadzeliger: There are several reasons as to why this feature only works with `maxRows:1`, among them also the fact that the `AggregateResultSet` cannot advance to the last result of the result set in order to get a continuation without irrevocably changing the state of the result set. 
Added an assertion that would alleviate the above issue, but I think that we would need a more elaborate scheme to get this to work with maxRows!=1 (that was the original intent, but it turned more complex than originally thought)

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/AggregateResultSet.java
ohadzeliger: Actually, NVM, this would work if we assume size of 1 and account for the state before we call next() for the first time.

ScottDugas (approved):

PR #3078, opened 2025-01-30
Update to sphinx documentation

alecgrieser:
This updates our website to use sphinx for its documentation site. This includes some new content around the SQL API, as well as our API Javadoc, which we will be publishing ourselves for the first time in a while. Publishing happens at the end of the release process (which I've been able to test by first applying these changes to a test repo of mine).


actions/release-build-publish/action.yml
```diff
@@ -51,8 +51,10 @@ runs:
       id: push_updates
       shell: bash
       run: git push origin
+      # Continue the build (including downstream steps). If the push fails, we'll create a PR
+      continue-on-error: true
     - name: Create Merge PR if conflict
-      if: failure() && steps.push_updates.conclusion == 'failure'
+      if: success() && steps.push_updates.outcome == 'failure'
```
hatyo, line 57: Why is this a `success`?

alecgrieser: It's because of the `continue-on-error` change to the `git push` step above. With that set, then even if the actual push fails, the whole job is marked as successful. We want that behavior, because that way, we continue on to push the documentation in the main action. But we only want to do that if the rest of the build has succeeded or if we've either pushed successfully or created a PR successfully.

So the behavior we want on this line is "if we have been successful so far (i.e., not been cancelled and not had any step fail) _but_ if the `git push` failed, then create a merge PR". FWIW, this is also why we have `.conclusion` here instead of `.outcome`. The `.conclusion` value reflects the success state _after_ `continue-on-error`, whereas `.outcome` reflects the success state _before_ `continue-on-error`. So, if the push was actually successful, we skip PR creation; if the push failed (but everything else succeeded), we create the PR

arnaud-lacurie (approved):

PR #3079, opened 2025-01-31
Fixes for Gradle tasks for Antlr

hatyo:
This adds two fixes:

- Puts the generated files in a directory that corresponds to their set package.
- Creates a post-`generateGrammarSource` task which copies all `*.token` files next to the grammar and lexer files in `src/main/antlr`. These files are added to `.gitignore` to avoid checking them in by mistake.

These effectively enable the IDE to correctly parse and present the grammar files as well as the generated files, before it was showing hundreds of errors due to file location mismatch and the inability to link the grammar file with the respective tokens array.

### DIFF
--- a/gradle/antlr.gradle
+++ b/gradle/antlr.gradle
@@ -21,6 +21,7 @@
 apply plugin: 'antlr'
 
 generateGrammarSource {
+    outputDirectory = layout.buildDirectory.dir('generated-src/antlr/main/com/apple/foundationdb/relational/generated').get().asFile
     maxHeapSize = "128m"
     arguments += ['-package','com.apple.foundationdb.relational.generated', '-listener','-visitor', '-long-messages']
 }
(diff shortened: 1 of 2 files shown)


gradle/antlr.gradle
```diff
@@ -29,3 +30,14 @@ sourceSets.configureEach {
...
+generateGrammarSource.doLast {
+    final source = layout.buildDirectory.dir('generated-src/antlr/main/com/apple/foundationdb/relational/generated/')
+    final tokensFile = "*.tokens"
+    final dest = layout.projectDirectory.dir('src/main/antlr')
+    copy {
+        from source
+        include tokensFile
+        into dest
+    }
+}
```
alecgrieser, line 43: It would be nice if `clean` could be updated to delete the generated `tokens` files as well. I think you can do that with a new task named something like `cleanGeneratedTokens` of type `Delete` that deletes the token files, and then `clean` can be made to depend on that task.

gradle/antlr.gradle
```diff
@@ -21,6 +21,7 @@
 apply plugin: 'antlr'
 
 generateGrammarSource {
+    outputDirectory = layout.buildDirectory.dir('generated-src/antlr/main/com/apple/foundationdb/relational/generated').get().asFile
```
alecgrieser, line 24: We'll have to think at some point how to make this more generic so that it accepts different package names, though for now, this should be fine. Something we will have to notice if we ever change the hardcoded package below

alecgrieser (approved):
I left one comment about cleaning these generated tokens, but this is also clearly a win, and I'm merging this one and taking a change to clean up the tokens files as a follow up

gradle/antlr.gradle
hatyo: Good idea, done.

gradle/antlr.gradle
hatyo: Yeah, let's cross that bridge when we reach it then.

alecgrieser (approved):

PR #3080, opened 2025-01-31
Update relase workflow to accept artifact version

alecgrieser:
This updates the release workflow so that it runs when someone manually triggers it rather than on every commit. It also takes an argument for the version, which hopefully will fix some tests that are failing during the current build.

It still doesn't update the build to do any of the other things we associate with releases (e.g., updating any of the values storing the current version), but it may give us insight into the release build itself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,11 @@
 name: Release
 
 on:
-  push:
-    branches:
-      - 'main'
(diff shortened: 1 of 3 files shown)


alecgrieser:
I think we'll want to go with something like #3082 instead. If we do, I'll close this one out.

alecgrieser:
Closing in favor of #3082

PR #3081, opened 2025-01-31
Implement global (within factory) state for alternating connections.

ohadzeliger:
This fixes an issue with the connection factory where YAML tests with a single statement always ended up using the same connection number.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
@@ -36,6 +36,7 @@ import java.sql.Struct;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -59,11 +60,12 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
 
     @Nonnull
     private final ConnectionSelectionPolicy connectionSelectionPolicy;
-    private final int initialConnection;
     @Nonnull
     private final YamlRunner.YamlConnectionFactory defaultFactory;
     @Nonnull
     private final List<YamlRunner.YamlConnectionFactory> alternateFactories;
+    @Nonnull
+    private final AtomicInteger currentConnection;
 
     public MultiServerConnectionFactory(@Nonnull final YamlRunner.YamlConnectionFactory defaultFactory,
                                         @Nonnull final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
@@ -75,18 +77,18 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
                                         @Nonnull final YamlRunner.YamlConnectionFactory defaultFactory,
                                         @Nonnull final List<YamlRunner.YamlConnectionFactory> alternateFactories) {
         this.connectionSelectionPolicy = connectionSelectionPolicy;
-        this.initialConnection = initialConnection;
         this.defaultFactory = defaultFactory;
         this.alternateFactories = alternateFactories;
         this.versionsUnderTest =
                 Stream.concat(Stream.of(defaultFactory), alternateFactories.stream())
                         .flatMap(yamlConnectionFactory -> yamlConnectionFactory.getVersionsUnderTest().stream())
                         .collect(Collectors.toSet());
+        this.currentConnection = new AtomicInteger(initialConnection);
     }
 
     @Override
     public RelationalConnection getNewConnection(@Nonnull URI connectPath) throws SQLException {
-        return new MultiServerRelationalConnection(connectionSelectionPolicy, initialConnection, defaultFactory.getNewConnection(connectPath), alternateConnections(connectPath));
+        return new MultiServerRelationalConnection(connectionSelectionPolicy, getNextConnectionNumber(), defaultFactory.getNewConnection(connectPath), alternateConnections(connectPath));
     }
 
     @Override
@@ -99,6 +101,10 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
         return true;
     }
 
+    public int getCurrentConnection() {
+        return currentConnection.get();
+    }
+
     @Nonnull
     private List<RelationalConnection> alternateConnections(URI connectPath) {
         return alternateFactories.stream().map(factory -> {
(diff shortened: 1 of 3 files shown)


yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        assertEquals(ic, classUnderTest.getCurrentConnection());
+        classUnderTest.getNewConnection(URI.create("Blah"));
+        assertEquals(ic, classUnderTest.getCurrentConnection());
+        classUnderTest.getNewConnection(URI.create("Blah"));
+        assertEquals(ic, classUnderTest.getCurrentConnection());
+    }
+
+    @ParameterizedTest
+    @CsvSource({"0", "1", "2", "3", "4"})
+    void testAlternatePolicy(String initialConnection) throws SQLException {
```
ScottDugas, line 54: I don't think this test has the correct coverage for the modulo. Namely, you don't test that it does `modulo`, which seems bad.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
```diff
@@ -110,6 +116,24 @@ private List<RelationalConnection> alternateConnections(URI connectPath) {
...
+     * and thus connections with a single query should increment their position between connection creation or else they
+     * will always execute with the same initial connection.
+     * @return the next initial connection number to use
+     */
+    private int getNextConnectionNumber() {
+        switch (connectionSelectionPolicy) {
+            case DEFAULT:
+                return DEFAULT_CONNECTION;
+            case ALTERNATE:
+                return currentConnection.addAndGet(1) % (1 + alternateFactories.size());
```
ScottDugas, line 131: This doesn't look right. Why are you doing `1+` here. If you have 2 connections, and `initialConnection` is `0` I think this would return:
1. 0 + 1 % 3 -> 1
2. 1 + 1 % 3 -> 2
3. 2 + 1 % 3 -> 0
4. 3 + 1 % 3 -> 1

But `2` is not valid.
And returning 1 seems wrong.

But, if my reading is right, this would fail tests, so maybe I'm wrong.

It might be clearer with
```
currentConnection.getAndUpdate(i -> (i + 1) % alternateFactories.size());
```

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        int ic = Integer.parseInt(initialConnection);
+        MultiServerConnectionFactory classUnderTest = new MultiServerConnectionFactory(
+                MultiServerConnectionFactory.ConnectionSelectionPolicy.ALTERNATE,
+                ic,
+                dummyConnectionFactory(),
+                List.of(dummyConnectionFactory()));
+        assertEquals(ic, classUnderTest.getCurrentConnection());
```
ScottDugas, line 61: Note: Because you use `addAndGet`, `getCurrentConnection()` isn't what the next one will be created with, and thus the initial connection is not `initialConnection`.
Perhaps it would make senes to change `getCurrentConnection` to return a `MultiServerRelationalConnection` and have that have a getter on its `initialConnection`.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        classUnderTest.getNewConnection(URI.create("Blah"));
+        assertEquals(ic + 1, classUnderTest.getCurrentConnection());
+        classUnderTest.getNewConnection(URI.create("Blah"));
+        assertEquals(ic + 2, classUnderTest.getCurrentConnection());
```
ScottDugas, line 65: I wonder whether it would make sense to try and have a more integrated test, similar to `SupportedVersionTest`, where the test would fail if it did anything other than alternate.

ScottDugas (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
ohadzeliger: the `1` is to account for the primary connection. Normally, we would have 1 `defaultFactory` and 1 `alternateFactory` in the list. 
It is incrementing prematurely, though, so the right statement is:
`currentConnection.getAndAdd(1) % (1 + alternateFactories.size())`
or
`currentConnection.getAndUpdate(i -> (i + 1) % (1 + alternateFactories.size()))`

ohadzeliger: The difference is that the `%` is either accounted for in the counter or after the fact - I thought it is nice that it actually count the entire number of connections created, but don't have strong opinion on that

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MultiServerConnectionFactoryTest {
+    @ParameterizedTest
+    @CsvSource({"0", "1", "2", "3", "4"})
+    void testDefaultPolicy(String initialConnection) throws SQLException {
```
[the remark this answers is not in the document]
ohadzeliger: Currently it takes in any `initialConnection` and `%` it after the fact. I can change that if you think it is better.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        int ic = Integer.parseInt(initialConnection);
```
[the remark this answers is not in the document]
ohadzeliger: Cool, done. JUnit rocks.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
ohadzeliger: Added field for the total number of connections, hopefully this will make it easier to follow.
Changed the logic to modulo the current connection to be within the range of the number of factories

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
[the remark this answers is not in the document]
ohadzeliger: Done (changed to the proposal) -- looks easier to follow.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Added coverage by looking into the returned connections and seeing that their initial connection alternates

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Done - added tests that inspect the returned connection.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Not sure what you mean. Does the added coverage meet that?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
ScottDugas: Oh, yeah, that is where I was confused, the fact that there is the "default" and the alternate.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ScottDugas: I was thinking something that has an actual `.yamsql` file, and would fail if it did anything other than alternate connections. But I think that would be a lot of work, and not a lot of additional coverage.

ohadzeliger: I see. That's a good idea, but like you said, may take some time. I'll keep it around for future improvements?

PR #3082, opened 2025-01-31
Increment version in gradle.properties automatically

alecgrieser:
This migrates the version so that the release build can use the version in gradle.properties fully. It adds a script to parse and increment the version, and then invokes that script at the end of the release process to update that number. The script by default updates the `build` number (the third argument in our four-tuple semver variant), but it can be adjusted to update an arbitrary spot, which also means it can be used for patch builds off of a release branch or for updating the major or minor version.


MMcM (dismissed):

MMcM (approved):

.idea/compiler.xml
```diff
@@ -1,92 +0,0 @@
...
-        <module name="fdb-record-layer.fdb-extensions.test" />
-        <module name="fdb-record-layer.fdb-record-layer-core.test" />
-        <module name="fdb-record-layer.fdb-record-layer-icu.main" />
-        <module name="fdb-record-layer.fdb-record-layer-lucene.main" />
-        <module name="fdb-record-layer.fdb-java-annotations.main" />
-      </profile>
-      <profile name="Gradle Imported" enabled="true">
-        <outputRelativeToContentRoot value="true" />
-        <processorPath useClasspath="false">
-          <entry name="$PROJECT_DIR$/fdb-java-annotations/.out/libs/fdb-java-annotations-4.0-SNAPSHOT.jar" />
```
alecgrieser, line 66: I'm removing this file from git entirely because this dependency would otherwise update now every time we did a release, generating conflicts. This file is actually already in the `.gitignore`, and I validated on my local checkout that the file is recreated when deleted. On a fresh checkout, it actually appears to no longer be needed, so both existing users upgrading and no users downloading should be okay with this change. But I wanted it in with this change to avoid conflicts on this file going forward.

docs/Versioning.md
```diff
@@ -25,36 +33,33 @@ For the most up-to-date builds, the following maven repo should be added to the
...
 }
 ```
 
-Then the following dependency can be added:
+For builds older than 4.1 but newer than 2.8, an additional repository must be added to the build configuration, as older versions were not published to Maven Central:
 
 ```
-"org.foundationdb:fdb-record-layer-core:4.0.+"
+repositories {
+  maven {
```
ScottDugas, line 56: Should this have `mavenCentral()` ?
```suggestion
  mavenCentral()
  maven {
```

.github/workflows/patch_release.yml
```diff
@@ -0,0 +1,47 @@
...
+
+
+jobs:
+  gradle:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: write
+      packages: write
+      pages: read
```
ScottDugas, line 18: Why does this have `read` permissions but `release.yml` has write permissions?

build/versionutils.py
```diff
@@ -0,0 +1,109 @@
+#!/bin/python3
+#
+# increment_version.py
```
ScottDugas, line 3:
```suggestion
# versionutils.py
```

build/versionutils.py
```diff
@@ -0,0 +1,109 @@
...
+import re
+import subprocess
+import sys
+
+
+VERSION_POSITIONS = ['MAJOR', 'MINOR', 'BUILD', 'PATCH']
+VERSION_LINE = re.compile(r'version\s*\=\s*(\d+)\.(\d+)\.(\d+)\.(\d+)')
+
+
+def incremented_version(version: (int, int, int, int), update_type: str) -> tuple[int, int, int, int]:
```
ScottDugas, line 32: This is more for my own efforts to learn python, but why is `version` `(int, int, int, int)`, but this returns a `tuple[int, int, int, int]`

build/versionutils.py
```diff
@@ -0,0 +1,109 @@
...
+def update_version(filename: str, update_type: str) -> tuple[int, int, int, int]:
+    lines = []
+    found = False
+    version = None
+    new_version = None
+    with open(filename, 'r') as fin:
+        for l in fin:
+            m = VERSION_LINE.match(l)
+            if m:
+                 if version is not None:
```
ScottDugas, line 58: You use both `found` and `version is not None` to determine whether you found the version... a bit redundant.

build/versionutils.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                 lines.append(l)
+
+    if not found:
+        raise ValueError(f'Unable to find version in {filename}')
+
+    with open(filename, 'w') as fout:
+        for l in lines:
+            fout.write(l)
+    print(f'Update version file {filename}')
```
ScottDugas, line 73: It might be nice to have this include the old and new version.

build/versionutils.py
```diff
@@ -0,0 +1,109 @@
...
+    return new_version
+
+
+def get_version(filename: str) -> tuple[int, int, int, int]:
+    with open(filename, 'r') as fin:
+        for l in fin:
+            m = VERSION_LINE.match(l)
+            if m:
+               return int(m.group(1)), int(m.group(2)), int(m.group(3)), int(m.group(4))
```
ScottDugas, line 82: Should this fail if the version is in there twice?

ScottDugas (changes requested):

.github/workflows/patch_release.yml
alecgrieser: We could probably downgrade the pages permission of `release.yml` until we actually publish documentation updates, but the plan would be that patch releases don't update the docs on GitHub pages, whereas non-patch releases do.

build/versionutils.py
alecgrieser: Hm, the official documentation say to style it like `tuple[int, int, int, int]`: https://docs.python.org/3/library/typing.html#annotating-tuples I guess `(int, int, int, int)` must be an alias. I've gone ahead and made it consistently go with `tuple[int, int, int, int]`

build/versionutils.py
alecgrieser: Added

build/versionutils.py
alecgrieser: Removed `found`

build/versionutils.py
alecgrieser: Sure. I went back and forth about how much we wanted to have duplicated between this function and `update_version`, but I can make it error if there's more than one

docs/Versioning.md
alecgrieser: I actually don't think so. You only need `mavenCentral()` for our dependencies, and it does note above that you need an _additional_ repository (beyond maven central) to get the RL build here

ScottDugas (approved):

PR #3083, opened 2025-01-31
Rework YamlIntegrationTests to not use inheritance

ScottDugas:
The inheritance tree makes it hard to add, or mix configurations. It also makes it impossible to have dynamically generated configurations (e.g. run mixed-mode with all of a bunch of different old versions). At some point it would also make it cumbersome, as all the tests have to live in one place.

The key thing here is the introduction of `@YamlTest` annotation which provides contexts for `@TestTemplate`, and is configured to run with a bunch of different `YamlTestConfig`s.

You may find this easier to review commit-by-commit. Or maybe not. The commits are pretty sensible, although a bunch of the classes don't get javadoc until the very end.

I left `YamlTestBase`, so I think this should be backwards-compatible, but I expect to remove that soon.


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -0,0 +1,161 @@
...
+        } else {
+            testConfigs = List.of(
+                    new EmbeddedConfig(),
+                    new JDBCInProcessConfig(),
+                    new MultiServerConfig(0, 1111, 1112),
+                    new MultiServerConfig(1, 1113, 1114)
+            );
+        }
+        for (final YamlTestConfig testConfig : testConfigs) {
+            testConfig.beforeAll();
```
ohadzeliger, line 66: Would be nice to find a way to run the `BeforeAll` prior to execution each config, to avoid cross-config conflicts.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public boolean supportsTestTemplate(final ExtensionContext context) {
+        return true; // TODO check type
+    }
+
+    @Override
+    public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(final ExtensionContext context) {
+        return testConfigs.stream().map(config -> new Context(config, context.getRequiredTestMethod().getAnnotation(ExcludeYamlTestConfig.class)));
```
ohadzeliger, line 93: Should we filter out the excluded configs here?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -0,0 +1,161 @@
...
+        public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+            return new YamlTest.Runner() {
+                @Override
+                public void run(final String fileName) throws Exception {
+                    run(fileName, false);
+                }
+
+                @Override
+                public void run(final String fileName, final boolean correctExplain) throws Exception {
+                    config.assumeSupport(fileName, excludedConfigs);
```
ohadzeliger, line 148: Why do we need this? I'm concerned that tests would get skipped but there won't be an obvious reason for it from the test code.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
```diff
@@ -281,17 +284,26 @@ private RelationalConnection getNextConnection(String op) {
...
+                    currentConnectionSelector = (currentConnectionSelector + 1) % relationalConnections.size();
                     return result;
                 default:
                     throw new IllegalStateException("Unsupported selection policy " + connectionSelectionPolicy);
             }
         }
+
+        @Override
+        public <T> T unwrap(final Class<T> iface) throws SQLException {
+            if (iface.equals(RelationalConnection.class)) {
```
ohadzeliger, line 302: Maybe appropriate to use `isWrapperFor`?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+public class MultiServerConfig implements YamlTestConfig {
+
+    private final JDBCInProcessConfig embeddedConfig;
+    private final ExternalServer externalServer;
+    private int initialConnection;
+
+    public MultiServerConfig(final int initialConnection, final int grpcPort, final int httpPort) {
+        this.initialConnection = initialConnection;
+        embeddedConfig = new JDBCInProcessConfig();
```
ohadzeliger, line 47: Nit: `this` looks like it could have extended `JDBCInProcessConfig`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -35,41 +32,42 @@
...
 
     private static final int SERVER_PORT = 1111;
     private final String jarName;
+    private final int grpcPort;
+    private final int httpPort;
     private String version;
     private Process serverProcess;
 
     /**
      * Create a new extension that will run latest released version of the server, as downloaded by gradle.
```
ohadzeliger, line 51:
```suggestion
     * Create a new process that will run latest released version of the server, as downloaded by gradle.
```

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -35,41 +32,42 @@
...
  */
-public class RunExternalServerExtension implements BeforeAllCallback, AfterAllCallback {
+public class ExternalServer {
 
-    private static final Logger logger = LogManager.getLogger(RunExternalServerExtension.class);
+    private static final Logger logger = LogManager.getLogger(ExternalServer.class);
     public static final String EXTERNAL_SERVER_PROPERTY_NAME = "yaml_testing_external_server";
     private static final boolean SAVE_SERVER_OUTPUT = false;
 
     private static final int SERVER_PORT = 1111;
```
ohadzeliger, line 43: Is this still in use?

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -18,232 +18,237 @@
...
+    @TestTemplate
+    public void groupByTests(YamlTest.Runner runner) throws Exception {
+        runner.run("groupby-tests.yamsql");
     }
 
-    @Test
-    public void standardTests() throws Exception {
-        doRun("standard-tests.yamsql");
+    @TestTemplate
+    @ExcludeYamlTestConfig(JDBCInProcessConfig.class)
```
ohadzeliger, line 43: Maybe add a reason for the exclusion (similar to `@Disabled`)?

ohadzeliger (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
ScottDugas: That being said, in order to get this to play well with `@ExcludeYamlTestConfig`, it was easier to change it to extend `JDBCInProcessConfig`

yaml-tests/src/test/java/YamlIntegrationTests.java
ScottDugas: 1. This was actually accidentally disabled, so I removed this
2. I added a `reason` field to the annotation, and filled it out with the new ones added for JDBC
3. At some point we may need to deal with having multiple configs, as they would probably have different reasons. Repeating the annotation, and changing it to not have an array for the values. But, I think it makes sense to figure that out when we have an opportunity.

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -40,7 +38,6 @@ public void groupByTests(YamlTest.Runner runner) throws Exception {
     }
 
     @TestTemplate
-    @ExcludeYamlTestConfig(JDBCInProcessConfig.class)
```
ohadzeliger: Nice - I'm assuming this test is now passing...

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/ConfigWithOptions.java
```diff
@@ -0,0 +1,74 @@
...
+import com.apple.foundationdb.relational.yamltests.YamlRunner;
+
+import javax.annotation.Nonnull;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An implementation of {@link YamlTestConfig} that sets additional options on top of a base config.
+ */
+public class ConfigWithOptions implements YamlTestConfig {
```
ohadzeliger, line 32: This is a nit (I don't think it should prevent us from merging this PR) but it would seem to be a simpler arrangement if the options were part of the base implementation of the `Config` rather than a delegation. This way you could create a `new MultiServerConfig(0, 1113, 1114, Map.of(Option, value)))`, and the test in YamlTestConfigExclusions would be `config instanceof JDBCInProcessConfig && config.getOptions(Option)`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -102,14 +105,17 @@ public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContex
      * method and a {@link YamlTestConfig}).
      */
     private static class Context implements TestTemplateInvocationContext {
+        @Nonnull
         private final YamlTestConfig config;
-        private final Set<Class<? extends YamlTestConfig>> excludedConfigs;
+        @Nonnull
         private final String excludedReason;
+        @Nullable
+        private final YamlTestConfigExclusions exclusion;
```
ohadzeliger, line 113: This allows for a single exclusion (was a collection before). Do you think this is enough?

ohadzeliger (approved):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
ohadzeliger: This is actually covered by a previous comment... Ignore.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/ConfigWithOptions.java
ScottDugas: That is true, but then you wouldn't be able to easily build an embedded, or combine different sets of options. 
There's actually a part of me that just wanted to make the fact that it is JDBC visible in the options, or make it a method such that `ConfigWithOptions` has `isJDBC()` which calls `underlying.isJDBC`, but I didn't feel like trying to engineer that at this point was the right call given that we have only the two things.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
ScottDugas: Yes, it only allows a single one. I started to add the ability to repeat the annotation, but none of the tests require two reasons for excluding, so I removed that, which keeps the code simpler. If we ever have a need it's not hard to add.

PR #3085, opened 2025-02-03
Metrics in explain

normen662:


yaml-tests/src/test/resources/bytes.yamsql
```diff
@@ -26,6 +26,7 @@ setup:
     - query: insert into lb values (1, x'deadbeef'), (2, b64'yv4='), (3, null);
 ---
 test_block:
+  name: bytes-tests
```
hatyo, line 29: please make the name optional, defaulting to the YAML file's name if not set explicitly.

normen662: The name is optional. The current name if not given is `unnamed-` followed by an index that indicates the position of the block in the file.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -195,15 +202,31 @@ String decorateQuery(@Nonnull String query) {
...
+                        .setQuery(currentQuery)
+                        .build();
+                final var expectedPlannerMetricsInfo = metricsMap.get(identifier);
+
                 if (success) {
                     logger.debug("✅️ plan match!");
                 } else {
+                    if (executionContext.testYamlRunnerOptions(YamlRunner.YamlRunnerOptions.SHOW_PLAN_ON_DIFF) &&
+                            actualDot != null && expectedPlannerMetricsInfo != null) {
+                        BrowserHelper.browse("/showPlanDiff.html",
```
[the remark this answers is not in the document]
normen662: Yes, it would. It is meant to be used for quickly looking what the graphical difference between last and current plan is. If you run `CORRECT_EXPLAIN` AND this option, you will see a lot of windows.

I have had this on the side for some time now. What I ended up doing was to also switch off parallelism if this option is on but I didn't do that here.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/TestBlock.java
```diff
@@ -436,6 +446,7 @@ private void executeInParallelizedMode(Collection<Consumer<RelationalConnection>
         }
         // Iterate through the futures to catch any uncaught errors/exceptions from the submitted tasks.
         for (var future : futures) {
+            Verify.verify(!future.isCancelled());
```
alecgrieser, line 449: What's the point of this `Verify`? Were there canceled futures sneaking in here causing trouble?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -236,22 +259,80 @@ void checkResultInternal(@Nonnull Object actual, @Nonnull String queryDescriptio
...
+                    if (expectedPlannerMetricsInfo == null) {
+                        executionContext.putMetrics(blockName, currentQuery, lineNumber, actualInfo, true);
+                        logger.debug("⭐️ Successfully inserted new planner metrics at line {}", getLineNumber());
+                    } else {
+                        final var actualCountersAndTimers = actualInfo.getCountersAndTimers();
+                        final var expectedCountersAndTimers = expectedPlannerMetricsInfo.getCountersAndTimers();
+                        boolean isDifferent =
+                                actualCountersAndTimers.getTaskCount() != expectedCountersAndTimers.getTaskCount()
+                                && log("‼️ taskCount differs; expected = " + expectedCountersAndTimers.getTaskCount() +
+                                        "; actual = " + actualCountersAndTimers.getTaskCount(), getLineNumber());
```
alecgrieser, line 295: It feels like this section could use some work to avoid repetition. I think it could be rewritten to:

1. Take a list of field descriptors that we want to check. In theory, this could be an allow list of fields we care about, or a deny list of fields we _don't_ care about, the latter being a little more unwieldy at first but allowing us to add fields to the metrics and get them checked in the future by default.
1. For a given field, use `Message.getField()` on the new and the old to compare the values
1. If they differ, use the name from the field descriptor to log

I think that basically gets the same behavior as here, but with less code duplication. As a style thing, I think using `x |= expr` is...probably slightly better than `x = expr || x`, but ¯\\\_(ツ)_/¯ I could see both ways on that.

yaml-tests/src/test/java/YamlTestBase.java
```diff
@@ -65,10 +66,11 @@ protected Map<String, Object> getAdditionalOptions() {
     }
 
     protected final void doRun(@Nonnull final String fileName) throws Exception {
-        doRun(fileName, false);
+        doRun(fileName, EnumSet.noneOf(YamlRunner.YamlRunnerOptions.class));
+        // doRun(fileName, EnumSet.of(YamlRunner.YamlRunnerOptions.CORRECT_STATS));
```
alecgrieser, line 70: Normally, I'd suggest removing this commented out line, but maybe what we really need here is documentation on how to run the tests with different options--maybe even pre-baked `EnumSet`s that, theoretically, we could even hook up to certain gradle args, so if you ran `./gradlew :yaml-tests:test -PcorrectStats` or something, it would do this rewriting for you.

I don't think the latter needs to happen now necessarily

yaml-tests/src/test/resources/aggregate-empty-table.metrics.binpb
```diff
@@ -0,0 +1,913 @@
+�
```
alecgrieser, line 1: Any idea why some `.binpb` files are being displayed as a text file whereas  others are displayed as binary? It...mostly shouldn't matter, unless hypothetically, someone were to try to download this onto a `CRLF` system and the "new-lines" were helpfully changed for them. That is, assuming we're okay with checking binary data into the codebase. (I think we're mostly okay with it. It can balloon the size of git history if the files are too big or updates are too frequent, but this is...probably fine?)

alecgrieser (dismissed):
LGTM. There are a few comments, though nothing of too much substance. This diff is straining the ability of the GitHub UI to handle, so I'd be fine if they were addressed as a follow up. The prb did fail with a checkstyle problem that could be addressed

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/TestBlock.java
normen662: I think there is a bug but I cannot prove it. So if you submit executables more of less in bulk like we do a couple of lines up, there is a chance that an executable is still in the queue of the executor pool when the pool is shut down. It will not wait for those tasks to start and finish but just cancel those futures. I cannot repro it so I left that in here.

normen662: I have no idea why we are not using a countdown latch or similar here.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
normen662: As the `!=` part of the comment, it is different, as I get to see every violation in the way I have written it versus only the first one in the way you are suggesting I think.

normen662: I think I want to keep this way of writing this for now because I think over time we may grow these tests apart from each other: this thing can go up by 10% per commit, this thing can only go down, yadda yadda.

yaml-tests/src/test/java/YamlTestBase.java
normen662: So I think @ScottDugas  added the ability to connect some options to the YamlTestConfig. I could try to get these other options for correcting the yaml files under that umbrella. Not sure, though, since these other options are somewhat orthogonal, and the options that we hand into to the runner are one-offs you never want to have on when you check in.

yaml-tests/src/test/resources/aggregate-empty-table.metrics.binpb
normen662: No idea why. `.binpb` is the recommendation in the `protobuf` documentation IIRC. It shows as binary in IntelliJ.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
alecgrieser: > As for the != part of the comment, it is different, as I get to see every violation in the way I have written it versus only the first one in the way you are suggesting I think.

I don't believe that's true. A single `|` is the "bitwise or" operator, and for Boolean expressions, that means that it evaluates each side of the expression first (it never early outs).

> I think I want to keep this way of writing this for now because I think over time we may grow these tests apart from each other: this thing can go up by 10% per commit, this thing can only go down, yadda yadda.

Hm, I'm still not a huge fan. It would be nice to at least extract a helper function to check exact matching. My guess is that if we do start having different fields with different tolerances for matching, we'd want to introduce different helper functions and comparisons anyway.

normen662: You are correct about the `!=`!

yaml-tests/src/main/proto/planner_metrics.proto
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+// make sure to use this package naming convention:
+// com.apple.foundationdb.relational.yamltests.generated.<name_of_test_suite>
+// adding "generated" is important to exclude the generated Java file from Jacoco reports.
```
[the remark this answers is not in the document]
normen662: removed the comment.

yaml-tests/src/test/java/YamlTestBase.java
normen662: I added a mechanism to the extension that allows test case maintenance through an annotation:

```
@MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
```

which can be on the test method or the test class in question. Other enum values are allowed to only update the explain or only update the metrics.

hatyo (approved):
Approving not to cause an unnecessary delay, we can always refactor this later if need be.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -66,23 +72,29 @@ public void beforeAll(final ExtensionContext context) throws Exception {
                     new MultiServerConfig(1, 1115, 1116),
                     new ForceContinuations(new MultiServerConfig(1, 1117, 1118))
             );
+            maintainConfigs = List.of(
+                    new CorrectExplains(new EmbeddedConfig()),
+                    new CorrectMetrics(new EmbeddedConfig()),
+                    new CorrectExplainsAndMetrics(new EmbeddedConfig())
```
alecgrieser, line 78: Do we need all three of these here? Or would

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -52,11 +56,13 @@
 public class YamlTestExtension implements TestTemplateInvocationContextProvider, BeforeAllCallback, AfterAllCallback {
     private static final Logger logger = LogManager.getLogger(YamlTestExtension.class);
     private List<YamlTestConfig> testConfigs;
+    private List<YamlTestConfig> maintainConfigs;
 
     @Override
     public void beforeAll(final ExtensionContext context) throws Exception {
         if (Boolean.parseBoolean(System.getProperty("tests.runQuick", "false"))) {
             testConfigs = List.of(new EmbeddedConfig());
+            maintainConfigs = List.of();
```
alecgrieser, line 65: If we wanted to rewrite all of the metrics, is there a way of doing that besides, like, setting the `testConfigs` to one of the maintain configs, and then running the tests? It seems like there's a fairly straightforward way in this design to update a single test or a single test file, but not all of the tests.

alecgrieser (approved):
I agree with @hatyo that we could take this now and then do some additional refactoring if we don't want to delay this further

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
normen662: You can annotate the test class which would rewrite the metrics file for every test in that class.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
normen662: As it turns out, whenever you add a new query and you kind of just want it to put the explain into the yaml file, you need the option to correct both at the same time, since either one would fail if run separately, i.e. `CorrectMetrics` would fail due to the missing explain, and `CorrectExplain` would fail on the missing metrics.

alecgrieser: Why not just have one (just `CorrectExplainAndMetrics`)?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
alecgrieser: I get that. I was asking if there's a way to get it correct all tests across all test files. In particular, I was thinking it would be a good idea to have one that doesn't require updating any of the Java files, as that makes it more difficult to accidentally commit the code in correction-mode

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
normen662: probably you are right but sometimes you really don't want to change the explain if you know only the counters change. It is easier to make a mistake if you always ask for both although in reality that is probably going to happen anyways.

alecgrieser: Maybe I'm misunderstanding how this works, but won't this run all three of these if you set the tag on the test/test class? I think you'd need separate annotations for each if you wanted to just correct the counters, say

PR #3088, opened 2025-02-03
Upgrade de.undercouch.download gradle plugin

ScottDugas:
This updates the download plugin to the latest version. It looks like this is used to download the geonames.


alecgrieser (approved):
I guess I appreciate the "one PR, one change" policy here with these version updates, but there are also a fair number of conflicts as a result of these. Given that, it may have been fine to have just updated all the version numbers in one go.

PR #3092, opened 2025-02-04
infinite continuations when NULL_ON_EMPTY is in the plan

normen662:
- remove default `...ON_EMPTY` behavior that previously was part of `RecordQueryStreamingAggregationPlan`
- have `RecordQueryDefaultOnEmptyPlan` enforce and adjust skips and limits correctly

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
@@ -193,4 +193,8 @@ public class StreamGrouping<M extends Message> {
         final EvaluationContext nestedContext = context.withBinding(Bindings.Internal.CORRELATION, alias, currentObject);
         return Objects.requireNonNull(groupingKeyValue).eval(store, nestedContext);
     }
+
+    public boolean isResultOnEmpty() {
+        return groupingKeyValue == null;
+    }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
@@ -193,8 +193,4 @@ public class StreamGrouping<M extends Message> {
         final EvaluationContext nestedContext = context.withBinding(Bindings.Internal.CORRELATION, alias, currentObject);
         return Objects.requireNonNull(groupingKeyValue).eval(store, nestedContext);
     }
-
-    public boolean isResultOnEmpty() {
-        return groupingKeyValue == null;
-    }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
@@ -93,9 +93,10 @@ public class RecordQueryDefaultOnEmptyPlan implements RecordQueryPlanWithChild, 
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties),
-                (executor, cont) -> RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context)))),
-                continuation);
+        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties.clearSkipAndLimit()),
+                        (executor, cont) ->
+                                RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context))), cont), continuation)
+                .skipThenLimit(executeProperties.getSkip(), executeProperties.getReturnedRowLimit());
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
```diff
@@ -93,9 +93,10 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties),
-                (executor, cont) -> RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context)))),
-                continuation);
+        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties.clearSkipAndLimit()),
+                        (executor, cont) ->
+                                RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context))), cont), continuation)
+                .skipThenLimit(executeProperties.getSkip(), executeProperties.getReturnedRowLimit());
```
alecgrieser, line 99: Do we have any coverage of the skip case? In particular, we have a behavior change here for a query like:

```sql
SELECT count(*) FROM t SKIP 1
```

I think the old code would skip the underlying count (which returns 1 result), so it would hit the `else` branch and return a single result of 0, whereas now, this would apply the skip after the `else`, and return an empty result

yaml-tests/src/test/resources/aggregate-empty-table.yamsql
```diff
@@ -36,7 +36,9 @@ test_block:
     -
       - query: select count(*) from T1;
       - explain: "SCAN(<,>) | TFILTER T1 | MAP (_ AS _0) | AGG (count_star(*) AS _0) | ON EMPTY NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)"
+      - maxRows: 1
```
alecgrieser, line 39: Should similar tests be added for the other aggregate types (sum, max, etc)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
```diff
@@ -93,9 +93,10 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties),
-                (executor, cont) -> RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context)))),
-                continuation);
+        return RecordCursor.orElse(cont -> getChild().executePlan(store, context, cont, executeProperties.clearSkipAndLimit()),
+                        (executor, cont) ->
+                                RecordCursor.fromList(ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context))), cont), continuation)
```
alecgrieser, line 98:
```suggestion
                                RecordCursor.fromList(executor, ImmutableList.of(QueryResult.ofComputed(onEmptyResultValue.eval(store, context))), cont), continuation)
```

This is kind of unrelated, but we should be threading the executor through here as well. Looking at this API, I think the _actual_ API here shouldn't pass the executor at all, and we should instead pass in the plan's executor directly, but that's a larger change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
```diff
@@ -193,8 +193,4 @@ private Object evalGroupingKey(@Nullable final Object currentObject) {
         final EvaluationContext nestedContext = context.withBinding(Bindings.Internal.CORRELATION, alias, currentObject);
         return Objects.requireNonNull(groupingKeyValue).eval(store, nestedContext);
     }
-
-    public boolean isResultOnEmpty() {
```
alecgrieser, line 197: If we go with this as is, I believe it means that deserialized plans that were created on older versions will have different semantics than they were originally planned. So, for example, a streaming aggregate with a null grouping that was planned without a `DefaultOnEmpty` plan on top will go from producing a single result to producing no values at all. That only becomes relevant if there's a continuation, which may not actually be possible here, but it feels a bit iffy.

If we choose not to make the cross-version compatibility work, we should at least call out how it's been broken.

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
normen662: added the `executor`

yaml-tests/src/test/resources/aggregate-empty-table.yamsql
normen662: I _converted_ a test doing `sum(...)` in the same yaml. This test ensures that `NULL_ON_EMPTY` also works without coalescing to a `0`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
alecgrieser: Sorry, this was a little unclear: I think the new behavior is better, but I was noting a lack of coverage if the behavior code change without any tests noticing.

Also, this should really be `OFFSET` instead of `SKIP`, but also, we don't currently support `OFFSET` in the SQL dialect. If it's not exposed, it shouldn't be an issue

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
alecgrieser: Did you miss a push? I'm not seeing the executor in the diff

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java
normen662: It's there!

PR #3101, opened 2025-02-04
Upgrade protobuf plugin from 0.8.19 to 0.9.4

ScottDugas:
https://github.com/google/protobuf-gradle-plugin/releases

> - Significantly improve support for task configuration avoidance. README.md has not yet been updated to encourage using configureEach instead of each within generateProtoTasks {}, but it is beneficial as long as you are not using any IDE plugins
> - Added generated code to java SourceSet instead of only adding it to JavaCompile and related tasks. This should cause tasks like sourcesJar and javadoc to now include the generated code. You may need to exclude the generated code from linters

I skimmed through the differences in the generated java files in protogen, and AFAICT the only change is that `Builder`s now have `@SuppressWarnings("deprecation")` on them.


alecgrieser:
> I skimmed through the differences in the generated java files in protogen, and AFAICT the only change is that Builders now have @SuppressWarnings("deprecation") on them.

Does that mean we can get rid of our explicit addition of the deprecation warning in `proto.gradle`?

PR #3105, opened 2025-02-05
Upgrade gradle serviceloader plugin from 1.1.8 to 1.1.9

ScottDugas:


ScottDugas:
The only change between 1.1.8 and 1.1.9 is:
> Release 1.1.9 Add kotlin support(https://github.com/harbby/gradle-serviceloader/pull/8 ) (untested)

So maybe not something we should take right now.
`1.1.8` already supports gradle 8.

PR #3108, opened 2025-02-05
preparing for up-level DEFAULTS_ON_EMPTY changes

normen662:
Cherry picks the cursor code from #3092 onto the 4.0.559 branch. That PR adapts the behavior of the `AggregateCursor` so that it no longer produces an incorrect "BEGIN" continuation in association with the final element of an empty aggregate. Such a continuation can produce an infinite loop if the continuation is then used to re-issue the query from the beginning (or throw an error if passed to `EXECUTE CONTINUATION`).

Because the fix changes the behavior of `AggregateCursor`, there would normally be some danger when continuing a plan from that version on older versions and vice versa To prevent this, the code in #3092 only goes to the new behavior if there is a flag set in the serialized continuation (so it will continue with the older semantics when deserializing an older continuation). We want to be able to reliably go to the new behavior in mixed-mode scenarios with 4.0.559 and the current `main` version, and so to make that work, we need to be able to parse this new flag on the 4.0.559 branch.

Note that the code in this PR _only_ goes to the new behavior when deserializing a new continuation with the flag set. So, behavior is preserved except in mixed-mode scenarios. Future upgrade testing will be done to validate that the continuation does get parsed correctly and produce the expected outcome, but the code is identical to the one taken in #3092 (except for defaulting to the old behavior instead of the new), so we are confident that this should be safe and correct.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -22,7 +22,7 @@ Our API stability annotations have been updated to reflect greater API instabili
 * **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Streaming aggregations to communicate whether to create a default on empty value via continuations which prepares a later fix for infinite continuations [(Issue #3093)](https://github.com/FoundationDB/fdb-record-layer/issues/3093)
 * **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Performance** Improvement 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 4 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -408,7 +421,7 @@ public static RecordQueryStreamingAggregationPlan of(@Nonnull final Quantifier.P
         final var referencedAggregateValue = ObjectValue.of(aggregateAlias, aggregateValue.getResultType());
 
         return new RecordQueryStreamingAggregationPlan(inner, groupingKeyValue, aggregateValue, groupingKeyAlias, aggregateAlias,
-                resultValueFunction.apply(referencedGroupingKeyValue, referencedAggregateValue));
+                resultValueFunction.apply(referencedGroupingKeyValue, referencedAggregateValue), true);
```
normen662, line 424: Note `true` here, but `false` in #3092.

docs/ReleaseNotes.md
```diff
@@ -22,7 +22,7 @@ Our API stability annotations have been updated to reflect greater API instabili
 * **Bug fix** Fix 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** ungrouped GROUP BY queries result in infinite continuations when maxRows is 1 [(Issue #3093)](https://github.com/FoundationDB/fdb-record-layer/issues/3093)
```
alecgrieser, line 25: Ideally, we'd note here that we can start to interpret continuations correctly that fix this bug, somehow, rather than just saying that this fixes the issue

normen662: done

alecgrieser (approved):

alecgrieser:
I'm going to merge now. We have a passing PRB run on this commit, but the PRB was re-run because the description was updated. That's definitely a mistake, so I'm going to take this now

PR #3110, opened 2025-02-05
PRBs shouldn't re-run on edited

alecgrieser:
The `edited` event for GitHub actions corresponds to title and description edits, which shouldn't re-run our PR builds.

It also handles the case where someone changes the base branch. In that case, we probably _do_ want to re-run the PR build, but that's a bit of an edge case, and one that we could probably solve by manually re-running the action if necessary.


alecgrieser:
Amending the most recent commit to validate that _that_ does produce a new build. It appears to have worked, as expected

MMcM (approved):

PR #3111, opened 2025-02-05
set skip and limit for the recursive union cursor correctly.

### LINKED ISSUE #3100: Continuations ignored for query
Summary:
Specifying maxRows:1 for recursive query has no effect
Steps To Reproduce:
1. Look at cte.yamsql
2. Add maxRows:1 to the test:
      - query: with c1 as (select col1, col2 from t1) select col1, col2 from c1
      - maxRows: 1
      - unorderedResult: [{COL1: 10, COL2: 1},
                          {COL1: 10, COL2: 2},
                          {COL1: 20, COL2: 6},
                          {COL1: 20, COL2: 7}]
3. The test still passes

Note: The same happens for UPDATE queries.

hatyo:
This sets the `limit` and `skip` (if any) on the recursive union cursor correctly. It also includes a number of small fixes to the YAML test framework error reporting in case of result mismatches, and a small optimization of handling unordered result matching.

This fixes #3100.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/Matchers.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/Matchers.java
@@ -30,7 +30,6 @@ import com.apple.foundationdb.relational.recordlayer.query.ParseHelpers;
 import com.apple.foundationdb.relational.util.SpotBugsSuppressWarnings;
 
 import com.google.common.collect.HashMultiset;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Multiset;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
@@ -64,7 +63,7 @@ public class Matchers {
         if (obj instanceof List) {
             return (List<?>) obj;
         }
-        fail(String.format("Expecting '%s' to be of type '%s'", desc, List.class.getSimpleName()));
+        fail(String.format(Locale.ROOT, "Expecting '%s' to be of type '%s'", desc, List.class.getSimpleName()));
         return null;
     }
 
@@ -80,7 +79,7 @@ public class Matchers {
         if (obj instanceof Map<?, ?>) {
             return (Map<?, ?>) obj;
         }
-        fail(String.format("Expecting %s to be of type %s", desc, Map.class.getSimpleName()));
+        fail(String.format(Locale.ROOT, "Expecting %s to be of type %s", desc, Map.class.getSimpleName()));
         return null;
     }
 
@@ -88,7 +87,7 @@ public class Matchers {
         if (obj instanceof Map.Entry<?, ?>) {
             return (Map.Entry<?, ?>) obj;
         }
-        fail(String.format("Expecting %s to be of type %s", desc, Map.Entry.class.getSimpleName()));
+        fail(String.format(Locale.ROOT, "Expecting %s to be of type %s", desc, Map.Entry.class.getSimpleName()));
         return null;
     }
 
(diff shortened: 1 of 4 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/Matchers.java
```diff
@@ -449,20 +451,24 @@ private static ImmutablePair<ResultSetMatchResult, ResultSetPrettyPrinter> match
 
     private static ImmutablePair<ResultSetMatchResult, ResultSetPrettyPrinter> matchUnorderedResultSet(final RelationalResultSet actual, final @Nonnull Multiset<?> expectedAsMultiSet) throws SQLException {
         final ResultSetPrettyPrinter resultSetPrettyPrinter = new ResultSetPrettyPrinter();
-        var i = 1;
-        // O(n^2) -- we all got M1s
```
normen662, line 453: This is super important! :-)

yaml-tests/src/test/resources/recursive-cte.yamsql
```diff
@@ -68,6 +68,49 @@ test_block:
                  {210, 20},
                  {250, 50},
                  {250, 50}]
+    -
+      - query: with recursive c1 as (
+            select id, parent from t1 where parent = -1
+            union all
+            select b.id, b.parent from c1 as a, t1 as b where a.id = b.parent) select id from c1
+      - maxRows: 1
+      - result: [{ID: 1}]
```
normen662, line 77: there is an outstanding PR that does this stuff automatically for up/downlevel testing. So you may not need these tests in a week from now.

normen662 (approved):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/Matchers.java
```diff
@@ -471,12 +477,12 @@ private static ImmutablePair<ResultSetMatchResult, ResultSetPrettyPrinter> match
...
                 printCurrentAndRemaining(actual, resultSetPrettyPrinter);
-                return ImmutablePair.of(ResultSetMatchResult.fail(String.format("result row at %d does not match any expected records", i)), resultSetPrettyPrinter);
+                return ImmutablePair.of(ResultSetMatchResult.fail(String.format("actual row at %d does not match any expected records", actualRowsCounter)), resultSetPrettyPrinter);
             }
-            i++;
+            actualRowsCounter++;
         }
         if (!expectedAsMultiSet.isEmpty()) {
-            return ImmutablePair.of(ResultSetMatchResult.fail("result does not contain all expected rows"), null);
+            return ImmutablePair.of(ResultSetMatchResult.fail(String.format("result does not contain all expected rows, expected %d row(s), got %d row(s) instead.", expectedRowCount, actualRowsCounter - 1)), null);
```
alecgrieser, line 485: I understand that this is tests, but I think we want to refrain from `String.format` (especially without a locale) if we can avoid it. That does seem to be a larger problem in this file, though

alecgrieser (approved):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/Matchers.java
hatyo: I made all the `String.format` calls here use `Locale.ROOT`. I will propose a separate PR that extends this fix to the entire codebase.

yaml-tests/src/test/resources/recursive-cte.yamsql
hatyo: Got it, I would leave it for now not to lose test coverage, and we can clean up later when we have a way to automatically test continuations in YAML.

alecgrieser (approved):

normen662 (approved):

yaml-tests/src/test/resources/recursive-cte.yamsql
```diff
@@ -68,6 +68,51 @@ test_block:
                  {210, 20},
                  {250, 50},
                  {250, 50}]
+    -
+      - query: with recursive c1 as (
+            select id, parent from t1 where parent = -1
+            union all
+            select b.id, b.parent from c1 as a, t1 as b where a.id = b.parent) select id from c1
+      - supported_version: !current_version
```
hatyo, line 76: It looks like without this, some JDBC tests were failing as they were incorrectly trying to run continuations against `4.0.575.0`. If I understand what this configuration means correctly, the query will be tested against current commit onwards only, which is the correct thing to do, considering that recursive CTE support of continuations was misbehaving before this PR. @ScottDugas can you please have a look?

normen662 (approved):

alecgrieser (dismissed):
LGTM. I'm not sure we have great documentation of the stability of this feature, which is for all intents and purposes experimental. We should probably still call out that limit and skip behavior have changed in the release notes, potentially calling it out as a breaking change. We'll be updating to 4.1 fairly soon (see #3082), and I don't think we'd usually update the minor version on an experimental change like this, but it wouldn't be a bad idea to piggy back on that minor version bump in this case (which basically just means merging but not releasing until the other PR goes in)

alecgrieser:
Given the state of the PRs, we could even update the release notes on #3082 to avoid merge conflicts

alecgrieser (approved):

PR #3112, opened 2025-02-05
Resolves #3172: Fix continuation bug in StreamAggregation plan

### LINKED ISSUE #3172: Continuation is not correct in AggregateCursor when a group finishes

pengpeng-lu:
The problem was: continuation is advanced after a group ends, so if we execute the query with the returned continuation, we'll start from the second row in the next group.
This PR fixes the problem, and adds more tests.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -84,7 +84,10 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
             } else {
                 final QueryResult queryResult = Objects.requireNonNull(innerResult.get());
                 boolean groupBreak = streamGrouping.apply(queryResult);
-                previousValidResult = innerResult;
+                if (!groupBreak) {
+                    // previousValidResult is the last one in the same group, it sets the continuation
+                    previousValidResult = innerResult;
+                }
```
alecgrieser, line 90: I almost wonder if this needs a little more thought. I'm worried about the following case:

1. There are two (three?) groups that each have a single key. plus more data
2. There's a request for groups with a limit of 2
3. Because each key represents a group break, we may never update `previousValidResult` which
4. In turn means we never advance the continuation

Or something like that--I'm not perhaps as familiar with the code as I should be.

It seems like what we actually want here is for the `StreamGrouping` to keep track of the continuation, updating it with each value we read in the group, and then associating the last continuation in the group with the result itself.

This may help with the other problem identified, namely that we drop data that is in the last group if we hit an out-of-band limit. (I think that code path is not tested at all, as there are some serious issues when continuing.) Ideally, we'd have the grouping serialize its intermediate state and resume, though we need to do some compatibility work to make that possible.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -208,14 +213,34 @@ void aggregateThreeGroupByTwo(final boolean useNestedResult) {
...
             assertResults(useNestedResult ? this::assertResultNested : this::assertResultFlattened, result, resultOf(0, "0", 1, 0, 0.5), resultOf(1, "0", 2, 2, 2.0), resultOf(1, "1", 3, 3, 3.0), resultOf(2, "1", 9, 4, 4.5));
         }
     }
 
     @ParameterizedTest(name = "[{displayName}-{index}] {0}")
-    @BooleanSource
-    void aggregateNoRecords(final boolean useNestedResult) {
+    @MethodSource("provideArguments")
+    void aggregateOneGroupByThree(final boolean useNestedResult, final int rowLimit) {
+        // each group only has one row
```
alecgrieser, line 224: Nice! Thanks!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -98,6 +101,25 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
             }
             // Use the last valid result for the continuation as we need non-terminal one here.
             RecordCursorContinuation continuation = Verify.verifyNotNull(previousValidResult).getContinuation();
+            /*
+            Here is how continuation(previousValidResult) moves:
```
alecgrieser, line 105: Okay, I had a bit of trouble understanding the main idea of this comment, but I think it could work with a few tweaks. First, it would be good to start out with something like, "Update the `previousValidResult` to the next continuation even though it hasn't been returned. This is to return the correct continuation when there are single-element groups." Then I think the table you have makes sense, but without that pre-amble, I glazed over the `previousValidResult = row2` line on 110, and on line 113, it's not clear why `row3` is necessarily of a different group.

On the other hand, I don't want to do too much bike shedding here, because I suspect that we'll change some of this when we re-do things to handle out of bounds limits.

alecgrieser (changes requested):
This probably needs a release note (though watch out for #3163 which will conflict with many release note changes)

alecgrieser:
There may be an infinite loop somewhere, because the `core-tests` had been running for five hours before I cancelled them. See: https://github.com/FoundationDB/fdb-record-layer/actions/runs/13383260519/job/37375412577

docs/ReleaseNotes.md
```diff
@@ -23,8 +23,8 @@ Users performing online updates are encouraged to update from [4.0.559.4](#40559
 * **Bug fix** Fix 2 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Bug fix** The relational component now specifies `Locale.ROOT` in all usages of `String.format` which may change number formatting if the default locale does not align [(Issue #3121)](https://github.com/FoundationDB/fdb-record-layer/issues/3121)
 * **Bug fix** Fix 3 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
-* **Bug fix** Fix 5 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
+* **Bug fix** Insert statement does not fully validate the column names with supplied values [(Issue #3069)](https://github.com/FoundationDB/fdb-record-layer/issues/3069)
```
alecgrieser, line 26:
```suggestion
```

I think that went into 4.1.6.0, so we can just remove this line

alecgrieser (changes requested):
One small release note change

alecgrieser (approved):

PR #3117, opened 2025-02-07
Chore: use `Locale.ROOT` in `String.format`.

### LINKED ISSUE #3121: use `Locale.ROOT` in `String.format`
Many usages of `String.format` in the relational component are relying on the default `Locale`, this could render resulting strings in unexpected way if the JVM process is running with a specific locale such as `lrc_IR` for example.

hatyo:
Use `String.format` in locale-neutral to avoid rendering message in an unexpected way when the JVM process is configured with a specific locale such as `lrc_IR`.

This fixes #3121.


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/ProtobufDataBuilder.java
```diff
@@ -71,7 +72,7 @@ public String getFieldType(String fieldName) throws SQLException {
     public boolean isPrimitive(int fieldNumber) throws SQLException {
         final var field = typeDescriptor.getFields().get(fieldNumber - 1);
         if (field == null) {
-            throw new RelationalException(String.format("Field with number <%d> does not exist", fieldNumber), ErrorCode.INVALID_PARAMETER).toSqlException();
+            throw new RelationalException(String.format(Locale.ROOT, "Field with number <%d> does not exist", fieldNumber), ErrorCode.INVALID_PARAMETER).toSqlException();
```
ScottDugas, line 75: We should update the release notes with this change, as it will be user visible for anyone with an alternate default locale. 
Something broad, like "The relational project is now specifying `Locale.ROOT` in all `String.format` calls, which may change number formatting if the default locale does not align`

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LogicalOperator.java
```diff
@@ -184,7 +185,7 @@ public static LogicalOperator generateAccess(@Nonnull Identifier identifier,
             return logicalOperatorCatalog.lookupTableAccess(identifier, alias, requestedIndexes, semanticAnalyzer);
         } else {
             final var correlatedField = semanticAnalyzer.resolveCorrelatedIdentifier(identifier, currentPlanFragment.getLogicalOperatorsIncludingOuter());
-            Assert.thatUnchecked(requestedIndexes.isEmpty(), ErrorCode.UNSUPPORTED_QUERY, () -> String.format("Can not hint indexes with correlated field access %s", identifier));
+            Assert.thatUnchecked(requestedIndexes.isEmpty(), ErrorCode.UNSUPPORTED_QUERY, () -> String.format(Locale.ROOT, "Can not hint indexes with correlated field access %s", identifier));
```
ScottDugas, line 188: Probably quite a few of these would be better off using string concatenation, but using the simple mechanical change of adding `Locale.ROOT` seems good for this broad change.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/SemanticAnalyzer.java
```diff
@@ -668,31 +668,31 @@ public static void validateLimit(@Nonnull Expression expression) {
         Assert.thatUnchecked(underlying instanceof LiteralValue<?>);
         final var value = ((LiteralValue<?>) underlying).getLiteralValue();
         Assert.notNullUnchecked(value, ErrorCode.INVALID_ROW_COUNT_IN_LIMIT_CLAUSE,
-                () -> String.format("limit value out of range [1, %s]", Integer.MAX_VALUE));
+                () -> String.format(Locale.ROOT, "limit value out of range [1, %s]", Integer.MAX_VALUE));
```
ScottDugas, line 671: It's weird that these are using `%s` instead of `%d`, given that it is an integer.
Not really a thing that should be changed in this PR though.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/PlanGenerationStackTest.java
```diff
@@ -162,7 +162,8 @@ public Stream<? extends Arguments> provideArguments(final ExtensionContext conte
                     Arguments.of(75, "with recursive c as (select * from t, c) select * from c", "recursive CTE does not contain non-recursive term"),
                     Arguments.of(76, "with recursive c1 as (select * from restaurant union all select * from c1) select * From c1", null),
                     Arguments.of(77, "with recursive c as (with recursive c1 as (select * from restaurant union all select * from c1) select * From c1 union all select * from restaurant, c) select * from c", null),
-                    Arguments.of(78, "with recursive c as (with c as (select * from restaurant) select * from c union all select * from restaurant) select * from c union all select * from restaurant", "ambiguous nested recursive CTE name")
+                    Arguments.of(78, "with recursive c as (with c as (select * from restaurant) select * from c union all select * from restaurant) select * from c union all select * from restaurant", "ambiguous nested recursive CTE name"),
+                    Arguments.of(79, "with recursive c1(name) as (select * from restaurant union all select * from c1) select * From c", "cte query has 7 column(s), however 1 aliases defined")
```
ScottDugas, line 166: I always approve of additional testing, but to make sure I didn't miss a code change, why an additional test here?

ScottDugas (changes requested):

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/SemanticAnalyzer.java
hatyo: Nice catch, I fixed it.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/PlanGenerationStackTest.java
hatyo: Right, the additional test was to confirm the fix done here [0] in `SemanticAnalyzer`'s validation of the CTE columns, where the incorrect `%s` in the error message is replaced with `%d` as it is a placeholder for column count which is a number.

[0] https://github.com/FoundationDB/fdb-record-layer/pull/3117/files#diff-02328f5aa27f48c66a951cf05b4a1f0e866d2f3488f7471bfba6bf49c108cb4aR695

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LogicalOperator.java
hatyo: Yes, I will propose a different PR to remove usage of `String.format` where applicable.

Admittedly, most (if not all) of the current usages of `String.format` in `relational` are in an exceptional code path; either in an assertion failure, or to formulate an exception message. I do not think optimizing exceptional code paths for performance is crucial (they should not be in a tight loop, stack unrolling is already _very_ expensive, ... etc).

Nevertheless, it is probably a good practice to avoid using `String.format` all together, and have a code analysis check that errors out when one uses it by mistake, considering how terrible its performance is.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/ProtobufDataBuilder.java
hatyo: Done.

ScottDugas (approved):

ScottDugas (approved):

PR #3120, opened 2025-02-08
Split out the PR coverage commenting workflow …

MMcM:
… so that it can run in the target (default) branch context.

Pull request tests run in the potentially unsafe context of the base, but can only produce artifacts.
A second workflow is triggered after that and runs in default branch context, so it can be given permission to add a comment to the PR.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -54,14 +54,6 @@ jobs:
         name: core-test-reports
         path: |
           test-reports/fdb-record-layer-core/
-    - name: Test Coverage Comment
-      uses: madrapps/jacoco-report@7c362aca34caf958e7b1c03464bd8781db9f8da7
-      with:
-        paths: |
-          fdb-record-layer-core/.out/reports/jacoco/test/jacocoTestReport.xml,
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request_comments.yml
```diff
@@ -0,0 +1,49 @@
...
+    workflows: [Pull Request]
+    types:
+      - completed
+
+jobs:
+  download:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
```
alecgrieser, line 14: Did this run in the PRB? Would we expect it to, or is that a sign that something is up?

.github/workflows/pull_request.yml
```diff
@@ -101,21 +101,3 @@ jobs:
             test-reports/fdb-relational-jdbc/
             test-reports/fdb-relational-server/
             test-reports/yaml-tests/
-      - name: Test Coverage Comment
```
alecgrieser, line 104: Post #3122, I think we need to remove the equivalent step from `core-tests` above as well.

alecgrieser (changes requested):

PR #3124, opened 2025-02-10
addressing previous comments in metrics in explain

normen662:


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryConfig.java
```diff
@@ -324,9 +328,18 @@ void checkResultInternal(@Nonnull String currentQuery, @Nonnull Object actual, @
...
-        logger.error(message + lineNumber);
-        return true;
+    private static boolean isMetricDifferent(@Nonnull final PlannerMetricsProto.CountersAndTimers expected,
+                                             @Nonnull final PlannerMetricsProto.CountersAndTimers actual,
+                                             @Nonnull final Descriptors.FieldDescriptor fieldDescriptor,
+                                             int lineNumber) {
+        final long expectedMetric = (long)expected.getField(fieldDescriptor);
+        final long actualMetric = (long)actual.getField(fieldDescriptor);
+        if (expectedMetric != actualMetric) {
+            logger.warn("‼️ metric {} differs; lineNumber = {}; expected = {}; actual = {}",
```
alecgrieser, line 338: Any reason this isn't using a `KeyValueLogMessage`?

yaml-tests/src/test/resources/aggregate-index-tests-count-empty.metrics.yaml
```diff
@@ -3,9 +3,9 @@ agg-index-tests-count-empty:
     explain: 'AISCAN(MV1 <,> BY_GROUP -> [_0: VALUE:[0]]) | MAP (_ AS _0) | ON EMPTY
         NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)'
     task_count: 434
-    task_total_time_ms: 13
+    task_total_time_ms: 19
```
alecgrieser, line 6: Does this PR need to update all of the checked in metrics? It we can avoid doing it, it would be nice just to avoid churn on these files if it's not needed

alecgrieser (changes requested):

yaml-tests/src/test/resources/aggregate-index-tests-count-empty.metrics.yaml
normen662: Yes -- there was a mistake int the previous PR that it also kept metrics for `explainContains:` which causes problems for downstream tests.

normen662: So I am recreating all metrics here since I don't want any of the `explainContains:` metrics to be kept at all.

alecgrieser (approved):

PR #3126, opened 2025-02-10
addressing more comments related to metrics pr

normen662:


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -75,7 +76,8 @@ public void beforeAll(final ExtensionContext context) throws Exception {
             maintainConfigs = List.of(
                     new CorrectExplains(new EmbeddedConfig()),
                     new CorrectMetrics(new EmbeddedConfig()),
-                    new CorrectExplainsAndMetrics(new EmbeddedConfig())
+                    new CorrectExplainsAndMetrics(new EmbeddedConfig()),
+                    new ShowPlanOnDiff(new EmbeddedConfig())
```
alecgrieser, line 80: The show-plan-on-diff config seems different enough that you wouldn't want to run it in the same step as running explain or metrics correction

alecgrieser (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
normen662: You don't run it like that. You can use the maintain annotation to pick the configs from here you want to run.

alecgrieser: Oh okay, I see. I was confused by the flow here, but I think I understand:

1. The `maintainConfigs` lists all possible configurations that you'd want to run to maintain the files. In this way, `ShowPlanOnDiff` is a bit different, because it never updates any of the pre-existing files, but it is kind of a `Debug` setting--one that you wouldn't want to run normally.
2. The annotation has a `value` that it sets
3. When choosing the set of tests to run, it filters out the `maintainConfigs` that don't align with its value

There is still something about this that feels a bit wrong, design wise. (I agree that it achieves the intended result.) Like, it feels like the annotation choice should result in the appropriate config being created, rather than filtering out all of the configs that aren't the one we want. It's also a bit strange that it happens when the configuration is invoked rather than when the configuration is created. But I see what you're getting at here.

alecgrieser (approved):

PR #3128, opened 2025-02-10
Change context options to a type-safe option with generics.

ohadzeliger:
Pay technical debt from https://github.com/FoundationDB/fdb-record-layer/pull/3075: make context options type safe.
The options are now a simplified version of similar option implementations: An Option class and a wrapper around a map of the options allows us to remove the type cast when getting options. This was done this way (as opposed to a specific class with getters) in order to support the merging of options in `ConfigWithOptions`.


ohadzeliger:
Closed and replaced by https://github.com/FoundationDB/fdb-record-layer/pull/3132

PR #3130, opened 2025-02-10
Make update_release_notes.bash executable

alecgrieser:
The most recent release build (for [4.1.3.0](https://github.com/FoundationDB/fdb-record-layer/actions/runs/13246575078)) failed to update the release notes, because the script is't executable. This updates the permission on the file so that it can be executed. We could alternatively invoke it via `bash` instead of directly, but this way, we get the expected shell and flags passed in from the `#!` at the top of the file


alecgrieser:
Closing in favor of #3131

PR #3132, opened 2025-02-10
Change context options to a type-safe option with generics.

ohadzeliger:
Pay technical debt from https://github.com/FoundationDB/fdb-record-layer/pull/3075: make context options type safe.
The options are now a simplified version of similar option implementations: An Option class and a wrapper around a map of the options allows us to remove the type cast when getting options. This was done this way (as opposed to a specific class with getters) in order to support the merging of options in ConfigWithOptions.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestConfigFilters.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestConfigFilters.java
@@ -80,6 +80,15 @@ public enum YamlTestConfigFilters {
             return config.getRunnerOptions().getOrDefault(YamlExecutionContext.OPTION_CORRECT_EXPLAIN, false) &&
                     config.getRunnerOptions().getOrDefault(YamlExecutionContext.OPTION_CORRECT_METRICS, false);
         }
+    },
+    /**
+     * Used to show the plan diffs graphically.
+     */
+    SHOW_PLAN_ON_DIFF {
+        @Override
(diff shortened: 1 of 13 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestConfigFilters.java
```diff
@@ -77,17 +77,8 @@ boolean filter(final YamlTestConfig config) {
...
         @Override
         boolean filter(final YamlTestConfig config) {
-            return (boolean)config.getRunnerOptions().getOrDefault(YamlExecutionContext.OPTION_CORRECT_EXPLAIN, false) &&
-                    (boolean)config.getRunnerOptions().getOrDefault(YamlExecutionContext.OPTION_CORRECT_METRICS, false);
-        }
-    },
-    /**
-     * Used to show the plan diffs graphically.
-     */
-    SHOW_PLAN_ON_DIFF {
```
ScottDugas, line 87: Why are you removing `SHOW_PLAN_ON_DIFF` ?

ScottDugas (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestConfigFilters.java
ohadzeliger: Sorry about that. Fixed.

ScottDugas (approved):

PR #3137, opened 2025-02-11
Resolve #3135: Allow scrubbing of indexes in READABLE_UNIQUE_PENDING state

### LINKED ISSUE #3135: Allow scrubbing of indexes in READABLE_UNIQUE_PENDING state
The index scrubbing should treat Indexes in this state as READABLE. 
If uniqueness violation will ever be a scrubbing feature (not likely), distinguishing the two states should be handled by the relevant `IndexScrubbingTool` implementation.

jjezra:
Indexes in `READABLE_UNIQUE_PENDING` state are fully built, and hence can be scrubbed.

   If uniqueness violation will ever become a scrubbing feature (not likely), distinguishing the two states should be handled by the relevant `IndexScrubbingTool` implementation.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
@@ -391,6 +391,7 @@ public class IndexingThrottle {
                 return;
             }
             throw new IndexingBase.UnexpectedReadableException(false, "Attempt to scrub a non readable index",
+                    LogMessageKeys.INDEX_NAME, common.getTargetIndexesNames(),
                     LogMessageKeys.INDEX_STATE, indexStates);
         }
         // Here: index building
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
@@ -62,7 +62,7 @@ public class IndexingThrottle {
     @Nonnull private static final Logger LOGGER = LoggerFactory.getLogger(IndexingThrottle.class);
     @Nonnull private final IndexingCommon common;
     @Nonnull private final Booker booker;
-    private final IndexState expectedIndexState;
+    private final boolean isScrubber;
     private Set<Index> mergeRequiredIndexes = new HashSet<>();
 
     static class Booker {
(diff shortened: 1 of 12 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -385,26 +385,34 @@ public <R> CompletableFuture<R> buildCommitRetryAsync(@Nonnull final BiFunction<
...
     private void expectedIndexStatesOrThrow(FDBRecordStore store, FDBRecordContext context) {
         List<IndexState> indexStates = common.getTargetIndexes().stream().map(store::getIndexState).collect(Collectors.toList());
-        if (indexStates.stream().allMatch(state -> state == expectedIndexState)) {
+        if (isScrubber) {
+            // index scrubbing requires a scannable state
+            if (indexStates.stream().allMatch(IndexState::isScannable)) {
+                return;
+            }
+            throw new IndexingBase.UnexpectedReadableException(false, "Attempt to scrub a non readable index",
+                    LogMessageKeys.INDEX_STATE, indexStates);
```
ScottDugas, line 394: It seems like this should include the index names too, otherwise you won't know which index is in the wrong state.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
```diff
@@ -385,26 +385,34 @@ public <R> CompletableFuture<R> buildCommitRetryAsync(@Nonnull final BiFunction<
...
+        if (indexStates.stream().allMatch(IndexState::isWriteOnly)) {
             return;
         }
         // possible exceptions:
         // 1. All the indexes are now readable.
         // 2. Some indexes are built, but all the others are in the expected state.
         // 3. Some indexes are not in the expected state (disabled?).
         // During mutual indexing, the first two may be part of the valid path
-        if (indexStates.stream().allMatch(state -> (state == IndexState.READABLE || state == IndexState.READABLE_UNIQUE_PENDING))) {
+        if (indexStates.stream().allMatch(IndexState::isScannable)) {
```
ScottDugas, line 405: Does the transition from `READABLE_UNIQUE_PENDING` to `READABLE` happen before this code, such that if you try to build indexes that are `READABLE_UNIQUE_PENDING` it will succeed? `testReadableUniquePendingSingle` implies that it should work.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingThrottle.java
jjezra: Yes. 
The `DesiredAction` for indexes in this state is `DesiredAction.MARK_READABLE`. This was designed to save resources when an index with uniqueness violation is attempted to be built.

ScottDugas (approved):

PR #3138, opened 2025-02-11
Support multiple server versions in multi-version tests

ohadzeliger:
This enables the execution of multi-version tests against various server versions. Assuming the availability of multiple server versions, this PR achieves the following:
- Reuses servers across different configurations, ensuring each version is run only once.
- Starts all servers located in the download directory.
- Runs all multi-server tests with all available servers.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
@@ -34,6 +34,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
@@ -72,6 +73,11 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
         } else {
             AtomicInteger serverPort = new AtomicInteger(1111);
             List<File> jars = ExternalServer.getAvailableServers();
+            // Fail the test if there are no available servers. This would force the execution in "runQuick" mode in case
+            // we don't have access to the artifacts.
+            // Potentially, we can relax this a little if all tests are disabled for multi-server execution, but this is
+            // not a likely scenario.
+            Assertions.assertFalse(jars.isEmpty(), "There are no external servers available to run");
             servers = jars.stream().map(jar -> new ExternalServer(jar, serverPort.getAndIncrement(), serverPort.getAndIncrement())).collect(Collectors.toList());
             for (ExternalServer server : servers) {
                 server.start();
@@ -111,11 +117,15 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
                         return e;
                     }
                 }).filter(Objects::nonNull).findFirst();
+        for (ExternalServer server : servers) {
+            try {
+                server.stop();
+            } catch (Exception ex) {
+                logger.info("Failed to stop server " + server.getVersion() + " on " + server.getPort());
+            }
+        }
         if (exception.isPresent()) {
             throw exception.get();
-        }
-        for (ExternalServer server : servers) {
-            server.stop();
         }
     }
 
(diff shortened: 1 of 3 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -58,21 +62,31 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
...
-                    new EmbeddedConfig(),
-                    new JDBCInProcessConfig(),
-                    new MultiServerConfig(0, 1111, 1112),
-                    new ForceContinuations(new MultiServerConfig(0, 1113, 1114)),
-                    new MultiServerConfig(1, 1115, 1116),
-                    new ForceContinuations(new MultiServerConfig(1, 1117, 1118))
-            );
+            AtomicInteger serverPort = new AtomicInteger(1111);
+            List<File> jars = ExternalServer.getAvailableServers();
+            servers = jars.stream().map(jar -> new ExternalServer(jar, serverPort.getAndIncrement(), serverPort.getAndIncrement())).collect(Collectors.toList());
```
ScottDugas, line 75: Just to be safe, I feel like it would be good to have this fail if `jars` is empty.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -100,6 +114,9 @@ public void afterAll(final ExtensionContext context) throws Exception {
         if (exception.isPresent()) {
             throw exception.get();
         }
+        for (ExternalServer server : servers) {
+            server.stop();
```
ScottDugas, line 118: I think we still want to make sure that it keeps stopping servers even if it fails to stop one.
Probably something similar to the above, or by making `stop` swallow and log exceptions.
We probably also want to stop the servers even if an `afterAll` throws an exception.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
```diff
@@ -88,9 +78,9 @@ public Set<String> getVersionsUnderTest() {
     @Override
     public String toString() {
         if (initialConnection == 0) {
-            return "MultiServer (Embedded then External)";
+            return "MultiServer version " + externalServer.getVersion() + " (Embedded then External)";
```
ScottDugas, line 81: Perhaps a more pithy version, as these go in the test names:
```suggestion
            return "MultiServer (Embedded then " + externalServer.getVersion() + ")";
```

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
```diff
@@ -88,9 +78,9 @@ public Set<String> getVersionsUnderTest() {
...
         } else {
-            return "MultiServer (External then Embedded)";
+            return "MultiServer version " + externalServer.getVersion() + " (External then Embedded)";
```
ScottDugas, line 83:
```suggestion
            return "MultiServer version (" + externalServer.getVersion() + " then Embedded)";
```

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -84,13 +86,12 @@ public String getVersion() {
...
-            jar = externalServers[0];
+        if (serverJar == null) {
+            final List<File> externalServers = getAvailableServers();
+            Assertions.assertEquals(1, externalServers.size());
+            jar = externalServers.get(0);
         } else {
-            jar = new File(jarName);
+            jar = serverJar;
         }
         Assertions.assertTrue(jar.exists(), "Jar could not be found " + jar.getAbsolutePath());
```
ScottDugas, line 96: Related improvement:
Perhaps we should move 
```
this.version = getVersion(jar)
```
Above starting the process.

ScottDugas (approved):

ScottDugas (changes requested):
Looks like I hit the wrong button, see my last review.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
ohadzeliger: Yeah, good point. Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
ohadzeliger: The `destroy` method does not throw exceptions, but better to be safe. Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/MultiServerConfig.java
ohadzeliger: Done.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
ohadzeliger: Done.

ScottDugas (approved):

PR #3142, opened 2025-02-12
Fixes #3139: Plan Hash Instability in recursive union query

### LINKED ISSUE #3139: Plan Hash Instability in recursive union query
After publishing 4.1.4.0, some of the recursive CTE queries began to fail when run in multi-server configuration mode. Namely:

```yaml
    -
      - query: with recursive c1 as (
            select id, parent from t1 where parent = -1
            union all
            select b.id, b.parent from c1 as a, t1 as b where a.id = b.parent) select id from c1
      - supported_version: 4.1.4.0
      - explain: RUNION q0, q1 { INITIAL { ISCAN(PARENTIDX [EQUALS promote(@c15 AS LONG)]) | INSERT INTO TEMP q1 } RECURSIVE  { ISCAN(CHILDIDX <,>) | FLATMAP q2 -> { TEMP SCAN base() | FILTER _.ID EQUALS q2.PARENT AS q3 RETURN (q2.ID AS ID, q2.PARENT AS PARENT) } | INSERT INTO TEMP q1 }} | MAP (_.ID AS ID)
      - maxRows: 1
```

Which would fail when running the second continuation with a query plan hash mismatch failure:

```
com.apple.foundationdb.relational.api.exceptions.ContextualSQLException: cannot continue query due to mismatch between serialized and actual plan hash
```

Looking into this more, it seems like the root cause is that the plan hash on this query plan isn't stable (some element of it is dependent on something JVM dependent, as it is stable across runs within the same JVM, but it changes every time we switch to a new process).

We didn't catch this before as the multi-version test configs are the only test configs we have that we continue across two different JVMs, and those queries were disabled prior to this most recent version being published.

alecgrieser:
This updates the plan hashes for some of the plan operators associated with recursive CTEs. They previously hashed the quantifier directly, which could include things like enum hashes, which are not stable across JVMs. This resulted in multi-server test failures. This updates the plan hash calculations to be based off of the child _plans_ instead of the child _quantifiers_, which have stable plan hashes.

To test this, I published this change to my maven local repo. I then configured the yaml tester to use that newly published library against my embedded driver, and I validated that the failing tests in `recursive-cte.yamsql` now passed. In order to make this actually mergeable, though, I needed to update the `supported_version` on those two tests to `!current_version`, which means we won't actually get coverage of this until the next release. We could consider adding additional configurations that compare embedded versus external on the same version, which would help cover things like this, but I didn't add that in this PR.

I did update `recursive-cte.yamsql` with query plans, which made it clearer what plans to look at and felt like good things to have. I've checked in the now necessary metrics files as well for those plans.

This fixes #3139.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/TempTableInsertPlan.java
```diff
@@ -184,7 +184,7 @@ public int hashCodeWithoutChildren() {
 
     @Override
     public int planHash(@Nonnull final PlanHashMode mode) {
-        return PlanHashable.objectsPlanHash(mode, BASE_HASH, getTempTableReferenceValue());
+        return PlanHashable.objectsPlanHash(mode, BASE_HASH, getChild());
```
normen662, line 187: This should actually use BOTH the reference value and the child. The reference value is safe (to be used for plan hashes; only quantifiers are not).

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/TempTableInsertPlan.java
alecgrieser: I've added back the `getTempTableReferenceValue()` (and the Boolean `isOwningTemplate`, which I think was an oversight). I'd thought that that was also unstable, but further local tests seem to suggest I was wrong, so either I was comparing the wrong thing or I was comparing this code against hashing the quantifier by mistake. I've re-done the final test (pushing a build to maven local and running the multi-server tests with two JVMs on the latest code), and the plan hashes were stable.

normen662 (approved):

PR #3143, opened 2025-02-12
Fix broken build in Main

ohadzeliger:
Two tests now fail on Main. The tests had a `supported_version` that was enabled once a release was cut but a plan hash change now make them fail when they are run on multi-server config.
This PR reintroduces the `!current_version` for the tests such that these will not run in multi-server mode


alecgrieser:
This is caused by #3131 and should be addressed by #3142. We can take this now as there is at least one open question on that PR.

alecgrieser (approved):

PR #3148, opened 2025-02-13
Don't include empty :destructiveTest in Test Summary

ScottDugas:
The test summary is nice, but most subprojects don't have anything run as part of `:destructiveTest`, so don't put suites in the summary that have nothing.


gradle/testing.gradle
```diff
@@ -189,7 +189,10 @@ tasks.withType(Test) { theTask ->
         }
     }
     afterSuite { descriptor, result ->
-        if (descriptor.parent == null) {
+        // Very few subprojects acutally have destructive tests, so strip those out of the summary.
+        // We leave the ones that have nothing under test, because that would be suspicious if a subproject had
+        // no tests
+        if (descriptor.parent == null && !(descriptor.toString().endsWith(":destructiveTest") && result.getTestCount() == 0)) {
```
alecgrieser, line 195: This seems fine, but do we want to just exclude any result with 0 tests? It's a bit simpler, though we would lose information if, say, someone forgot to add any tests to a given package

alecgrieser (approved):

gradle/testing.gradle
ScottDugas: Yeah, I commented above the code itself, I think it would be suspicious if a subproject had 0 tests, and probably an indication that someone made a mistake, so I wanted to make it easier to see.

PR #3149, opened 2025-02-13
Backport build changes to allow for 4.0.559 patch release

alecgrieser:
This takes all of the build changes necessary to get us to publish a new 4.0.559 patch release to maven central. They include:

1. Copying verbatim the versions of the `.github`, `actions`, and `build` directories from main. This gets us the most up to date actions, including the build-and-release action with the proper arguments, as well as updated scripts to increment the version, update the yamsql files, and push the release notes
2. Updated gradle changes to publish artifacts to GitHub and Maven Central. The latter includes signing work and the new nexus publishing plugin.
3. Updates to make the `yaml-tests` choose an appropriate version for multi-version testing

This is perhaps a bit more involved than we'd like, and so we could say that our official policy is to use the old flow for any 4.0 patches, and then only switch over to the new version for subsequent patches.


alecgrieser:
We've decided not to take this back to 4.0. We'll fall back to using our old build system if anything requires a 4.0 patch release

PR #3147, opened 2025-02-13
Publish a consolidated JaCoCo report from multiple subprojects and split test jobs

MMcM:
* Stop creating per-subproject JaCoCo reports.
* Add a new top-level Gradle task to take the outputs of build / test and generate a single report.
* Publish `.exec` files out of both `-test` jobs.
* Add a fourth `pull-request` workflow job that downloads those artifacts and generates a report.
* Adjust the separate `workflow_run` workflow (needed to get `pull_request` commenting access) to use that report.

PR #3150, opened 2025-02-13
Coverage comment experiment

alecgrieser:
I noticed that all of the comments are being sent to #3147, the change which introduced `pull_request_comments.yml`, which I think is a mistake. To avoid that, this tries to use a job within the PR

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -12,7 +12,10 @@ jobs:
   #  2. core-tests: Runs the fdb-record-layer-core tests. This is the longest test suite of the
   #     various subprojects, so separating it out allows us to speed up the PRB time.
   #  3. other-tests: Runs the rest of the tests. This tests everything else
-  #  4. coverage: Merges the JaCoCo output of 2 and 3 and generates reports.
+  #
+  # Once those complete, we have two jobs to handle publishing coverage information:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -139,3 +142,30 @@ jobs:
...
           path: |
             ${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/
+
+  coverage-comment:
+    runs-on: ubuntu-latest
+    needs: coverage
+    permissions:
+      checks: write
+      contents: read
+      pull-requests: write
```
MMcM, line 152: I don't think that this will work: `pull_request` workflows have their permissions always limited, since they are running code from the PR itself. (There is `pull_request_target`, but the protection is intentional.)

If the `pull_request` triggering the `workflow_run` isn't working properly, then we could write the PR number into an artifact and read from there in the `workflow_run` workflow, passing the number to the comment action. See https://github.com/Madrapps/jacoco-report/blob/main/src/action.ts#L92-L101 for how it tries to figure it out -- which is evidently not working -- and the available override.

alecgrieser:
This ended up not working out, for reasons @MMcM alludes to. Thanks to #3158, we're now threading through the PR number a different way, which gets us what we want.

PR #3151, opened 2025-02-13
Exclude protogen from code coverage reports

arnaud-lacurie:


gradle/root.gradle
```diff
@@ -85,7 +85,8 @@ tasks.register('codeCoverageReport', JacocoReport) {
     def exclusions = [
         "**/*Proto.class", "**/*Proto\$*.class",
         "**/planprotos/*", "**/eventprotos/*",
-        "**/generated/*", "**/grpc/v1/**/*"
+        "**/generated/*", "**/grpc/v1/**/*",
+        "**/protogen/*"
```
MMcM, line 89: These are class / package names, not build source directory names. What classes have `protogen` in their package name?

arnaud-lacurie: Ha! In that case, this won't work. 
This is from `com.apple.foundationdb.relational.continuation`. 
Basically the classes defined in https://github.com/FoundationDB/fdb-record-layer/blob/main/fdb-relational-core/src/main/proto/continuation.proto

MMcM: I guess there isn't anything more distinctive than the class name itself. Though perhaps renaming it to have `Proto` in it would be worth considering.

PR #3152, opened 2025-02-13
Kill abandoned external servers started during Yaml Tests

ScottDugas:
This adds a static constructor that will find all processes matching the directory for external servers and kill them.
This is most important when running/debugging locally, if you stop the debugger, or SIGKILL, there will be a bunch of processes left behind. This changes it so that we will kill them the next time you run the test.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
@@ -76,7 +76,9 @@ public class ExternalServer {
                                         argument.startsWith(serverPath)))
                         .orElse(false))
                 .forEach(process -> {
-                    logger.info("Killing existing server: pid=" + process.pid() + " " + process.info());
+                    if (logger.isInfoEnabled()) {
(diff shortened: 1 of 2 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -66,6 +67,22 @@ public ExternalServer(File serverJar, final int grpcPort, final int httpPort) {
         this.httpPort = httpPort;
     }
 
+    static {
```
ohadzeliger, line 70: This may "catch" other unrelated processes. Can we maybe:
- add "java" as the process name
- only turn this on in non-CI environments

ScottDugas: Restricting it to `java` makes sense. It can't be strict equality, at least locally, it's a long path, and then `/java`.
We don't (currently) have a way to know that we are in CI, but they are all run in docker images, so there shouldn't really be anything running that we didn't start, so I don't feel like we need to take in that change here.

ohadzeliger (approved):

Issue #3153, opened 2025-02-13
Server jar does not have Built-Date in the MANIFEST in the jar


ScottDugas:
We are supposed to be including the `Built-Date` and `Built-By` in the manifest in our jars, but looking at the recent server-all jars, it is not included.

PR #3154, opened 2025-02-13
Add gradle task to test against many versions

ScottDugas:
This adds a new gradle task: `mixedModeTest`, when you run this it will download the last 50 releases of the server, and run the mixed mode tests with each one against the current version.
The choice of 50 is arbitrary, iI think we will iterate once this is used. We may also want to check the date, but our jars are not currently including that (#3153). 
Note: Since 4.0.559.0 is the first version that includes, it won't include anything from before then.
Right now, it runs against 19 versions.
This doesn't actually start us to run this as part of any workflow, I am intending to add a new workflow to consume it


ScottDugas:
Replaced with #3166

PR #3156, opened 2025-02-13
Resolves #3155: FRL respects PLAN_CACHE_*_MAX_ENTRIES

### LINKED ISSUE #3155: FRL does not respect PLAN_CACHE_*_MAX_ENTRIES options
There are a few options for controlling plan cache sizes, but FRL does not pass them along.
- `PLAN_CACHE_PRIMARY_MAX_ENTRIES`
- `PLAN_CACHE_SECONDARY_MAX_ENTRIES`
- `PLAN_CACHE_TERTIARY_MAX_ENTRIES`

ScottDugas:
I discovered that running a lot of YamlTestConfigs against the same Embedded server can result in exhausting a heap of 512MB. The main cause (at least 300mb) was the plan cache, but there was no way to control the size, so this passes those options along when running `FRL`.
`TransactionBoundEmbeddedRelationalEngine` already passes these options along.

A limit of 10 should cover any tests we have, and any short-term parallesim, or tests with multiple schema templates.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -32,7 +32,7 @@ Users performing online updates are encouraged to update from [4.0.559.4](#40559
 * **Feature** Feature 1 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
 * **Feature** Add enum column support to relational server [(Issue #3073)](https://github.com/FoundationDB/fdb-record-layer/issues/3073)
 * **Feature** Allow scrubbing of indexes in READABLE_UNIQUE_PENDING state [(Issue #3135)](https://github.com/FoundationDB/fdb-record-layer/issues/3135)
-* **Feature** Feature 4 [(Issue #NNN)](https://github.com/FoundationDB/fdb-record-layer/issues/NNN)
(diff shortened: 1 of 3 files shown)


g31pranjal (approved):
LGTM! 

Just a small query: Does this also control the cache size for external server as well? If not do you think we should do that (can it be done?) just as a safety net?

ScottDugas:
I don't think this helps with the server: https://github.com/FoundationDB/fdb-record-layer/blob/119b268e598b0dbf5024a2a175db93a6b14cbf34/fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/RelationalServer.java#L137
I think setting values for these for the server would be good. I don't think it currently has any mechanism for controls.
Running locally, https://github.com/FoundationDB/fdb-record-layer/pull/3154,  did not run into any issues on the external server, perhaps because it leaves the ttls as the defaults (https://github.com/FoundationDB/fdb-record-layer/blob/119b268e598b0dbf5024a2a175db93a6b14cbf34/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java#L209) so it doesn't end up accruing thousands of entries.

I think, if that doesn't work well, perhaps we should revisit the defaults. While the server is very far from production-ready, I'm reticent to have it have a special configuration.

PR #3157, opened 2025-02-13
Backport prepared statement changes to the patch release

ohadzeliger:
Backport parts of PR #3116 in order to allow prepared statement multi-version tests to run against the 4.0.459 release.
This only takes in the JDBC server pieces that are required to allow prepared statement to run on JDBC, so that newer versions of the project can test against this server to prove backwards compatibility.
The original PR had additional changes to the client side that are not required for this purpose and so were left out of this PR.

### DIFF
--- a/fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
+++ b/fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
@@ -241,6 +241,64 @@ public class TypeConversion {
         return Struct.newBuilder().setColumns(listColumnBuilder.build()).build();
     }
 
+    /**
+     * Return the Java object stored within the proto.
+     * @param columnType the type of object in the column
+     * @param column the column to process
+     * @return the Java object from the Column representation
+     * @throws SQLException in case of an error
+     */
+    public static Object fromColumn(int columnType, Column column) throws SQLException {
+        switch (columnType) {
+            case Types.ARRAY:
+                checkColumnType(columnType, column.hasArray());
+                return fromArray(column.getArray());
+            case Types.BIGINT:
+                checkColumnType(columnType, column.hasLong());
+                return column.getLong();
+            case Types.INTEGER:
+                checkColumnType(columnType, column.hasInteger());
+                return column.getInteger();
+            case Types.BOOLEAN:
+                checkColumnType(columnType, column.hasBoolean());
+                return column.getBoolean();
+            case Types.VARCHAR:
+                checkColumnType(columnType, column.hasString());
+                return column.getString();
+            case Types.BINARY:
+                checkColumnType(columnType, column.hasBinary());
+                return column.getBinary().toByteArray();
+            case Types.DOUBLE:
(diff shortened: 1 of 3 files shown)


fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -241,6 +241,64 @@ private static Struct toStruct(RelationalStruct relationalStruct) throws SQLExce
...
+            case Types.BINARY:
+                checkColumnType(columnType, column.hasBinary());
+                return column.getBinary().toByteArray();
+            case Types.DOUBLE:
+                checkColumnType(columnType, column.hasDouble());
+                return column.getDouble();
+            default:
+                // NULL (java.sql.Types value 0) is not a valid column type for an array and is likely the result of a default value for the
+                // (optional) array.getElementType() protobuf field.
+                throw new SQLException("java.sql.Type=" + columnType + " not supported", ErrorCode.CANNOT_CONVERT_TYPE.getErrorCode());
```
ScottDugas, line 277: I noticed you have `CANNOT_CONVERT_TYPE`, is that because you wanted to pick something that already existed in this branch, and not change `ErrorCode`

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
```diff
@@ -230,11 +232,14 @@ private static void addPreparedStatementParameter(RelationalPreparedStatement re
                 relationalPreparedStatement.setString(index, parameter.getParameter().getString());
                 break;
             case Types.BIGINT:
-                relationalPreparedStatement.setInt(index, parameter.getParameter().getInteger());
```
ScottDugas, line 233: You can't change `prepared.yamsql` without the client-side changes, right? So the test has to stay disabled. Have you run the latest client against this build?

ScottDugas (changes requested):

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
ohadzeliger: Yes. Enabling the test on the patch release would require the client side changes that would require changes to the `fob-relational-api` module. As discussed, it is not required at this point, so was left out of the back port.
I tested the changes locally with a manually built revision of this branch's server, where the `supported_version` was manually set (see below). Note that newer versions of the `prepared` test will not pass since we haven't ported the change to all of them.
image
image

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
ohadzeliger: Yes, good catch. The `ErrorCode` is in the `fdb-relational-api` module and I wanted to avoid making any. change that could risk the production code. We are not asserting on the server return code in any of the tests so this seemed like a good trade off.

ScottDugas (approved):

PR #3161, opened 2025-02-14
Add support for running YAML tests against a server built from local branch

ohadzeliger:
In order to increase coverage - specifically, running the current version on multiple JVMs - this change adds the ability to build and run multi-server tests against the current local version.
The test run will treat the locally-built server the same as any other downloaded server. That means that 4 test runs will be performed:
- embedded then remote (with and without continuation forcing)
- remote then embedded (with and without continuation forcing)

This effectively adds one more downloaded server to the list.
An option was considered (and can be quite easily implemented) to reduce coverage by treating this server differently (and only run a subset of the configs) but given the large number of servers we already plan to download anyway, this didn't seem material.


yaml-tests/yaml-tests.gradle
```diff
@@ -128,8 +128,16 @@ task downloadExternalServer(type: Copy) {
     into project.layout.buildDirectory.dir('externalServer')
 }
 
+task buildLocalServer(type: Copy) {
+    dependsOn ":fdb-relational-server:package", "downloadExternalServer"
```
ScottDugas, line 132: Why does this depend on `downeladExternalServer`?
Should it depend on `cleanExternalServerDirectory` instead?

yaml-tests/yaml-tests.gradle
```diff
@@ -128,8 +128,16 @@ task downloadExternalServer(type: Copy) {
...
+    from rootProject.project("fdb-relational-server").layout.files('.dist')
+    include("**/*-all.jar")
```
ScottDugas, line 134: Should this be more specific instead of `*-all.jar`?

ScottDugas (changes requested):

yaml-tests/yaml-tests.gradle
ohadzeliger: Changed to be tighter.

yaml-tests/yaml-tests.gradle
ohadzeliger: I was trying to avoid a potential race condition where the cleanup happened after one of the defendants. I think this new task is better arranged.

ScottDugas: The race condition is not possible, gradle is creating a dependency graph, and will only run `cleanExternalServerDirectory` once.
That being said, the new code is cleaner.

ScottDugas (approved):

PR #3168, opened 2025-02-17
`RecordQueryPlanComplexityException` now includes only a shortened plan string as a log message key

### LINKED ISSUE #3167: RecordQueryPlanComplexityException includes the full plan string
The `RecordQueryPlanComplexityException` includes the full plan string in the error message. This is bad for two reasons:

1. It eschews are preferred practice of having a generic error message and putting details in the error keys and values. This approach allows us to have a small set of errors that can be aggregated together, and then it allows additional filtering to be done on the log message keys. (This is at the cost of the adopter having to know that they should log the keys.)
2. The plan strings can be very long. This is particularly the case because the `RecordQueryPlanComplexityException` is only likely to be logged when we have a very complex plan. This pollutes log storage systems and makes them hard to sift through

alecgrieser:
This updates the plan string in the default `RecordQueryPlanComplexityException` to include only the first 1k characters in the plan string. This decreases the total size of the data pushed to the exception for logging. It also updates the `RecordQueryPlanComplexityException` class so that it extends `RecordQueryException` and uses the log message keys now available in order to convey information instead of relying updating the error message with each error.

This resolves #3167.


hatyo (approved):
LGTM.

PR #3170, opened 2025-02-17
Resolves #3094: Improve multi-server connection testing

### LINKED ISSUE #3094: MultiServerConnectionFactory does not use the provided initialConnection for the first call
Building on the work here: https://github.com/FoundationDB/fdb-record-layer/pull/3083
I tried to change `MultiServerConnectionFactoryTest` so that it returns a mock statement, and assert about the statements that it returns, rather than asserting about the counters being used.

When I did this I added an assertion that the first statement comes from the `initialConnection`, and that failed.
I think it ends up incrementing weirdly.
I tried to get it to work, but kept failing.

Also, of note, the alternating may be confusing in the case where you have a variety of different calls to continuations.

ohadzeliger:
Add tests that assert on the returned statements from the multi-server-connection
Change connection logic to keep same current selector on read operations

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
@@ -174,12 +174,12 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
 
         @Override
         public RelationalStatement createStatement() throws SQLException {
-            return getNextConnection("createStatement").createStatement();
+            return getNextConnection(true, "createStatement").createStatement();
         }
 
         @Override
         public RelationalPreparedStatement prepareStatement(String sql) throws SQLException {
-            return getNextConnection("prepareStatement").prepareStatement(sql);
+            return getNextConnection(true, "prepareStatement").prepareStatement(sql);
         }
 
         @Override
@@ -195,7 +195,7 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
 
         @Override
         public boolean getAutoCommit() throws SQLException {
-            return getNextConnection("getAutoCommit").getAutoCommit();
+            return getNextConnection(false, "getAutoCommit").getAutoCommit();
         }
 
         @Override
@@ -218,32 +218,35 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
 
         @Override
         public boolean isClosed() throws SQLException {
-            return getNextConnection("isClosed").isClosed();
+            return getNextConnection(false, "isClosed").isClosed();
         }
 
         @Override
         public DatabaseMetaData getMetaData() throws SQLException {
-            return getNextConnection("getMetaData").getMetaData();
+            return getNextConnection(false, "getMetaData").getMetaData();
         }
 
         @Override
         public void setTransactionIsolation(int level) throws SQLException {
-            getNextConnection("setTransactionIsolation").setTransactionIsolation(level);
+            logger.info("Sending operation {} to all connections", "setTransactionIsolation");
+            for (RelationalConnection connection : relationalConnections) {
+                connection.setTransactionIsolation(level);
+            }
         }
 
         @Override
         public int getTransactionIsolation() throws SQLException {
-            return getNextConnection("getTransactionIsolation").getTransactionIsolation();
+            return getNextConnection(false, "getTransactionIsolation").getTransactionIsolation();
         }
 
         @Override
         public Array createArrayOf(String typeName, Object[] elements) throws SQLException {
-            return getNextConnection("createArrayOf").createArrayOf(typeName, elements);
+            return getNextConnection(true, "createArrayOf").createArrayOf(typeName, elements);
         }
 
         @Override
         public Struct createStruct(String typeName, Object[] attributes) throws SQLException {
-            return getNextConnection("createStruct").createStruct(typeName, attributes);
+            return getNextConnection(true, "createStruct").createStruct(typeName, attributes);
         }
 
         @Override
@@ -256,13 +259,13 @@ public class MultiServerConnectionFactory implements YamlRunner.YamlConnectionFa
 
         @Override
         public String getSchema() throws SQLException {
-            return getNextConnection("getSchema").getSchema();
+            return getNextConnection(false, "getSchema").getSchema();
         }
 
         @Nonnull
         @Override
         public Options getOptions() {
-            return getNextConnection("getOptions").getOptions();
+            return getNextConnection(false, "getOptions").getOptions();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
...
+        RelationalPreparedStatement statement = connection.prepareStatement("SQL");
+        assertStatement(statement, path[initialConnection]);
+        // next statement
+        statement = connection.prepareStatement("SQL");
+        assertStatement(statement, path[(initialConnection + 1) % 2]);
+
+        // Second run:
+        // - Factory current connection: alternate connection
+        // - connection current connection: alternate connection
+        // - statement: initial connection
```
ohadzeliger, line 105: Oops - type. This should be alternate

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
...
-        assertEquals(0, connection.getCurrentConnectionSelector());
+
+        MultiServerConnectionFactory.MultiServerRelationalConnection connection =
+                (MultiServerConnectionFactory.MultiServerRelationalConnection)classUnderTest.getNewConnection(URI.create("Blah"));
         assertEquals(initialConnection, classUnderTest.getCurrentConnectionSelector());
+        assertConnection(connection, 0, "Primary");
+        RelationalPreparedStatement statement = connection.prepareStatement("SQL");
+        assertStatement(statement, "Primary");
+        statement = connection.prepareStatement("SQL");
+        assertStatement(statement, "Primary");
```
ScottDugas, line 57: It feels like it would be cleaner to inline the `statement` variable, so that this becomes more like:
```
        assertStatement(connection.prepareStatement("SQL"), "Primary");
        assertStatement(connection.prepareStatement("SQL"), "Primary");
```
and then you can more easily parse the alternations.
You could also streamline it a bit more to have:
```
assertStatement(connection, "Primary");
```

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
...
-                List.of(dummyConnectionFactory()));
+                dummyConnectionFactory("Primary"),
+                List.of(dummyConnectionFactory("Alternate")));
         assertEquals(initialConnection, classUnderTest.getCurrentConnectionSelector());
-        MultiServerConnectionFactory.MultiServerRelationalConnection connection = (MultiServerConnectionFactory.MultiServerRelationalConnection)classUnderTest.getNewConnection(URI.create("Blah"));
-        assertEquals(0, connection.getCurrentConnectionSelector());
+
+        MultiServerConnectionFactory.MultiServerRelationalConnection connection =
+                (MultiServerConnectionFactory.MultiServerRelationalConnection)classUnderTest.getNewConnection(URI.create("Blah"));
         assertEquals(initialConnection, classUnderTest.getCurrentConnectionSelector());
```
ScottDugas, line 52: Are the assertions about `getCurrentConnectionSelector()` necessary anymore? I think asserting about the created statements is more trustworthy.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
...
+        // First run:
+        // - Factory current connection: initial connection
+        // - connection current connection: initial connection
+        // - statement: initial connection
+        MultiServerConnectionFactory.MultiServerRelationalConnection connection =
+                (MultiServerConnectionFactory.MultiServerRelationalConnection)classUnderTest.getNewConnection(URI.create("Blah"));
         assertEquals((initialConnection + 1) % 2, classUnderTest.getCurrentConnectionSelector());
+        assertConnection(connection, initialConnection, path[initialConnection]);
+        RelationalPreparedStatement statement = connection.prepareStatement("SQL");
+        assertStatement(statement, path[initialConnection]);
```
ScottDugas, line 97: I feel like it would be clearer to have the names flip above, so that all of these can just be
```
assertStatement(connection.prepareStatement("SQL"), "Initial");
assertStatement(connection.prepareStatement("SQL"), "Other");
assertStatement(connection.prepareStatement("SQL"), "Initial");
```

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
...
+        // - statement: initial connection
         connection = (MultiServerConnectionFactory.MultiServerRelationalConnection)classUnderTest.getNewConnection(URI.create("Blah"));
-        assertEquals((initialConnection + 1) % 2, connection.getCurrentConnectionSelector());
         assertEquals(initialConnection % 2, classUnderTest.getCurrentConnectionSelector());
+        assertConnection(connection, (initialConnection + 1) % 2, path[(initialConnection + 1) % 2]);
+        statement = connection.prepareStatement("SQL");
+        assertStatement(statement, path[(initialConnection + 1) % 2]);
+        // next statement
+        statement = connection.prepareStatement("SQL");
+        assertStatement(statement, path[initialConnection]);
```
ScottDugas, line 113: At least one of these should have 3 or 1 calls to `prepareStatement`

ScottDugas (changes requested):

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
```diff
@@ -42,61 +43,119 @@ void testDefaultPolicy(int initialConnection) throws SQLException {
         MultiServerConnectionFactory classUnderTest = new MultiServerConnectionFactory(
                 MultiServerConnectionFactory.ConnectionSelectionPolicy.DEFAULT,
                 initialConnection,
-                dummyConnectionFactory(),
-                List.of(dummyConnectionFactory()));
+                dummyConnectionFactory("Primary"),
```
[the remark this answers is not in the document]
ohadzeliger: Yes, I wanted the connection to actually create the statements, and since the statement is linked to the connection, I would verify that the correct underlying connection was used, which is why this test is useful.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
```diff
@@ -275,10 +278,19 @@ public void setOption(Options.Name name, Object value) throws SQLException {
...
+        /**
+         * Get the next connection to send requests to.
+         * This method conditionally advances the connection selector. This is done for ease of testing, where some
+         * commands that just return information will not advance the selector. This leads to better predictability
+         * of the command routing.
+         * @param advance whether to advance the connection selector
+         * @param op the name of the operation (for logging)
+         * @return the underlying connection to use
+         */
+        private RelationalConnection getNextConnection(boolean advance, String op) {
```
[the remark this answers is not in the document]
ohadzeliger: Renamed to the better `getCurrentConnection`, but left the boolean selector

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Done (first option, wanted to keep statement visible at this level too)

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: These can be removed. Done.

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Done (I think what you meant was to assign names to the `path[initial]` and `path[initial+1]` values)

yaml-tests/src/test/java/MultiServerConnectionFactoryTest.java
ohadzeliger: Good point, done.

ScottDugas (approved):

PR #3173, opened 2025-02-18
Put Build/Test/Documentation changes in separate section of release notes

ScottDugas:
This updates .github/release.yml so that it will combine build, test, and documentation changes into a separate section.

I highly doubt this is the final iteration on the release notes generation, even in the very near future, but when you generate notes it takes the config from the tag being generated, so I want something better for the next release than having _all_ the build/test changes under `Other Changes`

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -7,9 +7,6 @@ changelog:
     - title: New Features
       labels:
         - enhancement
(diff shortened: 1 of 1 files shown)


ScottDugas:
Decided not to use github's generate-release-notes feature, see #3174

PR #3174, opened 2025-02-18
Generate release notes automatically from pull requests

ScottDugas:
This alleviates two issues:
1. Developers routinely hit conflicts on their pull requests, requiring merging. The original code had attempted to alleviate this by having multiple different items that could be replaced by category, but it was insufficient.
2. In part due to the above, presumably, a lot of release notes in the past were sparse, by generating it automatically from the PR, this will be much harder.

I considered just using GitHub's `generate-notes` functionality, but that is rather limited. This change doesn't really expand on using that yet, but puts us in a better place to do so. 

This also updates the release workflow to use the new script, and updates the release notes so that they have the new format for every version since `4.0.559.0`. This also means the release notes now include the patch releases of `4.0.559.*`. I also did some manual additions to the release notes.


build/create_release_notes.py
```diff
@@ -0,0 +1,219 @@
...
+    return label_config['catch_all']
+
+def generate_note(prs, commit, label_config):
+    if len(prs) == 0:
+        return ("Direct Commit", commit[1])
+    if len(prs) > 1:
+        print("Too many PRs?")
+    pr = prs[0]
+    category = get_category(pr, label_config, commit)
+    text = '* ' + pr['title'] + " by @" + pr['user']['login'] + ' in ' + pr['html_url']
```
alecgrieser, line 94: I think we need a bit more than this in order to make these work for sphinx. Something like:

```suggestion
    text = '* ' + pr['title'] + ' by [@' + pr['user']['login'] + '](' + pr['user']['html_url'] + ') in [#' + str(pr['number']) + '](' + pr['html_url'] + ')'
```

Which I think would turn a release note like:

```md
* Commit automatic updates from 4.1.6.0 by @alecgrieser in https://github.com/FoundationDB/fdb-record-layer/pull/3163
```

into:

```md
* Commit automatic updates from 4.1.6.0 by [@alecgrieser](https://github.com/alecgrieser) in [#3163](https://github.com/FoundationDB/fdb-record-layer/pull/3163)
```

Which renders in GitHub like:

> * Commit automatic updates from 4.1.6.0 by [@alecgrieser](https://github.com/alecgrieser) in [#3163](https://github.com/FoundationDB/fdb-record-layer/pull/3163)

It doesn't tag users in quite the same way when rendered in the GitHub UI, which may include GitHub releases. But it does link correctly in Sphinx that way (and PRs are fine).

One other thing that is missing in the sphinx doc are links for issues. So a note like:

> * Resolves #3073: Add enum column support to relational server by @ScottDugas in https://github.com/FoundationDB/fdb-record-layer/pull/3074

Does not link to issue #3073 in Sphinx. I think we'd need to parse for the `#\d+` pattern and replace with issue links if we wanted that to work.

build/create_release_notes.py
```diff
@@ -0,0 +1,219 @@
...
+    text += f"\n\n<!-- MIXED_MODE_RESULTS {new_version} PLACEHOLDER -->\n"
+    return text
+
+def replace_note(lines, note):
+    print(f"Inserting note {note.old_version} -> {note.new_version}")
+    new_lines = []
+    added = False
+    for line in lines:
+        if not added and line == note.old_version_header():
+            new_lines.append(note.new_version_header())
```
alecgrieser, line 127:
```suggestion
            new_lines.append(note.new_version_header())
            new_lines.append('')
```

So that we get a new line between new version header and the release notes

build/create_release_notes.py
```diff
@@ -0,0 +1,219 @@
...
+            new_lines.append(note.notes)
+            new_lines.append('')
+            new_lines.append(line)
+            added = True
+        elif not added and note.greater_than_minor_version(line):
+            new_lines.append(note.new_minor_version_header())
+            new_lines.append('')
+            new_lines.append(note.new_version_header())
```
alecgrieser, line 135:
```suggestion
            new_lines.append(note.new_version_header())
            new_lines.append('')
```

build/create_release_notes.py
```diff
@@ -0,0 +1,219 @@
...
+    for new_version in args.new_version:
+        print(f"Generating release notes for {new_version}")
+        commits = get_commits(old_version, new_version)
+        prs = get_prs(commits, args.pr_cache)
+        prs = dedup_prs(prs)
+        new_notes = [generate_note(pr[0], pr[1], label_config) for pr in prs]
+        release_notes.append(Note(old_version, new_version, format_notes(new_notes, label_config, old_version, new_version)))
+        old_version = new_version
+    print("\n\n------------------------------\n\n")
+    if args.release_notes_md is None:
```
alecgrieser, line 211: If we wanted to publish to both the release notes file and to GitHub releases, then I take it we'd want to run both a configuration that adds to the checked in file and also to some other config that it can read separately. I almost feel like this needs an additional argument like `-o` or `--output` that lets it write to a file. In the release workflow, we can then call this with _both_ the `--release-notes-md` flag and `-o` set, then commit the release notes update and use the contents of the output file to power the GitHub release. We could theoretically just run this twice, though if we do, it sounds like we'd need to parse the standard out for the `------------------------------` to find the right lines.

This is also assuming that we want to use GitHub releases, at least in addition to the release notes file.

alecgrieser (changes requested):

build/create_release_notes.py
alecgrieser: Ah, I see. I think that makes sense as well (directly integrating GitHub into the script rather than making it a separate step)

build/create_release_notes.py
```diff
@@ -0,0 +1,219 @@
...
+            header = f"#### {title}"
+            if put_in_summary:
+                text += f"\n<details>\n<summary>\n\n{header} (click to expand)\n\n</summary>\n\n"
+            else:
+                text += f'{header}\n'
+            for note in notes:
+                text += f"{note}\n"
+            if put_in_summary:
+                text += '\n</details>\n'
+    text += f"\n\n**Full Changelog**: https://github.com/FoundationDB/fdb-record-layer/compare/{old_version}...{new_version}"
```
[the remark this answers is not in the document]
alecgrieser: I guess I'm okay with either one. It is kind of nice that with the link, you can see old build number explicitly, I guess. For the release notes file, that probably doesn't matter because the previous release is right there. I could see it going either way for GitHub Releases

build/create_release_notes.py
```diff
@@ -105,16 +106,17 @@ def format_notes(notes, label_config, old_version, new_version):
         if title in grouping:
             notes = grouping[title]
             put_in_summary = category.get('collapsed', False)
-            header = f"#### {title}"
+            # We use <h4> rather than #### because #### confuses sphinx
+            header = f"<h4> {title} </h4>"
             if put_in_summary:
                 text += f"\n<details>\n<summary>\n\n{header} (click to expand)\n\n</summary>\n\n"
```
alecgrieser, line 112: I believe that this doesn't quite work, because the `</h4>` now comes before the `(click to expand)`

alecgrieser (changes requested):

CONTRIBUTING.md
```diff
@@ -82,12 +82,22 @@ This will automatically create an association between the PR and the issue that
...
 
-If your pull request results in a user-visible change to the Record Layer, you should
-also update the [release notes](docs/ReleaseNotes.md). For most changes, it
-is sufficient to fill in one of the bullets in the "next release" section of that
-document. You should include a short description of the change as well as filling in
-the issue number. The "next release" section is commented out, so the change won't
-be visible in our documentation until the next time a release is cut.
+PRs should have titles that clearly indicate what the change, we use these to
+generate release notes, so you can glance at [release notes](docs/ReleaseNotes.md)
+for inspiration.
```
alecgrieser: There's a link problem (pointed out separately), but there are a few other stylin

```suggestion
Pull requests should have titles that clearly indicate what the change is trying to accomplish as we use these to
generate our release notes. You can glance at the [Release Notes](docs/sphinx/source/ReleaseNotes.md)
page for inspiration.
```

CONTRIBUTING.md
```diff
@@ -82,12 +82,22 @@ This will automatically create an association between the PR and the issue that
...
+
+They should also have one of the following labels:
```
alecgrieser, line 89: One thing that's a little unclear is what happens if a PR is given more than one label, which could happen (e.g., a `bug fix` or `feature` that introduces a `breaking change`). Is that allowed? And do the release notes include the PR multiple times?

actions/release-build-publish/action.yml
```diff
@@ -7,15 +7,50 @@ inputs:
...
+
+    # We want to do this before anything else, because if the later steps fail, we want to make sure that the full
+    # change log includes all changes. If a release fails, we should probably update the release notes manually after
+    # the fact, but this is less likely than the build or tests or publishing of the release failing.
+    - name: Update release notes
+      shell: bash
+      run: |
+        python ./build/create_release_notes.py \
+        --config ../fdb-record-layer/build/release-notes-config.json \
+        --release-notes-md docs/ReleaseNotes.md \
```
alecgrieser, line 44:
```suggestion
        --release-notes-md docs/sphinx/source/ReleaseNotes.md \
```

actions/release-build-publish/action.yml
```diff
@@ -7,15 +7,50 @@ inputs:
...
+    # at the beginning
+    - name: Increment version
+      run: python build/versionutils.py gradle.properties --increment --commit -u ${{ inputs.update_type }}
+    - name: Get new version
+      id: get_new_version
+      shell: bash
+      run: |
+        echo "version=$(python build/versionutils.py gradle.properties)" >> "$GITHUB_OUTPUT"
+
+    # Create and push the tag
```
alecgrieser: This isn't quite accurate now that the tag is pushed at a later step

.github/workflows/patch_release.yml
```diff
@@ -27,24 +27,17 @@ jobs:
         uses: ./actions/setup-base-env
       - name: Setup FDB
         uses: ./actions/setup-fdb
-
```
ScottDugas, line 30: I am going to try and test the changes to the actions in a separate repository.

build/create_release_notes.py
```diff
@@ -0,0 +1,234 @@
...
+                    dedupped.append((pr, commit))
+    return dedupped
+
+def get_category(pr, label_config, commit):
+    main_label = None
+    label_names = [label['name'] for label in pr['labels']]
+    for category in label_config['categories']:
+        for label in category['labels']:
+            if label in label_names:
+                return category['title']
```
alecgrieser, line 105: I believe this means that if we tag a PR multiple times, it only appears once (in the first label category in the config that it matches)

build/create_release_notes.py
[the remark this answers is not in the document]
ScottDugas: That's a good point, I changed it to:

```
**[Full Changelog (4.1.4.0...4.1.6.0)](https://github.com/FoundationDB/fdb-record-layer/compare/4.1.4.0...4.1.6.0)**
```

**[Full Changelog (4.1.4.0...4.1.6.0)](https://github.com/FoundationDB/fdb-record-layer/compare/4.1.4.0...4.1.6.0)**

actions/release-build-publish/action.yml
```diff
@@ -7,15 +7,50 @@ inputs:
...
+    - name: Create tag
+      shell: bash
+      run: git tag -m "Release ${{ steps.get_new_version.outputs.version }}" -f "${{ steps.get_new_version.outputs.version }}"
```
alecgrieser, line 35: I think we should move this to after the updating the release notes. If the release notes script fails, then nothing gets pushed (artifacts or commits), so no harm, no foul. However, at the end of the file (post #3078), we check out the tag at the very end, which we need so that we're running from the correct place when we build the documentation. At present, that would mean that the release notes file (published as part of the documentation step) would be missing the most recent update when we go to the docs, and it won't get published until the subsequent release.

alecgrieser: Maybe that means that the script needs to have a mode where instead of comparing tags to get the list of commits, it needs to compare the old tag to the current HEAD? As I think the script now will fail as the tag hasn't actually been pushed anyway

.github/workflows/patch_release.yml
```diff
@@ -27,24 +27,17 @@ jobs:
...
-      # Push a version bump back to main. There are failure scenarios that can result
-      # in published artifacts but an erroneous build, so it's safer to bump the version
-      # at the beginning
       - name: Configure Git
```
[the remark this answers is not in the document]
alecgrieser: I guess I would have been fine with moving this into the composite step in this PR

build/create_release_notes.py
alecgrieser: I think the main one I worry about is "breaking change". Most breaking changes will probably be also either features or bug fixes. In the past, with the hand-rolled release notes, I've done stuff like mention the fix/feature in the appropriate section, and then specified what the "breaking" part is in the breaking change. So you might have something like:

> #### Breaking changes
>
> * Continuations for streaming aggregate queries are no longer compatible with versions below 4.1.8.0 - PR #x
>
> #### Features
>
> * Streaming aggregate query continuations do not lose work if they are stopped by an out of band limit - PR #x

You can structure that as a single thing, but it means that someone looking at the breaking change may only see the bug fix/feature and have to then do more digging to see if they care about the way it is breaking

build/create_release_notes.py
```diff
@@ -165,8 +197,9 @@ def commit_release_notes(filename, new_versions):
 def get_minor_version(version):
     return '.'.join(version.split('.')[:2])
 
-version_header = re.compile('^#+ (\d+(?:\.\d+)+)$') # match only major or minor versions
+version_header = re.compile(r'^#+ (\d+(?:\.\d+)+)$') # match only major or minor versions
```
alecgrieser, line 200: I'm not sure about this comment. Doesn't this match all versions? I think it would need to be:

```py
r'^## (\d+\.\d+)$'
```

To match only major and minor versions. As it is, I think the `version_header` regex is used to match all versions, so I think it's just the comment that's off.

```suggestion
version_header = re.compile(r'^#+ (\d+(?:\.\d+)+)$') # match all version headers, including major or minor headers
```

docs/sphinx/source/ReleaseNotes.md
```diff
@@ -15,63 +15,130 @@ Builds and releases have been moved to a new CI system. This includes the resump
...
+<h4> Bug Fixes </h4>
+
+* Use `Locale.ROOT` in `String.format`, fixing messages if JVM has non-US default locale - [PR #3117](https://github.com/FoundationDB/fdb-record-layer/pull/3117)
+
+<details>
+<summary>
 
+<h4> Build/Test/Documentation Improvements (click to expand) </h4>
 
-#### Mixed Mode Test Results
```
alecgrieser, line 51: Did you mean to delete the 4.1.8.0 mixed mode results in this push?

alecgrieser (changes requested):

alecgrieser:
This mostly LGTM. I left a few comments about style, but the only thing that seems "blocking" is making sure we don't lose the mixed-mode results for 4.1.8.0. There's a bigger discussion about when the least erroneous time to push is, but I also think we can take this to unblock the new release notes process.

docs/sphinx/source/ReleaseNotes.md
ScottDugas: I did not, I will fix that...

ScottDugas: I guess, I should say that I meant to delete it, but I also meant to add it back, which I forgot to do.

alecgrieser (approved):

Issue #3107, opened 2025-02-05
Remove versioning shim code introduced in #3092


normen662:
In #3092 we introduced some code in `RecordQueryStreamingAggregationPlan` to ensure continuing (but incorrect behavior) for previously planned plans that are being continued in a new server. Take that code out again.

PR #3177, opened 2025-02-19
Fix the bug when aggregation stops before a group is finished

pengpeng-lu:
* Problem: in aggregate queries, if a scan is stopped in the middle of a group, we'll just return the current result, which is wrong, and continuation = current row, so the next scan will also return wrong result.
* Solution: doesn't return anything if we stop in the middle of a group, and resets the continuation to the last group, to prepare for the next scan.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
@@ -22,12 +22,14 @@ package com.apple.foundationdb.record.cursors.aggregate;
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.async.AsyncUtil;
+import com.apple.foundationdb.record.ByteArrayContinuation;
 import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.RecordCursorContinuation;
 import com.apple.foundationdb.record.RecordCursorEndContinuation;
 import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordCursorStartContinuation;
 import com.apple.foundationdb.record.RecordCursorVisitor;
+import com.apple.foundationdb.record.provider.foundationdb.KeyValueCursorBase;
 import com.apple.foundationdb.record.query.plan.plans.QueryResult;
 import com.google.common.base.Verify;
 import com.google.protobuf.Message;
@@ -58,13 +60,19 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
     // Previous non-empty record processed by this cursor
     @Nullable
     private RecordCursorResult<QueryResult> previousValidResult;
+    // last row in last group, is null if the current group is the first group
+    @Nullable
+    private RecordCursorResult<QueryResult> lastInLastGroup;
+    byte[] continuation;
 
     public AggregateCursor(@Nonnull RecordCursor<QueryResult> inner,
                            @Nonnull final StreamGrouping<M> streamGrouping,
-                           boolean isCreateDefaultOnEmpty) {
+                           boolean isCreateDefaultOnEmpty,
+                           byte[] continuation) {
         this.inner = inner;
         this.streamGrouping = streamGrouping;
         this.isCreateDefaultOnEmpty = isCreateDefaultOnEmpty;
+        this.continuation = continuation;
     }
 
     @Nonnull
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -280,6 +314,66 @@ void aggregateNoRecordsNoGroupNoAggregate(final boolean useNestedResult) {
...
+            openSimpleRecordStore(context, NO_HOOK);
+
+            final var plan =
+                    new AggregationPlanBuilder(recordStore.getRecordMetaData(), "MySimpleRecord")
+                            .withAggregateValue("num_value_2", value -> new NumericAggregationValue.Sum(NumericAggregationValue.PhysicalOperator.SUM_I, value))
+                            .withGroupCriterion("str_value_indexed")
+                            .build(false);
+
+            // In the testing data, there are 2 groups, each group has 3 rows.
+            // recordScanLimit = 5: scans 3 rows, and the 4th scan hits SCAN_LIMIT_REACHED
```
pengpeng-lu, line 329: not sure why it only scans 4 times instead of 5, but the offset=1 seems consistent.

alecgrieser:
It looks like there's a conflict in the release notes. PRs no longer need to update the release notes file as they will be auto-generated by #3174, so I think the solution is just "take what is on main"

docs/sphinx/source/ReleaseNotes.md
```diff
@@ -20,7 +20,7 @@ Users performing online updates are encouraged to update from [4.0.559.4](#40559
 <h4> New Features </h4>
 
 * FRL respects PLAN_CACHE_*_MAX_ENTRIES - [PR #3156](https://github.com/FoundationDB/fdb-record-layer/pull/3156)
-<h4> Bug Fixes </h4>
+  <h4> Bug Fixes </h4>
```
alecgrieser, line 23: I'm not sure what happened here, but we probably want to back this out and take what's on `main`

alecgrieser (changes requested):

docs/sphinx/source/ReleaseNotes.md
pengpeng-lu: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -57,77 +58,103 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
     // Previous non-empty record processed by this cursor
     @Nullable
     private RecordCursorResult<QueryResult> previousValidResult;
+    // last row in last group, is null if the current group is the first group
+    @Nullable
+    private RecordCursorResult<QueryResult> lastInLastGroup;
+    byte[] continuation;
```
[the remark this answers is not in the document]
alecgrieser: It looks like this still hasn't marked the `continuation` field as private. Is there a reason it needs package visibility?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -57,77 +58,104 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
...
+                * returns result (groupKey1, continuation = row2), and set previousValidResult = row3
+                *
+                * Both scenarios returns the correct result, and continuation are both set to row3 in the end, row2 is scanned twice if a new iteration starts.
+                */
+                previousValidResult = previousResult;
+                return RecordCursorResult.withNextValue(QueryResult.ofComputed(streamGrouping.getCompletedGroupResult()), continuation);
+            } else {
+                if (Verify.verifyNotNull(previousResult).getNoNextReason() == NoNextReason.SOURCE_EXHAUSTED) {
+                    if (previousValidResult == null) {
+                        return RecordCursorResult.exhausted();
```
alecgrieser, line 137: Does this need to check to insert the empty result if `isCreateDefaultOnEmpty`? It feels like that information is being lost here.

Perhaps it would be simpler if there were fewer moving parts here. If we wait until 4.2, we could drop support entirely for `isCreateDefaultOnEmpty` (see #3107) which may simplify this somewhat.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -57,77 +58,104 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
...
+                    RecordCursorContinuation currentContinuation;
+                    // in the current scan, if current group is the first group, set the continuation to the start of the current scan
+                    // otherwise set the continuation to the last row in the last group
+                    if (lastInLastGroup == null) {
+                        currentContinuation = continuation == null ? RecordCursorStartContinuation.START : ByteArrayContinuation.fromNullable(continuation);
+                    } else {
+                        currentContinuation = lastInLastGroup.getContinuation();
+                    }
+                    previousValidResult = lastInLastGroup;
+                    return RecordCursorResult.withoutNextValue(currentContinuation, Verify.verifyNotNull(previousResult).getNoNextReason());
```
alecgrieser, line 153: Should this also set `previousResult`? I think as written, if you call `getNext` a second time after getting one of these out-of-bounds values, you'll get an advanced continuation.

That is:

```java
RecordCursor<T> aggCursor = getAggregateCursor();
RecordCursorResult<T> withValueResult = aggCursor.getNext(); // has a value
RecordCursorResult<T> withoutValueResult1 = aggCursor.getNext(); // first result with out-of-bound limit; continuation borrowed from withValueResult
RecordCursorResult<T> withoutValueResult2 = aggCursor.getNext(); // second result with out-of-bound limit; continuation may be greater than withoutValueResult1
```

In general, once a cursor returns a `RecordCursorResult` without a next value, it should continue to return the same result if `getNext()` is returned again.

alecgrieser (changes requested):

PR #3185, opened 2025-02-21
Fix instability in the query planner

### LINKED ISSUE #3191: Instability in the planner
Cascades is sometimes unstable when planning complex queries that could result in having _valid_ duplications of sub-plans in the memoization structure. For a certain complex query we ended up generating too many binary selects (as a result of join enumeration) and the planner was able to find two variations of the same sub-plan in the memoizer. This is not a problem in itself, however the problem is related to choosing the right variation, which is done non-deterministically causing the planner to go through a different number of planning tasks each time it runs this query.

We were able to observe this thanks to the recent additions of planner metrics validation in YAML in #3113.

hatyo:
This fixes a number of instabilities found in Cascades when planning certain type of queries that end up generating a relatively large number of variations that might end up storing actual duplicates in the memoization structure (which is permissible).

Essentially, the following issues were detected and resolved in this PR:
1. **Fix instability in hash calculation**: Certain expressions had their hash unstable, because it includes hashing an `enum` member, which is [known to be unstable](https://bugs.openjdk.org/browse/JDK-8050217?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel). This issue made it very difficult to compare planner traces from different runs, since a planner trace might include the hash codes of some expressions, and the hash code could change across different JVMs, causing the comparator to report false negatives.
2. **Set instantiation with deterministic traversal order**: Certain instantiations of `Maps.newIdentityHashMap` and `Sets.newHashSet` are replaced with `LinkedIdentityMap`, resp. `LinkedIdentitySet`.
3. **Fix instability detected in Guava `StandardMutableNetwork`**: We use Guava [`Network` ](https://guava.dev/releases/20.0/api/docs/com/google/common/graph/Network.html) for the planner memoization structure, internally, the network uses a hash map to store the parallel edges between network vertices, this is exactly why we were getting the alternatives in different order as explained above. Due to the way the `Network` code is written, it is not possible to pass the (type of the) map to the builder, therefore, I went ahead and copied enough of the code to be able to modify the map construction and make it stable, which effectively fixed the issue, but ideally, Guava should make it possible to control how these maps are created, making it easier to use and reason about the code.
This fixes #3191.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -185,18 +185,18 @@ public boolean insert(@Nonnull final RelationalExpression newValue) {
      *         otherwise.
      */
     private boolean insert(@Nonnull final RelationalExpression newValue, @Nullable final Map<PlanProperty<?>, ?> precomputedPropertiesMap) {
-        Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(Debugger.Location.BEGIN)));
+        Debugger.withDebugger(debugger -> debugger.onEvent( new Debugger.InsertIntoMemoEvent(newValue, Debugger.Location.BEGIN)));
         try {
             final boolean containsInMemo = containsInMemo(newValue);
-            Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(containsInMemo ? Debugger.Location.REUSED : Debugger.Location.NEW)));
+            Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(newValue, containsInMemo ? Debugger.Location.REUSED : Debugger.Location.NEW)));
```
normen662, line 191: Wouldn't we want to have the reference (`this`) as the referring reference?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/Debugger.java
```diff
@@ -919,11 +922,29 @@ public PEvent.Builder toEventBuilder() {
...
+        private final RelationalExpression expression;
+
         @Nonnull
         private final Location location;
 
-        public InsertIntoMemoEvent(@Nonnull final Location location) {
+        @Nonnull
+        private final List<Reference> reusedExpressionReferences;
+
+        public InsertIntoMemoEvent(@Nonnull final RelationalExpression expression,
```
normen662, line 934: Nitpicking here and the other constructor. Can you please make the location come first?

fdb-record-layer-core/src/main/java/com/google/common/graph/StableDirectedMultiNetworkConnections.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * StableDirectedMultiNetworkConnections.java
```
normen662, line 2: Is this preamble ok like this?

fdb-record-layer-core/src/main/java/com/google/common/graph/StableStandardMutableNetwork.java
```diff
@@ -0,0 +1,272 @@
+/*
+ * StableStandardMutableNetwork.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
```
normen662, line 6:
```suggestion
 * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/google/common/graph/StableStandardMutableNetwork.java
```diff
@@ -0,0 +1,272 @@
+/*
+ * StableStandardMutableNetwork.java
```
normen662, line 2: same here

fdb-record-layer-core/src/main/java/com/google/common/graph/StableDirectedMultiNetworkConnections.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
```
normen662, line 6:
```suggestion
 * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/google/common/graph/StableStandardMutableNetwork.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/google/common/graph/StableDirectedMultiNetworkConnections.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/google/common/graph/StableStandardMutableNetwork.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/google/common/graph/StableDirectedMultiNetworkConnections.java
hatyo: Done.

normen662 (changes requested):

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -185,18 +185,23 @@ public boolean insert(@Nonnull final RelationalExpression newValue) {
      *         otherwise.
      */
     private boolean insert(@Nonnull final RelationalExpression newValue, @Nullable final Map<PlanProperty<?>, ?> precomputedPropertiesMap) {
-        Debugger.withDebugger(debugger -> debugger.onEvent(new Debugger.InsertIntoMemoEvent(Debugger.Location.BEGIN)));
+        Debugger.withDebugger(debugger -> debugger.onEvent( new Debugger.InsertIntoMemoEvent(Debugger.Location.BEGIN, newValue)));
```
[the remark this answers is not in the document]
hatyo: Thanks, this line is already refactored.

PR #3186, opened 2025-02-21
Add protection against test running into an infinite loop

ohadzeliger:
Trying to improve the test experience, abort tests in case they are running into infinite loops when processing continuations.
This assumes that the test data will not produce more than a maximum number of rows (currently set to 100), and aborts the continuation loop after that number.
Note that the non-forced-continuation loop does not need that protection since it will require a "result" line in the yaml to trigger a continuation , and those are bound by a finite number.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
@@ -53,6 +53,7 @@ import static com.apple.foundationdb.relational.yamltests.command.QueryCommand.r
 public class QueryExecutor {
     private static final Logger logger = LogManager.getLogger(QueryExecutor.class);
     private static final int FORCED_MAX_ROWS = 1; // The maxRows number to use when we are forcing it on the test
+    private static final int MAX_CONTINUATIONS_ALLOWED = 100;
 
     @Nonnull
     private final String query;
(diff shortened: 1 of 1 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -258,6 +260,10 @@ private Object executeStatementWithForcedContinuations(final @Nonnull Statement
                         Assertions.assertFalse(hasNext);
                     }
                 }
+                count = count + 1; // PMD failure for ++
```
ScottDugas, line 263:
```suggestion
                count += 1;
```

ScottDugas (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ohadzeliger: Yeah. Wouldn't it be nice if PMD allowed for `if (++count < MAX)`?

ohadzeliger: Done.

ScottDugas (approved):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -258,6 +260,10 @@ private Object executeStatementWithForcedContinuations(final @Nonnull Statement
                         Assertions.assertFalse(hasNext);
                     }
                 }
+                count += 1; // PMD failure for ++
+                if (count > MAX_CONTINUATIONS_ALLOWED) {
```
alecgrieser, line 264: This seems fine for now. In theory, we could try and assert on things like whether we've advanced the continuation or something more obscure, but we probably don't want to have too many tests that exceed 100 continuations anyway

alecgrieser (approved):

PR #3187, opened 2025-02-21
Make prepared.yamsql run against 4.0.559.6 as the earliest

ohadzeliger:
Prepared.yamsql should be able to run with continuations against the `4.0.559.6`

### DIFF
--- a/yaml-tests/src/test/resources/prepared.yamsql
+++ b/yaml-tests/src/test/resources/prepared.yamsql
@@ -19,7 +19,7 @@
 
 ---
 options:
-    supported_version: 4.1.6.0
+    supported_version: 4.0.559.6
 ---
 schema_template:
     create table ta(a bigint, b double, c boolean, d integer, e integer array, f string, primary key(a))


yaml-tests/src/test/resources/prepared.yamsql
```diff
@@ -19,7 +19,7 @@
 
 ---
 options:
-    supported_version: 4.1.6.0
+    supported_version: 4.0.559.6
```
ScottDugas, line 22: I feel like there should be a comment here, because it doesn't work on anything between `4.0.559.6` and `4.1.6.0`.
Maybe something like:
```suggestion
    # This does not work against any version between 4.0.559.6 and 4.1.6.0 because server support was backported
    # to a patch release. We would like to surface that in the mixed mode test results rather than hide it.
    supported_version: 4.0.559.6 # this does not actually work 
```

ScottDugas (changes requested):

yaml-tests/src/test/resources/prepared.yamsql
ohadzeliger: Done.

PR #3189, opened 2025-02-21
Add version information to yaml test exceptions

ScottDugas:
This does the following:
1. It introduces a `YamlConnection` interface that wraps the underlying connection.
2. This makes it easier to have the connection expose what versions it is testing against

The issue we found is that when running in mixed-mode it runs the same query multiple times in parallel (as intended, I just didn't realize). This makes it incredibly hard to tell if a failing test is failing because it fails when going from the old version to the new version, or vice-versa.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
@@ -30,6 +30,7 @@ import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Assertions;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.net.URI;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -173,16 +174,19 @@ public class MultiServerConnectionFactory implements YamlConnectionFactory {
             return getCurrentConnection(true, "prepareStatement").prepareStatement(sql);
         }
 
+        @Nullable
         @Override
         public MetricCollector getMetricCollector() {
             throw new UnsupportedOperationException("MultiServer does not support getting the metric collector");
         }
 
+        @Nullable
         @Override
         public EmbeddedRelationalConnection tryGetEmbedded() {
             return null;
         }
 
+        @Nonnull
         @Override
         public List<String> getVersions() {
             return this.versions;
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
@@ -20,21 +20,18 @@
 
 package com.apple.foundationdb.relational.yamltests;
 
-import com.apple.foundationdb.relational.api.Options;
-import com.apple.foundationdb.relational.api.RelationalConnection;
 import com.apple.foundationdb.relational.api.RelationalPreparedStatement;
 import com.apple.foundationdb.relational.api.RelationalStatement;
+import com.apple.foundationdb.relational.api.metrics.MetricCollector;
+import com.apple.foundationdb.relational.recordlayer.EmbeddedRelationalConnection;
+import com.apple.foundationdb.relational.util.Assert;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Assertions;
 
 import javax.annotation.Nonnull;
 import java.net.URI;
-import java.sql.Array;
-import java.sql.Connection;
-import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
-import java.sql.Struct;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -70,11 +67,6 @@ public class MultiServerConnectionFactory implements YamlConnectionFactory {
     @Nonnull
     private final AtomicInteger currentConnectionSelector;
 
-    public MultiServerConnectionFactory(@Nonnull final YamlConnectionFactory defaultFactory,
-                                        @Nonnull final List<YamlConnectionFactory> alternateFactories) {
-        this(ConnectionSelectionPolicy.DEFAULT, 0, defaultFactory, alternateFactories);
-    }
-
     public MultiServerConnectionFactory(@Nonnull final ConnectionSelectionPolicy connectionSelectionPolicy,
                                         final int initialConnection,
                                         @Nonnull final YamlConnectionFactory defaultFactory,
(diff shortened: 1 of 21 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
```diff
@@ -182,103 +174,35 @@ public RelationalPreparedStatement prepareStatement(String sql) throws SQLExcept
...
+        public EmbeddedRelationalConnection tryGetEmbedded() {
+            return null;
         }
 
+        @Nonnull
         @Override
-        public void rollback() throws SQLException {
-            throw new UnsupportedOperationException("rollback is not supported in YAML tests");
+        public List<String> getVersions() {
+            return this.versions;
```
ohadzeliger, line 192:
```suggestion
return getCurrentConnection(false, "getVersion").getVersion();
```

If done this way, maybe we can turn `getVersions` to getVersion and no longer need the list of versions in the multi-server connection?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlConnection.java
```diff
@@ -0,0 +1,95 @@
...
+     * Try to get the underlying embedded relational connection in support of a few specific setup methods.
+     * @return the underlying embedded connection, or {@code null} if one is not (easily) available.
+     */
+    @Nullable
+    EmbeddedRelationalConnection tryGetEmbedded();
+
+    /**
+     * Return the ordered list of versions that this will test against.
+     * <p>
+     *     This does not reset as {@link #createStatement}/{@link #prepareStatement}, etc. are called.
```
ohadzeliger, line 89: What do you mean by "reset" here?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -130,28 +129,28 @@ private QueryCommand(int lineNumber, @Nonnull QueryInterpreter interpreter, @Non
                 "SkipConfig should not have gotten into QueryCommand " + lineNumber);
     }
 
-    public void execute(@Nonnull final RelationalConnection connection, boolean checkCache, @Nonnull QueryExecutor executor) {
+    public void execute(@Nonnull final YamlConnection connection, boolean checkCache, @Nonnull QueryExecutor executor) {
         try {
-            if (!(connection instanceof EmbeddedRelationalConnection) && checkCache) {
+            if (!connection.supportsMetricCollector() && checkCache) {
```
ohadzeliger, line 134: Why `metricCollector` here?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -273,10 +272,10 @@ private static Object executeStatement(@Nonnull Statement s, @Nullable String q)
         return execResult ? s.getResultSet() : s.getUpdateCount();
     }
 
-    private void setParametersInPreparedStatement(@Nonnull RelationalPreparedStatement s, @Nonnull Connection connection) throws SQLException {
+    private void setParametersInPreparedStatement(@Nonnull RelationalPreparedStatement statement) throws SQLException {
         int counter = 1;
         for (var parameter : Objects.requireNonNull(parameters)) {
-            s.setObject(counter++, parameter.getSqlObject(connection));
+            statement.setObject(counter++, parameter.getSqlObject(statement.getConnection()));
```
ohadzeliger, line 278: While this "locks" the connection to the statement, I wonder whether that's actually required by the standard, and whether we should support `createStruct` and `createArray` that span multiple connections.

ohadzeliger (changes requested):

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlConnection.java
ScottDugas: I think I meant:
```suggestion
     *     This does not change as {@link #createStatement}/{@link #prepareStatement}, etc. are called.
```

But given the open discussion (https://github.com/FoundationDB/fdb-record-layer/pull/3189/files#r1966310538) I'm not going to change this right now.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
ScottDugas: Because access to the `MetricCollector` is what we need in order to check the cache.
I originally had `supportsCheckCache`, but that felt bat from the vantage point of the `YamlConnection`, so I changed it to `supportsMetricCollector`, but that could probably use a comment.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ScottDugas: I think, based on my reading of the spec, it is allowed, or at least should be. But, you should also be able to take a `Struct` or `Array` from any other database, and a whole bunch of other things.
I also feel like you're doing something weird if you are using two different connections here.
So, I feel like keeping the `YamlConnection` interface simpler at this point is worth more than automatically testing that an array can be passed between connections, and that testing that would be better done in a focused suite of tests that don't hide any of the complexity of JDBC behind a yaml file.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MultiServerConnectionFactory.java
alecgrieser: I think the most important thing for logging purposes is that the user is able to reconstruct what happened from the log they give. Getting a list of versions is probably not that helpful without context like where the query started, what version it failed against, etc. 

For my change, I ended up merging your branch in, and then I needed to add an additional method, `initialVersion`, so that it was clear from a connection what version it would start on. That ended up being the clearest way to write assertions on the version, I think. Technically, you could almost rely on the fact that the connection shouldn't have been used yet and so getting the current connection (without advancing) is enough to return the right value...but it seemed a bit shakier and also less clearly associated with the goals it was trying to accomplish than adding a separate method to say where the query started.

ohadzeliger: Sounds like two things are required to make this useful: The initial version (so that assertions can be made for Alec's change) and also the actual history of what transpired during the life of the connection (for reproduction purposes, per Scott's comment). It does look like they are both quite different and hard to combine into one method.

ScottDugas: The list as it is should provide the `initialVersion` (i.e. `getVersions().get(0)` should always work).

It is potentially less clear what happened. I think it should be clear, because I think you'll have the specific result that failed, so you should know whether it was the first or second continuation.

ScottDugas: That last comment was written without seeing @ohadzeliger's two comments that came before it.

ScottDugas: I think, based on my testing, the current list _works_ although it could probably be better, by having multiple different methods to support past vs future versions.
Perhaps worth not trying to iterate on the behavior here until we have more experience working with this in actual investigations.

ohadzeliger: Sure.

ohadzeliger (approved):

PR #3198, opened 2025-02-25
Add support for UUID as primitive type in SQL

g31pranjal:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -174,6 +176,14 @@ private static Object unwrapPrimitive(@Nonnull Type type, @Nullable Object field
             return returnList;
         } else if (type.getTypeCode() == Type.TypeCode.VERSION) {
             return FDBRecordVersion.fromBytes(((ByteString)fieldValue).toByteArray(), false);
+        } else if (type.isUuid()) {
+            if (fieldValue instanceof UUID) {
```
normen662, line 180: This cannot happen I think. Can you remove this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
```diff
@@ -174,6 +176,14 @@ private static Object unwrapPrimitive(@Nonnull Type type, @Nullable Object field
...
+                return fieldValue;
+            }
+            Verify.verify(fieldValue instanceof DynamicMessage);
+            final var message = (DynamicMessage) fieldValue;
+            return new UUID((Long) message.getField(message.getDescriptorForType().findFieldByName("most_significant_bits")),
```
normen662, line 185: Can we cache this? The lookup is a binary search iirc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
```diff
@@ -360,7 +360,9 @@ public static <M extends Message> Object transformMessage(@Nonnull final FDBReco
             } else {
                 if (currentMessage != null) {
                     final var currentFieldType = Verify.verifyNotNull(currentRecordType.getField(messageFieldDescriptor.getIndex())).getFieldType();
-                    final var fieldResult = NullableArrayTypeUtils.unwrapIfArray(getFieldOnMessage(currentMessage, messageFieldDescriptor), currentFieldType);
+                    var fieldResult = messageFieldDescriptor.isRepeated() || !messageFieldDescriptor.getType().equals(Descriptors.FieldDescriptor.Type.MESSAGE) ?
```
normen662, line 363: Can you add a comment? What is this doing?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
```diff
@@ -87,7 +88,8 @@ enum PhysicalOperator {
         NULL_TO_RECORD(Type.TypeCode.NULL, Type.TypeCode.RECORD, (descriptor, in) -> null),
         NONE_TO_ARRAY(Type.TypeCode.NONE, Type.TypeCode.ARRAY, (descriptor, in) -> in),
         NULL_TO_ENUM(Type.TypeCode.NULL, Type.TypeCode.ENUM, (descriptor, in) -> null),
-        STRING_TO_ENUM(Type.TypeCode.STRING, Type.TypeCode.ENUM, ((descriptor, in) -> stringToEnumValue((Descriptors.EnumDescriptor)descriptor, (String)in)));
+        STRING_TO_ENUM(Type.TypeCode.STRING, Type.TypeCode.ENUM, ((descriptor, in) -> stringToEnumValue((Descriptors.EnumDescriptor)descriptor, (String)in))),
+        STRING_TO_UUID(Type.TypeCode.STRING, Type.TypeCode.UUID, ((descriptor, in) -> UUID.fromString((String) in)));
```
normen662, line 92: Don't we want this to bomb for malformed `UUID`s in a way that it throws a proper SQL error message rather that an internal Java one?

fdb-record-layer-core/src/test/proto/test_records_2.proto
```diff
@@ -33,6 +34,11 @@ message MyLongRecord {
   optional bytes bytes_value = 2;
 }
 
+message MyUuidRecord {
```
normen662, line 37: 1. Can you make a test based on the primary key being a UUID?
2. Can you create indexes using an UUID?

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -131,7 +131,7 @@ columnType
     : primitiveType | customType=uid;
 
 primitiveType
-    : BOOLEAN | INTEGER | BIGINT | FLOAT | DOUBLE | STRING | BYTES;
+    : BOOLEAN | INTEGER | BIGINT | FLOAT | DOUBLE | STRING | BYTES | UUID;
```
normen662, line 134: So if we have a syntax construct to generate a UUID, why do we need `STRING` -> `UUID` promotion?

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -1211,7 +1211,8 @@ functionNameBase
     | TIMESTAMPADD | TIMESTAMPDIFF | TIME_FORMAT | TIME_TO_SEC
     | TOUCHES | TO_BASE64 | TO_DAYS | TO_SECONDS | UCASE
     | UNCOMPRESS | UNCOMPRESSED_LENGTH | UNHEX | UNIX_TIMESTAMP
-    | UPDATEXML | UPPER | UUID | UUID_SHORT
+    | UPDATEXML | UPPER
+//    | UUID | UUID_SHORT
```
normen662, line 1215:
```suggestion
```

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/SqlTypeSupport.java
```diff
@@ -98,6 +98,7 @@ public static int recordTypeToSqlType(Type.TypeCode typeCode) {
             case VERSION:
                 return Types.BINARY;
             case ENUM:
+            case UUID:
                 //TODO(bfines) should be string?
```
normen662, line 102: Remove or refine pls!

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/MessageHelpers.java
g31pranjal: With the addition of UUID, now there are certain messages that are represented differently in the `runtime`, essentially all the ones defines in the 'tuple_fields.proto'. For those cases, we do not want to get the `runtime` value but the actual message. 

Hence, if the field is a message, we ask for the raw message, else, we get the field value.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/PromoteValue.java
g31pranjal: Good Idea! Added a new Semantic error code for that and replaced occurrences of `UUID.fromString` with `stringToUuidValue` that throws proper exception when unable to parse the input string

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/SqlTypeNamesSupport.java
```diff
@@ -119,6 +120,8 @@ public static int getSqlTypeCodeFromObject(Object obj) {
             return Types.ARRAY;
         } else if (obj instanceof Struct) {
             return Types.STRUCT;
+        } else if (obj instanceof UUID) {
+            return Types.OTHER;
```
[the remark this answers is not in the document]
g31pranjal: It's a bit surprising that there is not. I guess then it defaults to OTHER, unless we extend this to have our own types. 

Currently, OTHER can either be an ENUM or a UUID, and we leave it to the customer to infer that based on the value returned. Agree, its bit not clean.

fdb-relational-core/src/main/antlr/RelationalParser.g4
g31pranjal: done!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
g31pranjal: Make sense!

fdb-record-layer-core/src/test/proto/test_records_2.proto
g31pranjal: Added a couple of tests in record-layer-core. There are also end-to-end tests in yaml framework which explicitly tests UUID as a primary key and in Indexes.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/SqlTypeSupport.java
g31pranjal: done

fdb-relational-core/src/main/antlr/RelationalParser.g4
g31pranjal: If I understand your concern correctly, we do not currently support have support for construct to generate UUID. hence, for simple (not prepared) statements, the constant value supplied for UUID is always a `String` that is promoted to `UUID` downstream. 

Even with the support, postgres allows String arguments anywhere UUID is expected.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
g31pranjal: The UUID seems to have been "unwrapped" before itself. On the contrary, when we reach this point Type UUID always has a value of Java UUID. 

This is a bit complex in the way how unwrapping is done. We have Nullalbe arrays and Versions that unwrap here, whereas TupleFields (UUID, Nullable primitives) are already unwrapped downstream.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -169,6 +178,15 @@ default boolean isEnum() {
         return getTypeCode().equals(TypeCode.ENUM);
     }
 
+    /**
+     * Checks whether a {@link Type} is {@link Enum}.
```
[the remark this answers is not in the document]
normen662: This still says enum.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
normen662: Caching aside, but should the descriptor not be one of our predefined external dependencies when the dynamic types are defined (as per this PR)? If that is the case the field is also always the same and can be stashed somewhere. If none of that is true (I would like to understand where the reasoning is wrong), I still do not like the literal use of `most_significant_bits`. We normally treat protobuf in a way that we can rename things without any problem. This line of code make that assumption break (even using an index is better here). You can also use the statically defined proto-class' field name, I suppose.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FieldValue.java
normen662: Ah crap, this is considered a tuple field.

normen662 (changes requested):
I resolved all comments except one or two. It's possible that those can be resolved as well without further code change (it's more that they are questions).

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -381,23 +408,23 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
...
+            return Enum.fromProtoValues(isNullable, enumDescriptor.getValues());
+        } else if (typeCode == TypeCode.RECORD) {
+            Objects.requireNonNull(descriptor);
+            final var messageDescriptor = (Descriptors.Descriptor)descriptor;
+            if (NullableArrayTypeUtils.describesWrappedArray(messageDescriptor)) {
+                // find TypeCode of array elements
+                final var elementField = messageDescriptor.findFieldByName(NullableArrayTypeUtils.getRepeatedFieldName());
+                final var elementTypeCode = TypeCode.fromProtobufType(elementField.getType());
+                return fromProtoTypeToArray(descriptor, protoType, elementTypeCode, true);
+            } else if (TupleFieldsProto.UUID.getDescriptor().equals(messageDescriptor)) {
```
hatyo, line 424: It is a bit complex to understand why the handling of the `UUID` is done here and not in an outer `if`. But after some discussion I understand it better now. Thanks @g31pranjal for the explanation.

hatyo (approved):
LGTM, great work!

PR #3205, opened 2025-02-27
Introduce a new `Option` to set `asyncToSync` timeout.

### LINKED ISSUE #3213: Set timeout limit for async calls
If FDB is unreachable, it is possible for some calls that attempt to interact with it (such as get read version) to block forever.

hatyo:
This introduces a new `Option` for setting `asyncToSyncTimeout` to avoiding blocking forever on futures that may never return. An example of such future is waiting for `getReadVersion` JNI call to comeback against an unreachable FDB instance.

This fixes #3213.

### DIFF
--- a/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
+++ b/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
@@ -192,6 +192,7 @@ public final class Options {
         /**
          * Timeout for asynchronous operations in milliseconds, this is usually used to set an upperbound time limit for
          * operations interacting with FDB.
+         * Scope: Engine
          */
         ASYNC_OPERATIONS_TIMEOUT_MILLIS
     }
--- a/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
+++ b/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
@@ -188,6 +188,12 @@ public final class Options {
          * that can be used in EXECUTE CONTINUATION statements.
          */
         CONTINUATIONS_CONTAIN_COMPILED_STATEMENTS,
+
+        /**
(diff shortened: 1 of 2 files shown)


fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -188,6 +188,12 @@ public enum Name {
          * that can be used in EXECUTE CONTINUATION statements.
          */
         CONTINUATIONS_CONTAIN_COMPILED_STATEMENTS,
+
+        /**
```
arnaud-lacurie, line 192: Could you document the scope of that option? 
Basically indicate if it is scoped to the query, the connection, the engine, etc. 
(See other options above)

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
```diff
@@ -88,6 +89,10 @@ public FRL() throws RelationalException {
     @SpotBugsSuppressWarnings(value = "CT_CONSTRUCTOR_THROW", justification = "Should consider refactoring but throwing exceptions for now")
     public FRL(@Nonnull Options options) throws RelationalException {
         final FDBDatabase fdbDb = FDBDatabaseFactory.instance().getDatabase();
+        final Long asyncToSyncTimeout = options.getOption(Options.Name.ASYNC_OPERATIONS_TIMEOUT_MILLIS);
```
arnaud-lacurie, line 92: I suppose that given the fact that this is set at the DB level, we wouldn't be able to have that per query / per connection, is that correct?

arnaud-lacurie (changes requested):
Could you adjust the PR description? It mentions this as a "draft", which I suppose was the case when the PR was introduced.

arnaud-lacurie (approved):

PR #3215, opened 2025-02-28
Fix continuation deserialization errors by deserializing streaming aggregate plans in the correct order

### LINKED ISSUE #3214: Continuation deserialization error encountered by `StreamingAggregatePlan` during FORCE_CONTINUATIONS tests
We're seeing `FORCE_CONTINUATIONS` tests fail with the `StreamingAggregatePlan` in some circumstances. The precise error is a `VerifyError` that is ultimately caused by the deserializer seeing a reference before it sees the underlying type referee.

Doing some digging at the example, I see that the referee actually is in there, but it looks like the way we deserialize things, if the serialization happens in the wrong order, we may come across the reference before the definition.

Example stack trace:

```
Caused by: com.google.common.base.VerifyException
	at com.google.common.base.Verify.verify(Verify.java:102)
	at com.apple.foundationdb.record.query.plan.cascades.typing.Type$Record.fromProto(Type.java:2099)
	at com.apple.foundationdb.record.query.plan.cascades.typing.Type$Record$Deserializer.fromProto(Type.java:2473)
	at com.apple.foundationdb.record.query.plan.cascades.typing.Type$Record$Deserializer.fromProto(Type.java:2461)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.invokeDeserializer(PlanSerialization.java:170)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.dispatchFromProto(PlanSerialization.java:162)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.dispatchFromProtoContainer(PlanSerialization.java:146)
	at com.apple.foundationdb.record.query.plan.cascades.typing.Type.fromTypeProto(Type.java:646)
	at com.apple.foundationdb.record.query.plan.cascades.values.RecordConstructorValue.fromProto(RecordConstructorValue.java:408)
	at com.apple.foundationdb.record.query.plan.cascades.values.RecordConstructorValue$Deserializer.fromProto(RecordConstructorValue.java:504)
	at com.apple.foundationdb.record.query.plan.cascades.values.RecordConstructorValue$Deserializer.fromProto(RecordConstructorValue.java:492)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.invokeDeserializer(PlanSerialization.java:170)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.dispatchFromProto(PlanSerialization.java:162)
	at com.apple.foundationdb.record.query.plan.serialization.PlanSerialization.dispatchFromProtoContainer(PlanSerialization.java:146)
	at com.apple.foundationdb.record.query.plan.cascades.values.Value.fromValueProto(Value.java:704)
	at com.apple.foundationdb.record.query.plan.plans.RecordQueryStreamingAggregationPlan.lambda$fromProto$0(RecordQueryStreamingAggregationPlan.java:391)
```

The exact error isn't that useful, though. The test case in `aggregate-index-tests-count` is more useful. This test:

```
    -
      - query: select count(*) from t2 group by col2
      - explain: "ISCAN(MV5 <,>) | MAP (_ AS _0) | AGG (count_star(*) AS _0) GROUP BY (_._0.COL2 AS _0) | MAP (_._1._0 AS _0)"
      - result: [{1}, {3}]
```

Fails when run in `FORCE_CONTINUATIONS` mode.

alecgrieser:
The error we were seeing was that for certain plans, we were getting errors trying to deserialize the elements of a `FDBStreamingAggregatePlan` continuation. In particular, we were seeing the type's reference before the type's definition. This was because the serialization and deserialization of the plan happened in different orders, which meant that at deserialization time, we were no longer guaranteed to come across the definition first. I've kept the order of serialization the same, and then updated the deserialization code path to align it with existing continuations that are out in the wild.

This fixes #3214. I've enabled FORCE_CONTINUATIONS mode on that test, though it mostly just works on the embedded vs. current configuration (which is enough to surface the bug). More test coverage is added by having the `StreamingAggregate` plans in one of the Record Layer tests all go through `verifySerialization`, which previously wasn't being done on those plans.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
@@ -334,7 +334,8 @@ class FDBStreamAggregationTest extends FDBRecordStoreQueryTestBase {
     }
 
     @Nonnull
-    private RecordCursor<QueryResult> executePlan(final RecordQueryPlan plan, final int rowLimit, final byte[] continuation) {
+    private RecordCursor<QueryResult> executePlan(final RecordQueryPlan originalPlan, final int rowLimit, final byte[] continuation) {
+        final RecordQueryPlan plan = verifySerialization(originalPlan);
         final var types = plan.getDynamicTypes();
         final var typeRepository = TypeRepository.newBuilder().addAllTypes(types).build();
         ExecuteProperties executeProperties = ExecuteProperties.SERIAL_EXECUTE;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
@@ -35,7 +35,6 @@ import com.apple.foundationdb.record.planprotos.PRecordQueryStreamingAggregation
 import com.apple.foundationdb.record.provider.common.StoreTimer;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
 import com.apple.foundationdb.record.provider.foundationdb.FDBStoreTimer;
-import com.apple.foundationdb.record.query.plan.explain.ExplainPlanVisitor;
 import com.apple.foundationdb.record.query.plan.cascades.AliasMap;
 import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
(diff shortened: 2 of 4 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -334,7 +334,8 @@ private void populateDB(final int numRecords) throws Exception {
     }
 
     @Nonnull
-    private RecordCursor<QueryResult> executePlan(final RecordQueryPlan plan, final int rowLimit, final byte[] continuation) {
+    private RecordCursor<QueryResult> executePlan(final RecordQueryPlan originalPlan, final int rowLimit, final byte[] continuation) {
+        final RecordQueryPlan plan = verifySerialization(originalPlan);
```
alecgrieser, line 338: This change alone is enough to surface the errors we were seeing in tests, but the other tests are nice to have and we want anyway (at least eventually).

normen662 (approved):

Issue #3220, opened 2025-03-03
RecordQueryFirstOrDefaultPlan does not handle out-of-band stopping reasons from inner correctly


alecgrieser:
Issue #3219 points out that we were incorrectly feeding back the continuation in `RecordQueryFirstOrDefaultPlan`s to the inner continuation when resuming it. That meant that we were not resuming the continuation from the right place. However, in reality, there are three states that the cursor needs to consider:

1. The inner cursor returned a result. In this case, the first value is returned
1. The inner cursor was empty. In this case, the default value is returned
1. The inner cursor hit an out of band limit. In this case, we are unable to determine if the underlying cursor is empty or not

In both of the first two cases, because we've returned a result, we no longer need to record any state about the inner cursor, and we should just return a continuation that amounts to "we are done after this". In the third case, we should provide the inner cursor its previous continuation and continue on. That implies a continuation like:

```
message ForOrDefaultContinuation {
   optional bytes innerContinuation = 1;
   optional bool returnedValue = 2;
}
```

Technically, we could get by with something like just the inner continuation, and if the field is empty, then that means that we have already exhausted the continuation, but one would have to be careful around not returning an empty array.

Note: this was originally the scope of #3219. However, that issue has been repurposed to fixing the fact that the continuation wasn't being properly handled _at all_ by the plan, which meant it was always resuming from the wrong place. This is to fix the other issue identified in a follow up.

PR #3221, opened 2025-03-03
`EXISTS` queries now honor continuation from previous runs to avoid infinite loops

### LINKED ISSUE #3219: RecordQueryFirstOrDefaultPlan mishandles continuation when resuming cursor
The `RecordQueryFirstOrDefaultPlan` currently does not resume from a continuation correctly. In particular, it creates a `FutureCursor` over an inner, and that future cursor will return 1 value--either the first value from the underlying cursor or the "default" value. But the future cursor's result always has constructs a single continuation value (the byte string `\x00`) which is supposed to indicate "the element has been completed". However, we are currently threading that value through into the child plan, which generally will result in resuming the underlying scan from the beginning.

This is causing problems when trying to run some `EXISTS` queries in "force continuations' mode. Those queries are looping forever, because the cursor is not being advanced by the continuation during each run (because it is re-running underlying scans).

There are also problems with out-of-band errors not being handled correctly; see: #3220. However, that is a slightly different problem, and we should be able to fix this (which enables turning on force_continuations tests) and then update the continuation format in a future version.

alecgrieser:
The `RecordQueryFirstOrDefaultPlan` constructs a special cursor which has cardinality at most one. The continuation it returns back to the user is not based on the inner cursor but only on whether the cursor has returned any data. Previously, we were taking the `FutureCursor`'s continuation (which is always `\x00`) and giving it back to the `inner` plan, which would generally result in the plan resuming from the beginning. That is not the appropriate way to interpret the continuation, so this updates the logic to instead skip creating the inner cursor if we get a non-beginning continuation, which is in line with what the `FutureCursor`'s continuation means.

This fixes #3219.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
@@ -59,6 +59,7 @@ import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -884,24 +885,57 @@ public interface RecordCursor<T> extends AutoCloseable {
 
     @Nonnull
     static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull CompletableFuture<T> future) {
-        return new FutureCursor<>(executor, future);
+        return fromFuture(executor, future, null);
     }
 
     /**
      * Get a new cursor that has the contents of the given future as its only record.
      * If the continuation is nonnull, return an empty cursor since a {@link FutureCursor} returns only a single record.
+     * Note that in general, this can lead to futures being started that are not properly waited on,
+     * so it is generally advised that the user use {@link #fromFuture(Executor, Supplier, byte[])} instead
+     * of this one. See also the warning on {@link #fromFuture(Executor, Supplier, byte[])} for a word
+     * about if the work powering the underlying future has more complicated state that should be
+     * included in the cursor's continuation.
+     *
      * @param executor an executor for executing the future
      * @param future a future that completes to the only element of the cursor
      * @param continuation a continuation from a future cursor to determine whether the cursor has a single element or no elements
      * @param <T> the result type of the future
      * @return a new cursor producing the contents of {@code future} if the continuation is nonnull and an empty cursor otherwise
+     * @see #fromFuture(Executor, Supplier, byte[])
      */
     @Nonnull
     static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull CompletableFuture<T> future, @Nullable byte[] continuation) {
+        return fromFuture(executor, () -> future, continuation);
+    }
+
+    /**
+     * Get a new cursor that has the contents of the lazily-evaluated future as its only record.
+     * If the continuation is nonnull, return an empty cursor since a {@link FutureCursor} returns only a single record.
+     * In that case, the {@code futureSupplier} will not be evaluated to avoid queuing up asynchronous work
+     * that is not guaranteed to have anyone wait on. For that reason, this should generally be preferred
+     * to {@link #fromFuture(Executor, CompletableFuture, byte[])}.
+     *
+     * <p>
+     * <em>Warning:</em> Using a {@link FutureCursor} here may lead to incorrect results if the underlying
+     * future is backed by something that has some kind of state that should be included in the continuation.
+     * For example, if the future is backed by a cursor that may stop for an out-of-band {@link NoNextReason},
+     * it may be desirable to include the progress included in the underlying cursor in the final continuation.
+     * That is not done by this class, so something more bespoke may be required.
+     * </p>
+     *
+     * @param executor an executor for executing the future
+     * @param futureSupplier a supplier of a future that completes to the only element of the cursor
+     * @param continuation a continuation from a future cursor to determine whether the cursor has a single element or no elements
+     * @param <T> the result type of the future
+     * @return a new cursor producing the contents of {@code future} if the continuation is nonnull and an empty cursor otherwise
+     */
+    @Nonnull
+    static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull Supplier<CompletableFuture<T>> futureSupplier, @Nullable byte[] continuation) {
         if (continuation == null) {
-            return new FutureCursor<>(executor, future);
+            return new FutureCursor<>(executor, futureSupplier.get());
         } else {
-            return RecordCursor.empty();
+            return RecordCursor.empty(executor);
         }
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFirstOrDefaultPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFirstOrDefaultPlan.java
@@ -96,10 +96,19 @@ public class RecordQueryFirstOrDefaultPlan implements RecordQueryPlanWithChild, 
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return new FutureCursor<>(store.getExecutor(),
-                getChild().executePlan(store, context, continuation, executeProperties).first()
-                        .thenApply(resultOptional ->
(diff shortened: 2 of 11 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFirstOrDefaultPlan.java
```diff
@@ -96,10 +96,19 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
-        return new FutureCursor<>(store.getExecutor(),
-                getChild().executePlan(store, context, continuation, executeProperties).first()
-                        .thenApply(resultOptional ->
-                                resultOptional.orElseGet(() -> QueryResult.ofComputed(onEmptyResultValue.eval(store, context)))));
+        if (continuation == null) {
+            return new FutureCursor<>(store.getExecutor(),
```
hatyo, line 100: Isn't dangerous to return `FutureCursor` in the first place? Because it seems we might be losing the state of the inner cursor in some cases.

I looked into the `FutureCursor` code, and I do not think it is aimed to be composable with other cursors, and therefore, passing a future representing a state of a resumable computation does not work nicely with it. Am I readying this correctly, or am I missing something?

yaml-tests/src/test/resources/join-tests.yamsql
```diff
@@ -67,16 +67,45 @@ test_block:
...
+      # Can remove once we do not care about cross-version compatibility before !current_version
+      # See: https://github.com/FoundationDB/fdb-record-layer/issues/3219
+      - query: select fname, lname from
+                  (select fname, lname, dept_id from emp where exists (select * from project where emp_id = emp.id)) as sq,
+                  dept
+               where sq.dept_id = dept.id and dept.name = 'Sales';
+      - maxRows: 1
+      - initialVersionLessThan: !current_version
+      - result: [{"Daniel", "Miller"}]
+      - result: []
```
ScottDugas, line 107: Isn't this the same as the above? It returns 1 row with `{"Daniel", "Miller"}`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFirstOrDefaultPlan.java
alecgrieser: Yeah, that's right that `FutureCursor`s lose the state of the inner cursor, and isn't really "composable" with them in quite the same way as, say, the `MapCursor` is. The `FutureCursor` is really intended to be used for something like "go complete this asynchronous task and integrate the result into a cursor stream". Grabbing the first element from a cursor doesn't _really_ play well with this, because the underlying cursor itself can have state.

As I point out in #3220, for example, the inner cursor can end for an out-of-band reason, which will not be correctly handled by this cursor (before or after). The version in the PR is still an improvement, as now it at least interprets the continuation it gets in a way that is consistent with what the producer of the continuation wanted, but there's more work to be done. However, fixing that bug would require changing the continuation format, which is why I've deferred that here.

ScottDugas: Actually, I think the problem is in the usage. Other than changing how continuations work on a broader scale, I'm not sure how to make this better, other than adding identifiers to our cursor to make sure you don't pass it into the wrong object.

`FutureCursor` is doing the right thing by the contract for a cursor.
It returns two things:
1. the result of the cursor, with a continuation
2. an exhausted continuation.

Since out-of-band limits are ignored until a cursor has made progress, it is not really relevant to the out-of-band limit.

The problem in the usage is that you are passing a continuation from one cursor (the `FutureCursor`) into another cursor (`executePlan`). 

Perhaps the medium-term solution is that the `FutureCursor` constructor should take a continuation. This usage would be more clearly wrong if you tried to pass the `continuation` parameter into two different methods.

ScottDugas, line 100: Is now the right time to add some tests of `FutureCursor` directly? It doesn't look like it has any unit tests.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryFirstOrDefaultPlan.java
alecgrieser: Maybe... I'm not convinced that there's much we're going to get out of just those tests, given that the tricky part is more about how it's used. I was thinking if there should be a method like:

```java
static RecordCursor<T> futureCursor(Supplier<CompletableFuture<T>> supplier, @Nullable byte[] continuation, Executor executor) {
   return continuation == null ? new FutureCursor(executor, supplier.get()) : RecordCursor.empty(executor);
}
```

And then we could assert that _that_ did the right thing...ish.

alecgrieser: I added some basic unit tests to `RecordCursorTest`. We also already had a `fromFuture` method that seems like it was trying to do much of the same thing as this logic here, so I switched over to it. I added some verbiage in the javadoc to reflect things we discussed in this PR review

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
```diff
@@ -884,24 +885,57 @@ static <T> RecordCursor<T> fromFuture(@Nonnull CompletableFuture<T> future) {
...
+     *
+     * @param executor an executor for executing the future
+     * @param futureSupplier a supplier of a future that completes to the only element of the cursor
+     * @param continuation a continuation from a future cursor to determine whether the cursor has a single element or no elements
+     * @param <T> the result type of the future
+     * @return a new cursor producing the contents of {@code future} if the continuation is nonnull and an empty cursor otherwise
+     */
+    @Nonnull
+    static <T> RecordCursor<T> fromFuture(@Nonnull Executor executor, @Nonnull Supplier<CompletableFuture<T>> futureSupplier, @Nullable byte[] continuation) {
         if (continuation == null) {
```
ScottDugas, line 935: It's weird that this is handled here, and not in the `FutureCursor` constructor, and exposes a risk that a user would call the constructor directly.

yaml-tests/src/test/resources/join-tests.yamsql
ScottDugas: Oh, so the key is that if you had, e.g. `4.1.8.0 then !currentVersion` it would return something like:
```
- result: [{"Daniel", "Miller"}]
- result: [{"Daniel", "Miller"}]
- result: []
```

Perhaps a comment in the other branch then?

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordCursor.java
alecgrieser: Maybe. There are a couple of things going on here that make it hard to use the constructor, mainly that that class is trying to be "simpler" by only covering the case that the original future supplies a single value. I tinkered around with moving the future supplier (and the continuation) into the `FutureCursor`, and, like, it can be done, but it's not clear that this is more in line with the way other cursors work. You can't, for example, return a `RecordCursor.empty()` from the constructor directly.

Other cursors (like, say, the `FlatMapPipelinedCursor`) employ a strategy where they have an internal constructor and then a static initializer in `RecordCursor` that handles things like initializing their children. I've tried to make `FutureCursor` as analogous to those as possible. I've marked the current `FutureCursor` constructor as `INTERNAL` and removed calls to it other than this one. The javadoc for that now points to this method.

In theory, we could mark that constructor as `private` and replace it with a static initializer on the `FutureCursor` to avoid mis-use. But this sort of quickly devolves into "should we refactor all cursors to make handling continuations more obvious", and like, maybe, but I don't want to boil the ocean too much given that the proximate goal of this PR is to fix a bug that is evident in our existing tests.

yaml-tests/src/test/resources/join-tests.yamsql
alecgrieser: That's what would happen if the initial version were `!current_version` and the second version were 4.1.8.0. I've added some comments about that in those branches. This is, however, sort of a general pattern, which maybe will become clearer when we try to merge my `enforce-force-continuations` branch in, and I'm not sure what the right balance here is between documenting everything and making everything a little unwieldy given that the same sorts of changes are applied multiple places.

ScottDugas (approved):

PR #3222, opened 2025-03-03
Fix unstable hash for NULL literal

### LINKED ISSUE #3218: Plan hash instability of `NullValue`
The `NullValue` class currently has a JVM-dependent plan hash, which leads to `planHash` instability when trying to resume a plan that contains a `NULL` literal.

The plan hash code is here: https://github.com/FoundationDB/fdb-record-layer/blob/aa6176fed9bfb5f9a57d3ccf152bde91028cdba5/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NullValue.java#L104

It includes the `NullValue`'s "type" in the return value, but `Type`s aren't `PlanHashable`. The `NullValue`s I saw seem to have type `NULL`, which is a singleton with a default `hashCode` implementation, so we're getting a new value with each JVM.

This is causing us trouble with enabling force continuations tests for certain queries in the YAML framework. For example:

https://github.com/FoundationDB/fdb-record-layer/blob/aa6176fed9bfb5f9a57d3ccf152bde91028cdba5/yaml-tests/src/test/resources/boolean.yamsql#L98

ohadzeliger:
Resolves #3218

This fixes an issue where the continuation for NULL literals is failing to execute since the hash code for the NULL literal is unstable. The tests now pass (and so enabled) for configurations other than forced continuations against older versions.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
@@ -1110,6 +1110,16 @@ public interface Type extends Narrowable<Type>, PlanSerializable {
                 return Null.fromProto(serializationContext, nullTypeProto);
             }
         }
+
+        @Override
+        public int hashCode() {
+            return getTypeCode().name().hashCode();
+        }
+
+        @Override
+        public boolean equals(final Object other) {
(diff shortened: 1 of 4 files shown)


normen662 (changes requested):
Please take the `resultType` out of the plan hash in `NullValue` altogether.

normen662 (approved):

PR #3230, opened 2025-03-05
`RecordQueryExplodePlan` plan adheres to `skip` and `limit` requirements

### LINKED ISSUE #3229: `RecordQueryExplodePlan` does not honor row skip and limit requirements
this plan operator is, so far, only used internally to represent a transposition of constant arrays such as e.g. `in` values as a leg of an in-join plan, and as long as the upper operator honors the skip and limit requirements it was probably ok to have this operator ignore these parameters internally. In fact, it is somewhat common for the upper operator to remove these requirements before proceeding to execute the inner operator.

I came across this issue when implementing table-functions, one example of these functions is to directly select from `values` that translates to a projection from this operator at runtime. In other words, it suddenly became a top-level operator, and therefore, it must respect skip and limit to align with the execution behavior of other operators.

hatyo:
The `RecordQueryExplodePlan` plan is, so far, only used internally to represent a transposition of constant arrays such as e.g. `in` values as a leg of an in-join plan, and as long as the upper operator honors the skip and limit requirements it was probably ok to have this operator ignore these parameters internally. In fact, it is somewhat common for the upper operator to remove these requirements before proceeding to execute the inner operator.

I came across this issue when implementing table-functions, one example of these functions is to directly select from `values` that translates to a projection from this operator at runtime. In other words, it suddenly became a top-level operator, and therefore, it must respect skip and limit to align with the execution behavior of other operators.

This PR explicitly sets the `limit` and `skip` on the resulting `RecordCursor` of the `RecordQueryExplodePlan`, it also adds a number of unit tests.

It resolves #3229.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
@@ -88,7 +88,8 @@ public class RecordQueryExplodePlan implements RecordQueryPlanWithNoChildren {
                                                                      @Nonnull final ExecuteProperties executeProperties) {
         final var result = collectionValue.eval(store, context);
         return RecordCursor.fromList(result == null ? ImmutableList.of() : (List<?>)result, continuation)
-                .map(QueryResult::ofComputed);
+                .map(QueryResult::ofComputed)
+                .skipThenLimit(executeProperties.getSkip(), executeProperties.getReturnedRowLimit());
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/ExplodePlanTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/ExplodePlanTest.java
@@ -1,0 +1,113 @@
+/*
+ * ExplodePlanTest.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.plans;
+
+import com.apple.foundationdb.record.EvaluationContext;
+import com.apple.foundationdb.record.ExecuteProperties;
+import com.apple.foundationdb.record.RecordCursor;
+import com.apple.foundationdb.record.query.plan.cascades.values.LiteralValue;
+import com.apple.foundationdb.record.query.plan.cascades.values.Value;
+import com.google.common.collect.ImmutableList;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Optional;
+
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
```diff
@@ -88,7 +88,8 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final ExecuteProperties executeProperties) {
         final var result = collectionValue.eval(store, context);
         return RecordCursor.fromList(result == null ? ImmutableList.of() : (List<?>)result, continuation)
-                .map(QueryResult::ofComputed);
+                .map(QueryResult::ofComputed)
+                .skipThenLimit(executeProperties.getSkip(), executeProperties.getReturnedRowLimit());
```
alecgrieser, line 92: I was trying to reason about whether there is a query where the result set may be affected by this, as it would be nice to have a test somewhere that reflects the top-level behavior change. I think the problem is that to see this bug, we'd need to find a plan that had an explode plan without a cardinality-modifying plan on top of it. For example, this query:

```sql
SELECT T.id, x.b FROM T, (SELECT b FROM T.arrayField) x LIMIT 10
```

I think this ends up getting planned with an explode:

```
SCAN(<,>) | TFILTER T | FLATMAP q0 -> { EXPLODE(q0.arrayField) AS q1 RETURN (q1.b AS b) }
```

But the flat map plan ends up clearing the skip and limit when it hands it down to the explode here anyway.

Likewise, something like:

```sql
SELECT A.* FROM A WHERE A.x IN ?xList
```

May likewise be planned with an explode:

```
EXPLODE([$inList]) | INJOIN ISCAN(xIndex [EQUALS _])
```

But the `INJOIN` plan also clears the skip and limit before it creates explode child. (This particular one may also get planned with a `FLATMAP` on the outer, which is similar in the way it handles limits.)

I _think_ the only way we could avoid this is by having only cardinality preserving operations from an explode to the final result. Something like:

```sql
SELECT * FROM VALUES ?listParam LIMIT $limit
```

Which would have to be planned with just something like:

```
EXPLODE([$listParam])
```

Which is fine, other than the fact that we don't support `FROM VALUES` (for the moment).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/ExplodePlanTest.java
```diff
@@ -0,0 +1,115 @@
...
+        if (verifyLimitExceeded) {
+            final var result = actualCursor.getNext();
+            Assertions.assertFalse(result.hasNext());
+            Assertions.assertEquals(RecordCursor.NoNextReason.RETURN_LIMIT_REACHED, result.getNoNextReason());
+        }
+    }
+
+    @Test
+    void explodeWithSkipAndLimitWorks() {
+        verifyCursor(ExplodeCursorBuilder.instance().withLimit(1).build(), ImmutableList.of(1), true);
```
alecgrieser, line 109: Relatively minor: it would be nice if this were structured as a `@ParameterizedTest` with the different cases as arguments rather than one test with all of the test cases. That way, if one of them fails, we know which one(s). Otherwise, a failure in an earlier case can mask failures in other test cases.

alecgrieser (approved):
LGTM. I did leave one comment about the way a test is structured, and another comment which amounts to "is there a SQL query that we could assert about that would be affected by this"? I think the answer to the latter is "no", but I did want to be sure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
hatyo: That's exactly right, this operator is used under another operator that handles skip and limit in all[*] potential compositions of plan operators we can think ok. This is why I had to write the test like that. As you mentioned, we start seeing this issue when this operator becomes responsible for handling skip and limit when used in table-function.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/ExplodePlanTest.java
hatyo: Done, I also tried to make the parameterized test output look more readable:

image

alecgrieser (approved):

PR #3232, opened 2025-03-06
Set *.metrics.binpb to diff as binary

ScottDugas:
These files have a large amount of binary data and are not inteded to be diffed by a human. They also have a fair amount of ascii, so sometimes git (and particularly github) considers them text files and displays the contens, which is not helpful.


ScottDugas:
I tried, but could not get GitHub to reliably treat these as binary files.

PR #3233, opened 2025-03-07
Run Mixed Mode Tests during the release parallel to the other tests

ScottDugas:
This changes our release build to run the MixedModeTest in parallel with the other tests so that it can populate the release notes at that time. One key advantage of this is the documentation that sphinx publishes will have this information.

This also changes the release build so that it is usable for both patch and non-patch builds, and figures that out depending on the target branch.

Note: there are a lot of commits here, I don't think I recommend going through this change commit-by-commit, there was a lot of trial-and-error to get here.


.github/workflows/release.yml
```diff
@@ -3,14 +3,82 @@ name: Release
...
+                throw new Error("Target must be a patch branch or main, but was: " + context.ref);
+            }
+          result-encoding: string
+      - name: Print update type
+        shell: bash
+        run: echo "${{steps.update-type.outputs.result}}"
+
+  # 2. In parallel, we run:
+  #   a) all the tests
+  #   b) the mixed-mode tests
```
alecgrieser, line 41: Do we know what the timings are? Would it make sense to also split up the core-tests and the non-core (non-mixed-mode) tests like we do in prb?

.github/workflows/release.yml
```diff
@@ -3,14 +3,82 @@ name: Release
...
+  test:
+    needs: [get-update-type]
```
alecgrieser, line 43: Does this actually need `get-update-type`?

.github/workflows/release.yml
```diff
@@ -3,14 +3,82 @@ name: Release
...
+        uses: actions/checkout@v4.2.2
+      - name: Mixed Mode Test
+        uses: ./actions/mixed-mode-test
+        with:
+          update_type: ${{ needs.get-update-type.outputs.update-type }}
+      # I think this _needs_ to be done at this level, rather than in the action
+      # so that "mixed-mode-results" gets passed around correctly
+      - name: Upload mixed mode results
+        id: mixed_mode_results
+        uses: actions/upload-pages-artifact@v3
```
alecgrieser, line 69: Why is this `upload-pages-artifact`? Shouldn't it just be `upload-artifact`?

actions/release-test/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Run standard tests for release
```
alecgrieser, line 1: It sort of feels like we should in-line this, if it's only used by one caller and it's pretty small (and having it spelled out in the main action means the reader doesn't have to drill down into another file to see what's going on).

actions/release-publish/action.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and publish release
+name: Tag and publish a release
```
alecgrieser, line 1: I could see the argument that we should in-line this one into the main action as well if it's only called from one place. This is a pretty substantial number of steps, so I could buy an argument that it's better to encapsulate them, but I'll also say that because the GitHub workflows UI shows composite steps as a single entity, it can be difficult to determine what's going on for long workflows like this. That probably becomes less of a problem if this doesn't include the tests (as the output from that currently takes up so much space that it's hard to see what's going on after about the Lucene tests), so it may be fine to leave this as its own thing

.github/workflows/release.yml
```diff
@@ -3,14 +3,82 @@ name: Release
 on:
   workflow_dispatch:
 
+
 jobs:
-  gradle:
+  # TOODO how do I ensure that nothing is running on a different commit
```
alecgrieser, line 8: What do you mean by this TODO? Are you worried that someone else might be concurrently running the release?

actions/mixed-mode-test/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      id: get_new_version
+      shell: bash
+      run: |
+        echo "version=$(python build/versionutils.py gradle.properties)" >> "$GITHUB_OUTPUT"
+    - name: Create results directory
+      shell: bash
+      run: mkdir artifact
+    - name: Create markdown
+      shell: bash
+      run: python build/publish-mixed-mode-results.py ${{ steps.get_new_version.outputs.version }} --run-link ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }} > artifact/mixed-mode-results.md
```
alecgrieser, line 35: Looking over `publish-mixed-mode-results.py`, I think this is probably fine, as the only thing it should print out are the mixed mode markdown information. I could also see the argument, though, that we should give it an explicit output argument (other than the "release notes" one that's already there) to make it less likely that someone accidentally prints some debugging information and we unwittingly put it into the final artifact.

.github/workflows/release.yml
```diff
@@ -3,14 +3,82 @@ name: Release
...
+    steps:
+      - name: Calculate update type
+        uses: actions/github-script@v7
+        id: update-type
+        with:
+          script: |
+            if (context.ref == "refs/heads/main") {
+                return "BUILD";
+            } else if (context.ref.startsWith("refs/heads/")) {
+                return "PATCH";
```
alecgrieser, line 30: One consequence of this approach is that if someone runs a patch build, they use the patch process that is baked into their branch, rather than the one on main. In some ways, that's good, because it means that if there's a relatively stable release process in a past branch, then past builds will continue to use that. But it also means that we need to be a little mindful of keeping past patch branches up to date if there are any changes that need to be reflected everywhere. Overall, this is probably fine, and there was actually already some dependency between the different branches (as any of the composite actions would borrow the version from the branch), but I did want to raise this nonetheless

build/create_release_notes.py
```diff
@@ -254,13 +257,24 @@ def main(argv):
 
     old_version = args.old_version
     release_notes = []
+    if len(args.new_version) == 0:
+        print("At least one new version must be provided", file=sys.stderr)
+        exit(1)
+    elif len(args.new_version) > 1 and args.mixed_mode_results is not None:
+        print("--mixed-mode-result cannot be provided with more than one new_version", file=sys.stderr)
+        exit(1)
+    mixed_mode_results = "<!-- MIXED_MODE_RESULTS {new_version} PLACEHOLDER -->"
```
alecgrieser, line 266: The old placeholder string was actually a formatted string, so this should be something like:

```python
mixed_mode_results = f"<!-- MIXED_MODE_RESULTS {new_version} PLACEHOLDER -->"
```

Except that that doesn't work because we haven't figured out the new version (yet). Maybe what this needs to do is set the field to `None` here, then set it to the contents of the file, and then in the loop (or back in the `format_notes` function) set it to the string with the interpolated placeholder if it's still `None`. Or we give up on the placeholder all together.

alecgrieser (changes requested):

.github/workflows/release.yml
ScottDugas: No it does not, removed.

actions/release-publish/action.yml
ScottDugas: Yeah, I think your right, and it is cleaner to just pull the actions into the workflow.

I was hoping that it would be much cleaner, like if each job was just a reference to another action, that would be nice, but it didn't end up being that way. `workflows/release.yml` is a bit big, but not crazy.

actions/release-test/action.yml
ScottDugas: Yeah, even more so than `release-publish/action.yml`

build/create_release_notes.py
ScottDugas: Yeah, I decided to just give up on the placeholder all together.
I didn't remove the code that can use the placeholder, I'll leave that for another day.

actions/mixed-mode-test/action.yml
ScottDugas: Yeah, that's probably a good callout.
I added a `--output` parameter and removed the code that allows it to inject into release notes.

.github/workflows/release.yml
ScottDugas: It also kind of looks like my patch release may be messing with the workflows in general, which seems super weird. 
It, all of a sudden, has 5 runs, and all of them say it's invalid: https://github.com/FoundationDB/fdb-record-layer/actions/workflows/patch_release.yml
I fixed the issue (I think), but nothing happened after that.

.github/workflows/release.yml
```diff
@@ -58,18 +64,40 @@ jobs:
...
       # I think this _needs_ to be done at this level, rather than in the action
       # so that "mixed-mode-results" gets passed around correctly
       - name: Upload mixed mode results
         id: mixed_mode_results
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-artifact@v4.6.0
         with:
           name: mixed-mode-results
-          path: ./artifact
+          path: mixed-mode-results.md
```
alecgrieser, line 99: We should consider adding `mixed-mode-results.md` to the `.gitignore`, or placing it in a directory that is already excluded. The way the PR creating workflow works, if there are files that aren't in the `.gitignore` present, it will include them in the PR that it creates. Excluding the file would make sure that we don't accidentally do that, particularly in the other step when we download it.

alecgrieser (changes requested):
This mostly LGTM. There's one last thing regarding `mixed-mode-results.yml` to make sure we don't accidentally include it in the PR that we (optionally) generate.

Issue #3096, opened 2025-02-04
Infinite continuations: sum(*) returns alternating rows forever


ohadzeliger:
Adding maxRows:1 to the yaml test keeps returning two rows forever (the result row and an empty row), the continuation is never atEnd().

Steps To Reproduce:
1. For example, in the test aggregate-index-tests.yamsql, modify the test to be:
2.        - query: select max(col2) from T1 use index (mv8);
       - maxRows: 1
       - explain: "ISCAN(MV8 <,>) | MAP (_ AS _0) | AGG (max_l(_._0.COL2) AS _0) | ON EMPTY NULL | MAP (_._0._0 AS _0)"
       - result: [{!l 13}]
       - result: [{!null _}]
       - result: [{!l 13}]
       - result: [{!null _}]
       - result: [{!l 13}]
       - result: [{!null _}]
       - result: [{!l 13}]
       - result: [{!null _}]
3. And the test will always fail with “Query returned more results, but no more were expected after line …”

Note that this behaves differently when running in mixed mode, where the error is:

Caused by: com.google.common.base.VerifyException
	at com.google.common.base.Verify.verify(Verify.java:102)
	at com.apple.foundationdb.relational.recordlayer.query.QueryPlan$ContinuedPhysicalQueryPlan.validatePlanAgainstEnvironment(QueryPlan.java:520)
	at com.apple.foundationdb.relational.recordlayer.query.QueryPlan$PhysicalQueryPlan.executePhysicalPlan(QueryPlan.java:361)
	at com.apple.foundationdb.relational.recordlayer.query.QueryPlan$PhysicalQueryPlan.executeInternal(QueryPlan.java:252)
	at com.apple.foundationdb.relational.recordlayer.query.QueryPlan$PhysicalQueryPlan.executeInternal(QueryPlan.java:112)

alecgrieser:
We now know what the difference between mixed-mode and non-mixed results here is about (it's a testing artifact based on how the framework handles receiving `null`/starting continuations--it's also been addressed by #3186). The continuation problem has to do with how the aggregate cursor handles continuations.

Namely, this is a consequence of the fact that the `AggregateCursor` does not remember whether it has already returned a value, so when `maxRows` is 1, it effectively goes:

1. First iteration: start from the beginning, return value and continuation pointing to the last record in the scan
[...]

PR #3246, opened 2025-03-13
Streaming aggregate cursor now requires 4.1.9.0 or later for continuation deserialization

alecgrieser:
This cleans up the `StreamingAggregateCursor` and its helper classes to remove the `createDefaultIfEmpty` option. We'd been carrying that option around as we introduced that field in #3092 and wanted to preserve behavior when upgrading from older versions. The intention had been that all new plans would set the option starting in 4.1, but a bug (fixed in #3211; see #3096) means that we didn't actually enable it until 4.1.9.0. That means that after this change, we'd require 4.1.9.0 or newer in order to safely continue these queries. In theory, we could wait with this change, but 4.1.9.0 has enough fixes that I actually think our recommendation should be that anyone upgrading from 4.0 or below go straight to 4.1.9.0, and then they can proceed safely to a newer version with this change.

I was able to validate that this change is compatible with 4.1.9.0 via the cross-version tests run during PRB. When I ran the full mixed mode tests with the `aggregate-index-tests.yamsql`, I found that only 4.1.9.0 worked, which is expected.

This resolves #3092.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
@@ -27,6 +27,7 @@ import com.apple.foundationdb.record.ObjectPlanHash;
 import com.apple.foundationdb.record.PlanDeserializer;
 import com.apple.foundationdb.record.PlanHashable;
 import com.apple.foundationdb.record.PlanSerializationContext;
+import com.apple.foundationdb.record.RecordCoreArgumentException;
 import com.apple.foundationdb.record.RecordCursor;
 import com.apple.foundationdb.record.cursors.aggregate.AggregateCursor;
 import com.apple.foundationdb.record.cursors.aggregate.StreamGrouping;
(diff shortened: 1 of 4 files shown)


pengpeng-lu (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -371,7 +361,7 @@ public PRecordQueryStreamingAggregationPlan toProto(@Nonnull final PlanSerializa
         builder.setGroupingKeyAlias(groupingKeyAlias.getId())
                 .setAggregateAlias(aggregateAlias.getId())
                 .setCompleteResultValue(completeResultValue.toValueProto(serializationContext))
-                .setIsCreateDefaultOnEmpty(isCreateDefaultOnEmpty);
+                .setIsCreateDefaultOnEmpty(false);
```
normen662, line 364: Can we mark this as deprecated in the proto file?

normen662 (commented):
My comment about deprecating a field is more or less about procedure going forward, not necessarily about this particular instance.

PR #3252, opened 2025-03-17
Remove yaml-tests dependency on fdb-record-layer-core test code

ScottDugas:
This changes the main code for `yaml-tests` to not depend on the test configuration of `fdb-record-layer-core`, allowing it to be more easily used in other projects.

`yaml-tests` had the following test dependencies:
1. `TestHelpers.assertThrows` - Replaced with junit's `Assertions.assertThrows`. We should probably do this across the code base, but I did not as part of this PR.
2. `TestHelpers.DangerousRunnable` - this was fairly easy to remove as it was only used by `QueryCommand.runWithDebugger`. I _think_ this is unnecessary as we set the debugger globally.
3. `DebuggerWithSymbolTables` - I believe this is only used to debug the behavior of the planner, which doesn't seem like something people should be testing outside, so moving this to the test code seemed reasonable. In order to accomplish this, I added a new extension: `YamlTestWithDebuggingExtension` that extends `YamlTestExtension` but initializes the debugger in the `beforeAll`. In order to use this, I moved `@YamlTest` the test code. This seems like a good idea anyways, as that adds the `MixedMode` category. 

With the above changes I removed `YamlTest.Runner`, and changed `YamlRunner` to be an acceptable parameter to the tests.


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -179,13 +174,13 @@ private void executeInternal(@Nonnull final YamlConnection connection, boolean c
                 // can result in the explain plan being put into the plan cache, so running without the debugger
                 // can pollute cache and thus interfere with the explain's results
                 Integer finalMaxRows = maxRows;
-                runWithDebugger(() -> executor.execute(connection, null, queryConfig, checkCache, finalMaxRows));
+                executor.execute(connection, null, queryConfig, checkCache, finalMaxRows);
```
normen662, line 177: This disables all sanity checks within the planner. See `Debugger.isSane()`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -179,13 +174,13 @@ private void executeInternal(@Nonnull final YamlConnection connection, boolean c
...
             } else if (QueryConfig.QUERY_CONFIG_EXPLAIN.equals(queryConfig.getConfigName()) || QueryConfig.QUERY_CONFIG_EXPLAIN_CONTAINS.equals(queryConfig.getConfigName())) {
                 Assert.that(!queryIsRunning, "Explain test should not be intermingled with query result tests");
                 // ignore debugger configuration, always set the debugger for explain, so we can always get consistent
                 // results
                 Integer finalMaxRows1 = maxRows;
-                runWithDebugger(() -> executor.execute(connection, null, queryConfig, checkCache, finalMaxRows1));
+                executor.execute(connection, null, queryConfig, checkCache, finalMaxRows1);
```
normen662, line 183: same here.

normen662 (changes requested):

ScottDugas:
I think it would still be valuable at some point to do this, perhaps by creating a test-utils package, but this is no longer a priority right now, so I'm going to close this.

PR #3253, opened 2025-03-17
Allow having YamlTestExtension include method name in test names

ScottDugas:
When publishing to JUnit xml style files there is no hierarchy, so some tools will just show the class name and description, making it hard to tell what tests were run, or failed. If YamlTestExtension is used in an environment where that is the case this can be used to have the output include the method name. For example:
- showcasingTests(Embedded)
- showcasingTests(MultiServer (Embedded then !current_version)) ...

Instead of
- YamlIntegrationTests
  - showcasingTests(Runner)
    - Embedded
    - MultiServer (Embedded then !current_version)

I thought about making this configurable via system property, but I think having it configured by a constructor parameter aligns more with PRs #3252 and #3251

Note: This is a breaking change if someone is depending on `YamlTestExtension`, but only if they are depending on the recently merged #3251

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
@@ -66,13 +66,23 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
     private List<ExternalServer> servers;
     @Nullable
     private final String clusterFile;
+    private final boolean includeMethodInDescriptions;
 
     public YamlTestExtension() {
-        this.clusterFile = null; // it will get it from the environment
+        this(null, false);
     }
 
-    public YamlTestExtension(@Nullable final String clusterFile) {
+    /**
+     * Create a new extension with some configuration.
+     * @param clusterFile a custom cluster file to use, or {@code null} to inherit it from the environment, namely
+     * {@code FDB_CLUSTER_FILE}.
+     * @param includeMethodInDescriptions Set this to {@code true} if publishing test results to something that cannot
+     * handle complex test hierarchies. In the record layer we maintain the full hierarchy in the output, so this is not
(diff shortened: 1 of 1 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -66,13 +66,23 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
...
-        this.clusterFile = null; // it will get it from the environment
+        this(null, false);
     }
 
-    public YamlTestExtension(@Nullable final String clusterFile) {
+    /**
+     * Create a new extension with some configuration.
+     * @param clusterFile a custom cluster file to use, or {@code null} to inherit it from the environment, namely
+     * {@code FDB_CLUSTER_FILE}.
+     * @param includeMethodInDescriptions Set this to {@code true} if publishing test results to something that cannot
```
alecgrieser, line 79: Why not just set this to `true` always? Does the name look bad when running in an IDE or something?

alecgrieser (approved):
This looks fine to me. I'm unclear as to why we wouldn't just always include the method names. I've definitely had trouble parsing our own JUnit reports when there are test failures due to the convention of not including the method name, and I'd be fine if they were just always there.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
ScottDugas: And I did some investigation into that HTML report, and it looks like the title there is hardcoded:
https://github.com/gradle/gradle/blob/1f2f66e0320559f3391cee8f79ba96dccd6cebcc/platforms/software/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/report/PageRenderer.java#L76

Issue #146, opened 2018-11-13
Add documentation/example on continuations/pagination


alecgrieser:
We have a little bit of this in our FAQ currently, but explaining how to use continuations to achieve pagination is probably a good idea. There are a few issues filed right now that will change how that is done, but there should probably be docs on the current code nonetheless.

PR #3254, opened 2025-03-17
Add partial result to AggregateCursor continuation

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -385,7 +434,7 @@ public static RecordQueryStreamingAggregationPlan fromProto(@Nonnull final PlanS
         final CorrelationIdentifier groupingKeyAlias = CorrelationIdentifier.of(Objects.requireNonNull(recordQueryStreamingAggregationPlanProto.getGroupingKeyAlias()));
         final CorrelationIdentifier aggregateAlias = CorrelationIdentifier.of(Objects.requireNonNull(recordQueryStreamingAggregationPlanProto.getAggregateAlias()));
         final Value completeResultValue = Value.fromValueProto(serializationContext, Objects.requireNonNull(recordQueryStreamingAggregationPlanProto.getCompleteResultValue()));
-        final boolean isCreateDefaultOnEmpty = recordQueryStreamingAggregationPlanProto.hasIsCreateDefaultOnEmpty() ? recordQueryStreamingAggregationPlanProto.getIsCreateDefaultOnEmpty() : true;
+        final boolean isCreateDefaultOnEmpty = recordQueryStreamingAggregationPlanProto.hasIsCreateDefaultOnEmpty() && recordQueryStreamingAggregationPlanProto.getIsCreateDefaultOnEmpty();
```
alecgrieser, line 437: I think this changes the semantics. I believe prior to this change, if the proto is missing `is_create_default_on_empty`, then it defaults to `true`. With this change, it now defaults to `false`. That's not what we want, because versions that didn't know about the field would default to `true`, and we wanted to continue defaulting to that old behavior.

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -133,6 +133,21 @@ message RecursiveCursorContinuation {
...
+}
+
+message PartialAggregationResult {
+  optional bytes group_key = 1;
+  repeated AccumulatorState accumulator_states = 2;
+}
+
+message AccumulatorState {
+  optional string physical_operator_name = 1;
+  repeated string state = 2; // 2 for avg
```
alecgrieser, line 148: I don't think `string` is the right data type for this. It seems like it would be better for the state object to be something like a `oneof` the following:

1. `int32`
2. `int64`
3. `float`
4. `double`
5. `bytes`

I think those are the value types currently supported by aggregates. I could also see `bytes` being used for `Message`s if we wanted to go down that route for average/record constructor values.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -351,27 +367,60 @@ public Value getCompleteResultValue() {
...
-                .setIsCreateDefaultOnEmpty(false);
-        return builder.build();
     }
 
     @Nonnull
     @Override
     public PRecordQueryPlan toRecordQueryPlanProto(@Nonnull final PlanSerializationContext serializationContext) {
-        return PRecordQueryPlan.newBuilder().setStreamingAggregationPlan(toProto(serializationContext)).build();
+        if (serializationMode == SerializationMode.TO_OLD) {
+            return PRecordQueryPlan.newBuilder().setStreamingAggregationPlan((PRecordQueryStreamingAggregationPlan)toProto(serializationContext)).build();
```
alecgrieser, line 401: It might make sense to just have to methods, `toProtoOld` and `toProtoNew`, which produce the correct types, and then you don't need to have to cast and you don't need to do another `if` statement on the `serializationMode`

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -133,6 +133,21 @@ message RecursiveCursorContinuation {
...
+    optional PartialAggregationResult partial_aggregation_results = 2;
+}
+
+message PartialAggregationResult {
+  optional bytes group_key = 1;
+  repeated AccumulatorState accumulator_states = 2;
+}
+
+message AccumulatorState {
+  optional string physical_operator_name = 1;
```
alecgrieser, line 147: Is the physical operator name actually used? From what I can tell, it grabs that information from the plan rather than the serialized `AccumulatorState`, which seems right anyway.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
```diff
@@ -377,14 +379,43 @@ public Object finish() {
...
                 return childAccumulatorsBuilder.build();
             }
+
+            @Nullable
+            @Override
+            public RecordCursorProto.PartialAggregationResult getPartialAggregationResult(Message groupingKey) {
+                List<RecordCursorProto.AccumulatorState> accumulatorStates = new ArrayList<>();
+                for (Accumulator accumulator: childAccumulators) {
+                    if (accumulator.getPartialAggregationResult(groupingKey) != null) {
+                        accumulatorStates.addAll(accumulator.getPartialAggregationResult(groupingKey).getAccumulatorStatesList());
```
alecgrieser, line 408: Updating this method so that it only returns the list of `AccumulatorState`s would seem to tidy this up a bit, in that it's currently attaching the grouping key to each state only to throw it away and then attach it at a higher level

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
```diff
@@ -377,14 +379,43 @@ public Object finish() {
...
+            private List<Accumulator> buildAccumulators(@Nullable List<RecordCursorProto.AccumulatorState> initialState) {
+                List<Value> childrenAsList = new ArrayList<>();
+                getChildren().forEach(childrenAsList::add);
                 final ImmutableList.Builder<Accumulator> childAccumulatorsBuilder = ImmutableList.builder();
-                for (final var child : getChildren()) {
-                    Verify.verify(child instanceof AggregateValue);
-                    childAccumulatorsBuilder.add(((AggregateValue)child).createAccumulator(typeRepository));
+
+                if (initialState != null) {
+                    Verify.verify(initialState.size() == childrenAsList.size());
```
alecgrieser, line 388: Can this be false due to some kind of weird composition? I'm thinking something like what if one of the child accumulators is an `avg`, and so it actually produces more state. It would seem like we'd either need to have state's nest, or we'd need to handle the fact that the list can be flattened

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -305,6 +313,96 @@ void aggregateNoRecordsNoGroupNoAggregate(final boolean useNestedResult, final i
...
+
+            final var plan =
+                    new AggregationPlanBuilder(recordStore.getRecordMetaData(), "MySimpleRecord")
+                            .withAggregateValue("num_value_2", value -> new NumericAggregationValue.Sum(NumericAggregationValue.PhysicalOperator.SUM_I, value))
+                            .withGroupCriterion("str_value_indexed")
+                            .build(false);
+
+            // In the testing data, there are 2 groups, each group has 3 rows.
+            // recordScanLimit = 5: scans 3 rows, and the 4th scan hits SCAN_LIMIT_REACHED
+            // although the first group contains exactly 3 rows, we don't know we've finished the first group before we get to the 4th row, so nothing is returned, continuation is back to START
```
alecgrieser, line 329: Well, the continuation isn't quite `START`, but I think the rest of this comment is true

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -455,9 +578,9 @@ public RecordQueryPlan build(final boolean useNestedResult) {
             final var groupingKeyValue = RecordConstructorValue.ofUnnamed(groupValues);
             final var aggregateValue = RecordConstructorValue.ofUnnamed(aggregateValues);
             if (useNestedResult) {
-                return RecordQueryStreamingAggregationPlan.ofNested(quantifier, groupingKeyValue, aggregateValue);
+                return RecordQueryStreamingAggregationPlan.ofNested(quantifier, groupingKeyValue, aggregateValue, RecordQueryStreamingAggregationPlan.SerializationMode.TO_NEW);
```
alecgrieser, line 581: We probably want to have a bit more of a mix here. Until we're done rolling this out, we probably need both `TO_NEW` and `TO_OLD` tests, which probably means adding this as a builder property and then setting it where appropriate. (That might mean parameterizing much of the existing tests.)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -305,6 +313,96 @@ void aggregateNoRecordsNoGroupNoAggregate(final boolean useNestedResult, final i
...
+
+    @Test
+    void partialAggregateBitmap() {
+        try (final var context = openContext()) {
+            openSimpleRecordStore(context, NO_HOOK);
+
+            final var plan =
+                    new AggregationPlanBuilder(recordStore.getRecordMetaData(), "MySimpleRecord")
+                            .withAggregateValue("num_value_2", value -> new NumericAggregationValue.BitmapConstructAgg(NumericAggregationValue.PhysicalOperator.BITMAP_CONSTRUCT_AGG_I, value))
+                            .withGroupCriterion("str_value_indexed")
```
alecgrieser, line 380: I think we could use a test here that uses the record scan limit but that doesn't have a grouping key.

Another interesting test might be: one where the aggregation plan sits on top of a filter cursor. I think as written, we won't make progress if, say, the filter cursor removes all of (or a lot of) the records below, but it would be nice to test that (and/or fix that bug with an appropriate test)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -97,6 +98,7 @@ public class RecordQueryStreamingAggregationPlan implements RecordQueryPlanWithC
     private final CorrelationIdentifier aggregateAlias;
     @Nonnull
     private final Value completeResultValue;
+    private final SerializationMode serializationMode;
```
alecgrieser, line 101: Should `serializationMode` be added to the plan `equals` method? I think that would be useful in test asserts to ensure that the plan's serialization mode is correctly encoded

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordLayerIterator.java
```diff
@@ -83,6 +83,7 @@ public boolean hasNext() {
 
     private void fetchNextResult() {
         if (result != null) {
+            continuation = ContinuationImpl.fromRecordCursorContinuation(result.getContinuation());
```
alecgrieser, line 86: What's this about? This change is a little bit notable given that it would affect more than just aggregate queries

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/GroupByQueryTests.java
```diff
@@ -46,6 +49,120 @@ public GroupByQueryTests() {
...
+                    Continuation continuation = null;
+                    // scan pk = 2 and pk = 3 and hit SCAN_LIMIT_REACHED
+                    Assertions.assertTrue(statement.execute(query), "Did not return a result set from a select statement!");
+                    try (final RelationalResultSet resultSet = statement.getResultSet()) {
+                        ResultSetAssert.assertThat(resultSet).hasNextRow()
+                                .isRowExactly(3.0)
+                                .hasNoNextRow();
+                        continuation = resultSet.getContinuation();
+                    }
+                    String postfix = " WITH CONTINUATION B64'" + Base64.getEncoder().encodeToString(continuation.serialize()) + "'";
```
alecgrieser, line 153: These tests should probably use `EXECUTE CONTINUATION` instead of `WITH CONTINUATION`

alecgrieser (changes requested):
There were a few things that looked a little concerning to me. I had some suggestions for a few more tests, but in places where I thought that there may be something wrong with the implementation, it may be a good idea to validate with a test whether the behavior is as I think it is, because I could be wrong about how some of these fit together!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
pengpeng-lu: Good catch! I removed the incorrect comment.

fdb-record-layer-core/src/main/proto/record_cursor.proto
pengpeng-lu: Makes sense, modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -862,5 +913,48 @@ public void accumulate(@Nullable final Object currentObject) {
...
+                case AVG_I:
+                case AVG_L:
+                case AVG_D:
+                case AVG_F:
+                    Pair<?, ?> pair = (Pair<?, ?>) state;
+                    builder.addState(String.valueOf(pair.getLeft())).addState(String.valueOf(pair.getRight()));
+                    break;
+                case BITMAP_CONSTRUCT_AGG_I:
+                case BITMAP_CONSTRUCT_AGG_L:
+                    builder.addState(new String(((BitSet) state).toByteArray(), StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
pengpeng-lu: I changed this to store a `bytes` in protobuf, so this problem is also resolved.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -852,6 +859,50 @@ public NumericAccumulator(@Nonnull final PhysicalOperator physicalOperator) {
...
+                    break;
+                case AVG_F:
+                    state = Pair.of(Float.parseFloat(initialState.getState(0)), Long.parseLong(initialState.getState(1)));
+                    break;
+                case BITMAP_CONSTRUCT_AGG_I:
+                case BITMAP_CONSTRUCT_AGG_L:
+                    state = BitSet.valueOf(initialState.getState(0).getBytes(StandardCharsets.UTF_8));
+                    break;
+                default:
+                    break;
```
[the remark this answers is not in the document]
pengpeng-lu: modified.

fdb-record-layer-core/src/main/proto/record_cursor.proto
pengpeng-lu: It is not used now, unless we want to use it as a verify step (the accumulator's operator name = initialState.operatorName). I removed it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/Accumulator.java
```diff
@@ -29,4 +32,7 @@ public interface Accumulator {
     void accumulate(@Nullable Object currentObject);
 
     @Nullable Object finish();
+
+    @Nullable
+    RecordCursorProto.PartialAggregationResult getPartialAggregationResult(Message groupingKey);
```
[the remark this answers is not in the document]
pengpeng-lu: absolutely a better design. thanks! modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
```diff
@@ -193,4 +208,17 @@ private Object evalGroupingKey(@Nullable final Object currentObject) {
         final EvaluationContext nestedContext = context.withBinding(Bindings.Internal.CORRELATION, alias, currentObject);
         return Objects.requireNonNull(groupingKeyValue).eval(store, nestedContext);
     }
+
+    public boolean isResultOnEmpty() {
+        return groupingKeyValue == null;
+    }
+
+    @Nullable
+    public RecordCursorProto.PartialAggregationResult getPartialAggregationResult(@Nonnull Message groupingKey) {
```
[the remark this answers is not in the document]
pengpeng-lu: removed. Good catch!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
pengpeng-lu: modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
pengpeng-lu: modified

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
pengpeng-lu: fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -113,13 +115,24 @@ private RecordQueryStreamingAggregationPlan(@Nonnull final Quantifier.Physical i
...
+        this.serializationMode = serializationMode;
+    }
+
+    private RecordQueryStreamingAggregationPlan(@Nonnull final Quantifier.Physical inner,
+                                                @Nullable final Value groupingKeyValue,
+                                                @Nonnull final AggregateValue aggregateValue,
+                                                @Nonnull final CorrelationIdentifier groupingKeyAlias,
+                                                @Nonnull final CorrelationIdentifier aggregateAlias,
+                                                @Nonnull final Value completeResultValue) {
+        this(inner, groupingKeyValue, aggregateValue, groupingKeyAlias, aggregateAlias, completeResultValue, SerializationMode.TO_OLD);
```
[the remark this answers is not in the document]
pengpeng-lu: refactored.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
pengpeng-lu: makes sense. modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
```diff
@@ -193,4 +208,17 @@ private Object evalGroupingKey(@Nullable final Object currentObject) {
         final EvaluationContext nestedContext = context.withBinding(Bindings.Internal.CORRELATION, alias, currentObject);
         return Objects.requireNonNull(groupingKeyValue).eval(store, nestedContext);
     }
+
+    public boolean isResultOnEmpty() {
```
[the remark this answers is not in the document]
pengpeng-lu: removed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/NumericAggregationValue.java
```diff
@@ -852,6 +859,50 @@ public NumericAccumulator(@Nonnull final PhysicalOperator physicalOperator) {
             this.physicalOperator = physicalOperator;
         }
 
+        public NumericAccumulator(@Nonnull final PhysicalOperator physicalOperator, @Nonnull final RecordCursorProto.AccumulatorState initialState) {
+            this.physicalOperator = physicalOperator;
+            switch (physicalOperator) {
+                case SUM_I:
+                case MAX_I:
+                case MIN_I:
+                    state = Integer.parseInt(initialState.getState(0));
```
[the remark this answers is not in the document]
pengpeng-lu: added size validation.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -52,29 +58,43 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
...
+                    previousResult.getNoNextReason()));
         }
 
         return AsyncUtil.whileTrue(() -> inner.onNext().thenApply(innerResult -> {
+            lastResult = previousResult;
             previousResult = innerResult;
             if (!innerResult.hasNext()) {
-                if (!isNoRecords()) {
-                    streamGrouping.finalizeGroup();
+                if (!isNoRecords() || streamGrouping.isResultOnEmpty()) {
```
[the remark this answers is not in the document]
pengpeng-lu: right it was from merge.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -147,4 +186,111 @@ public boolean accept(@Nonnull RecordCursorVisitor visitor) {
...
+        }
+
+        @Override
+        public boolean isEnd() {
+            return isEnd;
+        }
+
+        @Nullable
+        public byte[] getInnerContinuation() {
+            return innerContinuation == null ? null : innerContinuation.toByteArray();
```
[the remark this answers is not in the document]
pengpeng-lu: yes, good catch! I changed `innerContinuation` to `RecordCursorContinuation`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -147,4 +186,111 @@ public boolean accept(@Nonnull RecordCursorVisitor visitor) {
...
+        @Nullable
+        private final ByteString innerContinuation;
+
+        @Nullable
+        private final RecordCursorProto.PartialAggregationResult partialAggregationResult;
+
+        @Nullable
+        private RecordCursorProto.AggregateCursorContinuation cachedProto;
+
+        private final boolean isEnd;
```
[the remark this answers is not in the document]
pengpeng-lu: Got it, removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -147,4 +186,111 @@ public boolean accept(@Nonnull RecordCursorVisitor visitor) {
         }
         return visitor.visitLeave(this);
     }
+
+    public static class AggregateCursorContinuation implements RecordCursorContinuation {
+        @Nullable
+        private final ByteString innerContinuation;
```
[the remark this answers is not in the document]
pengpeng-lu: I see. modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -52,29 +58,43 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
...
+    @Nullable
+    // when previousResult = row x, lastResult = row (x-1); when previousResult = null, lastResult = null
+    private RecordCursorResult<QueryResult> lastResult;
     // Previous non-empty record processed by this cursor
     @Nullable
     private RecordCursorResult<QueryResult> previousValidResult;
+    @Nullable
+    private RecordCursorProto.PartialAggregationResult partialAggregationResult;
+    @Nullable
+    private final byte[] continuation;
```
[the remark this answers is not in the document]
pengpeng-lu: Agree, it is more convenient to differentiate between start and end.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -52,29 +58,43 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
     // Previous record processed by this cursor
     @Nullable
     private RecordCursorResult<QueryResult> previousResult;
+    @Nullable
+    // when previousResult = row x, lastResult = row (x-1); when previousResult = null, lastResult = null
```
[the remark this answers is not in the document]
pengpeng-lu: changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -87,41 +107,60 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
...
+                    } else {
+                        RecordCursorContinuation c = previousValidResult == null ? new AggregateCursorContinuation(continuation, false, serializationMode) : new AggregateCursorContinuation(previousValidResult.getContinuation(), serializationMode);
+                        previousValidResult = previousResult;
+                        return RecordCursorResult.withNextValue(QueryResult.ofComputed(streamGrouping.getCompletedGroupResult()), c);
+                    }
+                } else {
+                    // stopped in the middle of a group
+                    RecordCursorContinuation currentContinuation = new AggregateCursorContinuation(lastResult.getContinuation(), partialAggregationResult, serializationMode);
+                    previousValidResult = previousResult;
+                    return RecordCursorResult.withoutNextValue(currentContinuation, Verify.verifyNotNull(previousResult).getNoNextReason());
```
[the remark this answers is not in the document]
pengpeng-lu: so here is what actually happened.

result 1: (value = `(group1, 40)`, continuation = 40)
result 2: (value = None, continuation = 50, partialAggregationResult = {group2: aggregate_func(41:50) = 10 in this example}, noNextReason = SCAN_LIMIT_REACHED)

then on resuming,
result 3: (value = `(group2, 50)`, continuation = 80, partialAggregationResult = null)
result 4: (value = None, continuation = END, noNextReason = SOURCE_EXHAUSTED)
 

result 1/3 is returned by line#140, result 2 is returned by #156, result 4 is returned by #146.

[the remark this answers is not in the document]
pengpeng-lu: I added this test `FDBStreamAggregationTest.partialAggregateCount` like the above example.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -52,29 +58,43 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
...
+        this.serializationMode = serializationMode;
     }
 
     @Nonnull
     @Override
     public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
         if (previousResult != null && !previousResult.hasNext()) {
             // we are done
-            return CompletableFuture.completedFuture(RecordCursorResult.exhausted());
+            return CompletableFuture.completedFuture(RecordCursorResult.withoutNextValue(new AggregateCursorContinuation(previousResult.getContinuation(), serializationMode),
```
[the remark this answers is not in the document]
pengpeng-lu: I explained why `partialAggregationResult` will not be dropped in another comment, basically it'll be returned in line #156, as part of `continuation`. The change here is to preserve the `noNextReason` for `previousResult`, before it'll just return `sourceExhausted` even when `previousResult` hits other limit.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -87,41 +107,60 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
...
+                    if (previousValidResult == null && partialAggregationResult == null) {
+                        return RecordCursorResult.exhausted();
+                    } else {
+                        RecordCursorContinuation c = previousValidResult == null ? new AggregateCursorContinuation(continuation, false, serializationMode) : new AggregateCursorContinuation(previousValidResult.getContinuation(), serializationMode);
+                        previousValidResult = previousResult;
+                        return RecordCursorResult.withNextValue(QueryResult.ofComputed(streamGrouping.getCompletedGroupResult()), c);
+                    }
+                } else {
+                    // stopped in the middle of a group
+                    RecordCursorContinuation currentContinuation = new AggregateCursorContinuation(lastResult.getContinuation(), partialAggregationResult, serializationMode);
```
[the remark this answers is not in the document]
pengpeng-lu: you're right, removed `lastResult`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -52,29 +58,43 @@ public class AggregateCursor<M extends Message> implements RecordCursor<QueryRes
     // Previous record processed by this cursor
     @Nullable
     private RecordCursorResult<QueryResult> previousResult;
+    @Nullable
+    // when previousResult = row x, lastResult = row (x-1); when previousResult = null, lastResult = null
+    private RecordCursorResult<QueryResult> lastResult;
```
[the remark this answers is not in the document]
pengpeng-lu: removed `lastResult`, thanks for catching this!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RecordConstructorValue.java
pengpeng-lu: I added a test `FDBStreamAggregationPlanTest.partialAggregateAggregateThreeGroupByTwo`, where it has 3 aggregator, one of them is `avg`. It works correctly, because for `avg`, `initialState = [state, state]` as we defined it as a repeated field.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/StreamGrouping.java
```diff
@@ -167,20 +180,22 @@ private boolean isGroupBreak(final Object currentGroup, final Object nextGroup)
...
 
-    private void finalizeGroup(Object nextGroup) {
+    private RecordCursorProto.PartialAggregationResult finalizeGroup(Object nextGroup) {
         final EvaluationContext nestedContext = context.childBuilder()
                 .setBinding(groupingKeyAlias, currentGroup)
                 .setBinding(aggregateAlias, accumulator.finish())
                 .build(context.getTypeRepository());
         previousCompleteResult = completeResultValue.eval(store, nestedContext);
 
+        RecordCursorProto.PartialAggregationResult result = currentGroup == null ? null : getPartialAggregationResult((Message) currentGroup);
```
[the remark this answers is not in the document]
pengpeng-lu: good catch! I removed the `null` check part.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -87,41 +107,60 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
...
+                previousValidResult = previousResult;
+                return RecordCursorResult.withNextValue(QueryResult.ofComputed(streamGrouping.getCompletedGroupResult()), c);
+            } else {
+                // innerResult.hasNext() = false, might stop in the middle of a group
+                if (Verify.verifyNotNull(previousResult).getNoNextReason() == NoNextReason.SOURCE_EXHAUSTED) {
+                    // exhausted
+                    if (previousValidResult == null && partialAggregationResult == null) {
+                        return RecordCursorResult.exhausted();
+                    } else {
+                        RecordCursorContinuation c = previousValidResult == null ? new AggregateCursorContinuation(continuation, false, serializationMode) : new AggregateCursorContinuation(previousValidResult.getContinuation(), serializationMode);
```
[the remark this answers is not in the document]
pengpeng-lu: refactored.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordLayerIterator.java
pengpeng-lu: removed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
pengpeng-lu: so I added more parameterized tests in `FDBStreamAggregationTest`, do you think we need more in other tests? if that's the case, we'll probably need to put it in `RecordQueryPlanConfiguration`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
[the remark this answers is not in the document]
pengpeng-lu: changed the behavior of TO_OLD, and added tests for it in `FDBStreamAggregationTest`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -147,4 +193,84 @@ public boolean accept(@Nonnull RecordCursorVisitor visitor) {
...
+        @Nullable
+        @Override
+        public byte[] toBytes() {
+            ByteString byteString = toByteString();
+            return byteString.isEmpty() ? null : byteString.toByteArray();
+        }
+
+        @Override
+        public boolean isEnd() {
+            return innerContinuation.isEnd();
```
alecgrieser, line 238: There is something about this that is still a bit suspicious. In general, we should only return continuations where `isEnd()` returns `true` if we are also returning `RecordCursorResult.exhausted()`. In particular, we should never have a `RecordCursorResult` with a value _and_ a continuation where `isEnd` is true. If I'm understanding the code correctly, I actually think that this invariant is currently true (and may actually be necessary because of the check on line 113 where we create an `AggregateCursorContinuation` with the previous continuation value, though we could adjust that to return `RecordCursorResult.exhausted()` if `previousContinuationInGroup.isEnd()`).

The issue I have is that I don't think this works well with a slightly different variant we might want to make in the future, where we make the `innerContinuation` the previous _read_ continuation (with a partial aggregation result) rather than the previous _returned_ continuation, because in that case, we'd expect the final value to wrap an inner end continuation, but for the result to nevertheless have a value. We have to be a little careful in that case to not accidentally return an empty continuation (potentially we need a `bool inner_is_end` flag in the continuation to guarantee it won't be empty if both the `innerContinuation` is empty and there's no `partialAggregationResult`).

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/GroupByQueryTests.java
```diff
@@ -46,6 +49,120 @@ public GroupByQueryTests() {
...
+                    // scan pk = 2 and pk = 3 and hit SCAN_LIMIT_REACHED
+                    Assertions.assertTrue(statement.execute(query), "Did not return a result set from a select statement!");
+                    try (final RelationalResultSet resultSet = statement.getResultSet()) {
+                        ResultSetAssert.assertThat(resultSet).hasNextRow()
+                                .isRowExactly(1L, 1L, 20L)
+                                .hasNoNextRow();
+                        continuation = resultSet.getContinuation();
+                    }
+                    // scan pk = 5 and pk = 4 rows, hit SCAN_LIMIT_REACHED
+                    String postfix = "EXECUTE CONTINUATION " + Base64.getEncoder().encodeToString(continuation.serialize());
```
alecgrieser, line 82: This doesn't look right to me. `EXECUTE CONTINUATION` should be its own command rather than a post fix to the query

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
alecgrieser: I think having these be paramaterized on the serialization mode is probably sufficient

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
alecgrieser: Did these tests get added? I'm not actually sure if the filter problem is still there (that is, I'm not sure what will happen if a filter that removes all results from underneath will correctly update the inner continuation or not)

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -83,45 +101,73 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
...
+                } else {
+                    // stopped in the middle of a group
+                    RecordCursorContinuation currentContinuation = new AggregateCursorContinuation(Verify.verifyNotNull(previousResult).getContinuation(), partialAggregationResult, serializationMode);
+                    previousValidResult = previousResult;
+                    previousContinuationInGroup = Verify.verifyNotNull(previousValidResult).getContinuation();
+                    if (serializationMode == RecordQueryStreamingAggregationPlan.SerializationMode.TO_NEW) {
+                        return RecordCursorResult.withoutNextValue(currentContinuation, Verify.verifyNotNull(previousResult).getNoNextReason());
+                    } else {
+                        // for TO_OLD, return {groupKey: partialResult from row x to y}, next time will return {groupKey: partialResult from y+1 to next stop point}
+                        return RecordCursorResult.withNextValue(QueryResult.ofComputed(streamGrouping.getCompletedGroupResult()), currentContinuation);
```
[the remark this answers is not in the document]
pengpeng-lu: yes, I think at least we didn't ignore any rows with this result...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
```diff
@@ -83,45 +101,73 @@ public CompletableFuture<RecordCursorResult<QueryResult>> onNext() {
...
             }
         }), getExecutor()).thenApply(vignore -> {
-            if (isNoRecords()) {
-                // Edge case where there are no records at all
-                return RecordCursorResult.exhausted();
+            // either innerResult.hasNext() = false; or groupBreak = true
+            if (Verify.verifyNotNull(previousResult).hasNext()) {
+                // in this case groupBreak = true, return aggregated result and continuation, partialAggregationResult = null
+                // previousValidResult = null happens when 1st row of current scan != last row of last scan, results in groupBreak = true and previousValidResult = null
+                RecordCursorContinuation c = new AggregateCursorContinuation(previousContinuationInGroup, serializationMode);
```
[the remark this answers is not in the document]
pengpeng-lu: yes, the behavior is (say we are doing count):
row10    group1.     group1_count = 10
row11    group2

we return result = (group1: count = 10), continuation = row10, so the next scan starts with row11 again. There is a little bit of inefficiency here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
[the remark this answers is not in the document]
pengpeng-lu: you are right, added the `partialAggregationResult`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
pengpeng-lu: added the filterCursor tests.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/GroupByQueryTests.java
pengpeng-lu: modified.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/GroupByQueryTests.java
pengpeng-lu: modified.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
pengpeng-lu: changed `isEnd = innerContinuation.isEnd() && partialAggregationResult == null`, for the current implementation, `inner_is_end` is unnecessary so I didn't add it.

alecgrieser: I'm not sure I see what is gained here by also checking the `partialAggregationResult`, if I'm honest.

Yeah, I wouldn't expect `inner_is_end` to be needed unless we modified this to handle the case where the continuation is always the last key read. It is probably fine not to include it, if we decide that we're unlikely to do that

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/cursors/aggregate/AggregateCursor.java
pengpeng-lu: yes I think you are right, I removed the checking partialAggregationResult part here.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBStreamAggregationTest.java
```diff
@@ -372,6 +376,111 @@ void partialAggregateSumToNew() {
...
+        try (final var context = openContext()) {
+            openSimpleRecordStore(context, NO_HOOK);
+
+            // in the table, group1(str_value_indexed = "0") -> num_value_2 = 0, 1, 2, group2(str_value_indexed = "1") -> num_value_2 = 3, 4, 5.
+            final var plan =
+                    new AggregationPlanBuilder(recordStore.getRecordMetaData(), "MySimpleRecord")
+                            .withAggregateValue("num_value_2", value -> new NumericAggregationValue.Sum(NumericAggregationValue.PhysicalOperator.SUM_I, value))
+                            .withGroupCriterion("str_value_indexed")
+                            .withQueryPredicate("num_value_2", Comparisons.Type.LESS_THAN, 0)
+                            .build(false, RecordQueryStreamingAggregationPlan.SerializationMode.TO_NEW);
```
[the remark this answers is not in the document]
pengpeng-lu: made all tests here to test bother `TO_NEW` and `TO_OLD`

alecgrieser (approved):

PR #3256, opened 2025-03-18
Remove Artifactory publishing logic from gradle build

### LINKED ISSUE #3249: Remove Artifactory publishing logic
Since the resolution of #1288, we are now publishing to maven central and github packages instead of artifactory. We no longer need the logic that publishes to the old location, and so we should clean that up.

alecgrieser:
We no longer need the artifactory publishing logic that we used to use for publishing in the gradle build as new releases are no longer being published there. This removes the now dead code from the build, which means we no longer need to include the artficatory plugin in our build-time dependencies, which is nice.

There are a few other build files that appear to be dead code that we can remove, but this is more targeted. I've validated that I can still publish to maven local with this change, and all of the logic seems to be pretty specific to actually running an artifactory build, so this shouldn't disrupt our other build procedures, but I guess one never knows.

This resolves #3249.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,6 @@ plugins {
     alias(libs.plugins.versions)
     alias(libs.plugins.spotbugs)
     alias(libs.plugins.nexus)
-    alias(libs.plugins.artifactory)
     alias(libs.plugins.download)
 }
 
--- a/gradle/publishing.gradle
+++ b/gradle/publishing.gradle
@@ -18,7 +18,6 @@
  * limitations under the License.
  */
 
-apply plugin: 'com.jfrog.artifactory'
 apply plugin: 'signing'
 
 // Add various details to the pom file to allow for publishing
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -37,7 +37,6 @@ plugins {
     alias(libs.plugins.versions)
     alias(libs.plugins.spotbugs)
     alias(libs.plugins.nexus)
-    alias(libs.plugins.artifactory)
```
ScottDugas, line 40: Can this be removed from `libs.versions.toml` now?

gradle/publishing.gradle
```diff
@@ -74,17 +73,6 @@ if (ext.publishLibrary) {
             }
         }
     }
-    artifactoryPublish {
```
ScottDugas, line 77: This does mean that `build.py` will break, but we're not using that anywhere, and plan to remove it very shortly, so that seems fine.

ScottDugas (changes requested):

build.gradle
alecgrieser: Oh whoops. Definitely meant to remove it the first time. Yeah, it's unreferenced now. Removed

ScottDugas (approved):

PR #3258, opened 2025-03-19
Update protobuf dependency to 4.30.1 and grpc-commonProtos dependency to 2.54.1

alecgrieser:
This updates our protocol buffer related dependencies to more recent versions. For protobuf, that means the 4.30.1 and for grpc-commonProtos, that means 2.54.1.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataEvolutionValidator.java
```diff
@@ -251,10 +252,11 @@ private void validateMessage(@Nonnull Descriptor oldDescriptor, @Nonnull Descrip
         }
     }
 
-    @SuppressWarnings("deprecation") // checks the deprecated syntax field
     private void validateProtoSyntax(@Nonnull Descriptors.Descriptor oldDescriptor, @Nonnull Descriptors.Descriptor newDescriptor) {
-        if (!oldDescriptor.getFile().getSyntax().equals(newDescriptor.getFile().getSyntax())
-                || !oldDescriptor.getFile().getEdition().equals(newDescriptor.getFile().getEdition())) {
+        final DescriptorProtos.FileDescriptorProto oldFileDescriptorProto = oldDescriptor.getFile().toProto();
+        final DescriptorProtos.FileDescriptorProto newFileDescriptorProto = newDescriptor.getFile().toProto();
+        if (!oldFileDescriptorProto.getSyntax().equals(oldFileDescriptorProto.getSyntax())
```
alecgrieser: oh whoops, this is the bug that caused the tests to fail: it uses `oldFileDescriptorProto` on both sides.

arnaud-lacurie (approved):

PR #3260, opened 2025-03-19
Prepare for upgrade to Protobuf 4 by removing deprecated or removed calls

### LINKED ISSUE #3259: Prepare to upgrade to Protobuf 4
We may need to publish both versions for a while, so ideally keep the source compatible with both.

MMcM:
Address some small incompatibilities between Protobuf 3 and Protobuf 4, attempting to have code that will compile against both.
* `Descriptors.FileDescriptor.getSyntax`, which was deprecated, is now removed. Use `DescriptorProtos.FileDescriptorProto.getSyntax`.
* `TextFormat.shortDebugString` is deprecated. Remove newlines manually for now (only in test failure codepath).
* guava is no longer a transitive dependency; add it explicitly where needed.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataEvolutionValidator.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/MetaDataEvolutionValidator.java
@@ -251,10 +251,9 @@ public class MetaDataEvolutionValidator {
         }
     }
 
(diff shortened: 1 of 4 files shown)


alecgrieser:
See also #3258. That one seems to hit a problem where the syntax field isn't picked up correctly in the file descriptor, and I'm not sure what we'd want to do about it

MMcM:
#3261 seems to have passed tests; is that unexpected based on the other one?

PR #3266, opened 2025-03-21
Composite aggregates

normen662:
This PR adds the ability to plan arbitrary aggregation queries using aggregation indexes as well as intersections of aggregation indexes. It also corrects/adds the infrastructure to deal with complicated order-by requirements.

- refactor data access rules into a data access rule for value indexes and one for aggregation indexes
- have both of these data access rule implementations share the maximum code (only intersection planning is different)
- intersection planning for aggregation indexes
   - because of the same reasoning as regular index intersections AND in order to answer the query as these indexes carry data that is not part of the underlying table (the aggregation)
   - rollups of finer-granularity index scans if such a roll-up can be useful in answering the query
- order by requirements are treated in the same framework as all other kinds of expressions (using M3 maps, translations, etc.)

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
@@ -319,6 +319,12 @@ public abstract class AbstractDataAccessRule extends CascadesRule<MatchPartition
             return LinkedIdentitySet.of();
         }
 
+        if (logger.isTraceEnabled()) {
+            for (Vectored<SingleMatchedAccess> bestMaximumCoverageMatch : bestMaximumCoverageMatches) {
+                logger.trace("single access = {}", bestMaximumCoverageMatch.getElement());
+            }
+        }
+
         // create scans for all best matches
         final var bestMatchToPlanMap =
                 createScansForMatches(call.getContext(), call, bestMaximumCoverageMatches);
@@ -433,6 +439,11 @@ public abstract class AbstractDataAccessRule extends CascadesRule<MatchPartition
                             bestMatchToDistinctPlanMap,
                             binaryPartition,
                             requestedOrderings);
+
+            if (logger.isTraceEnabled()) {
+                logger.trace("binary intersection result = {}", binaryIntersections);
+            }
+
             if (binaryIntersections.hasViableIntersection()) {
                 updateIntersectionInfoMap(intersectionInfoMap, binaryPartition, binaryIntersections);
             } else {
@@ -490,6 +501,10 @@ public abstract class AbstractDataAccessRule extends CascadesRule<MatchPartition
                                     bestMatchToDistinctPlanMap,
                                     kPartition,
                                     requestedOrderings);
+
+                    if (logger.isTraceEnabled()) {
+                        logger.trace("intersection result = {}", intersectionResult);
+                    }
 
                     if (!intersectionResult.hasViableIntersection()) {
                         continue;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlan.java
@@ -31,7 +31,6 @@ import com.apple.foundationdb.record.planprotos.PRecordQueryPlan;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStoreBase;
 import com.apple.foundationdb.record.provider.foundationdb.cursors.IntersectionMultiCursor;
 import com.apple.foundationdb.record.query.plan.HeuristicPlanner;
-import com.apple.foundationdb.record.query.plan.cascades.AliasMap;
 import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.cascades.FinalMemoizer;
 import com.apple.foundationdb.record.query.plan.cascades.OrderingPart.ProvidedOrderingPart;
@@ -43,7 +42,6 @@ import com.apple.foundationdb.record.query.plan.cascades.explain.NodeInfo;
 import com.apple.foundationdb.record.query.plan.cascades.explain.PlannerGraph;
 import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
-import com.apple.foundationdb.record.query.plan.cascades.values.simplification.DefaultValueSimplificationRuleSet;
 import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
 import com.google.auto.service.AutoService;
 import com.google.common.base.Verify;
@@ -107,12 +105,7 @@ public class RecordQueryMultiIntersectionOnValuesPlan extends RecordQueryInterse
     @Override
     public List<? extends Value> getRequiredValues(@Nonnull final CorrelationIdentifier newBaseAlias,
                                                    @Nonnull final Type inputType) {
-        final var ruleSet = DefaultValueSimplificationRuleSet.instance();
-        return getComparisonKeyValues().stream()
-                .map(comparisonKeyValue ->
-                        comparisonKeyValue.rebase(AliasMap.ofAliases(Quantifier.current(), newBaseAlias))
-                                .simplify(ruleSet, EvaluationContext.empty(), AliasMap.emptyMap(), getCorrelatedTo()))
-                .collect(ImmutableList.toImmutableList());
+        throw new UnsupportedOperationException();
     }
 
     @HeuristicPlanner
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/QueryPredicate.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/QueryPredicate.java
@@ -238,19 +238,11 @@ public interface QueryPredicate extends Correlated<QueryPredicate>, TreeLike<Que
     }
 
     @Nonnull
-    default PredicateCompensationFunction computeCompensationFunction(@Nonnull final PartialMatch partialMatch,
-                                                                      @Nonnull final QueryPredicate originalQueryPredicate,
-                                                                      @Nonnull final Map<CorrelationIdentifier, ComparisonRange> boundParameterPrefixMap,
-                                                                      @Nonnull final List<PredicateCompensationFunction> childrenResults,
-                                                                      @Nonnull final PullUp pullUp) {
-        Verify.verify(this instanceof LeafQueryPredicate);
-        Verify.verify(childrenResults.isEmpty());
-
-        return toResidualPredicate()
-                .replaceValuesMaybe(pullUp::pullUpValueMaybe)
-                .map(PredicateCompensationFunction::ofPredicate)
-                .orElse(PredicateCompensationFunction.impossibleCompensation());
-    }
+    PredicateCompensationFunction computeCompensationFunction(@Nonnull PartialMatch partialMatch,
+                                                              @Nonnull QueryPredicate originalQueryPredicate,
+                                                              @Nonnull Map<CorrelationIdentifier, ComparisonRange> boundParameterPrefixMap,
+                                                              @Nonnull List<PredicateCompensationFunction> childrenResults,
+                                                              @Nonnull PullUp pullUp);
 
     /**
      * Create a {@link QueryPredicate} that is equivalent to {@code this} but which is evaluated as a residual
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
@@ -157,12 +157,12 @@ public class RecordQueryScanPlan implements RecordQueryPlanWithNoChildren, Recor
      */
     @VisibleForTesting
     public RecordQueryScanPlan(@Nullable Set<String> recordTypes,
-                                @Nonnull Type flowedType,
-                                @Nullable KeyExpression commonPrimaryKey,
-                                @Nonnull ScanComparisons comparisons,
-                                boolean reverse,
-                                boolean strictlySorted,
-                                @Nonnull final Optional<? extends WithPrimaryKeyMatchCandidate> matchCandidateOptional) {
+                               @Nonnull Type flowedType,
+                               @Nullable KeyExpression commonPrimaryKey,
+                               @Nonnull ScanComparisons comparisons,
+                               boolean reverse,
+                               boolean strictlySorted,
+                               @Nonnull final Optional<? extends WithPrimaryKeyMatchCandidate> matchCandidateOptional) {
         this.recordTypes = recordTypes == null ? null : ImmutableSet.copyOf(recordTypes);
         this.flowedType = flowedType;
         this.commonPrimaryKey = commonPrimaryKey;
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RootLogLevelExtension.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/RootLogLevelExtension.java
@@ -1,0 +1,52 @@
+/*
+ * RootLogLevelExtension.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.foundationdb;
+
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.Configurator;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class RootLogLevelExtension implements BeforeEachCallback, AfterEachCallback {
+    private final Level tempLevel;
+    private Level original;
+
+    public RootLogLevelExtension(Level tempLevel) {
+        this.tempLevel = tempLevel;
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext context) {
+        LoggerContext ctx = (LoggerContext)LogManager.getContext(false);
+        Configuration cfg = ctx.getConfiguration();
+        original = cfg.getLoggerConfig(LogManager.ROOT_LOGGER_NAME).getLevel();
+        Configurator.setRootLevel(tempLevel);
+    }
+
+    @Override
+    public void afterEach(ExtensionContext context) {
+        Configurator.setRootLevel(original);
+    }
+}
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java
@@ -30,6 +30,7 @@ import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.Key;
 import com.apple.foundationdb.record.metadata.expressions.KeyExpression;
 import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.RootLogLevelExtension;
 import com.apple.foundationdb.record.query.IndexQueryabilityFilter;
 import com.apple.foundationdb.record.query.expressions.Comparisons;
 import com.apple.foundationdb.record.query.plan.cascades.AccessHints;
@@ -61,11 +62,13 @@ import com.apple.test.Tags;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import org.apache.logging.log4j.Level;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.params.ParameterizedTest;
 
 import javax.annotation.Nonnull;
@@ -130,6 +133,9 @@ import static org.junit.jupiter.api.Assertions.assertNull;
  */
 @Tag(Tags.RequiresFDB)
 public class FDBSimpleQueryGraphTest extends FDBRecordStoreQueryTestBase {
+    @RegisterExtension
+    static RootLogLevelExtension logLevel = new RootLogLevelExtension(Level.TRACE);
+
     @DualPlannerTest(planner = DualPlannerTest.Planner.CASCADES)
     void testSimplePlanGraph() {
         CascadesPlanner cascadesPlanner = setUp();
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/CompensationTests.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/CompensationTests.java
@@ -18,21 +18,11 @@
  * limitations under the License.
  */
 
-package com.apple.foundationdb.record.provider.foundationdb.query.plan.cascades;
+package com.apple.foundationdb.record.query.plan.cascades;
 
 import com.apple.foundationdb.record.query.expressions.Comparisons;
-import com.apple.foundationdb.record.query.plan.cascades.AccessHints;
-import com.apple.foundationdb.record.query.plan.cascades.Compensation;
-import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
-import com.apple.foundationdb.record.query.plan.cascades.GraphExpansion;
-import com.apple.foundationdb.record.query.plan.cascades.GroupByMappings;
-import com.apple.foundationdb.record.query.plan.cascades.LinkedIdentityMap;
-import com.apple.foundationdb.record.query.plan.cascades.Memoizer;
-import com.apple.foundationdb.record.query.plan.cascades.PlannerStage;
 import com.apple.foundationdb.record.query.plan.cascades.PredicateMultiMap.PredicateCompensationFunction;
 import com.apple.foundationdb.record.query.plan.cascades.PredicateMultiMap.ResultCompensationFunction;
-import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
-import com.apple.foundationdb.record.query.plan.cascades.RuleTestHelper;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.FullUnorderedScanExpression;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.LogicalFilterExpression;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/MatchInfoTests.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/MatchInfoTests.java
@@ -1,0 +1,78 @@
+/*
+ * MatchInfoTests.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import com.apple.foundationdb.record.query.plan.cascades.expressions.FullUnorderedScanExpression;
+import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
+import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import javax.annotation.Nonnull;
+
+public class MatchInfoTests {
+    @Test
+    void testCollectPredicates() {
+        final var memoizer = Memoizer.noMemoization(PlannerStage.PLANNED);
+
+        final var fuseQuery = fuse();
+        final var fuseCandidate = fuse();
+        final var aCandidate = CorrelationIdentifier.of("a'");
+        final var quantifierOverFuseCandidate =
+                Quantifier.forEach(memoizer.memoizeExploratoryExpression(fuseCandidate), aCandidate);
+
+        final var selectOverFuseCandidate =
+                GraphExpansion.builder()
+                        .addQuantifier(quantifierOverFuseCandidate)
+                        .build().buildSimpleSelectOverQuantifier(quantifierOverFuseCandidate);
+
+        final var matchInfos =
+                fuseQuery.exactlySubsumedBy(fuseCandidate, AliasMap.emptyMap(), IdentityBiMap.create(), TranslationMap.empty());
+        Assertions.assertThat(matchInfos).hasSizeGreaterThan(0);
+        final var matchInfo = Iterables.getFirst(matchInfos, null);
+        final var maxMatchMap = matchInfo.getMaxMatchMap();
+        final var adjustedMaxMatchMapOptional =
+                maxMatchMap.adjustMaybe(aCandidate, selectOverFuseCandidate.getResultValue(), ImmutableSet.of(aCandidate));
+        Assertions.assertThat(adjustedMaxMatchMapOptional.isPresent()).isTrue();
+
+        final var adjustedMatch =
+                matchInfo.adjustedBuilder()
+                        .setMaxMatchMap(adjustedMaxMatchMapOptional.get())
+                        .build();
+
+        Assertions.assertThat(adjustedMatch.collectPulledUpPredicateMappings(selectOverFuseCandidate, ImmutableSet.of()))
+                .isEmpty();
+    }
+
+    @Nonnull
+    private static FullUnorderedScanExpression fuse() {
+        return new FullUnorderedScanExpression(ImmutableSet.of("someType"),
+                someRecordType(), new AccessHints());
+    }
+
+    @Nonnull
+    private static Type.Record someRecordType() {
+        return RuleTestHelper.TYPE_S;
+    }
+
+}
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlanTests.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlanTests.java
@@ -1,0 +1,89 @@
+/*
+ * RecordQueryMultiIntersectionOnValuesPlanTests.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.query.plan.plans;
+
+import com.apple.foundationdb.record.metadata.expressions.EmptyKeyExpression;
+import com.apple.foundationdb.record.query.plan.ScanComparisons;
+import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
+import com.apple.foundationdb.record.query.plan.cascades.Memoizer;
+import com.apple.foundationdb.record.query.plan.cascades.PlannerStage;
+import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
+import com.apple.foundationdb.record.query.plan.cascades.RuleTestHelper;
+import com.apple.foundationdb.record.query.plan.cascades.typing.Type;
+import com.apple.foundationdb.record.query.plan.cascades.values.QuantifiedObjectValue;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import javax.annotation.Nonnull;
+import java.util.Optional;
+
+public class RecordQueryMultiIntersectionOnValuesPlanTests {
+    @Test
+    public void testHeuristicPlannerEntryPoints() {
+        final var memoizer = Memoizer.noMemoization(PlannerStage.PLANNED);
+        final var scanA = memoizer.memoizePlan(scan());
+        final var qunA = Quantifier.physical(scanA, CorrelationIdentifier.of("a"));
+        final var scanB = memoizer.memoizePlan(scan());
+        final var qunB = Quantifier.physical(scanB, CorrelationIdentifier.of("b"));
+
+        final var intersectionPLan =
+                RecordQueryMultiIntersectionOnValuesPlan.intersection(ImmutableList.of(qunA, qunB),
+                        ImmutableList.of(),
+                        QuantifiedObjectValue.of(qunA),
+                        false);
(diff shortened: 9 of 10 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/combinatorics/EnumeratingIterable.java
```diff
@@ -44,6 +44,10 @@ static <T> EnumeratingIterable<T> singleIterable(@Nonnull final T singleElement)
         return new SingleIterable<>(singleElement);
     }
 
+    static <T> EnumeratingIterable<T> emptyOnEmptyIterable() {
+        return new SingleIterable<>();
```
hatyo, line 48: It looks like the reason for using `SingleIterator` (with `null` argument) instead of `EmptyIterator` is related to returning empty list, compared to returning `null` in `EmptyIterator` when calling `computeNext`, is that so, why not just use `EmptyIterator` instead and let the customer deal with the `null`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
```diff
@@ -750,12 +750,15 @@ static Optional<TranslationMap> pullUpAndComposeTranslationMapsMaybe(@Nonnull fi
...
+     * @param partialMatch the partial match already existing between {@code expression} and the only child of
+     *        {@code this}
+     * @param candidateQuantifier the quantifier we adjust along
      * @return {@code Optional.empty()} if the match could not be adjusted, Optional.of(matchInfo) for a new adjusted
      *         match, otherwise.
      */
     @Nonnull
-    default Optional<MatchInfo> adjustMatch(@Nonnull final PartialMatch partialMatch) {
+    default Optional<MatchInfo> adjustMatch(@Nonnull final PartialMatch partialMatch,
+                                            @Nonnull final Quantifier candidateQuantifier) {
```
hatyo: Isn't this quantifier the one of the child? can we not always infer it considering that `this` should have a single child.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/MatchableSortExpression.java
```diff
@@ -212,7 +212,8 @@ public MatchableSortExpression translateCorrelations(@Nonnull final TranslationM
 
     @Nonnull
     @Override
-    public Optional<MatchInfo> adjustMatch(@Nonnull final PartialMatch partialMatch) {
+    public Optional<MatchInfo> adjustMatch(@Nonnull final PartialMatch partialMatch,
+                                           @Nonnull final Quantifier candidateQuantifier) {
         final var childMatchInfo = partialMatch.getMatchInfo();
         final var maxMatchMap = childMatchInfo.getMaxMatchMap();
         final var innerQuantifier = Iterables.getOnlyElement(getQuantifiers());
```
hatyo: `innerQuantifier` and `candidateQuantifier` are always the same I think.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
```diff
@@ -70,7 +70,7 @@ public class YamlTestExtension implements TestTemplateInvocationContextProvider,
     private final boolean includeMethodInDescriptions;
 
     public YamlTestExtension() {
-        this(null, false);
+        this(null, true);
```
hatyo: is this necessary?

yaml-tests/src/test/resources/bitmap-aggregate-index.yamsql
```diff
@@ -47,16 +47,16 @@ setup:
...
-      - query: SELECT bitmap_construct_agg(bitmap_bit_position(id)) as bitmap, bitmap_bucket_offset(id) as offset FROM T1 GROUP BY bitmap_bucket_offset(id)
-      - explain: "AISCAN(BITMAPINDEX1 <,> BY_GROUP -> [_0: KEY:[0], _1: VALUE:[0]]) | MAP (_._1 AS BITMAP, _._0 AS OFFSET)"
-      - unorderedResult: [{BITMAP: xStartsWith_1250'060000c', 'OFFSET':0}, {BITMAP: xStartsWith_1250'02', 'OFFSET':10000}]
-    -
-      - query: SELECT bitmap_construct_agg(bitmap_bit_position(id)) as bitmap, category, bitmap_bucket_offset(id) as offset FROM T1 GROUP BY category, bitmap_bucket_offset(id)
-      - explain: "AISCAN(BITMAPINDEX2 <,> BY_GROUP -> [_0: KEY:[0], _1: KEY:[1], _2: VALUE:[0]]) | MAP (_._2 AS BITMAP, _._0 AS CATEGORY, _._1 AS OFFSET)"
-      - unorderedResult: [{BITMAP: xStartsWith_1250'0200004', 'CATEGORY': 'hello', 'OFFSET':0},
-                          {BITMAP: xStartsWith_1250'02', 'CATEGORY': 'hello', 'OFFSET':10000},
-                          {BITMAP: xStartsWith_1250'0400008', 'CATEGORY': 'world', 'OFFSET':0}]
+    #-
```
hatyo: why are these tests commented out?

yaml-tests/src/test/resources/recursive-cte.yamsql
```diff
@@ -29,7 +29,7 @@ test_block:
             select id, parent from t1 where parent = -1
             union all
             select b.id, b.parent from c1 as a, t1 as b where a.id = b.parent) select id from c1
-      - explain: "RUNION q0, q1 { INITIAL { ISCAN(PARENTIDX [EQUALS promote(@c15 AS LONG)]) | INSERT INTO TEMP q1 } RECURSIVE  { ISCAN(CHILDIDX <,>) | FLATMAP q2 -> { TEMP SCAN base() | FILTER _.ID EQUALS q2.PARENT AS q3 RETURN (q2.ID AS ID, q2.PARENT AS PARENT) } | INSERT INTO TEMP q1 }} | MAP (_.ID AS ID)"
+      - explain: "RUNION q0, q1 { INITIAL { ISCAN(PARENTIDX [EQUALS promote(@c15 AS LONG)]) ∩ ISCAN(CHILDIDX <,>) COMPARE BY (_.ID, recordType(_)) | INSERT INTO TEMP q1 } RECURSIVE  { ISCAN(CHILDIDX <,>) | FLATMAP q2 -> { TEMP SCAN base() | FILTER _.ID EQUALS q2.PARENT AS q3 RETURN (q2.ID AS ID, q2.PARENT AS PARENT) } | INSERT INTO TEMP q1 }} | MAP (_.ID AS ID)"
```
hatyo: That .... does not look ideal to me, why do we have an intersection here?

yaml-tests/src/test/resources/standard-tests.yamsql
```diff
@@ -81,7 +81,7 @@ test_block:
                  {ID: 13, NEWCOL: !null x} ]
     -
       - query: select * from T1 where COL1 = 20
-      - explain: "ISCAN(I1 [EQUALS promote(@c8 AS LONG)])"
+      - explain: "ISCAN(I1 [EQUALS promote(@c8 AS LONG)]) ∩ SCAN(<,>) COMPARE BY (recordType(_), _.ID)"
```
hatyo: I am not sure why we have an intersection here as well.

yaml-tests/src/test/resources/versions-tests.yamsql
```diff
@@ -54,7 +54,7 @@ test_block:
     -
       # Do not include __ROW_VERSION (as a pseudo-column) in *
       - query: select t1.* from t1 where col1 = 10;
-      - explain: "ISCAN(I1 [EQUALS promote(@c10 AS LONG)])"
+      - explain: "ISCAN(I1 [EQUALS promote(@c10 AS LONG)]) ∩ SCAN(<,>) COMPARE BY (recordType(_), _.ID)"
```
hatyo: same here ...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -114,10 +157,81 @@ public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final Correlat
         boolean isImpossible();
 
         @Nonnull
-        Set<QueryPredicate> applyCompensationForPredicate(@Nonnull CorrelationIdentifier baseAlias);
+        PredicateCompensationFunction amend(@Nonnull BiMap<CorrelationIdentifier, Value> unmatchedAggregateMap,
```
hatyo, line 168: Can you please add documentation?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -114,10 +157,81 @@ public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final Correlat
...
+                                            @Nonnull Map<Value, Value> amendedMatchedAggregateMap);
+
+        @Nonnull
+        Set<QueryPredicate> applyCompensationForPredicate(@Nonnull TranslationMap translationMap);
```
hatyo, line 172: I think you might be able to return `QueryPredicate` instead of a `Set<QueryPredicate>` based on the implementations, can you please double-check if this is feasible?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -131,8 +245,50 @@ public boolean isImpossible() {
...
+                @Override
+                public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final TranslationMap translationMap) {
+                    return result;
+                }
+            };
+        }
+
+        @Nonnull
+        static PredicateCompensationFunction ofChildrenCompensationFunctions(@Nonnull final List<PredicateCompensationFunction> childrenCompensationFunctions,
+                                                                             @Nonnull final BiFunction<List<PredicateCompensationFunction>, TranslationMap, Set<QueryPredicate>> compensationFunction) {
```
hatyo, line 271: non-trivial implementations of `compensationFunction` seem to return a `Set` of a single item (as in `OrPredicate` and `AndPredicate` (maybe - check comment there) implementations). Can we instead return a single `QueryPredicate` instead of the `Set<QueryPredicate>`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/AndPredicate.java
```diff
@@ -142,12 +142,13 @@ public PredicateCompensationFunction computeCompensationFunction(@Nonnull final
             return PredicateCompensationFunction.noCompensationNeeded();
         }
 
-        return PredicateCompensationFunction.of(
-                baseAlias -> childrenResults.stream()
-                        .filter(PredicateCompensationFunction::isNeeded)
-                        .flatMap(predicateCompensationFunction ->
-                                predicateCompensationFunction.applyCompensationForPredicate(baseAlias).stream())
-                        .collect(LinkedIdentitySet.toLinkedIdentitySet()));
+        return PredicateCompensationFunction.ofChildrenCompensationFunctions(childrenCompensationFunctions,
```
hatyo, line 145: Can we return a conjunction of the compensated children instead?

```java
        return PredicateCompensationFunction.ofChildrenCompensationFunctions(childrenCompensationFunctions,
                (functions, baseAlias) -> {
                    final var terms = functions.stream()
                            .filter(PredicateCompensationFunction::isNeeded)
                            .flatMap(predicateCompensationFunction ->
                                    predicateCompensationFunction.applyCompensationForPredicate(baseAlias).stream())
                            .collect(ImmutableList.toImmutableList())
                    return AndPredicate.and(terms);
                });
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -114,10 +157,81 @@ public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final Correlat
...
+
+                @Nonnull
+                @Override
+                public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final TranslationMap translationMap) {
+                    return LinkedIdentitySet.of(predicate.translateCorrelations(translationMap, shouldSimplifyValues));
+                }
+            };
+        }
+
+        private static boolean predicateContainsUnmatchedValues(final @Nonnull QueryPredicate pulledUpPredicate) {
```
hatyo: `final @Nonnull` -> `@Nonnull final`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -114,10 +157,81 @@ public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final Correlat
...
+                @Nonnull
+                @Override
+                public PredicateCompensationFunction amend(@Nonnull final BiMap<CorrelationIdentifier, Value> unmatchedAggregateMap,
+                                                           @Nonnull final Map<Value, Value> amendedMatchedAggregateMap) {
+                    final var amendedTranslatedPredicateOptional =
+                            predicate.replaceValuesMaybe(rootValue ->
+                                    Optional.of(replaceNewlyMatchedValues(unmatchedAggregateMap, amendedMatchedAggregateMap,
+                                            rootValue)));
+                    Verify.verify(amendedTranslatedPredicateOptional.isPresent());
+                    return ofPredicate(amendedTranslatedPredicateOptional.get(), true);
```
hatyo, line 204: Why is it ok to simplify the predicate `Value`(s) here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
```diff
@@ -103,7 +139,14 @@ public boolean isImpossible() {
 
                     @Nonnull
                     @Override
-                    public Set<QueryPredicate> applyCompensationForPredicate(@Nonnull final CorrelationIdentifier baseAlias) {
+                    public PredicateCompensationFunction amend(@Nonnull final BiMap<CorrelationIdentifier, Value> unmatchedAggregateMap,
```
hatyo, line 142: would it be cleaner just to pass `GroupByMappings` instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -574,33 +651,33 @@ default Compensation intersect(@Nonnull Compensation otherCompensation) {
                 Verify.verify(resultCompensationFunction.isNeeded());
                 Verify.verify(otherResultCompensationFunction.isNeeded());
                 // pick the one from this side -- it does not matter as both candidates have the same shape
-                newResultResultCompensationFunction = resultCompensationFunction;
+                newResultResultCompensationFunction =
+                        resultCompensationFunction.amend(unmatchedAggregateMap, newMatchedAggregatesMap);
```
hatyo, line 744: So, if want to intersect two compensations:

```
Compensation1: MatchedGrouping A->X1, B -> X2, MatchedAgg: COUNT(*), UnMatchedAgg: MAX(C)

Compensation2:  MatchedGrouping C->X3, B -> X2, MatchedAgg: MIN(Z), MAX(C), UnMatchedAgg: COUNT(*)
```

Both compensations, are individually impossible (because of the unmatched args), however their intersection is not:
```
Intersect(Compensation1, Compensation2): MatchedGrouping A->X1, B -> X2, C->X3, MatchedAgg: COUNT(*),MIN(Z), MAX(C), UnMatchedAgg: <EMPTY>
```

Is that the idea here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -574,33 +651,33 @@ default Compensation intersect(@Nonnull Compensation otherCompensation) {
...
             }
 
-            final var otherCompensationMap = otherWithSelectCompensation.getPredicateCompensationMap();
+            final var otherCompensationMap =
+                    otherWithSelectCompensation.getPredicateCompensationMap();
             final var combinedPredicateMap = new LinkedIdentityMap<QueryPredicate, PredicateCompensationFunction>();
             for (final var entry : getPredicateCompensationMap().entrySet()) {
                 // if the other side does not have compensation for this key, we don't need compensation
-                if (otherCompensationMap.containsKey(entry.getKey())) {
+                final var otherPredicateCompensationFunction = otherCompensationMap.get(entry.getKey());
```
hatyo, line 753: I think we need only to consider predicates which both compensations need to compensate by reapplication. The other predicates are _not_ relevant, considering that either compensation claims it is already handled correctly somewhere down the graph, because otherwise, it would required to compensate for it. Correct?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -738,34 +821,47 @@ public ResultCompensationFunction getResultCompensationFunction() {
...
+            return groupByMappings;
+        }
+
         /**
          * When applied to a reference this method returns a {@link RelationalExpression} consuming the
          * reference passed in that applies additional predicates as expressed by the predicate compensation map.
          *
          * @param memoizer the memoizer for new {@link Reference}s
          * @param relationalExpression root of graph to apply compensation to
+         * @param matchedToRealizedTranslationMapFunction a function that given an alias for the quantifier over the
```
hatyo, line 925: I am not sure I understand the purpose of this function.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -884,5 +983,22 @@ public RelationalExpression applyFinal(@Nonnull final Memoizer memoizer,
                     .build()
                     .buildSelectWithResultValue(resultValue);
         }
+
+        @Nonnull
+        @Override
+        public String toString() {
```
hatyo, line 1078: do you want to keep this or were you only using it for debugging?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
```diff
@@ -557,6 +564,44 @@ public Derivations visitInUnionOnKeyExpressionPlan(@Nonnull final RecordQueryInU
...
+            final var quantifiers = multiIntersectionOnValuesPlan.getQuantifiers();
+
+            for (final var quantifier : quantifiers) {
+                final var childDerivations = derivationsFromQuantifier(quantifier);
+                resultDerivationsBuilder.add(childDerivations.getResultValues());
+                localValuesBuilder.addAll(childDerivations.getLocalValues());
+            }
+
+            final var crossProductIterable =
+                    CrossProduct.crossProduct(resultDerivationsBuilder.build());
```
hatyo, line 585: The cross product here seems expensive, but I guess it is important for exhausting all possible intersections of child derivations.

```
Intersect(P1, P2, P3) | P1's derivations=(V11, V12), P2's derivations=(V21, V22), P3's derivations=(V31)
Intersect Result = RCV(P1', P2' + P3')
Derivations:
RCV(V11, V21 + V31)
RCV(V11, V22 + V31)
RCV(V12, V21 + V31)
RCV(V12, V22 + V31)
```

Correct? I wonder if there is a more compact form to represent these derivations losslessly.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -2126,7 +2126,7 @@ public ExplainTokens describe() {
                 }
 
                 if (i + 1 < getFields().size()) {
-                    return resultExplainTokens.addCommaAndWhiteSpace();
+                    resultExplainTokens.addCommaAndWhiteSpace();
```
hatyo, line 2131: Thanks, the output was very strange indeed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -123,6 +134,98 @@ protected BindingMatcher<R> getExpressionMatcher() {
...
 
+    @Override
+    public void onMatch(@Nonnull final CascadesRuleCall call) {
+        final var bindings = call.getBindings();
+        final var completeMatches = bindings.getAll(getCompleteMatchMatcher());
+        if (completeMatches.isEmpty()) {
+            return;
+        }
+
+        final var expression = bindings.get(getExpressionMatcher());
```
hatyo, line 144: The type of `expression` is generic type `R`. Is it needed? can we get rid of it considering all implementations are specializing `R` for `RelationalExpression`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -123,6 +134,98 @@ protected BindingMatcher<R> getExpressionMatcher() {
...
+            // that are covered by the match partition. Even though that logically forms a join, the existential
+            // quantifiers do not mutate the result of the join, they only cause filtering, that is, the resulting
+            // record is exactly what the for each quantifier produced filtered by the predicates expressed on the
+            // existential quantifiers.
+            //
+            for (final var matchPartitionEntry : matchPartitionsForAliasesByPredicates.entrySet()) {
+                final var matchPartition = matchPartitionEntry.getValue();
+
+                //
+                // The current match partition covers all matches that match the aliases in matchedAliases
```
hatyo: can you please rewrite this comment? I think we're generating here data access for a group of complete matches of a specific child alias, where the matches share the same set of matched query predicates IIUC.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -889,17 +939,83 @@ private static boolean isPartitionRedundant(@Nonnull final Map<BitSet, Intersect
...
         return false;
     }
 
+    @Nonnull
+    private static Optional<Set<CorrelationIdentifier>> unmatchedIdsMaybe(@Nonnull final Map<BitSet, IntersectionInfo> intersectionInfoMap,
+                                                                          @Nonnull final List<Vectored<SingleMatchedAccess>> partition) {
+        final var iterator = partition.iterator();
+
+        var unmatchedIdsOptional =
+                unmatchedIdsMaybe(Objects.requireNonNull(intersectionInfoMap.get(intersectionInfoKey(iterator.next()))));
```
hatyo, line 998: Should we add a check that the `partition` list is not empty?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GroupByMappings.java
```diff
@@ -0,0 +1,73 @@
...
+public class GroupByMappings {
+    @Nonnull
+    private final BiMap<Value, Value> matchedGroupingsMap;
+    @Nonnull
+    private final BiMap<Value, Value> matchedAggregatesMap;
+
+    @Nonnull
+    private final BiMap<CorrelationIdentifier, Value> unmatchedAggregatesMap;
+
+    private GroupByMappings(@Nonnull final BiMap<Value, Value> matchedGroupingsMap,
```
hatyo, line 57: It looks like, both `matchedGroupingMap` and `matchedAggregatesMap` contain the mapping from the _query_ values to the corresponding _match candidate_ values. However, the `unmatchedAggregatedMap` contains a mapping between unmatched aggregate generated `Value` and the corresponding _query_ aggregate `Value`.

This confused me a bit, but this is ok, as long as it is documented or the variables are named in a way the mapping clear.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -503,7 +572,24 @@ private ConstrainedBoolean groupingSubsumedBy(@Nonnull final Quantifier candidat
...
-        return unmatchedCandidateValues.isEmpty() ? booleanWithConstraint : falseValue();
+        if (!unmatchedCandidateValues.isEmpty()) {
+            Verify.verify(candidateGroupingValues.size() > translatedGroupingValuesSet.size());
+
+            //
+            // This is a potential roll-up case, but only if the query side's groupings are completely subsumed
+            // by the prefix of the candidate side. Iterate up to the smaller query side's grouping values to
+            // find out.
+            //
+            for (int i = 0; i < translatedGroupingValues.size(); i++) {
```
hatyo: use `translatedGroupingValuesSet` instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -329,79 +335,143 @@ public Iterable<MatchInfo> subsumedBy(@Nonnull final RelationalExpression candid
...
-                        .collect(ImmutableSet.toImmutableSet());
+                        .collect(ImmutableList.toImmutableList());
 
         //
         // If there are more groupingValues than candidateGroupingValues, we cannot match the index.
         //
-        if (translatedGroupingValues.size() > candidateGroupingValues.size()) {
-            return falseValue();
+        final var unmatchedCandidateValues = new LinkedHashSet<>(candidateGroupingValues);
+        if (translatedGroupingValuesSet.size() > unmatchedCandidateValues.size()) {
```
hatyo, line 471: can you use here in line `473` `candidateGroupingValues` instead of `unmatchedCandidateValues` in the comparison, just to make it a bit clearer.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
```diff
@@ -778,18 +899,19 @@ private static Optional<RelationalExpression> applyCompensationForSingleDataAcce
...
     
     /**
-     * Private helper method to plan an intersection and subsequently compensate it using the partial match structures
+     * Protected helper method to plan an intersection and subsequently compensate it using the partial match structures
      * kept for all participating data accesses.
      * Planning the data access and its compensation for a given match is a two-step approach as we compute
      * the compensation for intersections by intersecting the {@link Compensation} for the single data accesses first
      * before using the resulting {@link Compensation} to compute the compensating expression for the entire
      * intersection.
      * @param memoizer the memoizer
```
hatyo, line 927: documentation of `intersectionInfoMap` is missing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
```diff
@@ -0,0 +1,386 @@
...
+                                        .map(OrderingPart.MatchedOrderingPart::getValue))
+                        .collect(ImmutableSet.toImmutableSet());
+
+        final var isPartitionRedundant =
+                isPartitionRedundant(intersectionInfoMap, partition, equalityBoundKeyValues);
+        if (isPartitionRedundant) {
+            return IntersectionResult.noViableIntersection();
+        }
+
+        final var compensation =
```
hatyo, line 157: I believe we can immediately return `IntersectionResult.noViableIntersection()` if the resulting intersection `compensation` is impossible, correct?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
```diff
@@ -0,0 +1,386 @@
...
+            final var quantifier = quantifiers.get(i);
+            final var pickUpTranslationMap =
+                    TranslationMap.ofAliases(Quantifier.current(), quantifier.getAlias());
+            columnBuilder.add(Column.unnamedOf(pickUpValues.get(i).translateCorrelations(pickUpTranslationMap)));
+        }
+
+        return RecordConstructorValue.ofColumns(columnBuilder.build());
+    }
+
+    private static TranslationMap computeTranslationMap(@Nonnull final CorrelationIdentifier intersectionAlias,
```
hatyo, line 361: Can be tagged with `@Nonnull`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalTypeFilterExpression.java
```diff
@@ -198,27 +204,33 @@ public Compensation compensate(@Nonnull final PartialMatch partialMatch,
...
+                    rootOfMatchPullUp.getCandidateAlias())) {
                 resultCompensationFunction = ResultCompensationFunction.noCompensationNeeded();
             } else {
                 resultCompensationFunction =
-                        ResultCompensationFunction.ofTranslation(pulledUpResultValue, rootPullUp.getNestingAlias());
+                        ResultCompensationFunction.ofValue(pulledUpTranslatedResultValue);
             }
+            isCompensationImpossible |= resultCompensationFunction.isImpossible();
+
+            groupByMappings =
```
hatyo: It would be nice to add some comments explaining the handling of matched group by aggregations and groups here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlan.java
```diff
@@ -0,0 +1,271 @@
...
+        final ExecuteProperties childExecuteProperties = executeProperties.clearSkipAndLimit();
+        return IntersectionMultiCursor.create(
+                        getComparisonKeyFunction().apply(store, context),
+                        reverse,
+                        quantifiers.stream()
+                                .map(physical -> ((Quantifier.Physical)physical).getRangesOverPlan())
+                                .map(childPlan -> (Function<byte[], RecordCursor<QueryResult>>)
+                                        ((byte[] childContinuation) -> childPlan
+                                                .executePlan(store, context, childContinuation, childExecuteProperties)))
+                                .collect(Collectors.toList()),
```
hatyo: Can we use `ImmutableList` here instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
```diff
@@ -0,0 +1,386 @@
...
+                                    baseAlias -> computeTranslationMap(baseAlias,
+                                            newQuantifiers, candidateTopAliasesBuilder.build(),
+                                            (Type.Record)intersectionResultValue.getResultType(),
+                                            commonGroupingKeyValues.size()));
+                    expressionsBuilder.add(compensatedIntersection);
+                }
+            }
+        }
+
+        return IntersectionResult.of(hasCommonOrdering ? intersectionOrdering : null, compensation,
```
hatyo, line 237: I don't think it is possible to retrieve an `IntersectionResult` which non-empty list of `RelationalExpression`s and `null`-common ordering, if we examine the code flow above.

Having said that, we could probably cleanup the code, such that we can remove the `hasCommonFlag` maybe?

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/combinatorics/EnumeratingIterable.java
normen662: `EmptyIterator` does not return any item. It is `endOfData()` immediately. This one does return the empty list. Some combinatorics algorithms want nothing on empty while some others like the topological sort want an empty result on empty input. We thus need both I think.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
normen662: Yup. Changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/MatchableSortExpression.java
normen662: Yup. Changed.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlTestExtension.java
normen662: I like it better that way but I changed it back.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
normen662: No. It may be that we are currently impossible but that there is a common ordering (in fact there must be one for the follow up to work). If there is no common ordering the result will be discarded. If there is a common ordering but it is currently impossible to (think 2 out of 3 aggregate functions are satisfied), we want that intersection access to survive because otherwise we would never get to the intersection that creates the one where we get 3 out of 3.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
normen662: added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryMultiIntersectionOnValuesPlan.java
normen662: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AggregateDataAccessRule.java
normen662: No, You can have a common ordering but no expressions due to the compensation being impossible. You still want to signal the caller to keep this intersection (even though you didn't produce a single expression) in the sieve so you can potentially find a bigger intersection that is not impossible.

yaml-tests/src/test/resources/recursive-cte.yamsql
normen662: You are correct. Need some investigation into this.

yaml-tests/src/test/resources/standard-tests.yamsql
normen662: same as above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/AndPredicate.java
normen662: See above.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: It's always ok to do that. We most of the time don't want to even try since the translation would be from `QOV(q1)` to `QOV(q2)` and we would just let machinery kick in that that would not be able to simplify anything anyway. In this case we play enough shenanigans with the value tree that simplification becomes a need.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
normen662: That is correct.

yaml-tests/src/test/resources/bitmap-aggregate-index.yamsql
normen662: They are now active again!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/LogicalTypeFilterExpression.java
normen662: I extracted the common code between `SelectExpression`, `LogicalTypeFileExpression`, and `GroupByExpression`. The new code also has some documentation.

yaml-tests/src/test/resources/recursive-cte.yamsql
normen662: The _updated_ yaml files were not current. I regenerated everything and this artifact vanished.

yaml-tests/src/test/resources/versions-tests.yamsql
normen662: regenerated the expected plans. things should be fine now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: I feel like that this is a fruitless exercise. There is no clear benefit here from my point of view. You may as well argue that `SelectExpressions` should only contain one predicate as it is the conjunction of all its predicates. Second, this API existed before this PR, so if we want to do something about it, I would like to defer that to after.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PredicateMultiMap.java
normen662: That does not work. Even though `GroupdByMappings` contain an `unmatchedAggregatesMap` as well as the `matchedAggregatedMap`, the semantics of the parameters to this method have to be understood slightly differently: The former uses the `unmatchedAggregateMap` of the original `GroupByMappings` while the latter uses the already amended map of newly matched aggregates.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
normen662: Correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
normen662: Yeah I would like to keep it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
normen662: While you are matching a multitude of indexes (in this case), there will be different parts of the query side subsumed by different accesses which eventually gets planned as the multi intersection. When you then plan that intersection you need to keep track of which intersection leg gives you the aggregate you are looking for. So for instance if your query side says something like `q4` because `q4` gives you a `sum(...)` you need to track that back to the right intersection value. This is later on used for compensations as well as you may refer to `q4` in a predicate or result value. Then `q4` needs to become something like `q10._3._0` where `q10` is the intersection alias, there are 3 groups and we want the first aggregate.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
normen662: Yeah -- not sure if this could be done differently but nothing comes to mind immediately.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
normen662: bugged me for a long time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: Good observation! I removed that!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: I made that more succinct I hope. Please be aware that the code is tailored to the case that we may have a join where certain predicates are not matched unless they are considered local to the already matched aliases. We may want to make that simpler as well as I think with the current join partitioning logic this case does not arise anymore.

normen662: Opened #3565 to track this potentially new work.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: If `partition` is empty, that would constitute as a severe bug. The only thing, however, to gain from this is to throw at the entrance of the method as opposed to when `iterator.next()` is called. Let me just do that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/GroupByMappings.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
normen662: The set is being removed from. So it is in fact the remaining unmatched candidate grouping values set.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
normen662: You uncovered a bug with this comment. Nice! I fixed that. In a slightly different way than it seems obvious.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/AbstractDataAccessRule.java
normen662: added

hatyo (approved):
Great work, let's bring it in.

PR #3271, opened 2025-03-24
Synthetic record orphan policy

ohadzeliger:
Add support for `orphanPolicy` when loading synthetic records. This is necessary for the ability to scrub indexes with synthetic records as dangling index entries, by definition, have no records to back them up.
As part of this PR we've also disabled the repair of synthetic records as part of the scrubbing until we test this feature more thoroughly.
Also of note: Implementors of the `EXPERIMENTAL` `SyntheticRecordType` interface will see a change that requires the implementation of  another method (or add the new parameter to the existing one).
As far as the "breaking change" label: The scrubbing for dangling index entries now uses a `RETURN` policy (default was `ERROR`) which means that previous requests that used to fail now pass (if the record was missing). Also, any request to scrub synthetic record with repair `on` will now fail.


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -469,19 +469,24 @@ void testScrubUnnestedIndex(boolean legacy) {
                         .useLegacyScrubber(legacy)
                         .build())
                 .build()) {
-            indexScrubber.scrubDanglingIndexEntries();
-            indexScrubber.scrubMissingIndexEntries();
+            if (!legacy) {
```
jjezra, line 472: Nit: 
Since most of the other tests are structured as a positive question. I.E.: 
```
if (legacy) {
} else {
}
```
Can we change this to positive to make the code slightly more predictable?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
```diff
@@ -136,17 +139,37 @@ public CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(FDBRecordStor
             } else {
                 futures[i] = store.loadRecordAsync(constituentKey).thenApply(rec -> {
                     if (rec == null) {
```
ScottDugas, line 141: Outside the scope of this PR, `JoinedRecordType` supports outer joins, in theory, but I don't think this code has any handling for the possibility that it is an outer join.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/SyntheticRecordType.java
```diff
@@ -89,7 +90,13 @@ public boolean isSynthetic() {
...
     @API(API.Status.INTERNAL)
     @Nonnull
-    public abstract CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(FDBRecordStore store, Tuple primaryKey);
+    public CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(FDBRecordStore store, Tuple primaryKey) {
+        return loadByPrimaryKeyAsync(store, primaryKey, IndexOrphanBehavior.ERROR);
+    }
+
+    @API(API.Status.INTERNAL)
+    @Nonnull
+    public abstract CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(FDBRecordStore store, Tuple primaryKey, IndexOrphanBehavior orphanBehavior);
```
ScottDugas, line 99: I'm not sure what exactly we should do with `INTERNAL`, but I think the PR description should probably call out that `SyntheticRecordType` has changed the behavior for `loadByPrimaryKeyAsync`, in case anyone is extending it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/UnnestedRecordType.java
```diff
@@ -262,12 +263,21 @@ public NestedConstituent getParentConstituent() {
...
     @Nonnull
     @API(API.Status.INTERNAL)
-    public CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(@Nonnull final FDBRecordStore store, @Nonnull final Tuple primaryKey) {
+    public CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(@Nonnull final FDBRecordStore store, @Nonnull final Tuple primaryKey, IndexOrphanBehavior orphanBehavior) {
         Tuple parentPrimaryKey = primaryKey.getNestedTuple(1);
         return store.loadRecordAsync(parentPrimaryKey).thenApply(storedRecord -> {
             if (storedRecord == null) {
-                throw new RecordDoesNotExistException("constituent record not found: " + parentConstituent.getName())
-                        .addLogInfo(LogMessageKeys.PRIMARY_KEY, parentPrimaryKey);
+                switch (orphanBehavior) {
```
ScottDugas, line 270: Does the code below that checks the child element position need to consider `orphanBehavior`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
```diff
@@ -720,7 +720,23 @@ default CompletableFuture<FDBStoredRecord<M>> loadRecordAsync(@Nonnull final Tup
...
+    default CompletableFuture<FDBSyntheticRecord> loadSyntheticRecord(@Nonnull Tuple primaryKey) {
+        return loadSyntheticRecord(primaryKey, IndexOrphanBehavior.ERROR);
+    }
+
+    /**
+     * Load a {@link FDBSyntheticRecord synthetic record} by loading its stored constituent records and synthesizing it from them.
+     * In case any of the constituents are missing, the following will be returned:
+     * - ERROR: Exception is thrown
+     * - SKIP: Future(null) is returned
+     * - RETURN: Future(Record with no constituents) is returned
```
ScottDugas, line 732: This won't render well in our javadoc, better to be a `<ul>`, and use links for the behaviors, I think:
```suggestion
     * <ul>
     *     <li>{@link IndexOrphanBehavior#ERROR}: Exception is thrown</li>
     *     <li>{@link IndexOrphanBehavior#SKIP}: Future(null) is returned</li>
     *     <li>{@link IndexOrphanBehavior#RETURN}: Future(Record with no constituents) is returned</li>
     * </ul>
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -817,12 +817,12 @@ private CompletableFuture<Void> runSyntheticMaintainers(@Nonnull Map<RecordType,
     @API(API.Status.EXPERIMENTAL)
     @Nonnull
     @Override
-    public CompletableFuture<FDBSyntheticRecord> loadSyntheticRecord(@Nonnull Tuple primaryKey) {
+    public CompletableFuture<FDBSyntheticRecord> loadSyntheticRecord(@Nonnull Tuple primaryKey, final IndexOrphanBehavior orphanBehavior) {
```
ScottDugas, line 820: We should add some tests of this method with the two types of synthetic record types we provide.
I see there are calls in `UnnestedRecordTypeTest`. Unfortunately, I don't see any similar tests for joined record types. It might also make sense to do tests of `loadByPrimarkyKeyAsync` for the two types directly, rather than via this method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
```diff
@@ -136,17 +139,37 @@ public CompletableFuture<FDBSyntheticRecord> loadByPrimaryKeyAsync(FDBRecordStor
...
+            if ( ! isMissingConstituent.get()) {
+                // all constituents have been found
+                return FDBSyntheticRecord.of(this, constituentValues);
+            } else {
+                // some constituents are missing
+                if (orphanBehavior.equals(IndexOrphanBehavior.SKIP)) {
+                    return null;
+                } else {
+                    // This is for RETURN - return the shell of the record with no constituents
+                    return FDBSyntheticRecord.of(this, Map.of());
```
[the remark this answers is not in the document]
ohadzeliger: This was considered as an option, but existing code was testing `getConstituents().isEmpty()` to indicate an issue with the record loading, and also the fact that we would have to introduce some other flag as an indication that the constituent list is incomplete made me think that this option was the original intent. Note that with `ERROR`, which is the default used everywhere you also don't get the successful ones.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
ohadzeliger: I am not sure what kind of support we would need. Can you elaborate (maybe offline) and I'll get a ticket created?

ohadzeliger: Done (opened ticket)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/SyntheticRecordType.java
ohadzeliger: Done (changed PR description).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/UnnestedRecordType.java
ohadzeliger: Do you mean code like
```
int childElemIndex = (int) primaryKey.getNestedTuple(childConstituentIndex + 1).getLong(0);
```
I didn't think so because I thought the assumption was that the constituents are all present by now. Is that not the case?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreBase.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ohadzeliger: Done: Added tests for both `UnnestedRecordType` and `JoinedRecordType` for loading records when constituents are missing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/UnnestedRecordType.java
ScottDugas: As an example, if you have:
```
message Record {
    int32 primaryKey = 1;
    repeated nested = 2;
    message Nested {
        string text = 1;
        int32 num = 2;
    }
}
```

And then you define an Unnested type on `Record.nested`, with `recordTypeKey=7`, and a record:
```
{primaryKey: 5, nested: [{text: "foo", num: 8}]
```

The following primary key for the synthetic type is not valid:
```
[7, [5], [2]]
```
Because there are not 2 elements in `nested`, and it would hit the error below, because you can't actually read both constituents.

Also, if I'm thinking about when you would want to use the `Orphan` behavior, one use case would be if you updated the records, but failed to update the indexes (due to a bug/corruption or potentially to support some sort of optimization). And in that case it would seem reasonable to return, rather than erroring.

ohadzeliger: Done. Added handling of index "out of bounds" to return a result that consults the orphan policy. Also added tests for the same.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -0,0 +1,208 @@
...
+        final FDBStoredRecord<Message> otherRecord;
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, baseMetaData(addJoinedType()));
+
+            Tuple syntheticRecordTypeKey = recordStore.getRecordMetaData()
+                    .getSyntheticRecordType(JOINED_RECORD_NAME)
+                    .getRecordTypeKeyTuple();
+            simpleRecord = recordStore.saveRecord(createSimpleRecord(100, 10));
+            otherRecord = recordStore.saveRecord(createOtherRecord(101, 11));
+            joinedPrimaryKey = Tuple.from(
```
ScottDugas, line 85: hmm, it would be nice if something else provided this, but I don't think we want to be improving that aspect of the synthetic record type api as part of this work.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -0,0 +1,208 @@
...
+                    simpleRecord.getPrimaryKey().getItems(),
+                    otherRecord.getPrimaryKey().getItems());
+
+            context.commit();
+        }
+
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, baseMetaData(addJoinedType()));
+
+            recordStore.deleteRecord(otherRecord.getPrimaryKey());
```
ScottDugas, line 130: Is it worth testing again with `simpleRecord` deleted?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -0,0 +1,208 @@
...
+            assertEquals(simpleRecord, rec.getConstituent(SIMPLE_RECORD));
+            assertEquals(otherRecord, rec.getConstituent(OTHER_RECORD));
+
+            context.commit();
+        }
+    }
+
+    @Test
+    void loadSyntheticRecordFailsMissingConstituent() throws ExecutionException, InterruptedException {
+        final Tuple joinedPrimaryKey;
```
ScottDugas, line 108: All three of these tests have basically the same setup, it seems like it would be nice to extract this into a helper class (probably nested) that saves the two types.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -0,0 +1,208 @@
...
+            createOrOpenRecordStore(context, baseMetaData(addJoinedType()));
+
+            recordStore.deleteRecord(otherRecord.getPrimaryKey());
+
+            context.commit();
+        }
+
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, baseMetaData(addJoinedType()));
+            // Default policy (ERROR) should fail
```
ScottDugas, line 137: It's probably worth calling with explicitly specifying `ERROR`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+            recordStore.deleteRecord(simpleRecord.getPrimaryKey());
+            recordStore.deleteRecord(otherRecord.getPrimaryKey());
+
+            context.commit();
+        }
+
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, baseMetaData(addJoinedType()));
+            // Default policy (ERROR) should fail
```
ScottDugas, line 182: It's probably worth calling with explicitly specifying `ERROR`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ScottDugas: On the other hand, you could probably combine these two into a parameterized test with two booleans
`loadSyntheticRecordFailsMissingConstituent(boolean deleteSimple, boolean deleteOther)`
and then run with all the combinations, and make sure that if either is deleted you end up with the appropriate behaviors.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
```diff
@@ -1097,6 +1098,72 @@ void loadDoubleNestedType() {
...
+    @ParameterizedTest(name = "loadMapType[{index}]")
+    @MethodSource("mapMetaDataSuppliers")
+    void loadRecordNotFound(Function<RecordMetaDataHook, RecordMetaData> metaDataSource) {
+        final RecordMetaData metaData = metaDataSource.apply(addMapType());
+        final RecordType unnestedType = metaData.getSyntheticRecordType(UNNESTED_MAP);
+
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, metaData);
+
+            for (TestRecordsNestedMapProto.OuterRecord outerRecord : sampleMapRecords()) {
```
ScottDugas, line 1116: I don't think there's any reason to iterate over `sampleMapRecords()` here, none of them would save with a primary key of 100.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
```diff
@@ -44,6 +44,29 @@
...
+ * There is nothing special needed to be done to save a joined record. Once the constituent records are saved
+ * a joined record instance is implicitly assumed to exist. A subsequent {@link #loadByPrimaryKeyAsync(FDBRecordStore, Tuple, IndexOrphanBehavior)}
+ * or a query for the record or a scan of a joined index will create the joined record. similarly, a deletion
+ * of any or all of the constituents will implicitly render the joined record type deleted.
+ * <p>
+ * When loading a joined record (and similarly when scanning an index) an {@link IndexOrphanBehavior} can be used
+ * to determine the behavior in case one or more of the constituents are missing:
+ * <ul>
+ *     <li>{@link IndexOrphanBehavior#ERROR} (the default) will throw an exception</li>
+ *     <li>{@link IndexOrphanBehavior#RETURN} will return an instance of the records with no constituents</li>
```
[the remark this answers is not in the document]
ohadzeliger: For this PR, I would have thought that in the case of an outer joined constituent we would be allowed to _not_ error out if one of the constituents is null. I don't know whether this sits well with the rest of the assumptions of missing constituents. Need to discuss further.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ohadzeliger: Sure, done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ohadzeliger: Yes, done - one method for positive test and one for failing test

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/UnnestedRecordTypeTest.java
ohadzeliger: Yes, of course. That was unnecessary.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/JoinedRecordType.java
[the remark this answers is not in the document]
ohadzeliger: Created a ticket to track status of this feature.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
```diff
@@ -103,52 +108,13 @@ void loadSyntheticRecord() throws ExecutionException, InterruptedException {
...
-            assertEquals(RecordDoesNotExistException.class, exception.getCause().getClass());
-            // RETURN policy returns the shell of the synthetic record with no constituents
-            final FDBSyntheticRecord syntheticRecord = recordStore.loadSyntheticRecord(joinedPrimaryKey, IndexOrphanBehavior.RETURN).get();
-            assertEquals(0, syntheticRecord.getConstituents().size());
-            // SKIP policy returns null in case of missing constituents
-            assertNull(recordStore.loadSyntheticRecord(joinedPrimaryKey, IndexOrphanBehavior.SKIP).get());
-
-            context.commit();
-        }
+    static Stream<Arguments> twoBooleans() {
```
ScottDugas, line 111: `twoBooleans` is a bit vague. Outside of the implementation, I would assume it would have all 4, but it's skipping `false, false`, because that's tested above. Perhaps renaming this to just the name of the test, making it tightly coupled to the test itself. Or you could use `@CsvSource`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ScottDugas: Sorry, I meant, in addition to the default.

ScottDugas (changes requested):
Just a couple small test cleanup commens

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/JoinedRecordTypeTest.java
ohadzeliger: Sure, changed to CsvSource

PR #3275, opened 2025-03-26
Have create_release_notes.py get the old version from ReleaseNotes.md

ScottDugas:
This makes it dependent on ReleaseNotes.md, which makes sense. In addition, it now can only add one release at a time, which seems fine.

I also added some tests, because I wanted to make sure it works without having to fetch commit history and whatnot.
I still had to do some validation by running the command line, and found some bugs that way, so more tests could be added, but this felt good for right now.

I also, obviously, did not add any integration into prb for the tests, but that may be nice if we right more, they currently take less than a second.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,7 +165,7 @@ jobs:
           --repository ${{ github.repository }} \
           --commit \
           --mixed-mode-results mixed-mode-results.md \
-          ${{ steps.get_old_version.outputs.version }} ${{ steps.get_new_version.outputs.version }}
+          --version ${{ steps.get_new_version.outputs.version }}
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -165,7 +165,7 @@ jobs:
           --repository ${{ github.repository }} \
           --commit \
           --mixed-mode-results mixed-mode-results.md \
-          ${{ steps.get_old_version.outputs.version }} ${{ steps.get_new_version.outputs.version }}
```
alecgrieser, line 168: I think we can get rid of the `get_old_version` step entirely, as it doesn't seem to be used any more.

alecgrieser (approved):
LGTM

.github/workflows/release.yml
ScottDugas: Yeah, that sounds right, I'll do that as a followup so that we can get this merged, and can do releases again.

PR #3276, opened 2025-03-26
Fix wasNull state across toString() calls in RowStruct

g31pranjal:
This PR fixes the bug related to `wasNull` state not being preserved across `toString()` calls in a `RowStruct`. For instance, 

If a struct is defined as `{a -> null, b -> 1}`, `getObject(1)` should set the `wasNull` to `true` which should be uneffected if a call to `toString()` is made. 

Also adds tests around the same.


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/RowStruct.java
```diff
@@ -333,7 +336,8 @@ public String toString() {
         } catch (SQLException e) {
             throw new UncheckedRelationalException(new RelationalException(e));
         }
-
+        // restore wasNull
+        wasNull = preservedWasNullValue;
```
arnaud-lacurie, line 340: Could you restore wasNull in a finally block? 
To make sure that wasNull is not wrong if there is an exception being thrown.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/RowStructTest.java
```diff
@@ -0,0 +1,69 @@
...
+        struct.getObject(2);
+        Assertions.assertTrue(struct.wasNull());
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void wasNullWorksWithToString(boolean mutable) throws SQLException {
+        final var struct = createStruct(mutable);
+        struct.getObject(2);
+        Assertions.assertFalse(struct.toString().isEmpty());
```
arnaud-lacurie, line 51: Wouldn't this succeed even before your PR? Since it would override wasNull to false when serializing the first element but would set it back to true when serializing the second element.

arnaud-lacurie (changes requested):
Could you check the other classes that have a wasNull and make sure that this bug is not present there?

g31pranjal:
`wasNull` is defined in the `Struct` and `ResultSet` interfaces. Since `ResultSet` treats its rows as a `Struct` they (mostly, in Embedded) call the underlying `struct.wasNull()`. JDBC implementation has its own mechanism but does not have a `toString()`. 

Implementations of `Struct`, there is only one, which is fixed.

arnaud-lacurie (approved):

Issue #3231, opened 2025-03-05
`INSERT` with named columns errors with nested structs and arrays


hatyo:
Our SQL supports setting only a specific set of columns of a (nested) record, provided this is not prohibited e.g. due to causing constraint violation such as missing (parts of) the primary key constituents, i.e. this works as expected:

```sql
create type as struct S(S1 bigint, S2 bigint)
create table B(B1 bigint, B2 bigint, B3 S, primary key(B1))
insert into B (B1, B3(S2)) values (20, (5))
select * from B -- returns (20, null, (null, 5))
```

However, it is not possible to do this when we have nested arrays and structs:

```sql
create type as struct R(R1 bigint, R2 bigint)
create type as struct S(S1 bigint, S2 bigint)
create type as struct T(T1 R, T2 bigint)
create type as struct U(U1 T, U2 S array)
create table C(A1 bigint, A2 S array, primary key (A1))
create table D(A1 bigint, A2 U array, primary key (A1))
insert into C(A1, A2(S1)) values (1, [(10), (30)]), (2, [(100)]) -- error
insert into D(A1, A2(U2(S1))) values (1, [[(100), (200)]]) -- error
```

As it throws the following error:

```
Caused by: com.apple.foundationdb.relational.api.exceptions.ContextualSQLException: expected Record but got Array
	at com.apple.foundationdb.relational.api.exceptions.RelationalException.toSqlException(RelationalException.java:74)
	at com.apple.foundationdb.relational.recordlayer.EmbeddedRelationalPreparedStatement.execute(EmbeddedRelationalPreparedStatement.java:80)
	at com.apple.foundationdb.relational.yamltests.command.QueryExecutor.executeStatement(QueryExecutor.java:294)
```

Arguably, the above syntax can be ambiguous, or at least, it does not align exactly with the row we're trying to set. Perhaps we need to change the aliasing syntax to reflect the fact that `A2` is actually an array, but I am not sure.

```sql
insert into C(A1, A2[(S1)]) values (1, [(10), (30)]), (2, [(100)]) -- indicate A2 is an array
```

PR #3280, opened 2025-03-28
Support SQL Table Valued Functions

### LINKED ISSUE #3282: Support Table-Valued Functions
The relational engine should provide support to table-valued-functions, this ticket is about introducing the necessary framework support across different components (parser, SA, function catalog, resolution, planning, and execution) to table-valued functions.

A working example of these TVFs would be `values` function, which is implemented by almost all major SQL vendors:

```sql
select * from values (1, 'a', 3.0), (4, 'b', 4.5);
-- returns two rows:
-- 1, 'a', 3.0
-- 4, 'b', 4.5
```

Another example is the `range` TVF, which returns a range between begin (inclusive) and end (exclusive) with an optional step size:

```sql
select ID from range (1, 5)
-- returns 4 rows
-- 1
-- 2
-- 3
-- 4
```

The implementation of these TVFs should leverage streaming semantics where applicable. For example, the `range` function above should not pre-materialize all the values in the range, but rather return them on-demand.

hatyo:
This provides support to table-valued-functions, it introduces the necessary framework support across different components: parsing, semantic analysis, plan generation, planning, and execution, to pre-defined table-valued functions.

Two pre-defined table-functions are introduced in this PR:

#### 1. `values` function:
Most major SQL vendors support this function, the relational engine now supports this function as well:
```sql
select * from values (1, 'a', 3.0), (4, 'b', 4.5);
-- returns two rows:
-- 1, 'a', 3.0
-- 4, 'b', 4.5
```
It also enables naming the resulting table along with its columns:
```sql
select * from values (1, 2.0, (3, 4, 'foo')), (10, 90.2, (5, 6.0, 'bar')) as A(B, C, W(X, Y, Z))
-- returns two rows with the following metadata:
-- B  | C    | W(X, Y, Z)
-- 1  | 2.0  | (3, 4.0, 'foo')
-- 10 | 90.2 | (5, 6.0, 'bar')
```
Notice how the value `4` in the first row's nested structure `W.Y` was promoted to `4.0`, this is because this function analyzes all the rows, and for each leaf in every (nested) tuple, it looks for a type that accommodates for all the individual values of the leaf across all rows (max type) and injects type promotions if necessary.
Also note that the type aliasing works with nested array and structs, to some degrees addresses some of the issues reported in #3231, and in the future, the new logic can be leveraged to resolve that ticket as well.

#### 2. `range` function:
This is also a very common function, it returns a range between:
- given start (inclusive), defaulting to `0`.
- given end (exclusive).
- optional step size.
- 
```sql
select ID from range (1, 5)
-- returns 4 rows
-- 1
-- 2
-- 3
-- 4
```
The implementation of this TVF has streaming semantics. In other words, it does not pre-materialize all the values in the range, but rather return them on-demand.

With this PR, we can hopefully start adding more advanced TVFs easily.

This resolves #3282.


normen662 (changes requested):
General comment:
I am not sure if this changes the path of `INSERT INTO T VALUES (...)`. What is that QGM if it is changed. What is the QGM for `SELECT * FROM VALUES(...)`? is it supposed to use an `ExplodeValue`?

yaml-tests/src/test/resources/table-functions.yamsql
```diff
@@ -0,0 +1,131 @@
...
+    -
+      - query: select * from range(0, 12, 5)
+      - result: [{ID: 0}, {ID: 5}, {ID: 10}]
+    -
+      - query: select * from range(0, 11, 5)
+      - explain: "TFunc range()"
+      - result: [{ID: 0}, {ID: 5}, {ID: 10}]
+    -
+      - query: select * from range(6 - 6, 14 + 6 + 1, 20 - 10)
+      - explain: "TFunc range()"
```
normen662, line 101: `range()` needs to explain its arguments

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/explain/ExplainPlanVisitor.java
```diff
@@ -414,6 +415,13 @@ public ExplainTokens visitInsertPlan(@Nonnull final RecordQueryInsertPlan insert
                 .addIdentifier(insertPlan.getTargetRecordType());
     }
 
+    @Nonnull
+    @Override
+    public ExplainTokens visitTableFunctionPlan(@Nonnull final RecordQueryTableFunctionPlan tableFunctionPlan) {
+        return addKeyword("TFunc").addWhitespace()
```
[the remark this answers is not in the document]
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DistinctRecordsProperty.java
```diff
@@ -214,6 +215,12 @@ public Boolean visitInsertPlan(@Nonnull final RecordQueryInsertPlan insertPlan)
             return distinctRecordsFromSingleChild(insertPlan);
         }
 
+        @Nonnull
+        @Override
+        public Boolean visitTableFunctionPlan(@Nonnull final RecordQueryTableFunctionPlan element) {
+            return  false;
```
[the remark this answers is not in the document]
hatyo: I will work on a separate PR to enable this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
```diff
@@ -559,6 +567,12 @@ public Cardinalities visitLogicalIntersectionExpression(@Nonnull final LogicalIn
         return intersectCardinalities(fromChildren(logicalIntersectionExpression));
     }
 
+    @Nonnull
+    @Override
+    public Cardinalities visitTableFunctionExpression(@Nonnull final TableFunctionExpression element) {
+        return Cardinalities.unknownMaxCardinality();
```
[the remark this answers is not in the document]
hatyo: I will work on a separate PR to enable this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/OrderingProperty.java
```diff
@@ -302,6 +303,12 @@ public Ordering visitInsertPlan(@Nonnull final RecordQueryInsertPlan insertPlan)
             return Ordering.empty();
         }
 
+        @Nonnull
+        @Override
+        public Ordering visitTableFunctionPlan(@Nonnull final RecordQueryTableFunctionPlan element) {
+            return Ordering.empty();
```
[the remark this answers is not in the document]
hatyo: I will work on a separate PR to enable this.

fdb-relational-core/src/main/antlr/RelationalLexer.g4
```diff
@@ -166,7 +166,7 @@ PARTITION:                           'PARTITION';
 PRIMARY:                             'PRIMARY';
 PROCEDURE:                           'PROCEDURE';
 PURGE:                               'PURGE';
-RANGE:                               'RANGE';
+// RANGE:                               'RANGE';
```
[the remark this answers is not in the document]
hatyo: Done.

yaml-tests/src/test/resources/table-functions.yamsql
hatyo: I had a todo there, but fixed it in the PR.

hatyo:
> General comment: I am not sure if this changes the path of `INSERT INTO T VALUES (...)`. What is that QGM if it is changed. What is the QGM for `SELECT * FROM VALUES(...)`? is it supposed to use an `ExplodeValue`?

Yes, the `ExplodeValue` remains to be used in:
- `INSERT` from `VALUES` codepath (it is almost exactly as it was before this PR).
- `SELECT` from `VALUES` codepath.

I decided to keep it that way anticipating that a switch to using a more general `TableFunctionValue` et al will cause hundreds of plans to be changed, dramatically increasing the scope of this PR. I want to split this legwork across (multiple) PRs if possible.

The QGM of `SELECT FROM VALUES(...)` is the following:

image

Ending up with a single physical `VALUES` box.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
```diff
@@ -336,6 +338,15 @@ public Derivations visitInsertPlan(@Nonnull final RecordQueryInsertPlan insertPl
             return new Derivations(resultValuesBuilder.build(), localValuesBuilder.build());
         }
 
+        @Nonnull
+        @Override
+        public Derivations visitTableFunctionPlan(@Nonnull final RecordQueryTableFunctionPlan tableFunctionPlan) {
+            final var streamingValue = tableFunctionPlan.getValue();
+            final var elementType = streamingValue.getResultType();
+            final var values = ImmutableList.<Value>of(new FirstOrDefaultStreamingValue(streamingValue, new ThrowsValue(elementType)));
```
normen662, line 346: Since these values are fake anyway, when we spoke I thought we had agreed on using one of these for both arrays and the new stream value.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementTableFunctionRule.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final BindingMatcher<TableFunctionExpression> root =
+            tableFunctionExpression();
+
+    public ImplementTableFunctionRule() {
+        super(root);
+    }
+
+    @Override
+    public void onMatch(@Nonnull final CascadesRuleCall call) {
+        final var explodeExpression = call.get(root);
```
normen662, line 49: copy pasta

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementTableFunctionRule.java
```diff
@@ -0,0 +1,52 @@
...
+
+import javax.annotation.Nonnull;
+
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.RelationalExpressionMatchers.tableFunctionExpression;
+
+/**
+ * A rule that implements an table function expression into a {@link RecordQueryTableFunctionPlan}.
+ */
+@API(API.Status.EXPERIMENTAL)
+@SuppressWarnings("PMD.TooManyStaticImports")
```
normen662, line 38: Can you try removing that?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FirstOrDefaultStreamingValue.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A value that returns the first element of the streaming {@link Value} that is passed in. If the streaming value returns
+ * and empty stream or the first item is null, a default value of the same type is returned.
+ *
+ * @see RecordCursor#first() semantics for more information.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class FirstOrDefaultStreamingValue extends AbstractValue {
```
normen662, line 59: I thought we were going to use `FirstOrDefaultValue`. By introducing this you must also change `CompatibleTypeEvolutionPredicate` to reflect this change. Or you find a way to have one `FirstOrDefaultValue`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/FirstOrDefaultStreamingValue.java
```diff
@@ -0,0 +1,187 @@
+/*
+ * FirstOrDefaultValue.java
```
normen662, line 2:
```suggestion
 * FirstOrDefaultStreamingValue.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -0,0 +1,442 @@
...
+ * A {@link StreamingValue} that is able to return a range that is defined as the following:
+ * <ul>
+ *     <li>an optional inclusive start (0L by default).</li>
+ *     <li>an exclusive end.</li>
+ *     <li>an optional step (1L by default).</li>
+ * </ul>
+ * For more information, see relational SQL {@code range} table-valued function.
+ */
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+public class RangeValue extends AbstractValue implements StreamingValue, CreatesDynamicTypesValue {
```
normen662, line 77: This should not need `CreatesDynamicTypesValue` as you are using a regular RCV under the cover.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -0,0 +1,442 @@
...
+    private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Range-Value");
+
+    @Nonnull
+    private final Value endExclusive;
+
+    @Nonnull
+    private final Optional<Value> beginInclusive;
```
normen662, line 84: Can this be `LiteralValue(0L)` if not given? Seems a lot of code would get simpler. I would even argue the encapsulator should do this. That's what it's there for.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -0,0 +1,442 @@
...
+
+    @Nonnull
+    private final Optional<Value> step;
```
normen662, line 87: Same here. Let's use `LiteralValue(1L)` if not given in the encapsulator.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public <M extends Message> RecordCursor<QueryResult> evalAsStream(@Nonnull final FDBRecordStoreBase<M> store,
+                                                                      @Nonnull final EvaluationContext context,
+                                                                      @Nullable final byte[] continuation,
+                                                                      @Nonnull final ExecuteProperties executeProperties) {
+        final long endExclusiveValue = (Long)Verify.verifyNotNull(endExclusive.eval(store, context));
+        final var beginInclusiveMaybe = beginInclusive.map(b -> (Long)Verify.verifyNotNull(b.eval(store, context)));
```
normen662, line 113: In the planner I adopted for methods that return `Optional<>` to be named `...Maybe`, variables that are `Optional` are called `...Optional`. I think I picked that up from some framework in Scala. It's debatable if we want that, but we should not deviate from the rule unless we change it everywhere. We the other suggested change this should go away in any case (I hope)...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/StreamingValue.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * a {@link Value} that returns a stream of results upon evaluation.
+ */
+public interface StreamingValue extends Value {
+
+    @Nonnull
+    <M extends Message> RecordCursor<QueryResult> evalAsStream(@Nonnull FDBRecordStoreBase<M> store,
```
normen662, line 39: Can you add documentation?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryExplodePlan.java
```diff
@@ -86,6 +87,10 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nonnull final EvaluationContext context,
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
+        if (collectionValue instanceof StreamingValue) {
```
normen662, line 90: I think this should be removed?!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryTableFunctionPlan.java
```diff
@@ -0,0 +1,280 @@
...
+                                         @Nonnull final AliasMap equivalencesMap) {
+        if (this == otherExpression) {
+            return true;
+        }
+        if (getClass() != otherExpression.getClass()) {
+            return false;
+        }
+        final var otherTableFunctionPlan =  (RecordQueryTableFunctionPlan)otherExpression;
+
+        return value.semanticEquals(otherTableFunctionPlan.value, equivalencesMap) &&
```
normen662, line 184: I think this is doing too much work (by being redundant).

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -131,4 +131,8 @@ message RecursiveCursorContinuation {
     optional bool isInitialState = 1;
     optional planprotos.PTempTable tempTable = 2;
     optional bytes activeStateContinuation = 3;
+}
+
+message RangeCursorContinuation {
+    optional int64 nextPosition = 2;
```
normen662, line 137:
```suggestion
    optional int64 nextPosition = 1;
```

fdb-record-layer-core/src/main/proto/record_query_plan.proto
```diff
@@ -460,6 +461,11 @@ message PFirstOrDefaultValue {
   optional PValue on_empty_result_value = 2;
 }
 
+message PFirstOrDefaultStreamingValue {
```
normen662, line 464: Three options:
- `FirstOrDefaultStreamingValue` goes away with the suggested changes elsewhere,
- `FirstOrDefaultStreamingValue` properly implements plan serialization together with putting `PFirstOrDefaultStreamingValue` into `PValue`
- `FirstOrDefaultStreamingValue` does not implement plan serialization  at all (but I think the constraint on the types will then fail to serialize, so properly two options only).

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/exceptions/ErrorCode.java
```diff
@@ -122,6 +122,7 @@ public enum ErrorCode {
     INVALID_TABLE_DEFINITION("42F16"),
     UNKNOWN_TYPE("42F18"),
     INVALID_RECURSION("42F19"),
+    INCOMPATIBLE_TABLE_ALIAS("42F20"),
```
normen662, line 125: Hard to search the PR for usages of this new code. But where is it used?

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -258,4 +258,9 @@ public void enumTest(YamlTest.Runner runner) throws Exception {
     public void uuidTest(YamlTest.Runner runner) throws Exception {
         runner.runYamsql("uuid.yamsql");
     }
+
+    @TestTemplate
+    public void tableFunctionsTest(YamlTest.Runner runner) throws Exception {
```
normen662, line 263: I am not really sure how this can pass with the plan serialization problem I think exists due to the `FirstOrDefaultStreamingValue` problem.

yaml-tests/src/test/resources/table-functions.metrics.yaml
```diff
@@ -0,0 +1,134 @@
...
+    task_count: 100
+    task_total_time_ms: 7
+    transform_count: 33
+    transform_time_ms: 2
+    transform_yield_count: 4
+    insert_time_ms: 0
+    insert_new_count: 6
+    insert_reused_count: 0
+-   query: EXPLAIN select * from range(0, 11, 5)
+    explain: TFunc range()
```
normen662, line 83: The explain still shows the previous version without the arguments. Since we only check on the metrics here that won't break the test run. It would still be better to regenerate the metrics files to reflect that latest change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryTableFunctionPlan.java
```diff
@@ -0,0 +1,280 @@
...
+    public PRecordQueryPlan toRecordQueryPlanProto(@Nonnull final PlanSerializationContext serializationContext) {
+        return PRecordQueryPlan.newBuilder().setTableFunctionPlan(toProto(serializationContext)).build();
+    }
+
+    @Nonnull
+    public static RecordQueryTableFunctionPlan fromProto(@Nonnull final PlanSerializationContext serializationContext,
+                                                         @Nonnull final PRecordQueryTableFunctionPlan recordQueryTableFunctionPlanProto) {
+        final var value = Value.fromValueProto(serializationContext,
+                Objects.requireNonNull(recordQueryTableFunctionPlanProto.getValue()));
+        if (!(value instanceof StreamingValue)) {
```
normen662, line 255: This is guaranteed by the class definition to be a `StreamingValue`. In the very remote case that we change that without afterthought I think a lot of things would break anyway so why not just cast it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DistinctRecordsProperty.java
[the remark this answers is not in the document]
normen662: https://github.com/FoundationDB/fdb-record-layer/issues/3291

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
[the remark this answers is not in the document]
normen662: https://github.com/FoundationDB/fdb-record-layer/issues/3291

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/OrderingProperty.java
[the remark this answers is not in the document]
normen662: https://github.com/FoundationDB/fdb-record-layer/issues/3291

normen662:
Forgot one more thing: can you please update `ReleaseNotes.md`? Thanks!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/TableFunctionExpression.java
```diff
@@ -0,0 +1,168 @@
...
+        if (this == otherExpression) {
+            return true;
+        }
+        if (!(otherExpression instanceof TableFunctionExpression)) {
+            return false;
+        }
+
+        final var otherTableFunctionExpression = (TableFunctionExpression)otherExpression;
+
+        return value.semanticEquals(otherTableFunctionExpression.getValue(), equivalencesMap) &&
```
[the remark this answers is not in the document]
hatyo: fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
hatyo: Yes, but this can only happen once `StreamingValue` is able to handle both arrays (explode) and streamed values (e.g. range). Then we can use `FirstOrDefaultStreamingValue` that expects an underlying `StreamingValue`. please let me know if this is aligned with your thinking.

By the way, I think there is a different in the empty semantics between `FirstOrDefaultStreamingValue` and `FirstOrDefaultValue`. We can talk about it offline if you want.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
hatyo: No optionals anymore.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryTableFunctionPlan.java
hatyo: It is extra defensive, considering the `proto` message is something sent from an external client.

yaml-tests/src/test/java/YamlIntegrationTests.java
hatyo: what's the plan serialization problem?
the `FirstOrDefaultStreamingValue` is not even serialized, it is only used in the `DerivationProperty` to represent the semantics of the derivation.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/exceptions/ErrorCode.java
hatyo: `TypeUtils.java`.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
hatyo: `PFirstOrDefaultStreamingValue` already implements plan serialization, I forgot adding it as a possibility in `PValue`. The code works because it is only used currently in `DerivationProperty`.
I just added it to `PValue`.

hatyo:
> Forgot one more thing: can you please update `ReleaseNotes.md`? Thanks!

I don't think we should do that manually anymore. There should be a CI job that maintains this file.

fdb-record-layer-core/src/main/proto/record_query_plan.proto
normen662: Right, but the derivation is then put into a constraint that is then put into the continuation. I am not sure why this worked.

PR #3298, opened 2025-04-05
Index scrubbing should support range id & range reset

### LINKED ISSUE #3297: Index scrubbing should support range id & range reset
To allow users run different types of index scrubbing sessions, we should add the next items to the scrubbing policy: 
- range reset: This will force the scrubber to start from scratch (default is false)
- range id: user can request to use another range set, not to disturb existing one (default, backward compatible, is range 0)

jjezra:
Resolve #3297
   To allow users run different types of index scrubbing sessions, we should add the next items to the scrubbing policy:

   - scrubbing range reset: This will force the scrubber to start from scratch (default is false)
   - scrubbing range id: user can request to use another range set, not to disturb existing one (default, backward compatible, is range 0)
   - new scrubbing API function `eraseAllIndexingScrubbingData` can be used to delete all scrubbing data. This can be used as a scrubbing state cleanup.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -4710,8 +4710,8 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         IndexingRangeSet.forIndexBuild(this, index).clear();
         // clear even if non-unique in case the index was previously unique
         context.clear(indexUniquenessViolationsSubspace(index).range());
-        // Under the index build subspace, there are multiple lower level subspaces - the lock space and few others.
-        // We are not supposed to clear the lock subspace, which might have used to an online index job that had invoked this method.
+        // Under the index build subspace, there are multiple lower level subspaces - the lock subspace and few others. We are
+        // not supposed to clear the lock subspace, which might have been used to an online index job that had invoked this method.
         IndexingSubspaces.eraseAllIndexingDataButTheLock(context, this, index);
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -3916,7 +3916,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     @API(API.Status.INTERNAL)
     @Nonnull
     public CompletableFuture<IndexBuildProto.IndexBuildIndexingStamp> loadIndexingTypeStampAsync(Index index) {
-        byte[] stampKey = OnlineIndexer.indexBuildTypeSubspace(this, index).pack();
+        byte[] stampKey = IndexingSubspaces.indexBuildTypeSubspace(this, index).pack();
         return ensureContextActive().get(stampKey).thenApply(serializedStamp -> {
             if (serializedStamp == null) {
                 return null;
@@ -3944,7 +3944,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
      */
     @API(API.Status.INTERNAL)
     public void saveIndexingTypeStamp(Index index, IndexBuildProto.IndexBuildIndexingStamp stamp) {
-        byte[] stampKey = OnlineIndexer.indexBuildTypeSubspace(this, index).pack();
+        byte[] stampKey = IndexingSubspaces.indexBuildTypeSubspace(this, index).pack();
         ensureContextActive().set(stampKey, stamp.toByteArray());
     }
 
@@ -4710,14 +4710,9 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
         IndexingRangeSet.forIndexBuild(this, index).clear();
         // clear even if non-unique in case the index was previously unique
         context.clear(indexUniquenessViolationsSubspace(index).range());
-        // Under the index build subspace, there are 3 lower level subspaces, the lock space, the scanned records
-        // subspace, and the type/stamp subspace. We are not supposed to clear the lock subspace, which is used to
-        // run online index jobs which may invoke this method. We should clear:
-        // * the scanned records subspace. Which, roughly speaking, counts how many records of this store are covered in
-        // index range subspace.
-        // * the type/stamp subspace. Which indicates which type of indexing is in progress.
-        context.clear(Range.startsWith(OnlineIndexer.indexBuildScannedRecordsSubspace(this, index).pack()));
-        context.clear(Range.startsWith(OnlineIndexer.indexBuildTypeSubspace(this, index).pack()));
+        // Under the index build subspace, there are multiple lower level subspaces - the lock space and few others.
+        // We are not supposed to clear the lock subspace, which might have used to an online index job that had invoked this method.
+        IndexingSubspaces.eraseAllIndexingDataButTheLock(context, this, index);
     }
 
     @SuppressWarnings("PMD.CloseResource")
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
@@ -161,8 +161,8 @@ public class IndexScrubbing extends IndexingBase {
             if (range == null) {
                 // Here: no more missing ranges - all done
                 // This scrubbing is done. Clear the rangeSet - the next time scrubbing is called it will start from scratch
-                rangeSet.clear();
                 logScrubberRangeReset("range exhausted");
+                rangeSet.clear();
                 return AsyncUtil.READY_FALSE;
             }
             final Tuple rangeStart = RangeSet.isFirstKey(range.begin) ? null : Tuple.fromBytes(range.begin);
@@ -245,25 +245,23 @@ public class IndexScrubbing extends IndexingBase {
     @SuppressWarnings("PMD.CloseResource")
     @Override
     protected CompletableFuture<Void> setScrubberTypeOrThrow(FDBRecordStore store) {
-        // HERE: The index must be readable, checked by the caller
-        //   if scrubber had already run and still have missing ranges, do nothing
-        //   else: clear ranges and overwrite type-stamp
+        // HERE: The index must be readable, checked by the caller.
         IndexBuildProto.IndexBuildIndexingStamp indexingTypeStamp = getIndexingTypeStamp(store);
         validateOrThrowEx(indexingTypeStamp.getMethod().equals(IndexBuildProto.IndexBuildIndexingStamp.Method.SCRUB_REPAIR),
                 "Not a scrubber type-stamp");
 
-        final Index index = common.getIndex(); // Note: the scrubbers do not support multi target (yet)
+        final Index index = common.getIndex(); // Note: multi targets mode is not supported (yet)
         final IndexingRangeSet rangeSet = getRangeset(store, index);
         if (scrubbingPolicy.isRangeReset()) {
+            logScrubberRangeReset("forced reset");
             rangeSet.clear();
-            logScrubberRangeReset("force reset");
             return AsyncUtil.DONE;
         }
         return rangeSet.firstMissingRangeAsync()
                 .thenAccept(recordRange -> {
                     if (recordRange == null) {
-                        // Here: no un-scrubbed range was left for this call. We will
-                        // erase the 'ranges' data to allow a fresh records re-scrubbing.
+                        // Here: no un-scrubbed range is available for this call. Erase the 'ranges' data to allow
+                        // a new, fresh records re-scrubbing.
                         logScrubberRangeReset("range exhausted detected");
                         rangeSet.clear();
                     }
@@ -274,7 +272,7 @@ public class IndexScrubbing extends IndexingBase {
         if (LOGGER.isInfoEnabled()) {
             LOGGER.info(KeyValueLogMessage.build("Reset index scrubbing range")
                     .addKeysAndValues(common.indexLogMessageKeyValues())
-                            .addKeyAndValue(LogMessageKeys.REASON, reason)
+                    .addKeyAndValue(LogMessageKeys.REASON, reason)
                     .toString());
         }
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
@@ -162,6 +162,7 @@ public class IndexScrubbing extends IndexingBase {
                 // Here: no more missing ranges - all done
                 // This scrubbing is done. Clear the rangeSet - the next time scrubbing is called it will start from scratch
                 rangeSet.clear();
+                logScrubberRangeReset("range exhausted");
                 return AsyncUtil.READY_FALSE;
             }
             final Tuple rangeStart = RangeSet.isFirstKey(range.begin) ? null : Tuple.fromBytes(range.begin);
@@ -263,7 +264,7 @@ public class IndexScrubbing extends IndexingBase {
                     if (recordRange == null) {
                         // Here: no un-scrubbed range was left for this call. We will
                         // erase the 'ranges' data to allow a fresh records re-scrubbing.
-                        logScrubberRangeReset("range exhausted");
+                        logScrubberRangeReset("range exhausted detected");
                         rangeSet.clear();
                     }
                 });
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
@@ -30,6 +30,7 @@ import com.apple.foundationdb.record.RecordCursorResult;
 import com.apple.foundationdb.record.RecordMetaData;
 import com.apple.foundationdb.record.RecordMetaDataProvider;
 import com.apple.foundationdb.record.TupleRange;
+import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.MetaDataException;
@@ -86,6 +87,9 @@ public class IndexScrubbing extends IndexingBase {
         return Arrays.asList(
                 LogMessageKeys.INDEXING_METHOD, scrubberName,
                 LogMessageKeys.ALLOW_REPAIR, scrubbingPolicy.allowRepair(),
+                LogMessageKeys.RANGE_ID, scrubbingPolicy.getRangeId(),
+                LogMessageKeys.RANGE_RESET, scrubbingPolicy.isRangeReset(),
+                LogMessageKeys.SCRUB_TYPE, scrubbingType,
                 LogMessageKeys.SCAN_LIMIT, scrubbingPolicy.getEntriesScanLimit()
         );
     }
@@ -228,11 +232,49 @@ public class IndexScrubbing extends IndexingBase {
     IndexingRangeSet getRangeset(FDBRecordStore store, Index index) {
         switch (scrubbingType) {
             case MISSING:
-                return IndexingRangeSet.forScrubbingRecords(store, index);
+                return IndexingRangeSet.forScrubbingRecords(store, index, scrubbingPolicy.getRangeId());
             case DANGLING:
-                return IndexingRangeSet.forScrubbingIndex(store, index);
+                return IndexingRangeSet.forScrubbingIndex(store, index, scrubbingPolicy.getRangeId());
             default:
                 throw new RecordCoreArgumentException("Unpredicted scrubbing type ");
+        }
+    }
+
+    @Nonnull
+    @SuppressWarnings("PMD.CloseResource")
+    @Override
+    protected CompletableFuture<Void> setScrubberTypeOrThrow(FDBRecordStore store) {
+        // HERE: The index must be readable, checked by the caller
+        //   if scrubber had already run and still have missing ranges, do nothing
+        //   else: clear ranges and overwrite type-stamp
+        IndexBuildProto.IndexBuildIndexingStamp indexingTypeStamp = getIndexingTypeStamp(store);
+        validateOrThrowEx(indexingTypeStamp.getMethod().equals(IndexBuildProto.IndexBuildIndexingStamp.Method.SCRUB_REPAIR),
+                "Not a scrubber type-stamp");
+
+        final Index index = common.getIndex(); // Note: the scrubbers do not support multi target (yet)
+        final IndexingRangeSet rangeSet = getRangeset(store, index);
+        if (scrubbingPolicy.isRangeReset()) {
+            rangeSet.clear();
+            logScrubberRangeReset("force reset");
+            return AsyncUtil.DONE;
+        }
+        return rangeSet.firstMissingRangeAsync()
+                .thenAccept(recordRange -> {
+                    if (recordRange == null) {
+                        // Here: no un-scrubbed range was left for this call. We will
+                        // erase the 'ranges' data to allow a fresh records re-scrubbing.
+                        logScrubberRangeReset("range exhausted");
+                        rangeSet.clear();
+                    }
+                });
+    }
+
+    private void logScrubberRangeReset(String reason) {
+        if (LOGGER.isInfoEnabled()) {
+            LOGGER.info(KeyValueLogMessage.build("Reset index scrubbing range")
+                    .addKeysAndValues(common.indexLogMessageKeyValues())
+                            .addKeyAndValue(LogMessageKeys.REASON, reason)
+                    .toString());
         }
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -114,7 +114,8 @@ public abstract class IndexingBase {
 
 
     IndexingBase(@Nonnull IndexingCommon common,
-                 @Nonnull OnlineIndexer.IndexingPolicy policy, boolean isScrubber) {
+                 @Nonnull OnlineIndexer.IndexingPolicy policy,
+                 boolean isScrubber) {
         this.common = common;
         this.policy = policy;
         this.isScrubber = isScrubber;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -539,7 +539,7 @@ public abstract class IndexingBase {
     @Nonnull
     protected CompletableFuture<Void> setScrubberTypeOrThrow(FDBRecordStore store) {
         // This path should never be reached
-        throw new ValidationException("Called in a non-scrubbing path",
+        throw new ValidationException("Called setScrubberTypeOrThrow in a non-scrubbing path",
                 "isScrubber", isScrubber);
     }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -23,7 +23,6 @@ package com.apple.foundationdb.record.provider.foundationdb;
 import com.apple.foundationdb.FDBError;
 import com.apple.foundationdb.FDBException;
 import com.apple.foundationdb.MutationType;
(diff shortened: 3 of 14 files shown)


jjezra:
TODO: Add a test

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -576,44 +537,10 @@ private CompletableFuture<Void> throwUnlessNoRecordWasScanned(boolean noRecordSc
...
-                    if (LOGGER.isDebugEnabled()) {
-                        LOGGER.debug(KeyValueLogMessage.build("Reset scrubber's records range")
-                                .addKeysAndValues(common.indexLogMessageKeyValues())
-                                .toString());
-                    }
-                    recordsRangeSet.clear();
-                }
-            });
-        });
+    protected CompletableFuture<Void> setScrubberTypeOrThrow(FDBRecordStore store) {
```
ScottDugas, line 540: Can this be removed from this level, and only be defined in IndexScrubbing?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexScrubbing.java
```diff
@@ -228,14 +233,50 @@ private void reportIssues(List<IndexScrubbingTools.Issue> issueList, Throwable e
     IndexingRangeSet getRangeset(FDBRecordStore store, Index index) {
         switch (scrubbingType) {
             case MISSING:
-                return IndexingRangeSet.forScrubbingRecords(store, index);
+                return IndexingRangeSet.forScrubbingRecords(store, index, scrubbingPolicy.getRangeId());
             case DANGLING:
-                return IndexingRangeSet.forScrubbingIndex(store, index);
+                return IndexingRangeSet.forScrubbingIndex(store, index, scrubbingPolicy.getRangeId());
```
ScottDugas, line 238: Would it make sense to decouple the scrubbing type from the range id?
i.e. to have this method be something like:
```
if (scrubbingPolicy.getRangeId() == 0) {
    switch (scrubbingType) {
    case MISSING:
        return IndexingRangeSet.forScrubbingRecords(store, index);
        return IndexingRangeSet.forScrubbingRecords(store, index, scrubbingPolicy.getRangeId());
    case DANGLING:
        return IndexingRangeSet.forScrubbingIndex(store, index);
        return IndexingRangeSet.forScrubbingIndex(store, index, scrubbingPolicy.getRangeId());
    default:
        throw new RecordCoreArgumentException("Unpredicted scrubbing type ");
    }
} else {
    return IndexingRangeSet.forScrubbingId(store, scrubbingPolicy.getRange());
}
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingSubspaces.java
```diff
@@ -0,0 +1,100 @@
...
+import com.apple.foundationdb.record.metadata.Index;
+import com.apple.foundationdb.subspace.Subspace;
+import com.apple.foundationdb.tuple.Tuple;
+
+import javax.annotation.Nonnull;
+
+/**
+ * List of subspaces related to the indexing/index-scrubbing processes.
+ */
+public final class IndexingSubspaces {
```
ScottDugas, line 33: This should be annotated with `API.INTERNAL`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingSubspaces.java
```diff
@@ -0,0 +1,100 @@
...
+        throw new IllegalStateException("Utility class");
+    }
+
+    @Nonnull
+    private static Subspace indexBuildSubspace(@Nonnull FDBRecordStoreBase<?> store, @Nonnull Index index, Object key) {
+        return store.getUntypedRecordStore().indexBuildSubspace(index).subspace(Tuple.from(key));
+    }
+
+    @Nonnull
+    public static Subspace indexBuildLockSubspace(@Nonnull FDBRecordStoreBase<?> store, @Nonnull Index index) {
```
ScottDugas, line 52: I think some javadoc on the public methods here would be valuable.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4710,14 +4710,9 @@ void clearIndexData(@Nonnull Index index) {
         IndexingRangeSet.forIndexBuild(this, index).clear();
         // clear even if non-unique in case the index was previously unique
         context.clear(indexUniquenessViolationsSubspace(index).range());
-        // Under the index build subspace, there are 3 lower level subspaces, the lock space, the scanned records
```
ScottDugas, line 4713: This comment about why we aren't erasing the log seems valuable, and should be preserved, probably on the javadoc for `eraseAllIndexingDataButTheLock`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
ScottDugas: Looking at the usage, I think it would require a lot more untangling of the relationship between `IndexingBase`. The simplest approach I can see would be to introduce two additional abstract classes, an `IndexScrubbingBase` and `IndexBuildingBase`. If you see something trivial go for it, but otherwise, I guess it will stay as-is.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
```diff
@@ -576,44 +537,10 @@ private CompletableFuture<Void> throwUnlessNoRecordWasScanned(boolean noRecordSc
...
+        // This path should never be reached
+        throw new ValidationException("Called in a non-scrubbing path",
```
ScottDugas, line 542:
```suggestion
        throw new ValidationException("Called setScrubberTypeOrThrow in a non-scrubbing path",
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -126,21 +136,25 @@ public static Builder newBuilder() {
      * A builder for the scrubbing policy.
      */
     public static class ScrubbingPolicy {
-        public static final ScrubbingPolicy DEFAULT = new ScrubbingPolicy(1000, true, 0, false, false);
+        public static final ScrubbingPolicy DEFAULT = new ScrubbingPolicy(1000, false, 0, false, false, 0, false);
```
ScottDugas, line 139: You're changing the default policy to not repair.
This would be a breaking change, that I don't know it's worth making.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -126,21 +136,25 @@ public static Builder newBuilder() {
...
         private final int logWarningsLimit;
         private final boolean allowRepair;
         private final long entriesScanLimit;
         private final boolean ignoreIndexTypeCheck;
         private final boolean useLegacy;
+        private final int rangeId;
+        private final boolean rangeReset;
 
         public ScrubbingPolicy(int logWarningsLimit, boolean allowRepair, long entriesScanLimit,
```
ScottDugas, line 148: To avoid making a breaking change, it would be good ot have an override that does not take the new parameters, and provides default values.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -126,21 +136,25 @@ public static Builder newBuilder() {
...
-                               boolean ignoreIndexTypeCheck, boolean useLgacy) {
+                               boolean ignoreIndexTypeCheck, boolean useLgacy, int rangeId, boolean rangeReset) {
```
ScottDugas, line 149: Unrelated typo.
```suggestion
                               boolean ignoreIndexTypeCheck, boolean useLegacy, int rangeId, boolean rangeReset) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -263,8 +287,32 @@ public Builder useLegacyScrubber(boolean legacy) {
...
             }
 
+            /**
+             * Choose a specific "Already Scrubbed" rangeSet prefix. This may be useful for
+             * multiple scrubbing jobs that may not affect each other.
+             * 0 is the backward compatible default, which means no prefix.
+             * @param rangeId a rangeSet prefix
+             * @return this builder
+             */
+            public Builder setRangeId(final int rangeId) {
```
ScottDugas, line 297: I feel like `scrubbingId` would be clearer, and more use-focused, and less implementation-focused.

ScottDugas, line 297: It seems like it might be nice from a user perspective for this to be a string instead of an int.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
ScottDugas, line 148: Perhaps worth annotating this with `@Deprecated` and pointing users to the builder instead, and then we can make it private in the future.
Not needed in this PR, you could make a small followup PR, which would keep our release notes cleaner.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -263,8 +287,32 @@ public Builder useLegacyScrubber(boolean legacy) {
...
+                this.rangeId = rangeId;
+                return this;
+            }
+
+            /**
+             * If set to true, clear any previous "Already Scrubbed" range set (in the given rangeId) - which means that
+             * the index scrubbing will begin from scratch.
```
ScottDugas, line 304: I feel like the second part of this is much clearer, so perhaps this would be better:
```suggestion
             * Start the scrubbing from scratch, regardless of any progress already made for this scrubber id.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -263,8 +287,32 @@ public Builder useLegacyScrubber(boolean legacy) {
                 return this;
             }
 
+            /**
+             * Choose a specific "Already Scrubbed" rangeSet prefix. This may be useful for
+             * multiple scrubbing jobs that may not affect each other.
```
ScottDugas, line 292: I feel like referring to the range set is getting more into the implementation details than we want, perhaps this would be clearer:
```suggestion
             * Choose a specific id for this scrubbing operation. If the scrubbing stops, or fails, it can be resumed at a later time by constructing the same builder. Work done with other ids will not affect work done with this id.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexing/IndexingRangeSet.java
```diff
@@ -225,8 +227,15 @@ public static IndexingRangeSet forIndexBuild(@Nonnull FDBRecordStore store, @Non
      * @return a {@code WrappedRangeSet} for scrubbing the index's entries
      */
     @Nonnull
-    public static IndexingRangeSet forScrubbingIndex(@Nonnull FDBRecordStore store, @Nonnull Index index) {
-        RangeSet rangeSet = new RangeSet(IndexingBase.indexScrubIndexRangeSubspace(store, index));
+    public static IndexingRangeSet forScrubbingIndex(@Nonnull FDBRecordStore store, @Nonnull Index index, int rangeId) {
+        final Subspace subspace;
+        if (rangeId == 0) {
+            // Backward compatible
+            subspace = IndexingSubspaces.indexScrubIndexRangeSubspaceLegacy(store, index);
```
ScottDugas, line 234: This uses the word `legacy`, perhaps we should have a comment on the builder that the scrubbing id will be required in the future?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
```diff
@@ -117,6 +117,16 @@ public long scrubMissingIndexEntries() {
         return missingCount.get();
     }
 
+    /**
+     * Reset all index scrubbing data. This will make the index "forget" any previous index scrubbing
+     * sessions.
+     * This method was designed to be used as a global reset to clean unwanted partial index scrubbing information and
+     * should probably not be used as a routine.
+     */
+    public void eraseAllIndexingScrubbingData(@Nonnull FDBRecordContext context, @Nonnull FDBRecordStore store) {
```
ScottDugas, line 126: This should be called out in the PR description, as an additional part of the enhancements.

ScottDugas (commented):
This would have been better as 3 independent PRs, introducing 3 new scrubbing features, but I don't think it's worth untangling at this point.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                .setScrubbingPolicy(OnlineIndexScrubber.ScrubbingPolicy.newBuilder()
+                        .useLegacyScrubber(legacy)
+                        .setEntriesScanLimit(10)
+                        .setAllowRepair(true)
+                        .build())
+                .setLimit(10)
+                .build()) {
+            indexScrubber.scrubMissingIndexEntries();
+        }
+        assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
```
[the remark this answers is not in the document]
jjezra: Which library contains `Matchers.between`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                        .build())
+                .build()) {
+            res = indexScrubber.scrubMissingIndexEntries();
+        }
+        assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_MISSING_ENTRIES));
+        assertEquals(0, res);
+
+        // manually delete a few index entries
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+
+        Index tgtIndex = new Index("tgt_index", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS);
+        FDBRecordStoreTestBase.RecordMetaDataHook hook = myHook(tgtIndex);
+
+        populateData(numRecords);
+
+        openSimpleMetaData(hook);
+        buildIndexClean(tgtIndex);
+
+        try (OnlineIndexScrubber indexScrubber = newScrubberBuilder(tgtIndex, timer)
```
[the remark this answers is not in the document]
jjezra: I've capsulated all the full scans detect only to a common function. Leaving the others out as each has its own setup. Pleases let me know if that's enough.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                        .setRangeId(2)
+                        .setAllowRepair(false)
+                        .build())
+                .build()) {
+            indexScrubber.scrubMissingIndexEntries();
+        }
+        assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+
+        // Make sure range-id 0 was not affected
+        final ScrubbersMissingRanges range = getScrubbersMissingRange(tgtIndex);
```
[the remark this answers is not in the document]
jjezra: https://github.com/FoundationDB/fdb-record-layer/pull/2985#discussion_r1876827482

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                        .setEntriesScanLimit(10)
+                        .setAllowRepair(true)
+                        .build())
+                .setLimit(10)
+                .build()) {
+            indexScrubber.scrubMissingIndexEntries();
+        }
+        assertTrue(0 < timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+
+        // Run/Detect few records in range-id 1, no timer
```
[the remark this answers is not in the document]
jjezra: The timer is preserved, so later we can finish the original iteration with the timer and assert full scan.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                        .setEntriesScanLimit(21)
+                        .build())
+                .setLimit(20)
+                .build()) {
+            indexScrubber.scrubMissingIndexEntries();
+        }
+        final ScrubbersMissingRanges midRange = getScrubbersMissingRange(tgtIndex);
+        assertFalse(midRange.records.begin == null || RangeSet.isFinalKey(midRange.records.begin));
+
+        // finish fixing in range-id 0, with preserved timer. Check timer's values
```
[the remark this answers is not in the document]
jjezra: No, as the timer "continues" the original scan.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+        try (OnlineIndexScrubber indexScrubber = newScrubberBuilder(tgtIndex)
+                .setScrubbingPolicy(OnlineIndexScrubber.ScrubbingPolicy.newBuilder()
+                        .useLegacyScrubber(legacy)
+                        .setRangeId(1)
+                        .setAllowRepair(false)
+                        .setEntriesScanLimit(21)
+                        .build())
+                .setLimit(20)
+                .build()) {
+            indexScrubber.scrubMissingIndexEntries();
```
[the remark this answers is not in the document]
jjezra: Oops. 
✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
     }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testScrubberSimpleMissingMixedRanges(boolean legacy) throws ExecutionException, InterruptedException {
+        final FDBStoreTimer timer = new FDBStoreTimer();
+        final long numRecords = 50;
+        long res;
+
+        Index tgtIndex = new Index("tgt_index", field("num_value_2"), EmptyKeyExpression.EMPTY, IndexTypes.VALUE, IndexOptions.UNIQUE_OPTIONS);
```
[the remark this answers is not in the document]
jjezra: ✅

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -821,4 +828,620 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+                .build()) {
+            res = indexScrubber.scrubMissingIndexEntries();
+        }
+        assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
+        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_MISSING_ENTRIES));
+        assertEquals(0, res);
+
+        // manually delete a few index entries
+        openSimpleMetaData(hook);
```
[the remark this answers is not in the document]
jjezra: Nope. 
✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingSubspaces.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Fair enough. I think that we should change the default to be a scan only, but that should probably be done in another PR. 
✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: Mmm.. this constructor was never meant to be public. It also doesn't have any javadoc, test, or an API example. Would it make sense to convert it to private now?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: How about `scrubbingRangeId`? 
I am afraid that `scrubbingId` would be too general and misunderstanding.

jjezra: I've also elaborated the javadoc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: ✅

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: :whi

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/indexing/IndexingRangeSet.java
jjezra: Change to `IndexingSubspaces.indexScrubIndexRangeSubspaceZero(..)`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubber.java
jjezra: ✅

jjezra:
The style check believes that the "since" version in a deprecated tag is a hard coded IP.

ScottDugas (approved):

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -816,9 +713,462 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
+        assertFullIterationNoFix(tgtIndex, 77, numRecords, legacy, false, false);
+    }
+
+    private void assertFullIterationNoFix(Index index, int rangeId, long numRecords, boolean legacy, boolean missing, boolean expectIssues) {
+        FDBStoreTimer timer = new FDBStoreTimer();
+        long res;
+        try (OnlineIndexScrubber indexScrubber = newScrubberBuilder(index, timer)
+                .setScrubbingPolicy(OnlineIndexScrubber.ScrubbingPolicy.newBuilder()
+                        .useLegacyScrubber(legacy)
+                        .setAllowRepair(false)
```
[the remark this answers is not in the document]
jjezra: The allow repair? 
This function is not expected to repair.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -124,67 +93,22 @@ void testScrubberSimpleMissing(boolean legacy) throws ExecutionException, Interr
...
-                        .useLegacyScrubber(legacy)
-                        .build())
-                .build()) {
-            res = indexScrubber.scrubDanglingIndexEntries();
-        }
-        assertEquals(numRecords, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_SCANNED));
-        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.ONLINE_INDEX_BUILDER_RECORDS_INDEXED));
-        assertEquals(0, timer.getCount(FDBStoreTimer.Counts.INDEX_SCRUBBER_DANGLING_ENTRIES));
-        assertEquals(0, res);
+        Index tgtIndex = createValueIndexAndPopulateData(numRecords, false);
```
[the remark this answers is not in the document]
jjezra: Uniqueness should be meaningless for index scrubbing. I've alternated randomly.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/OnlineIndexScrubberTest.java
```diff
@@ -816,9 +713,462 @@ void testDetectMissingFromJoinedIndex(boolean legacy) {
...
-                assertThrows(UnsupportedOperationException.class, () -> indexScrubber.scrubMissingIndexEntries());
+                assertThrows(UnsupportedOperationException.class, indexScrubber::scrubDanglingIndexEntries);
+                assertThrows(UnsupportedOperationException.class, indexScrubber::scrubMissingIndexEntries);
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @BooleanSource
+    void testScrubberSimpleMissingMixedRanges(boolean legacy) throws ExecutionException, InterruptedException {
```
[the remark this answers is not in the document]
jjezra: Usually I use "Simple" to indicate a straight forward test. Typically the first few in a test file. Indeed, it doesn't fit here.

PR #3300, opened 2025-04-09
Support primitive type arrays contains

pengpeng-lu:
This fixes https://github.com/FoundationDB/fdb-record-layer/issues/3674

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizer.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizer.java
@@ -443,11 +443,11 @@ public final class AstNormalizer extends RelationalParserBaseVisitor<Object> {
 
     @Override
     public Object visitInPredicate(@Nonnull RelationalParser.InPredicateContext ctx) {
-        ctx.IN().accept(this);
-
         if (ctx.NOT() != null) {
             ctx.NOT().accept(this);
         }
+
+        ctx.IN().accept(this);
 
         if (ctx.inList().preparedStatementParameter() != null) {
             visit(ctx.inList().preparedStatementParameter());
--- a/fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizerTests.java
+++ b/fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizerTests.java
@@ -1328,6 +1328,24 @@ public class AstNormalizerTests {
                 Map.of(constantId(7), List.of(true, false)));
     }
 
+    @Test
+    void visitNotInPredicateWithConstants() throws Exception {
+        // Test NOT IN predicate with constants - verifies that NOT token is handled correctly
+        // This tests the visitInPredicate method's handling of NOT token in AstNormalizer
+        validate("select * from t1 where col1 not in (10, 20, 30)",
+                "select * from \"T1\" where \"COL1\" not in ( [ ] ) ",
+                Map.of(constantId(8), List.of(10, 20, 30)));
+    }
+
+    @Test
+    void visitNotInPredicateWithColumnReference() throws Exception {
+        // Test NOT IN predicate with column reference - this tests lines 453-454 in AstNormalizer
+        // where ctx.inList().fullColumnName() != null for a NOT IN predicate
+        validate("select * from T where 'apple' not in T.fruits",
+                "select * from \"T\" where ? not in \"T\" . \"FRUITS\" ",
+                Map.of(constantId(5), "apple"));
+    }
+
     @Nonnull
     private String normalizeQuery(@Nonnull final String functionDdl) throws RelationalException {
         final var normalizer = AstNormalizer.normalizeAst(fakeSchemaTemplate,
--- a/yaml-tests/src/test/resources/in-predicate.metrics.yaml
+++ b/yaml-tests/src/test/resources/in-predicate.metrics.yaml
@@ -1,24 +1,24 @@
-unnamed-3:
+unnamed-2:
 -   query: EXPLAIN select a, b from ta where b in (1, 1, 1, 1)
     explain: ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS
         ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }
     task_count: 991
-    task_total_time_ms: 277
+    task_total_time_ms: 116
     transform_count: 292
-    transform_time_ms: 143
+    transform_time_ms: 30
     transform_yield_count: 66
-    insert_time_ms: 10
+    insert_time_ms: 5
     insert_new_count: 124
     insert_reused_count: 8
 -   query: EXPLAIN select a, b from ta where b in (1, 2, 1, 3)
     explain: ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS
         ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }
     task_count: 991
-    task_total_time_ms: 277
+    task_total_time_ms: 116
     transform_count: 292
-    transform_time_ms: 143
+    transform_time_ms: 30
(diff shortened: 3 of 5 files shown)


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LogicalOperator.java
```diff
@@ -263,7 +263,15 @@ private static LogicalOperator generateCorrelatedFieldAccess(@Nonnull Expression
                 () -> String.format(Locale.ROOT, "join correlation can occur only on column of repeated type, not %s type", expression.getDataType()));
         final var explode = new ExplodeExpression(expression.getUnderlying());
         final var resultingQuantifier = Quantifier.forEach(Reference.initialOf(explode));
-        final var outputAttributes = Expressions.of(convertToExpressions(resultingQuantifier));
+
+        Expressions outputAttributes;
```
normen662, line 267: Why is this something we don't always do?

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
```diff
@@ -1362,6 +1361,69 @@ void unionIsNotSupported() throws Exception {
...
+                Assertions.assertTrue(statement.execute("SELECT * FROM T1 where exists (SELECT 1 FROM T1.A r where r in ('Grape', 'Orange'))"));
+                try (final RelationalResultSet resultSet = statement.getResultSet()) {
+                    ResultSetAssert.assertThat(resultSet).hasNextRow()
+                            .isRowExactly(42L, EmbeddedRelationalArray.newBuilder().addString("Apple").addString("Orange").build())
+                            .hasNextRow()
+                            .isRowExactly(44L, EmbeddedRelationalArray.newBuilder().addString("Grape").addString("Pear").build())
+                            .hasNoNextRow();
+                }
+
+                Assertions.assertTrue(statement.execute("SELECT * FROM T1 where 'Grape' in (a)"));
```
pengpeng-lu: I'd prefer to be able to remove the `()` here -> `select * from T1 where 'Grape' in a`, but keep hitting syntax ambuiguity errors.

.idea/misc.xml
hatyo, line 1: Can you please revert this?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizer.java
```diff
@@ -450,6 +450,8 @@ public Object visitInPredicate(@Nonnull RelationalParser.InPredicateContext ctx)
 
         if (ctx.inList().preparedStatementParameter() != null) {
             visit(ctx.inList().preparedStatementParameter());
+        } else if (ctx.inList().fullColumnName() != null) {
```
hatyo, line 454: Nice catch, can you please add a test for it?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LogicalOperator.java
hatyo: We _should_, but there is the assumption that `Quantifier`s flow `Record` after examining `getFlowedValues` and `getFlowedColumns`. That's why it looks like Pengpeng had to do the dispatch on primitive values here in order to support flowing (correlated) primitive fields such as in:
```sql
select * from T where exists (select b from T.A)
```
I think the fix should look different, considering that `Expressions` flow `Relation`s wrapping _any_ type, not exclusively `Record` types.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
```diff
@@ -516,6 +516,8 @@ public Expression visitInList(@Nonnull RelationalParser.InListContext ctx) {
         final Expression result;
         if (ctx.preparedStatementParameter() != null) {
             result = visitPreparedStatementParameter(ctx.preparedStatementParameter());
+        } else if (ctx.fullColumnName() != null) {
+            result = visitFullColumnName(ctx.fullColumnName());
```
hatyo, line 540: Do you think we should verify that the `result` type is an array?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
```diff
@@ -528,7 +530,12 @@ public Expression visitInList(@Nonnull RelationalParser.InListContext ctx) {
                     .processComplexLiteral(tokenIndex, arrayType));
         } else {
             final var inListItems = visitExpressions(ctx.expressions());
-            result = getDelegate().resolveFunction("__internal_array", inListItems.asList().toArray(new Expression[0]));
+            // if in (expressions), wrap it in an internal array
+            if (ctx.LEFT_ROUND_BRACKET() != null) {
```
hatyo: why is this needed? it doesn't look right to me.

hatyo (changes requested):

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizer.java
pengpeng-lu: yes, added tests in AstNormalizerTest.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
pengpeng-lu: you are right, I was confused, reverted the change.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
pengpeng-lu: sure, added

yaml-tests/src/test/resources/in-predicate.yamsql
```diff
@@ -16,11 +16,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+---
+options:
+  supported_version: !current_version
```
hatyo: I don't think we should mark the entire file as only testable against current version. I think you should add `supported_version` specifically around the queries that are only supported with this PR.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/AstNormalizerTests.java
hatyo, line 1: Can you please add a test for `not in` as well?

yaml-tests/src/test/resources/in-predicate.metrics.yaml
```diff
@@ -1,83 +1,83 @@
-unnamed-2:
+unnamed-3:
 -   query: EXPLAIN select a, b from ta where b in (1, 1, 1, 1)
     explain: ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS
         ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }
-    task_count: 945
-    task_total_time_ms: 307
-    transform_count: 288
-    transform_time_ms: 156
+    task_count: 991
```
hatyo, line 5: The increase in these tasks is not related to your PR. However, I think it is exposing a potential bug in the way we build the match candidates.

pengpeng-lu: task_count increases, but task_total_time and transform_time decreases?

hatyo: The reported times should not be taken by face value, the increased number of tasks is interesting, I noticed it is related to the addition of (unrelated) metadata, I reported it here: https://github.com/FoundationDB/fdb-record-layer/issues/3673

Issue #710, opened 2019-09-06
Document the different store format versions


alecgrieser:
The `FDBRecordStore` format versions could use more Javadoc comments. They currently each have a comment (but not a Javadoc comment), but we should move that into Javadoc, especially if we expect more people to start setting them (see #709).

PR #3304, opened 2025-04-15
Transition FormatVersion from constants to an enum

ScottDugas:
The FormatVersion was just an integers, with an associated set of constants, of the form: `FDBRecordStore.*_FORMAT_VERSION`. This PR starts the migration of the code to a new enum: `FormatVersion`, and introduces setters and getters that align with this.
This leaves around the existing constants, and methods to avoid being a breaking change, but they are all now flagged with `@API(API.Status.DEPRECATED)` and will be removed in the near future.

I did not replace the integer usages in the internals of the record store, because I felt that code was complicated enough that it would be worth focusing on independently. I did change everything else to use the new code.

This also adds a bunch of documentation, and thus resolves: #710.
I also added a bunch of documentation to the `DatastoreInfo`, because a lot of the details of that are tightly coupled with the FormatVersion.

Note: This is somewhat related to https://github.com/FoundationDB/fdb-record-layer/pull/3309 as some of the new paths will also throw an exception on invalid states, whereas the integer versions would not.


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreFormatVersionTest.java
```diff
@@ -40,23 +44,25 @@ public class FDBRecordStoreFormatVersionTest extends FDBRecordStoreTestBase {
 
     @Test
     public void testFormatVersionUpgrade() {
+        final List<FormatVersion> sortedVersions = Arrays.stream(FormatVersion.values()).sorted().collect(Collectors.toList());
+        FormatVersion penultimateVersion = sortedVersions.get(sortedVersions.size() - 1);
         try (FDBRecordContext context = openContext()) {
             recordStore = getStoreBuilder(context, simpleMetaData(NO_HOOK))
-                    .setFormatVersion(FDBRecordStore.MAX_SUPPORTED_FORMAT_VERSION - 1)
+                    .setFormatVersion(penultimateVersion)
```
ScottDugas, line 51: I thought about introducing a `FormatVersion.previous()`, but it would only be used for internal tests, and you can't change the numbers associated with the enums, so it seemed best to keep it simple.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreSplitRecordsTest.java
```diff
@@ -134,9 +134,9 @@ public void unsplitCompatibility(SplitRecordsTestConfig testConfig) throws Excep
         try (FDBRecordContext context = openContext()) {
             // Write a record using the old format
             recordStore = getStoreBuilder(context, simpleMetaData(NO_HOOK))
-                    .setFormatVersion(FDBRecordStore.SAVE_UNSPLIT_WITH_SUFFIX_FORMAT_VERSION - 1)
+                    .setFormatVersion(FormatVersion.FORMAT_CONTROL)
```
alecgrieser, line 137: It's probably fine either way, but one reason the tests were written the old way before is that they were trying to say "set the format version to the maximum that does _not_ support the save unsplit feature". That's a little bit less obvious with an explicit call to `FORMAT_CONTROL` like that. I'm not sure what you'd do instead, though, other than having either a `previous()` method on `FormatVersion`, or maybe introducing a local variable here:

```java
FormatVersion unsupportedVersion = FormatVersion.getFormatVersion(FormatVersion.SAVE_UNSPLIT_WITH_SUFFIX_FORMAT_VERSION.getValueForSerialization() - 1);
```

alecgrieser (approved):
I left a few smaller comments about some things that could be improved or were a bit erroneous, but overall, this LGTM, and I like the way this refactoring evolves the API

alecgrieser (commented):
There was one relatively minor typo, but otherwise, LGTM pending resolving the merge conflicts with #3309

alecgrieser (approved):

PR #3318, opened 2025-04-22
bump Gradle version to `8.13`.

hatyo:
This upgrades Gradle version to `8.13` as an attempt to solve a debugging error appearing in recent IntelliJ version 2025.1.


hatyo:
Closed in favor of #3316 which is doing more or less the same.

PR #3320, opened 2025-04-22
Fix broken gradle task: testShadowJar

MMcM:
It isn't used very often, but it is referenced in the comment to `FDBRecordStorePerformanceTest.main`.


MMcM:
See #3319

fdb-record-layer-core/fdb-record-layer-core.gradle
```diff
@@ -63,7 +63,7 @@ task testShadowJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.Shadow
     archiveClassifier = 'standalone-tests'
     from sourceSets.main.output
     from sourceSets.test.output
-    configurations = [ project.configurations.testRuntimeOnly ]
+    configurations = [ project.configurations.testRuntimeClasspath ]
     manifest {
         inheritFrom project.tasks.jar.manifest
         attributes 'Main-Class': 'com.apple.foundationdb.record.provider.foundationdb.FDBRecordStorePerformanceTest',
```
MMcM, line 69: Note that the `shadow` issue for `Multi-Release` is marked as fixed but that's only in the 9.0-beta, so leaving that alone for now.

PR #3324, opened 2025-04-28
New planner rule added to push predicates down in the query representation

alecgrieser:
This introduces a new query planner rule that will be a part of a larger framework to canonicalize query representations. The rule pushes down predicates to ensure that they appear as low in the tree as possible. So, for example, suppose you had a query like:

```sql
SELECT a, b, c
   FROM (SELECT a, b, c, d FROM T)
   WHERE d = ?
   ORDER BY a, b
```

This will produce a new query expression where the `d = ?` is pushed lower in the graph, something like:

```sql
SELECT a, b, c
   FROM (SELECT a, b, c, d FROM T where d = ?)
   ORDER BY a, b
```

This simplification helps us during planning to push down predicate execution to the lowest part of the graph where they can go.

Note that for the moment, this rule has _not_ been added to the default `PlanningRuleSet`. The reason for this is that some of the rewrite rules will have weird overlap with some of the other rules, and so we want to rewrite the query in one phase and then plan and optimize it in another. The framework for that is not there yet


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRule.java
```diff
@@ -0,0 +1,336 @@
...
+        private Quantifier.ForEach getPushQuantifier() {
+            return pushQuantifier;
+        }
+
+        @Nonnull
+        private List<QueryPredicate> updatedPredicates(@Nonnull TranslationMap translationMap, @Nonnull Collection<? extends QueryPredicate> preExistingPredicates) {
+            var predicatesBuilder = ImmutableList.<QueryPredicate>builderWithExpectedSize(getOriginalPredicates().size() + preExistingPredicates.size())
+                    .addAll(preExistingPredicates);
+            for (QueryPredicate originalPredicate : getOriginalPredicates()) {
+                predicatesBuilder.add(originalPredicate.translateValues(translationMap));
```
normen662, line 195: I think this needs to set `shoudSimplifyValues=true`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRule.java
```diff
@@ -0,0 +1,336 @@
...
+            return updatedPredicates(translationMap, ImmutableList.of());
+        }
+
+        @Nonnull
+        public Quantifier.ForEach pushOverChild(@Nonnull final Quantifier.ForEach child) {
+            final var translationMap =
+                    TranslationMap.rebaseWithAliasMap(AliasMap.ofAliases(getPushQuantifier().getAlias(),
+                            child.getAlias()));
+            final var newPredicates = updatedPredicates(translationMap);
+            return Quantifier.forEach(Reference.of(
```
normen662, line 211: Need a memoization call here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRule.java
```diff
@@ -0,0 +1,336 @@
...
+                            selectExpression.getQuantifiers(),
+                            newPredicates));
+        }
+
+        @Nonnull
+        @Override
+        public Optional<GroupByExpression> visitGroupByExpression(@Nonnull final GroupByExpression groupByExpression) {
+            // We have to be a little careful here. In particular, we can push down any predicates on a
+            // grouping column, but not any on the aggregate value. For now, just don't push anything down
+            return Optional.empty();
```
normen662, line 289: I think this is okay for now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRule.java
alecgrieser: Perhaps it was bad form, but that was actually implicit `translateValues(TranslationMap)`. I've removed that and replaced this call with a call to `translateCorrelations(translationMap, true)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java
```diff
@@ -187,17 +197,10 @@ void testSimplePlanGraphWithNullableArray() {
...
-                    graphExpansionBuilder.addQuantifier(qun);
-                    final var nameValue =
-                            FieldValue.ofFieldName(qun.getFlowedObjectValue(), "name");
-                    final var restNoValue =
-                            FieldValue.ofFieldName(qun.getFlowedObjectValue(), "rest_no");
-                    graphExpansionBuilder.addPredicate(new ValuePredicate(restNoValue, new Comparisons.SimpleComparison(Comparisons.Type.GREATER_THAN, 1L)));
-                    graphExpansionBuilder.addResultColumn(resultColumn(nameValue, "nameNew"));
-                    graphExpansionBuilder.addResultColumn(resultColumn(restNoValue, "restNoNew"));
-                    qun = Quantifier.forEach(Reference.of(graphExpansionBuilder.build().buildSelect()));
+                    qun = forEach(selectWithPredicates(qun,
```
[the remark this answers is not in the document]
normen662: Want to create an issue for that?

PR #3327, opened 2025-04-28
query uses grouping index with Is_Null predicate 

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -451,7 +451,7 @@ private BooleanWithConstraint groupingSubsumedBy(@Nonnull final Quantifier candi
         //
         // Consult already bound predicates from the child match.
         //
-        final var equalityPredicates =
+        final var equalityOrIsNullPredicates =
```
normen662, line 454: Let's revert that as all of those predicates are considered equality predicates in a sense that they only allow one value to pass the filter. Even an `IS NULL` is for these purposes a predicate that checks for the equality with `NULL`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -471,13 +471,13 @@ private BooleanWithConstraint groupingSubsumedBy(@Nonnull final Quantifier candi
                         }
 
                         return comparisons.stream()
-                                .anyMatch(comparison -> comparison.getType().isEquality() &&
-                                        comparison.getType() == Comparisons.Type.EQUALS);
+                                .anyMatch(comparison -> (comparison.getType().isEquality() &&
+                                        comparison.getType() == Comparisons.Type.EQUALS) || (comparison.getType() == Comparisons.Type.IS_NULL));
```
normen662, line 475: Can you use the comparison types equality test instead of enumerating the types you like: `comparison.getType().isEquality()`

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/GroupByQueryTests.java
```diff
@@ -46,6 +47,38 @@ public GroupByQueryTests() {
         Utils.enableCascadesDebugger();
     }
 
+    @Test
+    void isNullPredicateUsesGroupIndex() throws Exception {
```
normen662, line 51: Thank you for writing a proper test case so we won't have a problem like this the next time around!

normen662 (changes requested):
Thanks for finding out where the problem was. Tiny comments on how to fix them!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
pengpeng-lu: Got it, reverted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
pengpeng-lu: I changed it to : `c.isEquality() && (c==EQUALS || c== IS_NULL)`, a couple of `TEXT_CONTAINS` comparison types also returns `isEquality() = true`, so I guess we have to check whether it is EQUALS or IS_NULL.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -471,8 +471,8 @@ private BooleanWithConstraint groupingSubsumedBy(@Nonnull final Quantifier candi
                         }
 
                         return comparisons.stream()
-                                .anyMatch(comparison -> comparison.getType().isEquality() &&
-                                        comparison.getType() == Comparisons.Type.EQUALS);
+                                .anyMatch(comparison -> (comparison.getType().isEquality() &&
```
normen662, line 474: Why not just do

```
.anyMatch(comparison -> comparison.getType().isEquality())
```

normen662 (changes requested):

normen662 (approved):

PR #3330, opened 2025-04-30
Introduce planner rule to merge select expressions

alecgrieser:
This adds a new query rewrite rule for creating canonical forms of query expressions. This rule implements "cell merge", which is to say that it will merge together select expressions to form a smaller number of larger selects. This includes the simple case (e.g., of one select modifying the projection of another simple select), but it also means doing things like combining a join of an `n`-way and an `m`-way join into an `n + m`-way join.


alecgrieser:
This is superceded by #3388

PR #3334, opened 2025-05-02
Add RecordCountKey State

### LINKED ISSUE #3326: Add RecordCountKey State
RecordCountKey has been deprecated since 2018, but simply removing it from the metadata could be very risky. We should add a RecordCountKey state to the store header to allow controlling this.
Ideally we would be able to set the RecordCountKey to either:
1. DISABLED - The RecordCountKey is not maintained
2. WRITE_ONLY - The RecordCountKey is maintained, but cannot be used for queries
3. READABLE - The RecordCountKey is maintained and can be used for queries

This allows for a progression, on an individual store, and once that has been validated, the RecordCountKey can be removed from the metadata entirely, with minimal risk.

Note: We don't currently have a mechanism for building the RecordCountKey across multiple transactions, so WRITE_ONLY would only make sense as a sort of "safe" DISABLED, that can be transitioned back to READABLE trivially.

ScottDugas:
This adds a mechanism for controlling the state of the _deprecated_ RecordCountKey, that can be set on the RecordMetaData.

Note: This is different from `COUNT` indexes.

Resolves #3326 

As noted in the javadocs:

The new state is controlled by `FDBRecordStore.updateRecordCountStateAsync`, and allows:
1. Changing a READABLE RecordCount to WRITE_ONLY
2. Changing a WRITE_ONLY RecordCount back to READABLE
3. Changing any state to DISABLED

The default state is READABLE, and it will stay READABLE unless explicitly updated.
If the RecordCount is WRITE_ONLY then it will maintain the count, but not allow querying it.
If the RecordCount is DISABLED it will clear the data, and stop maintaining the values.

In order to support this new state, a new `FormatVersion` has been added: `RECORD_COUNT_STATE`
In order to use the new functionality, the stores must be built with at least this format version.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3237,6 +3273,46 @@ public void clearHeaderUserField(@Nonnull String userField) {
...
+                return builder;
+            }
+            throw new RecordCoreException("Invalid state transition for RecordCountState")
+                    .addLogInfo(LogMessageKeys.OLD, existing)
+                    .addLogInfo(LogMessageKeys.NEW, newState);
+        }).thenRun(() -> {
+            if (newState == RecordMetaDataProto.DataStoreInfo.RecordCountState.DISABLED) {
+                final byte[] begin = Tuple.from(RECORD_COUNT_KEY).pack(getSubspace().getKey());
+                final byte[] end = ByteArrayUtil.strinc(begin);
+                ensureContextActive().clear(begin, end);
```
ScottDugas, line 3311: This could use `Range.startsWith` instead

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
```diff
@@ -651,6 +649,374 @@ void countKeyOnNewStore() throws ExecutionException, InterruptedException {
...
+
+    private void rawAssertHasRecordCount(final RecordMetaData metaData, final boolean hasRecordCounts) {
+        try (FDBRecordContext context = openContext()) {
+            createOrOpenRecordStore(context, metaData);
+            final Tuple tuple = Tuple.from(FDBRecordStoreKeyspace.RECORD_COUNT.key());
+            final byte[] subspace = tuple.pack(recordStore.getSubspace().pack());
+            final List<KeyValue> counts = context.ensureActive()
+                    .getRange(subspace,
+                            ByteArrayUtil.strinc(subspace))
+                    .asList().join();
```
ScottDugas, line 963: This could use `Range.startsWith`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -1850,6 +1858,9 @@ private Key.Evaluated deleteRecordsWhereCheckRecordTypes() {
                 return null;
             }
 
+            // Note: it's possible that the RecordCount is DISABLED, and we could do the deleteWhere, but in a world
+            // where some stores may have it disabled, and others may not, it's probably best to keep this code
+            // simple
```
alecgrieser, line 1863: I'm not sure I buy this argument, as it doesn't seem like the code editions would be that different. We already look at the index state when selecting the set of `IndexMaintainer`s that we need to update, and it seems like we should mirror that behavior here. I could also see the argument that for this function, we should always look at the index, so that this function fails whenever the record meta-data is updated, rather than waiting for the index to be built, but that seems like a separate change.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -2078,6 +2089,9 @@ private CompletableFuture<Void> run() {
             }
 
             final KeyExpression recordCountKey = getRecordMetaData().getRecordCountKey();
+            // Note: The RecordCountKey may be disabled. But if it is, then the data should be clear, and there is no
```
alecgrieser, line 2092: This feels less defensible than in the other method. I think you're right that the clear is ultimately harmless, but it seems out of line with how we treat disabled indexes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3237,6 +3273,46 @@ public void clearHeaderUserField(@Nonnull String userField) {
         context.asyncToSync(FDBStoreTimer.Waits.WAIT_EDIT_HEADER_USER_FIELD, clearHeaderUserFieldAsync(userField));
     }
 
+    /**
+     * Update the RecordCount to have a new state.
+     * <p>
+     *     The state can go from {@code READABLE} to {@code WRITE_ONLY} and vice-versa, and it can always be set to
+     *     {@code DISABLED}, but it cannot transition out of {@code DISABLED}.
```
alecgrieser, line 3280: I assume this is a consequence of not adding a build record count key or some equivalent? That seems fine given that the key is effectively deprecated anyway. I'm not _totally_ sure I see the vision for `WRITE_ONLY` here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4841,6 +4917,8 @@ protected boolean checkPossiblyRebuildRecordCounts(@Nonnull RecordMetaData metaD
 
         if (rebuildRecordCounts) {
             // We want to clear all record counts.
+            // This code will leave data behind if the previous RecordCountKey was not grouped
+            // https://github.com/FoundationDB/fdb-record-layer/issues/3335
```
alecgrieser, line 4921: Are you planning on tackling that in a separate PR targeted to #3335?

fdb-record-layer-core/src/main/proto/record_metadata.proto
```diff
@@ -95,6 +95,27 @@ message DataStoreInfo {
   // When considering using this, one should also consider introducing a new record type to store the data.
   // This was introduced with FormatVersion.HEADER_USER_FIELDS.
   repeated UserFieldEntry user_field = 8;
+
+  // The state of the RecordCount as defined by record_count_key. This is similar to IndexState, but since there is no
+  // mechanism for building the RecordCount across multiple transactions, the state transitions are slightly different.
+  enum RecordCountState {
+    // The RecordCountKey is maintained and usable for FDBRecordStore.getSnapshotRecordCount
+    READABLE = 1;
```
alecgrieser, line 103: I'm not sure if we care, but if we wanted to be proto3 compatible, we should make this value `READABLE = 0`. That's still not entirely compatible with all proto3 guidance, as they suggest that value 0 be a special "uninitialized" state (which I think would mean that to see if the record count key is readable, we'd need to always check if it equals "uninitialized" or equals a different "readable" value). That being said, we already have a number of proto3 incompatible enums in the `record_metadata.proto` file. I suspect that we'll not end up updating the syntax here, but maybe instead updating it to one of the protobuf "editions", and they already have compatibility mode logic to allow proto2 enums to be integrated into newer code, so we could say that another one of those isn't so bad.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
```diff
@@ -651,6 +649,372 @@ void countKeyOnNewStore() throws ExecutionException, InterruptedException {
...
+    void disableRecordCountKey(boolean fromWriteOnly, boolean hasFallback, boolean grouped) {
+        final RecordMetaDataBuilder metaDataBuilder = simpleMetaDataBuilder();
+        addRecordCountKey(metaDataBuilder, grouped ? GROUP_EXPRESSION : EmptyKeyExpression.EMPTY);
+        if (hasFallback) {
+            addCountIndex(metaDataBuilder, grouped ? GROUP_EXPRESSION : EmptyKeyExpression.EMPTY);
+        }
+        IntConsumer assertCount = expected -> {
+            if (grouped) {
+                for (int groupingValue = 1; groupingValue < 6; groupingValue++) {
+                    assertGroupedCount(expected / 5, 1);
```
alecgrieser, line 670:
```suggestion
                    assertGroupedCount(expected / 5, groupingValue);
```

I think this is what you meant here?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
```diff
@@ -71,6 +81,9 @@
 @Tag(Tags.RequiresFDB)
 public class FDBRecordStoreCountRecordsTest extends FDBRecordStoreTestBase {
```
alecgrieser, line 82: I may be missing it, but this could probably use a test of `deleteRecordsWhere` that fails if the record count key isn't aligned, checking with various states of the record count key.

gradle.properties
```diff
@@ -19,7 +19,7 @@
 #
 
 rootProject.name=fdb-record-layer
-version=4.2.6.0
+version=4.3.0.0
```
alecgrieser, line 22: Not that I'm opposed to it, but was there anything in particular that made it seem like this was a good candidate for the 4.3 bump? The new format version?

alecgrieser (approved):
Overall, LGTM. I made a few suggestions, but I would also be happy with this as-is

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ScottDugas: I think that's a fairly solid point.
I also, realized, that if someone is changing something around the RecordCountState, they would probably look for usages, to double check what they need to update, if we're not referencing it here, that could be missed.

I would like to fix this as part of this PR.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ScottDugas: ok, I will elaborate. I thought I had explained, that but clearly I didn't do so here, and repeating here would be good.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ScottDugas: Yes, the existence of the state doesn't change things, but I noticed it, and didn't want it to be left unrecorded.

fdb-record-layer-core/src/main/proto/record_metadata.proto
ScottDugas: I have no strong objections to changing this to `READABLE = 0`, and going up from there. I mostly just copied `Index.Type` below.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
ScottDugas: Yes that is what I meant

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
ScottDugas: Yes, I think you are correct, that is not covered, and it would be good to add that test.

gradle.properties
ScottDugas: Yes, I bumped the minor version because there is a new format version. I can undo it.
I forgot to add a comment to the PR, but I now remember that I did this in part to make sure to start a discussion.

ScottDugas: Discussed offline, and decided that it is worth bumping the minor version in this instance.

fdb-record-layer-core/src/main/proto/record_metadata.proto
ScottDugas: I took a stab at adding "unspecified" to the enums, and erroring if it is unrecognized, that is in a different branch, at least for now, but the commit is here: https://github.com/ScottDugas/fdb-record-layer/commit/5dcb5760f62c1697aaec55dc475a69bfa8395f35 more testing is needed.

ohadzeliger (approved):
Approved - just a couple of small questions.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3276,9 +3275,12 @@ public void clearHeaderUserField(@Nonnull String userField) {
     /**
      * Update the RecordCount to have a new state.
      * <p>
-     *     The state can go from {@code READABLE} to {@code WRITE_ONLY} and vice-versa, and it can always be set to
-     *     {@code DISABLED}, but it cannot transition out of {@code DISABLED}.
-     *     When the RecordCount is set to {@code DISABLED}, the data will be cleared.
+     *     The state can go from {@code READABLE} to {@code WRITE_ONLY}, which is mostly useful to validate that if you
+     *     drop the {@link RecordMetaData#getRecordCountKey() recordCountKey} from the metadata, it won't have adverse
+     *     affects.
+     *     The state can always be set to {@code DISABLED}, at which point there is no way to any other state. This is
```
alecgrieser, line 3281:
```suggestion
     *     The state can always be set to {@code DISABLED}, at which point there is no way to any other state. This
```

fdb-record-layer-core/src/main/proto/record_metadata.proto
alecgrieser: One thing to note is that I believe the code in this PR currently takes advantage of the fact that calling `getRecordCountState()` on the enum field returns its default value in order to collapse the `READABLE` and unset states. Adding a new value 0 will change the default to `UNSPECIFIED`. That's technically against the protobuf evolution guidelines (if we did this later), but I think it's fine as long as the code is modified to either more explicitly treat `UNSPECIFIED` as a synonym of `READABLE` or if we add additional `hasRecordCountKey` checks if we add this later

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/recordrepair/RecordValidationTest.java
```diff
@@ -73,7 +73,7 @@ public static Stream<Arguments> splitAndVersion() {
      */
     @Test
     void monitorFormatVersion() {
-        assertEquals(FormatVersion.CHECK_INDEX_BUILD_TYPE_DURING_UPDATE, FormatVersion.getMaximumSupportedVersion(),
+        assertEquals(FormatVersion.RECORD_COUNT_STATE, FormatVersion.getMaximumSupportedVersion(),
```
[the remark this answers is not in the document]
alecgrieser: It may be a little annoying, but it's probably worth the cost of reminding a developer that they may need to modify other tests here if the record format changes

alecgrieser (approved):
One typo, but otherwise, LGTM

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/recordrepair/RecordValidationTest.java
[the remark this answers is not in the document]
ScottDugas: Yeah, the annoyingness wasn't problematic. it might be more annoying when I have less memory, but it did get me to pause, and confirm that this code didn't need to be updated.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -3237,6 +3272,49 @@ public void clearHeaderUserField(@Nonnull String userField) {
...
+                return builder;
+            }
+            throw new RecordCoreException("Invalid state transition for RecordCountState")
+                    .addLogInfo(LogMessageKeys.OLD, existing)
+                    .addLogInfo(LogMessageKeys.NEW, newState);
+        }).thenRun(() -> {
+            if (newState == RecordMetaDataProto.DataStoreInfo.RecordCountState.DISABLED) {
+                // Note: getSubspace().range(tuple) does not include getSubspace().pack(tuple), but this does
+                // ungrouped recordCountKey will be stored directly at getSubspace().pack(tuple).
+                ensureContextActive().clear(Range.startsWith(getSubspace().pack(Tuple.from(RECORD_COUNT_KEY))));
```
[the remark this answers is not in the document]
ScottDugas: Yeah, this block doesn't have access to the existing state.
However, reflecting, it is probably not great that this is not done as part of `updateStoreHeaderAsync`, because that has the `beginRecordStoreStateWrite` call, so this is not inside that lock.
Technically, since you can't go *from* `DISABLED`, I don't think you can have a race condition, because if the first one changes it to `DISABLED`, the second will not be able to change it.
But perhaps it would make more sense to have it clear inside the block, which would be before it actually updates the store header.

PR #3336, opened 2025-05-02
ParameterizedTest utilities

ScottDugas:
This adds a couple utilities to our testing frameworks:
1. An optional name for `@BooleanSource`, so that it doesn't just appear as `true` / `false`
2. Methods for producing similar named boolean streams for use in a `@MethodSource`
3. A method to produce the cartesian product of a bunch of streams, instead of having to chain a whole bunch of `flatMap`s.

Below is a screenshot of the test results after this change, for `FDBRecordStoreUniqueIndexTest` that takes full advantage of the new `@BooleanSource` functionality
image
vs the old way:
image
vs if you don't have the `name=` on all the tests
image

(Note: `removeUniquenessConstraintDuringTransactionWithNewDuplications` was taking `removeUniquenessConstraintArguments` but only actually used one of the parameters, causing it to run twice as many times as needed)

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
+++ b/fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
@@ -37,7 +37,7 @@ class BooleanArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<
 
     @Override
     public void accept(BooleanSource booleanSource) {
-        this.name = booleanSource.name();
+        this.name = booleanSource.value();
     }
 
     @Override
--- a/fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
+++ b/fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
@@ -20,7 +20,6 @@
 
 package com.apple.test;
 
-import org.junit.jupiter.api.Named;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.ArgumentsProvider;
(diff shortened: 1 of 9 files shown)


fdb-extensions/src/test/java/com/apple/test/ParameterizedTestUtils.java
```diff
@@ -0,0 +1,78 @@
...
+     * and so-on. Producing all combinations of an element from each source.
+     */
+    public static Stream<Arguments> cartesianProduct(Stream<?>... sources) {
+        return cartesianProduct(
+                Stream.of(sources)
+                        .map(stream -> stream.collect(Collectors.toList())).collect(Collectors.toList()))
+                .map(arguments -> Arguments.of(arguments.toArray()));
+    }
+
+    private static Stream<Stream<Object>> cartesianProduct(List<List<?>> sources) {
```
ohadzeliger, line 67: Neat!

ohadzeliger (changes requested):
This is very cool. Just a couple of small things.

fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
```diff
@@ -25,25 +25,29 @@
...
 
-import javax.annotation.Nonnull;
 import java.util.Arrays;
-import java.util.List;
 import java.util.stream.Stream;
 
 /**
  * Argument provider for the {@link BooleanSource} annotation for providing booleans to parameterized
- * tests. Regardless of the source or context, this always returns {@code false} and {@code true} in that order.
+ * tests. Regardless of the source or context, this always returns {@code false} and {@code true} in that order for each
```
ScottDugas, line 33: @ohadzeliger I mentioned in the commit message, but the original javadoc stated the order these would come back. I don't see a reason to change that behavior here, so I changed `booleans` to return in the same order.

fdb-extensions/src/test/java/com/apple/test/BooleanArgumentsProvider.java
```diff
@@ -20,30 +20,37 @@
...
     public Stream<? extends Arguments> provideArguments(ExtensionContext extensionContext) throws Exception {
-        return BOOLEANS.stream();
+        if (names.length == 0) {
+            throw new IllegalStateException("@BooleanSource has an empty strings");
+        }
+        return ParameterizedTestUtils.cartesianProduct(
+                Arrays.stream(names).map(name ->
+                        Stream.of(
+                                Named.of(name.isBlank() ? "true" : name, true),
+                                Named.of(name.isBlank() ? "false" : "! " + name, false)
```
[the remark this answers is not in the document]
ohadzeliger: It would likely become clearer with some miles of usage, we can change it as we go.

ohadzeliger (approved):

PR #3347, opened 2025-05-07
Create FutureManagerRunner and convert FDBDatabaseRunnerImpl to use it.

### LINKED ISSUE #3346: Add future management to TransactionalRunner
In order to allow extensions of TransactionalRunner for other purposes, the future management (the closing of futures once the runner is closed) should be added. This functionality is currently in the FDBDatabaseRunnerImpl class and should be made available separately.

ohadzeliger:
This refactoring moves the future management (completing of futures once the runner closes) out of the FDBDatabaseRunnerImpl into a subclass of TransactionalRunner. 
This allows the new subclass of TransactionalRunner to be used (e.g in the case where we want another throttling implementation).

Resolves #3346

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunnerImpl.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunnerImpl.java
@@ -29,7 +29,8 @@ import com.apple.foundationdb.record.RecordCoreRetriableTransactionException;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.provider.foundationdb.runners.ExponentialDelay;
-import com.apple.foundationdb.record.provider.foundationdb.runners.FutureManagerRunner;
+import com.apple.foundationdb.record.provider.foundationdb.runners.FutureAutoClose;
+import com.apple.foundationdb.record.provider.foundationdb.runners.TransactionalRunner;
 import com.apple.foundationdb.record.provider.foundationdb.synchronizedsession.SynchronizedSessionRunner;
 import com.apple.foundationdb.record.util.Result;
 import com.apple.foundationdb.subspace.Subspace;
@@ -55,7 +56,8 @@ public class FDBDatabaseRunnerImpl implements FDBDatabaseRunner {
 
     @Nonnull
     private final FDBDatabase database;
-    private final FutureManagerRunner innerRunner;
+    private final TransactionalRunner transactionalRunner;
+    private final FutureAutoClose futureManager;
     @Nonnull
     private FDBRecordContextConfig.Builder contextConfigBuilder;
     @Nonnull
@@ -72,7 +74,8 @@ public class FDBDatabaseRunnerImpl implements FDBDatabaseRunner {
         this.database = database;
         this.contextConfigBuilder = contextConfigBuilder;
         this.executor = database.newContextExecutor(contextConfigBuilder.getMdcContext());
-        this.innerRunner = new FutureManagerRunner(database, contextConfigBuilder);
+        this.transactionalRunner = new TransactionalRunner(database, contextConfigBuilder);
+        this.futureManager = new FutureAutoClose();
 
         final FDBDatabaseFactory factory = database.getFactory();
         this.maxAttempts = factory.getMaxAttempts();
@@ -159,7 +162,7 @@ public class FDBDatabaseRunnerImpl implements FDBDatabaseRunner {
     @Override
     @Nonnull
     public FDBRecordContext openContext() {
-        return innerRunner.openContext();
+        return transactionalRunner.openContext();
     }
 
     private class RunRetriable<T> {
@@ -220,7 +223,7 @@ public class FDBDatabaseRunnerImpl implements FDBDatabaseRunner {
                     if (getTimer() != null) {
                         future = getTimer().instrument(FDBStoreTimer.Events.RETRY_DELAY, future, executor);
                     }
-                    innerRunner.registerFuture(future);
+                    futureManager.registerFuture(future);
                     return future.thenApply(vignore -> {
                         currAttempt++;
                         return true;
@@ -235,10 +238,10 @@ public class FDBDatabaseRunnerImpl implements FDBDatabaseRunner {
         @SuppressWarnings("squid:S1181")
         public CompletableFuture<T> runAsync(@Nonnull final Function<? super FDBRecordContext, CompletableFuture<? extends T>> retriable,
                                              @Nonnull final BiFunction<? super T, Throwable, Result<? extends T, ? extends Throwable>> handlePostTransaction) {
-            CompletableFuture<T> future = innerRunner.newFuture();
+            CompletableFuture<T> future = futureManager.newFuture();
             AsyncUtil.whileTrue(() -> {
                 try {
-                    return innerRunner.runAsync(currAttempt != 0, retriable)
+                    return transactionalRunner.runAsync(currAttempt != 0, retriable)
                             .handle((result, ex) -> {
                                 Result<? extends T, ? extends Throwable> newResult = handlePostTransaction.apply(result, ex);
                                 return handle(newResult.getValue(), newResult.getError());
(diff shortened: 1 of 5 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
```diff
@@ -0,0 +1,83 @@
...
+        futuresToClose.removeIf(CompletableFuture::isDone);
+        futuresToClose.add(future);
+        return future;
+    }
+
+    @Override
+    public synchronized void close() {
+        if (!isClosed()) {
+            // Close super first, to prevent same-thread nested calls from proceeding
+            super.close();
```
ScottDugas, line 72: If this fails, it won't mark the futures as exceptional.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
```diff
@@ -0,0 +1,83 @@
...
+ * Futures created externally can be registered with this runner. This runner can also create new futures (and register them).
+ * Once registered, the incomplete futures will be completed exceptionally (with a
+ * {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner.RunnerClosed} exception) once the runner closes.
+ * It is the responsbility of the users of the runner to ensure futures that it creates are registered.
+ * <p>
+ * Normally, only the top level (root) of the future chain needs registration. Completing this future
+ * will cause the completion to trickle down to the dependent futures.
+ */
+@API(API.Status.INTERNAL)
+public class FutureManagerRunner extends TransactionalRunner {
```
ScottDugas, line 44: It would be good to have some isolated tests of this. You should be able to mimic the tests that already exist on the `FDBDatabaseRunnerImpl`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public synchronized void close() {
+        if (!isClosed()) {
+            // Close super first, to prevent same-thread nested calls from proceeding
+            super.close();
+
+            if (!futuresToClose.stream().allMatch(CompletableFuture::isDone)) {
+                final Exception exception = new FDBDatabaseRunner.RunnerClosed();
```
ScottDugas, line 75: I wonder if there's a good way to not have that class nested. I can't think of anything that wouldn't be fairly breaking.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ScottDugas, line 44: I _feel_ like it would be better to make this a separate class, and not extend `TransactionalRunner`. That gives two benefits:
1. You would be able to test this without having to worry about the database and contexts.
2. It would be usable in situations where it's not tied to a `TransactionalRunner`

I'm not sure those justify changing this as part of this PR.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A simple runner that extends {@link TransactionalRunner} by completing futures once the runner closes.
+ * Futures created externally can be registered with this runner. This runner can also create new futures (and register them).
+ * Once registered, the incomplete futures will be completed exceptionally (with a
+ * {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner.RunnerClosed} exception) once the runner closes.
+ * It is the responsbility of the users of the runner to ensure futures that it creates are registered.
```
jjezra, line 38: Please consider rewording to something like: 
When {@link close} is called,  any registered futures which is still incomplete will end excpetionally with a {@link com.apple.foundationdb.record.provider.foundationdb.FDBDatabaseRunner.RunnerClosed} exception. 
Note that no exception is thrown within the {@linke close} function.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ohadzeliger: Done (this is no longer a subclass)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ohadzeliger: Good point. The class is now separate and deserves a dedicated test suite (disconnected from its usage in other classes).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ohadzeliger: Discussed offline. Probably not at this point.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ohadzeliger: Good point. Tests, in particular, are now much easier to implement. Done.
Care would have to be taken when orchestrating this with other runners (e.g. order of closing, etc). If this becomes a concern, we can put together a combined class for this + runner.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureManagerRunner.java
ohadzeliger: Sure, done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/FutureAutoClose.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * FutureManagerRunner.java
```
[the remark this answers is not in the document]
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunnerImpl.java
```diff
@@ -311,12 +311,25 @@ public synchronized void close() {
...
+            caught = e;
+        }
+        try {
+            transactionalRunner.close();
+        } catch (RuntimeException e) {
+            if (caught != null) {
+                LOGGER.warn("Caught exception while closing", caught);
+            }
+            // replace
+            caught = e;
```
ScottDugas, line 328: I think the modern way to do this would be to add `cause` as a suppressed exception of `e`
```suggestion
            if (caught != null) {
                caught.addSuppressed(e);
            } else {
                caught = e;
            }
```
See: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Throwable.html#addSuppressed(java.lang.Throwable)

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBDatabaseRunnerImpl.java
ohadzeliger: Yes, that would work. Done.

ScottDugas (approved):

jjezra (approved):

PR #3354, opened 2025-05-10
Fix gradle 9 warnings and bump antlr version

arnaud-lacurie:


fdb-extensions/src/main/java/com/apple/foundationdb/async/rtree/RTreeHilbertCurveHelpers.java
```diff
@@ -69,7 +69,7 @@ private static BigInteger toIndex(int numBits, long... transposedIndexes) {
         for (int i = 0; i < numBits; i++) {
             for (final long transposedIndex : transposedIndexes) {
                 if ((transposedIndex & mask) != 0) {
-                    b[b.length - 1 - bIndex / 8] |= 1 << (bIndex % 8);
+                    b[b.length - 1 - bIndex / 8] |= (byte)(1 << (bIndex % 8));
```
ScottDugas, line 72: I'm confused how this change relates to Gradle 9. Looking at the logs, there still seem to be a bunch of gradle warnings, and no changes in this PR to fix the gradle files. Meanwhile #3358 Seems to be tackling a similar goal, and has no changes to the java code.
It also seems better to split the gradle 9 warning changes, and the antlr upgrade into separate PRs.

ScottDugas (changes requested):

fdb-extensions/src/main/java/com/apple/foundationdb/async/rtree/RTreeHilbertCurveHelpers.java
jen20: I actually hit this also and decided not to address it in #3358: this change is about upgrading the JDK, not Gradle. If you compile with JDK23 from `main`, these show up today.

Issue #196, opened 2018-11-30
Add a method on FDBRecordStore to check if a record exists


alecgrieser:
Currently, the only way to see if a record exists is to try and load the record. The FDB client doesn't really allow us to check for the presence of keys (at the moment), so the best we can do is something like read the record from the database but not deserialize it (and save some CPU work). If, at some point, the FDB client allows us to query for the presence of keys, we could theoretically change that function to save some network bandwidth as well.

PR #3357, opened 2025-05-16
Support isDistinctFrom and isNotDistinctFrom operators

### LINKED ISSUE #3504: Implement `is [not] distinct from` predicate
We aim to add support for these expressions to enable null-safe (in)equality comparisons as specified in the SQL:2023 standard:

```
<distinct predicate> ::=
    <row value predicand 3> <distinct predicate part 2>

<distinct predicate part 2> ::=
    IS [ NOT ] DISTINCT FROM <row value predicand 4>
```

pengpeng-lu:
This resolves #3504.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -234,24 +234,25 @@ public static Object toClassWithRealEquals(@Nullable Object obj) {
...
     public static int compare(@Nullable Object fieldValue, @Nullable Object comparand) {
-        if (fieldValue == null) {
-            if (comparand == null) {
-                return 0;
-            } else {
-                return -1;
-            }
-        } else if (comparand == null) {
-            return 1;
+        return toComparable(fieldValue).compareTo(toComparable(comparand));
```
pengpeng-lu, line 237: I think the original line 237~244 is wrong? If any of the left or right is null, the comparison result should be unknown? Also in the original code, we call: `if(value == null) return null` before we call this `compare` method, so the block of `if(fieldValue == null)` was never reached.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -707,28 +713,44 @@ public static Type invertComparisonType(@Nonnull final Comparisons.Type type) {
     @Nullable
     @SpotBugsSuppressWarnings("NP_BOOLEAN_RETURN_NULL")
     public static Boolean evalComparison(@Nonnull Type type, @Nullable Object value, @Nullable Object comparand) {
-        if (value == null) {
-            return null;
-        }
```
pengpeng-lu, line 712: as I mentioned in another comment, because of this is called first, we'll never reach the `if(fieldValue == null)` block in method `compare`. I removed this block here, and added to cases below.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -184,6 +184,13 @@ private static Optional<QueryPredicate> promoteOperandsAndCreatePredicate(@Nulla
                                                                               @Nonnull Value leftChild,
                                                                               @Nonnull Value rightChild,
                                                                               @Nonnull final Comparisons.Type comparisonType) {
+        if (leftChild.getResultType().getTypeCode() == Type.TypeCode.NULL && rightChild.getResultType().getTypeCode() == Type.TypeCode.NULL) {
```
pengpeng-lu, line 187: I'm a little confused of the difference between `TypeCode.Unknown` and `TypeCode.NULL`, should we use `NULL` or `Unknown` here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -763,6 +816,121 @@ private enum BinaryPhysicalOperator {
...
+        IS_DISTINCT_FROM_UE(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.UNKNOWN, Type.TypeCode.ENUM, (l, r) -> true),
+        IS_DISTINCT_FROM_IDID(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.UUID, Type.TypeCode.UUID, (l, r) -> Comparisons.evalComparison(Comparisons.Type.IS_DISTINCT_FROM, l, r)),
+        IS_DISTINCT_FROM_IDS(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.UUID, Type.TypeCode.STRING, (l, r) -> Comparisons.evalComparison(Comparisons.Type.IS_DISTINCT_FROM, l, PromoteValue.PhysicalOperator.stringToUuidValue((String) r))),
+        IS_DISTINCT_FROM_SID(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.STRING, Type.TypeCode.UUID, (l, r) -> Comparisons.evalComparison(Comparisons.Type.IS_DISTINCT_FROM, PromoteValue.PhysicalOperator.stringToUuidValue((String) l), r)),
+        IS_DISTINCT_FROM_UID(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.UNKNOWN, Type.TypeCode.UUID, (l, r) -> true),
+        IS_DISTINCT_FROM_IDU(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.UUID, Type.TypeCode.UNKNOWN, (l, r) -> true),
+        IS_DISTINCT_FROM_NN(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.NULL, Type.TypeCode.NULL, (l, r) -> false),
```
pengpeng-lu, line 873: again, should we use `NULL` or `Unknown` here?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: right now, if type = TEXT_CONTAINS_ALL, or anything that is not in the `switch` block, and `value == null`, this method returns null. Is this the expected behavior?

fdb-relational-core/src/main/antlr/RelationalLexer.g4
```diff
@@ -127,6 +127,8 @@ INSERT:                              'INSERT';
 INTERVAL:                            'INTERVAL';
 INTO:                                'INTO';
 IS:                                  'IS';
+IS_DISTINCT_FROM:                    'IS DISTINCT FROM';
+IS_NOT_DISTINCT_FROM:                'IS NOT DISTINCT FROM';
```
hatyo, line 131: This is unusual, you should split the tokens to enable tokenizing them correctly disregarding whitespaces.

pengpeng-lu: got it, modified.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/StandardQueryTests.java
```diff
@@ -881,6 +881,209 @@ void aliasingTableToResolveAmbiguityWorks() throws Exception {
         }
     }
 
+    @Test
```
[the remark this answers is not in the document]
pengpeng-lu: done.

pengpeng-lu:
> It feels like there should be some changes added to `ExpressionVisitor.java` that I do not see in this PR, but I am not sure.

is there any specific queries that you are concerned about? I think this is a feature on the "operator" level, not "expression" level.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
arnaud-lacurie: As far as I understood from conversations with @normen662 and @hatyo, `UNKNOWN` is used by the legacy planner, and `NULL` is used by the cascades planner to represent the literal `NULL`. You will want to add both to this map, as you've done here. 
You will also want to add tests in `BooleanValueTest`.
Currently missing from this list is `NULL` on the right hand side of all the other types, and `NULL` on the left hand side.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
hatyo: This is probably a premature constant folding optimization and should be instead done in the optimizer rewrites, I think you can just remove it for now, and add a dedicated predicate simplification rule separately.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -290,6 +299,20 @@ private static Value encapsulate(@Nonnull final String functionName,
...
+                    return encapsulate("notNull", Comparisons.Type.NOT_NULL, List.of(arg0));
+                }
+            }
+            if ("notDistinctFrom".equals(functionName)) {
+                if (res0.getTypeCode() == Type.TypeCode.NULL && res1.getTypeCode() != Type.TypeCode.NULL) {
+                    return encapsulate("isNull", Comparisons.Type.IS_NULL, List.of(arg1));
+                } else if (res1.getTypeCode() == Type.TypeCode.NULL && res0.getTypeCode() != Type.TypeCode.NULL) {
+                    return encapsulate("isNull", Comparisons.Type.IS_NULL, List.of(arg0));
+                }
+            }
```
hatyo: Same comment about premature optimization as in line 187.

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
normen662: We should not use either of the two types and this code should not be here. I agree with Youssef that it should be removed.

pengpeng-lu: so the reason that the code is here, is because of the check at line #196: `final var maxtype = Verify.verifyNotNull(Type.maximumType(leftChild.getResultType(), rightChild.getResultType()));`. It'll raise an error for queries like `where null is distinct from null`. Do we want to remove the check at line#196 instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
MMcM: This method currently implements full collation for use in contexts where `NULL` is to be considered just another value. For example, that is how index entries sort with `NULLS FIRST`. It is still reasonable to use such a function in contexts that implement tri-valued logic by guarding calls, as many callers do today.

It is equally reasonable to have this method _only_ implement comparison between non-null values and to expect _all_ callers to deal with nulls, either sorting them first or producing unknown, as is appropriate to their context. And _both_ contexts _are_ appropriate in the larger system, which must deal sometimes with the "query" view that has unknown and other times with the "storage" view that does not. It might be nice to have a clearer separation so that it is unambiguous which view applies, perhaps by using entirely different currency and not just bare `Object`.

What this change does is something in between, which doesn't seem right. Specifically, if this does not handle `null` at all, then it ought not to have `@Nullable`. The signature should also be changed. I am surprised that the IDE / style checker(s) doesn't flag this for a possible NPE. But otherwise, it's okay provided _all_ callers now check both sides.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -234,24 +234,24 @@ public static Object toClassWithRealEquals(@Nullable Object obj) {
...
+            return toClassWithRealEquals(value).equals(toClassWithRealEquals(comparand));
         }
     }
 
-    @Nullable
     @SpotBugsSuppressWarnings("NP_BOOLEAN_RETURN_NULL")
-    private static Boolean compareEquals(Object value, Object comparand) {
-        if (value == null || comparand == null) {
-            return null;
+    private static boolean compareNotDistinctFrom(Object value, Object comparand) {
```
MMcM: This should have `@Nullable` explicitly to be clear what the expectations are.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -285,6 +285,9 @@ private static Boolean compareStartsWith(@Nullable Object value, @Nullable Objec
     @Nullable
     @SpotBugsSuppressWarnings("NP_BOOLEAN_RETURN_NULL")
     private static Boolean compareLike(@Nullable Object value, @Nullable Object pattern) {
+        if (value == null) {
```
MMcM, line 285: Does this want to similarly handle `LIKE ?` with a `NULL` binding, rather than getting `Illegal pattern`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
MMcM: Yes, throwing for unknown type before checking either side does seem better.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
```diff
@@ -234,24 +234,24 @@ public static Object toClassWithRealEquals(@Nullable Object obj) {
...
+        return toComparable(fieldValue).compareTo(toComparable(comparand));
+    }
+
+    @SpotBugsSuppressWarnings("NP_BOOLEAN_RETURN_NULL")
+    private static boolean compareEquals(@Nonnull Object value, @Nonnull Object comparand) {
+        if (value instanceof Message) {
+            return MessageHelpers.compareMessageEquals(value, comparand);
         } else {
-            return toComparable(fieldValue).compareTo(toComparable(comparand));
+            return toClassWithRealEquals(value).equals(toClassWithRealEquals(comparand));
```
MMcM, line 242: Can we make `toClassWithRealEquals` also `@Nonnull` and remove the `null` check in it, to further ensure that all callers are similarly consistent with the rest of this change?

MMcM (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
normen662: The maximum type of two NullTypes should be the NullType. So maximumType seems to not do the right thing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: 👍 after removing the `if (value == null) return null` block, unknown type throws exception regardless of input value being null.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: I checked that in sql, `where x like a` returns null when x or a is null, so I think we should follow that here?

pengpeng-lu: I removed the `if(value == null)` block from here though, because `LikeOperatorValue.likeOperation` already handles nulls operations.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java
pengpeng-lu: Got it, I changed the method to compare between non-null values.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
pengpeng-lu: yes I think it was probably a different bug, removed the above pre-mature optimization and it works now.

MMcM (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -451,6 +455,36 @@ private static Value encapsulate(@Nonnull BuiltInFunction<Value> builtInFunction
         }
     }
 
+    /**
+     * The {@code isDistinctFrom} function.
+     */
+    @AutoService(BuiltInFunction.class)
+    public static class IsDistinctFromFn extends BuiltInFunction<Value> {
```
hatyo, line 462: Can you add unit tests for this function, as well as `NotDistinctFromFn`? You can extend the unit tests in `BooleanValueTest` and add tests for these functions with variations of argument types so you can test the underlying physical operators.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -1074,8 +1074,15 @@ private enum BinaryPhysicalOperator {
...
+            // handle nulls
             if (arg1 == null || arg2 == null) {
-                return null;
+                if (type == Comparisons.Type.IS_DISTINCT_FROM) {
+                    return arg1 != null || arg2 != null;
+                } else if (type == Comparisons.Type.NOT_DISTINCT_FROM) {
+                    return arg1 == null && arg2 == null;
+                } else {
+                    return null;
+                }
```
hatyo, line 1085: It is interesting that we have this early treatment of `null` values here instead of delegating it to the actual underlying physical operator.

I opened an issue to fix this, as well as introducing `NULL` type code handling (instead of the deprecated `UNKNOWN`).
https://github.com/FoundationDB/fdb-record-layer/issues/3526

hatyo (approved):
LGTM.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -894,6 +929,119 @@ private enum BinaryPhysicalOperator {
         GT_IDN(Comparisons.Type.GREATER_THAN, Type.TypeCode.UUID, Type.TypeCode.NULL, (l, r) -> null),
         GTE_NID(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.NULL, Type.TypeCode.UUID, (l, r) -> null),
         GTE_IDN(Comparisons.Type.GREATER_THAN_OR_EQUALS, Type.TypeCode.UUID, Type.TypeCode.NULL, (l, r) -> null),
+
+        IS_DISTINCT_FROM_BU(Comparisons.Type.IS_DISTINCT_FROM, Type.TypeCode.BOOLEAN, Type.TypeCode.NULL, (l, r) -> Objects.nonNull(l)),
```
arnaud-lacurie: Can you adjust the naming for these? 
I think you're using the letter `U` from when those were unknown, but looking at the rest of the file, you'd want to use `N` for nulls.
For example `IS_DISTINCT_FROM_BU` -> `IS_DISTINCT_FROM_BN`

arnaud-lacurie (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
pengpeng-lu: yes I'm not sure about this as well, but putting it in the operators will result in a lot of duplicated code. Fixing it in a separate PR sounds good.

normen662 (approved):

PR #3358, opened 2025-05-20
Fix deprecated Gradle syntax and configuration features

jen20:
When running `./gradlew shadowJar compileTestJava --warning-mode=all` on the current HEAD of `main` (68f805abea8), a number of warnings are produced for deprecated usages of Gradle configuration features. Upon fixing the trivial deprecations produced in the first pass, a second round of issues is also produced. This pull request fixes all of those, such that the command executes cleanly, without generating a Gradle problem report.

Each class of issue is addressed in a separate commit (which contains links to the Gradle documentation showing the fix mechanism to be the recommended approach), but broadly speaking there are five categories:

1. Use of `destination` instead of `outputLocation` when configuring JUnit report output location ([docs](https://docs.gradle.org/8.13/dsl/org.gradle.api.reporting.Report.html#org.gradle.api.reporting.Report:destination))
2. Consistent use of `key = value` assignment rather than "space assignment" or `propertyName(value)` assignment ([docs](https://docs.gradle.org/8.13/userguide/upgrading_version_8.html#groovy_space_assignment_syntax))
3. Use of plugin blocks instead of conventions for the `Java` plugin ([docs](https://docs.gradle.org/8.13/userguide/upgrading_version_8.html#java_convention_deprecation))
4. Use of plugin blocks instead of conventions for the `Application` plugin ([docs](https://docs.gradle.org/8.13/userguide/upgrading_version_8.html#application_convention_deprecation)) and applying the ensuing required cleanup from the Gradle 7 migration ([docs](https://docs.gradle.org/current/userguide/upgrading_version_7.html#javaapplication_api_cleanup))
5. Stopping the use of `Task.project` in Task actions ([docs](https://docs.gradle.org/8.13/userguide/upgrading_version_7.html#task_project)).

Other tasks which are not dependents of `shadowJar` or `compileTestJava` likely still use deprecated syntax or conventions - I will follow up with a second pull request to address those in future.


jen20:
Incidentally I did not happen to see #3354 from @arnaud-lacurie (hi Arnaud!) before doing this, but it is different: that fixes the warnings for building with a newer JDK _using_ Gradle 9, but does not address the issues of feature and configuration deprecation within Gradle itself.

arnaud-lacurie (approved):

PR #3365, opened 2025-05-22
Don't memoize correlation sets in planner objects

normen662:
We should not use a memoizing supplier to cache the correlation sets of a quantifier since the reference the quantifier ranges over is mutable with respect to the contained expressions. While these expressions are rarely exposing new correlations it may be that a new expression is added that has no correlations at all. If we prune to that expression later, the correlation set, if memoized, is wrong.

Example
```
SELECT *
FROM A, (SELECT * FROM B WHERE A = 5 AND FALSE)
```

The inner
```
...(SELECT * FROM B WHERE A = 5 AND FALSE)...
```
has `B` in its correlation set. That expression, however, can also be simplified to:
```
...(SELECT * FROM B WHERE FALSE)...
```
(and even further). The correlation set is empty.

This PR deals with the change of removing the `MemoizingSupplier` as well as the fallout from now exposed bugs.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAggregateIndexPlan.java
```diff
@@ -260,13 +260,11 @@ public RecordQueryAggregateIndexPlan translateCorrelations(@Nonnull final Transl
 
         final var translatedIndexPlan = indexPlan.translateCorrelations(translationMap,
                 shouldSimplifyValues, translatedQuantifiers);
-        final var newResultValue = resultValue.translateCorrelations(translationMap, shouldSimplifyValues);
-        final var newGroupByResultValue = groupByResultValue.translateCorrelations(translationMap, shouldSimplifyValues);
 
         // this is ok as there are no new quantifiers
-        if (translatedIndexPlan != indexPlan || newResultValue != resultValue || newGroupByResultValue != groupByResultValue) {
-            return new RecordQueryAggregateIndexPlan(translatedIndexPlan, recordTypeName, toRecord, newResultValue,
-                    newGroupByResultValue, constraint);
```
alecgrieser, line 269: Why is it okay to remove these translations? Overall, I'm not sure I understand what the changes to this class are getting at, if I'm honest

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationTest.java
```diff
@@ -174,7 +174,7 @@ void testPullUpValue2() {
         final var new_x_b = FieldValue.ofFieldNames(upperCurrentValue, ImmutableList.of("_1", "x", "xb"));
 
         final var expectedMap = ImmutableMap.of(
-                record_type, new RecordTypeValue(QuantifiedObjectValue.of(UPPER_ALIAS, new Type.AnyRecord(true))),
+                record_type, new RecordTypeValue(FieldValue.ofFieldName(upperCurrentValue, "_1")),
```
alecgrieser, line 177: What was the logic in changing this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAggregateIndexPlan.java
normen662: So there are two attributes (both `Values`) that are in this class which shouldn't be in here as the nothing in here needs it. So they are planner breadcrumbs. And worse, these `Values` are values that come from the match candidate that this index plan is associated with... together with quantifier references that make sense within the candidate but not within the query. So translating it here is outright wrong. And since these values are redundant at best, incorporating them into the hashing and the equality logic is wrong as well.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/ValueComputationTest.java
normen662: The `RecordTypeValue(...)` has a child but never advertised it causing other trouble. Correcting this lead me to realize that this test case expected the wrong output all along.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAggregateIndexPlan.java
normen662: This raises the question as to why they cannot be removed. They can and they should. But that would require us to bump the plan hash version.

alecgrieser: Do we have an issue about this? It would be nice if there were a link out to that issue in here, which may aid future readers of this code block

alecgrieser (approved):

PR #3368, opened 2025-05-26
reimplementation of `rebase(...)` and `translateCorrelations(...)` for references

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/InferredTranslationMap.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Optional;
+
+/**
+ * Translation map that allows for rebasing of graphs in a way that the resulting rebased graph is only using unique
+ * aliases. This property is needed when a sub-graph that is pre-compiled is later unified with the current query's
+ * graph, and we have to ensure that aliases between the sub-graph and the encompassing graph do not clash.
+ * The logic adds a unique source to target mapping when a source is seen for the first time. When the same source is
+ * then mapped again, the previously fixed target is returned.
+ */
+public class InferredTranslationMap implements TranslationMap {
```
alecgrieser: I'm not sure `Inferred` is the best name here. The more interesting thing about this translation map is that it generates unique quantifiers as it goes. Maybe `UniqueAliasTranslationMap` or even `ToUniqueAliasesTranslationMap`? (I would have assumed that an "inferred" translation map would do something like take some kind of subset of equivalencies and then "infer" missing ones, which isn't quite what what this is doing.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/RegularTranslationMap.java
```diff
@@ -0,0 +1,217 @@
...
+    @Nonnull
+    private final Map<CorrelationIdentifier, TranslationFunction> aliasToFunctionMap;
+
+    private RegularTranslationMap(@Nonnull final Map<CorrelationIdentifier, TranslationFunction> aliasToFunctionMap) {
+        this.aliasToFunctionMap = ImmutableMap.copyOf(aliasToFunctionMap);
+    }
+
+    @Nonnull
+    @Override
+    public Optional<AliasMap> getAliasMapMaybe() {
```
alecgrieser, line 56: Using inheritance here to determine the behavior of this method (that is, having the base class always return `Optional.empty()` and then having `AliasMapBasedTranslationMap` override it) I think confuses things a bit. I definitely thought that this method would just always return `Optional.empty()` at first. It almost feels like it would be better to have either a nullable or optional `AliasMap` member of this class, and then this method would be able to just return it. Then I don't think you'd need the overriding subclass at all

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/TranslationMap.java
```diff
@@ -20,137 +20,78 @@
...
+        return defaultAlias;
     }
 
-    @Nonnull
-    public Value applyTranslationFunction(@Nonnull final CorrelationIdentifier sourceAlias,
-                                          @Nonnull final LeafValue leafValue) {
-        final var translationTarget = Preconditions.checkNotNull(aliasToTargetMap.get(sourceAlias));
-        return translationTarget.translate(sourceAlias, leafValue);
+    @Nullable
+    default CorrelationIdentifier getTarget(@Nonnull final CorrelationIdentifier sourceAlias) {
```
alecgrieser: Given that the two implementations do different things with this method, it might make sense to just make it a non-default method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
```diff
@@ -205,11 +206,19 @@ public String getShorthand() {
...
+        }
+
+        @Override
+        @Nonnull
+        public ForEach overNewReference(@Nonnull final Reference reference,
+                                        @Nonnull final CorrelationIdentifier newAlias) {
             return Quantifier.forEachBuilder()
-                    .from(this)
+                    .withAlias(newAlias)
+                    .setNullOnEmpty(isNullOnEmpty())
```
alecgrieser: Would it be better to do something like:

```suggestion
                    .from(this)
                    .withAlias(newAlias)
```

I think it has more-or-less the same semantics, but it leaves the copying of all of the state to the quantifier's builder rather than having the caller here need to know about what fields it has. (That is, I think it's more likely that if we added a new field in the future that we'd remember to update `Quantifier.forEachBuilder.from` rather than here.)

The changes in here to the `Existential` and `Physical` quantifiers follow a similar pattern

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
```diff
@@ -205,11 +206,19 @@ public String getShorthand() {
             return "ƒ";
         }
 
-        @Override
         @Nonnull
+        @Override
         public ForEach overNewReference(@Nonnull final Reference reference) {
```
alecgrieser, line 211: Is this override maintained just to override the return value type?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ConstraintsMap.java
```diff
@@ -301,6 +301,18 @@ private long bumpTick() {
         return ++currentTick;
     }
 
+    /**
+     * Method to inherit all state from another constraints map.
+     * @param otherConstraintsMap the other constraints map
+     */
+    public void inheritFromOther(@Nonnull final ConstraintsMap otherConstraintsMap) {
```
alecgrieser, line 308: Just to make sure I understand the addition here: this is to prevent re-exploring references after translation. For example, in the select merge rule, if all we do is pull up grandchildren so that they are now the direct children of a `SelectExpression`, we don't want to re-explore all of those grandchildren (now children)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
```diff
@@ -0,0 +1,213 @@
...
+public class TranslateGraphTest extends FDBRecordStoreQueryTestBase {
+    @DualPlannerTest(planner = DualPlannerTest.Planner.CASCADES)
+    void rebaseGraphTest() {
+        final var cascadesPlanner = setUp();
+        var outerQun = fullTypeScan(cascadesPlanner.getRecordMetaData(), "RestaurantRecord");
+        final var explodeQun =
+                Quantifier.forEach(Reference.initialOf(
+                        new ExplodeExpression(FieldValue.ofFieldName(
+                                QuantifiedObjectValue.of(outerQun.getAlias(), outerQun.getFlowedObjectType()),
+                                "reviews"))));
```
alecgrieser, line 58: There are some helper methods in `FDBQueryGraphTestHelpers` that can make this a little more terse. I believe:

```java
forEach(new ExplodeExpression(fieldValue(outerQun, "reviews")));
```

Is equivalent. It calls `outerQun.getFlowedObjectValue()`, which I think is equivalent to the QOV you're constructing in-line here. (Unless there's a reason you're constructing the QOV manually that I'm missing.)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+        var graphExpansionBuilder = GraphExpansion.builder();
+        graphExpansionBuilder.addQuantifier(outerQun);
+        graphExpansionBuilder.addQuantifier(explodeQun);
+        graphExpansionBuilder.addPredicate(
+                new ValuePredicate(FieldValue.ofFieldName(QuantifiedObjectValue.of(outerQun.getAlias(),
+                        outerQun.getFlowedObjectType()), "name"),
+                        new Comparisons.SimpleComparison(Comparisons.Type.EQUALS, "name")));
```
alecgrieser, line 66: Likewise, this can be:

```java
fieldPredicate(outerQun, "name", new Comparisons.SimpleComparison(Comparisons.Type.EQUALS, "name"));
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
```diff
@@ -0,0 +1,213 @@
...
+import static com.apple.foundationdb.record.provider.foundationdb.query.FDBQueryGraphTestHelpers.fullScanExpression;
+import static com.apple.foundationdb.record.provider.foundationdb.query.FDBQueryGraphTestHelpers.fullTypeScan;
+import static com.apple.foundationdb.record.provider.foundationdb.query.FDBQueryGraphTestHelpers.reference;
+import static com.apple.foundationdb.record.provider.foundationdb.query.FDBQueryGraphTestHelpers.resultColumn;
+import static com.apple.foundationdb.record.query.plan.cascades.properties.ReferencesAndDependenciesProperty.referencesAndDependencies;
+
+public class TranslateGraphTest extends FDBRecordStoreQueryTestBase {
+    @DualPlannerTest(planner = DualPlannerTest.Planner.CASCADES)
+    void rebaseGraphTest() {
+        final var cascadesPlanner = setUp();
```
alecgrieser, line 52: Does this actually need the planner and `setUp()` call other than to get the meta-data?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
```diff
@@ -0,0 +1,213 @@
...
+        graphExpansionBuilder = GraphExpansion.builder();
+        graphExpansionBuilder.addQuantifier(outerQun);
+
+        var innerQun = fullTypeScan(cascadesPlanner.getRecordMetaData(), "RestaurantReviewer");
+        graphExpansionBuilder.addQuantifier(innerQun);
+
+        final var outerQuantifiedValue = QuantifiedObjectValue.of(outerQun.getAlias(), outerQun.getFlowedObjectType());
+        final var innerQuantifiedValue = QuantifiedObjectValue.of(innerQun.getAlias(), innerQun.getFlowedObjectType());
+
+        final var outerReviewerIdValue = FieldValue.ofFieldNames(outerQuantifiedValue, ImmutableList.of("review", "reviewer"));
```
alecgrieser, line 82: FWIW, this one could be:

```java
final var outerReviewerIdValue = fieldValue(outerQun, "review.reviewer");
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
```diff
@@ -0,0 +1,213 @@
...
+        final var explodeResultValue = QuantifiedObjectValue.of(explodeQun.getAlias(), explodeQun.getFlowedObjectType());
+        graphExpansionBuilder.addResultColumn(resultColumn(explodeResultValue, "review"));
+        outerQun = Quantifier.forEach(Reference.initialOf(
+                graphExpansionBuilder.build().buildSelect()));
+
+        graphExpansionBuilder = GraphExpansion.builder();
+        graphExpansionBuilder.addQuantifier(outerQun);
+
+        final var fullScanQuantifierForReviewer = forEach(fuseReference);
+        var innerQun = fullTypeScan(metaData, "RestaurantReviewer", fullScanQuantifierForReviewer);
```
alecgrieser, line 143: I take it this test and the other test are the same, except we get a diamond here because the inner quantifier re-uses part of the sub-graph from the outer, whereas they're separate in the other test case?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
```diff
@@ -62,80 +77,149 @@ public static List<? extends Reference> translateCorrelations(@Nonnull final Lis
...
+        final var references =
+                TopologicalSort.anyTopologicalOrderPermutation(partialOrder)
+                        .orElseThrow(() -> new RecordCoreException("graph has cycles"));
+
+        final var translationsCache =
+                Maps.<Reference, Reference>newIdentityHashMap();
+
+        for (final var reference : references) {
+            rebaseGraph(memoizer, translationMap, shouldSimplifyValues, reference, translationsCache,
+                    translationMap::getTarget);
```
alecgrieser: Instead of `translationMap::getTarget`, should this be:

```java
alias -> translationMap.getTargetOrDefault(alias, alias)
```

(Not that this matters for the `InferredTranslationMap`.) It also might be the case that we could use a `getTargetOrPreserve` helper method, or something

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
```diff
@@ -62,80 +77,149 @@ public static List<? extends Reference> translateCorrelations(@Nonnull final Lis
...
+        final var translatedFinalExpressionsBuilder = ImmutableList.<RelationalExpression>builder();
+        for (final var expression : reference.getAllMemberExpressions()) {
+            var allChildTranslationsSame = true;
+            final var translatedQuantifiersBuilder = ImmutableList.<Quantifier>builder();
+            for (final var quantifier : expression.getQuantifiers()) {
+                final var childReference = quantifier.getRangesOver();
+
+                // these must exist
+                final var translatedChildReference =
+                        translationsCache.getOrDefault(childReference, childReference);
```
alecgrieser: In the old code, this was:

```java
Verify.verify(translationsCache.containsKey(childReference));
translatedChildReference = translationsCache.get(childReference);
```

It this different now because we don't add non-translated references to the map at the end of the method? If that's the case, we may need to update this comment to state that the reference must have been translated, and so either it should have an entry in the map or we can retain the original reference as the translation didn't modify it

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
```diff
@@ -62,80 +77,149 @@ public static List<? extends Reference> translateCorrelations(@Nonnull final Lis
...
+
+                final var alias = quantifier.getAlias();
+                if (translatedChildReference == childReference &&
+                        !translationMap.containsSourceAlias(alias)) {
```
alecgrieser, line 163: It may be more straightforward to do something like:

```java
final var newAlias = quantifierAliasMapper.apply(alias);
```

Above, and then this check can be simplified to `alias.equals(newAlias)` (and below, we can just put in `newAlias`). I think that's the same, and we'd still get the side-effect in `translateCorrelations` if `alias` is in the source alias set (namely, an error thrown in the alias mapper)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
```diff
@@ -62,80 +77,149 @@ public static List<? extends Reference> translateCorrelations(@Nonnull final Lis
...
 
-        return refs.stream()
-                .map(ref -> Objects.requireNonNull(cachedTranslationsMap.get(ref)))
-                .collect(ImmutableList.toImmutableList());
+            final Reference translatedReference;
+            if (!translatedExploratoryExpressions.isEmpty()) {
+                translatedReference = memoizer.memoizeExploratoryExpressions(translatedExploratoryExpressions);
+            } else {
+                translatedReference = memoizer.memoizeFinalExpressions(translatedFinalExpressions);
+            }
```
alecgrieser: So, if we had a reference with a mix of final and exploratory members, this code will only return the translated exploratory members. Is that okay?

alecgrieser (approved):
I think this mostly looks good. There were a few smaller things that we could do later. The only conceptual thing is probably what happens if we translate an expression with a mix of exploratory and final expressions, which is I think the only comment I left that wasn't style (broadly defined)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
normen662: good catch. needs some rework

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/InferredTranslationMap.java
normen662: changed to `ToUniqueAliasesTranslationMap`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/TranslationMap.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
normen662: correct

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
normen662: no just for that.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TranslateGraphTest.java
normen662: yes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
normen662: changed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
normen662: removed comment

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/References.java
normen662: added a `memoizeExpressions(exploratoryExpressions, finalExpressions)` to `Memoizer` which at heart works exactly like `memoizeFinalExpressions(...)` but can deal with both kinds of expressions. The point is that the traversal is updated here, not that a reference is reused.

PR #3373, opened 2025-05-28
Add NULL operand type to binary rel op map

arnaud-lacurie:
SQL null literals are not falling under the `UNKNOWN` type which is mostly unused by the Cascades planner, but under the `NULL` type. 
This PR introduces `NULL` variants for the comparison operators.


hatyo (approved):
LGTM.

PR #3381, opened 2025-06-02
Introduce a new planner rewrite rule to facilitate in-lining function parameter values

### LINKED ISSUE #3380: Add planner rewrite rule to facilitate parameter inlining
The way the plan generator in the relational layer expands functions currently results in parameters being inserted as a join over a "values box" and another expression. See #3325.

However, to facilitate planning, we want to push the parameter values into the original expression. This would then allow us to remove the `range(1)` entirely, which simplifies the overall query needing to be planned. We should do this via a plan rewrite rule. In particular, the kind of rule we're thinking is one that:

1. Starts with a `SelectExpression`
2. Finds all of the children that are values boxes
3. Rewrites any projections or predicates referencing the values box(es) on the expression itself
4. Rewrites its immediate children so that they now reference the value box directly (if necessary)

This behaves similar to the predicate push down rule (see #3324), but it is more general, as parameters can be inlined into more kinds of child quantifiers (including existential and for-each-with-null-on-empty). But we'd still want to take an approach where:

* For any child that is a select expression, we can create a new select expression that exactly matches
* For any child that is not a select expression, we introduce a new select child that joins the values box with the existing child quantifier, and then rewrite the child over the new select

alecgrieser:
This adds the `DecorrelateValuesRule`. It operates on a single select expression to remove any values boxes (potentially introduced during function expansion). It will rewrite the immediate select expression's result value and predicates, and then it will rewrite its children so that each child now "owns" a copy of the values box(es) instead. By repeated application of this rule, the expression can be simplified so that there are no values boxes remaining.

Note: this does not currently take much care to make sure that each value is "pure". If there were an expression with a side-effect (like, say, `rand()`) that could not be pushed down, this would make some illegal transformations. Modifying this rule to exclude those values should be a relatively straightforward future modification, assuming one has a good oracle for function referential transparency.

This resolves #3380.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
@@ -24,7 +24,6 @@ import com.apple.foundationdb.record.query.plan.cascades.CorrelationIdentifier;
 import com.apple.foundationdb.record.query.plan.cascades.ExplorationCascadesRule;
 import com.apple.foundationdb.record.query.plan.cascades.ExplorationCascadesRuleCall;
 import com.apple.foundationdb.record.query.plan.cascades.ExploratoryMemoizer;
-import com.apple.foundationdb.record.query.plan.cascades.LinkedIdentitySet;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifiers;
 import com.apple.foundationdb.record.query.plan.cascades.Reference;
@@ -32,21 +31,27 @@ import com.apple.foundationdb.record.query.plan.cascades.expressions.LogicalFilt
 import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpressionVisitorWithDefaults;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.SelectExpression;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.TableFunctionExpression;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.AllOfMatcher;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.BindingMatcher;
 import com.apple.foundationdb.record.query.plan.cascades.matching.structure.TypedMatcherWithPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.QueryPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.properties.CardinalitiesProperty;
+import com.apple.foundationdb.record.query.plan.cascades.values.LiteralValue;
+import com.apple.foundationdb.record.query.plan.cascades.values.RangeValue;
+import com.apple.foundationdb.record.query.plan.cascades.values.StreamingValue;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 import com.apple.foundationdb.record.query.plan.cascades.values.translation.RegularTranslationMap;
 import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
+import com.apple.foundationdb.record.util.pair.NonnullPair;
+import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Streams;
 
 import javax.annotation.Nonnull;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -169,12 +174,12 @@ public class DecorrelateValuesRule extends ExplorationCascadesRule<SelectExpress
         return set1.stream().noneMatch(set2::contains);
     }
 
-    private static boolean correlatedToNone(@Nonnull RelationalExpression expression, @Nonnull Set<CorrelationIdentifier> ids) {
-        return emptyIntersection(expression.getCorrelatedTo(), ids);
+    private static boolean correlatedToNone(@Nonnull RelationalExpression expression, @Nonnull Set<CorrelationIdentifier> aliases) {
+        return emptyIntersection(expression.getCorrelatedTo(), aliases);
     }
 
-    private static boolean correlatedToNone(@Nonnull Quantifier qun, @Nonnull Set<CorrelationIdentifier> ids) {
-        return emptyIntersection(qun.getCorrelatedTo(), ids);
+    private static boolean correlatedToNone(@Nonnull Quantifier qun, @Nonnull Set<CorrelationIdentifier> aliases) {
+        return emptyIntersection(qun.getCorrelatedTo(), aliases);
     }
 
     @Override
@@ -186,50 +191,38 @@ public class DecorrelateValuesRule extends ExplorationCascadesRule<SelectExpress
         final SelectExpression selectExpression = call.get(root);
         final Set<CorrelationIdentifier> childAliases = Quantifiers.aliases(selectExpression.getQuantifiers());
         final List<? extends SelectExpression> valuesExpressions = call.getBindings().getAll(valuesExpressionMatcher);
-        final Set<SelectExpression> valueExpressionSet = new LinkedIdentitySet<>(valuesExpressions);
-        final Map<CorrelationIdentifier, SelectExpression> valuesByAlias = new HashMap<>();
+
+        final ImmutableMap.Builder<CorrelationIdentifier, SelectExpression> valuesByAlias = ImmutableMap.builderWithExpectedSize(valuesExpressions.size());
         final ImmutableMap.Builder<CorrelationIdentifier, Quantifier> qunsToPushDownBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
-        for (Quantifier.ForEach valueQunCandidate : valueQunCandidates) {
-            if (!correlatedToNone(valueQunCandidate, childAliases)) {
+        Streams.zip(valueQunCandidates.stream(), valuesExpressions.stream(), NonnullPair::of)
                 //
-                // Reject any quantifiers here that have correlations to their siblings
+                // Reject any quantifiers here that have correlations to their siblings.
                 // In theory, we could check the _expressions_, rather than the quantifier,
                 // but then we'd have to be careful not to carry along any expressions
-                // with correlations (which may be invalidated) when we pushed the quantifier down.
+                // with correlations (which may be invalidated) when we push the quantifier down.
                 //
                 // Note: simply re-creating a trimmed down reference without the correlated expressions
                 // leads to problems with the memoizer as it tries to re-use the reference with correlations
                 //
-                continue;
-            }
-            for (RelationalExpression expr : valueQunCandidate.getRangesOver().getAllMemberExpressions()) {
-                //
-                // Find the matched select expression(s) that correspond to this matched quantifier.
-                //
-                if (expr instanceof SelectExpression
-                        && valueExpressionSet.contains(expr)) {
-                    valuesByAlias.putIfAbsent(valueQunCandidate.getAlias(), (SelectExpression) expr);
-                    qunsToPushDownBuilder.put(valueQunCandidate.getAlias(), valueQunCandidate);
-                    break;
-                }
-            }
-        }
+                .filter(pair -> correlatedToNone(pair.getLeft(), childAliases))
+                .forEach(pair -> {
+                    Quantifier qun = pair.getLeft();
+                    SelectExpression valueExpression = pair.getRight();
+                    Verify.verify(qun.getRangesOver().containsExactly(valueExpression), "matched value box should be aligned with matched quantifier");
+                    valuesByAlias.put(qun.getAlias(), valueExpression);
+                    qunsToPushDownBuilder.put(qun.getAlias(), qun);
+                });
 
         final Map<CorrelationIdentifier, Quantifier> qunsToPushDownByAlias = qunsToPushDownBuilder.build();
         if (qunsToPushDownByAlias.isEmpty()) {
             // No actual values boxes here. Exit now
             return;
         }
-        if (qunsToPushDownByAlias.size() == selectExpression.getQuantifiers().size()) {
-            // All the quantifiers are values boxes. We can't push them around as this would
-            // leave no children at all. So we exit here.
-            return;
-        }
 
         //
         // Create a translation map, and use it to translate the result value of this select as well as any predicates
         //
-        final TranslationMap translationMap = createTranslationMapFromSelects(valuesByAlias);
+        final TranslationMap translationMap = createTranslationMapFromSelects(valuesByAlias.build());
         final Value newResultValue = selectExpression.getResultValue().translateCorrelations(translationMap, true);
         final List<QueryPredicate> newPredicates = selectExpression.getPredicates().stream()
                 .map(predicate -> predicate.translateCorrelations(translationMap, true))
@@ -239,7 +232,15 @@ public class DecorrelateValuesRule extends ExplorationCascadesRule<SelectExpress
         // Push the values box into each child for which it is relevant.
         //
         final PushValuesIntoVisitor visitor = new PushValuesIntoVisitor(qunsToPushDownByAlias, translationMap, call);
-        ImmutableList.Builder<Quantifier> newQuantifiersBuilder = ImmutableList.builderWithExpectedSize(selectExpression.getQuantifiers().size() - qunsToPushDownByAlias.size());
+        ImmutableList.Builder<Quantifier> newQuantifiersBuilder = ImmutableList.builderWithExpectedSize(Math.max(1, selectExpression.getQuantifiers().size() - qunsToPushDownByAlias.size()));
+        if (selectExpression.getQuantifiers().size() == qunsToPushDownByAlias.size()) {
+            //
+            // We're about to push down all the quantifiers. Introduce a range(1) box here to avoid creating a Select with no children
+            //
+            TableFunctionExpression rangeOneExpr = new TableFunctionExpression((StreamingValue) new RangeValue.RangeFn().encapsulate(ImmutableList.of(LiteralValue.ofScalar(1L))));
+            Quantifier newRangeQun = Quantifier.forEach(call.memoizeExploratoryExpression(rangeOneExpr));
+            newQuantifiersBuilder.add(newRangeQun);
+        }
         for (Quantifier qun : selectExpression.getQuantifiers()) {
             if (qunsToPushDownByAlias.containsKey(qun.getAlias())) {
                 //
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRuleTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRuleTest.java
@@ -583,6 +583,93 @@ class DecorrelateValuesRuleTest {
     }
 
     /**
+     * Trim uncorrelated (that is, extraneous) values boxes. Here, we start with a query like:
+     * <pre>{@code
+     * SELECT s.d
+     *    FROM values(x = @0, y = 42) AS v,
+     *       (SELECT b, c, d FROM T WHERE a = 42) AS s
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
```diff
@@ -823,9 +826,9 @@ private List<Cardinalities> fromQuantifiers(@Nonnull final List<? extends Quanti
         @Nonnull
         private Cardinalities fromQuantifier(@Nonnull final Quantifier quantifier) {
             if (quantifier instanceof Quantifier.Existential) {
-                return new Cardinalities(Cardinality.ofCardinality(1L), Cardinality.ofCardinality(1L));
+                return Cardinalities.exactlyOne();
             }
-
+            // Should this be adjusted to add .floor(1) to null-if-empty quantifiers?
```
normen662: Yes, I think!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+                // No expression was correlated. Put in the original quantifier
+                newQuantifiersBuilder.add(qun);
+            }
+        }
+        call.yieldExploratoryExpression(new SelectExpression(newResultValue, newQuantifiersBuilder.build(), newPredicates));
+    }
+
+    @SuppressWarnings("PMD.AvoidBranchingStatementAsLastInLoop")
+    @Nullable
+    private SelectExpression findSelectForQuantifier(@Nonnull Quantifier.ForEach qun,
```
normen662: We should try if possible to express this as a matcher. If we can't we should at least call this `findValuesForQuantifier(...)`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+                valuesByIdBuilder.put(qun.getAlias(), childSelect);
+                qunsByIdBuilder.put(qun.getAlias(), qun);
+            }
+        }
+        final Map<CorrelationIdentifier, SelectExpression> valuesById = valuesByIdBuilder.build();
+        if (valuesById.isEmpty()) {
+            // No actual values boxes here. Exit now
+            return;
+        }
+        if (valuesById.size() == selectExpression.getQuantifiers().size()) {
```
normen662: Isn't there a more general condition that you cannot find other quantifiers that are correlated to these ones so you cannot push at all?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+    private static boolean correlatedToNone(@Nonnull Quantifier qun, @Nonnull Set<CorrelationIdentifier> ids) {
+        return emptyIntersection(qun.getCorrelatedTo(), ids);
+    }
+
+    @Override
+    public void onMatch(@Nonnull final ExplorationCascadesRuleCall call) {
+        final SelectExpression selectExpression = call.get(root);
+        final List<? extends Quantifier.ForEach> valueQunCandidates = call.getBindings().getAll(forEach);
+
+        final Set<CorrelationIdentifier> selectChildQunIds = selectExpression.getQuantifiers().stream().map(Quantifier::getAlias).collect(ImmutableSet.toImmutableSet());
```
normen662:
```suggestion
        final Set<CorrelationIdentifier> childAliases = selectExpression.getQuantifiers().stream().map(Quantifier::getAlias).collect(ImmutableSet.toImmutableSet());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+        ImmutableMap.Builder<CorrelationIdentifier, SelectExpression> valuesByIdBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
+        ImmutableMap.Builder<CorrelationIdentifier, Quantifier> qunsByIdBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
```
normen662: Can you use `Quantifiers.aliasToQuantifierMap(...)`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+        return emptyIntersection(qun.getCorrelatedTo(), ids);
+    }
+
+    @Override
+    public void onMatch(@Nonnull final ExplorationCascadesRuleCall call) {
+        final SelectExpression selectExpression = call.get(root);
+        final List<? extends Quantifier.ForEach> valueQunCandidates = call.getBindings().getAll(forEach);
+
+        final Set<CorrelationIdentifier> selectChildQunIds = selectExpression.getQuantifiers().stream().map(Quantifier::getAlias).collect(ImmutableSet.toImmutableSet());
+        ImmutableMap.Builder<CorrelationIdentifier, SelectExpression> valuesByIdBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
```
normen662:
```suggestion
        ImmutableMap.Builder<CorrelationIdentifier, SelectExpression> valuesByAliasBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
normen662:
```suggestion
        ImmutableMap.Builder<CorrelationIdentifier, Quantifier> qunsByAliasBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+        ImmutableMap.Builder<CorrelationIdentifier, Quantifier> qunsByIdBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
+        for (Quantifier.ForEach qun : valueQunCandidates) {
+            @Nullable SelectExpression childSelect = findSelectForQuantifier(qun, selectChildQunIds);
+            if (childSelect != null) {
+                valuesByIdBuilder.put(qun.getAlias(), childSelect);
+                qunsByIdBuilder.put(qun.getAlias(), qun);
+            }
+        }
+        final Map<CorrelationIdentifier, SelectExpression> valuesById = valuesByIdBuilder.build();
```
normen662:
```suggestion
        final Map<CorrelationIdentifier, SelectExpression> valuesByAlias = valuesByIdBuilder.build();
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,375 @@
...
+        if (valuesById.size() == selectExpression.getQuantifiers().size()) {
+            // All the quantifiers are values boxes. We can't push them around as this would
+            // leave no children at all. So we exit here.
+            return;
+        }
+
+        //
+        // Create a translation map, and use it to translate the result value of this select as well as any predicates
+        //
+        final TranslationMap translationMap = createTranslationMapFromSelects(valuesById);
```
normen662: I think what we discussed earlier makes sense to me now, i.e. inline the values for the current expression and not immediately for the select your are potentially pushing into. That can be job of the next invocation. I think that's fine.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: I've updated all of the references to "IDs" to "aliases"

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: I'm not sure I understand the scenario here. If there are other quantifiers that are correlated to these, we can rewrite those other quantifiers so that the values box is now a child in that other quantifier's tree, I think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Updated to do more with matchers

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Updated the name from "ID" to "alias".

This can't quite use `Quantifiers.aliasToQuantifierMap` because it needs to only include the quantifiers that can be pushed down (for instance, it needs to exclude anything that is correlated to a sibling). I suppose I could put those quantifiers into a list or something, and then call `aliasToQuantifierMap` on the list, but that seems unecessary

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/CardinalitiesProperty.java
alecgrieser: Okay, I've added in the floor call. I'm a little concerned about the lack of test coverage here, as I don't think anything fails with this checked. I could move that out into a separate change

normen662: That's fine either way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
normen662: Nice, thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,381 @@
...
+        final SelectExpression selectExpression = call.get(root);
+        final Set<CorrelationIdentifier> childAliases = Quantifiers.aliases(selectExpression.getQuantifiers());
+        final List<? extends SelectExpression> valuesExpressions = call.getBindings().getAll(valuesExpressionMatcher);
+        final Set<SelectExpression> valueExpressionSet = new LinkedIdentitySet<>(valuesExpressions);
+        final Map<CorrelationIdentifier, SelectExpression> valuesByAlias = new HashMap<>();
+        final ImmutableMap.Builder<CorrelationIdentifier, Quantifier> qunsToPushDownBuilder = ImmutableMap.builderWithExpectedSize(valueQunCandidates.size());
+        for (Quantifier.ForEach valueQunCandidate : valueQunCandidates) {
+            if (!correlatedToNone(valueQunCandidate, childAliases)) {
+                //
+                // Reject any quantifiers here that have correlations to their siblings
```
normen662:
```suggestion
                // Reject any quantifiers here that have correlations to their siblings.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,381 @@
...
+                }
+            }
+        }
+
+        final Map<CorrelationIdentifier, Quantifier> qunsToPushDownByAlias = qunsToPushDownBuilder.build();
+        if (qunsToPushDownByAlias.isEmpty()) {
+            // No actual values boxes here. Exit now
+            return;
+        }
+        if (qunsToPushDownByAlias.size() == selectExpression.getQuantifiers().size()) {
```
normen662: This comment is to reiterate the question I had before. Isn't there a more general condition? If you find a values box that no one is correlated to, you cannot push that values box anywhere. So that values box should be excluded from the set of values to be decorrelated. In this particular case, none of the values boxes can be pushed anywhere, because there are obviously no places to push them to. The same, however, would be true if there was one uncorrelated select box which is underneath `selectExpression` and a bunch of values boxes. You still cannot push.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,381 @@
...
+    public void onMatch(@Nonnull final ExplorationCascadesRuleCall call) {
+        final List<? extends Quantifier.ForEach> valueQunCandidates = call.getBindings().getAll(valuesQunMatcher);
+        if (valueQunCandidates.isEmpty()) {
+            return;
+        }
+        final SelectExpression selectExpression = call.get(root);
+        final Set<CorrelationIdentifier> childAliases = Quantifiers.aliases(selectExpression.getQuantifiers());
+        final List<? extends SelectExpression> valuesExpressions = call.getBindings().getAll(valuesExpressionMatcher);
+        final Set<SelectExpression> valueExpressionSet = new LinkedIdentitySet<>(valuesExpressions);
+        final Map<CorrelationIdentifier, SelectExpression> valuesByAlias = new HashMap<>();
```
normen662: I guess you need a map and not a builder because of the `putIfAbsent(...)` call further down below. Can you make this a `LinkedHashMap`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,381 @@
...
+
+    private static <T> boolean emptyIntersection(@Nonnull Set<? extends T> set1, @Nonnull Set<? super T> set2) {
+        return set1.stream().noneMatch(set2::contains);
+    }
+
+    private static boolean correlatedToNone(@Nonnull RelationalExpression expression, @Nonnull Set<CorrelationIdentifier> ids) {
+        return emptyIntersection(expression.getCorrelatedTo(), ids);
+    }
+
+    private static boolean correlatedToNone(@Nonnull Quantifier qun, @Nonnull Set<CorrelationIdentifier> ids) {
```
normen662:
```suggestion
    private static boolean correlatedToNone(@Nonnull Quantifier qun, @Nonnull Set<CorrelationIdentifier> aliases) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -0,0 +1,381 @@
...
+
+    public DecorrelateValuesRule() {
+        super(root);
+    }
+
+    private static <T> boolean emptyIntersection(@Nonnull Set<? extends T> set1, @Nonnull Set<? super T> set2) {
+        return set1.stream().noneMatch(set2::contains);
+    }
+
+    private static boolean correlatedToNone(@Nonnull RelationalExpression expression, @Nonnull Set<CorrelationIdentifier> ids) {
```
normen662:
```suggestion
    private static boolean correlatedToNone(@Nonnull RelationalExpression expression, @Nonnull Set<CorrelationIdentifier> aliases) {
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/RangeConstraints.java
```diff
@@ -353,6 +353,29 @@ public RangeConstraints rebase(@Nonnull final AliasMap translationMap) {
     @Nonnull
     public RangeConstraints translateCorrelations(@Nonnull final TranslationMap translationMap,
                                                   final boolean shouldSimplifyValues) {
+        //
```
normen662, line 356: Just curious, how did you stumble over this problem?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Yes, that was the idea. Though I actually think it's unnecessary now. I'd initially written this to go through the values expressions and have it pick the first matched quantifier (and rejecting duplicates), but now it goes through the other way, and so we shouldn't see duplicate quantifier aliases. But also, I think I can re-write this with a zip anyway that accomplishes the same thing, so I've done that (and replaced the `HashMap` with an immutable map builder)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Okay, I see now. I think the answer is "no", we don't need to do that additional check. If a values box isn't correlated, then we don't have anywhere to push it, but that's fine because it can just be trimmed away. That happens naturally with the way the code is already structured. I've added two tests to this affect: in one of them, an entirely unreferenced values box gets trimmed away with no other modifications to the graph; in the second, a values box that is uncorrelated to any of its siblings but has references in predicates and the return value, and the appropriate modifications are made.

This check is more here as a safety check against having a select with _all_ values boxes to avoid creating a select with no children. However...I think we can actually replace that with `range(1)`. I've gone ahead and done that and then updated the relevant tests.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/RangeConstraints.java
alecgrieser: Some of the rewrites that happen as a part of translating the values boxes hit this. For example, if you had something like:

```sql
SELECT .... FROM value(x = 42) AS v, T
   WHERE T.a = v.x
```

Then when this gets translated into:

```sql
SELECT .... FROM T
   WHERE T.a = 42
```

The range comparison on `T.a` goes from being a deferred range to a compilable comparison. Some of the test cases I wrote hit that exact problem when it was trying to make its asserts

normen662 (approved):

PR #3382, opened 2025-06-02
(non)-partitioning properties for `Reference`s

normen662:
This PR provides a distinction among the properties that are being maintained for `References`. Before this PR, all properties were only used to derive `ExpressionPartition`s, i.e. they would define a set of properties all plans in the reference could be partitioned by. There are, however, also other properties which should be maintained in the same way but should be accessible individually for each expression in the reference.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
@@ -32,11 +32,25 @@ import java.util.Set;
  * @param <E> type parameter to indicate the type of expression this partition holds.
  */
 public class ExpressionPartition<E extends RelationalExpression> {
+    /**
+     * Grouping property map which holds all properties and their values that are common to all plans and therefore
+     * defining the partition.
+     */
     @Nonnull
     private final Map<ExpressionProperty<?>, ?> groupingPropertyMap;
+    /**
+     * Grouped property map which is a map from each individual plan in the partition to properties and their values
+     * for properties that are not defining the partition.
+     */
     @Nonnull
     private final Map<E, Map<ExpressionProperty<?>, ?>> groupedPropertyMap;
 
+    /**
+     * Constructor. Note that we do not defensively copy here as it is dependent on the use case if we need a copy or
+     * not. The caller needs to ensure that the maps being passed in are either immutable or owned.
+     * @param groupingPropertyMap grouping property map
+     * @param groupedPropertyMap grouped property map
+     */
     public ExpressionPartition(@Nonnull final Map<ExpressionProperty<?>, ?> groupingPropertyMap,
                                @Nonnull final Map<E, Map<ExpressionProperty<?>, ?>> groupedPropertyMap) {
         this.groupingPropertyMap = groupingPropertyMap;
@@ -54,8 +68,15 @@ public class ExpressionPartition<E extends RelationalExpression> {
     }
 
     @Nonnull
-    public <A> A getPropertyValue(@Nonnull final ExpressionProperty<A> expressionProperty) {
+    public <A> A getGroupingPropertyValue(@Nonnull final ExpressionProperty<A> expressionProperty) {
         return expressionProperty.narrowAttribute(Objects.requireNonNull(groupingPropertyMap.get(expressionProperty)));
+    }
+
+    @Nonnull
+    public <A> A getGroupedPropertyValue(@Nonnull final E expression,
+                                         @Nonnull final ExpressionProperty<A> expressionProperty) {
+        final var propertyMapForExpression = groupedPropertyMap.get(expression);
+        return expressionProperty.narrowAttribute(Objects.requireNonNull(propertyMapForExpression.get(expressionProperty)));
     }
 
     @Nonnull
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
@@ -37,50 +37,50 @@ public class ExpressionPartition<E extends RelationalExpression> {
      * defining the partition.
      */
     @Nonnull
-    private final Map<ExpressionProperty<?>, ?> groupingPropertyMap;
+    private final Map<ExpressionProperty<?>, ?> partitionPropertiesMap;
     /**
      * Grouped property map which is a map from each individual plan in the partition to properties and their values
      * for properties that are not defining the partition.
      */
     @Nonnull
-    private final Map<E, Map<ExpressionProperty<?>, ?>> groupedPropertyMap;
+    private final Map<E, Map<ExpressionProperty<?>, ?>> nonPartitioningPropertiesMap;
 
     /**
      * Constructor. Note that we do not defensively copy here as it is dependent on the use case if we need a copy or
      * not. The caller needs to ensure that the maps being passed in are either immutable or owned.
-     * @param groupingPropertyMap grouping property map
-     * @param groupedPropertyMap grouped property map
+     * @param partitionPropertiesMap property map common to the entire partition
+     * @param nonPartitioningPropertiesMap expression property map for expression-specific properties
      */
-    public ExpressionPartition(@Nonnull final Map<ExpressionProperty<?>, ?> groupingPropertyMap,
-                               @Nonnull final Map<E, Map<ExpressionProperty<?>, ?>> groupedPropertyMap) {
-        this.groupingPropertyMap = groupingPropertyMap;
-        this.groupedPropertyMap = groupedPropertyMap;
+    public ExpressionPartition(@Nonnull final Map<ExpressionProperty<?>, ?> partitionPropertiesMap,
+                               @Nonnull final Map<E, Map<ExpressionProperty<?>, ?>> nonPartitioningPropertiesMap) {
+        this.partitionPropertiesMap = partitionPropertiesMap;
+        this.nonPartitioningPropertiesMap = nonPartitioningPropertiesMap;
     }
 
     @Nonnull
-    public Map<ExpressionProperty<?>, ?> getGroupingPropertyMap() {
-        return groupingPropertyMap;
+    public Map<ExpressionProperty<?>, ?> getPartitionPropertiesMap() {
+        return partitionPropertiesMap;
     }
 
     @Nonnull
-    public Map<E, Map<ExpressionProperty<?>, ?>> getGroupedPropertyMap() {
-        return groupedPropertyMap;
+    public Map<E, Map<ExpressionProperty<?>, ?>> getNonPartitioningPropertiesMap() {
+        return nonPartitioningPropertiesMap;
     }
 
     @Nonnull
-    public <A> A getGroupingPropertyValue(@Nonnull final ExpressionProperty<A> expressionProperty) {
-        return expressionProperty.narrowAttribute(Objects.requireNonNull(groupingPropertyMap.get(expressionProperty)));
+    public <A> A getPartitionPropertyValue(@Nonnull final ExpressionProperty<A> expressionProperty) {
+        return expressionProperty.narrowAttribute(Objects.requireNonNull(partitionPropertiesMap.get(expressionProperty)));
     }
 
     @Nonnull
-    public <A> A getGroupedPropertyValue(@Nonnull final E expression,
-                                         @Nonnull final ExpressionProperty<A> expressionProperty) {
-        final var propertyMapForExpression = groupedPropertyMap.get(expression);
+    public <A> A getExpressionPropertyValue(@Nonnull final E expression,
+                                            @Nonnull final ExpressionProperty<A> expressionProperty) {
+        final var propertyMapForExpression = nonPartitioningPropertiesMap.get(expression);
         return expressionProperty.narrowAttribute(Objects.requireNonNull(propertyMapForExpression.get(expressionProperty)));
     }
 
     @Nonnull
     public Set<E> getExpressions() {
-        return groupedPropertyMap.keySet();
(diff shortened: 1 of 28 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
```diff
@@ -34,28 +32,55 @@
...
     @Nonnull
-    private final Map<ExpressionProperty<?>, ?> propertyValuesMap;
+    private final Map<ExpressionProperty<?>, ?> groupingPropertyMap;
+    /**
+     * Grouped property map which is a map from each individual plan in the partition to properties and their values
+     * for properties that are not defining the partition.
+     */
     @Nonnull
-    private final Set<E> expressions;
+    private final Map<E, Map<ExpressionProperty<?>, ?>> groupedPropertyMap;
```
alecgrieser: I'm not sure the grouping/grouped metaphor tracks here. I assume the intention here is to match what's going on with group-by expressions, but I'm not sure that that's intuitive. It almost seems like the `groupingPropertyMap` properties are "partition properties" (in that they define the partition) or "common properties", and the `groupedPropertyMap` properties are closer to "expression properties" (In that they are particular to the expression) or "individual properties"

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
```diff
@@ -34,28 +32,55 @@
...
 
-    public ExpressionPartition(final Map<ExpressionProperty<?>, ?> propertyValuesMap, final Collection<E> expressions) {
-        this.propertyValuesMap = ImmutableMap.copyOf(propertyValuesMap);
-        this.expressions = new LinkedIdentitySet<>(expressions);
+    /**
+     * Constructor. Note that we do not defensively copy here as it is dependent on the use case if we need a copy or
```
alecgrieser, line 49: Is the reason for this distinction basically effectively so that when we are doing roll-up, we can leave this map in a state where we can add in different expressions? Would the alternative be something like a "partition builder" that is mutable and then it could create an immutable version when built? Or is it more fundamental--like that `Reference`s want to be able to update their existing expressions map as new final expressions get added without needing to make a copy?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartitions.java
```diff
@@ -47,65 +45,96 @@ public static <E extends RelationalExpression> List<ExpressionPartition<E>> roll
...
+                            .entrySet()
+                            .stream()
+                            .filter(attributeEntry ->
+                                    rollupProperties.contains(attributeEntry.getKey()))
+                            .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+            rolledUpMap.compute(filteredPropertiesMap, (key, oldValue) -> {
+                if (oldValue == null) {
+                    return partition.getGroupedPropertyMap();
+                }
+                oldValue.putAll(partition.getGroupedPropertyMap());
```
alecgrieser: Suppose you have an expression partition set with two grouping properties, `A` and `B`, and one grouped property, `C`. If you roll up the partitions to one property, say `A`, then is the intention that the grouped property map in the rolled up `ExpressionPartition` would contain only the original grouped property `C`, or should it also have `B` as that property no longer defines the partition? As written, it looks like it drops `B`, but should it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
```diff
@@ -117,15 +120,32 @@ private E narrow(@Nonnull final RelationalExpression expression) {
...
         }
     }
 
+    @Nonnull
+    public Map<E, Map<ExpressionProperty<?>, ?>> getPropertiesMap() {
+        return propertiesMap;
+    }
+
+    @Nonnull
+    public Map<E, Map<ExpressionProperty<?>, ?>> computeNonGroupingPropertiesMap() {
```
alecgrieser: It feels like this should be `computeGroupedPropertiesMap` (or we should change all the "grouped" references to "non-grouped")

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
```diff
@@ -162,16 +182,49 @@ public void add(@Nonnull final RelationalExpression expression) {
...
+        }
+        final var groupingPropertyMap = groupingPropertyMapBuilder.build();
+        final var groupedPropertyMapBuilder = ImmutableMap.<ExpressionProperty<?>, Object>builder();
+        for (final var expressionProperty : trackedGroupedProperties) {
+            final var propertyValue = propertyMap.get(expressionProperty);
+            Verify.verify(propertyValue != null);
+            groupedPropertyMapBuilder.put(expressionProperty, propertyMap.get(expressionProperty));
+        }
+        final var groupedPropertyMap = groupedPropertyMapBuilder.build();
+        add(expression, groupingPropertyMap, groupedPropertyMap);
```
alecgrieser: If I'm following how this call to the other `add` works with this one, we create two maps here for the different properties, and then they get combined again together to form the combined map. It seems like either we should only have the grouped properties in the `propertiesMap` or at least we should try and avoid copying this data around twice

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ExpressionsPartitionMatchers.java
```diff
@@ -89,4 +90,38 @@ public static BindingMatcher<Collection<ExpressionPartition<RelationalExpression
     public static BindingMatcher<ExpressionPartition<RelationalExpression>> anyExpressionPartition() {
         return typed((Class<ExpressionPartition<RelationalExpression>>)(Class<?>)ExpressionPartition.class);
     }
+
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static <P extends Comparable<P>> BindingMatcher<ExpressionPartition<RelationalExpression>> argmin(@Nonnull final ExpressionProperty<P> expressionProperty,
```
alecgrieser, line 96: Looking forward to the fact that this is going to be a part of the select merge rule, does the binding matcher actually get used? That is, is the binding matcher logic sophisticated enough that we end up being able to use it?

alecgrieser (changes requested):
Mostly looks good. We could probably use a better PR title and description, as that is what will go into our release notes. It also seems like we could use more testing. I'm not sure if you looked at the coverage report: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15419532372/attempts/1#summary-43392883587 I haven't looked too closely, but there are a few files with regressions, and there's also the fact that the new `SelectCountProperty`, while it is technically covered in that it runs, there aren't any assertions that it returns a sensible value

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
normen662: The latter. I want to make sure the maps own their state. And since this sort of creating instances happens a lot, we want to make sure it's not always copied. In fact only the case where we create a partition from the initial reference needs a copy.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartitions.java
normen662: Good question. Maybe I should add it. Let me think about that one. It certainly is not needed right now but conceptionally you are right.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
normen662: _non-grouping_ is different that _grouped_ in current lingo. I will try to come up with better terms as indicated in the response to your other comment.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ExpressionsPartitionMatchers.java
normen662: I am using it in the select merge logic and it seems to work in the way envisioned.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
normen662: `partitioningProperties` and `nonPartitioningProperties` would work for me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartitions.java
normen662: Let me take that back. No, we have to drop `B` since there are no expressions I can associate that with. Even in SQL, when I roll up a
`SELECT a, b, SUM(c) FROM T GROUP BY a, b` to `SELECT a, sumc from (SELECT a, b, SUM(c) sumc FROM T) GROUP BY a`, the column `b` can only be aggregated over in the process but cannot get inherited.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
normen662: I made the add using precomputed values its own path that does not need to call the other `add(...)` anymore. That will avoid some copying. However, it's another path that updates the state.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartitions.java
alecgrieser: > we have to drop `B` since there are no expressions I can associate that with

I'm not sure I follow that. Isn't `B` associated with every expression in the original (finer) partition?

I'm thinking it's something like if you had a sql function `values_by_id` or something that allowed you to get a map linking row ID to a subset of columns, then you could roll up:

```sql
SELECT a, b, values_by_id(c) FROM T GROUP BY a, b
```

Into:

```sql
SELECT a, values_by_id(b, c) FROM T GROUP BY a
```

By taking the `b` value out of each group and then adding it to the `values_by_id` map

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
```diff
@@ -34,28 +32,55 @@
...
+        return nonPartitioningPropertiesMap;
+    }
+
+    @Nonnull
+    public <A> A getPartitionPropertyValue(@Nonnull final ExpressionProperty<A> expressionProperty) {
+        return expressionProperty.narrowAttribute(Objects.requireNonNull(partitionPropertiesMap.get(expressionProperty)));
+    }
+
+    @Nonnull
+    public <A> A getExpressionPropertyValue(@Nonnull final E expression,
```
alecgrieser: Should this be `getNonPartitioningPropertyValue`?  I don't necessarily dislike using "expression property" instead of non-partitioning, fwiw, if we're consistent

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartitions.java
```diff
@@ -47,65 +45,96 @@ public static <E extends RelationalExpression> List<ExpressionPartition<E>> roll
...
+            final Map<ExpressionProperty<?>, ?> filteredPropertiesMap =
+                    groupingPropertyMap
+                            .entrySet()
+                            .stream()
+                            .filter(attributeEntry ->
+                                    rollupProperties.contains(attributeEntry.getKey()))
+                            .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+            rolledUpMap.compute(filteredPropertiesMap, (key, oldValue) -> {
+                if (oldValue == null) {
+                    return partition.getNonPartitioningPropertiesMap();
```
alecgrieser: Sorry, probably should have asked about this earlier. But is there a problem here that the partition's map object is used directly here rather than creating a copy? Like, does that mean that we may get expressions from the other (non-rolled up) partitions in the finer-grained partitions.

Depending on the nature of the way these are used, I'm not actually sure if that's a problem or not. The `argmin`/`argmax` matcher could wind up returning an expression that's not in the overall expression partition, I guess, if the implementation just looks at the non-partitioning properties map. But this also depends on how long-lived the partitions provided are supposed to last

normen662: I changed the code in a way that I am copying the map when the initial map is created for the new key.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPartition.java
normen662: I tried to make it more consistent. There were some other variable names that still used `grouping` versus `grouped` as well which I got rid off.

PR #3386, opened 2025-06-09
Look for some code coverage

ScottDugas:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
```diff
@@ -4641,33 +4641,35 @@ protected CompletableFuture<Long> getRecordCountForRebuildIndexes(boolean newSto
...
-        MutableRecordStoreState writeOnlyState = recordStoreStateRef.get().withWriteOnlyIndexes(indexes.keySet().stream().map(Index::getName).collect(Collectors.toList()));
+        final IndexQueryabilityFilter indexQueryabilityFilter = new IndexQueryabilityFilter() {
+            @Override
+            public boolean isQueryable(@Nonnull final Index index) {
+                return !indexes.containsKey(index);
+            }
+
+            @Override
+            public int queryHash(@Nonnull final QueryHashKind hashKind) {
+                return indexes.hashCode();
```
alecgrieser: This should probably use `PlanHashables.iterablesPlanHash` or `PlanHashables.unorderedPlanHash`, though if I'm honest, it's unclear to me why this class is plan hashable at all. It's possible we should add a static `ObjectPlanHash` to serve as the basis here that we mix in with the iterables, like we do with the various key expressions

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
```diff
@@ -348,6 +348,10 @@ public IndexState needRebuildIndex(Index index, long recordCount, boolean indexO
         }
     }
 
+    /**
+     * We want to test that if the record count index has been modified, and requires rebuilding itself, we don't
```
alecgrieser, line 352: I believe you were looking into if this test would fail if there was a bug in the new approach (like, say, we used an index queryability filter of "true" or if the queryability filter were ignored). Were you able to validate that this test gives us coverage of that?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreOpeningTest.java
```diff
@@ -1033,6 +1037,113 @@ void testVersionChangedFlagAfterOpening() {
...
+                        @SuppressWarnings("deprecation")
+                        public CompletableFuture<Integer> checkUserVersion(final int oldUserVersion, final int oldMetaDataVersion, final RecordMetaDataProvider metaData) {
+                            assertEquals(0, oldUserVersion);
+                            assertEquals(initialVersion, oldMetaDataVersion);
+                            assertEquals(newMetaData, metaData);
+                            return CompletableFuture.completedFuture(1);
+                        }
+
+                        @Override
+                        public IndexState needRebuildIndex(final Index indexToRebuild, final long recordCount, final boolean indexOnNewRecordTypes) {
```
alecgrieser: FWIW, the plan with the user version checker is to remove the `needRebuildIndex` that takes a long and replace it with just the other variant, though we haven't formally deprecated the other method yet. Using this method is probably fine for this PR, and we could address that once the other method is deprecated.

Which, tbh, that class is already a bit of a mess, mainly for backwards compatibility reasons. It would be nice to be able to clean those all up at once

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
ScottDugas: Getting this to properly implement `queryHash` is tricky, because `Index` doesn't. Since we don't expose the query used anywhere, I opted to have this throw an exception.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreCountRecordsTest.java
ScottDugas: Yes, having the filter return a constant `true` caused this test to fail.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStoreOpeningTest.java
ScottDugas: ok, updated. I also converted this to an inner class, which I think makes the overall code more readable, and fixed a bug here, where it was putting with `index` rather than `indexToRebuild`.

alecgrieser (approved):

PR #3387, opened 2025-06-10
`EvaluationContext` and constraints for simplifications

normen662:
This PR is a direct precursor of enabling a `PredicateSimplification` rule in the `REWRITING` phase of planning.

This PR gives all planner rules including
- cascades planner rules
- simplification rules for `Value`s and `QueryPredicate`s
- computation rules for `Value`s and `QueryPredicate`s

the ability to access the current evaluation context and to produce `QueryPlanConstraint`s as part of the result of the rule when yielding.

Yielding has been extended to allow to yield `QueryPlanConstraint`s with the object that is being yielded.

For conciseness and readability I added a fluent-like API for rules to use:

```
call.yieldResultBuilder()
    .addConstraintsFrom(bindings.get(getMatcher()))
    .addConstraintsFrom(Sets.difference(allMajorOfMinors, retainedMajorOfMinors))
    .yieldResult(simplifiedPredicate);
```

Note that for the propagation to work, we need to make sure that in each individual rule, objects (i.e. normally in a tree of `QueryPredicate`s or `Value`s) that are not being preserved by the rule need to have their constraints preserved.

Example:

`COV(c58) IS NOT NULL AND FV(QOV(q, "x") = Literal(5)`

can be simplified to if `c58` is bound to `42`:

`TRUE AND FV(QOV(q, "x") = Literal(5)` with constraint that `COV(c58) IS NOT NULL` which is assigned to `TRUE`

that can be further simplified to:

`FV(QOV(q, "x") = Literal(5)` with constraint that `COV(c58) IS NOT NULL` which now needs to be assigned to the `ValuePredicate` as `TRUE` is gone. That means that `IdentityRule` must preserve the plan constraints that was previously assigned to the `TruePredicate`.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * BooleanWithConstraint.java
```
hatyo: Please fix.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
```diff
@@ -0,0 +1,83 @@
...
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import com.apple.foundationdb.record.query.plan.QueryPlanConstraint;
+import com.google.common.base.Verify;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A container that can carry a {@link QueryPlanConstraint}.
```
hatyo, line 29: I think you mean a "wrapper", container seems to indicate a collection type which is not the case here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/DefaultConstrained.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * BooleanWithConstraint.java
```
hatyo: ditto

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
```diff
@@ -0,0 +1,83 @@
...
+    @Nonnull
+    default T getUnconstrained() {
+        Verify.verify(!getConstraint().isConstrained());
+        return get();
+    }
+
+    boolean hasConstraint();
+
+    @Nonnull
+    QueryPlanConstraint getConstraint();
```
hatyo, line 47: it is a bit strange that the interface acts as decorator of a type, but at the same type, the thing it is decorated with can be null.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/DefaultConstrained.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return the wrapped object
+     */
+    @Nonnull
+    @Override
+    public T get() {
+        return object;
+    }
+
+    @Override
+    public boolean hasConstraint() {
```
hatyo: I am not sure this is necessary, factory methods of this class never produce null (if the object is without constraint, then it is created with `QueryPlanConstraint.noConstraint()`).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
```diff
@@ -96,39 +105,64 @@ public static Value simplify(@Nonnull final Value root,
...
     }
 
     /**
      * Main function that simplifies the given value using the {@link AbstractValueRuleSet} passed in.
      * @param current the {@link Value} to be simplified
+     * @param evaluationContext the evaluation context
      * @param aliasMap an alias map of equalities
      * @param constantAliases a set of aliases that are considered to be constant
      * @param ruleSet the rule set used to simplify the {@link Value} that is passed in
      * @return a new simplified {@link Value} of {@code root}
```
hatyo: While you're at it, can you please fix the documentation of the returned `Value`(s)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
```diff
@@ -293,25 +337,26 @@ private static <BASE extends TreeLike<BASE>> BASE computeCurrent(@Nonnull final
...
-    private static <BASE, ELEMENT> List<ELEMENT> collectChildrenElements(@Nonnull final Map<BASE, NonnullPair<BASE, List<ELEMENT>>> resultsMap,
-                                                                         @Nonnull final List<BASE> children) {
-        final var result = Lists.<ELEMENT>newArrayList();
-        for (final var child : children) {
-            @Nullable  final var childInResult = resultsMap.get(child);
-            if (childInResult != null) {
-                result.addAll(childInResult.getRight());
-            }
-        }
+    private static <BASE> BASE onResultsFunctionForSimplification(@Nonnull final Map<BASE, QueryPlanConstraint> constrainstsMap,
```
hatyo, line 340: Thanks for simplifying the generic definition of this method (and others), I find it much simpler to understand now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
```diff
@@ -444,58 +491,32 @@ public static <ARGUMENT> NonnullPair<QueryPredicate, List<QueryPlanConstraint>>
...
-            final var computedCurrent = computeCurrent(current, simplifiedChildren);
-            if (computedCurrent != current) {
-                if (resultsMap.containsKey(current)) {
-                    resultsMap.put(computedCurrent, NonnullPair.of(computedCurrent, resultsMap.get(current).getRight()));
-                } else {
-                    final var computedCurrentElements = collectChildrenElements(resultsMap, simplifiedChildren);
-                    final var computedCurrentPair = NonnullPair.of(computedCurrent, computedCurrentElements);
-                    resultsMap.put(computedCurrent, computedCurrentPair);
-                }
+                simplifiedChildren.add(simplifyWithReExploration(isRoot ? current : root, child, constraintsMap, ruleSet, ruleCallCreator));
```
hatyo, line 504: Is this correct? `current` can change in line `507`, however it is not matched against root as in `499`.

hatyo: Moreover, in the recursive calls, I think this condition will always be false. So we can probably simplify it to something like:

`simplifyWithReExploration(current, constraintsMap, ruleSet, ruleCallCreator)`
`simplifyWithReExplorationInternal(isRoot, current, constraintsMap, ruleSet, ruleCallCreator)`

Where:
- `simplifyWithReExploration` calls `simplifyWithReExplorationInternal` with `true` as the first argument. 
- `simplifyWithReExplorationInternal` calls itself recursively always with `false`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateSimplificationRule.java
```diff
@@ -0,0 +1,39 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.values.simplification.AbstractRuleCall;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Intermediate class that fixes the base type of the {@link AbstractRuleCall} to be {@link QueryPredicate}.
+ * @param <TYPE> the type of the object that this rule helps simplify
+ */
+@API(API.Status.EXPERIMENTAL)
+public abstract class QueryPredicateSimplificationRule<TYPE extends QueryPredicate> extends AbstractQueryPredicateRule<QueryPredicate, QueryPredicateSimplificationRuleCall, TYPE> {
```
hatyo, line 35: Can you remove the `QueryPredicateComputationRule` class?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateComputationRuleCall.java
```diff
@@ -51,12 +53,15 @@ public class QueryPredicateComputationRuleCall<ARGUMENT, RESULT> extends Abstrac
     public QueryPredicateComputationRuleCall(@Nonnull final PlannerRule<QueryPredicateComputationRuleCall<ARGUMENT, RESULT>, ? extends QueryPredicate> rule,
```
hatyo, line 51: should we remove this?

hatyo (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/DefaultConstrained.java
normen662: Removed that method.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
normen662: I changed that by modifying the semantics of `BooleanWithConstraint` accordingly. It's nicer now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/DefaultConstrained.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
normen662: A container class or container object can be wrapping anything, it does not necessarily have to be a collection. For instance in the preamble of `Optional` it states:

> A container object which may or may not contain a non-null value.

I am using the word container here in the same way.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Constrained.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/Simplification.java
normen662: I also need `root` in the calls as that is needed for the rules themselves. But then I will have `isRoot, root, current` where I know that `isRoot = root == current` creating a redundancy. I think I will leave it like it is unless I am overlooking a mistake.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateSimplificationRule.java
normen662: Discussed offline. Will leave it in here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/QueryPredicateComputationRuleCall.java
normen662: ditto. let's not do this right now.

PR #3388, opened 2025-06-10
Integrate rewrite rules

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -388,7 +388,7 @@ public Reference memoizeExploratoryExpressions(@Nonnull final Collection<? exten
                 expressions.forEach(expr ->
                         Debugger.withDebugger(debugger ->
                                 debugger.onEvent(InsertIntoMemoEvent.reusedExpWithReferences(expr, existingRefs))));
-                Verify.verify(existingReference != this.root);
+                // Verify.verify(existingReference != this.root);
```
normen662: What happened here?

normen662: Decided with @alecgrieser to reenable the check and deal with the fall out later.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -0,0 +1,163 @@
...
+        // Merge the quantifiers and predicates from each mergeable child expression
+        final var newQuantifiers = ImmutableList.<Quantifier>builder();
+        final var newPredicates = ImmutableList.<QueryPredicate>builder();
+        final var translationBuilder = TranslationMap.regularBuilder();
+        boolean atLeastOneQuantifierIsMergeable = false;
+        for (final var quantifier : selectExpression.getQuantifiers()) {
+            final var alias = quantifier.getAlias();
+            final var childSelectExpression = mergeableChildren.get(alias);
+            if (childSelectExpression != null &&
+                    !quantifier.narrow(Quantifier.ForEach.class).isNullOnEmpty()) {
```
alecgrieser: In theory, I think we could push this check into the matchers. We'd need to require that the `quantifiersMatcher` was only on `forEachWithoutNullOnEmpty`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -0,0 +1,163 @@
...
+                                childSelectExpression.getResultValue());
+                atLeastOneQuantifierIsMergeable = true;
+            } else {
+                //
+                // Not mergeable. Grab all final expressions and create a new reference.
+                //
+                final var childReference = quantifier.getRangesOver();
+                final var newChildReference =
+                        call.memoizeFinalExpressionsFromOther(childReference, childReference.getFinalExpressions());
+                newQuantifiers.add(quantifier.overNewReference(newChildReference));
```
alecgrieser, line 141: Is there a reason this can't just copy over the original quantifier directly? Why does it need to make a copy?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -0,0 +1,163 @@
...
+        boolean atLeastOneQuantifierIsMergeable = false;
+        for (final var quantifier : selectExpression.getQuantifiers()) {
+            final var alias = quantifier.getAlias();
+            final var childSelectExpression = mergeableChildren.get(alias);
+            if (childSelectExpression != null &&
+                    !quantifier.narrow(Quantifier.ForEach.class).isNullOnEmpty()) {
+                //
+                // Mergeable. Add the child quantifiers in the old one's place and scoop up any predicates.
+                //
+                newQuantifiers.addAll(childSelectExpression.getQuantifiers());
```
alecgrieser, line 128: Do we think that it's potentially the case that there are multiple quantifiers with the same alias, and we need to be careful when pulling them up that we don't get a clash? I know we hit some issues with that before, though it sounds like it was at least proximally fixed by changes to the way compilable functions worked to not re-use aliases. Do we still need to worry about that here? Or is the plan more like "let's not add handling for that until we are more confident that we need it"?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -201,6 +202,40 @@ public static RangeValue fromProto(@Nonnull final PlanSerializationContext seria
...
+        if (obj == this) {
+            return true;
+        }
+        if (!(obj instanceof RangeValue)) {
+            return false;
+        }
+        RangeValue otherRange = (RangeValue) obj;
+        return beginInclusive.equals(otherRange.beginInclusive)
+                && endExclusive.equals(otherRange.endExclusive)
+                && step.equals(otherRange.step);
```
alecgrieser: Should this be:

```java
return semanticEquals(other, AliasMap.emptyMap());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
```diff
@@ -201,6 +202,40 @@ public static RangeValue fromProto(@Nonnull final PlanSerializationContext seria
...
+    }
+
+    @Override
+    public int hashCode() {
+        return hashCodeWithoutChildren();
```
alecgrieser: Should this be `semanticHashCode()`? I think that's what some other values do in similar circumstances

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java
```diff
@@ -27,9 +27,9 @@
 
 import javax.annotation.Nonnull;
 
-import static com.apple.foundationdb.record.query.plan.cascades.properties.ExpressionCountProperty.selectCount;
 import static com.apple.foundationdb.record.query.plan.cascades.properties.ExpressionCountProperty.tableFunctionCount;
 import static com.apple.foundationdb.record.query.plan.cascades.properties.PredicateHeightProperty.predicateHeight;
+import static com.apple.foundationdb.record.query.plan.cascades.properties.SelectCountProperty.selectCount;
```
alecgrieser: Is there a reason to prefer the select count you added to the one I did? The main thing I'd say about mine is that it is more extensible to other expression types, like, for example, table function expressions

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveRangeOneRule.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.List;
+
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.AnyMatcher.any;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.QuantifierMatchers.forEachQuantifierOverRef;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ReferenceMatchers.members;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.RelationalExpressionMatchers.selectExpression;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.RelationalExpressionMatchers.tableFunctionExpression;
+
+@SuppressWarnings("PMD.TooManyStaticImports")
+public class RemoveRangeOneRule extends ExplorationCascadesRule<SelectExpression> {
```
alecgrieser, line 45: I suppose it doesn't hurt to keep this, but we should consider if we want to remove it, as we should hopefully be dealing with most of these through values decorrelation any way. It probably doesn't hurt to keep it for now and we can consider how it goes as we integrate these rules in.

If we do keep it, the matcher logic could probably be improved to do more in the matcher and less in `onMatch`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
```diff
@@ -109,8 +109,14 @@ public RuleTestHelper(@Nonnull CascadesRule<? extends RelationalExpression> rule
...
     @Nonnull
     private TestRuleExecution run(RelationalExpression original) {
-        ensureStage(PlannerStage.CANONICAL, original);
-        Reference ref = Reference.ofExploratoryExpression(PlannerStage.CANONICAL, original);
+        final var copiedOriginal =
+                Iterables.getOnlyElement(References.rebaseGraphs(ImmutableList.of(Reference.initialOf(original)),
+                        Memoizer.noMemoization(PlannerStage.INITIAL), new ToUniqueAliasesTranslationMap(), false)).get();
+        ensureStage(PlannerStage.CANONICAL, copiedOriginal);
+        if (rule instanceof ImplementationCascadesRule) {
+            preExploreForRule(copiedOriginal, false);
```
alecgrieser, line 165: Is there any reason not to call this on every test case (not just for implementation rules)? That is, if we expect in real life that `finalize` has always been called on all children and that all grand children have been pruned (which I believe is what this ensures), then should we just always call it here?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRuleTest.java
```diff
@@ -259,7 +260,7 @@ void canPushDownToMultipleChildren() {
                 baseQun, List.of("a", "b", "c"),
                 new ConstantPredicate(true)
         );
-        Reference lowerRef = Reference.ofExploratoryExpressions(PlannerStage.CANONICAL, ImmutableSet.of(lower1, lower2));
+        Reference lowerRef = Reference.ofFinalExpressions(PlannerStage.CANONICAL, ImmutableSet.of(lower1, lower2));
```
alecgrieser: Why was this changed from exploratory to final expressions? Shouldn't the predicate push down rule only be pushing down into exploratory expressions?

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/SqlFunctionsTest.java
```diff
@@ -77,4 +87,9 @@ public void queryNestingFunctionWithParameterLiterals() throws Exception {
         statement.execute("select * from f2(43) options (log query)");
         Assertions.assertTrue(logAppender.lastMessageIsCacheHit());
     }
+
+    @Test
+    void queryThreeLevelsDeep() throws Exception {
+        statement.execute("select * from f3(42, 'a') where col2a > col2b");
```
alecgrieser: At present, this test is a little half baked. I added it to test out some of the perf problems we were hitting. Perhaps we should remove it for now, or we should make it more precise

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -30,6 +30,7 @@
  * Note: Use {@link MaintainYamlTestConfig} using {@link YamlTestConfigFilters#CORRECT_EXPLAIN_AND_METRICS} or similar
  *       to correct explain strings and/or planner metrics. That annotation works both on class and on method level.
  */
+@MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
```
alecgrieser: We should back this out, as well as the YAML changes

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRuleTest.java
```diff
@@ -0,0 +1,1037 @@
...
+                .addPredicate(fieldPredicate(unionQun, "b", new Comparisons.ValueComparison(Comparisons.Type.EQUALS, fieldValue(explodeGQun, "two"))))
+                .build().buildSelect();
+
+        testHelper.assertYields(select, merged);
+    }
+
+    @Nonnull
+    private static Quantifier extractG(final Quantifier t3) {
+        return forEach(new ExplodeExpression(fieldValue(t3, "g")));
+    }
```
alecgrieser: This would be a pretty good candidate for a helper function that could go in `FDBQueryGraphTestHelpers`. I'm thinking something like a method like `explodeField` that takes a quantifier and field and then returns a quantifier over that field exploded

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRuleTest.java
```diff
@@ -0,0 +1,1037 @@
...
+        testHelper.assertYields(mergeL1Sigma2, mergeS3);
+
+        // Validate: last form is final
+
+        testHelper.assertYieldsNothing(mergeS3, true);
+    }
+
+    /**
+     * Test the case where we have multiple variants underneath a reference. We create additional variants
+     * for each such element at a higher level.
```
alecgrieser: We should probably update this to say that we only select the variants with the fewest select boxes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -205,11 +205,20 @@ public ExpressionPropertiesMap<? extends RelationalExpression> getPropertiesMap(
...
         Verify.verify(finalMembers.size() == 1);
+        advancePlannerStageUnchecked(newStage);
+    }
+
+    /**
+     * Advance the planner stage of this reference to the given one. This method cleans up internal state of this
+     * reference such that exploration in a new phase can start again.
+     * @param newStage the new stage we should advance to
+     */
+    void advancePlannerStageUnchecked(@Nonnull final PlannerStage newStage) {
```
alecgrieser, line 217: It might be good to mark this as `@VisibleForTesting` even though its package-private.

I'm also not sure I totally see why the unchecked-ness is necessary here. It seems like this is mainly here to deal with the test cases where there are multiple variants, but if those test cases put their references into the canonical phase (as they did before), then they shouldn't need to be advanced. I think as long as we ran the finalize expression rule on them to make sure they copy over (all) of their variants from exploratory to final, it should just work? But maybe I'm missing some subtlety

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
normen662: Because we don't want anybody else use that reference as well. It's the one thing that makes an implementation rule special. Now nobody else has another quantifier over this reference, thus you can prune the reference in the next step.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
normen662: I am fairly confident after a lot of thinking about this problem that if you run into this problem, the original graph was illegal to start with. As discussed offline, I think a rebase that I did earlier got rid of the problem I had found initially.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
normen662: yes, it should be!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
normen662: Yes! It is the same here though as this is a leaf value but still.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java
normen662: The one I added is maintained by the `ExpressionPropertiesMap`. So it's more efficient to grab it from the children.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
normen662: So all implementation rules will get you perfectly unshared singular (no variants) graphs that you can match your expected with. That is not the case for an exploration rule. If we run `FinalizeExpressionsRule` for the subgraph for exploring rules, you will get changes to the graph further down due to that and your assert won't work anymore. Again, for implementation rules that is not a problem as their final (yielded) expressions own the subgraph.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRuleTest.java
normen662: Yes, and it does because the `ensurePlanStage(...)` does that. The bug I actually have in here is that the method should be called using `PlannerStage.INITIAL`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRuleTest.java
normen662: Done exactly as suggested.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRuleTest.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
normen662: I added `@VisibleForTesting`. A bunch of other fragments in `RuleTestHelper` create as `INITIAL` and they depend on the stage getting advanced. I think we should have all the fragments created like that and all the steps that normally happen in the planner to also happen through the framework here.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/SqlFunctionsTest.java
normen662: removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
normen662: done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RangeValue.java
normen662: yes

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveRangeOneRule.java
normen662: Unless you object, let's just keep it here and in its current shape. We can talk offline if the decorrelation rule itself will make this superfluous. If that's the case we should remove this one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
normen662: There already is a matcher for this: `forEachQuantifierWithoutDefaultOnEmptyOverRef(...)`. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
alecgrieser: How is that linked to needing to skip the verification though? For the other fragments marked as `INITIAL`, as long as they don't have multiple variants, shouldn't we be able to just call the normal `advancePlannerStage` method?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java
alecgrieser: I guess embedded in my question was the assumption that we'd make whichever one we chose also be maintained by the expression properties map. I was more asking about whether we should align the select count and table function count properties so they don't have separate implementations

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
alecgrieser: Hm, okay. I'm not sure I totally see the way that all connects here, but this seems fine to me. Is the problem more that (assuming we're also running the finalize rules on the expected list) that because of the way the memo structure works, you can have nodes in the two DAGs being compared where the expected DAG is one subset of the expected structure? And it's _that_ sharing that means that we start getting asserts that don't work?

alecgrieser: Okay, we talked offline, and I think I understand this now. It has to do with extra variants that are introduced to expressions, both during the initial call to `FinalizeExpressionsRule` and during the actual process of planning. Those extra variants don't exist in the expected graph, so we can't get a match.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
alecgrieser: Okay, we discussed this a bit off line, and the desire here is to start with an entirely `INITIAL` graph. We could have a mix, but it gets messy and a little ugly, and so relaxing the constraint here to allow for multiple final variants in an  `INITIAL` reference is the most straightforward way to achieve this.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/PredicatePushDownRuleTest.java
alecgrieser: Okay, this makes sense, given: https://github.com/FoundationDB/fdb-record-layer/pull/3388#discussion_r2144745856

alecgrieser (approved):

Issue #3206, opened 2025-02-27
Scanning an aggregate index can result in an empty, non-end continuation


alecgrieser:
If you have an ungrouped aggregate index (though I think this also works on a grouped aggregate index if your scan boundary aligns with the group), then when the aggregate index is scanned, we can end up creating a malformed continuation. In particular, the underlying scan will ultimately be based on a `KeyValueBase.Continuation`. This continuation is based off of the final key with a prefix removed that is common to all elements in the scan range. (That way, it doesn't have to be repeated with continuation.) For an ungrouped aggregate index, there's only a single key, and that key is equal to the scan range prefix. That means that the serialized continuation will be the empty string, although that doesn't mean the end in this case because this will be associated with the value from the aggregate scan.

So, for example, you might have a value/continuation pair like:

```
res1: value = 100, continuation = []
res2: no_next_reason=EXHAUSTED, continuation = []
```

Which is bad, especially as a reader who is doing something like reading one value at a time and then resuming the scan from a continuation can get into an infinite loop if the empty continuation is interpreted as a "start" continuation.

We noticed this in some of our yaml tests, as they do not expect to get a "begin" continuation back from running a query, which can happen if one of these lives in the plan stack. For example, a plan like:

```
AISCAN(T2_I1 <,> BY_GROUP -> [_0: VALUE:[0]]) | MAP (_ AS _0) | ON EMPTY NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)
```

(A plausible plan for a query like `select count(*) from t2` with an ungrouped count index `T2_I1`--see: [`aggregate-empty-table.yamsql`](https://github.com/FoundationDB/fdb-record-layer/blob/8a3c17f81a4a9f1423887212f72fdcb94793e2de/yaml-tests/src/test/resources/aggregate-empty-table.yamsql#L334).) If the underlying scan returns a single result, then the `map` plan passes the continuation along, and the `ON EMPTY` plan's `OrElseCursor` returns an empty continuation as well, though it is not marked as an `END` continuation. This results in machinery in the Relational Layer eventually producing a new `BEGINNING` continuation, which can result in infinite looping.

PR #3393, opened 2025-06-12
Re-enable force continuation tests in Yaml

pengpeng-lu:
- Because of https://github.com/FoundationDB/fdb-record-layer/issues/3096, some force continuation tests cannot be run before, re-enable these tests.
- Cleans up tests that care about versions < 4.1.9.0


yaml-tests/src/test/resources/aggregate-index-tests.yamsql
```diff
@@ -162,9 +162,6 @@ test_block:
     -
        - query: select max(col2) from T1 use index (mv8);
        - maxRows: 1
-       # Cannot use FORCE_CONTINUATIONS with older versions due to: https://github.com/FoundationDB/fdb-record-layer/issues/3096
-       # (Extra values being produced after exhausting source of an aggregate cursor)
-       # Can remove once we do not care about backwards compatibility before 4.1.9.0
```
alecgrieser, line 167: I think we either want to retain this comment or remove the conditional behavior. Otherwise, it'll be unclear to us why we have the version check when we look at this.

At present, it actually looks like we're no longer testing against pre-4.1.9.0 versions in our mixed mode testing. So we could probably just remove this conditional behavior. In fact, it looks like this is actually the same query as the one before, and we could probably clean this up by: (1) removing this test case entirely and (2) removing the `supported_version` directive from the previous query

yaml-tests/src/test/resources/aggregate-index-tests.yamsql
```diff
@@ -278,23 +269,25 @@ test_block:
       - result: [{COL3: 1, S: 1}, {COL3: 2, S: 2}, {COL3: 100, S: 1}, {COL3: 200, S: 2}]
     -
       - query: select col3, sum(col2) as s from t2 use index (mv9) where col1 = 1 group by col1, col3 order by col3 asc;
-      # Cannot use FORCE_CONTINUATIONS with older versions due to: https://github.com/FoundationDB/fdb-record-layer/issues/3096
-      # (Extra values being produced after exhausting source of an aggregate cursor)
-      # Can remove once we do not care about backwards compatibility before 4.1.9.0
-      - maxRows: 0
-      - result: [{COL3: 1, S: 1}, {COL3: 2, S: 2}, {COL3: 100, S: 1}, {COL3: 200, S: 2}]
+      - maxRows: 1
```
alecgrieser: I don't think we need to add `maxRows: 1` here. In general, we can just remove the `maxRows: 0` clause, and then the framework will generate the force-continuations stuff for us. In this case, this is another instance where we already have such a test case (in this case, the immediately previous query), and we should be able to just remove the `supported_version` line from there and then delete this now extraneous test case

yaml-tests/src/test/resources/catalog.yamsql
```diff
@@ -62,8 +62,9 @@ test_block:
       - query: select sum(cnt) from (select count(*) as cnt, template_name, template_version from schemas
           group by template_name, template_version having template_name = 'TEST_TEMPLATE_1') as t;
       - explain: "AISCAN(TEMPLATES_COUNT_INDEX [EQUALS promote(@c29 AS STRING)] BY_GROUP -> [_0: KEY:[0], _1: KEY:[1], _2: VALUE:[0]]) | MAP (_._2 AS CNT, _._0 AS TEMPLATE_NAME, _._1 AS TEMPLATE_VERSION) | MAP (_ AS _0) | AGG (sum_l(_._0.CNT) AS _0) | ON EMPTY NULL | MAP (_._0._0 AS _0)"
-      - maxRows: 0 # Disable force continuations because of empty continuation due to: https://github.com/FoundationDB/fdb-record-layer/issues/3206
+      - maxRows: 1
```
alecgrieser: This one claims to be from #3206, which I think is one that you're currently working on. If that fix isn't actually required, then that's interesting. This may have been miscategorized, or maybe there are interactions with #3206 and #3219 or something

alecgrieser (changes requested):
I left some comments on some indicative examples, but in general, I think we could clean these up mostly by removing things. There are a bunch of test cases that require 4.1.9.0 or have conditional behavior on that version. I think removing the conditional behavior and `supported_version` things and then deleting duplicative tests is closer to the kind of clean-up we want

yaml-tests/src/test/resources/catalog.yamsql
pengpeng-lu: yes maybe it was miscategorized?

yaml-tests/src/test/resources/groupby-tests.yamsql
```diff
@@ -154,127 +118,34 @@ test_block:
...
-      - initialVersionAtLeast: 4.1.9.0
     -
       - query: select AVG(x.col2) from (select col1,col2 from t1) as x group by x.col1;
       - explain: "ISCAN(I1 <,>) | MAP (_.COL1 AS COL1, _.COL2 AS COL2) | MAP (_ AS _0) | AGG (avg_l(_._0.COL2) AS _0) GROUP BY (_._0.COL1 AS _0) | MAP (_._1._0 AS _0)"
-      - maxRows: 0 # Disable force_continuations. Can be removed when we longer care about testing with versions older than 4.1.9.0
-      - result: [{3.0}, {9.5}]
-    -
-      # Same as above but simulates force_continuations mode. Can be removed after we no longer care about compatibility with versions before 4.1.9.0
-      - query: select AVG(x.col2) from (select col1,col2 from t1) as x group by x.col1;
       - maxRows: 1
```
alecgrieser: I think we want to take out the `maxRows: 1` case here in favor of just leaving in the full result and relying on `FORCE_CONTINUATIONS` mode for the max-rows-1 case

alecgrieser (changes requested):
Sorry for the delay in getting back to this. Got distracted by some other things and am just getting back to it now. It mostly looks good now except for one query which still has `maxRows: 1`. There are some conflicts, though, and it looks like the metrics files will probably need to be regenerated.

yaml-tests/src/test/resources/groupby-tests.yamsql
pengpeng-lu: good catch! removed.

PR #3395, opened 2025-06-12
Adopt dataType for Literals and preparedValues

g31pranjal:
This PR moves away from using `java.sql.Types` in the internals of `fdb-relational-core`, while still maintaining JDBC-compliance for external `getType()`-like operations. It is replaced with `DataType`, which provides richer and more diverse type information available for downstream operations. Notably, we can now have types like `Uuid`, `Enum`, `Version` expressed natively. There are several consequences of this:
- It eliminates SQLType <-> RecordType conversions and instead relies on DataType <-> RecordType conversions, which were already being used for DDL work. This is more consistent as it is used throughout the codebase.
- `StructMetadata` and `ArrayMetadata` wrap `DataType` as the source of truth for their type info. However, they still map to `sql.Types` externally. Moreover, they now expose `getRelationalDataType()` that returns `DataType` to the external users for more transparency. 
- Aligning to the above changes, `grpc` protobufs now have a `Type` enum, moving away from `sql.Types` too. This was needed to support `struct` and `array` operations homogenously across embedded and JDBC drivers. It keeps the old types too for backward compatibility, which I think should be removed in future to simplify conversions.


fdb-relational-server/src/testFixtures/java/com/apple/foundationdb/relational/server/ServerTestUtil.java
```diff
@@ -21,13 +21,12 @@
 package com.apple.foundationdb.relational.server;
 
 import com.apple.foundationdb.annotation.API;
-
 import io.prometheus.client.CollectorRegistry;
+import org.apache.logging.log4j.LogManager;
```
hatyo: please revert.

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
```diff
@@ -235,49 +234,33 @@ public Response execute(String database, String schema, String sql, List<Paramet
 
     private static void addPreparedStatementParameter(RelationalPreparedStatement relationalPreparedStatement,
                                                       Parameter parameter, int index) throws SQLException {
```
hatyo: Can you please add `@Nonnull` annotations where applicable to this method signature?

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/jdbc.proto
```diff
@@ -106,6 +106,7 @@ message Parameter {
   // Reuse the column type here. It has much of what we need carrying
   // across parameters as well as null support, etc.
   Column parameter = 2;
+  Type type = 3;
```
hatyo, line 109: Can we remove the `java_sql_types_code` above? the mapping between `Type` and java SQL type is hardcoded in the codebase, and so it is redundant here, right?

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/column.proto
```diff
@@ -26,20 +26,39 @@ option java_multiple_files = true;
...
-  int32 nullable = 4;
+  bool nullable = 4;
   // Indicates a column that isn't part of the DDL for a query, but is part of the returned
   // tuple (and therefore necessary to keep so that our positional ordering is intact)
   bool phantom = 5;
   // If this column has a Struct or an Array, their metadata is here.
   oneof metadata {
     ListColumnMetadata structMetadata = 6;
     ColumnMetadata arrayMetadata = 7;
+    EnumMetadata enumMetadata = 8;
```
hatyo, line 59: please adapt the comment at line 55.

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/column.proto
```diff
@@ -26,20 +26,39 @@ option java_multiple_files = true;
 option java_package = "com.apple.foundationdb.relational.jdbc.grpc.v1.column";
 option java_outer_classname = "ColumnProto";
 
+enum Type {
+  NO_TYPE = 0;
```
hatyo: UNKNOWN?

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -73,7 +77,7 @@ public class TypeConversion {
      * @return {@link RelationalStruct} instance pulled from <code>resultSet</code>
      * @throws SQLException If failed get of <code>resultSet</code> metadata.
      */
-    // Manipulation of protobufs. Exploits package-private internal ottributes of {@link RelationalStructFacade}.
+    // Manipulation of protobufs. Exploits package-private internal attributes of {@link RelationalStructFacade}.
```
hatyo: Please add to the method documentation.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/DataType.java
```diff
@@ -162,6 +167,52 @@ public interface Named {
         String getName();
     }
 
+    /**
+     * Trait representing composite type.
+     */
+    public interface CompositeType {
```
hatyo, line 173: Can you remove this? Why not using `equals()` instead?

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/ArrayMetaData.java
```diff
@@ -59,4 +61,7 @@ default StructMetaData getElementStructMetaData() throws SQLException {
      * @throws SQLException if the type of the column is not an array, or if something else goes wrong.
      */
     ArrayMetaData getElementArrayMetaData() throws SQLException;
+
+    @Nonnull
+    DataType.ArrayType getRelationalDataType() throws SQLException;
```
hatyo: I think this is returning (this) as a relational data type, correct? If so, can we name this method differently to something like: `asRelationalType`.
Also, can you please document it?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/EmbeddedRelationalStruct.java
```diff
@@ -41,95 +38,87 @@ static RelationalStructBuilder newBuilder() {
...
 
-        final List<FieldDescription> fields = new ArrayList<>();
+        final List<DataType.StructType.Field> fields = new ArrayList<>();
 
         final List<Object> elements = new ArrayList<>();
 
         @Override
         public EmbeddedRelationalStruct build() {
-            return new ImmutableRowStruct(new ArrayRow(elements.toArray()), new RelationalStructMetaData(fields.toArray(new FieldDescription[0])));
+            // Ideally, the name of the struct should be something that the user can specify, however, this actually
```
hatyo, line 47: There is already an open issue about this. The struct building and verification is something the client, and the structure name should be something the user can specify as well.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/EmbeddedRelationalConnection.java
```diff
@@ -369,40 +360,31 @@ public SQLWarning getWarnings() throws SQLException {
...
-            throw ve.toSqlException();
+            return new RowArray(Arrays.stream(elements).collect(Collectors.toList()), RelationalArrayMetaData.of(DataType.ArrayType.from(elementType, false)));
         }
-        return new RowArray(Arrays.stream(elements).collect(Collectors.toList()), arrayMetaData);
     }
 
     @Override
     public Struct createStruct(String typeName, Object[] attributes) throws SQLException {
+        // We do not preserve the typeName, as there is no validation around that currently.
+        final var builder = EmbeddedRelationalStruct.newBuilder();
```
hatyo, line 382: I do like the simplification, nice work.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/RowStruct.java
```diff
@@ -240,8 +240,6 @@ public RelationalArray getArray(int oneBasedPosition) throws SQLException {
             for (final var t : coll) {
                 if (t instanceof Message) {
                     elements.add(new ImmutableRowStruct(new MessageTuple((Message) t), arrayMetaData.getElementStructMetaData()));
-                } else if (t instanceof ByteString) {
```
hatyo, line 243: I see that in `MessageTuple.sanitizeField` we return `ByteString::toByteArray()` when encountering a `ByteString` object. However, in this `else if` branch I see that you removed the dispatch on `ByteString` without sanitizing `t`, is this expected? Perhaps it is called indirectly above but I am not sure.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
```diff
@@ -738,16 +737,32 @@ private Expression parseRecordField(@Nonnull ParserRuleContext parserRuleContext
...
+    private static Value coerceValueIfNecessary(@Nonnull Value value, @Nonnull Type targetType) {
         final var resultType = value.getResultType();
         if (resultType.isUnresolved() ||
                 (resultType.isPrimitive() && PromoteValue.isPromotionNeeded(resultType, targetType))) {
-            final var promoteValue = PromoteValue.inject(value, targetType);
-            return new Expression(expression.getName(), DataTypeUtils.toRelationalType(promoteValue.getResultType()), promoteValue);
+            return PromoteValue.inject(value, targetType);
         }
-        return expression;
+        if (resultType.isArray() && value instanceof AbstractArrayConstructorValue) {
```
hatyo: You should check whether the promotion is necessary in case of arrays as well, otherwise,  a new `LightArrayConstructorValue` is constructed, potentially unnecessarily.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/StructMetaData.java
```diff
@@ -87,4 +89,6 @@ public interface StructMetaData extends Wrapper {
     default int getLeadingPhantomColumnCount() {
         return 0;
     }
+
+    DataType.StructType getRelationalDataType() throws SQLException;
```
hatyo, line 101: Please add nullability annotations and documentation.

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
g31pranjal: done!

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/jdbc.proto
g31pranjal: We cannot actually remove the `java_sql_types_code` straightaway - otherwise mixed-mode testing will start breaking as they still populate the `java_sql_types_code` field. Part of the code in JDBC deals with using the new Types, if available, otherwise fall to sql types. 

We can probably remove sql types in futue release though.

g31pranjal: done!

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/column.proto
g31pranjal: done!

fdb-relational-grpc/src/main/proto/grpc/relational/jdbc/v1/column.proto
g31pranjal: I think I am fine with `UNKNOWN` - just that my early idea came from the fact that there will be no type supplied by older versions.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
g31pranjal: done

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/ArrayMetaData.java
g31pranjal: done!

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/api/RowStruct.java
g31pranjal: Yes. The hidden contract here is that whatever comes out of `MessageTuple#getObjectInternal`, is already sanitised, which is depicted by the altered logic here. However, I had to "sanitise" for a special case when we think its a `Message` but is actually a special-cased nullable wrapper. For this particular case, we will have to sanitise the "repeated" field (content of that array).

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/StructMetaData.java
g31pranjal: done!

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/DataType.java
g31pranjal: I am not sure if this can be removed, but I think structural equality is needed. It circles back to the same issue of having a naming clash for structs of different types. We expect the elements of arrays to have the same data type or shape in case of a struct. Using `equals()` would not work since we give randomised identifiers to each struct. We cannot give same names internally or ask user to give a name that is unique.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/visitors/ExpressionVisitor.java
g31pranjal: nice catch!

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/DataType.java
```diff
@@ -1184,6 +1323,39 @@ public boolean equals(Object other) {
...
+            if (!(other instanceof StructType)) {
+                return false;
+            }
+            final var otherStructType = (StructType) other;
+            final var fields = this.getFields();
+            final var otherFields = otherStructType.getFields();
+            if (this.isNullable() != otherStructType.isNullable() || fields.size() != otherFields.size()) {
+                return false;
+            }
+            for (int i = 0; i < fields.size(); i++) {
```
hatyo, line 1341: This is not exactly correct. If the other structure has more fields it would still pass although it shouldn't I think.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LiteralsUtils.java
```diff
@@ -55,25 +50,18 @@ private LiteralsUtils() {
...
-        if (theType != Types.STRUCT) {
-            return new Type.Array(Type.primitiveType(SqlTypeSupport.sqlTypeToRecordType(theType), false));
-        }
-        final var distinctStructMetadata = objects.stream().filter(Objects::nonNull).map(o -> {
-            try {
-                return ((RelationalStruct) o).getMetaData();
-            } catch (SQLException e) {
-                throw new RelationalException(e).toUncheckedWrappedException();
+        DataType distinctType = null;
+        for (var object: objects) {
```
hatyo, line 56: This is not entirely correct I think in the presence of nullable fields.

hatyo (changes requested):

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalArrayFacade.java
```diff
@@ -66,9 +74,16 @@ class RelationalArrayFacade implements RelationalArray {
 
     RelationalArrayFacade(@Nonnull ColumnMetadata delegateMetadata, Array array) {
         this.delegateMetadata = delegateMetadata;
+        this.type = Suppliers.memoize(this::computeType);
```
hatyo, line 77: It is almost always a good idea to memoize repeated computations, but I think this should be pushed to the underlying `delegate` if possible (since this class is a mere facade presumably).

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalResultSetMetaDataFacade.java
```diff
@@ -47,6 +49,12 @@ public ArrayMetaData getArrayMetaData(int oneBasedColumn) throws SQLException {
         throw new SQLException("Not implemented", ErrorCode.UNSUPPORTED_OPERATION.getErrorCode());
     }
 
+    @Nonnull
+    @Override
+    public DataType.StructType getRelationalDataType() throws SQLException {
+        throw new SQLException("Not implemented", ErrorCode.UNSUPPORTED_OPERATION.getErrorCode());
```
hatyo, line 55: Can we implement this?

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -213,29 +262,30 @@ private static ColumnMetadata toColumnMetadata(@Nonnull ArrayMetaData metadata)
         return columnMetadataBuilder.build();
     }
 
-    private static ListColumnMetadata toListColumnMetadataProtobuf(@Nonnull RelationalStructMetaData metadata) throws SQLException {
+    private static ListColumnMetadata toListColumnMetadataProtobuf(@Nonnull StructMetaData metadata) throws SQLException {
         var listColumnMetadataBuilder = ListColumnMetadata.newBuilder();
         for (int oneBasedIndex = 1; oneBasedIndex <= metadata.getColumnCount(); oneBasedIndex++) {
-            var columnMetadata = toColumnMetadata(metadata, oneBasedIndex);
+            var columnMetadata = toColumnMetadata(metadata, oneBasedIndex, oneBasedIndex - 1 + metadata.getLeadingPhantomColumnCount());
```
hatyo, line 268: was this wrong all the time (in the presence of phantom columns, that is)?

hatyo (changes requested):

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/DataType.java
g31pranjal: I check on the sizes of both the `StructType`s and do element-wise comparison only if sizes are equal. Hence the case of `otherField` having more fields is covered imo.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
g31pranjal: No. `StructMetadata` methods, by default, used to hide phantom columns. Because of this. phantom columns get dropped when converting to JDBC protobuf. Hence the above logic worked correctly. However,  the new relational dataType do not hide any fields, hence we need "absolute" zero-based index. 

We will and should remove reliance on one-based indexes for internal purposes. This can be done as a cleenup after this PR to not break mixed-mode tests

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalArrayFacade.java
g31pranjal: Hmm, looks like `delegate` and `delegateMetadata` are not actually "delagate" in a sense, they are simply "backing" protobuf structures from which array content and metadata are read from. Hence, it will not be possible to push `type` into `delagate`.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalResultSetMetaDataFacade.java
g31pranjal: `RelationalResultSetMetaDataFacade` is created only when the `relationalDataType` is not available. That is, for the requests with continuations from older builds. Where we have the new `relationalDataType`, the `RelationalResultSetMetaDataFacade` can be completely avoided in favour of `StructResultSetMetaData`. 

Again, this is in place to support mixed-mode testing, and should be removed in cleanup!

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/LiteralsUtils.java
g31pranjal: Agreed, this won't work if the "shape" is even slightly different from one. Based on our discussion offline, I think we know how to improve this, but we probably want to do it separately as a separate line of work with rigorous testing.

g31pranjal: Opened issue on the same: https://github.com/FoundationDB/fdb-record-layer/issues/3472

hatyo (approved):

PR #3397, opened 2025-06-13
Keyvalue cursor

pengpeng-lu:


yaml-tests/src/test/resources/aggregate-index-tests-count.yamsql
```diff
@@ -108,8 +106,6 @@ test_block:
     -
       - query: select count(col1) from t2
       - explain: "ISCAN(MV5 <,>) | MAP (_ AS _0) | AGG (count(_._0.COL1) AS _0) | ON EMPTY NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)"
-      # Cannot run with FORCE_CONTINUATIONS due to: https://github.com/FoundationDB/fdb-record-layer/issues/3206
-      - maxRows: 0
```
pengpeng-lu, line 99: These 2 were also mis-categorized to issue 3206.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/ExecuteProperties.java
```diff
@@ -72,12 +72,13 @@ public class ExecuteProperties {
     // how record scan limit reached is handled -- false: return early with continuation, true: throw exception
     private final boolean failOnScanLimitReached;
     private final boolean isDryRun;
+    private final boolean kvCursorContSerializeToNew;
```
pengpeng-lu: will add comments in the next commit.

alecgrieser: Hm. I'm still not sure this really belongs in the `ExecuteProperties`. The point of the `ExecuteProperties` configuration is that it's supposed to contain stuff that a user might want to configure differently for other identical invocations of some operation. For example, various limits are in here, which could be applied to different plans. This is different, because for a given plan, it's actually very important that the user correctly serialize/deserialize KV cursor continuations. Which doesn't seem to fit the pattern

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +164,79 @@ public boolean isEnd() {
         @Nonnull
         @Override
         public ByteString toByteString() {
-            if (lastKey == null) {
-                return ByteString.EMPTY;
+            System.out.println("Keyvaluecursor toByteString serialization mode:" + serializationMode.name());
```
alecgrieser:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +164,79 @@ public boolean isEnd() {
...
+            if (serializationMode == SerializationMode.TO_OLD) {
+                if (lastKey == null) {
+                    return null;
+                }
+                return Arrays.copyOfRange(lastKey, prefixLength, lastKey.length);
+            } else {
+                byte[] result = toProto().toByteArray();
+                System.out.println("result:" + result);
+                return toProto().toByteArray();
+            }
```
alecgrieser: I suppose there are some performance implications, but it would be nice if this were able to be based on the `toByteString()` implementation

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -131,21 +137,23 @@ public RecordCursorResult<K> getNext() {
 
     @Nonnull
     private RecordCursorContinuation continuationHelper() {
-        return new Continuation(lastKey, prefixLength);
+        return new Continuation(lastKey, prefixLength, serializationMode);
     }
 
-    private static class Continuation implements RecordCursorContinuation {
+    public static class Continuation implements RecordCursorContinuation {
```
alecgrieser, line 143: Does this need to be `public` just so that the `RecordQueryIndexPlan` can update its continuation appropriately?

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -165,3 +165,8 @@ message OneOfTypedState {
 message RangeCursorContinuation {
     optional int64 nextPosition = 1;
 }
+
+message KeyValueCursorContinuation {
+    optional bytes continuation = 1;
+    optional bool isEnd = 2;
```
alecgrieser: We really shouldn't be returning any "end" continuations that are of type `KeyValueCursorContinuation`, so this seems a little suspect. To support something #1076, it could be useful to have the prefix length, though that actually has some portability problems, though that's not necessarily required. For the purposes of just solving this problem, I think the issue was just that we would sometimes get the empty string if the prefix size was the same as the whole key size. If we can confirm that wrapping it in a continuation and setting the `continuation` field to the empty string results in a non-empty continuation, that might actually be enough. If we did go with just a wrapped `bytes`, we should probably note why we went with that in a comment lest someone be confused why it's not just returning the underlying bytes.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +164,79 @@ public boolean isEnd() {
...
+                builder.setContinuation(base.substring(prefixLength, lastKey.length))
+                        .setIsEnd(false);
+            }
+            return builder.build();
+        }
+    }
+
+    public enum SerializationMode {
+        TO_OLD,
+        TO_NEW
```
alecgrieser, line 242: Using "old" and "new" here may be fine (it's certainly what we did with #3254). The danger is that we leave it in for too long and then we're confused what's "new" about it. I think the goal is for this to be pretty temporary, and for us to move pretty quickly to a mode where we only have the new way, and if that's the case, then we're probably okay with this nomenclature

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -247,10 +314,27 @@ protected void prepare() {
...
+                } else {
+                    try {
+                        RecordCursorProto.KeyValueCursorContinuation keyValueCursorContinuation = RecordCursorProto.KeyValueCursorContinuation.parseFrom(continuation);
+                        if (keyValueCursorContinuation.hasIsEnd()) {
+                            realContinuation = keyValueCursorContinuation.getContinuation().toByteArray();
+                        } else {
+                            realContinuation = continuation;
+                        }
+                    } catch (InvalidProtocolBufferException ex) {
+                        realContinuation = continuation;
```
alecgrieser: I'm not sure about this. If we hit this part, then it's a sign that someone has used the `TO_NEW` mode but the array hasn't been parseable as a protobuf. We may want to just throw an error here rather than try and paper over it, which could result in incorrect results.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -247,10 +314,27 @@ protected void prepare() {
...
+                if (serializationMode == SerializationMode.TO_OLD) {
+                    realContinuation = continuation;
+                } else {
+                    try {
+                        RecordCursorProto.KeyValueCursorContinuation keyValueCursorContinuation = RecordCursorProto.KeyValueCursorContinuation.parseFrom(continuation);
+                        if (keyValueCursorContinuation.hasIsEnd()) {
+                            realContinuation = keyValueCursorContinuation.getContinuation().toByteArray();
+                        } else {
+                            realContinuation = continuation;
+                        }
```
alecgrieser: I'm not sure I understand this `if` statement. Is the idea that the only way we'd get a protobuf where `isEnd` is not set is that this is actually an old-style continuation?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -208,6 +274,7 @@ public abstract static class Builder<T extends Builder<T>> {
         private StreamingMode streamingMode;
         private KeySelector begin;
         private KeySelector end;
+        protected SerializationMode serializationMode;
```
alecgrieser, line 280: It seems like there's some confusion about the lifecycle of this value. It's a member variable of the builder, and there's a setter, but it also gets cribbed from the execute properties during `prepare`. I'm not sure it needs to be a member variable of the builder, at least as it is currently. (It's possible that it would make sense for it to stay here if we end up removing it from the `ExecuteProperties`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +164,79 @@ public boolean isEnd() {
...
+                return continuationProto.getContinuation().toByteArray();
+            } catch (InvalidProtocolBufferException ipbe) {
+                return rawBytes;
+            }
+        }
+
+        @Nonnull
+        private RecordCursorProto.KeyValueCursorContinuation toProto() {
+            RecordCursorProto.KeyValueCursorContinuation.Builder builder = RecordCursorProto.KeyValueCursorContinuation.newBuilder();
+            if (lastKey == null) {
```
alecgrieser, line 229: The `null` handling here, in general, seems kind of suspect to me. Looking things over, I actually think that `lastKey` should never be `null`. It seems to be marked as `@Nullable` mainly because the `lastKey` variable in the outer class is `@Nullable`, but I don't think it should ever still be `null` when it creates a continuation. That's because:

1. The `lastKey` starts as `null`
2. If there is no data at all, then `lastKey` remains `null` but we return `RecordCursorResult.exhausted()` which does not create one of these continuations
3. If there is data, then `lastKey` will be set to a non-`null` value. All the scenarios where we create one of these continuations should come in this case

So, I don't think we'll ever hit this code path. I think it also points to the possibility that this isn't quite the right format for the continuation, though that's something that I said in a comment on `record_cursor.proto`

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/ForceContinuationQueryTests.java
```diff
@@ -0,0 +1,242 @@
...
+        try (final var s = connection.createStatement()) {
+            s.setMaxRows(1);
+            try (var resultSet = s.executeQuery("select count(*) from t2 group by col2")) {
+                Assertions.assertTrue(resultSet.next());
+                Assertions.assertEquals(1L, resultSet.getLong(1));
+                continuation = resultSet.getContinuation();
+            }
+        }
+        connection.setOption(Options.Name.KEYVALUE_CURSOR_CONTINUATION_SERIALIZE_TO_NEW, true);
+        try (final var preparedStatement = connection.prepareStatement("EXECUTE CONTINUATION ?param")) {
```
alecgrieser: What actually happens here? Is this testing the case that we parse the bytes, we see that it didn't become a protobuf, and then we fall back to the legacy mode?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -162,6 +163,7 @@ public class RecordQueryIndexPlan implements RecordQueryPlanWithNoChildren,
     @Nonnull
     private final Supplier<ComparisonRanges> comparisonRangesSupplier;
 
+
```
alecgrieser, line 168:
```suggestion
```

Looks like an extra newline snuck in that I don't think we want

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -328,7 +330,7 @@ private <M extends Message> RecordCursor<IndexEntry> executeEntriesWithOverScan(
                                                                                     @Nonnull FDBRecordStoreBase<M> store, @Nonnull Index index,
                                                                                     @Nullable byte[] continuation, @Nonnull ExecuteProperties executeProperties) {
         final byte[] prefixBytes = getRangePrefixBytes(tupleScanRange);
-        final IndexScanContinuationConvertor continuationConvertor = new IndexScanContinuationConvertor(prefixBytes);
+        final IndexScanContinuationConvertor continuationConvertor = new IndexScanContinuationConvertor(prefixBytes, executeProperties.isKvCursorContSerializeToNew() ? KeyValueCursorBase.SerializationMode.TO_NEW : KeyValueCursorBase.SerializationMode.TO_OLD);
```
alecgrieser: I'm not sure that this is clear necessarily, but one of the goals of the continuation convertor is to make the decision to run in overscan mode something that the user can turn on and off. So the continuation running in overscan mode needs to be the same as the continuation running without it.

I _think_ there are some tests to that effect somewhere, though I'm not certain. If there are, we should make sure that we have tests that that property is maintained with wrapped continuations. (Looking at the code, I _think_ that it is, I just wanted to make sure that this is under test.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -757,15 +762,21 @@ public byte[] unwrapContinuation(@Nullable final byte[] continuation) {
...
 
         @Override
         public RecordCursorContinuation wrapContinuation(@Nonnull final RecordCursorContinuation continuation) {
             if (continuation.isEnd()) {
                 return continuation;
             }
-            final byte[] continuationBytes = continuation.toBytes();
+            byte[] continuationBytes;
+            if (continuation instanceof KeyValueCursorBase.Continuation) {
+                continuationBytes = ((KeyValueCursorBase.Continuation) continuation).getInnerContinuationInBytes();
```
alecgrieser: Are there cases where the continuation is not of this type? Should assert on the continuation type here instead of attempting to work around it?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
```diff
@@ -44,6 +44,8 @@
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
```
alecgrieser, line 53: It would probably be good to add a test here where there is a scan over a single key range. The expectation would be that the continuation from the (only) key should be empty in the old serialization mode, but it should be non-empty in the new serialization mode

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/MultidimensionalIndexTestBase.java
```diff
@@ -496,7 +496,7 @@ void basicReadWithNullsTest(final boolean useAsync, @Nonnull final String storag
     }
 
     void indexReadTest(final boolean useAsync, final long seed, final int numRecords, @Nonnull final String storage,
-                   final boolean storeHilbertValues, final boolean useNodeSlotIndex) throws Exception {
+                           final boolean storeHilbertValues, final boolean useNodeSlotIndex) throws Exception {
```
alecgrieser: minor:

```suggestion
                       final boolean storeHilbertValues, final boolean useNodeSlotIndex) throws Exception {
```

I think. The main goal here would be to align the `final`s from the two lines

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlanWithOverScanTest.java
```diff
@@ -248,8 +250,12 @@ private static RecordCursorContinuation assertSameResults(@Nonnull FDBRecordStor
             do {
                 indexResult = indexCursor.getNext();
                 overscanResult = overscanCursor.getNext();
-                assertEquals(indexResult.getContinuation().toByteString(), overscanResult.getContinuation().toByteString(), "Continuation byte strings should match");
-                assertArrayEquals(indexResult.getContinuation().toBytes(), overscanResult.getContinuation().toBytes(), "Continuation byte arrays should match");
```
alecgrieser, line 252: Does this assertion actually need to change? I would have thought that we'd expect the continuations to stay the same, even if both are wrapped in protobuf

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordLayerIterator.java
```diff
@@ -83,6 +83,7 @@ public boolean hasNext() {
 
     private void fetchNextResult() {
         if (result != null) {
+            continuation = ContinuationImpl.fromRecordCursorContinuation(result.getContinuation());
```
alecgrieser: Hm, I'm not sure this is quite right. It seems like there is perhaps a deficiency in the previous code, but perhaps it should set the continuation on below near where it sets the continuation if the `noNextReason` is `SOURCE_EXHAUSTED`. That is, instead of just setting it to `END` if the no next reason is `SOURCE_EXHAUSTED`, it should also set it to the copied over continuation if the no-next-reason is not `SOURCE_EXHAUSTED`

alecgrieser (changes requested):
Sorry for the delay in getting to reviewing this!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/MultidimensionalIndexTestBase.java
pengpeng-lu: fixed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordLayerIterator.java
pengpeng-lu: you're right, changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: thanks for the catch! removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: Thanks for catching this! I removed the `is_end` from the proto, and removed the `null` handling here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: yes it was preventing the `parseFrom` to parse things in a weird way. I removed this since `is_end` field is removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: This is for the case that someone serializes TO_OLD, another one with TO_NEW needs to deserialize it. I think this is for phase 2. Phase 1: serializationMode = TO_OLD, Phase 2: serializationMode = TO_NEW, but can deal with old data; Phase 3: throw an ex here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: There seems no with core test... replaced it with the assert.

fdb-record-layer-core/src/main/proto/record_cursor.proto
pengpeng-lu: makes sense, removed the isEnd field.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/ForceContinuationQueryTests.java
pengpeng-lu: First we roll out with TO_OLD, then we roll out TO_NEW, I assume we should be able to deal with some continuation generated with TO_OLD? We should be able to parse it and it'll be the old behavior?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: yes, changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: Sorry for the debugging prints, removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: yes `RecordIndexPlan` calls this method: `KeyValueCursorBase.Continuation.fromRawBytes(continuation, serializationMode);`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlanWithOverScanTest.java
pengpeng-lu: yes I reverted the change, continuations stay the same.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +165,83 @@ public boolean isEnd() {
...
+                // proto.hasContinuation() = false when lastKey = null
+                return builder.setPrefixLength(prefixLength).build();
+            } else {
+                ByteString base = ZeroCopyByteString.wrap(Objects.requireNonNull(lastKey));
+                // proto.hasContinuation() = ByteString.EMPTY when prefixLength = lastKey.length
```
alecgrieser: Is it a problem if `proto.getContinuation()` returns the empty string if the prefix length and is equal to the last key length? I'd think that to fix #3206, the main thing would be to make sure that the final wrapped protobuf value isn't empty in that case, and that parsing it leads to `proto.hasContinuation()` returning `true`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -156,21 +165,83 @@ public boolean isEnd() {
...
+            try {
+                RecordCursorProto.KeyValueCursorContinuation continuationProto = RecordCursorProto.KeyValueCursorContinuation.parseFrom(rawBytes);
+                if (continuationProto.hasPrefixLength()) {
+                    return continuationProto.getContinuation().toByteArray();
+                } else {
+                    // parseFrom can parse an old serialization result as the new proto, wrong deserialization
+                    return rawBytes;
+                }
+            } catch (InvalidProtocolBufferException ipbe) {
+                return rawBytes;
```
alecgrieser: Hm, it seems weird here to return `rawBytes` in these situations. It seems like we should only ever hit them if the user misconfigures things so that an old continuation is deserialized by an old serialization cursor, and we may want to just throw in that case rather than try and be smart about it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -214,6 +217,21 @@ public RecordQueryIndexPlan(@Nonnull final String indexName,
                                 @Nonnull final Optional<? extends MatchCandidate> matchCandidateOptional,
                                 @Nonnull final Type resultType,
                                 @Nonnull final QueryPlanConstraint constraint) {
+        this(indexName, commonPrimaryKey, scanParameters, indexFetchMethod, fetchIndexRecords, reverse, strictlySorted, matchCandidateOptional, resultType, constraint, KeyValueCursorBase.SerializationMode.TO_NEW);
```
alecgrieser: Are we going to need to revert this to `TO_OLD` for now, and then we flip it to `TO_NEW` in the future?

yaml-tests/src/test/resources/aggregate-index-tests.yamsql
```diff
@@ -112,453 +112,8 @@ setup:
 test_block:
   name: agg-index-tests
   tests:
-    -
-      - query: select col1, sum(col2) from T1 group by col1;
```
alecgrieser, line 116: Why'd these all get taken out?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -231,6 +249,7 @@ public RecordQueryIndexPlan(@Nonnull final String indexName,
         }
         this.constraint = constraint;
         this.comparisonRangesSupplier = Suppliers.memoize(this::computeComparisonRanges);
+        this.serializationMode = serializationMode;
```
alecgrieser, line 253: It looks like the `serializationMode` here is only being used to unwrap the continuation before handing it to the underlying cursor. Shouldn't it also inform the way the continuation is wrapped when it comes back from the underlying? I'm assuming the flow is something like:

1. We add the serialization mode to the plan. This informs how the cursor continuations are wrapped/unwrapped, but it defaults to `TO_OLD`.
2. We flip the default to `TO_NEW`
3. We drop support for `TO_OLD`. At this point, we can also update the default values in the key value cursor so that they use the new format

The awkward part is that it actually seems like you don't need to change the `KeyValueCursor` until step 3, except for manual constructors of the `KeyValueCursorBase`.

The other other option would be to expose the serialization mode on the `FDBRecordStore` scan index/records methods, and then these could just set those as desired, which would be nice in that it would mean less interpretation of the continuation up here, but it is another API entry point, at least for the time being

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
```diff
@@ -143,6 +146,16 @@ public RecordQueryScanPlan(@Nullable Set<String> recordTypes,
...
     }
 
+    public RecordQueryScanPlan(@Nullable Set<String> recordTypes,
+                               @Nonnull Type flowedType,
+                               @Nullable KeyExpression commonPrimaryKey,
+                               @Nonnull ScanComparisons comparisons,
+                               boolean reverse,
+                               boolean strictlySorted,
+                               @Nonnull final Optional<? extends WithPrimaryKeyMatchCandidate> matchCandidateOptional) {
+        this(recordTypes, flowedType, commonPrimaryKey, comparisons, reverse, strictlySorted, matchCandidateOptional, KeyValueCursorBase.SerializationMode.TO_NEW);
```
alecgrieser: Should this one be `TO_OLD` as well?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
```diff
@@ -180,8 +195,10 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
         final TupleRange range = comparisons.toTupleRange(store, context);
+        byte[] innerContinuation = KeyValueCursorBase.Continuation.fromRawBytes(continuation, serializationMode);
+
         return store.scanRecords(
-                range.getLow(), range.getHigh(), range.getLowEndpoint(), range.getHighEndpoint(), continuation,
+                range.getLow(), range.getHigh(), range.getLowEndpoint(), range.getHighEndpoint(), innerContinuation,
```
alecgrieser: Like in the index plan, it seems like this should use the serialization mode to determine how to serialize the continuation

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
alecgrieser: I'm not sure I see the test over a single key range, but maybe I'm missing it?

yaml-tests/src/test/resources/disabled-planner-rewrites/aggregate-index-tests.yamsql
```diff
@@ -343,10 +343,6 @@ test_block:
       - query: select col1, sum(col2) from T1 where col1 > 15 group by col1;
       - explain: "AISCAN(MV1 [[GREATER_THAN promote(@c13 AS LONG)]] BY_GROUP -> [_0: KEY:[0], _1: VALUE:[0]]) | MAP (_._0 AS COL1, _._1 AS _1)"
       - result: [{!l 20, !l 76}]
-    -
-      - query: select sum(col2) from T1 where col1  = 10 group by col1
```
alecgrieser, line 347: Is there a reason this one got taken out as well?

alecgrieser (changes requested):

yaml-tests/src/test/resources/disabled-planner-rewrites/aggregate-index-tests.yamsql
pengpeng-lu: I added it back, it failed because of a bug and I forgot to put it back after fixing the bug, thanks for catching this!

yaml-tests/src/test/resources/aggregate-index-tests.yamsql
pengpeng-lu: It was a mistake, put them back.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
pengpeng-lu: here in line#198, `innerContinuation = fromRawBytes(continuation, serialization)`, we did use `serializationMode` to find out the `innerContinuation`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: yes you're right, reverted to `TO_OLD`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: I added a unit test in `KeyValueCursorTest` to verify:
```
proto = builder.build(); // proto.hasContinuation() = false in this case
proto = builder.setContinuation(ByteString.EMPTY).build();   // proto.hasContinuation() = true in this case.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
pengpeng-lu: yes reverted.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: I think in step2 when serializationMode is TO_OLD in old version, and TO_NEW in new version. If in version TO_NEW, we could try to deserialization a continuation generated by TO_OLD, in this case we'd like to return rawBytes, so that it behaves like TO_OLD.
From step 3 onwards (after all versions we care is in TO_NEW), InvalidProtocolBufferException should throw an exception. I modified line#221 to throw exception, I think that is a real exception even in step2.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: Yes I totally agree with step1,2,3 here. I'm a little confused of where you think we need to change `KeyValueCursor`? I think I changed `KeyValueCursorBase` even for step 1, and `KeyValueCursor` inherits these changes? Is there more changes you think are missing? We can chat offline about this -:)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
alecgrieser: Ah, I see. I think the fundamental disagreement is then on the middle phase. I still don't think that we want to have a "duck-typing" middle phase where we provide continuations and the `KeyValueCursor` takes a guess at what it could mean. I think we want this to mirror what we did with the aggregate cursor where we communicate the serialization mode in the plan(s), and then we know how the cursor is supposed to be parsed.

To make that work, we'd either need to expose the serialization mode through `scanIndex` and `scanRecords`, or we'd need to move the continuation wrapping/unwrapping logic into the calling plans

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
alecgrieser: There are more details here: https://github.com/FoundationDB/fdb-record-layer/pull/3397#discussion_r2301456524 But I was assuming a mode where the plans are the ones that need to be responsible for continuation wrapping/unwrapping, not the `KeyValueCursor`, in the intermediate steps

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
alecgrieser: Yes, that's true that we're using it to _deserialize_ the continuation, but I was saying that the continuation _serialization_ should also use it

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
alecgrieser: It looks like the most recent version of this is still relying on duck typing the serialization rather than using the `serializationMode` in the `RecordQueryIndexPlan` and `RecordQueryScan` plan to adjust how the continuation gets serialized. Is that the intention?

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -169,10 +184,52 @@ public byte[] toBytes() {
...
+            return byteString.isEmpty() ? new byte[0] : byteString.toByteArray();
+        }
+
+        public static byte[] fromRawBytes(@Nullable byte[] rawBytes, SerializationMode serializationMode) {
+            if (rawBytes == null) {
+                return null;
+            }
+            if (serializationMode == SerializationMode.TO_OLD) {
+                return rawBytes;
+            }
```
alecgrieser: Should this really be using the serialization mode to determine how to interpret the bytes? Shouldn't it always try to parse? Otherwise, it won't be able to switch over to interpreting new continuation values when we update some instances to `TO_NEW`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -131,21 +137,24 @@ public RecordCursorResult<K> getNext() {
...
+        return new Continuation(lastKey, prefixLength, serializationMode);
     }
 
-    private static class Continuation implements RecordCursorContinuation {
+    public static class Continuation implements RecordCursorContinuation {
         @Nullable
         private final byte[] lastKey;
         private final int prefixLength;
+        private final SerializationMode serializationMode;
+        private static final long MAGIC_NUMBER = 1234L;
```
alecgrieser: I was thinking a bit about this magic number. We should probably pick one with a few more bits, as this one, `1234`, has a hex representation of `0x4D2` or binary `0b10011010010`. In Protobuf's 7-bit variable length representation, that becomes `0b11010010 0b00001001` or `\xD2\x09`. That can hide within an FDB Tuple, so something like: `\x10\xD2\x09\x0A\x01\x14` would parse as a Tuple as `(-771159550, 0L)` and as a Protobuf message as `magic_number = 1234, inner_continuation = '\x14'`. If we picked a value that requires 5 bytes when packed into a var int, then I think we'd be safer, as long as we ensured that the final byte does not represent a valid Tuple code. (Also, a longer value would have more entropy, so it becomes much less likely that we accidentally happen to find it in the wild by chance.)

If we made magic number an `sfixed64` then that may also help. The type code for an `sfixed64` at field 2 would be `\x11`, which is a three byte negative number. If we ensure that the fourth byte (little-Endian) encoded is not a Tuple type code, then that would also do a good job to protect us from accidental ambiguous parsing

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: I changed it to 1234567890L and `sfixed64`, and explained why this is a safe choice, please let me know if it sounds correct to you.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -145,7 +145,20 @@ public static class Continuation implements RecordCursorContinuation {
         private final byte[] lastKey;
         private final int prefixLength;
         private final SerializationMode serializationMode;
-        private static final long MAGIC_NUMBER = 1234L;
+        /*
+        how we chose this "magic number":
+        The goal is to make sure an old continuation won't be accidentally parsed as: {magic_number = 1234567890L, inner_continuation = some byte array}
+        An example that can be parsed as above is:
+        byte[] data = new byte[] {0x11, (byte) 0xD2, 0x02, (byte) 0x96, 0x49, 0x00, 0x00, 0x00, 0x00, 0x0A, 0x01, 0x14};
+        where 0x11 is wire tag for sfixed64,
```
alecgrieser: This would be field 2, wire type `sfixed64`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -145,7 +145,20 @@ public static class Continuation implements RecordCursorContinuation {
...
+        [(byte) 0xD2, 0x02, (byte) 0x96, 0x49, 0x00, 0x00, 0x00, 0x00] is 1234567890L
+        0x0A is wire tag for bytes
+        0x01 represents that the byte array contains 1 byte
+        0x14 is the value of the byte array
+
+        Because 0xD2 is a negative number in java, it is not possible for a continuation to reach this number.
```
alecgrieser: I don't think that this is quite complete/correct. The more important thing is that `\x11` is the `Tuple` type code for a 3-byte negative integer. So, if there were some Tuple that created a byte string that somehow resulted in a message that could be so parsed, it would need to insert four bytes (`\x11` + `\xd2\x02\x96`) as a negative `Tuple`-encoded integer, and then it would need to start some new value from the next bytes. But the required next byte (`\x49`) is not a valid `Tuple` code, which makes it harder to accidentally generate one. I think the only way would be to doing a `PREFIX_STRING` scan (which is the only scan over keys where we expect non-`Tuple` values).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -145,7 +145,20 @@ public static class Continuation implements RecordCursorContinuation {
         private final byte[] lastKey;
         private final int prefixLength;
         private final SerializationMode serializationMode;
-        private static final long MAGIC_NUMBER = 1234L;
+        /*
+        how we chose this "magic number":
+        The goal is to make sure an old continuation won't be accidentally parsed as: {magic_number = 1234567890L, inner_continuation = some byte array}
+        An example that can be parsed as above is:
+        byte[] data = new byte[] {0x11, (byte) 0xD2, 0x02, (byte) 0x96, 0x49, 0x00, 0x00, 0x00, 0x00, 0x0A, 0x01, 0x14};
```
alecgrieser: I still think that this is not quite enough entropy, given the number of zero bits, though I like byte `\x49` isn't a Tuple code.

If I had to pick one, I might pick something like:

```java
private static final long MAGIC_NUMBER = 6773487359078157740L;
```

Or in hex:

```java
private static final long MAGIC_NUMBER = 0x5e0042dd9873cdacL;
```

In little-Endian, this has a representation of:

```java
new byte[]{ (byte) 0xAC, (byte) 0xCD, 0x73, (byte) 0x98, (byte) 0xDD, 0x42, 0x00, 0x5E };
```

Note that none of those bytes are valid Tuple codes (except for the `0x00`--which is also deliberate). That includes the byte at position 4, that is `\x98`, which is the relevant byte if we had a Tuple that began with `\x11`. The choice of 0x00 as the penultimate byte is there to protect against the only case that I know of where we don't get a valid `Tuple` back from a scan, namely `PREFIX_STRING` scan. In that case, the byte string will begin with some `String` suffix, which theoretically could be a valid Protobuf value. But that `String` suffix will have any `\x00` bytes escaped by following it with an `\xff` byte. So the sequence `\x00\x5e` would mean "end-of-string" followed by the beginning of a new Tuple value with code `\x5e`, which again, is invalid.

I think this one has the right balance of enough entropy to make such a sequence unlikely to appear in the wild, while also providing enough of a protection to stop us from accidentally parsing the new value.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -304,13 +324,15 @@ public <M extends Message> RecordCursor<IndexEntry> executeEntries(@Nonnull FDBR
         final RecordMetaData metaData = store.getRecordMetaData();
         final Index index = metaData.getIndex(indexName);
         final IndexScanBounds scanBounds = scanParameters.bind(store, index, context);
+        byte[] innerContinuation = KeyValueCursorBase.Continuation.fromRawBytes(continuation);
```
alecgrieser: I think this is wrong. We don't want to get the inner continuation here--we want to let the `KeyValueCursor` be the one that decides to un-wrap the continuation. The only case here that would be special would be the `executeEntriesWithOverScan`, which has to grab the inner continuation in order to modify it properly

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -768,19 +793,20 @@ public byte[] unwrapContinuation(@Nullable final byte[] continuation) {
                 return null;
             }
             // Add the prefix back to the inner continuation
-            return ByteArrayUtil.join(prefixBytes, continuation);
+            byte[] innerContinuation = KeyValueCursorBase.Continuation.fromRawBytes(continuation);
+            return ByteArrayUtil.join(prefixBytes, innerContinuation);
```
alecgrieser: One consequence of this here is that we always end up returning an unwrapped continuation to the underlying cursor. There's part of me that wants to say that we should return something like:

```java
new KeyValueCursorBase.Continuation(ByteArrayUtil.join(prefixBytes, innerContinuation), 0, serializationMode).toBytes();
```

Which would ensure that we're properly wrapping the bytes in the way that the underlying cursor expects. This is a little bit wasteful, in that we're parsing, unwrapping, modifying, wrapping, then unwrapping the same array a few times, but the abstractions are more cleanly respected

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -821,7 +849,8 @@ public byte[] toBytes() {
             @Nonnull
             @Override
             public ByteString toByteString() {
-                return baseContinuation.toByteString().substring(prefixLength);
+                byte[] result = KeyValueCursorBase.Continuation.fromRawBytes(baseContinuation.toBytes());
+                return result == null ? ByteString.EMPTY : ByteString.copyFrom(result).substring(prefixLength);
```
alecgrieser: The way this is written, I think we will always return an unwrapped continuation from this plan, which seems wrong. The goal of this class is to have the continuations returned by an index plan with overscan exactly match the continuations returned by one that doesn't overscan, which I believe would mean wrapping the continuation bytes back up.

One way to do that might actually be to add a method on the `KeyValueCursorBase.Continuation` that allows the user to create a copy that adjusts the prefix length. Then rather than manipulating the raw byte arrays here, we have the `KeyValueCursorBase.Continuation` adjust the part of the continuation that it returns. (We also probably need an `innerContinuationStartsWith` method or something to correctly know when to return an `END` continuation.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -795,10 +821,12 @@ private static class PrefixRemovingContinuation implements RecordCursorContinuat
             @SuppressWarnings("squid:S3077") // array immutable once initialized, so AtomicByteArray not necessary
             @Nullable
             private volatile byte[] bytes;
+            private final KeyValueCursorBase.SerializationMode serializationMode;
```
alecgrieser, line 821: Is this actually used any more? If we did the thing I mentioned and had the calling method manipulate a `KeyValueCursorBase.Continuation`, I think we could retain that object's serialization mode rather than needing to have it independently set here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
```diff
@@ -180,8 +180,10 @@ public <M extends Message> RecordCursor<QueryResult> executePlan(@Nonnull final
                                                                      @Nullable final byte[] continuation,
                                                                      @Nonnull final ExecuteProperties executeProperties) {
         final TupleRange range = comparisons.toTupleRange(store, context);
+        byte[] innerContinuation = KeyValueCursorBase.Continuation.fromRawBytes(continuation);
```
alecgrieser: I think this should also be unnecessary now, as the `KeyValueCursor` should be the one doing the unwrapping anyway

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -169,10 +197,49 @@ public byte[] toBytes() {
             if (lastKey == null) {
                 return null;
             }
-            return Arrays.copyOfRange(lastKey, prefixLength, lastKey.length);
+            ByteString byteString = toByteString();
+            return byteString.isEmpty() ? new byte[0] : byteString.toByteArray();
+        }
+
+        public static byte[] fromRawBytes(@Nullable byte[] rawBytes) {
```
alecgrieser: The more I think about it, I think this should be `getInnerContinuation`. That is, its contract is that it unwraps the continuation (if wrapped) and then returns the inner value

yaml-tests/src/test/resources/disabled-planner-rewrites/aggregate-index-tests.yamsql
alecgrieser: In the most recent version, I'm still seeing this query missing

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
```diff
@@ -535,4 +580,17 @@ public void buildWithRequiredProperties() {
         });
     }
 
+    @Test
+    void emptyByteStringSerDesTest() throws InvalidProtocolBufferException {
```
alecgrieser, line 664: I think this is actually the opposite test from the one I was thinking of. I was asking for a single-key range. In that case, if the results are non-empty, the inner continuation will be empty, To ensure that we don't mistake that as a begin continuation, we need to ensure that the (wrapped) continuation is not empty. That's the ultimate cause of #3206

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorTest.java
```diff
@@ -90,18 +96,24 @@ public void runBefore() {
         });
     }
 
-    @Test
-    public void all() {
+    @ParameterizedTest
```
alecgrieser, line 102: We should probably have a test in here that ensures that we get the right behavior if we have a Protobuf-parseable continuation (that is, that the magic number will actually save us). I think we have some good examples, like:

```java
Tuple.fromBytes(new byte[]{ 0x11, (byte) 0xac,  (byte) 0xcd, (byte) 0x73, 0x01, (byte) 0xdd, 0x42, (byte) 0x98, 0x5e, 0x0A, 0x04, 0x0f, (byte) 0xdb, 0x00, 0x14 } ),  // should set the magic_number to 0x5e9842dd0173cdacL and the inner continuation to `\x0f\xdb\x00\x14`
Tuple.fromBytes(new byte[]{ 0x18, 0x01, 0x0A, 0x02, 0x01, 0x14 } ) // should set an unknown field 3 to 1 and the inner_continuation to `\x01\x14`
```

We should be able to iterate over a subspace containing both of those keys using both serialization modes, and each time, we should resume from the correct location thanks to the magic number comparison

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: Thanks a lot! This is super helpful!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: Understand, I also remove the special null handling above, since line796~797 should correctly handle `null` as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
pengpeng-lu: changed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryScanPlan.java
pengpeng-lu: yes, removed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
pengpeng-lu: agree, `getInnerContinuation` is a better name.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -147,18 +147,17 @@ public static class Continuation implements RecordCursorContinuation {
...
+        new byte[]{ (byte) 0xAC, (byte) 0xCD, 0x73, (byte) 0x98, (byte) 0xDD, 0x42, 0x00, 0x5E };
+        Note that none of those bytes are valid Tuple codes (except for the 0x00--which is also deliberate).
+        That includes the byte at position 4, that is \x98, which is the relevant byte if we had a Tuple that began with \x11.
+        The choice of 0x00 as the penultimate byte is there to protect against the only case where we don't get a valid Tuple back from a scan,
+        namely PREFIX_STRING scan. In that case, the byte string will begin with some String suffix, which theoretically could be a valid Protobuf value.
+        But that String suffix will have any \x00 bytes escaped by following it with an \xff byte.
+        So the sequence \x00\x5e would mean "end-of-string" followed by the beginning of a new Tuple value with code \x5e, which again, is invalid.
          */
-        private static final long MAGIC_NUMBER = 1234567890L;
+        private static final long MAGIC_NUMBER = 677_348_735_907_815_774_0L;
```
alecgrieser:
```suggestion
        private static final long MAGIC_NUMBER = 6_773_487_359_078_157_740L;
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -147,18 +147,17 @@ public static class Continuation implements RecordCursorContinuation {
...
-        0x0A is wire tag for bytes
-        0x01 represents that the byte array contains 1 byte
-        0x14 is the value of the byte array
-
-        Because 0xD2 is a negative number in java, it is not possible for a continuation to reach this number.
+        The goal is to make sure an old continuation won't be accidentally parsed as: {magic_number = 6773487359078157740L, inner_continuation = some byte array}
+        In little endian, MAGIC_NUMBER: 6773487359078157740L is:
+        new byte[]{ (byte) 0xAC, (byte) 0xCD, 0x73, (byte) 0x98, (byte) 0xDD, 0x42, 0x00, 0x5E };
+        Note that none of those bytes are valid Tuple codes (except for the 0x00--which is also deliberate).
+        That includes the byte at position 4, that is \x98, which is the relevant byte if we had a Tuple that began with \x11.
```
alecgrieser:
```suggestion
        Note that the Protobuf tag for the magic_number field is \x11, which is a valid Tuple type code, corresponding to a three-byte
        negative integer. So a Tuple may legitimately begin \x11\xAC\xCD\x73, but then the next byte would need to be
        \x98 to match the magic number, and as that is not a Tuple code, a valid serialized Tuple cannot begin with a sequence
        that deserializes to a Protobuf that sets the magic number.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/KeyValueCursorBase.java
```diff
@@ -365,7 +364,7 @@ public T setContext(FDBRecordContext context) {
 
         @SpotBugsSuppressWarnings(value = "EI2", justification = "copies are expensive")
         public T setContinuation(@Nullable byte[] continuation) {
-            this.continuation = continuation;
+            this.continuation = KeyValueCursorBase.Continuation.getInnerContinuation(continuation);
```
alecgrieser:
```suggestion
            this.continuation = continuation;
```

In theory, I suppose we could get the inner continuation here, but if we do, we shouldn't unwrap the inner continuation in `prepare`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryIndexPlan.java
```diff
@@ -807,11 +832,7 @@ public byte[] toBytes() {
                 if (bytes == null) {
                     synchronized (this) {
                         if (bytes == null) {
-                            byte[] baseContinuationBytes = baseContinuation.toBytes();
-                            if (baseContinuationBytes == null) {
-                                return null;
-                            }
-                            this.bytes = Arrays.copyOfRange(baseContinuationBytes, prefixLength, baseContinuationBytes.length);
+                            bytes = KeyValueCursorBase.Continuation.getInnerContinuation(new KeyValueCursorBase.Continuation(baseContinuation.toBytes(), prefixLength, serializationMode).toBytes());
```
alecgrieser:
```suggestion
                            bytes = KeyValueCursorBase.Continuation.getInnerContinuation(new KeyValueCursorBase.Continuation(KeyValueCursorBase.Continuation.getInnerContinuation(baseContinuation.toBytes()), prefixLength, serializationMode).toBytes());
```

I believe. Either that, or we need to store the inner continuation bytes on the prefix removing continuation. It also seems like we could probably just...remove this entire class and use a `KeyValueCursorBase.Continuation` in its place if that's what we're going for.

My initial suggestion, fwiw, had been to add a boolean method `bool innerContinuationStartsWith(byte[] prefix)` as well as another method to return a new `KeyValueCursorBase.Continuation` with an updated prefix. The downside is that you'd need to explicitly cast the continuation in `wrapContinuation` to get the underlying object's methods. Either may be fine, as long as we're consistent here.

pengpeng-lu: changed, I agree we can remove the entire `PrefixRemovingContinuation` class, but I think it is a little bit more clear the way it is now. Maybe we can remove this and use a `KeyValueCursorBase.Continuation` in its place after we completely switch to `TO_NEW`, so the `KeyValueCursorBase.Continuation` is more concise?

alecgrieser: Maybe... In my mind, the decision to keep these two continuation classes really should be independent of the choice to return `TO_NEW` versus `TO_OLD` serialized forms. I'm fine with this as it is in the PR now, though

yaml-tests/src/test/resources/disabled-planner-rewrites/aggregate-index-tests.yamsql
alecgrieser, line 347: Is there a reason this test case is still deleted? Can we add it back?

PR #3398, opened 2025-06-13
Have ExternalServer dynamically find available ports

ScottDugas:
Developers have run into a fair number of issues with servers having different paths (e.g. if they have two copies of record layer), and thus not being cleaned up by the store. Instead, it now dynamically picks the ports (within a range), so if you have some leftover processes sitting around, new processes will find different ports.

This still leaves the code that kills processes around, so that if you keep killing tests, you don't end up with hundreds of servers running.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
@@ -59,17 +59,11 @@ public class ExternalServer {
      *
      * @param serverJar the path to the jar to run
      */
-    public ExternalServer(@Nullable File serverJar,
+    public ExternalServer(@Nonnull final File serverJar,
                           @Nullable final String clusterFile) throws IOException {
         this.clusterFile = clusterFile;
 
-        if (serverJar == null) {
-            final List<File> externalServers = getAvailableServers();
-            Assertions.assertEquals(1, externalServers.size());
-            this.serverJar = externalServers.get(0);
-        } else {
-            this.serverJar = serverJar;
-        }
+        this.serverJar = serverJar;
         Assertions.assertTrue(this.serverJar.exists(), "Jar could not be found " + serverJar.getAbsolutePath());
         this.version = getVersion(this.serverJar);
     }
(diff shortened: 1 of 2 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -201,6 +192,25 @@ private boolean startServer(ProcessBuilder processBuilder) throws IOException, I
         return serverProcess.isAlive();
     }
 
+    private int getAvailablePort(final int unavailablePort) {
```
ohadzeliger, line 196: It seems that the start method would only return once the server is fully started therefore the ports are in use (no race condition). Maybe worth while to add a comment to that effect?

yaml-tests/src/test/java/com/apple/foundationdb/relational/yamltests/server/ExternalServerTest.java
```diff
@@ -0,0 +1,73 @@
...
+        for (int i = 0; i < 3; i++) {
+            servers.add(new ExternalServer(currentServerPath, null));
+        }
+        try {
+            for (final ExternalServer server : servers) {
+                server.start();
+            }
+            // we can't assert about the actual values, because one of the ports may be busy
+            assertEquals(servers.stream().map(ExternalServer::getPort).distinct().collect(Collectors.toList()),
+                    servers.stream().map(ExternalServer::getPort).collect(Collectors.toList()));
```
ohadzeliger, line 67: Which is Similar to the assertion that the distinct size of the port list is of size 3, right?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/server/ExternalServer.java
```diff
@@ -58,21 +59,19 @@ public class ExternalServer {
...
-    public ExternalServer(@Nullable File serverJar, final int grpcPort, final int httpPort,
-                          @Nullable final String clusterFile) {
-        this.serverJar = serverJar;
-        this.grpcPort = grpcPort;
-        this.httpPort = httpPort;
+    public ExternalServer(@Nullable File serverJar,
+                          @Nullable final String clusterFile) throws IOException {
         this.clusterFile = clusterFile;
+
+        if (serverJar == null) {
```
ohadzeliger: I didn't see a usage where the server is null in the codebase.

ohadzeliger (changes requested):

PR #3401, opened 2025-06-17
WIP: Integrate Planner Rewrite Rules

alecgrieser:
This is a WIP PR to integrate the planner rewrite rules that have been added in order to canonicalize a query before planning.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -125,16 +135,45 @@ public void onMatch(@Nonnull final ImplementationCascadesRuleCall call) {
...
+                    }
+                }
+                final TranslationMap childAliasMap = TranslationMap.rebaseWithAliasMap(childAliasMapBuilder.build());
+                if (childAliasMap.definesOnlyIdentities()) {
+                    // No conflicts with other child aliases. Pull things up directly
+                    newQuantifiers.addAll(childSelectExpression.getQuantifiers());
+                    newPredicates.addAll(childSelectExpression.getPredicates());
+                } else {
+                    // At least one child needs to be rewritten. Rebase the children on top of the new name
+                    final List<Reference> oldChildRefs = childSelectExpression.getQuantifiers().stream().map(Quantifier::getRangesOver).collect(ImmutableList.toImmutableList());
```
normen662: Can you create a helper in `Quantifiers` named `rangedOver(...)` or something that does this one-liner? I have seen this pattern before and I meant to introduce such a function (just like `Quantifiers.aliases(...)`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -125,16 +135,45 @@ public void onMatch(@Nonnull final ImplementationCascadesRuleCall call) {
                 //
                 // Mergeable. Add the child quantifiers in the old one's place and scoop up any predicates.
                 //
-                newQuantifiers.addAll(childSelectExpression.getQuantifiers());
-                newPredicates.addAll(childSelectExpression.getPredicates());
+                AliasMap.Builder childAliasMapBuilder = AliasMap.builder();
```
normen662: This code looks to be conceptually correct (minus one-offs or something). NICE!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/RegularTranslationMap.java
```diff
@@ -71,9 +71,11 @@ public boolean containsSourceAlias(@Nullable CorrelationIdentifier sourceAlias)
     @Nullable
     @Override
     public CorrelationIdentifier getTarget(@Nonnull final CorrelationIdentifier sourceAlias) {
-        return getAliasMapMaybe()
-                .map(aliasMap -> aliasMap.getTarget(sourceAlias))
-                .orElseThrow(() -> new RecordCoreException("translation map is not backed by an alias map"));
+        AliasMap aliasMap = getAliasMapMaybe().orElse(null);
```
normen662: How is this code different than the old one in this context?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -349,6 +349,7 @@ public Reference memoizeExploratoryExpressions(@Nonnull final Collection<? exten
             final var expressionToReferenceMap = new LinkedIdentityMap<RelationalExpression, Reference>();
             referencePathsList.stream()
                     .flatMap(Collection::stream)
+                    .filter(path -> path.getReference().getPlannerStage() == plannerPhase.getTargetPlannerStage() && path.getReference().isExploratory(path.getExpression()))
```
normen662: I think there is a little to unpack here:
1. I think you identified that there is a _leak_ in `Traversal`. We should fix that leak.
2. I am not sure if (after 1. is fixed) we can leave this in here or make these conditional `Verify`s since there are references in the current phase that have not yet been advanced into the current target stage. Those, however, don't necessarily have their children advanced either, so there will be no way they can be reused. But after they have been advanced it may be possible to reuse them. Let's chat offline about this problem.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
```diff
@@ -282,7 +283,9 @@ public <P> Map<RecordQueryPlan, P> propertyValueForPlans(@Nonnull final Expressi
 
     @Nonnull
     public static ExpressionPropertiesMap<RelationalExpression> defaultForRewritePhase() {
-        return new ExpressionPropertiesMap<>(RelationalExpression.class, ImmutableSet.of(),
-                ImmutableSet.of(ExpressionCountProperty.selectCount()), ImmutableList.of());
+        return new ExpressionPropertiesMap<>(RelationalExpression.class,
+                ImmutableSet.of(CardinalitiesProperty.cardinalities()),
```
normen662: Do you need `CardinalitiesProperty.cardinalities()`?. I thought this made it easier in `DecorrelateValuesRule` but not sure. `DecorrelateValuesRule` is an exploration rule but all the lower expressions including the `TF` and the `SELECT` over the `TF` are already explored and have final expressions and would have that property. So that's good. On the other hand, the entire reference that `SELECT` is in, must be `minCard, maxCard == 1` if that `SELECT` has `minCard, maxCard == 1` so it's kind of pointless in keeping and maintaining this property for all expressions in the reference. I looked at the PRE for the rule again and I think something is not right in the way the `SELECT` of `TF` is matched. Another topic for a discussion I think.

alecgrieser: I may not actually need it. I put it in to use it in `DecorrlateValuesRule`, but I ended up taking it out. We could probably take it back out

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/translation/RegularTranslationMap.java
alecgrieser: It has to do with the semantics of `Optional::map`. If the lambda within `Optional::map` returns `null`, then it makes the whole thing empty. So previously, calling `getTarget` on an `AliasMap`-based `TranslationMap` with an absent source alias would result in this method throwing "translation map is not backed by an alias map". Now that case returns `null`, which I think was the intention of the method from the beginning

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
alecgrieser: Apparently, there already was one that was called `translateCorrelations`. I switched that to use the new `References.rebaseGraphs` method and then called it from here

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
alecgrieser: I tried to do some more digging into this. When I first encountered it, it was in a state where I was able to reliably get duplicates from expressions that were in different phases of exploration. However, after making some other changes, none of those tests are reproducing 🫠

I'm leaving them in for now, as I think it's pretty clear that we don't want to re-use a reference if it's in the wrong planner stage. I'll try and look more into it

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java
```diff
@@ -80,6 +71,15 @@ public int compare(final RelationalExpression a, final RelationalExpression b) {
             return Integer.compare(aTableFunctions, bTableFunctions);
         }
 
+        //
```
normen662, line 74: Thanks!

gradle.properties
```diff
@@ -19,7 +19,7 @@
 #
 
 rootProject.name=fdb-record-layer
-version=4.3.6.0
+version=4.4.0.0
```
normen662, line 22: NOICE

yaml-tests/src/test/resources/table-functions.metrics.yaml
```diff
@@ -56,35 +56,35 @@ table-functions:
...
     explain: EXPLODE array((@c20 AS B, @c22 AS C, (@c25 AS X, promote(@c27 AS DOUBLE)
         AS Y, @c29 AS Z) AS W), (@c34 AS B, @c36 AS C, (@c39 AS X, @c41 AS Y, @c43
-        AS Z) AS W)) | MAP (_.B AS B, _.C AS Q, _.W.X AS X) | FILTER _.B LESS_THAN
-        @c68
-    task_count: 172
-    task_total_time_ms: 1
-    transform_count: 38
-    transform_time_ms: 0
-    transform_yield_count: 8
+        AS Z) AS W)) | FILTER _.B LESS_THAN @c68 | MAP (_.B AS B, _.C AS Q, _.W.X
```
normen662, line 72: That is interesting... I don't think it's wrong, though.

normen662 (approved):

yaml-tests/src/test/resources/table-functions.metrics.yaml
alecgrieser: I think this makes sense. It's basically just saying that it prefers to push the filter below the `map` which is what we want

PR #3402, opened 2025-06-17
Type rename proto

ScottDugas:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditorUnitTest.java
```diff
@@ -275,7 +329,131 @@ public void renameOuterTypeWithNestedTypeWithSameName() throws Descriptors.Descr
...
+                        () -> "No resource: " + name))) {
+            RecordMetaDataProto.MetaData.Builder builder = RecordMetaDataProto.MetaData.newBuilder();
+            JsonFormat.parser().ignoringUnknownFields().merge(reader, builder);
+            return builder;
+        }
+    }
+
+    @ParameterizedTest
+    // Note: Explicitly having the .json here so that you can Cmd+click in the IDE to jump to the file
+    @ValueSource(strings = {"OneBoringType.json",
```
ScottDugas: It would be good to have a test where renaming one of the types would become another.
It would also be good to have a test where a type just kind of matches the renaming logic.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
```diff
@@ -113,16 +126,27 @@ private static DescriptorProtos.DescriptorProto.Builder fetchUnionBuilder(@Nonnu
...
+        }
         for (DescriptorProtos.FieldDescriptorProto.Builder field : unionBuilder.getFieldBuilderList()) {
-            final FieldTypeMatch unionFieldMatch = fieldIsType(recordsBuilder, unionBuilder, field, recordTypeName);
-            if (unionFieldMatch.isAmbiguousMatch()) {
-                throw new AmbiguousTypeNameException(recordsBuilder.getPackage(), unionBuilder, field, fullyQualifiedTypeName(recordsBuilder, recordTypeName));
-            } else if (FieldTypeMatch.MATCHES.equals(unionFieldMatch)
-                       && (foundField == null || field.getName().equals("_" + recordTypeName) || field.getNumber() > foundField.getNumber())) {
-                // Choose the matching field with either a matching name or the highest number.
-                foundField = field;
+            if (field.getTypeName().equals(recordTypeName)) {
```
alecgrieser: Does this need to check `field.getTypeName()` for some reason, or is this just an optimization? That is, could it rely on the `fieldIsType` check that you have below?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
```diff
@@ -517,28 +488,44 @@ private static void renameRecordTypeUsages(@Nonnull DescriptorProtos.FileDescrip
...
+        isHidden |= messageTypeBuilder.getNestedTypeList().stream()
+                .anyMatch(nestedType -> nestedType.getName().equals(oldRecordTypeName));
+
         // Rename any fields within the record type to the new type name
         for (DescriptorProtos.FieldDescriptorProto.Builder field : messageTypeBuilder.getFieldBuilderList()) {
-            final FieldTypeMatch fieldTypeMatch = fieldIsType(namespace, messageTypeBuilder, field, fullOldRecordTypeName);
-            if (fieldTypeMatch.isAmbiguousMatch()) {
-                throw new AmbiguousTypeNameException(namespace, messageTypeBuilder, field, fullOldRecordTypeName);
-            } else if (FieldTypeMatch.MATCHES.equals(fieldTypeMatch)) {
+            if (!isHidden && field.getTypeName().equals(oldRecordTypeName)) {
```
alecgrieser: This also feels a little bit sketchy. It seems like we should be able to just retain the `else` condition of this `if` statement, and then we don't need to worry about the `isHidden` complication

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
```diff
@@ -579,10 +566,41 @@ private static void renameTopLevelRecordType(@Nonnull RecordMetaDataProto.MetaDa
...
+                                            @Nonnull String oldRecordTypeName,
+                                            @Nonnull String newRecordTypeName) {
+        for (var unnested : metaDataBuilder.getUnnestedRecordTypesBuilderList()) {
+            for (var constituent : unnested.getNestedConstituentsBuilderList()) {
+                // The nested constituents would most likely be nested types, not record types, and thus would not be
+                // renamed
+                if (constituent.getParent().isEmpty() && constituent.getTypeName().equals(oldRecordTypeName)) {
+                    constituent.setTypeName(newRecordTypeName);
+                } else if (constituent.getTypeName().startsWith(metaDataBuilder.getRecords().getPackage()) &&
+                        constituent.getTypeName().endsWith(oldRecordTypeName)) {
```
alecgrieser: This seems like it would require similar logic to `MATCHES_AS_NESTED`, in that we'd need to update the type name of an unnested constituent if one of the types along its path changed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/metadata/SyntheticRecordType.java
```diff
@@ -37,6 +37,10 @@
  * A <i>synthetic</i> record type is made up of other record types and not actually stored separately in the record store.
  * It can, however, be indexed, by passing it to an {@link com.apple.foundationdb.record.provider.foundationdb.IndexMaintainer}
  * that will evaluate {@link KeyExpression}s against it.
+ * Note: If you are adding a new implementation of this, check out
```
alecgrieser, line 40: It would be nice, in general, if there were some kind of solution to "how do we know we've updated a thing that references a type as part of the type modification logic"? I'm not sure I have a _great_ solution to that, though. If you've given some thought to that, that would be good to know

alecgrieser: It's almost like we could use a test that randomly generates a meta-data, and then we could try to do things like make random modifications to it, and validate that it's the same to, say, generate a random meta-data with type set 1, then rewrite the types, versus generating the same meta-data with a different set of types. But that's pretty complicated

ScottDugas: This specific situation should be covered by `MetaDataProtoEditorUnitTest.validateMetaDataCoverage` although certainly not exhaustive. 

https://github.com/FoundationDB/fdb-record-layer/pull/3402/files#diff-e6160b248f4e8f9fc48fce82f1be8d46b39f9d706175ce9e2cc0c7ad2dd4fa8cR448

And yes, something that can randomly generated metadata's, modify them and then validate that the modification is correct does seem complicated, and seems like it might end up just pushing it to a different part of the code that you have to make sure to update.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
ScottDugas: I'm not 100% certain I have the right "below" reference, but;
Looking back at it, I think it was from before I decided to use the `FileDescriptor`, and I just didn't have to update this branch, only the one below, and didn't end up cleaning it up.
I removed this part of the if branch, just having the `else` below.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
ScottDugas: Yeah, this is another instance where I was adding sketchy stuff before switching to FileDescriptors, and forgot to  clean it up. Removed this whole branch, the `isHidden` variable, and `oldRecordTypeName` parameter.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
ScottDugas: Yeah, I think so. I'll try to add some tests for coverage of that scenario.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
```diff
@@ -566,31 +548,53 @@ private static void renameTopLevelRecordType(@Nonnull RecordMetaDataProto.MetaDa
...
+                    }
                 }
             }
         }
     }
 
+    private static boolean isNested(@Nonnull Descriptors.Descriptor typeDescriptor,
+                                    @Nonnull Descriptors.Descriptor targetDescriptor) {
+        if (typeDescriptor.equals(targetDescriptor)) {
+            return true;
```
alecgrieser, line 596: It's maybe a bit interesting to me that this method considers a type is considered to be nested within itself, though I think that's consistent with how the method is used. I'm trying to think if there's a way that makes that more obvious, like "is sub-type of", but that also depends on whether people think of "improper sub-types" (like an improper subset)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
```diff
@@ -566,31 +548,53 @@ private static void renameTopLevelRecordType(@Nonnull RecordMetaDataProto.MetaDa
...
+                } else {
+                    final Descriptors.Descriptor constituentTypeDescriptor = UnnestedRecordTypeBuilder.findDescriptorByName(fileDescriptor,
+                            constituent.getTypeName());
+                    if (constituentTypeDescriptor == null) {
+                        throw new MetaDataException("missing descriptor for nested constituent")
+                                .addLogInfo(LogMessageKeys.EXPECTED, constituent.getTypeName())
+                                .addLogInfo(LogMessageKeys.CONSTITUENT, constituent.getName());
+                    }
+                    if (isNested(constituentTypeDescriptor, oldTypeDescriptor)) {
+                        throw new MetaDataException("Renaming types used by non-parent unnested constituents is not supported");
```
alecgrieser, line 580: It's probably fine to leave this as un-supported for now, though it's also not that strange to think that we could actually just support it. I don't think it would be that different from re-naming fields that reference (potentially) nested types.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
alecgrieser: Yeah, that was the "below" part I was referencing. Makes sense

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBMetaDataStoreTest.java
```diff
@@ -811,7 +811,10 @@ private void addRecordType(@Nonnull DescriptorProtos.DescriptorProto newRecordTy
     }
 
     private void deprecateRecordType(@Nonnull String recordType) {
-        metaDataStore.mutateMetaData((metaDataProto) -> MetaDataProtoEditor.deprecateRecordType(metaDataProto, recordType));
+        metaDataStore.mutateMetaData((metaDataProto) -> {
+            Descriptors.FileDescriptor[] dependencies = RecordMetaDataBuilder.getDependencies(metaDataProto.build(), Map.of());
+            MetaDataProtoEditor.deprecateRecordType(metaDataProto, recordType, dependencies);
```
alecgrieser, line 816: I almost wonder if it would make sense from an API design perspective to retain a variant of some of these methods that doesn't take a `dependencies` list for cases that don't have dependencies. I could also see the argument that having a bunch of overloads just clutters the API. (I know that's a criticism levied against our API already)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditorUnitTest.java
```diff
@@ -100,17 +130,17 @@ public void fieldIsTypeUnqualified() throws Descriptors.DescriptorValidationExce
         assertNotNull(simpleRecordDescriptor);
         assertSame(simpleRecordDescriptor, modifiedFileDescriptor.findMessageTypeByName(RecordMetaDataBuilder.DEFAULT_UNION_NAME).findFieldByName("_MySimpleRecord").getMessageType());
 
-        assertEquals(FieldTypeMatch.MIGHT_MATCH,
+        assertEquals(FieldTypeMatch.MATCHES,
```
alecgrieser, line 136: I'm sure you felt this while you were trying to edit these tests, but this does seem like the kind of thing that we'd want to consider refactoring into a parameterized test. It probably would have sped up fixing all of these

fdb-record-layer-core/src/test/resources/DuplicateUnionFields.json
```diff
@@ -0,0 +1,72 @@
+{
+  "records": {
```
alecgrieser, line 2: Did you have some way to generate these json files (besides by hand)?

fdb-record-layer-core/src/test/resources/OneTypeWithIndexes.json
```diff
@@ -0,0 +1,60 @@
...
+        "field": [
+          {
+            "name": "ID",
+            "number": 1,
+            "type": "TYPE_INT64"
+          },
+          {
+            "name": "VALUE",
+            "number": 2,
+            "type": "TYPE_INT64"
```
alecgrieser, line 29: What would happen here if we had an index set via a field annotation? Would that get added into the `indexes` list (with a name based on the original type) and then type renaming would just keep the old name? Or would the annotation be persisted and potentially re-evaluated leading to an index with the new type name?

fdb-record-layer-core/src/test/resources/NestedMessage.json
```diff
@@ -0,0 +1,57 @@
...
+          {
+            "name": "ID",
+            "number": 1,
+            "type": "TYPE_INT64"
+          },
+          {
+            "name": "T1",
+            "number": 2,
+            "type": "TYPE_MESSAGE",
+            "type_name": "T1"
```
alecgrieser, line 40: Do we have tests here that attempt to do things like use different package specifications for the type name here? I think we have tests for that in the unit test, but maybe not through these JSON files. That level of testing may be fine

alecgrieser (approved):
LGTM. I had a few thoughts as I was going through, but it also looks good to me. Feel free to merge if you don't think there's anything that needs responding to

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
ScottDugas: Yeah, I think it makes it a little bit simpler to implement it's one usage. Although, maybe it would be better as a `do-while`. I added a javadoc to make it clear.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/MetaDataProtoEditor.java
ScottDugas: Yeah, it was something I didn't need, and I feel pretty ok with having this fail, rather than trying to add support, and tests.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/FDBMetaDataStoreTest.java
ScottDugas: I would agree with that sentiment that our API is overwhelmed with convenience overloads.

fdb-record-layer-core/src/test/resources/DuplicateUnionFields.json
ScottDugas: Basically by hand. I had a copy of one in use, and massively simplified it, and made modifications for some of them, and renamed all the types/fields.

fdb-record-layer-core/src/test/resources/NestedMessage.json
ScottDugas: I don't think I want to get into different package specifications at this level, because that's, in some ways, testing protobuf's resolution.
I did, however, add a test where there is a dependency with the same name as one of the types.

fdb-record-layer-core/src/test/resources/OneTypeWithIndexes.json
ScottDugas: It does not rename the index, the index is preserved in the metadata. Which is essential because it needs to persist things like subspace, and modified versions. I added a test of this behavior. The test uses the `RecordMetaData.toProto`, but that does appear to fully qualify, so I'm glad we still have all these handwritten json files.

ScottDugas (commented):
Yeah, I thought some of the additional tests would be important to have.

fdb-record-layer-core/src/test/resources/DuplicateUnionFields.json
alecgrieser: Yeah, that makes sense. We may need some utility to craft these kind of things more easily at some point, though today isn't that day

alecgrieser (approved):

PR #3405, opened 2025-06-18
Add validation/verification to the planner to avoid alias ambiguities and unresolved aliases

normen662:
This PR adds logic to
1. validate a query graph handed in to the `CascadesPlanner`
2. validate every expression yielded during `REWRITING` and `PLANNING`.

This validation logic is implemented in `Reference.verifyCorrelationsRecursive` and `Reference.verifyCorrelationsForNewExpression`.

I ran extensive verification tests with these checks enabled, but as coded up in the PR it will only be run if an insane `Debugger` is installed. That is currently the case for `:fdb-record-layer-core:test` but not for `:fdb-relational-layer-core:test`.

Upon running these tests the following problems surfaced:
1. There were quite a few tests using `RecordQuery` that caused `InComparisonToExplodeRule` to yield an incorrect `ExplodeExpression` of a `QOV(parameter)` where `parameter` is a bound parameter marker. Since `QOV(...)` is only properly defined over correlations, `QOV(parameter)` is an illegal alias reference as `parameter` is not a correlation. This was fixed by introducing `ParameterValue` that can dereference an actual parameter. Other rules such as `ImplementInJoinRule` and `ImplementInUnionRule` had to be adapted to create the proper `InSource` flavors.
2. `RecursiveUnionExpression` (and `RecordQueryRecursiveUnionPlan`) can correlate AND descendants can refer to the temp aliases. Logic was added to allow the alias resolution to understand that these temp table aliases are valid aliases to be used in subgraphs underneath `RecursiveUnionExpression` (and `RecordQueryRecursiveUnionPlan`).
3. We created illegal plans when `QueryPlan.strictlySorted(...)` was called (we incorrectly did not inherit the right quantifier alias). The fix is a one-liner in `RecordQueryPredicatesFilterPlan`.
4. Some test cases were written in a way that they handed only incomplete fragments to the planner. I sometimes changed the test case if that was easy. The more general fix is to allow an `EvaluationContext` to be passed in that the verification logic can use to identify additionally visible aliases.
5. Tons of adaptations in `RuleTestHelper` as
  a. some query graphs were incomplete (dangling unresolved aliases)
  b. the `Traversal` that is used inside of `CascadesRuleCall` to validate new expressions was incorrectly maintained in `RuleTestHelper`

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
@@ -717,6 +717,11 @@ public class Reference implements Correlated<Reference>, Typed {
         return PlannerGraphVisitor.show(renderSingleGroups, this);
     }
 
+    @Nonnull
+    public String showExploratory() {
+        return PlannerGraphVisitor.show(PlannerGraphVisitor.REMOVE_FINAL_EXPRESSIONS | PlannerGraphVisitor.RENDER_SINGLE_GROUPS, this);
+    }
+
     @SuppressWarnings("PMD.CompareObjectsWithEquals")
     private static boolean isMemoizedExpression(@Nonnull final RelationalExpression member,
                                                 @Nonnull final RelationalExpression otherExpression,
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ConstraintsMap.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ConstraintsMap.java
@@ -97,6 +97,18 @@ public class ConstraintsMap {
         this.attributeToConstraintMap = Maps.newLinkedHashMap();
     }
 
+    public long getCurrentTick() {
+        return currentTick;
+    }
+
+    public long getWatermarkGoalTick() {
+        return watermarkGoalTick;
+    }
+
+    public long getWatermarkCommittedTick() {
+        return watermarkCommittedTick;
+    }
+
     /**
      * Method to return if a particular attribute is contained in the map.
      * @param attribute the attribute key to check
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PartialMatch.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/PartialMatch.java
@@ -100,7 +100,7 @@ public class PartialMatch {
     private final Supplier<Map<CorrelationIdentifier, ComparisonRange>> boundParameterPrefixMapSupplier;
 
     @Nonnull
-    private final Supplier<Set<QueryPredicate>> boundPlaceholdersSupplier;
+    private final Supplier<Set<Placeholder>> boundPlaceholdersSupplier;
 
     @Nonnull
     private final Supplier<Set<Quantifier>> matchedQuantifiersSupplier;
@@ -217,14 +217,14 @@ public class PartialMatch {
     }
 
     @Nonnull
-    public final Set<QueryPredicate> getBoundPlaceholders() {
+    public final Set<Placeholder> getBoundPlaceholders() {
         return boundPlaceholdersSupplier.get();
     }
 
     @Nonnull
-    private Set<QueryPredicate> computeBoundPlaceholders() {
+    private Set<Placeholder> computeBoundPlaceholders() {
         final var boundParameterPrefixMap = getBoundParameterPrefixMap();
-        final var boundPlaceholders = Sets.<QueryPredicate>newIdentityHashSet();
+        final var boundPlaceholders = Sets.<Placeholder>newIdentityHashSet();
 
         //
         // Go through all accumulated parameter bindings -- find the query predicates binding the parameters. Those
@@ -246,7 +246,7 @@ public class PartialMatch {
 
             final var placeholder = (Placeholder)candidatePredicate;
             if (boundParameterPrefixMap.containsKey(placeholder.getParameterAlias())) {
-                boundPlaceholders.add(predicateMapping.getCandidatePredicate());
+                boundPlaceholders.add(placeholder);
             }
         }
 
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpression.java
@@ -853,4 +853,9 @@ public interface RelationalExpression extends Correlated<RelationalExpression>, 
     default String show(final boolean renderSingleGroups) {
         return PlannerGraphVisitor.show(renderSingleGroups, this);
     }
+
+    @Nonnull
+    default String showExploratory() {
+        return PlannerGraphVisitor.show(PlannerGraphVisitor.REMOVE_FINAL_EXPRESSIONS | PlannerGraphVisitor.RENDER_SINGLE_GROUPS, this);
+    }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
@@ -626,21 +626,31 @@ public class SelectExpression implements RelationalExpressionWithChildren.Childr
     @Nonnull
     @Override
     public PlannerGraph rewriteInternalPlannerGraph(@Nonnull final List<? extends PlannerGraph> childGraphs) {
+        final var predicateString = "WHERE " + getPredicates().stream()
+                .map(Object::toString)
+                .collect(Collectors.joining(" AND "));
+
+        final var abbreviatedPredicateString = predicateString.length() > 30 ? String.format("%02x", predicateString.hashCode()) : predicateString;
         return PlannerGraph.fromNodeAndChildGraphs(
                 new PlannerGraph.LogicalOperatorNode(this,
                         "SELECT " + resultValue,
                         getPredicates().isEmpty()
                         ? ImmutableList.of()
-                        : ImmutableList.of("WHERE " + getPredicates().stream()
-                                .map(Object::toString)
-                                .collect(Collectors.joining(" AND "))),
+                        : ImmutableList.of(abbreviatedPredicateString),
                         ImmutableMap.of()),
                 childGraphs);
     }
 
     @Override
     public String toString() {
-        return "SELECT " + resultValue + " WHERE " + AndPredicate.and(getPredicates());
+        final var selectFrom =
+                "SELECT " + resultValue + " FROM " + Quantifiers.aliases(getQuantifiers())
+                        .stream()
+                        .map(CorrelationIdentifier::toString).collect(Collectors.joining(", "));
+        if (!getPredicates().isEmpty()) {
+            return selectFrom + " WHERE " + AndPredicate.and(getPredicates());
+        }
+        return selectFrom;
     }
 
     /**
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ExpressionsPartitionMatchers.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ExpressionsPartitionMatchers.java
@@ -48,7 +48,7 @@ public class ExpressionsPartitionMatchers {
     @Nonnull
     @SuppressWarnings("unchecked")
     public static BindingMatcher<Reference> expressionPartitions(@Nonnull final BindingMatcher<? extends Iterable<? extends ExpressionPartition<? extends RelationalExpression>>> downstream) {
-        return TypedMatcherWithExtractAndDownstream.typedWithDownstream((Class<Reference>)(Class<?>)Reference.class,
+        return TypedMatcherWithExtractAndDownstream.typedWithDownstream(Reference.class,
                 Extractor.of(Reference::toExpressionPartitions, name -> "expressionPartitions(" + name + ")"),
                 downstream);
     }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/PlanPartitionMatchers.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/PlanPartitionMatchers.java
@@ -41,8 +41,8 @@ public class PlanPartitionMatchers {
 
     @Nonnull
     @SuppressWarnings("unchecked")
-    public static <R extends Reference> BindingMatcher<R> planPartitions(@Nonnull final BindingMatcher<? extends Iterable<PlanPartition>> downstream) {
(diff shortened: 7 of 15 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -229,6 +229,12 @@ private void yieldExpression(@Nonnull final RelationalExpression expression, fin
         }
     }
 
+    protected void validateNewExpression(@Nonnull final RelationalExpression expression) {
+        Debugger.sanityCheck(() -> verifyChildrenMemoized(expression));
```
alecgrieser: Do we want this to be part of a sanity check? The old code always validated that the children were memoized, so this is making things laxer, at least in the production configuration of the code

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -708,6 +710,132 @@ public boolean addPartialMatchForCandidate(final MatchCandidate candidate, final
         return partialMatchMap.put(candidate, partialMatch);
     }
 
+    public void verifyCorrelationsForNewExpression(@Nonnull final Traversal traversal,
+                                                   @Nonnull final RelationalExpression expression,
+                                                   @Nonnull final EvaluationContext evaluationContext) {
+        final Set<CorrelationIdentifier> correlatedToWithoutChildren;
+        if (expression instanceof RelationalExpressionWithChildren) {
+            correlatedToWithoutChildren = ((RelationalExpressionWithChildren)expression).getCorrelatedToWithoutChildren();
```
alecgrieser: Is there a reason this doesn't go down the children of this expression and validate those correlations as well?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -708,6 +710,132 @@ public boolean addPartialMatchForCandidate(final MatchCandidate candidate, final
...
+        } else {
+            correlatedToWithoutChildren = expression.getCorrelatedTo();
+        }
+
+        final var visibleThroughEvaluationContext = evaluationContext.getBindings().getBoundCorrelationAliases();
```
alecgrieser: Is this for things like constants?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -708,6 +710,132 @@ public boolean addPartialMatchForCandidate(final MatchCandidate candidate, final
...
+
+        final var currentUnresolvedCorrelatedTo =
+                currentUnresolvedCorrelatedToBuilder.build();
+        final var currentResolvedCorrelatedTo =
+                currentResolvedCorrelatedToBuilder.build();
+
+        final var parentRefPaths = traversal.getParentRefPaths(this);
+
+        if (parentRefPaths.isEmpty()) {
+            Verify.verify(currentUnresolvedCorrelatedTo.isEmpty(), "unresolved aliases: " + currentUnresolvedCorrelatedTo);
```
alecgrieser:
```suggestion
            Verify.verify(currentUnresolvedCorrelatedTo.isEmpty(), "unresolved aliases: %s", currentUnresolvedCorrelatedTo);
```

To prevent us from calculating the error message even if the verify succeeds

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -708,6 +710,132 @@ public boolean addPartialMatchForCandidate(final MatchCandidate candidate, final
...
+
+        final var currentUnresolvedCorrelatedTo =
+                currentUnresolvedCorrelatedToBuilder.build();
+        final var currentResolvedCorrelatedTo =
+                currentResolvedCorrelatedToBuilder.build();
+
+        final var parentRefPaths = traversal.getParentRefPaths(this);
+
+        if (parentRefPaths.isEmpty()) {
+            Verify.verify(currentUnresolvedCorrelatedTo.isEmpty(), "unresolved aliases: " +
```
alecgrieser:
```suggestion
            Verify.verify(currentUnresolvedCorrelatedTo.isEmpty(), "unresolved aliases: %s",
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -708,6 +710,132 @@ public boolean addPartialMatchForCandidate(final MatchCandidate candidate, final
...
+        } else {
+            for (final var parentRefPath : parentRefPaths) {
+                final var parentReference = parentRefPath.getReference();
+                parentReference.verifyCorrelationsForNewExpressionRecursive(traversal, parentRefPath.getExpression(),
+                        parentRefPath.getQuantifier(), currentUnresolvedCorrelatedTo, currentResolvedCorrelatedTo);
+            }
+        }
+    }
+
+    private void verifyCorrelationsForNewExpressionRecursive(@Nonnull final Traversal traversal,
```
alecgrieser: It feels like there's a fair amount of duplicated code between this method and `verifyCorrelationsForNewExpression`. Would it be possible to have the base case call the recursive case (potentially with like a `null` value for `quantifier` if need be)? I think the base case would still be responsible for constructing the set of unresolved correlation aliases, etc., but you then wouldn't need separate methods that both walked up the tree

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ParameterValue.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * QuantifiedObjectValue.java
```
alecgrieser:
```suggestion
 * ParameterValue.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ParameterValue.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2022 Apple Inc. and the FoundationDB project authors
```
alecgrieser:
```suggestion
 * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ParameterValue.java
```diff
@@ -0,0 +1,190 @@
...
+import com.google.common.collect.Iterables;
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * A value representing the quantifier as an object. For example, this is used to represent non-nested repeated fields.
```
alecgrieser: I think this is cribbed from `QuantifiedObjectValue`. We probably want to change this to something about how this represents a query parameter value, and that it wraps a value taken from the `EvaluationContext`. I suppose someone might want to know what the difference between this and a constant object value are, and ¯\\\_(ツ)_/¯ they are pretty similar

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
```diff
@@ -405,6 +405,11 @@ private void planPartial(@Nonnull final Supplier<Reference> referenceSupplier,
                              @Nonnull final Function<Reference, PlanContext> contextCreatorFunction,
                              @Nonnull final EvaluationContext evaluationContext) {
         this.currentRoot = referenceSupplier.get();
+
+        // run sanity check to make sure that all aliases handed in can be uniquely resolved
+        Debugger.sanityCheck(() ->
+                currentRoot.verifyCorrelationsRecursive(evaluationContext.getBindings().getBoundCorrelationAliases()));
```
alecgrieser: How expensive do you think it would be to run this check here at the beginning of planning? Running it with every new expression (outside of a sanity check) is probably too much, but it would probably be a good idea to verify at the start that the initial input is okay. Maybe there's a concern that we'd start failing someone's query that is technically illegal (or that we identify as illegal as a bug) which results in failures in production code. We may also need to switch this away from the `Verify` framework so that we generate an appropriate error code if we're using this validate user input

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/RecursiveUnionTest.java
```diff
@@ -321,9 +321,9 @@ private List<Long> multiplesOf(@Nonnull final List<Long> initial, long limit) th
 
             final var logicalPlan = Reference.initialOf(LogicalSortExpression.unsorted(Quantifier.forEach(Reference.initialOf(recursiveUnionPlan))));
             final var cascadesPlanner = (CascadesPlanner)planner;
-            final var plan = cascadesPlanner.planGraph(() -> logicalPlan, Optional.empty(), IndexQueryabilityFilter.TRUE, EvaluationContext.empty()).getPlan();
+            final var evaluationContext = putTempTableInContext(seedingTempTableAlias, seedingTempTable, null);
+            final var plan = cascadesPlanner.planGraph(() -> logicalPlan, Optional.empty(), IndexQueryabilityFilter.TRUE, evaluationContext).getPlan();
```
alecgrieser: This was "wrong" before, but by calling `cascadesPlanner::planGraph` directly instead of `FDBRecordStoreQueryTest::planGraph`, it's not sending the plans through serialization verification

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/RecursiveUnionTest.java
```diff
@@ -321,9 +321,9 @@ private List<Long> multiplesOf(@Nonnull final List<Long> initial, long limit) th
 
             final var logicalPlan = Reference.initialOf(LogicalSortExpression.unsorted(Quantifier.forEach(Reference.initialOf(recursiveUnionPlan))));
             final var cascadesPlanner = (CascadesPlanner)planner;
-            final var plan = cascadesPlanner.planGraph(() -> logicalPlan, Optional.empty(), IndexQueryabilityFilter.TRUE, EvaluationContext.empty()).getPlan();
+            final var evaluationContext = putTempTableInContext(seedingTempTableAlias, seedingTempTable, null);
```
alecgrieser: It seems a bit strange for the user to need to specify the temp table in this way, but I guess it was something that was already true about the way the API worked

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -245,6 +245,7 @@ public void bitmap(YamlTest.Runner runner) throws Exception {
     }
 
     @TestTemplate
+    @MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAINS)
```
alecgrieser:
```suggestion
```

Are there explains in the `recursive-cte.yamsql` test file that need to be modified?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
```diff
@@ -230,27 +243,45 @@ private static RelationalExpression costModel(final Reference reference) {
...
         final var expectedList = expectedListBuilder.build();
+
+        //
+        // If the rule that is being tested is an implementation rule, we need to prepare the subgraph underneath
+        // the group to be tested in a way that the implementation rule can find finalized expressions.
+        //
         if (rule instanceof ImplementationCascadesRule) {
             for (RelationalExpression expression : expectedList) {
-                preExploreForRule(expression, true);
+                preExploreForRule(expression, Traversal.withRoot(Reference.initialOf(expression)),
```
alecgrieser: Should this traversal be based on the `root` rather than the `expression`? It's possible the root and the expression can share subgraphs, so does that mean we'd need to copy the root and the expression as a unit to get the right traversal?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
normen662: That is correct. My thinking was that this has never triggered and we possibly can use the cycles elsewhere.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
normen662: Yeah, there are bunch of test cases written using incomplete graphs, and sometimes using temp tables. This seems to be the easier way out.

normen662: Constant themselves live in a different namespace among the bindings, so this is for actual `___corr_XXX`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
normen662: We could always run it, sure! If this triggers I would always view it as a bug answer the plan generator should have caught it, so I guess `verify` is fine.

yaml-tests/src/test/java/YamlIntegrationTests.java
normen662: Oh, I need to revert that. No, I sometimes do that as it doesn't run the multi server stuff if I repeatedly run it.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
normen662: So, I assume that `root` includes `group` which directly contains `expression`. The rule under testing is run on `group`. So the pre-explore has to run on `group`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/RecursiveUnionTest.java
normen662: let me see if I can fix that.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
normen662: That gets a little more tricky. In a lot of cases the expression that is yielded is the only thing that has changed. In those cases, this is exhaustive. If there is a memoization call that returns a new reference, and an expression in that new reference has a problem, we wouldn't catch it. Maybe I can call it when the exploration of that new reference starts. That however, would make the graph validation at beginning of planning superfluous.

Issue #3410, opened 2025-06-19
Implement planner rewrite rule for simplifying predicates in `SelectExpression`


hatyo:
With #3404 we have support in the predicate simplification for folding constant `QueryPredicate`s, under certain conditions. We could now leverage this work to simplify predicates in `SelectExpression`s by means of a new planner rewrite rule.

In the first step, we will just introduce the rewrite rule and add a number of tests for it, later on we'll integrate it in the planner, along with the rest of the rewrite rules.

Issue #3404, opened 2025-06-18
Constant folding support for `QueryPredicate`


hatyo:
This issue requests the implementation of constant folding optimizations for `QueryPredicate`s.  When possible, `QueryPredicate`s should be simplified to constant boolean values (`ConstantPredicate.TRUE`, `ConstantPredicate.FALSE`, or `ConstantPredicate.NULL`) during query planning.

Example:

Given `ValuePredicate(Cov(P1, Boolean), Comparison.ValueComparison(Equals, Cov(P2, Boolean)))`, and knowing that `P1` and `P2` are both `TRUE`, the predicate can be simplified to `ConstantPredicate.TRUE`.

There are numerous other scenarios where constant folding can be applied.

In case of `ValuePredicate`s, the implementation should also simplify the `Value`s inside of it, be delegating the actual simplification to the `Value` simplification engine.

PR #3411, opened 2025-06-19
Planner rewrite rule for simplifying `SelectExpression` predicates

### LINKED ISSUE #3407: `QueryPredicate` simplification rule for `PredicateWithValueAndRanges`
There is an interesting behavior in the construction of `PredicateWithValueAndRanges` that prevents simplification from working correctly.

If we have the following two predicates in a query:
```
P1 = PredicateWithValueAndRanges('False, [Null, Null])
P2 = PredicateWithValueAndRanges('False, ['False, 'False])
```

That is, the customer created a conjunction of the above predicates and passed it to the `SelectExpression`. If these predicates are kept as-is, then the simplification engine will be to recognize that they are constant predicates and simplifies them correctly:

```
P1 ^ P2
≡ PredicateWithValueAndRanges('False, [Null, Null]) ^ PredicateWithValueAndRanges('False, ['False, 'False])
≡ ConstantPredicate.FALSE ^ PredicateWithValueAndRanges('False, ['False, 'False]) -- ConstantFoldingPredicateWithRangesRule
≡ ConstantPredicate.FALSE ^ ConstantPredicate.FALSE -- ConstantFoldingPredicateWithRangesRule
≡ ConstantPredicate.FALSE -- AnnulmentAndRule
```

However, in the `SelectExpression` construction logic, there is handling of conjunctions that attempt to simplify predicates in a way that makes it easier to match against (filtered) indexes:

https://github.com/FoundationDB/fdb-record-layer/blob/e3fb56d2b2eb22df5829c11892cc33975dea8f19/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java#L703-L727

That logic creates a new `PredicateWithValueAndRanges` that looks like this:
```
P' = PredicateWithValueAndRanges('False, [Null, Null] AND ['False, 'False])
```
In other words, it creates a predicate that attempts to match `'False` against a range of two singletons `[Null, Null]`, and `[False, False]`. Albeit being a clear contradiction, the simplification engine will fail to pattern match it, since it has two ranges not one. And therefore, will not be simplified.

hatyo:
This introduces a new planner rewrite rule called `QueryPredicateSimplificationRule`, it matches `SelectExpression`s with at least one predicate, and attempts to simplify the conjunction of the predicates through the `QueryPredicate` simplification engine and rules (which were introduced in #3404).

If the predicate was simplified, it proceeds to create a new `SelectExpression` with that simplified predicate, and if the predicate is a conjunction, it breaks it apart, and supply its terms to the `SelectExpression` to align with the way the `SelectExpression` is constructed where its predicates are heavily preprocessed.

This PR also introduces a new `QueryPredicate` simplification rule called `ConstantFoldingMultiConstraintPredicateRule` that, as the name suggests, attempt to simplify a `PredicateValueWithRangeAndConstraints` that has 1 `RangeConstraints` with _at least_ 2 constraints. There are some interesting cases where the resulting constraints are contradicting and result in disjoint singleton ranges that are impossible to satisfy (more about that explained #3407).

Finally, the PR introduced a new randomized predicate testing framework that is able to generate complex predicate trees and validate them. It integrates with the `RandomizedTestUtils` so it can also benefit from any nightly testing capabilities.

This resolves #3407 and #3410.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/MultiMatcher.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/MultiMatcher.java
@@ -22,6 +22,7 @@ package com.apple.foundationdb.record.query.plan.cascades.matching.structure;
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.RecordQueryPlannerConfiguration;
+import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 
 import javax.annotation.Nonnull;
@@ -62,11 +63,11 @@ public abstract class MultiMatcher<T> implements CollectionMatcher<T> {
 
         // The children need to be merged in the same order that they appear to satisfy the contract of
         // PlannerBindings.getAll().
+        int counter = 0;
         for (final T item : in) {
             final List<PlannerBindings> individualBindings = downstream.bindMatches(plannerConfiguration, outerBindings, item).collect(Collectors.toList());
             if (individualBindings.isEmpty()) {
-                final Optional<Stream<PlannerBindings>> onEmptyStreamOptional =
-                        onEmptyIndividualBindings(bindingStream);
+                final Optional<Stream<PlannerBindings>> onEmptyStreamOptional = onEmptyIndividualBindings(bindingStream);
                 if (onEmptyStreamOptional.isEmpty()) {
                     return Stream.empty();
                 } else {
@@ -74,16 +75,20 @@ public abstract class MultiMatcher<T> implements CollectionMatcher<T> {
                 }
             } else {
                 items.add(item);
+                counter++;
                 bindingStream = bindingStream.flatMap(existing -> individualBindings.stream().map(existing::mergedWith));
             }
         }
 
         bindingStream = bindingStream.flatMap(existing -> Stream.of(PlannerBindings.from(this, items.build()).mergedWith(existing)));
-        return bindingStream;
+
+        return onNumberOfMatches(counter, bindingStream);
     }
 
     @Nonnull
     protected abstract Optional<Stream<PlannerBindings>> onEmptyIndividualBindings(@Nonnull Stream<PlannerBindings> accumulatedStream);
+
+    protected abstract Stream<PlannerBindings> onNumberOfMatches(int numberOfMatches, @Nonnull Stream<PlannerBindings> accumulatedStream);
 
     /**
      * A multi matcher that binds a sub collection of objects of the collection it is being matched. That includes the empty
@@ -109,6 +114,12 @@ public abstract class MultiMatcher<T> implements CollectionMatcher<T> {
             return "some " + nestedId + " in " + boundId + " that match {" + newLine(nestedIndentation) +
                    getDownstream().explainMatcher(Object.class, nestedId, nestedIndentation) + newLine(indentation) +
                    "}";
+        }
+
+        @Override
+        protected Stream<PlannerBindings> onNumberOfMatches(final int numberOfMatches,
+                                                            @Nonnull final Stream<PlannerBindings> accumulatedStream) {
+            return accumulatedStream;
         }
     }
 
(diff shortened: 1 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/MultiMatcher.java
```diff
@@ -22,6 +22,7 @@
 
 import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.query.plan.RecordQueryPlannerConfiguration;
+import com.google.common.base.Verify;
```
normen662, line 25: Thanks for writing this new matcher! I looked at all the users of `some(...)` in the rules. There are 20 references. In most of them, if not all you can see the `some(...)` matcher to be used immediately followed by a check that the matched result is not empty. Can you please update those usages of `some(...)` to use `atLeastOne(...)`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
```diff
@@ -104,6 +113,11 @@ public static Quantifier baseTau() {
         return forEach(new LogicalTypeFilterExpression(ImmutableSet.of("TAU"), fuseQun(), TYPE_TAU));
     }
 
+    @Nonnull
+    public static Quantifier baseBlah() {
```
normen662: Can we name this something other than `blah` (same for the type)? Maybe you can add these columns to the `T` and/or `TAU`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/QueryPredicateSimplificationRule.java
```diff
@@ -0,0 +1,128 @@
...
+ *  SELECT a, b, c
+ *  FROM t
+ *  WHERE null
+ *  }</pre>
+ */
+@SuppressWarnings({"PMD.TooManyStaticImports", "PMD.CompareObjectsWithEquals"})
+public class QueryPredicateSimplificationRule extends ExplorationCascadesRule<SelectExpression> {
+
+    @Nonnull
+    private static final CollectionMatcher<QueryPredicate> predicateMatcher = atLeastOne(anyPredicate());
```
normen662, line 93: I like this new matcher.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/QueryPredicateSimplificationRule.java
```diff
@@ -0,0 +1,128 @@
...
+        final var predicates = call.get(predicateMatcher);
+        final var conjunction = AndPredicate.and(predicates);
+        final var aliasMap = AliasMap.emptyMap();
+        final var constantAliases = Sets.difference(conjunction.getCorrelatedTo(),
+                Quantifiers.aliases(selectExpression.getQuantifiers()));
+
+        final var simplifiedConjunction = Simplification.optimize(conjunction, call.getEvaluationContext(), aliasMap,
+                constantAliases, ConstantFoldingRuleSet.ofSimplificationRules());
+
+        if (simplifiedConjunction.get() == conjunction) {
```
normen662: Can you use `semanticEquals(...)` instead?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/MultiMatcher.java
hatyo: Thanks, I was thinking about doing that actually, but not in this PR, so I opened this issue to track it later:
https://github.com/FoundationDB/fdb-record-layer/issues/3415

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/RuleTestHelper.java
hatyo: Let me try to remove that completely.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/QueryPredicateSimplificationRule.java
hatyo: That seems unnecessarily expensive, considering that the premise of `optimize` method is that it returns the _same_ object should it fail to optimize it, but I might be wrong.

normen662: It's pretty much the same in terms of amortized performance:
- if it is the same object, `semanticEquals(...)` is supposed to short circuit and immediately return
- if it is not the same object the, then we are testing for inequality really and that should be fairly fast as well
- the only case where it takes longer is where it is not the same object but semantically equal (which is they case we want to catch). Optimize doesn't _fail_ to optimize anything, it just returns an equivalent predicate that hopefully is simpler, but there is nothing in the contract that says that optimize has to return the same instance of an object under any circumstances. I think using `semanticEquals(...)` is just more defensive and sturdy against future changes.

hatyo: done.

normen662 (approved):

PR #3421, opened 2025-06-23
Add new `Option`s to disable planner rules through the Relational API

### LINKED ISSUE #3420: Expose options to disable planner rules through the Relational interface
There are already options to disable certain planner rules in the `RecordQueryPlanConfiguration`. However, that is not currently exposed through the Relational (JDBC) interface. We should add some `Option`s to the query/connection that allow that to happen.

While we are there, we should make it easy to disable the new plan rewrite rules that are going in (see #3401). Those rules fundamentally shift the way that more complicated queries get planned, and they could have significant performance implications. Therefore, it would be good to expose a way to disable all of those new rules at once.

alecgrieser:
This exposes the existing methods on the `RecordQueryPlanConfiguration` to allow us to disable planner rules through the Relational API. It also adds a new helper method that allows us to turn off all of the planner rewrite rules at once, which can be useful as a stop-gap as we work to integrate them further.

This resolves #3420.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
@@ -27,12 +27,14 @@ import com.apple.foundationdb.record.query.RecordQuery;
 import com.apple.foundationdb.record.query.plan.cascades.CascadesRule;
 import com.apple.foundationdb.record.query.plan.cascades.PlannerRule;
 import com.apple.foundationdb.record.query.plan.cascades.PlanningRuleSet;
+import com.apple.foundationdb.record.query.plan.cascades.RewritingRuleSet;
 import com.apple.foundationdb.record.query.plan.cascades.rules.PredicateToLogicalUnionRule;
 import com.apple.foundationdb.record.query.plan.plans.QueryPlan;
 import com.apple.foundationdb.record.query.plan.serialization.PlanSerialization;
 import com.apple.foundationdb.record.query.plan.sorting.RecordQueryPlannerSortConfiguration;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -434,6 +436,7 @@ public class RecordQueryPlannerConfiguration {
             this.protoBuilder = RecordPlannerConfigurationProto.PlannerConfiguration.newBuilder();
         }
 
+        @CanIgnoreReturnValue
         @Nonnull
         public Builder setIndexScanPreference(@Nonnull QueryPlanner.IndexScanPreference indexScanPreference) {
             protoBuilder.setIndexScanPreference(SCAN_PREFERENCE_BI_MAP.get(indexScanPreference));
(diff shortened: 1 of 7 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -434,6 +436,7 @@ public Builder() {
             this.protoBuilder = RecordPlannerConfigurationProto.PlannerConfiguration.newBuilder();
         }
 
+        @CanIgnoreReturnValue
```
normen662, line 439: Apparently, you can use `@CanIgnoreReturnValue` on type level.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
```diff
@@ -609,18 +629,43 @@ public Builder setDisabledTransformationRules(@Nonnull final Set<Class<? extends
...
          */
+        @CanIgnoreReturnValue
         @Nonnull
         public Builder setDisabledTransformationRuleNames(@Nonnull final Set<String> disabledTransformationRuleNames, @Nonnull PlanningRuleSet planningRuleSet) {
             protoBuilder.clearDisabledTransformationRules()
                     .addAllDisabledTransformationRules(disabledTransformationRuleNames);
             return this;
         }
 
+        /**
```
normen662, line 640: Can we have a mirror of this for the `PLANNING` rule set? Or is that useless.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
alecgrieser: I thought about that. One thing that gave me pause on that is: https://github.com/google/error-prone/blob/150417a3d52794649b1fe1289a358a1a97771f20/annotations/src/main/java/com/google/errorprone/annotations/CanIgnoreReturnValue.java#L35

> // Note: annotating a type with `@CanIgnoreReturnValue` is discouraged (and banned inside of Google)

I didn't annotate the `build()` method with `@CanIgnoreReturnValue`, if we care about being that precise. But it was mostly this note here suggesting that annotating that way would be a bad idea that gave me pause.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
alecgrieser: I don't think we'd want to do this for (all) the planning rules, though maybe there are some other sub-sets of rules that we'd want to make macros for like this. I don't really see it making sense for planning rules though as then you'd just get "Cascades could not plan query" for everything

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
normen662: Okay!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
normen662: You could disable `ImplementDistinctUnion` and get `UNION ALL` with `DISTINCT` on top.
Sometimes, we saw IN UNION win over IN JOIN for no apparent reason. We fixed that bug but it would have been useful to just disable IN UNION.

 So it's definitely useful.

normen662 (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/RecordQueryPlannerConfiguration.java
alecgrieser: Maybe I'm misunderstanding the suggestion, but you can already do that with the existing `RecordQueryPlannerConfiguration` API:

```java
configurationBuilder
   .disableTransformationRule(ImplementDistinctUnion.class)
   .build()
```

And now you can also do that through the relational API:

```java
connection.setOption(Option.Name.DISABLED_PLANNER_RULES, ImmutableSet.of("ImplementDistinctUnion"));
```

The only difference with the rewriting rules is that there's _also_ this method (that the comment is on) that allows one to disable all of the rewriting rules at once, which is the part that I think wouldn't make sense for the planning rules.

normen662: Ah right!

PR #3422, opened 2025-06-23
addressing matching problems when integrating `REWRITING` phase

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/WindowedValue.java
```diff
@@ -134,23 +134,23 @@ protected int basePlanHash(@Nonnull final PlanHashMode mode, ObjectPlanHash base
     @SuppressWarnings("PMD.ForLoopCanBeForeach")
     public ExplainTokensWithPrecedence explain(@Nonnull final Iterable<Supplier<ExplainTokensWithPrecedence>> explainSuppliers) {
         int i = 0;
-        final var argumentsBuilder = ImmutableList.<ExplainTokens>builder();
+        final var partitioningBuilder = ImmutableList.<ExplainTokens>builder();
```
normen662, line 137: explain was wrong

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
```diff
@@ -406,7 +406,7 @@ void testOrQuery1(OrQueryParams orQueryParams) throws Exception {
 
         // Fetch(Covering(Index(MySimpleRecord$num_value_3_indexed [[1],[1]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]) ∪ Covering(Index(MySimpleRecord$num_value_3_indexed [[2],[2]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]) ∪ Covering(Index(MySimpleRecord$num_value_3_indexed [[4],[4]]) -> [num_value_3_indexed: KEY[0], rec_no: KEY[1]]))
         RecordQueryPlan plan = planQuery(query);
-        final KeyExpression comparisonKey = planner instanceof CascadesPlanner ? concat(primaryKey("MySimpleRecord"), field("num_value_3_indexed")) : primaryKey("MySimpleRecord");
```
normen662, line 409: @alecgrieser I think this is ok. Do you think there is a problem?

alecgrieser: It seems fine, at least for a scalar field like `num_value_3_indexed` where it's not possible for there to be duplicate records from the different legs of the union (and the lack of ordering constraint). It is interesting that this order effectively transforms the query execution into something more like an in-join (in that the union legs are returned in sequence). I suppose it raises questions like "should we try and re-write queries like this so that we can execute them as in-joins", or potentially we could use a "concat" operator to execute unions without ordering constraints. There's already the unordered union operator, which currently executes all the legs in parallel, but we could consider modifying that to execute things in serial, which would decrease the amount of data that we load at any given time, at the cost of less parallel processing (especially when legs have few results).

All of that seems secondary, and this change seems like it's still semantically correct. Do you have an idea what caused it? Does it have something to do with, like, different compensation values meaning the cost model now favors the other plan (potentially because of plan hash changes)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
```diff
@@ -125,6 +125,11 @@ public boolean isNeeded() {
             return false;
         }
 
+        @Override
+        public boolean isFinalNeeded() {
```
alecgrieser, line 129: What is this supposed to signify?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
```diff
@@ -142,49 +141,26 @@ public GraphExpansion expand(@Nonnull final Quantifier.ForEach baseQuantifier,
...
-            // this is needed for reapplication of the component if the sub query cannot be matched or only matched with
-            // compensation
-            QueryComponent withPrefix = this;
-            for (int i = fieldNamePrefix.size() - 1; i >= 0;  i--) {
-                final String fieldName = fieldNamePrefix.get(i);
-                withPrefix = Query.field(fieldName).matches(withPrefix);
-            }
+            rankExpansionBuilder.pullUpQuantifier(innerBaseQuantifier);
+            partitioningAndArgumentExpansion.getQuantifiers()
+                    .forEach(quantifier -> rankExpansionBuilder.addAllResultColumns(quantifier.getFlowedColumns()));
```
alecgrieser, line 159: Why do we care about adding the result columns here if we're going to then wrap the `rankExpansionBuilder` in an existential quantifier? Can things access the data of an existential quantifier's record?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
```diff
@@ -262,17 +265,57 @@ public GraphExpansion visitExpression(@Nonnull final NestingKeyExpression nestin
...
+                    final SelectExpression selectExpression =
+                            sealedBaseAndChildExpansion.buildSelect();
+                    final Quantifier childQuantifier = Quantifier.forEach(Reference.initialOf(selectExpression));
+                    return sealedBaseAndChildExpansion
+                            .builderWithInheritedPlaceholders()
+                            .pullUpQuantifier(childQuantifier)
+                            .build();
                 } else {
-                    baseAndChildExpansion = childExpansion.withBase(childBaseQuantifier);
+                    final var baseAndChildExpansionBuilder =
```
alecgrieser, line 284: If I'm understanding the logic here, the new structure is something like:

1. The child quantifier gets all of the the predicates (and placeholders) defined on that child and returns all of its original columns
2. The parent quantifier translates all of the result values so that they are now defined on the (new) child quantifier and can be put at the parent level

So something like a fan out index on `field("a", FanOut).nest(concatenateFields("b", "c"))` will get translated into something like:

```sql
SELECT x.b, x.c FROM T,
   (SELECT * FROM T.a WHERE T.b <placeholder1> AND T.c <placeholder2> ) x
```

Whereas before, it would have been:

```sql
SELECT x.b, x.c FROM T,
   (SELECT b, c FROM T.a WHERE T.b <placeholder1> AND T.c <placeholder2> ) x
```

And the proximate reason for this is that it makes it easer to match queries like:

```sql
SELECT x.c FROM T, T.a AS x WHERE T.b = ?bParam
```

Because now the quantifiers corresponding to `x` now have the same columns. This all seems fine, though it could probably use a comment

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -573,9 +574,14 @@ public Compensation compensate(@Nonnull final PartialMatch partialMatch,
...
-                    PredicateMultiMap.ResultCompensationFunction.of(baseAlias -> pulledUpResultValue.translateCorrelations(
-                            TranslationMap.ofAliases(rootPullUp.getNestingAlias(), baseAlias), false));
+            if (QuantifiedObjectValue.isSimpleQuantifiedObjectValueOver(pulledUpResultValue,
+                    rootPullUp.getNestingAlias())) {
+                resultCompensationFunction = ResultCompensationFunction.noCompensationNeeded();
+            } else {
+                resultCompensationFunction =
+                        ResultCompensationFunction.of(baseAlias -> pulledUpResultValue.translateCorrelations(
+                                TranslationMap.ofAliases(rootPullUp.getNestingAlias(), baseAlias), false));
+            }
```
alecgrieser: I think this is repeated a few times, and it might be nice to encapsulate it into a function in `ResultCompensationFunction`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -573,9 +574,14 @@ public Compensation compensate(@Nonnull final PartialMatch partialMatch,
 
             final var pulledUpResultValue = pulledUpResultValueOptional.get();
 
-            resultCompensationFunction =
-                    PredicateMultiMap.ResultCompensationFunction.of(baseAlias -> pulledUpResultValue.translateCorrelations(
-                            TranslationMap.ofAliases(rootPullUp.getNestingAlias(), baseAlias), false));
+            if (QuantifiedObjectValue.isSimpleQuantifiedObjectValueOver(pulledUpResultValue,
+                    rootPullUp.getNestingAlias())) {
+                resultCompensationFunction = ResultCompensationFunction.noCompensationNeeded();
```
alecgrieser, line 579: Why don't we need to compensate in this case (that is, if I understand correctly, turn a `QOV(rootPullUp.getNestingAlias())` into a `QOV(baseAlias)`, which is what will be different between the old code and the new code)?

alecgrieser (approved):
LGTM. I had a few comments, mainly about things that were unclear to me/documentation, but the overall change looks good

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/QueryRecordFunctionWithComparison.java
normen662: Good question, has to do with the way the matching code is structured. What happens is that we don't need to compensate for the difference in result values, however, there is a check in `SelectExpression.compensate(...)` approx line 901 that cannot accept any kind of compensation as such compensation would not be expressible with the tools we currently have. So even though we know that we only have to potentially have to compensate for a different result (which doesn't need to happen in case of an existential above), this code (which needs to be there right now for other reasons) ruins the party and declares and impossible match.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
normen662: Each compensation has an `isFinalNeeded()` and an `isNeededForFiltering()` which signify if the compensation messes with the expression by imposing new filters, or in this case just changes the shape of the output. The shape of the output is only corrected once at the very end of compensation, thus the name `...Final...`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
normen662: You are correct.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
normen662: This just helps us to avoid a `SELECT qov(t) FROM T t` and to just use a  `T`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBOrQueryToUnionTest.java
normen662: So when we plan any operator that uses a comparison key, we iterate through all possible such comparison keys since they tend to give different resulting orderings. Sometimes, it does not matter like in this case. So we create both variations, but the cost model cannot really find a tie breaker so it resorts to plan hashes. Since we have a different structure underneath (`SELECT QOV(t) FROM T t` is missing) that hash is different and presumably orders differently in the cost model.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Compensation.java
alecgrieser: It would probably be good to have this in the javadoc

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
alecgrieser: I guess I should have been more specific, in that I was thinking that the refactored method could have the entire `if` statement

PR #3426, opened 2025-06-24
Address nullability of null-on-empty quantifiers and nested fields

alecgrieser:
This PR addresses two problems, both of which are necessary to resolve problems encountered by `FDBSimpleGraphQueryTest.testSubselectHasNullOnEmptyAndIsNullPredicate` when trying to integrate the planner rewrite rules (see: #3401).

The first is that `Quantifier.forEachWithNullOnEmpty` quantifiers were not taking into account the fact that they could return `null` in their return type. That meant that predicate simplification rules based on simplifying `IS (NOT) NULL` based on the field type could make incorrect inferences about the result of such quantifiers. This updates the result type method for that class so that we now take that into account.

The second is that if a type had non-nullable fields, but it was a nested type that lived inside a nullable field, then we’d report that field values on the child field were non-nullable. This is incorrect, as if the whole struct is `null`, field accesses on the child can return `null`. This could also result in the same rules making incorrect inferences about the type. This addresses this by looking at the types as we walk down field access paths, and then taking into account their nullability when we return the field value's return type.

I also looked into a different approach, where we’d just pro-actively mark all fields of a nullable type as nullable. This unfortunately has cross version compatibility problems with this verify:

https://github.com/FoundationDB/fdb-record-layer/blob/296b4e5b1e1196c92dd78ceda54bc9f82b1dbe4d/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryDefaultOnEmptyPlan.java#L79

Essentially, if the inner type has some (nested) non-nullable fields, then the nullable on empty value will differ in more ways than just the top-level type being nullable. This causes cross-version continuation tests to fail. We could solve _that_ by loosening the assert to only require that the types by the same (recursively) ignoring nullability, and then constructing a new type that combines field nullabilities.

There are also problems with the fact that this erases the nullability constraints entirely, which means that you wouldn’t be able to assert that, say, either the whole parent struct is null _or_ all of its non-null fields are not null.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBRecordStore.java
@@ -3636,7 +3636,7 @@ public class FDBRecordStore extends FDBStoreBase implements FDBRecordStoreBase<M
     @Nonnull
     private CompletableFuture<Boolean> markIndexReadable(@Nonnull Index index, boolean allowUniquePending) {
         if (recordStoreStateRef.get() == null) {
-            return preloadRecordStoreStateAsync().thenCompose(vignore -> markIndexReadable(index));
+            return preloadRecordStoreStateAsync().thenCompose(vignore -> markIndexReadable(index, allowUniquePending));
         }
 
         addIndexStateReadConflict(index.getName());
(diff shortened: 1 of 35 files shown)


normen662 (approved):
All good! Can you check `ComposeFieldValueOverFieldValueRule` if there is some impact there? I don't think so but it would be nice if you can stare at that for a second.

alecgrieser:
I don't think there should be an impact to that rule, though I'm not sure if you had something more specific you were thinking about.

The way I'm thinking about it, that rule performs a transform like:

```
fieldValue(fieldValue(v, path1), path2) -> fieldValue(v, path1.path2)
```

Under the new code, the nullability of the rhs will be `true` iff `v` or any field along `path1.path2` is nullable. On the lhs, `fieldValue(v, path1)` is nullable iff any field along `v` or `path1` is nullable, and `fieldValue(fieldValue(v, path1), path2)` is nullable iff `fieldValue(v, path1)` or any field along `path2` is nullable, which is to say that it's nullable iff and only if `v` is nullable or any field along `path1` is nullable or any field along `path2` is nullable, which is equivalent to `v` or any field along `path1.path2` being nullable.

So, it should be fine. Maybe we could use some more tests around that

PR #3427, opened 2025-06-24
Scoring and costing query predicates

### LINKED ISSUE #3419: Introduce costing algorithm that prefers simpler `QueryPredicate`s
The costing model should have a reasonable algorithm for picking `SelectExpression` with simpler `QueryPredicate`s.

The distinction can be based on this simple formula:

`SelectExpression(P1)` is considered to be less costly compared to a semantically-equivalent `SelectExpression(P2)` if `diameter(P1) < diameter(P2)`.

hatyo:
This introduces a simple algorithm for scoring predicate complexity based on the predicate's tree diameter (defined as the longest path between any two leaf nodes in the predicate tree, assuming an undirected tree, as per Cormen et al.'s *Introduction to Algorithms*). To facilitate this, a `diameter()` method has been added to the `TreeLike` class for calculating the diameter of a `SelectExpression`'s predicates.

Furthermore, a `PredicateComplexityProperty` is introduced, which returns the maximum predicate diameter found within a `RelationalExpression` tree. This property is then leveraged in the `RewritingCostModel` as a tie-breaker: when expressions have identical costs, the expression with the smallest maximum predicate diameter is preferred.

This fixes #3419.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateComplexityProperty.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ This property traverses a {@link RelationalExpression} to find the maximum diameter of any {@link QueryPredicate} associated
+ with either a {@link SelectExpression} or a {@link LogicalFilterExpression}.
+ */
+public class PredicateComplexityProperty implements ExpressionProperty<Integer> {
+    @Nonnull
+    private static final PredicateComplexityProperty PREDICATE_COMPLEXITY = ofScoredTyped(SelectExpression.class, LogicalFilterExpression.class);
```
normen662: you can simplify by looking at `RelationExpressionWithPredicates` only. That makes all that filter logic superfluous.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateComplexityProperty.java
hatyo: That's how I started the implementation, but I thought we wanted to be more selective about which expressions we want to look at when scoring their predicates.

normen662: `RelationExpressionWithPredicates` is all the expressions you want to look at which I believe is exactly as what you have here.

Issue #3390, opened 2025-06-11
`EvaluationContext` inconsistent constant binding dereferencing behavior


hatyo:
The `EvaluationContext` exhibits inconsistent behavior when attempting to dereference non-existent constant bindings.

https://github.com/FoundationDB/fdb-record-layer/blob/560b04893bf3d82ecf94198a17fa7ca9f7d6bcc8/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/EvaluationContext.java#L148-L154

If the user attempts to dereference a non-existent constant reference (which is invalid), the following scenarios may occur:

- If the `EvaluationContext` is empty or contains no constant bindings, the `constantMap` will be `null`, and an exception will be thrown.
- If the `EvaluationContext` contains some constant bindings (excluding the one being dereferenced), the `constantMap` will not be empty. Attempting to retrieve the non-existent item from it at line `154` will return `null` instead of throwing an exception.

This subtle approach allows certain bugs to be concealed. For instance, there is a bug that prevents the execution of a cached plan containing the `like` operator (see #3389). However, this bug only manifests when `like` is the only literal in the query. If the query contains any other literal, no exception will be thrown. Instead, the plan cannot be reused because the plan constraint requiring the type of the `like` pattern to be `String` will always fail since it receives `null` from the evaluation context. Consequently, the query will need to be replanned.

PR #3435, opened 2025-06-25
Address planner bugs encountered while integrating rewrite rules

alecgrieser:
This contains a number of planner fixes that were first identified when trying to integrate the planner rewrite rules (see: #3401). They are:

1. There were some values being leaked into the `Traversal` which would end up polluting the memo structure with additional values that weren't in the final graph. This adds a sanity check to validate the traversal contains precisely the information in the tree. There were a bunch of places where things got out of line, mainly due to something being added to the memo structure during rule execution (e.g., a new child) but that was never actually integrated into the final expression (e.g., because the expression was a duplicate of another in the reference). This adds a new "garbage collection" like mechanic, where new references are tracked during rule execution, and then after we're done, any that are still not in the memo are pruned.
2. Another memoizer issue: The first was that expressions could be re-used when they were in the wrong planning stage. This would be bad, and so the memoizer now looks for only references that match the expected stage.
3. Additionally, we could end up memoizing a reference that had an incorrect correlation set. Say, for example, that one expression was added that dropped a correlation (maybe due to predicate simplification). If we tried to re-use this reference in a place where that correlation wasn't guaranteed to be defined, which means we'd have an illegal graph. This addresses this by only looking for pre-existing references in the memoizer that have the right correlation set. This also allows us to update `DecorrelateValues` so that it can pick up any _expression_ that is not sideways correlated without having to worry about those correlations in any quantifier it pushes down
4. The `PredicateMultiMap` map hit an issue if one tried to add the same (semantically) equivalent predicate multiple times. This can happen if the user had the same predicate multiple times (like `a > 4 AND a > 4`) though it's actually harder to construct than one might think. The rewrite rules hit it because there was a query like `SELECT * FROM (SELECT * FROM T WHERE T.a > 4) X WHERE X.a > 4` and then select merge would be two copies of the same predicate at the same level. The builder for the `PredicateMultiMap` was always based on `LinkedIdenityMap`, so it used pointer equality. This updates the built `PredicateMultiMap` to be based on pointer equality as well instead of semantic equality (through immutable collections)
5. The `Reference` used to have a check to stop any expression that had a different correlation set from the reference as a whole from being in the memo. This is wrong, and it would result in way too many members of the memo. That check has been removed
6. Expression partitions can now match for the argmin of a tuple of properties. This allows us to pick the best child select with tiebreakers to favor expressions that are simpler (with tiebreakers)
7. There was a bug in the expression count property where it would always report the number of selects on references rather than the number of the expression requested. This has been updated to correctly return the expression type desired
8. The `InComparisonToExplodeRule` would allow us to do things like take an expression like `FROM T WHERE (T.a, T.b) IN ?list` and then explode the list and turn it into something like `FROM T, ?list x WHERE T.a = x._0 AND T.b = x._1`. That actually required that the original comparison had child values, and it wouldn't work if `(T.a, T.b)` could be re-expressed as some kind of simpler thing. This has been modified to work on more generic items of type `Record`
9. The exploration rules now only match against exploratory examples
10. Select merge now checks its children to see if pulling them up introduces duplicate quantifiers, and it uses the new `rebaseGraphs` method to rename any that are affected if it would do so
11. Constant object values now can be ignored during simplification is not in the evaluation context. This was mainly for tests, but it would be fine for actual code as well

As one can see, it's a lot of things. Most of them are actually fairly minor code changes, though some of them are a bit conceptually tricky

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Quantifier.java
@@ -193,6 +193,15 @@ public abstract class Quantifier implements Correlated<Quantifier> {
             return isNullOnEmpty;
         }
 
+        @Nonnull
+        @Override
+        public Type getFlowedObjectType() {
+            // If null on empty, then we may return null, so we need to update
+            // the nullability of our returned type to reflect that
+            Type baseType = super.getFlowedObjectType();
+            return baseType.overrideIfNullable(isNullOnEmpty);
+        }
+
         @Override
         @Nonnull
         public Builder<? extends Quantifier, ? extends Builder<?, ?>> toBuilder() {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/ConstantFoldingMultiConstraintPredicateRule.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/ConstantFoldingMultiConstraintPredicateRule.java
@@ -26,14 +26,12 @@ import com.apple.foundationdb.record.query.plan.cascades.matching.structure.Bind
 import com.apple.foundationdb.record.query.plan.cascades.predicates.ConstantPredicate;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.PredicateWithValueAndRanges;
 import com.apple.foundationdb.record.query.plan.cascades.predicates.RangeConstraints;
-import com.apple.foundationdb.record.query.plan.cascades.predicates.simplification.ConstantPredicateFoldingUtil.EffectiveConstant;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 
 import javax.annotation.Nonnull;
-
 import java.util.Optional;
 
 import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ListMatcher.exactly;
@@ -104,113 +102,45 @@ public class ConstantFoldingMultiConstraintPredicateRule extends QueryPredicateS
 
     @Nonnull
     private Optional<ConstantPredicate> foldConjunctionRangesMaybe(final Value lhs, @Nonnull final RangeConstraints rangeConstraints) {
-        final var rangeCategoriesBuilder = ImmutableSet.<RangeCategory>builder();
-        final var lhsOperand = EffectiveConstant.from(lhs);
-
         // degenerate case, give up.
         if (rangeConstraints.getComparisons().isEmpty()) {
             return Optional.empty();
         }
 
-        for (final var comparison : rangeConstraints.getComparisons()) {
-            if (comparison.getType().isUnary()) {
-                if (comparison.getType() == Comparisons.Type.IS_NULL) {
-                    switch (lhsOperand) {
-                        case NULL:
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_TRUE);
-                            break;
-                        case TRUE: // fallthrough
-                        case FALSE: // fallthrough
-                        case NOT_NULL:
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_FALSE);
-                            break;
-                        default:
-                            break;
-                    }
-                } else {
-                    rangeCategoriesBuilder.add(RangeCategory.UNKNOWN);
-                }
-            } else {
-                if (!(comparison instanceof Comparisons.ValueComparison)) {
-                    rangeCategoriesBuilder.add(RangeCategory.UNKNOWN);
-                    continue;
-                }
-                final var valueComparison = (Comparisons.ValueComparison)comparison;
-                final var rhsOperand = EffectiveConstant.from(valueComparison.getValue());
-                switch (comparison.getType()) {
-                    case EQUALS:
-                        if (lhsOperand == EffectiveConstant.NULL || rhsOperand == EffectiveConstant.NULL) {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_NULL);
-                            break;
-                        } else if (lhsOperand.equals(rhsOperand)) {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_TRUE);
-                            break;
-                        } else {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_FALSE);
-                            break;
-                        }
-                    case NOT_EQUALS:
-                        if (lhsOperand == EffectiveConstant.NULL || rhsOperand == EffectiveConstant.NULL) {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_NULL);
-                            break;
-                        } else if (!lhsOperand.equals(rhsOperand)) {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_TRUE);
-                            break;
-                        } else {
-                            rangeCategoriesBuilder.add(RangeCategory.SINGLETON_FALSE);
-                            break;
-                        }
-                    default:
-                        rangeCategoriesBuilder.add(RangeCategory.UNKNOWN);
-                }
-            }
+        final var predicateCategoriesBuilder = ImmutableSet.<ConstantPredicateFoldingUtil.PredicateCategory>builder();
+        for (Comparisons.Comparison comparison : rangeConstraints.getComparisons()) {
+            predicateCategoriesBuilder.add(ConstantPredicateFoldingUtil.foldComparisonMaybe(lhs, comparison));
+
         }
-        // if we have more than one distinct range then it is impossible to find a single value that satisfies all of them
-        final var rangeCategories = rangeCategoriesBuilder.build();
+        final var predicateCategories = predicateCategoriesBuilder.build();
+
+        // if we have only one distinct range, then that's the one we go with
+        if (predicateCategories.size() == 1) {
+            return Iterables.getOnlyElement(predicateCategories).getPredicateMaybe();
+        }
 
         // if at least one range is FALSE, the returned result must be FALSE
-        if (rangeCategories.contains(RangeCategory.SINGLETON_FALSE)) {
+        if (predicateCategories.contains(ConstantPredicateFoldingUtil.PredicateCategory.SINGLETON_FALSE)) {
             return Optional.of(ConstantPredicate.FALSE);
         }
 
         // if at least one range is UNKNOWN, give up.
-        if (rangeCategories.contains(RangeCategory.UNKNOWN)) {
+        if (predicateCategories.contains(ConstantPredicateFoldingUtil.PredicateCategory.UNKNOWN)) {
             return Optional.empty();
-        }
-
-        if (rangeCategories.size() == 1) {
-            final var singleRangeCategory = Iterables.getOnlyElement(rangeCategories);
-            switch (singleRangeCategory) {
-                case SINGLETON_TRUE:
-                    return Optional.of(ConstantPredicate.TRUE);
-                case SINGLETON_FALSE:
-                    return Optional.of(ConstantPredicate.FALSE);
-                case SINGLETON_NULL:
-                    return Optional.of(ConstantPredicate.NULL);
-                default:
-                    return Optional.empty();
-            }
         }
 
         // we have multiple disjoint singleton ranges then it is impossible to find a single value satisfying all
         // of them ...
 
         // ... unless we have a combination of TRUE and NULL then the result is NULL
-        if (rangeCategories.stream().noneMatch(p -> p == RangeCategory.SINGLETON_FALSE)) {
+        if (predicateCategories.stream().noneMatch(p -> p == ConstantPredicateFoldingUtil.PredicateCategory.SINGLETON_FALSE)) {
             // we must have TRUE and NULL
-            Verify.verify(rangeCategories.contains(RangeCategory.SINGLETON_TRUE));
-            Verify.verify(rangeCategories.contains(RangeCategory.SINGLETON_NULL));
+            Verify.verify(predicateCategories.contains(ConstantPredicateFoldingUtil.PredicateCategory.SINGLETON_TRUE));
+            Verify.verify(predicateCategories.contains(ConstantPredicateFoldingUtil.PredicateCategory.SINGLETON_NULL));
             return Optional.of(ConstantPredicate.NULL);
         }
 
         // ... ok range is impossible, bailout.
         return Optional.of(ConstantPredicate.FALSE);
-    }
-
-    private enum RangeCategory {
-        SINGLETON_NULL,
-        SINGLETON_TRUE,
-        SINGLETON_FALSE,
-        UNKNOWN
     }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/ConstantFoldingPredicateWithRangesRule.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/predicates/simplification/ConstantFoldingPredicateWithRangesRule.java
@@ -92,6 +92,6 @@ public class ConstantFoldingPredicateWithRangesRule extends QueryPredicateSimpli
         final var booleanSingletonRange = call.getBindings().get(booleanSingletonRangeMatcher);
         final var comparison = (Iterables.getOnlyElement(booleanSingletonRange.getComparisons()));
         final var lhsValue = root.getValue();
-        ConstantPredicateFoldingUtil.foldComparisonMaybe(lhsValue, comparison).ifPresent(call::yieldResult);
+        ConstantPredicateFoldingUtil.foldComparisonMaybe(lhsValue, comparison).getPredicateMaybe().ifPresent(call::yieldResult);
     }
 }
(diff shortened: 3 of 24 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -346,10 +370,18 @@ public Reference memoizeExploratoryExpressions(@Nonnull final Collection<? exten
                             .map(traversal::getParentRefPaths)
                             .collect(ImmutableList.toImmutableList());
 
+            // Only look for references that are: (1) in the right planner stage and (2) have appropriate correlations.
+            // It's important that there aren't any correlations as if we choose to re-use a reference that
+            // contains a correlation that the original set of expressions don't have, then we might try to use it
+            // in a place that we are not allowed to
+            final Predicate<Traversal.ReferencePath> referencePathFilter = path ->
```
normen662: Can you write a helper method and refer to that in the filters rather than an anonymous lambda?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -423,17 +455,26 @@ private Reference memoizeLeafExpressions(@Nonnull final Collection<? extends Rel
...
     }
 
+    @Nonnull
+    private Set<CorrelationIdentifier> correlatedTo(@Nonnull Collection<? extends RelationalExpression> expressions) {
+        if (expressions.isEmpty()) {
+            return ImmutableSet.of();
+        } else if (expressions.size() == 1) {
+            return Iterables.getOnlyElement(expressions).getCorrelatedTo();
+        } else {
+            return expressions.stream()
```
normen662, line 476: do you think it would be better to use `Sets.union(...)` here since it avoids the copying around?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
```diff
@@ -1004,6 +1009,12 @@ public void execute() {
         protected void executeRuleCall(@Nonnull CascadesRuleCall ruleCall) {
             ruleCall.run();
 
+            //
+            // During rule construction, references can be added to the memoizer that don't end up in
+            // any final expression. Prune them here
+            //
+            ruleCall.pruneUnusedNewReferences();
```
normen662, line 1016: Can't you already prune the new references when we realize that the `insert` into the memo yielded a duplicate? If so, you don't have to keep the new references around.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ReferenceMatchers.java
```diff
@@ -82,4 +82,12 @@ public static <R extends Reference, E extends RelationalExpression> BindingMatch
                 Extractor.of(Reference::getExploratoryExpressions, name -> "allMembers(" + name + ")"),
                 downstream);
     }
+
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static <E extends RelationalExpression> BindingMatcher<Reference> exploratoryMember(@Nonnull final BindingMatcher<E> downstream) {
+        return TypedMatcherWithExtractAndDownstream.typedWithDownstream(Reference.class,
+                Extractor.of(Reference::getExploratoryExpressions, name -> "allMembers(" + name + ")"),
```
normen662:
```suggestion
                Extractor.of(Reference::getExploratoryExpressions, name -> "exploratoryMember(" + name + ")"),
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/ExpressionCountProperty.java
```diff
@@ -60,7 +60,7 @@ private ExpressionCountProperty(@Nonnull final Predicate<? super RelationalExpre
     @Nonnull
     @Override
     public ExpressionCountVisitor createVisitor() {
-        return new ExpressionCountVisitor(filter, isTracked);
+        return new ExpressionCountVisitor(filter, this);
```
normen662: If you want that, you can just make the inner class non-static.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -134,8 +137,10 @@
 public class DecorrelateValuesRule extends ExplorationCascadesRule<SelectExpression> {
     // TODO: This could use filtered expression partitions, but we have to make modifications to the test infrastructure to ensure there are final children
     @Nonnull
-    private static final BindingMatcher<RelationalExpression> baseExpressionMatcher = TypedMatcherWithPredicate.typedMatcherWithPredicate(RelationalExpression.class,
-            expr -> CardinalitiesProperty.Cardinalities.exactlyOne().equals(CardinalitiesProperty.cardinalities().evaluate(expr)));
+    private static final BindingMatcher<Quantifier.ForEach> rangeOneMatcher = forEachQuantifierWithoutDefaultOnEmptyOverRef(
+            TypedMatcherWithPredicate.typedMatcherWithPredicate(Reference.class,
```
normen662: Hmmm, can we express the `ref.getAllMemberExpressions().stream().anyMatch(expr -> expr instanceof TableFunctionExpression` using a matcher? I guess the complications comes from the cardinalities part.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
```diff
@@ -196,21 +205,33 @@ public void onMatch(@Nonnull final ExplorationCascadesRuleCall call) {
...
+                        // We can push down the entire quantifier
+                        qunsToPushDownBuilder.put(alias, qun);
+                    } else {
+                        // There is at least one expression in the quantifier that is correlated to one of the
+                        // siblings. Extract out only the variations that are not correlated to a sibling
+                        Set<RelationalExpression> uncorrelatedExpressions = qun.getRangesOver().getExploratoryExpressions().stream()
+                                .filter(expr -> correlatedToNone(expr, childAliases))
+                                .collect(LinkedIdentitySet.toLinkedIdentitySet());
+                        Verify.verify(uncorrelatedExpressions.contains(valueExpression), "target value expression should be in the uncorrelated subset of quantifier expressions");
+                        Quantifier newQun = qun.overNewReference(call.memoizeExploratoryExpressions(uncorrelatedExpressions));
```
normen662, line 231: wait! wouldn't this `memoizeExploratoryExpressions` give you that reference you just grabbed it from? Or is that exactly the check on the correlation sets in `CascadesRuleCall` that prevents this?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveRangeOneRule.java
```diff
@@ -37,7 +37,7 @@
 
 import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.AnyMatcher.any;
 import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.QuantifierMatchers.forEachQuantifierOverRef;
-import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ReferenceMatchers.members;
+import static com.apple.foundationdb.record.query.plan.cascades.matching.structure.ReferenceMatchers.exploratoryMember;
```
normen662, line 40: Is this rule still needed?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java
```diff
@@ -66,7 +76,11 @@ public class SelectMergeRule extends ImplementationCascadesRule<SelectExpression
 
     @Nonnull
     private static final BindingMatcher<ExpressionPartition<RelationalExpression>> childPartitionsMatcher =
-            argmin(ExpressionCountProperty.selectCount(), childExpressionMatcher);
+            argmin(ExpressionsPartitionMatchers.<RelationalExpressionWithPredicates>comparisonByPropertyList(
```
normen662, line 79: clever!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/DereferenceConstantObjectValueRule.java
```diff
@@ -54,6 +54,14 @@ public DereferenceConstantObjectValueRule() {
     public void onMatch(@Nonnull final ValueSimplificationRuleCall call) {
         final var constantObjectValue = call.getBindings().get(rootMatcher);
         final var evaluationContext = call.getEvaluationContext();
+        if (!evaluationContext.containsConstantBinding(constantObjectValue.getAlias(), constantObjectValue.getConstantId())) {
```
normen662, line 57: Isn't this an error?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/InComparisonToExplodeRule.java
```diff
@@ -202,14 +205,20 @@ public void onMatch(@Nonnull final ExplorationCascadesRuleCall call) {
      * This method creates an equality predicate for all constituent parts of a tuple.
      */
     @Nonnull
-    private static List<QueryPredicate> createSimpleEqualities(@Nonnull final Value value,
-                                                               @Nonnull final Quantifier.ForEach newQuantifier) {
-        List<Value> valueChildren = ImmutableList.copyOf(value.getChildren());
+    private static List<QueryPredicate> createSimpleEqualitiesForRecordTypeValue(@Nonnull final Value value,
+                                                                                 @Nonnull final Quantifier.ForEach newQuantifier) {
+        Verify.verify(value.getResultType().isRecord(), "value should be of type record and not %s", value.getResultType());
```
normen662, line 208: This all maybe can use `Values.deconstructRecord(...)`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
alecgrieser: Done

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/DereferenceConstantObjectValueRule.java
hatyo: Indeed, but let's address this later, as in #3390 fix + tests.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
alecgrieser: One problem with `Sets.union` is that it's strictly for 2 sets, not `n`. I tried using `reduce`, with something like:

```java
return expressions.stream()
    .map(RelationalExpression::getCorrelatedTo)
    .reduce(ImmutableSet.of(), (s1, s2) -> s1.isEmpty() ? s2 : (s2.isEmpty() ? s1 : Sets.union(s1, s2)));
```

Which works, though I'm not sure how clear it is. But more to the point, for references, I'm actually expecting that most of the expressions to have lots of elements in common, which isn't done proactively by `Sets.union`. So, copying here uses more memory, but the benefit is that we don't have to re-dedupe multiple times. I've added some commentary about that choice.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
alecgrieser: I did initially try this, but it wasn't actually sufficient, and more to the point, it led to verification errors in `verifyChildrenMemoized`. You can see some of them here: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15764888894

There are maybe two problems here:

1. First, `SelectDataAccessRule`. That method makes calls to the memoizer when it creates scans, and then it doesn't do a great job of keeping track of which ones it actually uses and which ones it doesn't. So without some kind of garbage collection phase, it will leave un-referenced data in the memo structure
2. Second, there was at least one rule that would end up creating multiple parent references that pointed to the same (new) child reference. If the first instance ended up being excluded from the graph, we could end up pruning them a bit too eagerly, which is what led to the `verifyChildrenMemoized` errors. One thing I'm not sure about is whether this is a sign that we're actually doing something wrong there, in that we may have multiple final expressions that share a reference. Those would all be references in the same expression, so it...may be okay? If that's not okay, I can file an issue to look into that

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/ExpressionCountProperty.java
alecgrieser: Yeah, maybe that makes sense. Non-static inner classes can be a bit of a code smell. I've gone ahead and done that as a separate commit so we can see how we like it (or remove it)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: Maybe you have a suggestion for this. Getting this to be written with a matcher isn't too hard. You can even see it on main: https://github.com/FoundationDB/fdb-record-layer/blob/331b53d349aaf52e57c518adacfe83e5e3d3f157/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java#L137

The problem is that the way that's written, if we have a reference over multiple `range(1)` expressions, we'll end up creating a new match for each child. Protecting against that was the genesis behind this test assertion: https://github.com/alecgrieser/fdb-record-layer/blob/1fda78545fe07b3f2241cd8d4a9c500ccbc82429/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRuleTest.java#L794

I think the intended fix for that is to try and match the expression partition rather than the expression directly. The problem with that is that the _tests_ (at least) don't necessarily have final values here, so we don't have final expressions to roll up, even if the real rules don't.

If there's some other matcher trickery that will avoid the duplicate matching, I'd be happy to switch to it. I've added a note for now as to why it is the way it is

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/DecorrelateValuesRule.java
alecgrieser: This has been resolved, but yeah, for any one else looking, this would be the case if it weren't for the correlation set check in the memoizer

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveRangeOneRule.java
alecgrieser: I'm not sure. There may be some niche use cases that require it. One that came to mind is that if there's a `SelectMergeRule` that runs over a set of expressions that includes a values box, we might end up with an extra `range(1)` that we could otherwise remove. It should be the case that `DecorrelateValuesRule` removes most of the `range(1)` boxes as it goes, but I'm not certain there isn't some case that might require it. I sort of think we should keep it for now, see if it makes any difference when integrated further, and then potentially remove it

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/simplification/DereferenceConstantObjectValueRule.java
alecgrieser: Maybe. I agree that it's an error during query _execution_ to have a COV that does not have a referenced value, but it's less clear to me that it should be an error during query _planning_, though it can (as here) result in us not applying every optimization. The actual genesis of this here is that we have some tests that run but don't bother setting the evaluation context for some cov's they create, mainly because it's not relevant for them. In general, I actually think it would make sense for certain kinds of prepared statements that don't care about the missing optimization to leave out a constant from the evaluation context

I've updated the comment below this highlighted line to reflect the fact that this (at some level) comes from tests. Note that there is already text there with a use case that (imo) makes sense that would explain why it's not set

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/InComparisonToExplodeRule.java
alecgrieser: Ah, yes, indeed, good call. Updated

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
normen662: Thanks, I understand but this really should not happen I think. Yes, an issue would be appreciated.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/RemoveRangeOneRule.java
normen662: Okay.

PR #3436, opened 2025-06-25
Avoid Boolean simplification that would lead to expressions like `x IS TRUE` being simplified to `FALSE`

alecgrieser:
The boolean value tests in #3401 were failing in this PRB: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15878200321/attempts/1#summary-44771112943

I tracked this down, and the problem ended up being in some of the constant folding logic in `ConstantFoldingMultiConstraintPredicateRule`. Essentially, there was a check to see if two Boolean comparisons were of the same type (e.g., `TRUE` and `TRUE`), and if they were, then they'd be simplified to `ConstantPredicate.TRUE` or `ConstantPredicate.FALSE`. However, there was a problem with the logic in that it didn't handle the `NOT_NULL` and `UKNOWN` range types, which it would note as different from, say, `TRUE`, and then mark the whole predicate as `FALSE`.

This would happen when given SQL predicates like `x IS TRUE` or `x IS FALSE`. The `IS` there is important, because it represents two range comparisons, one on `NOT_NULL` and the other on `IS TRUE/FALSE`. That meant that it would use `ConstantFoldingMultiConstraintPredicateRule` to simplify the predicate instead of `ConstantFoldingValuePredicateRule`. That other rule would use `ConstantFoldingUtil.foldComparisonMaybe`, which appears to have logic handling that case.

To avoid this problem, I modified `ConstantFoldingMultiConstraintPredicateRule` so that it uses `ConstantFoldingUtil.foldComparisonMaybe` instead of its own logic. The rule then is now only responsible for `AND`ing the different comparison ranges together.


alecgrieser:
This currently is built on top of #3426 because I modified some of the same test classes in that PR that I would need here. My hope is that we'll merge #3426 first and then I can rebase this one

hatyo (approved):
LGTM.

Issue #3432, opened 2025-06-25
Nullability problems with FieldValue and null-on-empty quantifiers


alecgrieser:
Two issues were identified with #3401 when we tried to integrate the planner rewrite rules into the planner. They both involved cases where we were getting null values with types that claimed to be non-nullable.

The first involves `FieldValue`s. If a field `Type` is not nullable, that only really means that the field will not be null given the fact that its parent is not null. So, for example, something like:

```sql
CREATE TYPE lower AS STRUCT (
    x STRING NOT NULL
)
CREATE TYPE upper AS STRUCT (
   y lower
)
```

Then the field value `_.y.x` should be nullable, as the nullability of `y` means that accesses to `x` are also nullable (unless one can prove that a particular reference to `y` is not nullable). We should propagate that information up to avoid incorrectly transforming a predicate like `y.x IS (NOT) NULL` into constant predicates.

The second involves `Quantifer.ForEach` instances with `nullOnEmpty`. Those can return `null` if the underlying is empty. The result type should therefore record that it returns a nullable type. Likewise, the `RecordQueryDefaultOnEmptyPlan` (effectively the implementation of that quantifier) should record that it can return null in its return type.

The two of those together are leading to a test failure of this test:

https://github.com/FoundationDB/fdb-record-layer/blob/29cfcd0c8de0ebbd0b4be339fe169d1c1e0dcd44/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/FDBSimpleQueryGraphTest.java#L334-L356

The simplification engine can be tricked into turning the `IS NULL` predicate into `false` because the underlying field is not nullable even though it's coming from a quantifier that can return null.

PR #3438, opened 2025-06-26
Added ability to hand in connection options through the yaml framework and temporary tests that disable all rewrite rules

normen662:
…ork and creating a set of temporary test case yamls that disable all rewrite rules


fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -346,6 +358,16 @@ public Options build() {
         }
     }
 
+    @Nullable
+    private static Object parseStringOption(@Nonnull final Name name, String valueAsString) throws SQLException {
+        for (OptionContract contract : Objects.requireNonNull(OPTIONS).get(name)) {
+            if (contract instanceof TypeContract<?>) {
```
alecgrieser, line 364: I believe the recently added `DISABLED_PLANNER_RULES` option does not have a type contract. Instead it has a `CollectionContract` which is sort of like a `TypeContract`, except it has a child contract to apply to each element. Given that we're not using this option in the tests, we could probably leave without this one working (for now), or we could extend this check to also allow for `CollectionContract`s

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/TestBlock.java
```diff
@@ -101,6 +104,7 @@ public final class TestBlock extends ConnectedBlock {
     static final String OPTION_CONNECTION_LIFECYCLE = "connection_lifecycle";
     static final String OPTION_CONNECTION_LIFECYCLE_TEST = "test";
     static final String OPTION_CONNECTION_LIFECYCLE_BLOCK = "block";
+    static final String OPTION_CONNECTION_OPTIONS = "connection_options";
```
alecgrieser, line 107: Should this go into `FileOption`s as well? At this point it may be late, but it would be nice to be able to set these kinds of things at a file level

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -316,6 +322,12 @@ private Builder() {
             optionsMap = Maps.newHashMap();
         }
 
+        @Nonnull
+        public Builder withOptionFromString(Name name, String valueAsString) throws SQLException {
```
alecgrieser, line 326: Did this end up getting used? From what I can tell, the logic in `TestBlock` is letting YAML parse the string into a type

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
```diff
@@ -0,0 +1,265 @@
+/*
+ * YamlIntegrationTests.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2021-2024 Apple Inc. and the FoundationDB project authors
```
alecgrieser:
```suggestion
 * Copyright 2021-2025 Apple Inc. and the FoundationDB project authors
```

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
```diff
@@ -0,0 +1,265 @@
+/*
+ * YamlIntegrationTests.java
```
alecgrieser:
```suggestion
 * YamlIntegrationDisabledPlannerRewritesTests.java
```

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
```diff
@@ -0,0 +1,265 @@
...
+ * limitations under the License.
+ */
+
+import com.apple.foundationdb.relational.yamltests.MaintainYamlTestConfig;
+import com.apple.foundationdb.relational.yamltests.YamlTest;
+import com.apple.foundationdb.relational.yamltests.YamlTestConfigFilters;
+import org.junit.jupiter.api.TestTemplate;
+
+/**
+ * Class covering the standard integration tests specified by yamsql files.
```
alecgrieser: It may be good to note here that these are essentially all copies of the tests in `YamlIntegrationTests` but with the planner rewrites disabled in order to test that the rewriter can be safely flipped on and off as we roll it out. (And probably noting that this file is supposed to be temporary, as we don't think we'll need it once the rewriter is more mature.)

yaml-tests/src/test/resources/showcasing-tests.yamsql
```diff
@@ -168,6 +168,9 @@ test_block:
     # Like the options at the top of the file, a supported_version can be specified for
     # an individual test_block. This overrides the version specified in the file
     supported_version: 4.0.560.0
+    # specifies the connection options. This may not be supported by every connection type
+    connection_options:
+      DISABLE_PLANNER_REWRITING: false
```
alecgrieser: Is this just to make sure we can do this? Does this also mean that we end up skipping the whole file even though it's only this block that sets the connection option?

yaml-tests/src/test/resources/disabled-planner-rewrites/uuid-proto.yamsql
```diff
@@ -0,0 +1,96 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+setup:
+  connect: "jdbc:embed:/__SYS?schema=CATALOG"
+  steps:
+    - query: drop schema template if exists uuid_proto_template
+    - query: drop database if exists /FRL/UUID_PROTO_YAML
+    - query: create database /FRL/UUID_PROTO_YAML
```
alecgrieser, line 26: In theory, we should pick a different database path for this test. There are a couple of others (mostly the `-proto` and `-metadata` tests, I think) that hard-code their database URI. I think it only actually matters if we try and parallelize running two test files at the same time, which we currently don't do though we've wanted to at certain points, as they need different paths to ensure isolation

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
```diff
@@ -0,0 +1,265 @@
...
+        runner.runYamsql(PREFIX + "/aggregate-index-tests-count-empty.yamsql");
+    }
+
+    @TestTemplate
+    public void maxRows(YamlTest.Runner runner) throws Exception {
+        runner.runYamsql(PREFIX + "/maxRows.yamsql");
+    }
+
+    @TestTemplate
+    public void nested(YamlTest.Runner runner) throws Exception {
```
alecgrieser, line 129: PR #3432 added some new files, `nested-with-nulls.yamsql` and `nested-with-nulls-proto.yamsql`. It would probably be good to have them added to this test suite, especially as the latter actually has different behavior when the simplification engine is used

alecgrieser (changes requested):
Mostly looks good. There are a few documentation things, though we could probably go with it as is (once we have #3435 in to address the failures, I think)

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
```diff
@@ -273,24 +274,28 @@ public List<Block> getFinalizeBlocks() {
...
         String fileName;
         if (resourcePath.contains("/")) {
             final String[] split = resourcePath.split("/");
             fileName = split[split.length - 1];
         } else {
             fileName = resourcePath;
         }
-        if (e instanceof YamlExecutionError) {
-            final var oldStackTrace = e.getStackTrace();
+        try {
```
hatyo: this changes the backtrace and makes the exception appear to originate from here, right?. Can you just use `instanceOf` similar to how it was done before?

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
```diff
@@ -0,0 +1,265 @@
...
+ * Note: Use {@link MaintainYamlTestConfig} using {@link YamlTestConfigFilters#CORRECT_EXPLAIN_AND_METRICS} or similar
+ *       to correct explain strings and/or planner metrics. That annotation works both on class and on method level.
+ */
+@YamlTest
+public class YamlIntegrationDisabledPlannerRewritesTests {
+    private static final String PREFIX = "disabled-planner-rewrites";
+
+    @TestTemplate
+    public void showcasingTests(YamlTest.Runner runner) throws Exception {
+        runner.runYamsql(PREFIX + "/showcasing-tests.yamsql");
```
hatyo, line 40: I am not sure it is a good idea to duplicate the test files, considering that one can easily forget to update both files. Can you instead create a single test that iterates over the original test files, and run them with that option instead?

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
normen662: Added support for that. Thank you!

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/TestBlock.java
normen662: It could, though `FileOption` does not have a way to dictate defaults to the individual blocks. It feels like that something like that could/should be done, but not as part of this PR.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
normen662: Discussed offline. No, it's currently not being used but it could get used fairly soon once we want to pass around options from client to server. I wrote this first, not understanding that the YAML parser actually does its best to interpret the correct type, so it's probably ok for this case. Still leaving it in for now.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
normen662: done.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
normen662: done

yaml-tests/src/test/resources/showcasing-tests.yamsql
normen662: you are right, I made this a comment to restore the original behavior while still documenting what someone could do to set the options.

yaml-tests/src/test/resources/disabled-planner-rewrites/uuid-proto.yamsql
normen662: As far as I can tell this is the same problem when running `YamlIntegrationTests`. I'll leave this for another time.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
normen662: done.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
normen662: That does not work, as many files have plan changes in them. This is temporary, so we should not have to update both for an extended period of time.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
normen662: No, it does not fill in the stack trace when you `throw`. That happens only during the creation of a new exception. I opted to use this way as this allows me to do exception pattern matching as I don't have to down cast the exception. I picked up this style from Crunch/Hadoop back in the days and find it quite nice actually.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
normen662: done.

yaml-tests/src/test/java/YamlIntegrationDisabledPlannerRewritesTests.java
hatyo: You're right.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
hatyo: ok.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/options/CollectionContract.java
```diff
@@ -24,21 +24,26 @@
...
  * of collection type and (2) every element in the collection satisfies the "element
  * contract" over which the contract is defined. For example, the child contract
  * may be that each element is of a given type or within a particular range.
+ * @param <T> the type parameter of the collection
  */
-public class CollectionContract implements OptionContract {
+public class CollectionContract<T> implements OptionContract, OptionContractWithConversion<Collection<T>> {
     @Nonnull
-    private final OptionContract elementContract;
+    private final TypeContract<T> elementContract;
```
alecgrieser, line 44: In theory, this should be an `OptionContractWithConversion<T>`, I think, which would let us do things like nesting. Though right now the serialization format of collection contracts don't support that.

This is probably fine for now, though

alecgrieser (approved):

PR #3439, opened 2025-06-26
Remove MaintainYamlTestConfig annotation

ScottDugas:
These should not be committed, and was caught in a recent nightly test run.

### DIFF
--- a/yaml-tests/src/test/java/YamlIntegrationTests.java
+++ b/yaml-tests/src/test/java/YamlIntegrationTests.java
@@ -285,7 +285,6 @@ public class YamlIntegrationTests {
     }
 
     @TestTemplate
-    @MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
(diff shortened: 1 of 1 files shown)


alecgrieser:
It looks like one of the plans in `literal-tests.yamsql` changed and the results were never checked in, and so PRBs were passing just because they were re-writing the explain every time

alecgrieser:
Believe it or not, this actually breaks our mixed mode tests. See: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15925583405/job/44921988612

The problem is that with this config, we end up generating an "embedded" config (to do the corrections). That ends up hitting an error during the run when it tries to determine the associated version of a test run. The "embedded" configs don't have a test associated with it, so we get an error

ScottDugas:
This was removed in https://github.com/FoundationDB/fdb-record-layer/pull/3448/files#diff-737b24ee2e542fa2d6bd98ea9f9d0ad7b51262c6b7f0fb60b4738b9f5079a6e8

PR #3440, opened 2025-06-26
Separate out Lucene tests so they run separately in PRB

alecgrieser:
I was looking through recent PRB runs, and it seemed like the time to run other-tests was now around twice the time to run the core-tests. Most of the time seemed to be coming from lucene-tests, so I wanted to experiment with separating those out. This hopefully should decrease our total PRB time

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -64,6 +64,40 @@ jobs:
           include-hidden-files: true
           retention-days: 1
 
+  lucene-tests:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
+      pull-requests: write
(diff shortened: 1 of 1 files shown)


alecgrieser:
Looking at the timings here, the other- plus lucene-tests took 30 minutes together, which is consistent with some other PR runs I've seen. The core-tests took 15 minutes, which is about equal to the other two tests jobs (and the style job), so this seems like a pretty even split to me

ScottDugas (commented):
I feel like we should validate that a failing lucene test fails the PRB build. I created a draft PR to test this: https://github.com/FoundationDB/fdb-record-layer/pull/3444

ScottDugas (approved):

PR #3443, opened 2025-06-26
Fix eager constant folding in `RelOpValue` and other fixes

### LINKED ISSUE #3441: Fix eager predicate folding logic in `toQueryPredicate`
In `RelOpValue::toQueryPredicate` we have the following logic:

https://github.com/FoundationDB/fdb-record-layer/blob/c323aa8a9e61b89c24e009203dccdf4c759ec0c7/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java#L147-L150

This logic is incorrect as it assumes a predicate is constant if the predicate does not correlate to the innermost alias, which is not true, since the predicate can for example have an outer correlation that breaks this assumption.

If the predicate has an outer correlation, and is then passed to `tryBoxSelfAsConstantPredicate` it will immediately fail because compile-time evaluation of the predicate will not be able to find that outer correlation.

hatyo:
This change addresses a bug related to the premature folding of constant `QueryPredicate`s within `RelOpValue` legs. The previous logic incorrectly identified a `QueryPredicate` as constant if its correlations didn't include the innermost aliases. This led to runtime exceptions ("missing binding") when the `QueryPredicate` had only correlations to outer quantifiers, as the predicate would be evaluated at compile time before these outer bindings were available.

The fix modifies the constant check: a `QueryPredicate` is now only considered constant if it has **no** correlations at all. Furthermore, the interface is improved to accept a list of inner aliases instead of just the innermost one, signaling proper handling of join predicates internally.

This PR also includes several smaller improvements:

*   Function parameters are now nullable by default, aligning with the default nullability of relation attributes.
*   Double-normalization of function names during schema template metadata reading is prevented.
*   Unnecessary application of the entire `OR` predicate as residual is avoided when an exact match exists with the match candidate's `OR` predicate.
*   Moved extremely fine-grained logging messages in `MaxMatchMap` and `DebuggerWithSymbolTables` from `log` to `trace`. This is necessary to avoid potential slowdowns in planning with `debug` level (which is enabled by default as per the global log settings in `gradle/scripts/log4j-test.properties`).
*   Folding constant promotion rules are applicable for any child value, not only `LiteralValue`.
*   Modify predicate costing algorithm, it uses now a special version of `diameter` that adds all the edges between a node and its left / right heights instead of just 2 at max.

This fixes #3441.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
```diff
@@ -162,7 +161,8 @@ public Optional<QueryPredicate> toQueryPredicate(@Nullable final TypeRepository
             if (leftChildCorrelatedTo.isEmpty() && rightChildCorrelatedTo.isEmpty() && typeRepository != null) {
                 return tryBoxSelfAsConstantPredicate(typeRepository);
             }
-            if (rightChildCorrelatedTo.contains(innermostAlias) && !leftChildCorrelatedTo.contains(innermostAlias)) {
+            if (localAliases.containsAll(rightChildCorrelatedTo) &&
```
normen662, line 164: Hmm, so what happens if `localAliases` does not contain all of `right`, and/or none of `left`? You just flip over to the else branch. That creates an asymmetry I do not like. On the other hand I don't have something better for you. Any ideas?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/InComparisonToExplodeRule.java
```diff
@@ -214,7 +215,7 @@ private static List<QueryPredicate> createSimpleEqualitiesForRecordTypeValue(@No
         for (int i = 0; i < fieldValues.size(); i++) {
             final Value fieldValue = fieldValues.get(i);
             BooleanValue currentVal = (BooleanValue) new RelOpValue.EqualsFn().encapsulate(List.of(fieldValue, comparandValueChildren.get(i).getValue()));
-            Optional<QueryPredicate> currentQueryPredicate = currentVal.toQueryPredicate(null, Quantifier.current());
+            Optional<QueryPredicate> currentQueryPredicate = currentVal.toQueryPredicate(null, ImmutableSet.of(Quantifier.current()));
```
normen662: can you create an override for these callers that only have exactly one alias? So we don't have to explicitly create `ImmutableSet.of(...)` every time.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -328,19 +328,58 @@ default int height() {
...
+     * Using a custom connection function can be beneficial for scoring or evaluating the tree based on criteria beyond
+     * the traditional diameter definition.  However, be aware that modifying the connection cost may cause the returned
+     * value to deviate from the strictly defined diameter of the tree.
+     *
+     * @param customConnectingEdgeFunc A function that determines the connection cost between the current root and the
+     *                                 heights of its left and right subtrees. This function allows for customizing
+     *                                 the diameter calculation based on tree-specific properties.
+     * @return The diameter of the tree, calculated using the specified connection function.
+     */
+    default int diameterWithConnectingEdgeFunction(Function<TreeLike<T>, Integer> customConnectingEdgeFunc) {
```
normen662, line 373: `IntFunction`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
```diff
@@ -328,19 +328,58 @@ default int height() {
...
+     * Returns the diameter of the tree. i.e. the maximum path between two leaves in the tree.
+     *
      * @return the diameter of the tree.
      */
     default int diameter() {
+        return diameterWithConnectingEdgeFunction(current -> Math.min(Iterables.size(current.getChildren()), 2));
+    }
+
+    /**
+     * Calculates the diameter of the tree, incorporating the number of children of each node in the calculation.
```
normen662, line 340: Consider to do this: Instead of adding `2` or the arity of the node, I think the proper way to do this would be:
1. calculate each (height/depth, diameter) for each child
2. for each child `i, j` pairing, compute the max of `(diameter_i, diameter_j, height_i + height_j + (j -i) + 2)` -- use the largest of these pairings.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/RelOpValue.java
hatyo: Yeah, I think the else branch here means: give up and leave the predicate as it. Let's discuss offline the alternatives.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
hatyo: I am not sure what you mean by `(j-i) + 2` and why is it potentially better, if by `i` and `j` you mean the index of child, it doesn't make too much sense to me.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/TreeLike.java
hatyo: Hmm ... have you reviewed the latest commit? the function takes in a `TreeLike<T>` as an argument now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/InComparisonToExplodeRule.java
hatyo: done.

normen662 (approved):

hatyo:
@normen662 I pushed a new commit which fixes the expected plan in `literal-tests.yamsql`: 
previously:
```sql
SCAN(<>) | FILTER @c6 EQUALS coalesce_long(@c10, @c6)
```
now:
```sql
SCAN(<>) | FILTER coalesce_long(@c10, @c6) EQUALS @c6
```
This is the result of doing the flipping of the predicate terms consistently.

arnaud-lacurie (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/AbstractArrayConstructorValue.java
```diff
@@ -171,7 +171,9 @@ private static Type resolveElementType(@Nonnull final Iterable<? extends Typed>
     @Nonnull
     private static List<? extends Value> injectPromotions(@Nonnull Iterable<? extends Value> children, @Nonnull final Type elementType) {
         return Streams.stream(children)
-                .map(child -> PromoteValue.inject(child, elementType))
+                .map(child -> child.getResultType().nullable().equals(elementType.nullable())
```
normen662, line 174: This is dealt with in `PromoteValue.inject(...)` already. Why doesn't that already work as desired?

PR #3450, opened 2025-06-27
Update code version before running tests in release build

### LINKED ISSUE #3449: Release job does not choose right set of external server versions
The release job currently does not select the right set of external server versions for the YAML tests. In particular, the logic goes like:

1. Download a bunch of external servers (so far so good)
2. Also construct a local external server based off of the current code
3. Pick the most recent of the downloaded and locally constructed servers for mixed mode and external server testing
4. Run the tests against both of those versions

The problem is in step 3. The logic looks for two different servers, one of which has a version equal to the current code version and one which looks for an older version. The way we handle versions is that we bump the version after we've run the tests but before we publish. We then check that version in, which means that when the most recent release is version `x`, the `gradle.properties` file on `main` says that we're on version `x`. Normal builds append `-SNAPSHOT` to the build number, so the "pick two versions" logic will typically end up selecting `x` and `x-SNAPSHOT`, which works fine. But in the release build specifically, we run with the `-PreleaseBuild` flag, so we _don't_ append the `-SNAPSHOT` number, and so we end up picking two builds with version `x`.

We could fix this a couple of different ways. First, we could modify the way we update versions so that the final version that gets checked in is the version of the _upcoming_ release instead of the _previous_ release. This would also align our versions to the way most systems expect `SNAPSHOT` and non-`SNAPSHOT` builds are ordered. A second approach would be to increment the version in the test section before we run the tests, and then we would run `x` against `x + 1`. That would align our testing job closer to the publishing job.

This broke one of our release builds: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15925583405 This became relevant because there were changes in the current version that necessitated disabling tests when run with certain versions. The cross-version tests in the release (but not PRB) ended up thinking that the oldest selected version were the same as the `!current_version` and then hit a testing issue.

If we wanted to get a release out sooner, we could manually update the release number. It will result in our final builds skipping a number, but that might not be the end of the world.

alecgrieser:
This modifies the release build workflow so that we increment the version of local code before we go to run the tests. This makes it so that we correctly distinguish between the latest external version (which should have a version less than `!current_version`) and the current code version. This is necessary to unblock the release build, which is failing currently due to certain tests failing when run in mixed mode against the lateset published version due to expected differences.

Overall, I think this is a better set-up for our tests, as the version that will actually run the tests is now closer to version that we will actually end up publishing. The only difference should be the release number, which _shouldn't_ matter, but it's good to be sure.

This fixes #3449.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,7 @@ jobs:
   #   a) all the tests
   #   b) the mixed-mode tests
   test:
+    needs: [get-update-type]
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -50,6 +51,12 @@ jobs:
         uses: ./actions/setup-base-env
       - name: Setup FDB
         uses: ./actions/setup-fdb
+      # Increment the version here so that the version used in tests matches the eventually published version.
+      # This is actually important in order to correctly choose appropriate external server versions during
+      # mixed mode tests (see: https://github.com/FoundationDB/fdb-record-layer/issues/3449)
+      - name: Increment version
+        shell: bash
+        run: python build/versionutils.py gradle.properties --increment -u ${{ needs.get-update-type.outputs.update-type }}
       - name: Run Gradle Test
         uses: ./actions/gradle-test
         with:


.github/workflows/release.yml
```diff
@@ -50,6 +51,12 @@ jobs:
         uses: ./actions/setup-base-env
       - name: Setup FDB
         uses: ./actions/setup-fdb
+      # Increment the version here so that the version used in tests matches the eventually published version.
+      # This is actually important in order to correctly choose appropriate external server versions during
+      # mixed mode tests (see: https://github.com/FoundationDB/fdb-record-layer/issues/3449)
+      - name: Increment version
+        shell: bash
+        run: python build/versionutils.py gradle.properties --increment -u ${{ needs.get-update-type.outputs.update-type }}
```
alecgrieser, line 59: FWIW, I didn't actually try this, mainly because getting the tests going is a bit of a pain. This is, however, just copied from the code below, so I'm fairly confident in it. Note that it re-uses the `get-update-type` task that the other tasks use

ScottDugas: You didn't remove any other increment code, will this cause a double increment, or is it ok, because this is on a different machine from the publish, and we don't commit anything?

alecgrieser: Right. All three (now--previously two) increments happen in different jobs, and only the last one commits

PR #3451, opened 2025-06-27
add placeholders in all appropriate select expressions

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingRuleSet.java
```diff
@@ -37,8 +37,9 @@
 @SuppressWarnings("java:S1452")
 public class RewritingRuleSet extends CascadesRuleSet {
     private static final Set<ExplorationCascadesRule<? extends RelationalExpression>> EXPLORATION_RULES = ImmutableSet.of(
-    // new PredicatePushDownRule(),
-    // new RemoveRangeOneRule()
+    //            new PredicatePushDownRule(),
```
alecgrieser: Could we revert this change just to avoid causing trouble with merging #3401 (not that that conflict is a particularly difficult one to address)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
```diff
@@ -328,6 +318,65 @@ public GraphExpansion visitExpression(@Nonnull final NestingKeyExpression nestin
...
+                    final var pulledUpValue = pulledUpPlaceholderValuesMap.get(value);
+                    final var parameterAlias =
+                            Objects.requireNonNull(childExpansionPlaceholderValuesMap.get(value));
+                    return Placeholder.newInstanceWithoutRanges(pulledUpValue, parameterAlias);
+                })
+                .collect(ImmutableList.toImmutableList());
+    }
+
+    @Nonnull
+    private static ImmutableList<Column<? extends Value>> pullUpResultColumns(@Nonnull final GraphExpansion childExpansion,
```
alecgrieser, line 357: Just to make sure I haven't missed something, this is identical to the original code for pull up the result columns, right? The only substantive difference made in the whole PR is adding placeholder pull-up, right?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
```diff
@@ -294,31 +297,18 @@ public GraphExpansion visitExpression(@Nonnull final NestingKeyExpression nestin
...
+                    final var pulledUpExpansionColumns =
+                            pullUpResultColumns(childExpansion, childResultValue, childQuantifier);
+
+                    final var pulledUpPlaceholders =
+                            pullUpPlaceholders(childExpansion, childResultValue, childQuantifier);
+
+                    return GraphExpansion.builder()
                             .addQuantifier(childQuantifier)
+                            .addAllPredicates(pulledUpPlaceholders)
+                            .addAllPlaceholders(pulledUpPlaceholders)
```
alecgrieser, line 311: Could you speak to what the "problem" was here? Like, is the issue that without pulling up the placeholders on nested fields, matching logic down the line wasn't able to associate sargable predicates with the index, or something like that?

It does sort of concern me that there are no test cases that are added here that depend on the new behavior. There are some tests of nested repeated (aggregate) queries in `FDBNestedRepeatedQueryTest`. Those could be modified to include some more cases, or we could add some more YAML tests with some nested repeated queries

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
normen662: Yes, the same. I just wanted the logic to be extracted into their own little functions.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/KeyExpressionExpansionVisitor.java
normen662: This problem only occurs in aggregate indexes over a fanned out group or aggregation column. Thus, things like `FDBNestedRepeatedQueryTest` are not able to pick up a difference in behavior.

 It is however, as you say. Because of differences in the structure of the query and the match candidate, things wouldn't match in _with rewrites_. Changing things in the match candidates made them match but sometimes fails when run _without rewrites_. We are now relying on homomorphic matching (more boxes in the candidate) and duplicates placeholder to be
1. not reliant upon whether we run with or without rewrites
2. have a match candidate that is as canonical as possible

In the future we can be ease up on 1.) but I still think it is useful to have the candidates expose a high degree of flexibility as to what they can match. Also, we should think about running rewrites on candidates in order to reduce the dependency on the candidates to fulfill 2. out of the box.

alecgrieser: Okay, yeah, the actual difference here makes sense. It may be useful to note the reason in a code comment

> This problem only occurs in aggregate indexes over a fanned out group or aggregation column. Thus, things like `FDBNestedRepeatedQueryTest` are not able to pick up a difference in behavior.

That test class is designed, amongst other things, to test aggregate indexes over fanned out groups or values. I think none of the problematic cases that exhibit those differences are currently under test, though, which is why none of them failed. For example, there may not be queries that put predicates on the fanned-out columns of the select-where, and so they don't require the duplicated placeholders.

alecgrieser (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingRuleSet.java
normen662: done.

PR #3452, opened 2025-06-28
Allow yaml-test load template from json to add dependencies

pengpeng-lu:
now users can put dependencies in src/test/proto, and `load schema template` command will automatically load these dependencies.


arnaud-lacurie (changes requested):
Why does the `LOAD` comment need to specify the dependency? It seems that all the information should be already present in the json, via the type name and the dependencies that you declared at the top.

pengpeng-lu:
> Why does the `LOAD` comment need to specify the dependency? It seems that all the information should be already present in the json, via the type name and the dependencies that you declared at the top.

In the json file dependencies, it only has `test_student.proto` but not the file path to the proto file, so theoretically we would at least need to modify the file name to a full file path. I think it is a bit cleaner to add the compiled class name to the `load schema template` command than adding a somewhat random file path (it'll be something like `src/test/proto/` in practice but still theoretically can be random).

`record_metadata.proto` and `record_metadata_options.proto` work because we added them in `RecordMetaDataBuilder.initGeneratedDependencies`, not because they are declared in the json file.

PR #3455, opened 2025-07-01
Pick correct locally generated build for `!current_version` tests

### LINKED ISSUE #3454: `ExternalServerTest` fails due to inability to find current version jar
The most recent release build failed when `ExternalServerTest::startMultiple` failed to find more than one version: https://github.com/FoundationDB/fdb-record-layer/actions/runs/15996475482/attempts/1#summary-45120979352

It also appears that our release builds run the yaml tests with fewer configurations (in that it excludes `!current_version`), which I think is related.

alecgrieser:
The previous logic for choosing the locally generated build for `!current_version` required that there be the word `SNAPSHOT` in the jar name. This had the effect that during release builds, we wouldn't run tests with the current version as an external version. This resulted in test failures after `ExternalServerTest` was added.

This fixes #3454.

### DIFF
--- a/yaml-tests/yaml-tests.gradle
+++ b/yaml-tests/yaml-tests.gradle
@@ -179,7 +179,7 @@ task serverJars(type: Copy) {
     into project.layout.buildDirectory.dir('externalServer')
     from resolveOtherServer(["4.2.3.0", "4.2.4.0"].toSet())
     from (rootProject.project("fdb-relational-server").layout.files('.dist')) {
-        include("*-SNAPSHOT-all.jar")
+        include("fdb-relational-server-${rootProject.version}-all.jar")
     }
 }
 


alecgrieser:
Regarding testing: the PRB here won't show that this fixes the release build, but I did run the tests locally with the `-PreleaseBuild=true` flag, and that (1) reproduced the error seen during the release build before I made this change and (2) showed that this change fixes it.

yaml-tests/yaml-tests.gradle
```diff
@@ -179,7 +179,7 @@ task serverJars(type: Copy) {
     into project.layout.buildDirectory.dir('externalServer')
     from resolveOtherServer(["4.2.3.0", "4.2.4.0"].toSet())
     from (rootProject.project("fdb-relational-server").layout.files('.dist')) {
-        include("*-SNAPSHOT-all.jar")
+        include("fdb-relational-server-${rootProject.version}-all.jar")
```
ohadzeliger, line 182: Will this impact the PRBs? The `-SNAPSHOT` versions will still be generated by PRBs, right? Should we still be including them in that case?

alecgrieser: Right. During PRBs, the value of `${rootProject.version}` will have `-SNAPSHOT` in it, so we'll pick that value up

ohadzeliger (approved):

Issue #3456, opened 2025-07-01
Switch away from Sonatype for publishing because it is sunset


alecgrieser:
We've had two release builds fail now with 403 errors from Sonatype. See: https://github.com/FoundationDB/fdb-record-layer/actions/runs/16003773419

Looking into it, it looks like it is now sunset and has been shut down as of yesterday, June 30, 2025 (womp womp): https://central.sonatype.org/pages/ossrh-eol/

There is some guidance on how to overcome this on that page that we'll need to implement.

PR #3457, opened 2025-07-01
Update the Maven Central publishing URL to avoid sunset service

alecgrieser:
The service we were using to publish to Maven Central was sunset. There is a legacy API that allows us to continue to use our old publising gradle platform, and so I've updated the configuration so that it follows the advice available on their website: https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central

This is advertised to be pretty straightforward. The two main steps are (1) update the publishing URL to one that uses a bridge API and (2) modify the credentials to ones generated by the new portal. I've done that (and already updated the secrets on the repo with the new credentials). So we should be able to get a new build published through the new gateway once this is in.

I've also gone ahead and removed the release notes for the builds that never actually were published. That way, when we get a new build out, we'll get the correct set of release notes automatically without needing to tweaks those by hand later.

This fixes https://github.com/FoundationDB/fdb-record-layer/issues/3456.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -310,8 +310,8 @@ if (Boolean.parseBoolean(centralPublish)) {
     nexusPublishing {
         repositories {
             sonatype {
-                username = System.getenv("MAVEN_USER")
-                password = System.getenv("MAVEN_PASSWORD")
+                // Update the URL now that the OSSRH service has been sunset: https://central.sonatype.org/news/20250326_ossrh_sunset/
+                nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
             }
         }
     }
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -310,8 +310,8 @@ if (Boolean.parseBoolean(centralPublish)) {
     nexusPublishing {
         repositories {
             sonatype {
-                username = System.getenv("MAVEN_USER")
-                password = System.getenv("MAVEN_PASSWORD")
+                // Update the URL now that the OSSRH service has been sunset: https://central.sonatype.org/news/20250326_ossrh_sunset/
+                nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
```
ScottDugas, line 314: The documentation also has a
```
snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
```

Do we not need that because we aren't publishing `-SNAPSHOT`?

alecgrieser: Yes, thats right. That should only be needed for snapshot builds

PR #3460, opened 2025-07-02
Escape the ' character in the loggable method to help with log parser frameworks.

### LINKED ISSUE #3461: The ' (single quote) character in a log message can confuse log parsing frameworks.
The ' character can confuse log parsing frameworks and should be handled specifically. The `ByteArrayUtil2.loggable` method already handles a few characters and can be extended the special handling to '.

ohadzeliger:
The ' (single quote) character can confuse log parsers. This PR escapes this character when used within a log message.
This change alters the serialized representation of the byte array as a string. Any usage of this representation would be therefore modified. Searching the codebase showed usage within Exception messages and logging, which is likely OK, but care should be taken to verify that no ill impact is exposed.
Resolves #3461

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/tuple/ByteArrayUtil2Test.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/tuple/ByteArrayUtil2Test.java
@@ -51,7 +51,7 @@ public class ByteArrayUtil2Test {
         for (int i = (byte)'"' + 1; i < (byte)'\''; i++) {
             assertEquals(Character.toString((char) i), ByteArrayUtil2.loggable(new byte[]{(byte)i}));
         }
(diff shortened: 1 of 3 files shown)


fdb-extensions/src/test/java/com/apple/foundationdb/tuple/ByteArrayUtil2Test.java
```diff
@@ -92,7 +96,7 @@ void testRandomBytes(long seed) {
         assertFalse(loggable.contains("="), "loggable string should not contain equals sign");
         assertFalse(loggable.contains("\""), "loggable string should not contain quote");
```
MMcM, line 97: Might as well include single quote in these assertions, too.

ohadzeliger: Right. Done.

MMcM (approved):

PR #3463, opened 2025-07-03
WIP - add `PlannerConfiguration` to `QueryCacheKey`.

hatyo:


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/PlannerConfiguration.java
```diff
@@ -49,47 +49,70 @@
...
 
+    @Nonnull
+    public RecordQueryPlannerConfiguration getRecordQueryPlannerConfiguration() {
+        return recordQueryPlannerConfiguration;
+    }
+
     @Override
-    public boolean equals(Object o) {
-        if (this == o) {
+    public boolean equals(Object other) {
```
hatyo, line 92: Essentially this is the important part of this PR.

`PlannerConfiguration` is added to `QueryCacheKey`.

The implementation of `equals` and _memoized_ `hashCode` picks the bits that are only relevant for the `QueryCacheKey`, they are:

1. The `IndexFetchMethod`.
2. The set of disabled transformation rules.

cc @normen662 @alecgrieser

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -133,7 +133,9 @@ public enum Name {
          * By default, no rules are disabled, and the user should be able to leave this
          * unset. This is intended as an escape hatch in case the introduction of a new
          * planner rule causes trouble for an existing query.
-         *
+         * <p>
+         * This option is experimental, and is meant to be used for debugging purposes only.
+         * <p>
```
alecgrieser:
```suggestion
         * </p>
```

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/PlannerConfiguration.java
alecgrieser: Hm. I'm not crazy about the way that `equals` and `hashCode` are overridden here. It seems like it could lead to a problem in that implementors wouldn't expect it to exclude parts of its members from `equals` calculations. (We should certainly add documentation in the code if we did make that choice.)

Is there a reason it can't use the _whole_ planner configuration (assuming we add `equals` and `hashCode` implementations to that class)? There are a bunch of options there, though most of those don't get set through the options. That's seems the safest, in that if we did add another relevant option, this logic handling the plan cache keys would automatically work. The next safest would be to extract the relevant fields from the `RecordQueryPlannerConfiguration`, and then we're more explicit about what fields are relevant to the plan cache

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/OptionsUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import com.apple.foundationdb.relational.util.Assert;
+import com.google.common.collect.ImmutableSet;
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Set;
+
+public final class OptionsUtils {
```
alecgrieser, line 37: Does this need to be `public`? This probably should be marked as `API(API.Status.INTERNAL)` if it can't be made package-private

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/PlannerConfiguration.java
```diff
@@ -49,47 +49,70 @@
...
     }
 
     @Nonnull
     public Optional<Set<String>> getReadableIndexes() {
         return readableIndexes;
     }
 
+    @Nonnull
+    public RecordQueryPlannerConfiguration getRecordQueryPlannerConfiguration() {
+        return recordQueryPlannerConfiguration;
```
alecgrieser, line 88: Do we have any tests that validate that if the same query is planned multiple times with different planner configurations, then we use/don't reuse plans correctly?

hatyo: sure, I added a new test that verifies exactly that. I might few more tests (PR is still WIP).

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/OptionsUtils.java
hatyo: It has to be public. But I added `API(API.Status.INTERNAL)`. Thanks for the suggestion.

alecgrieser (approved):

PR #3471, opened 2025-07-07
Yaml transactions

ScottDugas:


fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/jdbc/v1/TransactionRequestHandler.java
```diff
@@ -82,9 +82,13 @@ public void onNext(final TransactionalRequest transactionRequest) {
                 final FRL.Response response = frl.transactionalExecute(transactionalToken, request.getSql(),
                         request.getParameters().getParameterList(), fromProto(request.getOptions()));
 
-                StatementResponse.Builder statementResponseBuilder = StatementResponse.newBuilder()
-                        .setResultSet(response.getResultSet())
-                        .setRowCount(response.getRowCount());
+                StatementResponse.Builder statementResponseBuilder = StatementResponse.newBuilder();
```
ScottDugas: creating temporary tables does not return a resultSet, so this was failing because the `Optional` underneath was not present.

fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
```diff
@@ -345,7 +345,8 @@ public void setAutoCommit(boolean autoCommit) throws SQLException {
             this.autoCommit = false;
         } else {
             // commit any remaining work
-            commit();
+            // TODO do we need better protection if there is uncommitted work
+            //      does the autoCommit-ness need to be passed down
```
ScottDugas: This was doing the right thing
>   <B>NOTE:</B>  If this method is called during a transaction and the
> auto-commit mode is changed, the transaction is committed.  If
> <code>setAutoCommit</code> is called and the auto-commit mode is
> not changed, the call is a no-op.

But, it would end up calling `commit` on the server when there was no active transaction, which failed.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/configs/ExternalSingleServerConfig.java
```diff
@@ -0,0 +1,64 @@
...
+import com.apple.foundationdb.relational.yamltests.YamlExecutionContext;
+import com.apple.foundationdb.relational.yamltests.connectionfactory.ExternalServerYamlConnectionFactory;
+import com.apple.foundationdb.relational.yamltests.server.ExternalServer;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Run against multiple external servers, alternating commands that go against each.
+ */
+public class ExternalSingleServerConfig implements YamlTestConfig {
```
ScottDugas: This isn't necessary, I created it while debugging, I should pull out into a separate, unrelated PR.

yaml-tests/src/test/resources/transactions-tests.yamsql
```diff
@@ -37,21 +37,29 @@
...
 ---
 options:
-    supported_version: 4.3.5.0
+    supported_version: !current_version
 ---
 schema_template:
-    create table t1(id bigint, col1 bigint, primary key(id))
+    create table table_t1(id bigint, col1 bigint, primary key(id))
+---
+transaction_setups:
```
ScottDugas, line 33: I looked into using yaml's repeated nodes feature, but since each test_block is a separate "document" you can't really do that, so instead I'm adding our own reference section.

yaml-tests/src/test/resources/transactions-tests.yamsql
```diff
@@ -37,21 +37,29 @@
...
+  t1_less_than_50: create temporary function t1() on commit drop function AS SELECT * FROM table_t1 where id < 50
```
ScottDugas: In theory, it wouldn't be hard to change this to allow an array of strings if there's a bunch of setup steps

fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
ScottDugas: Created Issue #3473

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/jdbc/v1/TransactionRequestHandler.java
ScottDugas: Created issue: #3475

scripts/YAML-SQL.xml
```diff
@@ -53,8 +53,8 @@
     multi_repetition_parallelized" />
     <keywords4 keywords="connect:;query:;load schema template:;set schema state:;result:;unorderedResult:;explain:;
     explainContains:;count:;error:;planHash:;setup:;schema_template:;supported_version:;test_block:;options:;tests:;
-    maxRows:;initialVersionAtLeast:;initialVersionLessThan:;;mode:;repetition:;seed:;
-    check_cache:;connection_lifecycle:;steps:;preset:;statement_type:;!r;!in;!a;" />
+    maxRows:;initialVersionAtLeast:;initialVersionLessThan:;;mode:;repetition:;seed:;setup:;transaction_setups:;
+    setupReference:;check_cache:;connection_lifecycle:;steps:;preset:;statement_type:;!r;!in;!a;" />
```
ScottDugas, line 57: We seem pretty incosistent in general about whether things should be `snake_case` or `camelCase`

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
```diff
@@ -207,6 +207,10 @@ private void executeInternal(@Nonnull final YamlConnection connection, boolean c
             } else if (QueryConfig.QUERY_CONFIG_NO_OP.equals(queryConfig.getConfigName())) {
                 // Do nothing for noop execution.
                 continue;
+            } else if (QueryConfig.QUERY_CONFIG_SETUP.equals(queryConfig.getConfigName())) {
+                Assert.that(!queryIsRunning, "Transaction setup should not be intermingled with query results");
+                executor.addSetup(Matchers.string(Matchers.notNull(queryConfig.getVal(), "Setup Config Val"),
+                        "Transaction setup"));
```
g31pranjal: There seems to be some merit in enforcing "setup" configs to always appear consecutively and directly above "query"? I suppose that will be more visibly clear given what "setup" is supposed to do.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
```diff
@@ -188,6 +197,22 @@ private Continuation executeQuery(@Nonnull YamlConnection connection, @Nonnull Q
...
     }
 
+    @Nullable
+    private Continuation executeWithSetup(final @Nonnull YamlConnection connection,
+                                          SQLFunction<YamlConnection, Continuation> execute) throws SQLException, RelationalException {
+        if (!setup.isEmpty()) {
+            return connection.executeTransactionally(singleConnection -> {
+                for (var setupStatement : setup) {
+                    final RelationalStatement statement = singleConnection.createStatement();
+                    statement.execute(setupStatement);
```
g31pranjal, line 207: Hmm, while I see this to be working well for creating temporary functions, I am a unclear on the scope of SQL statements that can be used in "setup". For instance, `UPDATE .... RETURNING ...` won't work here as it requires slurping the result set. Now that I think of it, DQL works fine and so does DML except the case I mentioned above.

g31pranjal: I think we should either, 
1) restrict the scope of statements here (maybe to only CREATE TEMPORARY FUNCTION ....) by some sort of patern matching, or we can error if the `execute` returns a `ResultSet` (barring DQL completely). 
2) or, we could softly specify in documentation about its specific usages and cautions.   
 
 Open to your thoughts on this!

yaml-tests/src/test/resources/transactions-tests.yamsql
```diff
@@ -17,41 +17,45 @@
...
+      - query: select * from t1 where id > 15;
+      - setup: create temporary function t1() on commit drop function AS
+               SELECT * FROM table_t1 where id < 30;
+      - result: []
+    -
+      # This query references the transaction setup from above. This behaves exactly the same as the inline version
+      # it just allows easier use of the same setup for many queries.
       - query: select * from t1 where id > 15;
+      - setupReference: t1_less_than_50
+      - result: [{id: 30, col1: 40}]
```
g31pranjal, line 60: It will go to add cases here to showcase/tests that utilizes multiple setup configs (looks like that is supported!) and that the order of execution is preserved!

g31pranjal: Another case could be to have a mix of `setup` and `setupReference` intermingled

yaml-tests/src/test/resources/transactions-tests.yamsql
```diff
@@ -17,41 +17,45 @@
...
 schema_template:
-    create table t1(id bigint, col1 bigint, primary key(id))
+    create table table_t1(id bigint, col1 bigint, primary key(id))
+---
+# This document type creates a map of transaction setups that can be referenced in individual queries
+# This helps to reduce duplication.
+# The map is basically from a name (e.g. t1_less_than_50) to a statement to run at the start of every query.
+# They're referenced in the query using `setupReference`. Right now it only supports a single statement as a string, but
+# it could be expanded to support an array in the future.
+transaction_setups:
```
g31pranjal, line 33: it seems like multiple `transaction_setups` are allowed, so they can be spread all over the file. However, since all of them are processed in `parsing`, the lifecycle of all the kv pairs is entire runtime. I think it will be good to disallow that or have an example citing this behavior.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/SQLFunction.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * SQLSupplier.java
```
g31pranjal: name

g31pranjal (changes requested):
I remember that you introduced some unit tests to support the `supported_version` functionality. Will it be good to have some like those that could check for disallowed yaml layouts here, like having setups before query and key collison.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
ScottDugas: We could do that, but the way we figure out what Command it is, is by the first entry, so we would have to change that logic in order to support putting the `setup` first. I feel like that's a bit more refactoring than is worth bringing in here.
It's also possible that we're pushing up against the limits of the original design, and changing it to be a map might make more sense, but that would be a pretty big change.
One other way to think about it is that the query itself is the test name, and thus should go first.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryExecutor.java
ScottDugas: Yeah, that's a good callout.
We currently commit at the end of the query (maybe that's a mistake), and so having other DDL would cause problems. I forbid it for now, and put a comment there that this might not stay if we can figure out how we want to use it in other situations.

yaml-tests/src/test/resources/transactions-tests.yamsql
ScottDugas: It appears to work as expected; I added some tests that cover this, and show that only `test_block`s after the `transaction_setup` can refer to it. It doesn't make a lot of sense if you have just tests, but if you have some tests, some setup, and then some transaction_setups it may make more sense.

yaml-tests/src/test/resources/transactions-tests.yamsql
ScottDugas: Yeah, that's a good callout. I added a bunch of tests.

yaml-tests/src/test/resources/transaction-setup/shouldFail/setup-reference-after-results.yamsql
```diff
@@ -0,0 +1,41 @@
...
+  steps:
+    - query: INSERT INTO table_T1 VALUES (10, 20), (30, 40), (50, 60), (70, 80)
+---
+test_block:
+  name: transactions-tests
+  tests:
+    -
+      - setupReference: t1_less_than_50
+      - result: [{id: 30, col1: 40}]
+      - query: select * from t1 where id > 10;
```
g31pranjal: The order of commands here is not in line with the filename!

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/command/QueryCommand.java
g31pranjal: Aah, I see! Completely missed the fact. I agree that's a good to have but is too much to do.

g31pranjal (changes requested):
LGTM! 
There is just a minor fix required in one of the tests.

Issue #3473, opened 2025-07-08
Enabling autoCommit against server without active transaction throws an error


ScottDugas:
If you enable autoCommit against a server without an active transaction, it will fail, e.g.:

```
connection.setAutoCommit(false);
connection.setAutoCommit(true);
```
throws: `Transaction was not initialized`

```
connection.setAutoCommit(false);
// do something in the transaction
connection.commit();
connection.setAutoCommit(true);
```
throws: `No transaction to commit`

```
connection.setAutoCommit(false);
// do something in the transaction
connection.rollback();
connection.setAutoCommit(true);
```
throws: `No transaction to commit`

PR #3477, opened 2025-07-08
Allow enabling autoCommit without an active transaction

ScottDugas:
This changes the client so that it sends a request to enable autoCommit when it is enabled on the client, rather than calling `commit` directly. By doing so, the server can determine whether it wants to commit or not based on whether it has an active transaction.

Fixes: #3473

### DIFF
--- a/fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
+++ b/fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
@@ -31,6 +31,7 @@ import com.apple.foundationdb.relational.jdbc.grpc.GrpcConstants;
 import com.apple.foundationdb.relational.jdbc.grpc.GrpcSQLExceptionUtil;
 import com.apple.foundationdb.relational.jdbc.grpc.v1.CommitRequest;
 import com.apple.foundationdb.relational.jdbc.grpc.v1.DatabaseMetaDataRequest;
+import com.apple.foundationdb.relational.jdbc.grpc.v1.EnableAutoCommitRequest;
 import com.apple.foundationdb.relational.jdbc.grpc.v1.InsertRequest;
 import com.apple.foundationdb.relational.jdbc.grpc.v1.InsertResponse;
 import com.apple.foundationdb.relational.jdbc.grpc.v1.JDBCServiceGrpc;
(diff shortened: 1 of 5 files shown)


fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
```diff
@@ -345,7 +346,23 @@ public void setAutoCommit(boolean autoCommit) throws SQLException {
             this.autoCommit = false;
         } else {
             // commit any remaining work
-            commit();
+            try {
+                TransactionalRequest.Builder transactionRequest = TransactionalRequest.newBuilder()
+                        .setEnableAutoCommitRequest(EnableAutoCommitRequest.newBuilder().build());
+                // wait here for response
+                final TransactionalResponse response = serverConnection.sendRequest(transactionRequest.build());
```
ohadzeliger, line 353: Just to clarify: This request will always be sent prior to closing the stateful connection, and so (for the time being) can be considered an equivalent of "connection about to be orderly closed", right? 
It is feasible that in the future we will use this same request to signify the change to the state of the connection (since I think we will change the protocol to be only using the "stateful" path to better sync the state of the connection between client and server).

fdb-relational-jdbc/src/test/java/com/apple/foundationdb/relational/jdbc/JDBCAutoCommitTest.java
```diff
@@ -80,6 +80,65 @@ void autoCommitOn() throws SQLException, IOException {
...
+        try (RelationalConnection connection = DriverManager.getConnection(getTestDbUri()).unwrap(RelationalConnection.class)) {
+            connection.setAutoCommit(false);
+
+            try (RelationalStatement statement = connection.createStatement()) {
+                insertOneRow(statement);
+            }
+            connection.commit();
+            connection.setAutoCommit(true);
+            try (RelationalStatement statement = connection.createStatement()) {
+                RelationalResultSet resultSet = statement.executeQuery("SELECT * FROM test_table");
```
ohadzeliger, line 120: Maybe this would be more exhaustive with another insert after the setAutoCommit(true)?

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/FRL.java
```diff
@@ -371,6 +371,16 @@ public void transactionalRollback(TransactionalToken token) throws SQLException
         token.getConnection().rollback();
     }
 
+    public void enableAutoCommit(TransactionalToken token) throws SQLException {
+        // we don't actually call setAutoCommit(false) until an operation happens, so if there is no token
```
ohadzeliger: I think that - aligning this towards the future, where the client connection state will be in sync with the server-side connection state (TransactionalRequestHandler) - the logic of looking at the token should reside in the TransactionalRequestHandler. 
FRL would be a simpler, stateless pass through to the EmbeddedConnection.

fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/JDBCRelationalConnection.java
ScottDugas: Yeah, I think that is one way of thinking about it. It will make even more sense when we combine the two.

ohadzeliger:
test this please

PR #3482, opened 2025-07-10
Preserve `IntermingleTables` flag in `RecordLayerSchemaTemplate.toBuilder`

### LINKED ISSUE #3483: `IntermingleTables` flag in not preserved
When converting a `RecordLayerSchemaTemplate` to `Builder` in:

https://github.com/FoundationDB/fdb-record-layer/blob/a2b404726578b091f0c19e54271b4febd2726a0d/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java#L679-L685

The `IntermingleTables` flag is not carried over, and therefore, it can permit adding new tables with an incorrect primary key.

hatyo:
This preserves `IntermingleTables` when turning a `RecordLayerSchemaTemplate` to its `Builder`.

It resolves #3483.

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
@@ -39,6 +39,7 @@ import com.apple.foundationdb.relational.recordlayer.metadata.serde.RecordMetada
 import com.apple.foundationdb.relational.recordlayer.metadata.serde.RecordMetadataSerializer;
 import com.apple.foundationdb.relational.util.Assert;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
@@ -95,7 +96,7 @@ public final class RecordLayerSchemaTemplate implements SchemaTemplate {
     @Nonnull
     private final Supplier<String> transactionBoundMetadataSupplier;
 
-    private final boolean tablesAreIntermingled;
+    private final boolean intermingleTables;
 
     private RecordLayerSchemaTemplate(@Nonnull final String name,
                                       @Nonnull final Set<RecordLayerTable> tables,
@@ -103,14 +104,14 @@ public final class RecordLayerSchemaTemplate implements SchemaTemplate {
                                       int version,
                                       boolean enableLongRows,
                                       boolean storeRowVersions,
-                                      boolean tablesAreIntermingled) {
+                                      boolean intermingleTables) {
         this.name = name;
         this.tables = ImmutableSet.copyOf(tables);
         this.invokedRoutines = ImmutableSet.copyOf(invokedRoutines);
         this.version = version;
         this.enableLongRows = enableLongRows;
         this.storeRowVersions = storeRowVersions;
-        this.tablesAreIntermingled = tablesAreIntermingled;
+        this.intermingleTables = intermingleTables;
         this.metaDataSupplier = Suppliers.memoize(this::buildRecordMetadata);
         this.tableIndexMappingSupplier = Suppliers.memoize(this::computeTableIndexMapping);
         this.indexesSupplier = Suppliers.memoize(this::computeIndexes);
@@ -124,7 +125,7 @@ public final class RecordLayerSchemaTemplate implements SchemaTemplate {
                                       int version,
                                       boolean enableLongRows,
                                       boolean storeRowVersions,
-                                      boolean tablesAreIntermingled,
+                                      boolean intermingleTables,
                                       @Nonnull final RecordMetaData cachedMetadata) {
         this.name = name;
         this.version = version;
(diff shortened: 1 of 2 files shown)


fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/metadata/SchemaTemplateSerDeTests.java
```diff
@@ -433,6 +433,59 @@ public void onlyQueriedSqlFunctionsAreCompiled() throws RelationalException {
...
+        Assertions.assertTrue(newBuilder.isIntermingleTables());
+
+        sampleRecordSchemaTemplate = RecordLayerSchemaTemplate.newBuilder()
+                .setName("TestSchemaTemplate")
+                .setVersion(42)
+                .addAuxiliaryType(DataType.StructType.from(
+                        "Subtype",
+                        List.of(DataType.StructType.Field.from("field1", DataType.Primitives.INTEGER.type(), 0)),
+                        true))
+                .setIntermingleTables(false)
```
alecgrieser: These two cases are the same except for how they set intermingle tables, right? This could be a parameterized test where the value of `setIntermingleTables` is a parameter to the test, which would help to avoid some code repetition here

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/metadata/SchemaTemplateSerDeTests.java
```diff
@@ -433,6 +433,59 @@ public void onlyQueriedSqlFunctionsAreCompiled() throws RelationalException {
         Assertions.assertTrue(peekingDeserializer.hasOneCompilationRequestFor("SqlFunction3"));
     }
 
+    @Test
+    public void schemaTemplateToBuilderPreservesIntermingledTablesFlag() {
```
alecgrieser: Given that I believe this was raised in response to #3480, would it be good to have a more explicit test of adding a temporary function to a schema template with intermingled tables?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
```diff
@@ -95,18 +95,22 @@ public final class RecordLayerSchemaTemplate implements SchemaTemplate {
     @Nonnull
     private final Supplier<String> transactionBoundMetadataSupplier;
 
+    private final boolean tablesAreIntermingled;
```
alecgrieser: Should we expose a method to probe the value of this flag from a `RecordLayerSchemaTemplate`? (Like the builder  method, `isIntermingleTables`?)

I might also prefer to keep the names consistent, though it's not like `tablesAreIntermingled` and `intermingleTables` are that hard to decipher as a pair. But `intermingleTables` is closer to the flag in the DDL

alecgrieser (approved):
LGTM. I have some testing and naming thoughts, but overall, this bug fix looks like it accomplishes what it sets out to do

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
hatyo: I don't think we require this flag anywhere in the codebase, we just use it to tweak the way we construct the tables' primary key.

I changed the name to `intermingleTables` to make it consistent with the `Builder`'s.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/metadata/SchemaTemplateSerDeTests.java
hatyo: done.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
alecgrieser: That sounds true (that nothing else should require introspecting this flag). My main thought is that it makes the tests for this _slightly_ more straightforward, in that you don't need to inspect the way the builder sets it, but it's mostly tomato/tomato.

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/metadata/SchemaTemplateSerDeTests.java
hatyo: done.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerSchemaTemplate.java
hatyo: yeah, I see your point, let's do that.

PR #3486, opened 2025-07-16
resolves: #3487 restrict select expressions in rewrite matchers to be exploratory

### LINKED ISSUE #3487: rewrite rules create variations based on final root expressions
Since none of the rewrite rules are optimizing a final expression, they all should only use exploratory expressions.

normen662:
This fixes #3487

### DIFF
--- a/yaml-tests/src/test/java/YamlIntegrationTests.java
+++ b/yaml-tests/src/test/java/YamlIntegrationTests.java
@@ -33,7 +33,6 @@ import org.junit.jupiter.api.TestTemplate;
  * up the {@link com.apple.foundationdb.record.query.plan.cascades.debug.PlannerRepl} debugger implementation.
  */
 @YamlTest
-@MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
 public class YamlIntegrationTests {
     @TestTemplate
     public void showcasingTests(YamlTest.Runner runner) throws Exception {


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
```diff
@@ -626,21 +626,31 @@ public Optional<MatchInfo> adjustMatch(@Nonnull final PartialMatch partialMatch)
     @Nonnull
     @Override
     public PlannerGraph rewriteInternalPlannerGraph(@Nonnull final List<? extends PlannerGraph> childGraphs) {
+        final var predicateString = "WHERE " + getPredicates().stream()
+                .map(Object::toString)
+                .collect(Collectors.joining(" AND "));
+
+        final var abbreviatedPredicateString = predicateString.length() > 30 ? String.format("%02x", predicateString.hashCode()) : predicateString;
```
hatyo, line 633: Would it make sense to display the first 30 characters, maybe with `...` at the end, instead of the hash code?

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -33,6 +33,7 @@
  * up the {@link com.apple.foundationdb.record.query.plan.cascades.debug.PlannerRepl} debugger implementation.
  */
 @YamlTest
+@MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
```
hatyo: Please remove.

hatyo (changes requested):

yaml-tests/src/test/java/YamlIntegrationTests.java
normen662: yoinks!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/SelectExpression.java
normen662: It's nice to stare at the boxes and know that they are probably the same // definitely not the same.

PR #3490, opened 2025-07-17
Add support for NULL type fields in ExecuteInsert

g31pranjal:
This PR adds support for inserting structs with NULL typed fields into the table via `ExecuteInsert` in the Direct Access API. Also adds a test for the same. 

This bug was introduced by https://github.com/FoundationDB/fdb-record-layer/pull/3484, which essentially shifts the Relational-layer from using Sql Type codes for internal operations to relying on Relational DataTypes for getting a richer information. 

Fixes: https://github.com/FoundationDB/fdb-record-layer/issues/3491

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordTypeTable.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordTypeTable.java
@@ -250,6 +250,8 @@ public class RecordTypeTable extends RecordTypeScannable<FDBStoredRecord<Message
                             }
                         }
                         break;
+                    case NULL:
+                        break;
                     default:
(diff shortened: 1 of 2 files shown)


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordTypeTable.java
```diff
@@ -250,6 +250,8 @@ public static Message toDynamicMessage(RelationalStruct struct, Descriptors.Desc
                             }
                         }
                         break;
+                    case NULL:
+                        break;
```
hatyo, line 254: so a `null`-typed field is reflected by non-existing field in the respective dynamic message.

g31pranjal: yep!

PR #3494, opened 2025-07-21
Directory thread safety

ScottDugas:


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneRecordCursor.java
```diff
@@ -422,9 +422,7 @@ private CompletableFuture<RecordCursorResult<IndexEntry>> switchToNextPartitionA
 
     @Override
     public void close() {
-        if (indexReader != null) {
-            IOUtils.closeWhileHandlingException(indexReader);
```
ScottDugas: `IOUtils.closeWhileHandlingException` checks for `null` itself.

ohadzeliger: It may be better to leave this in place in this case, for readability. There is no ambiguity over who is null here (the closable or the array of closeables) but one can easily get confused when the parameter to a varargs method is a literal `null`.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
```diff
@@ -84,25 +84,21 @@ public class FDBDirectoryManager implements AutoCloseable {
...
         this.mergeDirectoryCount = getMergeDirectoryCount(state);
-        if (FDBTieredMergePolicy.usesCreationStack()) {
-            this.exceptionAtCreation = new Exception();
-        } else {
-            this.exceptionAtCreation = null;
-        }
         final var fieldInfos = LuceneIndexExpressions.getDocumentFieldDerivations(state.index, state.store.getRecordMetaData());
         this.analyzerSelector = LuceneAnalyzerRegistryImpl.instance().getLuceneAnalyzerCombinationProvider(state.index, LuceneAnalyzerType.FULL_TEXT, fieldInfos);
         this.writerAnalyzer = analyzerSelector.provideIndexAnalyzer();
+        this.exceptionAtCreation = FDBTieredMergePolicy.usesCreationStack() ? new Exception() : null;
```
ohadzeliger: Why move this statement down? If nothing else, the generated stack trace will be a little different.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
```diff
@@ -72,35 +72,94 @@ public class FDBDirectoryWrapper implements AutoCloseable {
     private final Tuple key;
     @Nonnull
     private final LuceneAnalyzerWrapper analyzerWrapper;
-    @SuppressWarnings({"squid:S3077"}) // object is thread safe, so use of volatile to control instance creation is correct
-    private volatile IndexWriter writer;
-    @SuppressWarnings({"squid:S3077"}) // object is thread safe, so use of volatile to control instance creation is correct
-    private volatile DirectoryReader writerReader;
+    private volatile boolean useWriter = false;
```
ohadzeliger, line 82: I believe I was able to verify that the previous behavior was maintained through this change. It would be beneficial, though, maybe later, to add a little more documentation to the class - esp. around the writer/writerReader and `useWriter`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -779,9 +780,9 @@ void mergeLosesLockTest(int failurePercentage) throws IOException {
                 FDBDirectory fdbDirectory = new InvalidLockTestFDBDirectory(recordStore.indexSubspace(index).subspace(directoryKey), context, options, failurePercentage);
                 FDBDirectoryWrapper fdbDirectoryWrapper = new FDBDirectoryWrapper(state, fdbDirectory, directoryKey,
                         1, AgilityContext.agile(context, 1L, 1L),
-                        indexAnalyzerSelector.provideIndexAnalyzer());
+                        indexAnalyzerSelector.provideIndexAnalyzer(), null);
```
ohadzeliger:
```suggestion
                        indexAnalyzerSelector.provideIndexAnalyzer(), new Exception());
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -880,8 +881,9 @@ void sampledDelete(boolean isSynthetic, boolean isGrouped, long seed) throws IOE
     private static Stream<Arguments> concurrentParameters() {
         // only run the individual tests with synthetic during nightly, the mix runs both
         return Stream.concat(Stream.of(false),
-                TestConfigurationUtils.onlyNightly(Stream.of(true)))
-                .map(isSynthetic -> Arguments.of(isSynthetic));
+                TestConfigurationUtils.onlyNightly(
+                        IntStream.range(0, 3).boxed().flatMap(i -> Stream.of(true, false))))
```
ohadzeliger, line 885: This doesn't seem to match the `boolean` param of the test method?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryManager.java
ScottDugas: Side effect of history. I removed it, and added it back, but apparently not in the same place.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryWrapper.java
ScottDugas: Yeah, good callout, I added some documentation around these three things.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ScottDugas: It's doing a flatMap, so this produces:
1. `true`
2. `false`
1. `true`
2. `false`
1. `true`
2. `false`

(Note: It will also get the non-nightly `false`, so nightly builds will have 7 runs, and one more `false` than` true`)

ohadzeliger (approved):

PR #3495, opened 2025-07-21
Enable AggregateCursorContinuation serialize to the new implementation

pengpeng-lu:
AggregateCursorContinuation was introduced in this PR: https://github.com/FoundationDB/fdb-record-layer/pull/3397/files#, but it was not enabled, the continuation was still serialized in the old way, this PR enables it to serialize in the new way.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
@@ -222,7 +222,7 @@ public class RecordQueryStreamingAggregationPlan implements RecordQueryPlanWithC
                 groupingKeyAlias,
                 aggregateAlias,
                 completeResultValue,
-                SerializationMode.TO_OLD);
+                SerializationMode.TO_NEW);
     }
 
     @Nonnull
@@ -234,7 +234,7 @@ public class RecordQueryStreamingAggregationPlan implements RecordQueryPlanWithC
                 groupingKeyAlias,
                 aggregateAlias,
                 completeResultValue,
-                SerializationMode.TO_OLD);
+                SerializationMode.TO_NEW);
     }
 
     @Nonnull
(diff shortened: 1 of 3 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -222,7 +222,7 @@ public RecordQueryStreamingAggregationPlan translateCorrelations(@Nonnull final
                 groupingKeyAlias,
                 aggregateAlias,
                 completeResultValue,
-                SerializationMode.TO_OLD);
+                SerializationMode.TO_NEW);
```
alecgrieser: It's a bit unfortunate that we didn't change this prior to this, but I believe that this should actually be:

```suggestion
                serializationMode);
```

That is, calling `translateCorrelations` should preserve this setting rather than forcing it one way or the other. I don't think that this is actually symptomatic, as we should only be creating plans with non-default values of this parameter in the context of deserializing protos, and we don't call `translateCorrelations` (or `withChild`) on those, but it would be good for this to be consistent.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
```diff
@@ -234,7 +234,7 @@ public RecordQueryStreamingAggregationPlan withChild(@Nonnull final Reference ch
...
```
alecgrieser:
```suggestion
                serializationMode);
```

Similar to the other instance

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/RecordLayerIterator.java
```diff
@@ -90,6 +90,8 @@ private void fetchNextResult() {
             noNextReason = result.getNoNextReason();
             if (noNextReason == RecordCursor.NoNextReason.SOURCE_EXHAUSTED) {
                 this.continuation = ContinuationImpl.END;
+            } else {
+                this.continuation = ContinuationImpl.fromUnderlyingBytes(result.getContinuation().toBytes());
```
alecgrieser, line 94: I'm pretty sure that this new code is correct. It's interesting to me that we didn't seem to catch it before this. I suppose that's because most of our tests check for limit-reached cases, in which case the final continuation of the last returned result is the same as the continuation of the element with no-next-reason, so we were hiding this error. I assume the test you enabled caught this?

pengpeng-lu: yes the test I introduced returns a result (hasNext = False, nextValue = null, continuation = (continuation_bytes, partial_aggregation_result)), so this is the case that hasNext = False but noNextReason != SOURCE_EXHAUSTED.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryStreamingAggregationPlan.java
pengpeng-lu: yes you are absolutely right. Thanks for catching this!

alecgrieser (approved):

PR #3506, opened 2025-07-25
`NULL` handling fixes in plan generator

### LINKED ISSUE #3500: Plan generator constraints incorrectly handles `null` constants
When encountering literals, the plan generator generates equality constraints for duplicate literals, the equality constraints do not produce the expected semantics when both values are null. This is because the equality constraints (correctly) returns `null` when either operand is `null`:

https://github.com/FoundationDB/fdb-record-layer/blob/8cfcff6481936130233b014c2ce632c500f04a42/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java#L1620-L1624

and

https://github.com/FoundationDB/fdb-record-layer/blob/8cfcff6481936130233b014c2ce632c500f04a42/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/expressions/Comparisons.java#L250-L253

To understand why this fails, consider this query:

```
select * from t1 where col1 = ?foo and col2 = ?bar
```
where both `foo` and `bar` bind to `null`, the generated constraint will be a conjunction of predicates, including this one:

```
ValuePredicate(Cov7, Comparisons.Comparison(Equals, Cov11))
```
where `Cov7` is the placeholder for `foo` and `Cov11` is the placeholder of `bar` in the `EvaluationContext`. 

The above terms will _always_ evaluate to `null`, meaning the overall predicate will not be true. And therefore, we will fail to pick up the plan from the cache against the exact same query.

### LINKED ISSUE #3503: `null` Literals are silently dropped
When a query contains `null`, an `OrderedLiteral` is created with its literal object set to `null`. The original implementation intentionally filtered out any `OrderedLiteral` with null values when creating the corresponding `Literals` list, likely because `ImmutableMap` doesn't permit `null` keys or values:

[View the original implementation](https://github.com/FoundationDB/fdb-record-layer/commit/a78c038e6e8634aacfd957f66bf83a3f45676380#diff-fe0449f37785fb7417b9c5bb7a7f9bf343cf836407cd3f758522e3db18f96a1dR88-R90)

```java
public Literals(@Nonnull final List<OrderedLiteral> orderedLiterals) {
    this.orderedLiterals = ImmutableList.copyOf(orderedLiterals);
    this.asMapSupplier = Suppliers.memoize(() ->
            this.orderedLiterals
                    .stream()
                    .filter(orderedLiteral -> orderedLiteral.getLiteralObject() != null)
                    .collect(ImmutableMap.toImmutableMap(OrderedLiteral::getConstantId,
                            OrderedLiteral::getLiteralObject)));
}
```

While this logic may have been acceptable when first implemented, it now represents a significant flaw that subtly prevents certain rewrites in the newly introduced multi-phased planner (#3328). The `DereferenceConstantObjectValue` rule from #3403 only performs replacements when the `EvaluationContext` contains the corresponding reference. This approach is problematic because the `EvaluationContext` should contain _all_ literal values from query plan references, and this filtering behavior masks the underlying issue.

The failure to replace `ConstantObjectValue` references to `null` blocks subsequent predicate simplifications, which can only process expression trees with `LiteralValue` leaves. This limitation can cascade into non-optimal query plans or even complete planning failures for certain queries.

hatyo:
This commit implements two fixes to improve `null` literal handling in query processing.

**Fix #3500: Implement null-safe equality constraints in plan generator**

Updated the plan generator's literal equality constraint logic to properly handle `null` comparisons. The previous implementation used a simple `EQUALS` comparison, which incorrectly returned `null` when either operand was `null`, breaking equality constraints between identical literals when `null`s are involved.

The new predicate implements proper null-safe equality semantics: `(left IS NULL AND right IS NULL) OR (left IS NOT NULL AND right IS NOT NULL AND left = right)`. While this predicate is more complex and may impact plan cache performance, it will be replaced with a simpler `IS NOT DISTINCT FROM` operator once issue #3504 is resolved.

**Fix #3503: Prevent silent exclusion of null-bound OrderedLiterals**

Removed legacy behavior that silently excluded `null`-bound `OrderedLiteral` objects from query processing. This undocumented behavior was preventing predicate simplification optimizations, resulting in suboptimal query plans.

`null`-bound `OrderedLiteral` objects are now properly included in both the `Literals` collection and `EvaluationContext`, enabling correct dereferencing of `CoV` objects in query plans, and therefore, improved predicate simplification opportunities, resulting in more efficient physical query plans.

**Issues resolved:** #3500, #3503


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/Literals.java
```diff
@@ -57,11 +57,10 @@ public class Literals {
     private Literals(@Nonnull final List<OrderedLiteral> orderedLiterals) {
         this.orderedLiterals = ImmutableList.copyOf(orderedLiterals);
         this.asMapSupplier = Suppliers.memoize(() ->
-                this.orderedLiterals
-                        .stream()
-                        .filter(orderedLiteral -> orderedLiteral.getLiteralObject() != null)
-                        .collect(ImmutableMap.toImmutableMap(OrderedLiteral::getConstantId,
-                                OrderedLiteral::getLiteralObject)));
+                Collections.unmodifiableMap(
```
alecgrieser, line 63: I assume we're using `unmodifiableMap` here instead of `ImmutableMap` because the latter doesn't allow for nulls. If so, there should probably be a comment about that to explain the choice of data structure

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/Literals.java
```diff
@@ -57,11 +57,10 @@ public class Literals {
...
+                        this.orderedLiterals
+                                .stream()
+                                .collect(HashMap::new, (m, v) -> m.put(v.getConstantId(), v.getLiteralObject()), HashMap::putAll)));
```
alecgrieser: Should this be a `LinkedHashMap` so that the iteration order is the insertion order (rather than pseudorandom)? It's probably a code smell to iterate over this data structure given that `orderedLiterals` is right there, but it seems safer

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/LiteralsTests.java
```diff
@@ -0,0 +1,104 @@
...
+ * building a {@link Literals} object.
+ */
+public class LiteralsTests {
+
+    @Nonnull
+    private final Random random = new Random(42L);
+
+    @Test
+    void nullLiteralsAreNotExcluded() {
+        // TODO use RandomizedTestUtils.
```
alecgrieser, line 49: If you wanted to make this use `RandomizedTestUtils`, I don't think the steps are that hard. I think the steps are basically:

1. Make this test a parameterized test that takes the seed as its input
2. Introduce a method to the class that supplies the random seed(s). That method should call `RandomizedTestUtils.randomSeeds`
3. Move the `Random` initialization into the method. It already looks like the member variable `random` isn't referenced directly outside of this test method

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/MutablePlanGenerationContext.java
```diff
@@ -164,8 +166,24 @@ private void addEqualityConstraint(@Nonnull String leftTokenId, @Nonnull String
...
-        equalityConstraints.add(equalityPredicate);
+        final var notNullComparison = AndPredicate.and(ImmutableList.of(leftIsNotNull, rightIsNotNull, equalityPredicate));
+
+        // Term2: left = null AND right = null
+        final var leftIsNull = new ValuePredicate(leftCov, new Comparisons.NullComparison(Comparisons.Type.IS_NULL));
+        final var rightIsNull = new ValuePredicate(rightCov, new Comparisons.NullComparison(Comparisons.Type.IS_NULL));
+        final var bothAreNullComparison = AndPredicate.and(leftIsNull, rightIsNull);
+
+        // Term1 OR Term2
+        final var constraint = OrPredicate.or(notNullComparison, bothAreNullComparison);
```
alecgrieser, line 185: I was giving this query plan constraint a little more thought. The way I initially understood it, I thought the idea is that we want to assert that two extracted literals are not distinct so that we can re-use plans where the same parameter/literal value is used in multiple places, and sometimes that parameter/literal value is null and sometimes that parameter/literal value is not null. So, a query like:

```sql
SELECT * FROM T WHERE T.x = ?xParam AND T.y = ?yParam
```

The plan for `{xParam: 3, yParam: 3}` should be re-used for `{xParam: 4, yParam: 4}`. My thought twas that this was intending to make it also be-reused for `{xParam: null, yParam: null}`. But if there are _also_ query plan constraints validating the nullness of each parameter/literal, then we'd not expect an actual re-use there, right? Are there typing issues, even, with the COV sometimes being extracted as a `NULLABLE` type and sometimes a non-nullable type? Is this looking slightly forward to a world where we don't always include is null/non-null constraints if it's not relevant to the simplifications that were made to the query?

Is the actual idea more that we want to be able to re-use the same plan for multiple invocations of the query when the parameters are `{xParam: null, yParam: null}`? And it's easier to just make the constraint agnostic to the nullability rather than have separate branches, one for non-nullable types where only the equals predicate is necessary and another where `null` values just get `IS_NULL` constraints? Because there's part of me that wonders if it wouldn't be nicer to just always use only the equality predicate for non-null COVs, and then just rely on the `IS_NULL` constraint for nullable COVs

yaml-tests/src/test/resources/null-extraction-tests.yamsql
```diff
@@ -0,0 +1,60 @@
...
+        (5, 'c', 6)
+---
+test_block:
+  options:
+    statement_type: prepared
+    check_cache: true
+  tests:
+    -
+# With constant folding enabled, the condition in parentheses gets optimized away, allowing
+# the optimizer to use an efficient index scan. Without constant folding, the optimizer
```
alecgrieser, line 45: If we cared about asserting about the behavior without constant folding, we could add a version of this test case where we've disabled the transformation rule via a connection option

yaml-tests/src/test/resources/null-extraction-tests.yamsql
```diff
@@ -0,0 +1,60 @@
...
+      - query: select * from B where b3 = 4 and b2 = 'b' and (!! !n null !! is null or b1 < 20)
+      - explain: "ISCAN(I1 [EQUALS promote(@c8 AS LONG), EQUALS @c12])"
+      - result: [{3, 'b', 4}]
+    -
+# This query contains two `null` values, creating an equality constraint that handles `null`
+# comparisons with proper semantics. The constraint correctly evaluates equality for both
+# `null` and non-`null` objects. After the initial query execution stores the plan in cache,
+# subsequent runs of the same query will reuse the cached plan since the equality constraint
+# is satisfied when both arguments are `null`.
+      - query: select * from B where !! !n null !! = 42 and !! !n null !! = 43
```
alecgrieser, line 57: Are there tests about what happens if we sometimes put a null value in for these parameters and sometimes don't? That may be hard to make any concrete assertions about in the YAML framework.

alecgrieser (changes requested):

yaml-tests/src/test/resources/null-extraction-tests.yamsql
hatyo: It is probably a good idea to test pre-rewriter behaviour. I added a test block that performs the same test with planner rewriter disabled.

yaml-tests/src/test/resources/null-extraction-tests.yamsql
hatyo: I think so, but these tests are by no means providing good coverage. We should probably add more. This particular test though is focused about testing what happens when we have _two_ `null` parameters, as the manifestation of the bug the PR is fixing is related to having an improper equality constraint between `null` params. Nevertheless, I added few tests with variations of non-`null` and `null` parameters, with / without planner rewriter rules.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/Literals.java
hatyo: That's right, I will add a comment.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/Literals.java
hatyo: done

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/LiteralsTests.java
hatyo: Thanks, it would also require adding a dependency on `fdb-extensions`. I think I will do that in a separate PR.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/MutablePlanGenerationContext.java
hatyo: Equality constraints provide the plan generator with flexibility to reorganize literals during plan construction. This allows the planner to add constraints based on literal values independent of their position in the query, prioritizing safety over plan reusability.


The issue with multiple `null` literals is that they generate equality conditions that can never be satisfied. This prevents identical queries (containing multiple `null` literals) from being retrieved from cache, which is suboptimal. You correctly noted that `null` literals can create additional constraints that may prevent plan reuse for non-`null` literals. However, such conditions should only be added when justified by actual optimization decisions (e.g., simplifying `IS_NULL(?xParam => null)` only when `xParam` is definitively `null`). As you suggested, these optimization decisions should be handled by the rewrite engine or other appropriate components in the future.


This PR aims to make plan generator constraints behave correctly while remaining agnostic to actual literal values.

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -297,7 +297,6 @@ public void transactionalCallsTest(YamlTest.Runner runner) throws Exception {
     }
 
     @TestTemplate
-    @MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
```
alecgrieser: Good catch! I thought there was an issue somewhere to make this a PRB failure, but I can't actually find it

yaml-tests/src/test/resources/null-extraction-tests.yamsql
```diff
@@ -55,6 +55,55 @@ test_block:
 # subsequent runs of the same query will reuse the cached plan since the equality constraint
 # is satisfied when both arguments are `null`.
       - query: select * from B where !! !n null !! = 42 and !! !n null !! = 43
-      - explain: "COVERING(I1 <,> -> [B1: KEY[3], B2: KEY[1], B3: KEY[0]]) | FILTER promote(@c8 AS INT) EQUALS @c6 AND promote(@c12 AS INT) EQUALS @c6 | FETCH"
+      - explainContains: "COVERING(I1 <,> -> [B1: KEY[3], B2: KEY[1], B3: KEY[0]])"
```
alecgrieser, line 58: Is there a reason these are `explainContains` instead of `explain`? I've found `explainContains` can be harder to work with as it won't necessarily get fixed by the correction mode. Does this need to use `explainContains` because it's unstable which plan gets used or something due to the plan cache or something like that?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/MutablePlanGenerationContext.java
alecgrieser: Sure, fair enough. This seems forward looking, though I did want to make sure that we understood the short-term implications as well (in a world where we do always get the nullability constraints). Most importantly, I think it would have implications on what test assertions around cache hits we'd expect for running the same plan with different null/non-null values placed in for parameters.

alecgrieser (approved):
LGTM. The use of `explainContains` is probably fine, but I'll leave it up to you whether this would be appropriate to switch to `explain`

yaml-tests/src/test/resources/null-extraction-tests.yamsql
hatyo: The tests within the `test_block` are running on parallel. Therefore, there is a chance for multiple permissible concurrent plan caching and retrieval schedules, the problem is, the test is documenting only _one_ with `explain`, and therefore, it will "fail" if another schedule was chosen at runtime. To avoid this, I am verifying only the common part of the plans with `explainContains`.

To illustrate this, here is an example:

```sql
# query1:
select * from t where x = 4 and y = 5

# query2:
select * from t where x = 4 and y = 4
```
There are _two_ permissible schedules when we run this concurrently, let's say the constants are:
```
select * from t where x = 4 and y = 5
                          ^         ^
                          @7        @11
```

Schedule1:
```
time
   1    plan and cache query1, the constraint is: OfTypeInt(@7) and OfTypeInt(@11)
   2    retrieve plan for query1 to execute query2, because query2's constants satisfy the constraint OfTypeInt(@7) and OfTypeInt(@11)
```

Schedule2:
```
time
   1    plan and cache query2, the constraint is: OfTypeInt(@7) and OfTypeInt(@11) and Equals(@7, @11)
   2    plan and cache query1, the constraint is: OfTypeInt(@7) and OfTypeInt(@11), (query1 constants do NOT satisfy the constraint for query2
```

(there is another schedule, it exhibits Schedule2 and represents the case where both queries are planned at the same time, so when each thread probed the cache, they got empty result, in that case, we end up storing two different physical plans).

hatyo:
> LGTM. The use of `explainContains` is probably fine, but I'll leave it up to you whether this would be appropriate to switch to `explain`

Thanks, I've left a comment explaining why `explainContains` was necessary for this test.

yaml-tests/src/test/resources/null-extraction-tests.yamsql
alecgrieser: I see. I figured that it might be something like that (that's what I meant by "it's unstable which plan gets used or something due to the plan cache"). It would probably good to have that written down in the code, as we've had other cases with similar setups, and then been surprised later when we try to "fix" the test case and it is now broken

PR #3508, opened 2025-07-28
Update to minor version 4.5 in recognition of incompatibility with older 4.3 versions

alecgrieser:
With #3495, we are no longer compatible in mixed modes with versions that do not have #3254, which was introduced in version 4.3.4.0. To highlight that this version can be safely upgraded from 4.4 but requires more care when upgrading from 4.3, this updates the minor version to 4.5.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@
 #
 
 rootProject.name=fdb-record-layer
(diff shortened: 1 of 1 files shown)


hatyo (approved):
LGTM

PR #3510, opened 2025-07-29
Bring back full support for UUID

g31pranjal:
This PR brings back the complete support for UUID as a primitive data type. The support was first brought in with PR:https://github.com/FoundationDB/fdb-record-layer/pull/3198, but was later partially rolled back in https://github.com/FoundationDB/fdb-record-layer/pull/3293. 

It also brings in UUID-relevant parts from patch: https://github.com/FoundationDB/fdb-record-layer/pull/3395, that was based on top 4.2.4.0, onto the main.


g31pranjal:
We probably want to cut a release with the intermediate support for UUID in JDBC server, before bringing this PR in.
Ref: https://github.com/FoundationDB/fdb-record-layer/pull/3515

g31pranjal:
This is good to review now!

yaml-tests/src/test/resources/uuid.yamsql
```diff
@@ -0,0 +1,247 @@
...
+    -
+      - query: INSERT INTO TE VALUES (1, !! !uuid 'c35ba01f-f8fc-47d7-bb00-f077e8a75682' !!), (2, !! !uuid '64120112-4e39-40c3-94b9-2cc88a52e8df' !!)
+      - count: 2
+    -
+      - query: INSERT INTO TD VALUES (1, (2, !! !uuid 'c35ba01f-f8fc-47d7-bb00-f077e8a75682' !!)), (2, (3, !! !uuid '64120112-4e39-40c3-94b9-2cc88a52e8df' !!))
+      - count: 2
+---
+test_block:
+  name: uuid-as-a-field-tests
+  preset: single_repetition_ordered
```
hatyo: is there a reason for adding an explicit preset? if not, can we remove it?

yaml-tests/src/test/resources/uuid.yamsql
```diff
@@ -0,0 +1,247 @@
...
+      - unorderedResult: [{3, !uuid 'c35ba01f-f8fc-47d7-bb00-f077e8a75682'},
+                          {5, !uuid 'a8708750-d70f-4800-8c3b-13700d5b369f'}]
+    -
+      - query: select * from ta where b < !! !uuid '99e8e8b1-ac34-4f4d-9f01-1f4a7debf4d6' !!
+      - unorderedResult: [{1, !uuid '0920df1c-be81-4ec1-8a06-2180226f051d'},
+                          {2, !uuid '64120112-4e39-40c3-94b9-2cc88a52e8df'},
+                          {4, !uuid '5394a912-aa8e-40fc-a4bb-ddf3f89ac45b'}]
+    -
+      - query: select * from ta where b >= !! !uuid '99e8e8b1-ac34-4f4d-9f01-1f4a7debf4d6' !!
+      - unorderedResult: [{3, !uuid 'c35ba01f-f8fc-47d7-bb00-f077e8a75682'},
```
hatyo, line 101: can you please verify the plan as well?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -434,6 +434,8 @@ private static Type fromProtoType(@Nullable Descriptors.GenericDescriptor descri
                 final var elementField = messageDescriptor.findFieldByName(NullableArrayTypeUtils.getRepeatedFieldName());
                 final var elementTypeCode = TypeCode.fromProtobufType(elementField.getType());
                 return fromProtoTypeToArray(descriptor, protoType, elementTypeCode, true);
+            } else if (TupleFieldsProto.UUID.getDescriptor().equals(messageDescriptor)) {
```
hatyo, line 437: I think we may have discussed this before, but can you please remind me again why are handling the conversion of `UUID` PB message within a `Struct` but not within and `Array`?

hatyo (changes requested):
LGTM, very minor comments.

yaml-tests/src/test/resources/uuid.yamsql
g31pranjal: Yeah, that was stray code. Removed!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
g31pranjal: We are adding an extra check here for the case of `RECORD`, since the descriptor could be that representing a `UUID`. If so, we do not consider it a "Record" type, but a "UUID" type. This is only needed for this particular case, and not also in Arrays as that dispatches the element processing to this function only.

hatyo (approved):

Issue #3498, opened 2025-07-23
Implement file-wide connection options in YAML


hatyo:
YAML test files have a dedicated region for file options that currently supports setting the supported version. It would be helpful to extend this region to support global connection `Options`. This enhancement would bring YAML testing closer to simulating end-to-end usage of RelationalFDB, where customers may define multiple options when establishing a connection and use these options (or override them) when running SQL statements.

Similar work was completed in #3438, where connection `Options` were introduced at the test-block level only.,

PR #3511, opened 2025-07-29
Fix case-sensitivity handling

### LINKED ISSUE #3513: Handle case-sensitivity option `CASE_SENSITIVE_IDENTIFIERS` correctly
The `CASE_SENSITIVE_IDENTIFIERS` option controls how the system processes SQL identifiers based on case sensitivity. When enabled, unquoted identifiers are processed exactly as written, preserving their original case. When disabled, unquoted identifiers are treated case-insensitively, while double-quoted identifiers always retain their exact case regardless of the setting.

Currently, this option suffers from implementation issues in the codebase where it is not handled correctly or consistently applied throughout the system. Additionally, the testing infrastructure lacks proper support for validating this functionality, making it difficult to verify correct behavior across both DDL and DQL operations. These problems prevent reliable case-sensitivity behavior and hinder comprehensive testing of the feature.

hatyo:
This PR addresses the correct handling of `CASE_SENSITIVE_IDENTIFIERS` throughout the codebase by fixing multiple issues where this option was previously ignored during parsing, logical plan generation, and `SchemaTemplate` interactions. The implementation ensures consistent case-sensitivity behavior across all components that process SQL identifiers.

The PR also enhances both testing frameworks to properly support this functionality. The JUnit test framework now accepts `Options` when performing DDL operations through the custom `Extension`s for database, schema template, and schema creation, enabling comprehensive testing scenarios.

Additionally, the YAML test framework introduces file-wide connection `Options` that influence how schema template DDL statements are parsed and persisted, while allowing developers to test various case-sensitivity scenarios by contrasting file-wide Options with block-specific Options for thorough validation coverage.

This fixes #3513, and #3498.

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/catalog/RecordLayerStoreSchemaTemplateCatalog.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/catalog/RecordLayerStoreSchemaTemplateCatalog.java
@@ -173,7 +173,7 @@ class RecordLayerStoreSchemaTemplateCatalog implements SchemaTemplateCatalog {
         final var tupleRange = new TupleRange(key, key, EndpointType.RANGE_INCLUSIVE, EndpointType.RANGE_INCLUSIVE);
         try (var cursor = recordStore.scanRecords(tupleRange, ContinuationImpl.BEGIN.getExecutionState(), ScanProperties.REVERSE_SCAN)) {
             final var cursorResult = cursor.getNext();
-            final var schemaExists = cursorResult.get() != null && !cursorResult.getContinuation().isEnd();
+            final var schemaExists = !cursorResult.getContinuation().isEnd() && cursorResult.get() != null;
             Assert.thatUnchecked(schemaExists, ErrorCode.UNKNOWN_SCHEMA_TEMPLATE,
                     "SchemaTemplate '" + templateName + "' is not in catalog");
             return toSchemaTemplate(Assert.notNullUnchecked(cursorResult.get()).getRecord());


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/Block.java
```diff
@@ -42,17 +43,38 @@
...
+        try {
+            if (blockKey.equals(PreambleBlock.OPTIONS)) {
+                return Optional.of(PreambleBlock.parse(lineNumber, entry.getValue(), executionContextBuilder));
+            }
+        } catch (Exception e) {
+            throw YamlExecutionContext.wrapContext(executionContextBuilder.getResourcePath(), e,
+                    () -> "Error parsing block at line " + lineNumber, blockKey, lineNumber);
+        }
+        return Optional.empty();
+    }
```
g31pranjal: I think I am a bit concerned about the out-of-the-ordinary "behaviour" of how we deal with the preamble block. I am a bit afraid that this could lead to other cases where "parsing" and "execution" of a block is different from the usual. Minor, but it seems to replicate a bit of logic too from the general parsing.

It would be great if we could follow the general notion of "parsing" and "execution".

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlRunner.java
```diff
@@ -59,43 +62,55 @@ public final class YamlRunner {
...
+                // Step 1/3: process preamble block (if any)
+                Optional<PreambleBlock> preambleBlockMaybe = Optional.empty();
+                if (!regions.isEmpty()) {
+                    preambleBlockMaybe = Block.parsePreambleBlock(regions.get(0), executionContextBuilder);
+                    preambleBlockMaybe.ifPresent(preambleBlock ->
+                            executionContextBuilder.setConnectionOptions(preambleBlock.getBlockOptions().getOptions()));
+                }
+
+                // Step 2/3: build the YAML execution context, potentially with file-specific options set in the preamble block
+                executionContext = executionContextBuilder.build();
```
g31pranjal: It seems to me that step 1/3 is `parse` of PreambleBlock, while 2/3 is `execute` of it which ends up changing the underlying `ExecutionContext` by adding CONNECTION_OPTION to it (which could then be overridden by test_block options). I think my suggestion won't work with builder() pattern as the same ExecutionContext is shared by all the blocks in parse-time. However, we already do "add to" the executionContext, as in `TransactionSetupsBlock.java`. I believe it could be the same for the PreambleBlock as well!

In general, is there any specific reason to deal with PreambleBlock separately ? My understanding is that we may want to stick to the existing paradigm of `parse` and `execute` as much as possible - as it maintain homogeneity. And, it feels this use-case could be covered well.

yaml-tests/src/test/resources/case-sensitivity.yamsql
```diff
@@ -0,0 +1,153 @@
+#
+# case-sensitivity.yamsql
+#
+# This source file is part of the FoundationDB open source project
+#
+# Copyright 2021-2024 Apple Inc. and the FoundationDB project authors
```
g31pranjal: Is it possible somehow to sanitize this everywhere? I see a couple of places being changed to 2015-2025, which we probably want to do here as well?

yaml-tests/src/test/resources/case-sensitivity.yamsql
```diff
@@ -0,0 +1,153 @@
...
+        values (10, 'blah'),
+               (20, 'bleh')
+    - query: insert into table2
+        values (10, 100, 10),
+               (20, 100, 20),
+               (60, 200, 60),
+               (70, 200, 70)
+---
+test_block:
+  name: case-sensitivity-on
```
g31pranjal, line 60: Do you think it makes sense to add tests here that tests with temporary functions as well? These tests seem to be a good contender for using the new transactions_setup_block that Scott introduced recently to create a temporary function inside the transaction.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlRunner.java
hatyo: I initially attempted to make `executionContext` immutable by using a builder pattern instead of direct state modification. However, this approach proved overly complex, so I reverted to the original design and simply added an additional setter method for configuring connection options.

Additionally, I refactored `PreambleBlock` to better align with the existing block parsing architecture. As you noted, this makes it more consistent and streamlined with the rest of the codebase.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/Block.java
hatyo: See my other comment.

yaml-tests/src/test/resources/case-sensitivity.yamsql
hatyo: Nice idea, let me do that after bringing in this PR.

yaml-tests/src/test/resources/case-sensitivity.yamsql
hatyo: https://github.com/FoundationDB/fdb-record-layer/issues/3527

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/PreambleBlock.java
```diff
@@ -1,9 +1,9 @@
 /*
- * FileOptions.java
+ * NoOpBlock.java
```
g31pranjal: filename

g31pranjal (changes requested):
Pretty minor request!

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/block/PreambleBlock.java
hatyo: nice catch. Done.

PR #3512, opened 2025-07-29
Include stats for each planner phase in EXPLAIN metrics

hazefully:
This PR adds new planner statistics  to the `PLANNER_METRICS` struct produced for `EXPLAIN` SQL queries which reflect the number of tasks in each planner phase and the total time the tasks for each phase took to execute. It also adds additional metrics for rule call events that are emitted with the locations `DISCARDED_INTERSECTION_COMBINATIONS` and `ALL_INTERSECTION_COMBINATIONS` to the `PLANNER_METRICS`. 

After the changes in this PR, this is the `PLANNER_METRICS` struct returned by an `EXPLAIN` query:

```
PLANNER_METRICS -> { 
    TASK_COUNT -> 248
    , TASK_TOTAL_TIME_NS -> 112758330
    , TRANSFORM_COUNT -> 65
    , TRANSFORM_TIME_NS -> 89059163
    , TRANSFORM_YIELD_COUNT -> 29
    , INSERT_TIME_NS -> 1683416
    , INSERT_NEW_COUNT -> 22
    , INSERT_REUSED_COUNT -> 6
    , TRANSFORM_DISCARDED_INTERSECTION_COMBINATIONS_COUNT -> 0
    , TRANSFORM_ALL_INTERSECTION_COMBINATIONS_COUNT -> 0
    , REWRITING_PHASE_TASK_COUNT -> 44
    , PLANNING_PHASE_TASK_COUNT -> 204
    , REWRITING_PHASE_TASKS_TOTAL_TIME_NS -> 9047705
    , PLANNING_PHASE_TASKS_TOTAL_TIME_NS -> 103710625
}
```

To make this `PLANNER_METRICS` struct more useful, this PR changes the events which are emitted by the `CascadesPlanner` throughout the planning of the query in a fundamental way, to make them first-class objects that are no longer tied to the existence of a `Debugger`. This is because the implementations of `Debugger` that we have are only meant to be used in development or tests, and never in production, as they may do heavyweight processing that can impact performance. The PR achieves this by doing the following changes:

- Create a new package `com.apple.foundationdb.record.query.plan.cascades.events`, and move all planner events that were inner classes in `Debugger.java` to this new package under the name `PlannerEvent`. This makes it clear that planner events are no longer objects that are specifically tied to the Debugger anymore.

- Introduce a new entry point for emitting planner events throughout the planning of a query, a new class `PlannerEventListeners`, which has hooks for adding objects as receivers of planner events when they are emitted. The thread local storage of `Debugger` instances and other objects that would like to listen to planner events is moved to this class as well. All objects that would like to subscribe to planner events and do some action on them would need to implement the interface `PlannerEventListeners.Listener`, which the `Debugger` interface now extends.

- Create a new interface called `PlannerEventStatsCollector`, which takes over the functionality of collecting and aggregating statistics of planner events from the different `Debugger` implementations and makes them available via `PlannerEventStatsMaps`, which provide access to planner event statistics by Java class (a class that implements `PlannerEvent`) or planner phase. This makes the `Debugger` implementation only responsible for storing events as protos, comparing them with previous recorded events, conditionally running sanity checks during planning and keeping track of different symbols/entities seen during planning. An implementation of that interface,  `DefaultPlannerEventStatsCollector`, is provided which can be used to enable stats collection. The default collector can be installed by just calling the method `PlannerEventStatsCollector.enableDefaultStatsCollector`, which is called whenever a `Debugger` is set for connivence.

- The event statistics aggregation logic, which can now be found in `PlannerEventStatsCollectorState`, takes care of aggregating events stats by class and planner phase to provide these different views via `PlannerEventStatsMaps` later when needed.

- Finally, all places where a planner event is emitted are updated to make use of the new entry point `PlannerEventListeners.dispatchEvent` instead of calling the Debugger directly.

As part of this PR, unit tests are added for the different Debugger implementations we have to improve their test coverage, with the exception of the `PlannerEventStatsViewer` class which is used to render the stats as HTML in a browser during an interactive Java debugging session (so it is only used manually during debugging), as there is no easy way to unit test that without major refactoring. This is how the HTML output of `PlannerEventStatsViewer` would look like after this PR: 

image

Note: This PR took different approaches initially, which are documented below only for historical reasons. There is no need to read them for purposes of reviewing this PR. Because of this, it is better not to review this PR commit by commit.

<details>

<summary>Previous approaches this PR took</summary>

### Approach 1

This PR previously took a different approach to expose the duration each planner phase took ([12b5817](https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/12b5817799845aa3f75a147fb9015c23f349d9c4)), but after offline discussions we decided to make use of the existing Cascades debugger to expose these metrics.

### Approach 2 

This PR makes use of the `Debugger` interface to collect more statistics about events by planner phase in the following way:

- 2582ff8: This commit splits the `cascades.debug.Debugger` interface into two
separate interfaces (`StatsDebugger` and `SymbolDebugger`) which both extend the base Debugger interface, in
order to ensure that operations related to recording Debugger events (with the purpose of calculating stats
based on the profiling of these events) make use of the `StatsDebugger` interface, and similarly operations related to
registering different objects used within the planner make use of the SymbolDebugger interface. The `State` class which kept track of symbols and debugger events is also split into two different classes: `EventState` and `SymbolTables`.
- bdf2e16: This commit extracts the functionality of rendering the different stats for debugger events from the `State` class to a separate `StatsViewer` class.
- cf9848b and a9edd0f: These two commits include the main work of grouping debugger events by the planner phase they were emitted in, and exposing that grouping in the `EXPLAIN` planner metrics. As a result of this grouping the "Event Profiling" rendered by the `StatsViewer` class is now split by phase as well:

   image
 
    In addition, the result set of an `EXPLAIN` query contains the expected planner metrics:
  ```
    { TASK_COUNT -> 248
    , TASK_TOTAL_TIME_NS -> 98967370
    , TRANSFORM_COUNT -> 61
    , TRANSFORM_TIME_NS -> 77887334
    , TRANSFORM_YIELD_COUNT -> 29
    , INSERT_TIME_NS -> 2012374
    , INSERT_NEW_COUNT -> 22
    , INSERT_REUSED_COUNT -> 6
    , REWRITING_PHASE_TASK_COUNT -> 44
    , PLANNING_PHASE_TASK_COUNT -> 204
    , REWRITING_PHASE_TASKS_TOTAL_TIME_NS -> 7869126
    , PLANNING_PHASE_TASKS_TOTAL_TIME_NS -> 91098244
    } 
   ```
- 2ee4763: This commit adds the new lightweight debugger implementation that only implements the `StatsDebugger` interface.

</details>

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,7 +16,7 @@ jobs:
   # The mixed-mode coverage is a little different, we want to see explicitly what code is being covered in mixed-mode
   # so that we can find potential uncovered code that could expose bugs when multiple versions are running at the same
   # time.
-  # It's possible that we could change the nightly tests to just be the new tests, and not the regular tests in the future.
+  # It's possible that we could change the nightly tests to just be the nightly tests, and not the regular tests in the future.
   test:
     if: github.repository == 'FoundationDB/fdb-record-layer'
     runs-on: ubuntu-latest
@@ -45,11 +45,38 @@ jobs:
           files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
           teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
 
-  mixed-mode-test:
+  ### Mixed mode
+  # Get the versions of mixed-mode that we want to run
+  get-mixed-mode-versions:
     if: github.repository == 'FoundationDB/fdb-record-layer'
     runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.versions.outputs.versions }}
     permissions:
       contents: read
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4.2.2
+      - name: Setup Base Environment
+        uses: ./actions/setup-base-env
+      - name: Calculate versions
+        id: versions
+        uses: ./actions/list-recent-versions
+
+  # Run the actual mixed-mode tests
+  # This runs them in parallel by version, which is unnecessary, because it's nightly and we don't care if it takes
+  # 20 minutes or 2 hours, but this way we have the same config for PRB, nightly, and release
+  mixed-mode:
+    if: github.repository == 'FoundationDB/fdb-record-layer'
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    needs: get-mixed-mode-versions
+    strategy:
+      matrix:
+        version: ${{ fromJSON(needs.get-mixed-mode-versions.outputs.versions) }}
+    permissions:
+      contents: read
+    timeout-minutes: 40
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.2.2
@@ -61,26 +88,26 @@ jobs:
         uses: ./actions/gradle-test
         with:
           gradle_command: jar mixedModeTest codeCoverageReport
-          gradle_args: -PreleaseBuild=false -PpublishBuild=false
-          report_name: mixed-mode-test-reports
-      # We don't commit the incremented version, but we use this to know the version when generating
-      # the resulting markdown
-      - name: Increment version
-        shell: bash
-        run: python build/versionutils.py gradle.properties --increment -u PATCH
-      - name: Get new version
-        id: get_new_version
-        shell: bash
-        run: |
-          echo "version=$(python build/versionutils.py gradle.properties)" >> "$GITHUB_OUTPUT"
-      - name: Create markdown
-        shell: bash
-        run: python build/publish-mixed-mode-results.py ${{ steps.get_new_version.outputs.version }} --run-link ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }} --output mixed-mode-results.md
-      - name: Preview results
-        shell: bash
-        run: cat mixed-mode-results.md >> $GITHUB_STEP_SUMMARY
+          # Setting tests.nightly would cause it to ignore failures at the gradle level.
+          # But, in the case of mixed-mode-test we are runnning with continue-on-error and using the result of the gradle
+          # task to populate the summary. Thus, there's no benefit to ignoring failures at the gradle level, and if we
+          # did, we'd need more complicated logic to have the results correctly display that an individual version
+          # failed.
+          # If we ever have more than one module that has mixed-mode test, or yaml-tests starts using nightly in a way
+          # that we want to impact mixedModeTest we'll probably want to change this, but for now we can keep it simple
+          gradle_args: -PreleaseBuild=false -PpublishBuild=false -Ptests.mixedModeVersion=${{ matrix.version }}
+          report_name: mixed-mode-${{ matrix.version }}-test-reports
+      - name: Publish Coverage Data
+        uses: actions/upload-artifact@v4.6.0
+        with:
+          name: mixed-mode-${{ matrix.version }}-coverage-data
+          path: |
+            **/.out/jacoco/*.exec
+            **/.out/libs/*.jar
+          include-hidden-files: true
+          retention-days: 1
+      # This will upload multiple times, once for each of the old versions
       - name: Upload coverage to teamscale
-        # temporary until we validate that this is working correctly
         continue-on-error: true
         uses: ./actions/teamscale-upload
         with:
@@ -89,6 +116,26 @@ jobs:
           files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
           teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
 
+  # Print out the mixed mode results to the summary
+  mixed-mode-results:
+    if: github.repository == 'FoundationDB/fdb-record-layer'
+    needs: [mixed-mode]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout HEAD sources
+        uses: actions/checkout@v4.2.2
+      - name: Setup Base Environment
+        uses: ./actions/setup-base-env
+      - name: 'Download results'
+        uses: actions/download-artifact@v4
+        with:
+          pattern: 'mixed-mode-*-test-reports'
+      - name: Generate mixed mode results
+        uses: ./actions/publish-mixed-mode-results
+
+  ### Nightly
   nightly_test:
     if: github.repository == 'FoundationDB/fdb-record-layer'
     runs-on: ubuntu-latest
@@ -115,6 +162,3 @@ jobs:
           revision: ${{ github.sha }}
           files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
           teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
-
-
-
--- a/.github/workflows/pr_mixed_mode.yml
+++ b/.github/workflows/pr_mixed_mode.yml
@@ -1,0 +1,75 @@
+name: Pull Request Mixed Mode
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+
+jobs:
+  get-mixed-mode-versions:
+    # Run these jobs only when adding the label, or when updating HEAD when it already has the label
+    if: ${{ github.event.label.name == 'Run mixed-mode' || (github.event.label.name == '' && contains(github.event.pull_request.labels.*.name, 'Run mixed-mode')) }}
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.versions.outputs.versions }}
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4.2.2
+      - name: Setup Base Environment
+        uses: ./actions/setup-base-env
+      - name: Calculate versions
+        id: versions
+        uses: ./actions/list-recent-versions
+
+  mixed-mode:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    needs: get-mixed-mode-versions
+    strategy:
+      matrix:
+        version: ${{ fromJSON(needs.get-mixed-mode-versions.outputs.versions) }}
+    permissions:
+      checks: write
+      contents: read
+      pull-requests: write
+    timeout-minutes: 40
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4.2.2
+      - name: Setup Base Environment
+        uses: ./actions/setup-base-env
+      - name: Setup FDB
+        uses: ./actions/setup-fdb
+      - name: Run Gradle Test
+        uses: ./actions/gradle-test
+        with:
+          gradle_command: mixedModeTest
+          gradle_args: -PreleaseBuild=false -PpublishBuild=false -Ptests.mixedModeVersion=${{ matrix.version }}
+          report_name: mixed-mode-${{ matrix.version }}-test-reports
+      - name: Publish Coverage Data
+        uses: actions/upload-artifact@v4.6.0
+        with:
+          name: mixed-mode-${{ matrix.version }}-coverage-data
+          path: |
+            **/.out/jacoco/*.exec
+            **/.out/libs/*.jar
+          include-hidden-files: true
+          retention-days: 1
+
+  mixed-mode-results:
+    needs: [mixed-mode]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout HEAD sources
+        uses: actions/checkout@v4.2.2
+      - name: Setup Base Environment
+        uses: ./actions/setup-base-env
+      - name: 'Download results'
+        uses: actions/download-artifact@v4
+        with:
+          pattern: 'mixed-mode-*-test-reports'
+      - name: Generate mixed mode results
+        uses: ./actions/publish-mixed-mode-results
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,6 +152,18 @@ Similar to `quickTest` above, except instead of running with an
 `EmbeddedRelationalConnection` it runs against JDBC-In-Process, which can be helpful if
 you are working on an aspect of the RPC protocol. This can be run with `./gradlew rpcTest`
 
+#### mixedModeTest
+
+For tests that use `@YamlTest`, this will run all the tests with the test alternating between talking to the current
+version and some older version. This helps ensure that we remain backwards compatibility for data and continuations.
+This can be run with `./gradlew mixedModeTest`.
+By default it runs against the 10 most recent versions, but it can be run against a single version by:
+`./gradlew mixedModeTest -Ptests.mixedModeVersion=${{ matrix.version }}`.
+
+These tests are run during the nightly build and also as part of the release process. By default, they are not run
+during PRB, but if you want to validate, you can add the
+[Run mixed-mode](https://github.com/FoundationDB/fdb-record-layer/labels/Run%20mixed-mode) label.
+
 #### singleVersionTest
 
 For tests that use `@YamlTest`, this will run all the tests with the test connecting just
--- a/actions/gradle-test/action.yml
+++ b/actions/gradle-test/action.yml
@@ -42,6 +42,16 @@ runs:
            ln -s "../$d/.out/reports" "test-reports/$d"
          fi
        done
+  - name: Overall failure
+    if: ${{ failure() }}
+    shell: bash
+    run: |
+      echo "FAILURE" >> test-reports/overall.txt
+  - name: Overall success
+    if: ${{ success() }}
+    shell: bash
+    run: |
+      echo "SUCCESS" >> test-reports/overall.txt
   - name: Publish Test Reports
     if: ${{ !cancelled() }}
     uses: actions/upload-artifact@v4.6.0
@@ -65,3 +75,4 @@ runs:
             test-reports/fdb-relational-jdbc/
             test-reports/fdb-relational-server/
             test-reports/yaml-tests/
+            test-reports/overall.txt
--- a/actions/list-recent-versions/action.yml
+++ b/actions/list-recent-versions/action.yml
@@ -1,0 +1,22 @@
+name: List Recent Server Versions
+
+outputs:
+  versions:
+    description: "The versions as json"
+    value: ${{ steps.versions.outputs.versions }}
+
+runs:
+  using: "composite"
+  steps:
+  - name: Calculate versions
+    uses: ./actions/run-gradle
+    with:
+      gradle_command: listRecentVersions
+  - name: create output
+    id: versions
+    shell: bash
+    # This is, somewhat unfortunately tied to yaml-tests, but we don't currently have anything else that supports
+    # mixed-mode in reality. We'll have to assess how to deal with that in order to have mixed-mode tests elsewhere
+    run: |
+     cat yaml-tests/.out/recentVersions.txt >> "$GITHUB_OUTPUT"
+
--- a/actions/publish-mixed-mode-results/action.yml
+++ b/actions/publish-mixed-mode-results/action.yml
@@ -1,0 +1,11 @@
+name: Publish mixed mode results to step summary
+
+runs:
+  using: "composite"
+  steps:
+  - name: Generate markdown
+    shell: bash
+    run: python3 build/publish-mixed-mode-results.py --results-path "mixed-mode-*-test-reports/overall.txt" --output mixed-mode-results.md $(python build/versionutils.py gradle.properties)
+  - name: Preview mixed-mode results
+    shell: bash
+    run: cat mixed-mode-results.md >> $GITHUB_STEP_SUMMARY
--- a/build/publish-mixed-mode-results.py
+++ b/build/publish-mixed-mode-results.py
@@ -24,6 +24,8 @@
 # ./gradlew mixedModeTest
 
 import argparse
+import glob
+import re
 import subprocess
 import sys
 
@@ -52,6 +54,24 @@ def get_results(results_path):
                 results[version] = result
     return results
 
+def get_results_from_glob(results_glob):
+    version_pattern = re.compile(r'\d+(?:\.\d+)+')
+    results = {}
+    for filepath in glob.glob(results_glob):
+        versions = version_pattern.findall(filepath)
+        if len(versions) != 1:
+            raise Exception("Expected exactly one version in path " + filepath + ", found: " + str(versions))
+        version = versions[0]
+        with open(filepath) as f:
+            result = f.read().strip()
+        if result not in ('SUCCESS', 'FAILURE'):
+            raise Exception("File " + filepath + " contains invalid result: " + result)
+        if result == 'FAILURE':
+            results[version] = result
+        elif version not in results:
+            results[version] = result
+    return results
+
 def emoji(result_word):
     if result_word == 'FAILURE':
         return '❌'
@@ -69,14 +89,19 @@ def generate_markdown(version, results, header_size):
 def main(argv):
     '''Process the output of a mixedModeTest run and convert it into a short markdown'''
     parser = argparse.ArgumentParser()
-    parser.add_argument('--results-path', help='Path to the results', default='.out/reports/mixed-mode-results.log')
+    parser.add_argument('--results-path', help='Path to the results file, or a glob matching per-version result files', default='.out/reports/mixed-mode-results.log')
     parser.add_argument('--header-size', help='Markdown header level (e.g. # or ##)', default='####')
     parser.add_argument('--run-link', help='A link to the test run that generated the results')
     parser.add_argument('--output', required=True, help='Output to print the markdown to')
     parser.add_argument('version', help='Version of the server that was tested')
     args = parser.parse_args(argv)
 
-    markdown = generate_markdown(args.version, get_results(args.results_path), args.header_size)
+    if glob.has_magic(args.results_path):
+        results = get_results_from_glob(args.results_path)
+    else:
+        results = get_results(args.results_path)
+
+    markdown = generate_markdown(args.version, results, args.header_size)
     if args.run_link is not None:
         markdown = markdown + "\n\n[See full test run](" + args.run_link +")\n"
     with open(args.output, mode='w') as fout:
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Primitives.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Primitives.java
@@ -323,7 +323,7 @@ public class Primitives {
                                         @Nonnull final Map<Tuple, AbstractNode<N>> nodeCache) {
         return fetchSomeNodesAndApply(storageAdapter, readTransaction, storageTransform, layer, neighborReferences,
                 neighborReference -> {
-                    if (neighborReference.isNodeReferenceWithVector()) {
+                    if (storageAdapter.isInliningStorageAdapter() && neighborReference.isNodeReferenceWithVector()) {
                         return neighborReference.asNodeReferenceWithVector();
                     }
                     final AbstractNode<N> neighborNode = nodeCache.get(neighborReference.getPrimaryKey());
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Search.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Search.java
@@ -496,74 +496,95 @@ public class Search {
                             @Nonnull final ToDoubleFunction<Transformed<RealVector>> objectiveFunction,
                             @Nonnull final Map<Tuple, AbstractNode<N>> nodeCache) {
         final Primitives primitives = primitives();
-        final Queue<NodeReferenceWithDistance> candidates =
-                // This initial capacity is somewhat arbitrary as m is not necessarily a limit,
-                // but it gives us a number that is better than the default.
-                new PriorityQueue<>(getConfig().getM(), NodeReferenceWithDistance.comparator());
-        candidates.addAll(nodeReferences);
-        final Set<Tuple> visited = Sets.newConcurrentHashSet(NodeReference.primaryKeys(nodeReferences));
-        final Queue<NodeReferenceWithDistance> nearestNeighbors =
-                new PriorityQueue<>(efSearch + 1, // prevent reallocation further down
-                        NodeReferenceWithDistance.reversedComparator());
-        nearestNeighbors.addAll(nodeReferences);
 
-        return AsyncUtil.whileTrue(() -> {
-            if (candidates.isEmpty()) {
-                return AsyncUtil.READY_FALSE;
-            }
+        //
+        // Refetch node references which we should start from as they potentially come from a higher layer.
+        // Semantically, that is fine, however, on a lower level, vectors may be encoded differently.
+        // We first refetch the node references we already synthesized on a higher level for this level in order
+        // to ensure that everything is identical to what is stored in the database.
+        //
+        return primitives.fetchNeighborhoodReferences(storageAdapter, readTransaction, storageTransform, layer,
+                        nodeReferences, nodeCache)
+                .thenApply(refetchedNodeReferences -> {
+                    final ImmutableList.Builder<NodeReferenceWithDistance> nodeReferencesWithDistanceBuilder =
+                            ImmutableList.builder();
+                    for (final NodeReferenceWithVector refetchedNodeReference : refetchedNodeReferences) {
+                        nodeReferencesWithDistanceBuilder.add(new NodeReferenceWithDistance(refetchedNodeReference.getPrimaryKey(),
+                                refetchedNodeReference.getVector(),
+                                objectiveFunction.applyAsDouble(refetchedNodeReference.getVector())));
+                    }
+                    return nodeReferencesWithDistanceBuilder.build();
+                })
+                .thenCompose(refetchedNodeReferences -> {
+                    final Queue<NodeReferenceWithDistance> candidates =
+                            // This initial capacity is somewhat arbitrary as m is not necessarily a limit,
+                            // but it gives us a number that is better than the default.
+                            new PriorityQueue<>(getConfig().getM(), NodeReferenceWithDistance.comparator());
+                    candidates.addAll(refetchedNodeReferences);
+                    final Set<Tuple> visited = Sets.newConcurrentHashSet(NodeReference.primaryKeys(refetchedNodeReferences));
+                    final Queue<NodeReferenceWithDistance> nearestNeighbors =
+                            new PriorityQueue<>(efSearch + 1, // prevent reallocation further down
+                                    NodeReferenceWithDistance.reversedComparator());
+                    nearestNeighbors.addAll(refetchedNodeReferences);
 
-            final NodeReferenceWithDistance candidate = candidates.poll();
-            final NodeReferenceWithDistance furthestNeighbor = Objects.requireNonNull(nearestNeighbors.peek());
-
-            if (candidate.getDistance() > furthestNeighbor.getDistance()) {
-                return AsyncUtil.READY_FALSE;
-            }
-
-            return primitives.fetchNodeIfNotCached(storageAdapter, readTransaction, storageTransform, layer, candidate, nodeCache)
-                    .thenApply(candidateNode ->
-                            candidateNode == null
-                            ? ImmutableList.<N>of()
-                            : Iterables.filter(candidateNode.getNeighbors(),
-                                    neighbor -> !visited.contains(Objects.requireNonNull(neighbor).getPrimaryKey())))
-                    .thenCompose(neighborReferences -> primitives.fetchNeighborhoodReferences(storageAdapter, readTransaction,
-                            storageTransform, layer, neighborReferences, nodeCache))
-                    .thenApply(neighborReferences -> {
-                        for (final NodeReferenceWithVector current : neighborReferences) {
-                            visited.add(current.getPrimaryKey());
-                            final double furthestDistance =
-                                    Objects.requireNonNull(nearestNeighbors.peek()).getDistance();
-
-                            final double currentDistance = objectiveFunction.applyAsDouble(current.getVector());
-                            if (currentDistance < furthestDistance || nearestNeighbors.size() < efSearch) {
-                                final NodeReferenceWithDistance currentWithDistance =
-                                        new NodeReferenceWithDistance(current.getPrimaryKey(), current.getVector(),
-                                                currentDistance);
-                                candidates.add(currentWithDistance);
-                                nearestNeighbors.add(currentWithDistance);
-                                if (nearestNeighbors.size() > efSearch) {
-                                    nearestNeighbors.poll();
-                                }
-                            }
+                    return AsyncUtil.whileTrue(() -> {
+                        if (candidates.isEmpty()) {
+                            return AsyncUtil.READY_FALSE;
                         }
-                        return true;
+
+                        final NodeReferenceWithDistance candidate = candidates.poll();
+                        final NodeReferenceWithDistance furthestNeighbor = Objects.requireNonNull(nearestNeighbors.peek());
+
+                        if (candidate.getDistance() > furthestNeighbor.getDistance()) {
+                            return AsyncUtil.READY_FALSE;
+                        }
+
+                        return primitives.fetchNodeIfNotCached(storageAdapter, readTransaction, storageTransform, layer, candidate, nodeCache)
+                                .thenApply(candidateNode ->
+                                        candidateNode == null
+                                        ? ImmutableList.<N>of()
+                                        : Iterables.filter(candidateNode.getNeighbors(),
+                                                neighbor -> !visited.contains(Objects.requireNonNull(neighbor).getPrimaryKey())))
+                                .thenCompose(neighborReferences -> primitives.fetchNeighborhoodReferences(storageAdapter, readTransaction,
+                                        storageTransform, layer, neighborReferences, nodeCache))
+                                .thenApply(neighborReferences -> {
+                                    for (final NodeReferenceWithVector current : neighborReferences) {
+                                        visited.add(current.getPrimaryKey());
+                                        final double furthestDistance =
+                                                Objects.requireNonNull(nearestNeighbors.peek()).getDistance();
+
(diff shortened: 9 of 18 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/StatsDebugger.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>
+ * This interface extends the {@link Debugger} interface and provides functionality related to
+ * recording instances of {@link Debugger.Event} and producing statistics based on these events
+ * in the form of {@link StatsMaps}.
+ * <p>
+ * Similar to the {@link Debugger} interface, the main mean of communication with the debugger
+ * for operations related to recording events is the set of statics defined within this interface.
+ * </p>
+ */
+public interface StatsDebugger extends Debugger {
```
normen662: `StatsKeeper` ?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
...
+        THREAD_LOCAL.get().put(listenerClass, listener);
+    }
+
+    /**
+     * Remove the {@link Listener} instance associated with the provided {@code listenerClass}, if it is set.
+     * @param listenerClass the listener class to remove the listener instance of.
+     */
+    public static void removeListener(@Nonnull final Class<? extends Listener> listenerClass) {
+        THREAD_LOCAL.get().remove(listenerClass);
+    }
```
hazefully: @normen662 I made an attempt in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/2508f05ab6b4c0e70fb320cfca2ca3a531e7ede8 to have only a single thread-local storage with all listeners and make use of that in different classes, but not sure if that makes things better or worse, I guess it depends on how much it is important to avoid multiple thread-local variables.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -342,29 +353,43 @@ private RelationalResultSet executeExplain(@Nonnull ContinuationImpl parsedConti
...
+                                aggregateInsertIntoMemoStats.map(s -> s.getCount(PlannerEvent.Location.NEW)).orElse(0L),
+                                aggregateInsertIntoMemoStats.map(s -> s.getCount(PlannerEvent.Location.REUSED)).orElse(0L),
+                                aggregateTransformRuleCallStats.map(s -> s.getCount(
+                                        PlannerEvent.Location.DISCARDED_INTERSECTION_COMBINATIONS)).orElse(0L),
+                                aggregateTransformRuleCallStats.map(s -> s.getCount(
+                                        PlannerEvent.Location.ALL_INTERSECTION_COMBINATIONS)).orElse(0L),
+                                executingTasksStatsForRewritingPhase.map(s -> s.getCount(PlannerEvent.Location.BEGIN)).orElse(0L),
+                                executingTasksStatsForPlanningPhase.map(s -> s.getCount(PlannerEvent.Location.BEGIN)).orElse(0L),
+                                executingTasksStatsForRewritingPhase.map(PlannerEventStats::getTotalTimeInNs).orElse(0L),
+                                executingTasksStatsForPlanningPhase.map(PlannerEventStats::getTotalTimeInNs).orElse(0L)
```
sergei-pustovykh, line 402: A question is not related to this PR: is it appropriate to construct data and schema separately without any connections? It is very simply to accidentally change the order of these lines and schema won't match but it is very hard to spot. 
I had to spent some time to understand that `13` in this line
`DataType.StructType.Field.from("PLANNING_PHASE_TASKS_TOTAL_TIME_NS", DataType.Primitives.LONG.type(), 13)`
is basically the index of this line here
`executingTasksStatsForPlanningPhase.map(PlannerEventStats::getTotalTimeInNs).orElse(0L)`

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -320,7 +325,13 @@ private RelationalResultSet executeExplain(@Nonnull ContinuationImpl parsedConti
...
                             DataType.StructType.Field.from("INSERT_NEW_COUNT", DataType.Primitives.LONG.type(), 6),
-                            DataType.StructType.Field.from("INSERT_REUSED_COUNT", DataType.Primitives.LONG.type(), 7)),
+                            DataType.StructType.Field.from("INSERT_REUSED_COUNT", DataType.Primitives.LONG.type(), 7),
+                            DataType.StructType.Field.from("TRANSFORM_DISCARDED_INTERSECTION_COMBINATIONS_COUNT", DataType.Primitives.LONG.type(), 8),
+                            DataType.StructType.Field.from("TRANSFORM_ALL_INTERSECTION_COMBINATIONS_COUNT", DataType.Primitives.LONG.type(), 9),
+                            DataType.StructType.Field.from("REWRITING_PHASE_TASK_COUNT", DataType.Primitives.LONG.type(), 10),
+                            DataType.StructType.Field.from("PLANNING_PHASE_TASK_COUNT", DataType.Primitives.LONG.type(), 11),
+                            DataType.StructType.Field.from("REWRITING_PHASE_TASKS_TOTAL_TIME_NS", DataType.Primitives.LONG.type(), 12),
+                            DataType.StructType.Field.from("PLANNING_PHASE_TASKS_TOTAL_TIME_NS", DataType.Primitives.LONG.type(), 13)),
```
sergei-pustovykh, line 344: Just accidentally spotted that we have `planner_metrics.proto` file which partially match this schema. Should we add the missed fields to the proto file?

https://github.com/FoundationDB/fdb-record-layer/blob/99aedf7f6c21522b88c962ddf8d5f06e43e82dfa/yaml-tests/src/main/proto/planner_metrics.proto#L44

normen662: I think Hazem wanted to do that separately as this would cause all the metrics files to go stale.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
+/*
+ * PlannerEventDispatcher.java
```
normen662:
```suggestion
 * PlannerEventListeners.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```
normen662:
```suggestion
 * Copyright 2015-2026 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
...
+        THREAD_LOCAL.get().clear();
+    }
+
+    /**
+     * Get the current active {@link Listener} instance for the provided {@link Class}.
+     * @param listenerClass a class that implements the {@link Listener} interface.
+     * @return the instance that is currently set for the provided {@code listenerClass}, null otherwise.
+     */
+    @Nullable
+    public static Listener getListener(@Nonnull final Class<? extends Listener> listenerClass) {
```
normen662: I think this should use a generic that it can return. As the caller asks for a particular class (extending `Listener`) we can cast and then return an instance of that class.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
...
+    static final ThreadLocal<Map<Class<? extends Listener>, Listener>> THREAD_LOCAL =
+            ThreadLocal.withInitial(HashMap::new);
+
+    /**
+     * Set the active thread-local instance for the provided listener class to the provided instance.
+     * This will override any currently set instances for the same {@link Class}.
+     * @param listenerClass the class that the provided listener is an instance of.
+     * @param listener the new listener instance to use
+     */
+    public static void addListener(@Nonnull final Class<? extends Listener> listenerClass,
```
normen662: This should also be generified (if that's a word). We want to avoid the caller to do a 

```
PlannerEventListeners.addListener(Debugger.class, new OtherListenerClass());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * <p>
+     * Classes implementing this interface listen to {@link PlannerEvent}s emitted by
+     * {@link com.apple.foundationdb.record.query.plan.cascades.CascadesPlanner} and do some action on them.
+     * </p>
+     * <p>
+     * See {@link com.apple.foundationdb.record.query.plan.cascades.debug.Debugger} and {@link PlannerEventStatsCollector}.
+     * </p>
+     */
+    public interface Listener {
```
normen662: Let's pull this up to top level and name it `EventListener`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
```diff
@@ -0,0 +1,164 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.events.eventprotos.PPartialMatch;
+import com.apple.foundationdb.record.query.plan.cascades.events.eventprotos.PRegisteredReference;
+import com.apple.foundationdb.record.query.plan.cascades.events.eventprotos.PRegisteredRelationalExpression;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
+import com.google.protobuf.Message;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Tag interface for all events.
```
normen662:
```suggestion
 * Interface for all events.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
```diff
@@ -0,0 +1,164 @@
...
+        } else if (bindable instanceof PartialMatch) {
+            builder.setPartialMatch(toPartialMatchProto((PartialMatch)bindable));
+        } else if (bindable instanceof MatchPartition) {
+            builder.setMatchPartition(toMatchPartitionProto((MatchPartition)bindable));
+        }
+        return builder.build();
+    }
+
+    @Nonnull
+    static PPartialMatch toPartialMatchProto(@Nonnull final PartialMatch partialMatch) {
```
normen662: This should be in `PartialMatch` and not here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
```diff
@@ -0,0 +1,164 @@
...
+        return PPartialMatch.newBuilder()
+                .setMatchCandidate(partialMatch.toString())
+                .setQueryRef(toReferenceProto(partialMatch.getQueryRef()))
+                .setQueryExpression(toExpressionProto(partialMatch.getQueryExpression()))
+                .setCandidateRef(toReferenceProto(partialMatch.getCandidateRef()))
+                .build();
+    }
+
+    @Nonnull
+    static PMatchPartition toMatchPartitionProto(@Nonnull final MatchPartition matchPartition) {
```
normen662: This should be in `MatchPartition` and not here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
```diff
@@ -0,0 +1,164 @@
...
+    @Nonnull
+    static PRegisteredRelationalExpression toExpressionProto(@Nonnull final RelationalExpression expression) {
+        return PRegisteredRelationalExpression.newBuilder()
+                .setName(Debugger.mapDebugger(debugger -> debugger.nameForObject(expression)).orElseThrow())
+                .setSemanticHashCode(expression.semanticHashCode())
+                .build();
+    }
+
+    @Nonnull
+    static PRegisteredReference toReferenceProto(@Nonnull final Reference reference) {
```
normen662: This should be in `Reference` and it the class it is creating should be renamed `PReference` (as per convention).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
```diff
@@ -0,0 +1,164 @@
...
+                .setDescription(getDescription())
+                .setShorthand(getShorthand().name())
+                .build();
+    }
+
+    @Nonnull
+    PPlannerEvent.Builder toEventBuilder();
+
+    @Nonnull
+    static PRegisteredRelationalExpression toExpressionProto(@Nonnull final RelationalExpression expression) {
```
normen662: This should be in `RelationalExpression` and the name of the class it is returning should be `PExpression` as per convention.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/package-info.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * package-info.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```
normen662:
```suggestion
 * Copyright 2015-2026 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsMaps.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * This class stores statistics for {@link PlannerEvent}s emitted during the planning of a query by the
+ * {@link com.apple.foundationdb.record.query.plan.cascades.CascadesPlanner}, and provides
+ * different ways to access these statistics: by planner event type (i.e. one of {@link PlannerEvent} implementations),
+ * {@link PlannerPhase} and {@link com.apple.foundationdb.record.query.plan.cascades.PlannerRule}.
+ */
+public class PlannerEventStatsMaps {
```
normen662, line 40: This class is de-facto immutable now and is directly created by `PlannerEventStatsCollectorState.getStatsMaps()`. Unfortunately, there are no comments here that explain the reasoning why that was done. We want to make sure that, whoever consumes this structure can be certain that the counters will not change afterwards.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsMaps.java
```diff
@@ -0,0 +1,167 @@
...
+    private final Supplier<Map<PlannerPhase, Map<Class<? extends PlannerEventWithState>, PlannerEventStats>>> immutableEventWithStateClassStatsByPlannerPhaseMapSupplier;
+    @Nonnull
+    private final Supplier<Map<Class<? extends CascadesRule<?>>, PlannerEventStats>> immutablePlannerRuleClassStatsMapSupplier;
+
+
+    public PlannerEventStatsMaps(@Nonnull final Map<Class<? extends PlannerEvent>, ? extends PlannerEventStats> eventWithoutStateClassStatsMap,
+                                 @Nonnull final Map<PlannerPhase, Map<Class<? extends PlannerEventWithState>, ? extends PlannerEventStats>> eventWithStateClassStatsByPlannerPhaseMap,
+                                 @Nonnull final Map<Class<? extends CascadesRule<?>>, ? extends PlannerEventStats> plannerRuleClassStatsMap) {
+
+        this.eventWithoutStateClassStatsMap = eventWithoutStateClassStatsMap;
```
normen662: I think this is done wrong. The (only) caller to this just creates unmodifiable maps around its own mutable maps.

1. The constructor here should deal with this maps and make that step towards immutability.
2. At least one map uses a nested map for its values which still modifiable.
3. We don't want to see if the caller potentially modifies the values later.
4. We normally always use `ImmutableMap`, `ImmutableList`, `ImmutableSet`, we never create an unmodifiable view.

The downside of doing this more rigorously here is that the call the `getStatsMaps()` gets more expensive. That should be ok though, as it should not be set for every query anyway.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsCollectorState.java
```diff
@@ -0,0 +1,203 @@
...
+                .computeIfAbsent(event.getClass(), (eC) -> new MutableStats());
+    }
+
+    private MutableStats getEventStatsForPlannerRuleClass(@Nonnull Class<? extends CascadesRule<?>> plannerRuleClass) {
+        return plannerRuleClassStatsMap.computeIfAbsent(plannerRuleClass, (eC) -> new MutableStats());
+    }
+
+    @Nonnull
+    PlannerEventStatsMaps getStatsMaps() {
+        return new PlannerEventStatsMaps(
```
normen662, line 141: see biggish comment in `PlannerEventStatsMaps`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -111,6 +111,9 @@ public interface Type extends Narrowable<Type>, PlanSerializable {
     @Nonnull
     Function FUNCTION = new Function();
 
+    @Nonnull
```
normen662: Not sure how this change got in here. Can you please undo?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
```diff
@@ -370,6 +373,11 @@ static None noneType() {
         return Type.NONE;
     }
 
+    @Nonnull
```
normen662: same with this one.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
```diff
@@ -437,16 +450,16 @@ private void planPartial(@Nonnull final Supplier<Reference> referenceSupplier,
                 }
                 taskCount++;
 
-                Debugger.withDebugger(debugger -> debugger.onEvent(
-                        new Debugger.ExecutingTaskEvent(currentRoot, taskStack, Location.BEGIN,
-                                Objects.requireNonNull(taskStack.peek()))));
+                PlannerEventListeners.dispatchEvent(
```
normen662: Let's import `PlannerEvent.Location`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -534,15 +534,14 @@ private Reference memoizeExpressionsExactly(@Nonnull final Collection<? extends
                                                 @Nonnull BiFunction<Set<? extends RelationalExpression>, Set<? extends RelationalExpression>, Reference> referenceCreator) {
         final var allExpressions =
                 Iterables.concat(exploratoryExpressions, finalExpressions);
-        Debugger.withDebugger(debugger -> allExpressions.forEach(
-                expression -> debugger.onEvent(InsertIntoMemoEvent.begin())));
+        allExpressions.forEach(expression -> PlannerEventListeners.dispatchEvent(InsertIntoMemoPlannerEvent.begin()));
```
normen662: Hmmm, this causes an iteration through `allExpressions` even if there is no event listener installed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
```diff
@@ -534,15 +534,14 @@ private Reference memoizeExpressionsExactly(@Nonnull final Collection<? extends
...
         try {
             final var exploratoryExpressionSet = new LinkedIdentitySet<>(exploratoryExpressions);
             final var finalExpressionSet = new LinkedIdentitySet<>(finalExpressions);
             return addNewReference(referenceCreator.apply(exploratoryExpressionSet, finalExpressionSet));
         } finally {
-            Debugger.withDebugger(debugger -> allExpressions.forEach(
-                    expression -> debugger.onEvent(InsertIntoMemoEvent.end())));
+            allExpressions.forEach(
```
normen662: same here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -326,23 +327,23 @@ public boolean insertFinalExpression(@Nonnull final RelationalExpression newValu
...
-            Debugger.withDebugger(debugger -> {
-                if (containsInMemo) {
-                    debugger.onEvent(InsertIntoMemoEvent.reusedExpWithReferences(newExpression, ImmutableList.of(this)));
-                } else {
-                    debugger.onEvent(InsertIntoMemoEvent.newExp(newExpression));
-                }
-            });
+
+            if (containsInMemo) {
+                PlannerEventListeners.dispatchEvent(InsertIntoMemoPlannerEvent.reusedExpWithReferences(newExpression, ImmutableList.of(this)));
```
normen662: I think the API here has regressed (as pointed out in the other file). This call creates a new event even if there is no listener installed. It previously was shielded by the `Debugger.withDebugger(...)` call.

normen662: I won't continue to point this out. I think we should come up with the proper way of calling this method and unfortunately you would need to look at every caller and change it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/QueryPlanInfoKeys.java
```diff
@@ -32,7 +32,7 @@ public class QueryPlanInfoKeys {
             new QueryPlanInfo.QueryPlanInfoKey<>("maxTaskQueueSize");
     public static final QueryPlanInfo.QueryPlanInfoKey<QueryPlanConstraint> CONSTRAINTS =
             new QueryPlanInfo.QueryPlanInfoKey<>("constraints");
-    public static final QueryPlanInfo.QueryPlanInfoKey<StatsMaps> STATS_MAPS =
+    public static final QueryPlanInfo.QueryPlanInfoKey<PlannerEventStatsMaps> STATS_MAPS =
```
normen662, line 35: Careful here. This has effects on downstream projects!!!

fdb-record-layer-debugger/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/PlannerRepl.java
```diff
@@ -453,25 +458,6 @@ public void onDone() {
         reset();
     }
 
-    @Override
```
normen662, line 456: Where did these go?

fdb-record-layer-debugger/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/PlannerRepl.java
```diff
@@ -995,8 +981,8 @@ public boolean onCallback(final PlannerRepl plannerRepl, final Event event) {
         public void onList(final PlannerRepl plannerRepl) {
             super.onList(plannerRepl);
             plannerRepl.print("; ");
-            plannerRepl.printKeyValue("shorthand", Shorthand.RULECALL + "; ");
-            plannerRepl.printKeyValue("location", Location.END.name().toLowerCase(Locale.ROOT) + "; ");
+            plannerRepl.printKeyValue("shorthand", PlannerEvent.Shorthand.RULECALL.name().toLowerCase(Locale.ROOT) + "; ");
```
normen662: let's import `PlannerEvent.Shorthand`

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/ExplainTests.java
```diff
@@ -114,6 +139,56 @@ void explainWithNoContinuationTest() throws Exception {
...
+                            .hasValue("PLANNING_PHASE_TASK_COUNT", 204L);
+                    assertResult.hasNoNextRow();
+                }
+            }
+        }
+    }
+
+    @Test
+    void explainContainsEventStatsWithDefaultEventStatsCollectorEnabled() throws Exception {
+        Debugger.setDebugger(null);
```
normen662: Can we put this into a

```
try {
...
} finally {
...
}
```

We must leave this method using the same debugger/eventlistener installed as when this method was entered.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/DebuggerWithSymbolTables.java
```diff
@@ -323,28 +320,10 @@ protected PEvent computeNext() {
         };
     }
 
-    @Override
```
normen662, line 326: Where did these two methods go?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/DebuggerWithSymbolTables.java
```diff
@@ -396,24 +370,6 @@ public static DebuggerWithSymbolTables withPrerecordedEvents(@Nonnull final Stri
         return new DebuggerWithSymbolTables(true, true, fileName);
     }
 
-    public static void printForEachExpression(@Nonnull final Reference root) {
```
normen662, line 399: Where did these two go?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/State.java
```diff
@@ -21,40 +21,28 @@
...
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.TimeUnit;
 import java.util.function.IntUnaryOperator;
 
 @SuppressWarnings("PMD.SystemPrintln")
 public class State {
```
normen662: I think the name of this class should be changed to something more appropriate. Maybe `RegisteredEntities`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/StatsDebugger.java
normen662: This class was deleted in the meantime.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
sergei-pustovykh: I update the year in all changed source files https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/61c533f6ed5ef05f14940044781be5e5229a9ace

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
sergei-pustovykh: renamed in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/1a1cd8d4b5f664781df812ab85964b3f0f00c167

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/a5c221b4e287d837a53e7735b72dc5353a1e4fec

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/a5c221b4e287d837a53e7735b72dc5353a1e4fec

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
sergei-pustovykh: We can't simply generify it because the original type of the debugger does not match Debugger.class, it just implements it. (see `Debugger.setDebugger` and `Debugger.getDebugger`)
We probably can implement `Debugger.setDebugger` as a generic too, but I don't see a way how `Debugger.getDebugger` should work in this case.

We can leave it as is, or do some redesign and add to `EventListener` interface additional static method `getStorageKey()` and use it like `THREAD_LOCAL.get().put(listener. getStorageKey(), listener);` 
All `EventListener`s have to override and return unique key, but for debuggers it has to be overridden in the `Debugger` interface and reports the same key for all debuggers. This solution looks awkward, maybe something more elegant could be found...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/3107b387709f43023ebd94b43da05a26fbff6f9e

please check the methods names, Im not sure about right naming here. 
```
PartialMatch.toPlannerEventPartialMatchProto
MatchPartition.toPlannerEventMatchPartitionProto
Reference.toPlannerEventReferenceProto
RelationalExpression.toPlannerEventExpressionProto
...
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/3107b387709f43023ebd94b43da05a26fbff6f9e

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/commit/3107b387709f43023ebd94b43da05a26fbff6f9e

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEvent.java
sergei-pustovykh: done in https://github.com/FoundationDB/fdb-record-layer/commit/3107b387709f43023ebd94b43da05a26fbff6f9e

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/package-info.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/commit/61c533f6ed5ef05f14940044781be5e5229a9ace

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsMaps.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/617840b923f48c08edf674a5d27e43751999bf06

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsMaps.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/617840b923f48c08edf674a5d27e43751999bf06
redo https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/78617b20617c1e087c6660f3ff5394974f084976

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventStatsCollectorState.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/617840b923f48c08edf674a5d27e43751999bf06

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/a8f1ed0e65d34f80e8887dd18823551dc1fff886

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/Type.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/a8f1ed0e65d34f80e8887dd18823551dc1fff886

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/7099ba5f372ac237a360acd5bea89a52d6b13a6b

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/1ed33941269ff95a2befb2fb9745dfdcf63d34bc

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesRuleCall.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/1ed33941269ff95a2befb2fb9745dfdcf63d34bc

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
sergei-pustovykh: I implemented `dispatchEvent()` with supplier in parameter, which is called only if any Listener exists.  https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/1ed33941269ff95a2befb2fb9745dfdcf63d34bc

fdb-record-layer-debugger/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/PlannerRepl.java
sergei-pustovykh: Debugger does not store stats now
instead of `Debugger.showStats()` we have `PlannerEventStatsViewer.showStats()`
instead of `Debugger.getStatsMaps()` we have `PlannerEventStatsCollector.getStatsMaps()`
Should I leave some comments about that?

fdb-record-layer-debugger/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/PlannerRepl.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/55dbb1f05f436963eed1aa3411ae0fc12820d782

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/DebuggerWithSymbolTables.java
sergei-pustovykh: Debugger does not store stats any more.
`PlannerEventStatsViewer.showStats()`
`PlannerEventStatsCollector.getStatsMaps()`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/DebuggerWithSymbolTables.java
sergei-pustovykh: returned https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/540d52fad712cfc42306446f91927bb734a9002c

fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/ExplainTests.java
sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/78934e0289e459e45e2f4c053816d4e0f29896b0

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/State.java
sergei-pustovykh: renamed https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/89966c7d97a47a2c65d6340c5b116c5699ac2c6c

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
normen662:
```
/**
     * Set the active thread-local instance for the provided listener class to the provided instance.
     * This will override any currently set instances for the same {@link Class}.
     * @param listenerClass the class that the provided listener is an instance of.
     * @param listener the new listener instance to use
     */
    public static <T extends EventListener, T1 extends T> void addListener(@Nonnull final Class<T> listenerClass,
                                                                           @Nonnull final T1 listener) {
        THREAD_LOCAL.get().put(listenerClass, listener);
    }
```

fdb-record-layer-debugger/src/main/java/com/apple/foundationdb/record/query/plan/cascades/debug/PlannerRepl.java
normen662: No -- my bad. All good -- that makes sense.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/events/PlannerEventListeners.java
sergei-pustovykh: surprisingly it does not compile for such cases:
```
        final var t = new DummyEventListener();
        PlannerEventListeners.addListener(t.getClass(), t);
```
`T1 extends T` - should be determined during compilation, but before execution the result of `t.getClass()` is unknown. it is known only that it is `extends DummyEventListener` not precise `DummyEventListener`

looking for other solution

sergei-pustovykh: done https://github.com/FoundationDB/fdb-record-layer/pull/3512/commits/cc2313bfa1c6b64e0ba1560dd117afdd94abea80

PR #3515, opened 2025-07-30
Interpret full-supported UUID type correctly in JDBC server 

g31pranjal:
This PR (1) adds the missing support in `DataType` and gRPC-related classes for the future UUID type. This is needed to maintain compatibility in mixed-mode testing with the future full-support version.

It is relatively difficult to test this on the `main` as we cannot bring in the support in the current PR. However, here is the PR for re-introducing the full support: https://github.com/FoundationDB/fdb-record-layer/pull/3510 (2). I have tested that (2) is backward compatible with fixes in (1). Hence, if (1) releases first, and then we bring in (2) we will maintain compatibility to "older" version JDBC servers.


alecgrieser (approved):
LGTM. In normal circumstances, we'd probably want some more tests, but given the difficulty here, and that this will be tested once we can add UUID support back in the next minor release, I'd say this is fine

PR #3516, opened 2025-07-30
HNSW poc

normen662:


fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/HNSWHelpers.java
```diff
@@ -0,0 +1,63 @@
...
+        // nothing
+    }
+
+    /**
+     * Helper method to format bytes as hex strings for logging and debugging.
+     * @param bytes an array of bytes
+     * @return a {@link String} containing the hexadecimal representation of the byte array passed in
+     */
+    @Nonnull
+    static String bytesToHex(byte[] bytes) {
```
MMcM: Can this be shared with the `NodeHelpers` one?

fdb-extensions/src/main/java/com/apple/foundationdb/async/rtree/NodeHelpers.java
```diff
@@ -1,5 +1,5 @@
 /*
- * NodeHelpers.java
+ * HNSWHelpers.java
```
MMcM, line 2: Looks like an accident.

fdb-record-layer-core/src/test/proto/evolution/test_field_type_change.proto
```diff
@@ -24,7 +24,7 @@ package com.apple.foundationdb.record.evolution.fieldtypechange;
 option java_package = "com.apple.foundationdb.record.evolution";
 option java_outer_classname = "TestFieldTypeChangeProto";
 
-import "record_metadata_options.proto";
+//import "record_metadata_options.proto";
```
MMcM, line 27: Why? What was broken?

MMcM (changes requested):

fdb-record-layer-core/src/test/proto/evolution/test_field_type_change.proto
normen662: Not sure what happened. I changed something somewhere in a test proto file and suddenly got all these warnings about unnecessary imports. And since we run everything in `-Werror`...

normen662:
This has found its way into main in the meantime.

PR #3522, opened 2025-07-31
Extend serialized record encoding to allow for multiple encryption keys

MMcM:
A key number is encoded in the serialized record, in a format that is backward compatible. An interface can be provided to resolve key numbers into keys for encryption and decryption. The byte format of existing encrypted data is such that it is interpreted as having key number 0.

Compatibility of this change in persistent data is checked by the `encrypted.yamsql` test when run against `4.6.1.0`. The YAML framework still uses the old serializer API to set encryption keys, so this also demonstrates that this is compatible.

A followup PR will change YAML to use the new serializer API while continuing to pass mixed tests.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
@@ -34,7 +34,6 @@ import com.apple.test.ParameterizedTestUtils;
 import com.apple.test.RandomSeedSource;
 import com.apple.test.RandomizedTestUtils;
 import com.google.common.base.Strings;
-import com.google.common.collect.Streams;
 import com.google.common.primitives.Bytes;
 import com.google.protobuf.Message;
 import org.junit.jupiter.api.BeforeAll;
@@ -508,12 +507,11 @@ public class TransformedRecordSerializerTest {
         assertThat(e.getMessage(), containsString("unrecognized transformation encoding"));
     }
 
-    @SuppressWarnings("UnstableApiUsage")
     public static Stream<Arguments> randomAndCompressed() {
-        return Streams.zip(
+        return ParameterizedTestUtils.cartesianProduct(
                 RandomizedTestUtils.randomSeeds(0xC0DE6EEDL, 0x6EEDC0DEL),
-                Stream.of(Boolean.TRUE, Boolean.FALSE),
-                Arguments::arguments);
+                ParameterizedTestUtils.booleans("compressToo")
+        );
     }
 
     @ParameterizedTest


MMcM:
These changes can be tested for compatibility after #3557, which will include a test that passes serializer options to an old external version, which is then able to interoperate with the same options passed through the new manager API.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/SerializationKeyManager.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * TransformedRecordSerializerKeyManager.java
```
ScottDugas:
```suggestion
 * SerializationKeyManager.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/SerializationKeyManager.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2018 Apple Inc. and the FoundationDB project authors
```
ScottDugas:
```suggestion
 * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerPrefix.java
```diff
@@ -0,0 +1,188 @@
+/*
+ * TransformedRecordSerializerEncoding.java
```
ScottDugas:
```suggestion
 * TransformedRecordSerializerPrefix.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerPrefix.java
```diff
@@ -0,0 +1,188 @@
...
+ * The encoded form begins with a Protobuf <a href="https://protobuf.dev/programming-guides/encoding/#varints">varint</a>.
+ * The low three bits of this prefix specify how it was encoded and the remaining bits are the encryption key number.
+ * <ul>
+ * <li><code>1</code> encrypted</li>
+ * <li><code>2</code> clear text or no prefix at all</li>
+ * <li><code>4</code> compressed</li>
+ * <li><code>5</code> compressed and encrypted (in that order)</li>
+ * </ul>
+ * Recall that a serialized record is a message using the {@link RecordMetaData#getUnionDescriptor union descriptor}
+ * from the {@link RecordMetaData record metadata}. That means is will be wire-type <code>2</code> plus a field numnber
```
ScottDugas:
```suggestion
 * from the {@link RecordMetaData record metadata}. That means this will be wire-type <code>2</code> plus a field number
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerState.java
```diff
@@ -0,0 +1,120 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+
+/**
+ * The internal state of serialization / deserialization, pointing to a portion of a byte array.
+ * Also includes information on intended / found serialization format.
+ */
+@SpotBugsSuppressWarnings("EI_EXPOSE_REP")
+class TransformedRecordSerializerState {
```
ScottDugas, line 35: Should have an internal api annotation

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerPrefix.java
```diff
@@ -0,0 +1,188 @@
...
+ * serialized message. Finally, a key number of zero becomes a single byte prefix of <code>1</code>, <code>2</code>,
+ * <code>4</code> or <code>5</code>, formerly representing an encrypted and compressed bitmask.
+ * </p>
+ *
+ * <p>
+ * The encrypted form currently begins with a random IV, although this might change for other algorithms.
+ * The compressed form begins with a compression level, which again might someday be extended.
+ * </p>
+ */
+@API(API.Status.UNSTABLE)
```
ScottDugas: Should this be internal instead?
```suggestion
@API(API.Status.INTERNAL)
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
         }
     }
 
+    @ParameterizedTest
+    @ValueSource(ints = {6, 10})
+    public void malformedVarintEncoding(int length) {
```
ScottDugas: Tests should have package visibility. I don't think you necessarily should change the existing tests but the new ones would be good to have.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -359,24 +367,29 @@ public void unrecognizedEncoding() {
...
         assertThat(e.getMessage(), containsString("unrecognized transformation encoding"));
-        assertEquals(15, e.getLogInfo().get("encoding"));
+        assertEquals(15L, e.getLogInfo().get("encoding"));
     }
 
-    @Test
-    public void encryptWhenSerializing() throws Exception {
+    @ParameterizedTest
+    @BooleanSource
+    public void encryptWhenSerializing(boolean compressToo) throws Exception {
```
ScottDugas: It seems like it would be good to have a `@ParameterizedTest` that has serialized bytes created on `main` for each of the permutations for allowed records-on-disk. 
Of note: https://github.com/FoundationDB/fdb-record-layer/pull/3557/files covers encryption, but does not cover compression, or the two together.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -359,24 +367,29 @@ public void unrecognizedEncoding() {
             deserialize(serializer, Tuple.from(1066L), serialized);
         });
         assertThat(e.getMessage(), containsString("unrecognized transformation encoding"));
-        assertEquals(15, e.getLogInfo().get("encoding"));
+        assertEquals(15L, e.getLogInfo().get("encoding"));
     }
 
-    @Test
-    public void encryptWhenSerializing() throws Exception {
+    @ParameterizedTest
```
ScottDugas, line 375: Would it be valuable to have some unit tests of the `readVarint`/`writeVarint` section? I'm not as good reasoning about all the bitwise operators. I believe we also want it to be the same as protobuf, so maybe something that writes with protobuf and reads with `TransformedRecordSerializerPrefix` and vice-versa would be valuable.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
...
+            byte[] serialized = new byte[10];
+            TransformedRecordSerializerPrefix.writeVarint(serialized,
+                    TransformedRecordSerializerPrefix.PREFIX_ENCRYPTED + ((long)Integer.MAX_VALUE + 1 << 3));
+            deserialize(serializer, Tuple.from(1066L), serialized);
+        });
+        assertThat(e.getMessage(), containsString("unrecognized transformation encoding"));
+    }
+
+    @Test
+    public void encryptRollingKeys() throws Exception {
```
ScottDugas: `RollingTestKeyManager` picks random integers for the keys, would it make sense to use `RandomizedTestUtils` to seed this in a predictable way, in case a specific integer causes problems with the serialization.

ScottDugas: Perhaps worth parameterizing this for whether it is combined with compression too

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
...
+        RollingTestKeyManager keyManager = new RollingTestKeyManager();
+        TransformedRecordSerializer<Message> serializer = TransformedRecordSerializerJCE.newDefaultBuilder()
+                .setEncryptWhenSerializing(true)
+                .setKeyManager(keyManager)
+                .setWriteValidationRatio(1.0)
+                .build();
+
+        List<MySimpleRecord> records = new ArrayList<>();
+        for (int i = 0; i < 10; i++) {
```
ScottDugas: This test is quite fast, perhaps worth running for 100 records instead to get better coverage of integers.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerJCE.java
```diff
@@ -282,17 +304,18 @@ public Builder<M> clearSecureRandom() {
...
-                if (cipherName == null) {
-                    cipherName = CipherPool.DEFAULT_CIPHER;
+            } else {
+                if (encryptionKey != null) {
+                    throw new RecordCoreArgumentException("cannot specify both key manager and encryption key");
                 }
-                if (secureRandom == null) {
-                    secureRandom = new SecureRandom();
+                if (cipherName != null) {
+                    throw new RecordCoreArgumentException("cannot specify both key manager and cipher name");
```
ScottDugas, line 318: Should this error if `secureRandom` is specified?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
...
+    public void cannotDecryptWithoutKey(boolean jce) throws Exception {
+        KeyGenerator keyGen = KeyGenerator.getInstance("AES");
+        keyGen.init(128);
+        TransformedRecordSerializer<Message> serializer = TransformedRecordSerializerJCE.newDefaultBuilder()
+                .setEncryptWhenSerializing(true)
+                .setEncryptionKey(keyGen.generateKey())
+                .setWriteValidationRatio(1.0)
+                .build();
+        MySimpleRecord simpleRecord = MySimpleRecord.newBuilder().setRecNo(1066L).setStrValueIndexed("Hello").build();
+        RecordTypeUnion unionRecord = RecordTypeUnion.newBuilder().setMySimpleRecord(simpleRecord).build();
```
ScottDugas: Did you intend to serialize `unionRecord` instead?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
...
+                    @Override
+                    public Random getRandom(final int keyNumber) {
+                        return random;
+                    }
+                })
+                .setWriteValidationRatio(1.0)
+                .build();
+
+        MySimpleRecord simpleRecord = MySimpleRecord.newBuilder().setRecNo(1066L).setStrValueIndexed("Hello").build();
+        RecordTypeUnion unionRecord = RecordTypeUnion.newBuilder().setMySimpleRecord(simpleRecord).build();
```
ScottDugas: Did you intend to serialize `unionRecord` instead?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -450,10 +463,294 @@ public void corruptAnyBit() {
...
+        KeyGenerator keyGen2 = KeyGenerator.getInstance("DES");
+        keyGen2.init(56);
+        TransformedRecordSerializer<Message> deserializer = TransformedRecordSerializerJCE.newDefaultBuilder()
+                .setEncryptWhenSerializing(true)
+                 .setCipherName("DES")
+                .setEncryptionKey(keyGen2.generateKey())
+                .setWriteValidationRatio(1.0)
+                .build();
+        RecordSerializationException e = assertThrows(RecordSerializationException.class,
+                () -> deserialize(deserializer, Tuple.from(1066L), serialized));
```
ScottDugas: perhaps worth extracting a constant `PRIMARY_KEY = Tuple.from(1066L)`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerState.java
MMcM: Added.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: I let the IDE fix all of the warnings in this file, including this, `assertThrows` number of statements, `record` variable name. I made it a separate commit to facilitate review.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: Did you mean as literal byte arrays in the test itself? Or via some other means?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerJCE.java
MMcM: Yes. Added that check, too.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: No. The `unionRecord` was for checking (uncompressed) size against, and doesn't apply once compression / encryption is in play. I removed it from such tests.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: See above.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: Included in IDE-managed changes.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: I made it use a seeded random from that utils class. I think this is what you had in mind. That does not behave quite as I had expected, in that the specified (or defaulted) seeds are fixed, so that the randomness is the same every time, not just reported in the JUnit test name.

MMcM: Yes. It's a little complicated because it also needs the random seed.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: Increased.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
MMcM: I added such a test. I made it use `@RandomSeedSource`, but perhaps I am not using it correctly because it seems to be the opposite of random, using the same fixed seed every time and so the same test varints.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
```diff
@@ -484,17 +486,27 @@ void invalidKeyNumberEncoding() {
...
 
-    @Test
-    void encryptRollingKeys() throws Exception {
-        RollingTestKeyManager keyManager = new RollingTestKeyManager();
+    @SuppressWarnings("UnstableApiUsage")
+    public static Stream<Arguments> randomAndCompressed() {
+        return Streams.zip(
+                RandomizedTestUtils.randomSeeds(0xC0DE6EEDL, 0x6EEDC0DEL),
+                Stream.of(Boolean.TRUE, Boolean.FALSE),
+                Arguments::arguments);
```
ScottDugas: This will break in nightly builds, because those run with non-0 iterations. 
I think I would use:
```
return ParameterizedTestUtils.cartesianProduct(RandomizedTestUtils.randomSeeds(), ParameterizedTestUtils.booleans("compressToo"));
```

In PRB, and when running locally, it will run with:
- `0xf17ed5eedL`, true
- `0xf17ed5eedL`, false

But in nigthly it will add two extra iterations per `compressToo` value (if I remember correctly).

I see now that `RandomizedTestUtils` does not explain this (or at least not well), I created a PR to hopefully make that clearer: https://github.com/FoundationDB/fdb-record-layer/pull/3617

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/common/TransformedRecordSerializerTest.java
ScottDugas: Yeah, that was what I was thinking. Although, perhaps it's better to expand on the work in https://github.com/FoundationDB/fdb-record-layer/pull/3557/files#diff-ae5f04f5574f523120e88b8fe6f5560944690f591bb4046b5a2fbb40fdbdc416R2 to add coverage of compression, and compression+encryption.

ScottDugas (changes requested):

MMcM:
> perhaps it's better to expand on the work

I guess that is more consistent for the testing. It's not really useful to turn off compression, as it already checks whether it helped. In fact, all the records in the existing test aren't compressed, even though compression is allowed, because they are so small.

I created #3618 to add an option for compression and test it. By rights, that should be merged and a new build created against which this PR (after rebasing) can be tested. The draft #3606 represents the final state with all changes.

MMcM:
Rebased against `4.6.3.0` and `YamlIntegrationTests.serializationOptions` passes, in `4.6.3.0 then In-process`, in particular, which I believe shows that the compatibility API is holding up at the persistence level.

PR #3531, opened 2025-08-05
Handle metrics for the same query but with different setups

ScottDugas:
Planner metrics previously assumed that it was only reasonable to have a query once within a single block. Having setup can make this more complicated, so this changes the keys for the metrics to have the setup too.

There were a couple unrelated cleanups along the way:
1. `markDirty()` was annotated `@Nullable`, but is `void`
2. `putMetrics` took `isDirectyMetrics`, but that parameter was never used, so I removed it.

### DIFF
--- a/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
+++ b/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
@@ -47,7 +47,6 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -107,9 +106,7 @@ public final class YamlExecutionContext {
         this.additionalOptions = additionalOptions;
         this.connectionOptions = Options.none();
         this.expectedMetricsMap = loadMetricsResource(resourcePath);
-        this.actualMetricsMap = new TreeMap<>(Comparator.comparing(QueryAndLocation::getLineNumber)
-                .thenComparing(QueryAndLocation::getBlockName)
-                .thenComparing(QueryAndLocation::getQuery));
+        this.actualMetricsMap = new TreeMap<>();
         if (isNightly()) {
             logger.info("ℹ️ Running in the NIGHTLY context.");
             if (shouldCorrectExplains() || shouldCorrectMetrics()) {
@@ -179,11 +176,10 @@ public final class YamlExecutionContext {
                                                             @Nonnull final String query,
                                                             final int lineNumber,
                                                             @Nonnull final PlannerMetricsProto.Info info,
-                                                            boolean isDirtyMetrics) {
-        return actualMetricsMap.put(new QueryAndLocation(blockName, query, lineNumber), info);
+                                                            @Nonnull final List<String> setups) {
+        return actualMetricsMap.put(new QueryAndLocation(blockName, query, lineNumber, setups), info);
     }
 
-    @Nullable
     @SuppressWarnings("UnusedReturnValue")
     public synchronized void markDirty() {
         this.isDirtyMetrics = true;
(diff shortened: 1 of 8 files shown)


yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
```diff
@@ -107,9 +106,7 @@ public static class YamlExecutionError extends RuntimeException {
         this.additionalOptions = additionalOptions;
         this.connectionOptions = Options.none();
         this.expectedMetricsMap = loadMetricsResource(resourcePath);
-        this.actualMetricsMap = new TreeMap<>(Comparator.comparing(QueryAndLocation::getLineNumber)
-                .thenComparing(QueryAndLocation::getBlockName)
-                .thenComparing(QueryAndLocation::getQuery));
+        this.actualMetricsMap = new TreeMap<>();
```
normen662: I am more in the camp of not making things `Comparable<T>` just because it is fixed at that point. I am not opposed to it, though.

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
```diff
@@ -179,11 +176,10 @@ public synchronized PlannerMetricsProto.Info putMetrics(@Nonnull final String bl
                                                             @Nonnull final String query,
                                                             final int lineNumber,
                                                             @Nonnull final PlannerMetricsProto.Info info,
-                                                            boolean isDirtyMetrics) {
-        return actualMetricsMap.put(new QueryAndLocation(blockName, query, lineNumber), info);
+                                                            @Nonnull final List<String> setups) {
```
normen662, line 182: Is this sensitive to the order of the setup strings?

yaml-tests/src/main/proto/planner_metrics.proto
```diff
@@ -32,6 +32,7 @@ message Entry {
 message Identifier {
   optional string blockName = 1;
   optional string query = 2;
+  repeated string setups = 3;
```
normen662, line 35: Do you output this in yaml? If not, won't you end up with duplicate yaml sections?

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
ScottDugas: Yes, (or at least it should be) but so are the semantics, in theory. Right now the setups only allow creating temporary functions, so the only way it can really matter with both being valid is if you try to replace the existing temporary function. Which might make sense.
i.e.:
```
- query: SELECT * FROM T2
- setup: create temporary function t2() on commit drop function AS
               SELECT * FROM t1 WHERE id < 10;
- setup: create temporary function t2() on commit drop function AS
               SELECT * FROM t1 WHERE id > 10;
```

ANd then you could have another version where the two setups are reversed.
That test is really more about whether `create temporary function` replaces correctly (I think it does), so the metrics are probably unimportant, and an explain check might be excessive. But I think it's better to support it, than not.

yaml-tests/src/main/proto/planner_metrics.proto
ScottDugas: Yes, it does in `saveMetricsAsYaml` (I can't seem to get a link to the change...)
And you can see an example here: https://github.com/FoundationDB/fdb-record-layer/pull/3531/files#diff-6c04e317b2477f04a1eb883fc8c9c7464e8d1a930979f59ed6eb98ede8aaa5bb

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlExecutionContext.java
ScottDugas: This seemed like the natural ordering of the class to me, based on its definition.

ScottDugas: That being said, I think I only added this because I was intending to fully encapsulate the block name and query with the addition of the setups, but ended up not doing that so, it's not as valuable, so I changed it back.

normen662 (approved):

PR #3536, opened 2025-08-08
Add more prefix-based tests to like.yamsql

alecgrieser:
This adds an additional set of tests to `like.yamsql` that are designed to handle prefix scans where the prefix can be satisfied by an index. Our logic currently doesn't handle transforming `LIKE '${str}%'` into a tuple scan with a range like `[PREFIX_STRING ${str}]`, which means that we don't use the index as efficiently as we could. This is shown in the `explain`s here, where we only use the index in order to get a set of covering fields, but we don't add it to a scan range.

Note that this is relevant to #3523 (addressing #1076), as if we did turn those kinds of `LIKE` expressions into prefix string scans, we'd get mixed-mode coverage of the fix for the bug being addressed there. AFAIK, there isn't any other SQL comparison that we have that we could use to generate prefix string scans in some other way.


arnaud-lacurie:
In theory, we could also add tests like:

```
LIKE 'cove_t'
```

which could be broken down into a prefix search and a residual string comparison, to find `covert` and `covent` for example.

alecgrieser:
I've added a set of additional test cases that have prefixes but have tails that need to be compensated for. That includes some patterns with underscores (like `ca_al`) and with wildcards (like `ca%al`).

PR #3541, opened 2025-08-15
Add serialized plan complexity and hash to the continuation info in EXPLAIN

hazefully:
This PR adds two new columns, `PLAN_HASH` and `SERIALIZED_PLAN_COMPLEXITY`, to the `PLAN_CONTINUATION` struct in the result from EXPLAIN which contain the plan hash stored in the continuation and the complexity of the plan serialized within the continuation (if the continuation contains a compiled statement )respectively . 

The new columns are populated for EXPLAINs on an `EXECUTE CONTINUATION` statement and a regular SELECT statement that uses the `WITH CONTINUATION` clause.

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
@@ -81,6 +81,7 @@ import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
+import com.google.protobuf.Option;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -469,6 +470,9 @@ public abstract class QueryPlan extends Plan<RelationalResultSet> implements Typ
         @Nonnull
         private final PlanHashMode serializedPlanHashMode;
 
+        @Nonnull
+        private final Supplier<Integer> serializedPlanHashSupplier;
+
         public ContinuedPhysicalQueryPlan(@Nonnull final RecordQueryPlan recordQueryPlan,
                                           @Nonnull final TypeRepository typeRepository,
                                           @Nonnull final QueryPlanConstraint continuationConstraint,
@@ -479,6 +483,7 @@ public abstract class QueryPlan extends Plan<RelationalResultSet> implements Typ
             super(recordQueryPlan, null, typeRepository, QueryPlanConstraint.noConstraint(),
                     continuationConstraint, queryExecutionParameters, query, currentPlanHashMode);
             this.serializedPlanHashMode = serializedPlanHashMode;
+            this.serializedPlanHashSupplier = Suppliers.memoize(() -> recordQueryPlan.planHash(serializedPlanHashMode));
         }
 
         @Nonnull
@@ -524,6 +529,11 @@ public abstract class QueryPlan extends Plan<RelationalResultSet> implements Typ
         @Nonnull
         protected Optional<RecordQueryPlan> getSerializedPlanFromContinuation(@Nonnull ContinuationImpl parsedContinuation,
                                                                               @Nonnull ExecutionContext executionContext) {
+            if (parsedContinuation.getPlanHash() == null ||
+                    !parsedContinuation.getPlanHash().equals(serializedPlanHashSupplier.get())) {
+                return Optional.empty();
+            }
+
             // No need to deserialize the plan from the continuation again
             return Optional.of(getRecordQueryPlan());
         }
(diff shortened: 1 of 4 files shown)


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -492,6 +520,26 @@ protected <M extends Message> void validatePlanAgainstEnvironment(@Nonnull final
...
+
+            final var currentPlan = this.getRecordQueryPlan();
+            final var continuationPlanHash = parsedContinuation.getPlanHash();
+            final var serializedPlanHashMode = PlanHashable.PlanHashMode.valueOf(compiledStatement.getPlanSerializationMode());
+            if (continuationPlanHash != null && continuationPlanHash.equals(currentPlan.planHash(serializedPlanHashMode))) {
+                return Optional.of(currentPlan);
+            }
+
+            // Fallback to deserializing the plan from the continuation
+            return super.getPlanFromContinuation(parsedContinuation, executionContext);
```
g31pranjal: IIUC, this should never happen. Because, the entire premise of `ContinuedPhysicalQueryPlan` is based on the fact that the plan is from the continuation. I believe, we should throw an exception here that the plan hashes don't match.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -280,12 +281,37 @@ protected <M extends Message> void validatePlanAgainstEnvironment(@Nonnull final
...
+                );
+                final var serializationContext = new PlanSerializationContext(
+                        DefaultPlanSerializationRegistry.INSTANCE, planSerializationMode
+                );
+                return Optional.of(
+                        RecordQueryPlan.fromRecordQueryPlanProto(serializationContext, compiledStatement.getPlan()));
+            } catch (RelationalException ex) {
+                return Optional.empty();
+            }
+        }
```
g31pranjal: This is probably fine in the interim, but we want to move away from the `SELECT .... WITH CONTINUATION ...` syntax and instead have `EXECUTE CONTINUATION ...` for using continuations that always contain compiled statements. This is something that we can do today, as I don't think the former syntax is in active use.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
```diff
@@ -280,12 +281,37 @@ protected <M extends Message> void validatePlanAgainstEnvironment(@Nonnull final
...
+
+        @Nonnull
+        private RelationalResultSet executeExplain(@Nonnull ContinuationImpl parsedContinuation, ExecutionContext executionContext) {
             final var continuationStructType = DataType.StructType.from(
                     "PLAN_CONTINUATION", List.of(
                             DataType.StructType.Field.from("EXECUTION_STATE", DataType.Primitives.BYTES.type(), 0),
                             DataType.StructType.Field.from("VERSION", DataType.Primitives.INTEGER.type(), 1),
-                            DataType.StructType.Field.from("PLAN_HASH_MODE", DataType.Primitives.STRING.type(), 2)),
+                            DataType.StructType.Field.from("PLAN_HASH_MODE", DataType.Primitives.STRING.type(), 2),
+                            DataType.StructType.Field.from("PLAN_HASH", DataType.Primitives.INTEGER.type(), 3),
```
g31pranjal, line 313: We might not have a successful run when the actual plan hash is different from the continuation plan hash. 

Hence, this is redundant in some sense but valuable if the scenarios ever arise.  Having a test for that scenario will be nice!

g31pranjal (changes requested):

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
hazefully: As discussed offline, AFAICT a difference between the actual plan hash and the plan hash in the continuation can only happen with `SELECT .. WITH CONTINUATION` queries, but not with `EXECUTE CONTINUATION` queries. I added a test in 616b73095b5e93695dd9cb1c81a78c6760c9936f that shows this possibility.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
hazefully: As discussed offline, I will leave this in the PR which could be cleaned up once the support for the `SELECT ... WITH CONTINUATION` syntax is removed.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/QueryPlan.java
hazefully: I removed the fallback in 9954155 and added an assertion to compare the plan hash from the continuation and the hash of the plan `ContinuedPhysicalQueryPlan` in ee349c9 to catch future potential mis-uses of this method that may end up passing a continuation that is not originating from the plan's `QueryExecutionContext` (i.e. a different continuation than the one used to construct the `ContinuedPhysicalQueryPlan`).

PR #3542, opened 2025-08-18
Fix insertion of Struct with Enum

g31pranjal:
This PR fixes a minor bug that prevented the insertion of 

1. `struct with enum`: This was because of the fact that target `enum` descriptor was not getting propagated rightly during the message coercion.   
2. `array of struct with enum`: This was because of the fact that the array of structs was restricted to do promotion of struct, if needed.

### DIFF
--- a/fdb-relational-core/src/testFixtures/java/com/apple/foundationdb/relational/utils/ArrayAssert.java
+++ b/fdb-relational-core/src/testFixtures/java/com/apple/foundationdb/relational/utils/ArrayAssert.java
@@ -21,15 +21,17 @@
 package com.apple.foundationdb.relational.utils;
 
 import com.apple.foundationdb.annotation.API;
-
 import com.apple.foundationdb.relational.api.RelationalResultSet;
-
+import com.apple.foundationdb.relational.api.RelationalStruct;
 import org.assertj.core.api.AbstractAssert;
 import org.assertj.core.api.Assertions;
+import org.assertj.core.api.SoftAssertions;
 
 import java.sql.Array;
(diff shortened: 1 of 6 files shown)


fdb-relational-core/src/testFixtures/java/com/apple/foundationdb/relational/utils/ArrayAssert.java
```diff
@@ -79,4 +86,33 @@ public ArrayAssert isEqualTo(Array expected) {
...
+    @SuppressWarnings("unchecked")
+    private ArrayAssert isEqualTo(List<?> expected) {
+        if (expected == null) {
+            Assertions.assertThat(actual).isNull();
+            return this;
+        }
+        try {
+            ResultSet rs = actual.getResultSet();
+            Assertions.assertThat(rs).isInstanceOf(RelationalResultSet.class);
+            var actualValues = (Object[]) actual.getArray();
```
pengpeng-lu, line 99: should we check `actualValues` is not null here?

pengpeng-lu (approved):
just a small comment about checking nulls.

fdb-relational-core/src/testFixtures/java/com/apple/foundationdb/relational/utils/ArrayAssert.java
g31pranjal: I think, yes, it is theoretically possible that getArray() returns null for a nullable array field. And hence, will be a good call to add that check here when dealing with asserts on that.

PR #3551, opened 2025-08-19
Fix for Lucene block leaks on file deletion

### LINKED ISSUE #3550: Lucene data blocks are leaking when files are deleted
Lucene files are leaking blocks when deleted. There seem to be dangling blocks that remain that belong to no valid file.

ohadzeliger:
The source of the leaking data blocks has been found and fixed.
Added more tests to verify the leakage was fixed directly as well as an addition to the `LuceneIndexTestValidator` to get some additional coverage "for free" whenever we use it in other tests.

Resolves #3550

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -122,7 +122,8 @@ public class FDBDirectory extends Directory  {
     public static final int DEFAULT_INITIAL_CAPACITY = 128;
     private static final int SEQUENCE_SUBSPACE = 0;
     private static final int META_SUBSPACE = 1;
-    private static final int DATA_SUBSPACE = 2;
+    @VisibleForTesting
+    public static final int DATA_SUBSPACE = 2;
     @SuppressWarnings({"Unused", "PMD.UnusedPrivateField"}) // preserved to document that this is reserved
     private static final int SCHEMA_SUBSPACE = 3;
     private static final int PRIMARY_KEY_SUBSPACE = 4;
@@ -657,7 +658,7 @@ public class FDBDirectory extends Directory  {
 
     @Nonnull
     @VisibleForTesting
-    protected CompletableFuture<Map<String, FDBLuceneFileReference>> getFileReferenceCacheAsync() {
+    public CompletableFuture<Map<String, FDBLuceneFileReference>> getFileReferenceCacheAsync() {
         if (fileReferenceCache.get() != null) {
             return CompletableFuture.completedFuture(fileReferenceCache.get());
         }
@@ -741,8 +742,8 @@ public class FDBDirectory extends Directory  {
         if (fieldInfosStorage.delete(id)) {
             agilityContext.clear(fieldInfosSubspace.pack(id));
         }
-        // Nothing stored here currently.
-        agilityContext.clear(dataSubspace.subspace(Tuple.from(id)).range());
+        // Clear all data blocks for the file
+        agilityContext.clear(dataSubspace.subspace(Tuple.from(value.getId())).range());
 
         // Delete K/V data: If the deferredDelete flag is on then delete all content from K/V subspace for the segment
         // (this is to support CFS deletion, that will be disjoint from the actual file deletion).
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
@@ -734,16 +734,16 @@ public class FDBDirectory extends Directory  {
     @VisibleForTesting
     protected boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference> cache, @Nonnull String name) throws IOException {
         // TODO make this transactional or ensure that it is deleted in the right order
-        FDBLuceneFileReference value = cache.remove(name);
-        if (value == null) {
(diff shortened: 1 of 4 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -741,8 +742,8 @@ protected boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference
         if (fieldInfosStorage.delete(id)) {
             agilityContext.clear(fieldInfosSubspace.pack(id));
         }
-        // Nothing stored here currently.
-        agilityContext.clear(dataSubspace.subspace(Tuple.from(id)).range());
+        // Clear all data blocks for the file
+        agilityContext.clear(dataSubspace.subspace(Tuple.from(value.getId())).range());
```
ScottDugas: Perhaps also worth renaming `value` to `fileReference`

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
```diff
@@ -132,14 +136,74 @@ void validate(Index index, final Map<Tuple, ? extends Map<Tuple, Tuple>> expecte
                     .collect(Collectors.toList());
             if (partitionHighWatermark > 0) {
                 validatePartitionedGroup(index, universalSearch, allowDuplicatePrimaryKeys, groupingKey, partitionLowWatermark, partitionHighWatermark, records, missingDocuments);
+                validatePartitionedDanglingBlocks(index, groupingKey);
```
ScottDugas: Why do this at this level, rather than inside the existing `validatePartitionedGroup`/`validateUnpartitionedGroup` which already cover iterating over the partitions, if appropriate?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
```diff
@@ -251,4 +253,75 @@ private void assertMetricCountAtMost(StoreTimer.Event metric, int maximumValue)
         assertThat("Metric " + metric + " should be called at most " + maximumValue + " times",
                 timer.getCount(metric), lessThanOrEqualTo(maximumValue));
     }
+
+    @Test
+    public void testDataBlocksProperlyCleared() throws Exception {
+        // Create a file with data blocks
+        long fileId = 1L;
+        FDBLuceneFileReference reference = new FDBLuceneFileReference(fileId, 100, 100, 1024);
```
ScottDugas, line 261: These tests are using lower-level, kind of internal aspects of `FDBDirectory`, perhaps it would be better to use public methods like `createOutput`, and `deleteFile`, and then assert that there is nothing in the `DATA` subspace. You still have to look at the raw data in FDB, so maybe not a big difference.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -741,8 +742,8 @@ protected boolean deleteFileInternal(@Nonnull Map<String, FDBLuceneFileReference
         if (fieldInfosStorage.delete(id)) {
             agilityContext.clear(fieldInfosSubspace.pack(id));
         }
-        // Nothing stored here currently.
-        agilityContext.clear(dataSubspace.subspace(Tuple.from(id)).range());
+        // Clear all data blocks for the file
```
ScottDugas, line 745: Also, confirming that `LuceneIndexMaintenanceTest` fails without this change, but with the new assertions in `LuceneIndexTestValidator`, correct?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: Done.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
ohadzeliger: Yes, undoing the fix with the additional assertions in the validator causes a bunch of failures, looking like: 
```
Found orphaned data blocks for file IDs: (388, 198), (422, 37), (388, 197), (422, 38), (388, 200), (422, 35), (229, 0), (388, 199), (422, 36), (388, 202), (422, 41), (388, 201), (422, 42), (388, 204), (422, 39), (388, 203), (422, 40), (388,
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexTestValidator.java
ohadzeliger: Based on the comment above `A broad validation of the lucene index`, I thought that was the original intent - for some set of validations that can be made somewhat modular. Either way would be good. Changed.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/directory/FDBDirectoryTest.java
ohadzeliger: Yes, these are the lower-level tests that assert that the fix works, quite directly. The `TestValidator` provides more of a broad functional coverage. I thought this is a good coverage.

PR #3555, opened 2025-08-20
Add support for detailed file info in the LuceneMetadata operation. Make metadata class API.EXPERIMENT

### LINKED ISSUE #3554: Need more info in LuceneMetadata index operation
The Lucene metadata info Index opratation lists the file names. In order to perform more elaborate matching of file ID to blocks, adding more information to the metadata is needed.

ohadzeliger:
Add more info to the Lucene Index Operation metadata (file ID and size)
Make the `LuceneGetMetadataInfo` class `EXPERIMENTAL` (it is used from outside the RL).
Deprecate the metadata's file name list (users can substitute the new field to get access to the file names).
 
Resolves #3554

### DIFF
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
@@ -119,7 +119,6 @@ public class LuceneMetadataInfo extends IndexOperationResult {
         /**
          * The detailed file info from the directory.
          * (Optional) - if the request did not specify that this is needed, this would be null.
-         * Note: This method returns the internal collection. It should not be modified.
          * @return the list of fileInfos in the directory
          */
         @Nullable


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
```diff
@@ -62,13 +66,27 @@ public Map<Integer, LuceneInfo> getLuceneInfo() {
      */
     public static class LuceneInfo {
         private final int documentCount;
+        @API(API.Status.DEPRECATED) // replaced by the detailedFileInfos field
         private final Collection<String> files;
         private final int fieldInfoCount;
+        @Nullable
+        private final Collection<LuceneFileInfo> detailedFileInfos;
 
+        @Deprecated(forRemoval = true)
```
jjezra: There was some discussion about it, and I think that `@API(API.Status.DEPRECATED)` is preferred.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
```diff
@@ -62,13 +66,27 @@ public Map<Integer, LuceneInfo> getLuceneInfo() {
...
             this.files = files;
             this.fieldInfoCount = fieldInfoCount;
+            this.detailedFileInfos = null;
+        }
+
+        public LuceneInfo(final int documentCount,
+                          final int fieldInfoCount,
+                          @Nonnull final Collection<LuceneFileInfo> detailedFileInfos) {
+            this.documentCount = documentCount;
+            this.files = detailedFileInfos.stream().map(LuceneFileInfo::getName).collect(Collectors.toList());
```
jjezra, line 88: Can "files" here have a null assingment, and `getFiles` below be something like: 
```
if (files == null && detailedFileInfos != null) {
  return detailedFileInfos.stream().map(LuceneFileInfo::getName).collect(Collectors.toList());
} 
// else return an error...
```
?

I assume that since `files` is private, the callers will always use the `getFiles` function to access this data.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
```diff
@@ -62,13 +66,27 @@ public Map<Integer, LuceneInfo> getLuceneInfo() {
...
+            this.fieldInfoCount = fieldInfoCount;
+            this.detailedFileInfos = detailedFileInfos;
```
jjezra: Is there a good reason to store a copy of the `detailedFileInfos` as teamscale suggests? 
(The class is public, is it likely to be constructed outside this code? )

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
ohadzeliger: Right. I was trying to get Teamscale to not mark it as a test gap, but that didn't work.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
ohadzeliger: Yes, we can do that, but then we will recompute the derivation every time instead of once.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
ohadzeliger: Personally, I'd like to minimize cloning of data structures for use with more public APIs and mutable classes. This is a data holder and the intended use is to store results as they are produced before they are going away. 
Added unmodifiableCollection here for extra safety, where users of the class cannot modify the internals.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
jjezra:
```
if (files == null && detailedFileInfos != null) {
  files = detailedFileInfos.stream().map(LuceneFileInfo::getName).collect(Collectors.toList());
  return files;
} 
```
?

ohadzeliger: I think this is exactly what  constructors are doing - except that this way the class is immutable: First constructor has populates `files` and second constructor has `files==null` and `detailedFileInfos != null` and is therefore exactly what you have in your comment.
Since the `files` field is deprecated (to be removed once usage is gone), then we will be left with a simple constructor initialization.

ohadzeliger: I changed the constructor to take a list and added `unmodifieableList` to the constructor. This allows the `equals` method to work properly.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneMetadataInfo.java
ohadzeliger: Done.

PR #3556, opened 2025-08-20
Extract explain query config into a separate file

ScottDugas:
Teamscale informed me that this method/anonymous class is quite long, so I extracted it into a separate class. 

I recommend reviewing commit-by-commit.


g31pranjal (approved):
Nice!

PR #3558, opened 2025-08-26
Re enable auto test

ScottDugas:
This allows one to run `@AutomatedTest` tests. It does not add it to any regular test run.

### DIFF
--- a/fdb-relational-core/fdb-relational-core.gradle
+++ b/fdb-relational-core/fdb-relational-core.gradle
@@ -63,11 +63,12 @@ dependencies {
     testFixturesImplementation(project(coreProject))
 }
 
-/*
-task autoTest(type: Test){
+
+// The engine is defined in relational-core. If we move it somewhere such that other subprojects can use it, this
+// task should be moved to testing.gradle
+tasks.register('autoTest', Test) {
     description = 'Runs the Automated end-to-end tests.'
     group = 'verification'
-//    testClassesDir = sourceSets.test.output.classesDirs
     classpath = sourceSets.test.runtimeClasspath
 
     useJUnitPlatform {
@@ -79,7 +80,7 @@ task autoTest(type: Test){
         junitXml.outputLocation = layout.buildDirectory.dir("autotest-test-results")
     }
 }
- */
+
 
 publishing {
     publications {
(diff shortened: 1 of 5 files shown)


gradle/testing.gradle
```diff
@@ -282,7 +282,7 @@ tasks.withType(Test) { theTask ->
 }
 
 tasks.withType(Test).configureEach { task ->
-    if (task.name != 'performanceTest') {
+    if (task.name != 'performanceTest' && task.name != 'autoTest') {
```
alecgrieser: Is there a reason we can't validate async-to-sync calls in auto test? Does something break?

ScottDugas: No Idea, that wasn't what actually broke. I can't recall what failed.
That being said, I think it would be better to just add the AutoTestEngine to nightly, and not have a separate gradle testing task.

fdb-relational-core/fdb-relational-core.gradle
```diff
@@ -63,23 +63,15 @@ dependencies {
     testFixturesImplementation(project(coreProject))
 }
 
-/*
-task autoTest(type: Test){
```
alecgrieser, line 67: With this task gone, is there a way to run the auto-tests locally? Would you have to do something like `./gradlew test -Ptests.nightly`? It might be nice if we had the `autoTest` task just to facilitate locally running things, but I'm not too fussed one way or the other.

I guess a related question would be: would it make sense to just run these during PRB as well, or do they add too much time/instability?

alecgrieser (approved):

fdb-relational-core/fdb-relational-core.gradle
ScottDugas: I don't know how much instability they might add, so I'm reticent to put them in PRB. The original authors, I think were under the same opinion.
If we wanted to get them to run during PRB, I would advocate that they have a way to have a fixed seed, but it would be better to get them running in nightly first, and then work to get them running during PRB.

The current way to run them locally is to run with `-Ptests.nightly`. I would rather add a `nightlyTest` task then a specific `autoTest` task, but I'm also somewhat reticent to add more types of test tasks.

ScottDugas: I changed it to have a fixed seed across any `@AutomatedTest`, but in doing so discovered that they all had fixed seeds (`1l`). I tried running with `-Ptests.includeRandom=true`, and it failed reliably because it couldn't find a Struct. I didn't want to push something that would pretty reliably fail the nightly build, so these are just being enabled with a fixed seed, but they will always run.

Issue #3559, opened 2025-08-27
Duplicate elements provided to an IN predicate can lead to duplicate results


hazefully:
Cascades attempts to transform an `IN` predicate to a nested-loop join (i.e. `RecordQueryFlatMapPlan` or `RecordQueryInJoinPlan`). This can lead to duplicate records in the result if the in-list contains duplicate values, example:

```sql
SELECT a, b FROM ta WHERE b IN (1, 1, 1, 1)
```

Should result in something like `[{A: 8, 1}]`, but would actually return `[{A: 8, 1}, {A: 8, 1}, {A: 8, 1}, {A: 8, 1}]` as it would match the record with each instance of `1` from the in-list.

For the results of the plan to be correct, the in-list should be de-duplicated before executing the plan.

PR #3560, opened 2025-08-27
Add new wrapper value for removing duplicates from array values

hazefully:
This PR prepares for fixing the issue in #3559 by adding a new `Value` to Cascades,`ArrayDistinctValue`, which can be used to wrap an underlying array value to remove duplicates from the underlying array during evaluation. The purpose of this new value is to use it to wrap in-lists used in `InJoin`, `InUnion` and `FlatMap` plans to avoid duplicate records if the in-list contains duplicate values.

~~A new function `array_distinct` is added to the SQL layer that allows users to make use of the same functionality in their queries.~~ (this will be introduced in a subsequent PR).

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
@@ -79,6 +79,7 @@ public class ArrayDistinctValue extends AbstractValue implements ValueWithChild 
 
     @Nonnull
     @Override
+    @SuppressWarnings("PMD.CompareObjectsWithEquals")
     public ValueWithChild withNewChild(@Nonnull final Value rebasedChild) {
         if (getChild() == rebasedChild) {
             return this;


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
```diff
@@ -0,0 +1,171 @@
...
+    @Nonnull
+    @Override
+    public Value getChild() {
+        return childValue;
+    }
+
+    @Nonnull
+    @Override
+    public ValueWithChild withNewChild(@Nonnull final Value rebasedChild) {
+        return new ArrayDistinctValue(rebasedChild);
```
normen662, line 87: Can you do an `== ` comparison here and not return a new instance if that is the same?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public int planHash(@Nonnull final PlanHashMode mode) {
+        return PlanHashable.objectsPlanHash(mode, BASE_HASH, childValue);
+    }
+
+    @Nonnull
+    @Override
+    public ExplainTokensWithPrecedence explain(@Nonnull final Iterable<Supplier<ExplainTokensWithPrecedence>> explainSuppliers) {
+        return ExplainTokensWithPrecedence.of(new ExplainTokens().addFunctionCall("arrayDistinct",
```
normen662, line 118: Is there a test for the explain?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
```diff
@@ -0,0 +1,171 @@
...
+        return resultType;
+    }
+
+    @Override
+    public <M extends Message> Object eval(@Nullable final FDBRecordStoreBase<M> store, @Nonnull final EvaluationContext context) {
+        final var childResult = childValue.eval(store, context);
+        if (childResult == null) {
+            return null;
+        }
+        return ((List<?>)childResult).stream().distinct().collect(ImmutableList.toImmutableList());
```
normen662, line 102: What happens for `Message`s? Is there a test for that?

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
hazefully: Yes, `testExplain` in `ArrayDistinctValueTest`, but it is probably easy to miss because of how I was ordering the methods in the test class; I moved all the static methods to the top of the test class which as far as I can tell is what most other test classes do

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/ArrayDistinctValue.java
hazefully: I added test cases to confirm that duplicates elements in a `List<Message>` are removed correctly (duplicate `Message` with the same descriptor and the same values for fields are removed).

PR #3562, opened 2025-08-28
Ensure that `TransactionalRunner::runAsync` closes a context that it creates even if future creation fails

### LINKED ISSUE #3561: Potential transaction leak in `TransactionalRunner::runAsync`
In `TransactionalRunner::runAsync`, there's a potential transaction leak here:

https://github.com/FoundationDB/fdb-record-layer/blob/68f805abea8b16ee461e95abde65dde53e4bd45d/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/runners/TransactionalRunner.java#L101-L104

Note that the transaction is opened on line 101 and then it is supposed to be closed in the callback on line 104. However, if there's an exception when invoking the `runnable` function or even in `commitAsync()` that prevents the future from being created, then we may never actually create the callback. We should be sure to guard the creation of the lambda and close the context if we don't create the future.

alecgrieser:
There was a case that could result in a transaction being created that was not closed. One thing to note is that the exception actually is closed with the current set up when the `TransactionalRunner` is closed, so this leak should only be a problem if the runner isn't being closed (an API error), but it would be good anyway to ensure that this gets closed.

This fixes #3561.

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/runners/TransactionalRunnerTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/runners/TransactionalRunnerTest.java
@@ -140,6 +140,11 @@ class TransactionalRunnerTest {
         }
     }
 
+    /**
+     * Validate the behavior when the future returned by {@link TransactionalRunner#runAsync(boolean, Function)}
+     * completes exceptionally. The work done by the transaction should not be committed, and the underlying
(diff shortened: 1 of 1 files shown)


alecgrieser:
I'd believe that there are other places where a similar pattern is needed. (And a need for something similar in adopter logic as well in similar circumstances.) I didn't do an audit this time, but I think I've seen similar places that needed _slightly_ different logic, but there very well may be a way that we could unify those. It also is the kind of thing that a macro would be useful for, but alas

PR #3566, opened 2025-09-03
Keyspacepath export

ScottDugas:


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/EnvironmentKeySpace.java
```diff
@@ -0,0 +1,195 @@
...
+import com.apple.foundationdb.Transaction;
+import com.apple.foundationdb.record.provider.foundationdb.FDBDatabase;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordStore;
+import com.apple.foundationdb.tuple.Tuple;
+
+import javax.annotation.Nonnull;
+import java.util.UUID;
+
+/**
```
ScottDugas, line 32: In case you're reviewing this top-to-bottom, this class was extracted from KeySpaceDirectoryTest, but `setupSampleData` was added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -337,6 +341,21 @@ public String toString() {
         return toString(null);
     }
 
+    @Nonnull
+    @Override
+    public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext context,
+                                                          @Nullable byte[] continuation,
+                                                          @Nonnull ScanProperties scanProperties) {
+        return new LazyCursor<>(toTupleAsync(context)
+                .thenApply(tuple -> KeyValueCursor.Builder.withSubspace(new Subspace(tuple))
```
alecgrieser, line 380: Just to try and head off some potential problems: in #3397, there's a modification being made to the `KeyValueCursor` continuation to wrap it in a protobuf. This would allow us to avoid returning an empty `ByteString` as the continuation in case there is a single key range, which may be something that a `KeySpacePath` could run into. (It would require that the first key in the keyspace path is set, I think.)

There's sort of an open question as to how we migrate all of the uses over from the old format to the new format. It would be nice if we could set the serialization mode on this one to the new format so that we don't have to worry about migrating it later, I think. Though that would introduce a dependency on #3397 being merged first.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
```diff
@@ -0,0 +1,76 @@
...
+public class DataInKeySpacePath {
+
+    final CompletableFuture<ResolvedKeySpacePath> resolvedPath;
+    final KeyValue rawKeyValue;
+
+    public DataInKeySpacePath(KeySpacePath path, KeyValue rawKeyValue, FDBRecordContext context) {
+        this.rawKeyValue = rawKeyValue;
+
+        // Convert the raw key to a Tuple and resolve it starting from the provided path
+        Tuple keyTuple = Tuple.fromBytes(rawKeyValue.getKey());
```
alecgrieser: Do we need to be concerned with keys that are not `Tuple` parseable? I believe all of the keys that we currently generate are `Tuple` parseable, though we could adjust that in the future

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
```diff
@@ -0,0 +1,76 @@
...
+        
+        // First resolve the provided path to get its resolved form
+        this.resolvedPath = path.toResolvedPathAsync(context).thenCompose(resolvedPath -> {
+            // Now use the resolved path to find the child for the key
+            // We need to figure out how much of the key corresponds to the resolved path
+            Tuple pathTuple = resolvedPath.toTuple();
```
alecgrieser: Should this validate that `pathTuple` is a prefix of `keyTuple`? Strictly speaking, I don't think the check is necessary, but it could be a sensible thing to validate

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
```diff
@@ -0,0 +1,76 @@
...
+    final KeyValue rawKeyValue;
+
+    public DataInKeySpacePath(KeySpacePath path, KeyValue rawKeyValue, FDBRecordContext context) {
+        this.rawKeyValue = rawKeyValue;
+
+        // Convert the raw key to a Tuple and resolve it starting from the provided path
+        Tuple keyTuple = Tuple.fromBytes(rawKeyValue.getKey());
+        
+        // First resolve the provided path to get its resolved form
+        this.resolvedPath = path.toResolvedPathAsync(context).thenCompose(resolvedPath -> {
```
alecgrieser: It almost seems like the method provided should be a method on the `KeySpacePath`. I was sort of surprised that we didn't already have it. I can see how we'd have to be clear in the name that we're starting with the full key rather than just the suffix. I suppose we _could_ structure this here to go the other way: we resolve the `KeySpacePath` first, use it to get a raw byte prefix, and then resolve the `Tuple` suffix using `findChildForKey`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
```diff
@@ -0,0 +1,333 @@
...
+            byte[] keyBytes = keyTuple.pack();
+            byte[] valueBytes = Tuple.from("accessor_test").pack();
+            
+            KeyValue originalKeyValue = new KeyValue(keyBytes, valueBytes);
+            DataInKeySpacePath dataInPath = new DataInKeySpacePath(testPath, originalKeyValue, context);
+            
+            // Verify accessor methods
+            KeyValue retrievedKeyValue = dataInPath.getRawKeyValue();
+            assertNotNull(retrievedKeyValue);
+            assertEquals(originalKeyValue.getKey(), retrievedKeyValue.getKey());
```
alecgrieser: I believe that for arrays, the assertion here ends up asserting that the two arrays are pointer equal. Which I guess makes sense given that the original `KeyValue` is just being wrapped. That may be what you inteded to assert here, but I just wanted to make sure that that's what was going on here.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
```diff
@@ -0,0 +1,333 @@
...
+            // Create DataInKeySpacePath from the binary_store-level path
+            KeySpacePath storePath = root.path("binary_store");
+            DataInKeySpacePath dataInPath = new DataInKeySpacePath(storePath, keyValue, context);
+            
+            ResolvedKeySpacePath resolved = dataInPath.getResolvedPath().join();
+            
+            // Verify the path using assertNameAndValue
+            assertEquals("blob_id", resolved.getDirectoryName());
+            byte[] resolvedBytes = (byte[]) resolved.getResolvedValue();
+            assertArrayEquals(blobId, resolvedBytes);
```
alecgrieser, line 259: Can this not use `assertNameAndValue` because `assertEquals` will check pointer-equality when provided with on `byte[]`s?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
```diff
@@ -0,0 +1,333 @@
...
+ * Tests for {@link DataInKeySpacePath}.
+ */
+@Tag(Tags.RequiresFDB)
+class DataInKeySpacePathTest {
+    @RegisterExtension
+    final FDBDatabaseExtension dbExtension = new FDBDatabaseExtension();
+
+    @ParameterizedTest
+    @ValueSource(ints = {0, 1, 2, 3, 4, 5})
+    void resolution() {
```
alecgrieser: Is the parameter actually used in this test?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
```diff
@@ -0,0 +1,333 @@
...
+            final Tuple remainderTuple = Tuple.from("salary", 75000L, "start_date", "2023-01-15");
+            byte[] keyBytes = employeePath.toSubspace(context).pack(remainderTuple);
+            byte[] valueBytes = Tuple.from("employee_record").pack();
+
+            tr.set(keyBytes, valueBytes);
+            KeyValue keyValue = new KeyValue(keyBytes, valueBytes);
+
+            // Create DataInKeySpacePath from the company-level path
+            KeySpacePath companyPath = root.path("company");
+            DataInKeySpacePath dataInPath = new DataInKeySpacePath(companyPath, keyValue, context);
```
alecgrieser: It might be nice to have an assertion somewhere that states that you can choose any path along the way, and all of those will return the same `DataInKeySpacePath`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
```diff
@@ -0,0 +1,76 @@
...
+import com.apple.foundationdb.record.provider.foundationdb.FDBRecordContext;
+import com.apple.foundationdb.tuple.Tuple;
+import com.apple.foundationdb.tuple.TupleHelpers;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Class representing a {@link KeyValue} pair within in {@link KeySpacePath}.
+ */
+public class DataInKeySpacePath {
```
alecgrieser, line 37: Should this implement `equals` and `hashCode`?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
```diff
@@ -0,0 +1,619 @@
...
+
+    /**
+     * Export all the data, and make some assertions that can always be done.
+     * This combines a lot of assertions, but most of the underlying behavior should be well covered by the objects
+     * that {@link KeySpacePath#exportAllData} is built on.
+     * @param pathToExport the path being exported
+     * @param context the context in which to export
+     * @return a list of the raw {@code KeyValue}s being exported
+     */
+    private static List<KeyValue> exportAllData(final KeySpacePath pathToExport, final FDBRecordContext context) {
```
alecgrieser: Any reason this returns `KeyValue`s instead of `DataInKeySpacePath`s?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
```diff
@@ -0,0 +1,619 @@
...
+        }
+        if (20 % limit == 0) {
+            expectedBatches.add(List.of());
+        }
+        
+        // Export with continuation support
+        try (FDBRecordContext context = database.openContext()) {
+            KeySpacePath continuationPath = root.path("continuation");
+
+            final ScanProperties scanProperties = ScanProperties.FORWARD_SCAN.with(props -> props.setReturnedRowLimit(limit));
```
alecgrieser: Should we also test reverse scans like this?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
```diff
@@ -0,0 +1,619 @@
...
+            for (KeyValue kv : allData) {
+                String valueStr = new String(kv.getValue());
+                assertTrue(valueStr.startsWith("binary_data_"));
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = {1, 2, 3, 30})
+    void exportAllDataWithContinuation(int limit) {
```
alecgrieser: Another case that is relevant to continuations: scanning a single record key space path. Something like:

1. Resolving a path to a tuple then
2. Setting that specific key to some value and
3. Validating that we get that value back and that the continuation that comes back is sensible (not null and ideally not empty)

I'm not actually sure we'd get that key back (that is, whether the scan is over a strict subspace or not). And if it does return a continuation, I actually think it would return an empty continuation for the first element (because of #3206, see #3397), but it would be nice to know.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: That's a good point.
I changed `exportAllData` on the interface to be `EXPERIMENTAL`, and added a note that we expect the continuation change without preserving backwards compatibility when that merges. If the rest of the pieces are in place such that this is useful before #3397 gets merged, we can potentially update the javadoc, and make this be backwards compatible.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
ScottDugas: Yeah, probably, but I'm not entirely sure how to do that with the `CompletableFuture` for the resolved path.
It can either fall through and say it's equal if the original path, and raw value or equal, or, alternately, I can change it so that the constructor takes the resolved path.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
ScottDugas: `KeySpacePathImpl` always produces tuples, and AFAIK every key we have ever used has been in tuples. I feel like committing to that at this point, is worthwhile. If we find a strong reason to put non-tuple data in a key, we'll have to cross that bridge when we get to it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
```diff
@@ -566,4 +566,20 @@ default List<ResolvedKeySpacePath> listSubdirectory(@Nonnull FDBRecordContext co
...
+     * Supports continuation to resume scanning from a previous position.
+     *
+     * @param context the transaction context in which to perform the data export
+     * @param continuation optional continuation from a previous export operation, or null to start from the beginning
+     * @param scanProperties properties controlling how the scan should be performed
+     * @return a RecordCursor that iterates over all KeyValue pairs under this path
+     */
+    @API(API.Status.UNSTABLE)
+    @Nonnull
+    RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext context,
```
ScottDugas: I realized that this is an interface, and while I don't think we intended anyone to extend it, I should probably put a default implementation in.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
ScottDugas: Yeah, I was also kind of surprised. I was a bit fixated on getting this working, that I didn't pause to think that I should add it there. I'll move most of this there.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
ScottDugas: Yeah, I think that's a good double check.

ScottDugas: Note: I took advantage of `TupleHelpers.isPrefix`, which apparently didn't have any unit tests, but I'm adding them in a followup PR. https://github.com/FoundationDB/fdb-record-layer/pull/3578/files#diff-c7f3ee5c693bba884071e40ab98f13c12a86350a819f71f1c6a60eb7b750623c
I'm not sure if it's worth bringing into this PR.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ScottDugas: True, updated, but I'm probably going to make more substantial changes to the api here, namely hiding the key entirely.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ScottDugas: Yeah, but I updated `assertNameAndValue` to check for `byte[]` and use `assertArrayEquals` appropriately.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ScottDugas: It is not, but it was supposed to be used to resolve from a path deeper than the root. Updated the test to actually use it.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ScottDugas: Yeah, that was the intent of the parameter. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
ScottDugas: I started doing this path, an discovered that `ResolvedKeySpacePath.equals` is broken, and I've started working on a fix for that. Its a fair amount of new tests, and a couple other equals/hashcode implementations, so I'm going to pull that out into a separate PR, and come back to this once it is fixed. 

Issue: #3594

ScottDugas: So, I went ahead and added `ResolvedKeySpacePath.equals`, and updated the tests to take advantage of it, but this class holds onto a future for the resolved path, making `equals` hard. It's possible that as I go along, i'll see that it makes more sense to have `exportAllData` resolve these futures so that `DataInKeySpacePath` has a concrete `ResolvedKeySpacePath`, but I think in the short term I'd rather leave this as-is, and see how it shakes out as the surrounding code is implemented.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: #3397 merged first, and so this just uses the new continuation type.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
ScottDugas: Yes, added.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
ScottDugas: OK, added a test that it does a sane thing if you have only a single element under the keyspace. I tried to parameterize to get a couple other edge cases.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
ScottDugas: Historical. Originally I had `exportAllData` returning `KeyValue`, and at the time it seemed easier to just have this return the raw data. But I went back and encapsulated the raw key entirely, so this now returns a `DataInKeySpacePath`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
```diff
@@ -198,6 +198,18 @@ default Tuple toTuple(@Nonnull FDBRecordContext context) {
...
+    /**
+     * Resolves the given key within this path.
+     * @param context the transaction to lookup any necessary directory layer entries.
+     * @param key a raw key from the database
+     * @return the {@link ResolvedKeySpacePath} corresponding to that key, with a potential remainder.
+     */
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    default CompletableFuture<ResolvedKeySpacePath> toResolvedPathAsync(@Nonnull FDBRecordContext context, byte[] key) {
+        throw new UnsupportedOperationException("toResolvedPath is not supported");
```
ohadzeliger:
```suggestion
        throw new UnsupportedOperationException("toResolvedPathAsync is not supported");
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
```diff
@@ -198,6 +198,18 @@ default Tuple toTuple(@Nonnull FDBRecordContext context) {
...
 
+    /**
+     * Resolves the given key within this path.
+     * @param context the transaction to lookup any necessary directory layer entries.
+     * @param key a raw key from the database
+     * @return the {@link ResolvedKeySpacePath} corresponding to that key, with a potential remainder.
+     */
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    default CompletableFuture<ResolvedKeySpacePath> toResolvedPathAsync(@Nonnull FDBRecordContext context, byte[] key) {
```
ohadzeliger, line 216: Is this doing work similar to `keySpace.resolveFromKey`? What is the difference? Is it because of efficiency (not needing to parse the key from the start every time)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
```diff
@@ -566,4 +578,22 @@ default List<ResolvedKeySpacePath> listSubdirectory(@Nonnull FDBRecordContext co
...
+     * Supports continuation to resume scanning from a previous position.
+     *
+     * @param context the transaction context in which to perform the data export
+     * @param continuation optional continuation from a previous export operation, or null to start from the beginning
+     * @param scanProperties properties controlling how the scan should be performed
+     * @return a RecordCursor that iterates over all KeyValue pairs under this path
+     */
+    @API(API.Status.EXPERIMENTAL)
+    @Nonnull
+    default RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext context,
```
ohadzeliger, line 601: Why is this added to the API at this level? Would it make sense to add this to the `FDBRecordStore`?
Update: Discussed with Scott, and the reason is that this would allow exporting multiple stores with a single call.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePath.java
```diff
@@ -271,4 +271,15 @@ public static void appendValue(StringBuilder sb, Object value) {
...
         }
     }
+
+    /**
+     * Returns a new {@code ResolvedKeySpacePath} that is the same, except with the provided {@link #getRemainder()}.
+     * @param newRemainder a new remainder. This can be {@code null} to remove the remainder entirely.
+     * @return a new {@code ResolvedKeySpacePath} that is the same as this, except with a different {@link #getRemainder()}.
+     */
+    @Nonnull
+    public ResolvedKeySpacePath withRemainder(@Nullable final Tuple newRemainder) {
```
ohadzeliger:
```suggestion
    @VisibleForTesting
    public ResolvedKeySpacePath withRemainder(@Nullable final Tuple newRemainder) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
```diff
@@ -0,0 +1,352 @@
...
+            ResolvedKeySpacePath appLevel;
+            appLevel = assertNameAndValue(resolvedPath, "data", EnvironmentKeySpace.DATA_VALUE);
+            ResolvedKeySpacePath userLevel = assertNameAndDirectoryScopedValue(appLevel, "application", "app1",
+                    appPath, context);
+            ResolvedKeySpacePath envLevel = assertNameAndValue(userLevel, "userid", 100L);
+            assertNull(assertNameAndDirectoryScopedValue(envLevel, keySpace.root().getDirectoryName(),
+                    keySpace.root().getValue(), keySpace.root(), context));
+        }
+    }
+
```
ohadzeliger, line 334: Maybe add a test that fails to resolve a key?
Would it make sense to add test for resolve with NULL value?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
```diff
@@ -0,0 +1,700 @@
...
+                                            String expectedValueContents, String contentMessage) {
+        assertEquals(expectedCount, app1User100Data.size(), expectedCountMessage);
+
+        for (DataInKeySpacePath data : app1User100Data) {
+            String value = Tuple.fromBytes(data.getValue()).getString(0);
+            assertTrue(value.contains(expectedValueContents), contentMessage);
+        }
+    }
+
+}
```
ohadzeliger, line 696: Maybe add a test that fails to export?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
```diff
@@ -0,0 +1,700 @@
...
+    @Test
+    void exportAllDataWithDifferentKeyTypes() {
+        KeySpace root = new KeySpace(
+                new KeySpaceDirectory("mixed", KeyType.STRING, UUID.randomUUID().toString())
+                        .addSubdirectory(new KeySpaceDirectory("strings", KeyType.STRING))
+                        .addSubdirectory(new KeySpaceDirectory("longs", KeyType.LONG))
+                        .addSubdirectory(new KeySpaceDirectory("bytes", KeyType.BYTES))
+                        .addSubdirectory(new KeySpaceDirectory("uuids", KeyType.UUID))
+                        .addSubdirectory(new KeySpaceDirectory("booleans", KeyType.BOOLEAN)));
+
```
ohadzeliger, line 231: Add NULL?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePath.java
ScottDugas: I think this could be used by external clients, but I'm on board with giving limited visibility and exposing as needed. I also reduced it to package-scoped.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
ScottDugas: Yes, that is correct. If a client wants to export a bunch of stores at once, that would be challenging if this were on the store itself. And, if a user wants to export a single store that is still possible.
Also, the store takes a subspace (although IMHO it should take a KeySpacePath), so it would be awkward to deal with `DirectoryLayerDirectory`, as eventually we want to be providing the logical keys, rather than the actual tuples.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ScottDugas: I added a test that fails to resolve (has a different root)

Regarding `NULL`, are you suggesting creating a `DataInKeySpacePath` with `KeyValue(key, null)`, or a path that has a `NULL` type in it?
For the former, I don't think FDB will ever return such a `KeyValue`, and if you have one, you would never be able to insert it, because `set` does not allow a null value. Give that, it probably makes sense to have `DataInKeySpacePath` error if it gets a `null` value.

For the later, I think that should be better tested in `KeySpacePath.toResolvedPathAsync`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
ScottDugas: Actually, it might be worthwhile to have it cover all `KeyType`s. I reworked the test a bit to do that. I think it also makes the overall test a little bit cleaner.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathDataExportTest.java
ScottDugas: I don't think there is a way at this level to cause errors that would be relevant from a business-logic perspective. We could have it export from a closed transaction, or something like that, but that's doesn't seem particularly valuable here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
ScottDugas: It's primarily here because we want to export from a `KeySpacePath`, and that does not have a reference back to the `KeySpace`. The majority of the logic is shared.
It probably has some efficiency gains if this path is on one of the last sub-directories, but it is still resolving this `KeySpacePath` on every call.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePathTest.java
ohadzeliger: I meant the later - NULL type.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/DataInKeySpacePath.java
alecgrieser: Yeah, the main place we run into trouble is when we have a key range like "all keys starting with a tuple", that leads to keys like "original tuple + \x00" (which is Tuple parseable) and "original tuple + \xff" (which is not). That isn't a concern for data in a key space path like this, but it would be if we used part of this for parsing, say, conflict range keys.

fdb-extensions/src/main/java/com/apple/foundationdb/tuple/TupleHelpers.java
```diff
@@ -141,17 +141,17 @@ public static int packedSizeAsTupleItem(Object item) {
...
-     * @param t2 the whole tuple
-     * @return {@code true} if {@code t1} is a prefix of {@code t2}
+     * @param potentialPrefix the potential prefix
+     * @param wholeTuple the whole tuple
+     * @return {@code true} if {@code potentialPrefix} is a prefix of {@code wholeTuple}
      */
-    public static boolean isPrefix(@Nonnull Tuple t1, @Nonnull Tuple t2) {
-        final int len = t1.size();
-        if (t2.size() < len) {
+    public static boolean isPrefix(@Nonnull Tuple potentialPrefix, @Nonnull Tuple wholeTuple) {
```
alecgrieser, line 148: Nice, like the new names 👍

alecgrieser (approved):

PR #3568, opened 2025-09-04
Upload coverage from nightly builds

ScottDugas:
Add new jobs to the nightly flow to run standard tests and mixed-mode tests. This is primarily there so that we can upload coverage to teamscale. 
These new teamscale steps are set to continue-on-error, to make sure they are working correctly before impacting the nightly build.

I also extracted the teamscale upload into a separate action to decrease repetition, this is the biggest risk; if I messed this up, it could impact the post-pr teamscale upload.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -113,6 +113,7 @@ jobs:
           test-reports/fdb-relational-jdbc/
           test-reports/fdb-relational-server/
           test-reports/yaml-tests/
+
     - name: Upload coverage to teamscale
         # temporary until we validate that this is working correctly
         continue-on-error: true
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -42,6 +42,7 @@ jobs:
           partition: 'CI Tests'
           revision: ${{ github.sha }}
           files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
+          teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
 
   mixed-mode-test:
     if: github.repository == 'FoundationDB/fdb-record-layer'
@@ -84,6 +85,7 @@ jobs:
           partition: 'Mixed Mode Tests'
           revision: ${{ github.sha }}
           files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
+          teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
 
   nightly_test:
     if: github.repository == 'FoundationDB/fdb-record-layer'
@@ -129,6 +131,7 @@ jobs:
         partition: 'Nightly Tests'
         revision: ${{ github.sha }}
         files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
+        teamscaleKey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
 
 
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,7 +10,9 @@ jobs:
   #   a) all the tests
   #   b) the mixed-mode tests
   #   c) the nightly tests
-  # At the moment the nightly tests include the regular tests, but we do both to get valuable coverage in teamscale.
+  # The nightly tests (currently) include all the regular tests, but we want to publish to teamscale the coverage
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly.yml
```diff
@@ -41,4 +41,19 @@ jobs:
...
+      continue-on-error: true
+      uses: 'cqse/teamscale-upload-action@8d10b6693b9242420ef2062dbefc36af6e88d587'
+        with:
+          server: 'https://fdb.teamscale.io'
+          project: 'foundationdb-fdb-record-layer'
+          user: 'fdb-record-layer-build'
+          partition: 'Nightly Tests'
+          accesskey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
+          format: 'JACOCO'
+          revision: ${{ github.event.workflow_run.head_sha }}
```
ScottDugas: This is wrong, I think.

.github/workflows/nightly.yml
```diff
@@ -41,4 +41,19 @@ jobs:
...
+          files: |
+            ${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml
```
ScottDugas: I'm not certain this is the correct path

.github/workflows/nightly.yml
```diff
@@ -6,7 +6,79 @@ on:
...
+        with:
+          gradle_args: -PreleaseBuild=true -PpublishBuild=true
+      - name: Upload coverage to teamscale
+        # temporary until we validate that this is working correctly
+        continue-on-error: true
+        uses: ./actions/teamscale-upload
+        with:
+          partition: 'CI Tests'
+          revision: ${{ github.sha }}
+          files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
```
ScottDugas, line 44: I'm not 100% certain this is the correct path, but it's the same path that we use to upload to artifacts, and then download in the teamscale_upload task, so it should be right.

.github/workflows/nightly.yml
```diff
@@ -6,7 +6,79 @@ on:
...
+        # temporary until we validate that this is working correctly
+        continue-on-error: true
+        uses: ./actions/teamscale-upload
+        with:
+          partition: 'Mixed Mode Tests'
+          revision: ${{ github.sha }}
+          files: "${{ github.workspace }}/.out/reports/jacoco/codeCoverageReport/codeCoverageReport.xml"
+
+  nightly_test:
     if: github.repository == 'FoundationDB/fdb-record-layer'
```
alecgrieser, line 91: I think we'd want this on the other two jobs as well, so that the nightly tasks don't run on forks

.github/workflows/nightly.yml
```diff
@@ -6,7 +6,79 @@ on:
   workflow_dispatch:
 
 jobs:
-  gradle:
+  # In parallel, we run:
+  #   a) all the tests
+  #   b) the mixed-mode tests
+  #   c) the nightly tests
+  # At the moment the nightly tests include the regular tests, but we do both to get valuable coverage in teamscale.
```
alecgrieser: I'm not sure I understand the point you're making here. If the `nightly_test` task also runs the regular tests, what coverage data would we miss if we didn't have the new `test` task?

actions/teamscale-upload/action.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: Upload coverage to teamscale
+    # temporary until we validate that this is working correctly
+    continue-on-error: true
+    uses: 'cqse/teamscale-upload-action@8d10b6693b9242420ef2062dbefc36af6e88d587'
+      with:
+        server: 'https://fdb.teamscale.io'
+        project: 'foundationdb-fdb-record-layer'
+        user: 'fdb-record-layer-build'
+        partition: inputs.partition
+        accesskey: ${{ secrets.TEAMSCALE_ACCESS_KEY }}
```
alecgrieser: I believe that composite actions can't access secrets directly: https://stackoverflow.com/questions/70098241/using-secrets-in-composite-actions-github

I think the teamscale access key needs to be an input to the step, and then the callers can set the secret. (There are protections in place to stop the secret value from being logged, even if it's provided as input into the action.)

alecgrieser (changes requested):

.github/workflows/nightly.yml
ScottDugas: ok, I think I improved the wording, hopefully that makes it clear.

actions/teamscale-upload/action.yml
ScottDugas: Good catch, I was unaware of that.

PR #3569, opened 2025-09-05
Pass through Options to different kinds of connections.

MMcM:


MMcM:
Closing in favor of #3557, which motivates the change.

Issue #3574, opened 2025-09-08
Support core layer tests that interact with multiple clusters


ScottDugas:
To better support testing of copying from one cluster to another we need some core-tests that can interact with multiple clusters.

PR #3575, opened 2025-09-08
Add support for testing against multiple FDB clusters

ScottDugas:
This changes the way we specify cluster files in tests. It replaces the `fdb-environment.properties` with `fdb-environment.yaml`. The new format allows for specifying multiple cluster files.

The majority of tests now randomly chose a file at the start of the test from the list of the available ones.
The following sub-projects set the environment variable in the gradle task because getting the project under test to support custom cluster files seemed like a bigger task:
- fdb-relational-cli
- fdb-relational-jdbc

Our CI builds also, now, setup two FDB clusters, and create an `fdb-environment.properties`.

I added one new "test" to validate this all works. It connects to two different clusters. It `assume`s that there are at least two clusters. At some point in the future we will want to make that a stronger contract to ensure that a bug in our CI setup doesn't cause us to not run all the multi-cluster tests.

I created Draft PR: https://github.com/FoundationDB/fdb-record-layer/pull/3623 to ensure that everything still works if you don't specify `fdb-environment.yaml`
I also created this PR off the same version of main to confirm the test counts didn't change: https://github.com/FoundationDB/fdb-record-layer/pull/3621 Apparently we do merge main to do the tests so PR #3621 saw fewer tests than this branch when it was last run. I re-ran and they may be the same depending on whether new tests were merged. I validated some of the differences, and they look to be expected.

Resolves: #3574

### DIFF
--- a/fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/RelationalServer.java
+++ b/fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/RelationalServer.java
@@ -95,8 +95,8 @@ public class RelationalServer implements Closeable {
         this.clusterFile = clusterFile;
     }
 
-    public RelationalServer(int grpcPort, int httpPort) {
-        this(grpcPort, httpPort, CollectorRegistry.defaultRegistry, null);
+    public RelationalServer(int grpcPort, int httpPort, String clusterFile) {
+        this(grpcPort, httpPort, CollectorRegistry.defaultRegistry, clusterFile);
     }
 
     @Override
@@ -253,6 +253,10 @@ public class RelationalServer implements Closeable {
         Option httpPortOption = Option.builder().option("p").longOpt("httpPort").hasArg(true).type(Number.class)
                 .desc("Port for HTTP to listen on; default=" + DEFAULT_HTTP_PORT + ".").build();
         options.addOption(httpPortOption);
+        Option clusterFileOption = Option.builder().option("c").longOpt("clusterFile").optionalArg(true)
+                .hasArg(true).type(String.class)
+                .desc("Path to the cluster file; default=null.").build();
+        options.addOption(clusterFileOption);
         CommandLineParser parser = new DefaultParser();
         CommandLine cli = null;
         try {
@@ -267,12 +271,22 @@ public class RelationalServer implements Closeable {
             return;
         }
 
+        final String clusterFile;
+        if (cli.hasOption(clusterFileOption)) {
+            clusterFile = cli.getOptionValue(clusterFileOption);
+        } else {
+            clusterFile = System.getenv("FDB_CLUSTER_FILE");
+        }
+
         if (logger.isInfoEnabled()) {
             logger.info("FDB_CLUSTER_FILE: " + System.getenv("FDB_CLUSTER_FILE"));
             logger.info("DYLD_LIBRARY_PATH: " + System.getenv("DYLD_LIBRARY_PATH"));
         }
         int grpcPort = getPort(cli, grpcPortOption, GrpcConstants.DEFAULT_SERVER_PORT);
         int httpPort = getPort(cli, httpPortOption, DEFAULT_HTTP_PORT);
-        new RelationalServer(grpcPort, httpPort).start().awaitTermination();
+        try (RelationalServer relationalServer = new RelationalServer(grpcPort, httpPort, clusterFile)) {
+            relationalServer.start();
+            relationalServer.awaitTermination();
+        }
     }
 }
--- a/fdb-relational-server/src/test/java/com/apple/foundationdb/relational/server/RelationalServerTest.java
+++ b/fdb-relational-server/src/test/java/com/apple/foundationdb/relational/server/RelationalServerTest.java
@@ -20,13 +20,15 @@
 
 package com.apple.foundationdb.relational.server;
 
+import com.apple.foundationdb.record.RecordCoreInterruptedException;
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/BlockingInAsyncDetectionTest.java
```diff
@@ -57,7 +58,7 @@ void testBlockingInAsyncException() {
         // Make sure that we aren't holding on to previously created databases
         factory.clear();
 
-        FDBDatabase database = factory.getDatabase();
+        FDBDatabase database = factory.getDatabase(FDBTestEnvironment.randomClusterFile());
```
ScottDugas, line 61: This test could probably just use `dbExtension.getDatabase()`, but I didn't want to change too much in this PR. And you have to take care to make sure you're not reusing a database created in the setup.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/indexes/RankIndexTest.java
```diff
@@ -745,34 +746,27 @@ void containsNullScore() throws Exception {
...
-                                        scanPlan().where(scanComparisons(unbounded())))
-                                        .where(recordTypes(containsAll(ImmutableSet.of("BasicRankedRecord")))))
-                                .where(queryComponents(exactly(equalsObject(filter)))));
-                recordStore.executeQuery(plan)
-                        .map(rec -> TestRecordsRankProto.BasicRankedRecord.newBuilder().mergeFrom(rec.getRecord()).build()).asList().get();
-            } catch (ExecutionException e) {
-                throw e.getCause();
-            }
-        });
+    void writeOnlyRankQuery() throws Exception {
```
ScottDugas, line 749: This test was failing, and reworking it to isolate the assertThrows helped me debug what was wrong, and I felt like it would be good to keep it.

fdb-relational-core/fdb-relational-core.gradle
```diff
@@ -47,11 +47,13 @@ dependencies {
     testImplementation(libs.bundles.test.runtime)
     testCompileOnly(libs.bundles.test.compileOnly)
 
+    testImplementation project(path: ':fdb-extensions', configuration: 'tests')
```
ScottDugas: I thought about putting the new class in a `testFixtures`, but I couldn't get that to work without putting all the classes that anyone already depends on via this approach in the testFixture, and thought that would be better as a followup.

fdb-relational-server/src/main/java/com/apple/foundationdb/relational/server/RelationalServer.java
```diff
@@ -131,7 +138,7 @@ RelationalServer start() throws IOException {
         // Create access to backing database.
         // TODO: Make this multi-query/-tenant/-database!
         try {
-            frl = new FRL();
+            frl = new FRL(com.apple.foundationdb.relational.api.Options.NONE, clusterFile);
```
ScottDugas, line 137: We also have `Options` from apache cli, so this needs to be fully qualified (or statically imported)

yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/YamlConnection.java
```diff
@@ -108,4 +108,11 @@ public interface YamlConnection extends AutoCloseable {
     SemanticVersion getInitialVersion();
 
     <T> T executeTransactionally(SQLFunction<YamlConnection, T> transactionalWork) throws SQLException, RelationalException;
+
+    /**
+     * The Cluster File that this connection is connected to, so that other connections can point to the same
```
ScottDugas, line 113: I'm not super committed to this API, but at some point we'll want multi-cluster testing in yaml-tests, and I think that's a better time to really think about it.

fdb-test-utils/src/main/java/com/apple/foundationdb/test/FDBTestEnvironment.java
```diff
@@ -0,0 +1,86 @@
...
+
+import javax.annotation.Nullable;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
```
ScottDugas, line 34: I thought about putting this in `testFixtures`, but since other projects already depend on the tests configuration, doing so would require moving all the other classes into a `testFixtures`, and I figured that would be better as a separate PR.

fdb-relational-core/fdb-relational-core.gradle
ScottDugas: Actually, I think it would be better to move everything to a `testFixtures`. I'll do that in a separate PR

fdb-test-utils/src/main/java/com/apple/foundationdb/test/FDBTestEnvironment.java
ScottDugas: I made a change in main to move them into a test fixture, and then went to test publishing of `testFixtures` so that `yaml-tests` could be used by clients for their tests, but it looks like we can't publish `testFixtures`: https://github.com/FoundationDB/fdb-record-layer/commit/e9db5c04ba449c64d846ca389879ffae56c6d730#diff-9515853bb3a0c957c759e9c453a9e17941f4333751715e75d74efed681fadac0R60 I'm going to look into creating a `test-utils` library....

PR #3577, opened 2025-09-08
Implementation of grouping size aggregator

### LINKED ISSUE #3576: Create an aggregate size calculator that can group size by subspace
The existing `SizeStatisticsCollectorCursor` returns a single result with the entire size of the subspace calculated.
For a more nuanced approach, a calculator that can break the size into sub-subspaces is needed (specifically, it can help with troubleshooting storage usage per store components).

ohadzeliger:
This implementation is creating a flavor of the `SizeStatisticsCollectorCursor` that can return sizes aggregated by sub-subspaces.
This relies on the original implementation but had to go into a separate class as the generic type of the cursor has changed.
Eventually, the original cursor can be replaced by this one with the aggregation size being "0".
For backwards compatibility, the original class was marked as "deprecated". The original `SizeStatisticsResults` was marked as deprecated and a new (non-embedded version) of it was added.

Initially, I thought that we can make a backwards-compatible change to the `SizeStatisticsCollectorCursor` class, but that proved more complicated with the changes to the result type, so I decided to go with a new implementation instead.
 
Resolves #3576


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingContinuation.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * A continuation for the {@link SizeStatisticsGroupingCursor}.
+ */
+class SizeStatisticsGroupingContinuation implements RecordCursorContinuation {
+    /**
+     * A continuation for the case where we have one last result to send, after which the cursor would be done.
+      */
+    public static final SizeStatisticsGroupingContinuation LAST_RESULT_CONTINUATION = new SizeStatisticsGroupingContinuation();
```
alecgrieser, line 41: Is the serialized form of the `LAST_RESULT_CONTINUATION` the empty byte string? That's generally a bad idea, as it can cause confusion when we try and compose cursors together, though that's probably rarer for a class like this than many other cursors. It's possible we should add a Boolean to the Protobuf like `is_last` that this can set

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsCollectorCursor.java
```diff
@@ -65,8 +65,9 @@
  * for all keys and values in the specified subspace. Any cursor results emitted prior to that are a manifestation of
  * hitting an execution limit while scanning the subspace. These results provide a continuation that can be used
  * to resume aggregating the remaining keys and values of the subspace.
+ * Deprecated, use the {@link SizeStatisticsGroupingCursor} instead.
  */
-@API(API.Status.EXPERIMENTAL)
+@API(API.Status.DEPRECATED)
```
alecgrieser, line 70: I suppose that given that this is `EXPERIMENTAL`, we could remove it instead of deprecating it, though we could also keep it for a bit longer to give people a chance to transition over.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -0,0 +1,353 @@
...
+ * the
+ * constituents of the subspace key values.
+ * The cursor performs a full scan of the subspace backing whatever it is collecting statistics on. It
+ * should therefore be run relatively sparingly, and it should not be expected to finish within a single
+ * transaction.
+ * Note that this class makes no attempt at keeping its results transactional. As a result, if this is run on an
+ * actively mutating data set, there are no guarantees that the values it returns were ever actually true for any
+ * version in the transaction history. However, as long as the data set is not too volatile, it should produce an
+ * approximate answer
+ * for the statistics it yields.
```
alecgrieser: The newlines of this comment are a bit strange. I think there are tools that will re-justify this stuff, though I'm actually not super familiar with them

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -0,0 +1,353 @@
...
+ * <p>
+ * A note about security: scanning subspaces may reveal sensitive information. Specifically, in the case of value
+ * indexes,
+ * the constituents of the key may reveal the values indexed by the index. DO NOT use this tool without proper
+ * authentication and
+ * authorization in place that ensures data integrity.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class SizeStatisticsGroupingCursor implements RecordCursor<SizeStatisticsGroupedResults> {
+    private static final Tuple EMPTY_TUPLE = new Tuple();
```
alecgrieser: You could use `TupleHelpers.EMPTY` for this, I think

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -0,0 +1,353 @@
...
+        }
+        // instantiate the underlying KV cursor, aggregates as much of its input as it can provide, and close it
+        return subspaceProvider.getSubspaceAsync(context).thenCompose(subspace -> {
+            KeyValueCursor kvCursor = KeyValueCursor.Builder
+                    .withSubspace(subspace)
+                    .setContext(context)
+                    .setContinuation(kvCursorContinuation)
+                    .setScanProperties(scanProperties)
+                    .build();
+            return AsyncUtil.whileTrue(() ->
```
alecgrieser: This needs to supply an `Executor` to `whileTrue`. You should be able to use the result of `getExecutor()`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -0,0 +1,353 @@
...
+    @Nonnull
+    @Override
+    public CompletableFuture<RecordCursorResult<SizeStatisticsGroupedResults>> onNext() {
+        // if this cursor instance has previously hit a limit then keep returning the same result
+        if (nextStatsResult != null && !nextStatsResult.hasNext()) {
+            return CompletableFuture.completedFuture(nextStatsResult);
+        }
+        // instantiate the underlying KV cursor, aggregates as much of its input as it can provide, and close it
+        return subspaceProvider.getSubspaceAsync(context).thenCompose(subspace -> {
+            KeyValueCursor kvCursor = KeyValueCursor.Builder
```
alecgrieser: Is there a reason why this constructs the `KeyValueCursor` each time? I would have thought that we could do something like:

1. Store as a member variable a `LazyCursor` that handles getting the subspace and then constructing the `KeyValueCursor`
2. In this method, we call `handleOneItem` in a loop by pulling from that cursor

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -0,0 +1,353 @@
...
+    public static SizeStatisticsGroupingCursor ofSubspace(@Nonnull Subspace subspace, @Nonnull FDBRecordContext context,
+                                                          @Nonnull ScanProperties scanProperties, @Nullable byte[] continuation, int aggregationDepth) {
+        return new SizeStatisticsGroupingCursor(new SubspaceProviderBySubspace(subspace), context, scanProperties, continuation, aggregationDepth);
+    }
+
+    @Nonnull
+    @Override
+    public CompletableFuture<RecordCursorResult<SizeStatisticsGroupedResults>> onNext() {
+        // if this cursor instance has previously hit a limit then keep returning the same result
+        if (nextStatsResult != null && !nextStatsResult.hasNext()) {
```
alecgrieser: Should this also check to see if the `nextStatsResult` has a last-result continuation, and if so, update the `nextStatsResult` to `RecordCursorResult.exhausted()` before going to the `whileTrue` loop?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
```diff
@@ -0,0 +1,298 @@
...
+            stats.updateStatistics(new KeyValue("ab".getBytes(), "cd".getBytes())); // size 4
+        }
+        
+        double median = stats.getMedian();
+        double p90 = stats.getP90();
+        double p95 = stats.getP95();
+        
+        Assertions.assertThat(median).isEqualTo(stats.getProportion(0.5));
+        Assertions.assertThat(p90).isEqualTo(stats.getProportion(0.9));
+        Assertions.assertThat(p95).isEqualTo(stats.getProportion(0.95));
```
alecgrieser: These tests seem a tad tautological, as it's basically showing (at best) that median is syntactic sugar for `getProportion(0.5)`. Should there be a variant that asserts on the absolute values returned here? We may also need to test this out on a non-uniform distribution to get signal here

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Unit tests for {@link SizeStatisticsResults}.
+ */
+public class SizeStatisticsResultsTest {
+
+    @Test
+    public void emptyStatistics() {
+        SizeStatisticsResults stats = new SizeStatisticsResults();
+        
+        Assertions.assertThat(stats.getKeyCount()).isZero();
```
alecgrieser: The codebase is somewhat inconsistent about this, but we should be able to statically import `Assertions.assertThat` to make this a little bit less verbose

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
```diff
@@ -0,0 +1,479 @@
...
+            Assertions.assertThat(continuationResult.hasNext()).isTrue();
+            Assertions.assertThat(continuationResult.getContinuation().isEnd()).isFalse();
+            // scan again, to get the continuation to the end
+            RecordCursorResult<SizeStatisticsGroupedResults> finalResult = getStoreSizeResult(recordStore, ScanProperties.FORWARD_SCAN, continuationResult.getContinuation().toBytes(), 0);
+            Assertions.assertThat(finalResult.hasNext()).isFalse();
+            Assertions.assertThat(finalResult.getContinuation().isEnd()).isTrue();
+
+            final RecordCursorResult<SizeStatisticsGroupedResults> controlResult = getStoreSizeResult(recordStore);
+
+            Assertions.assertThat(continuationResult.get().getStats()).usingRecursiveComparison().isEqualTo(controlResult.get().getStats());
```
alecgrieser: Do we need the `usingRecursiveComparison()` thing because `StatisticsResult` doesn't extend `equals`? Should it?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
```diff
@@ -0,0 +1,479 @@
...
+        return result;
+    }
+
+    private RecordCursorResult<SizeStatisticsGroupedResults> getStoreSizeResult(@Nonnull FDBRecordStore store) {
+        return getStoreSizeResult(store, ScanProperties.FORWARD_SCAN, null, 0);
+    }
+
+    private RecordCursorResult<SizeStatisticsGroupedResults> getStoreSizeResult(@Nonnull FDBRecordStore store, final ScanProperties scanProperties, byte[] continuation, int aggregationDepth) {
+        try (SizeStatisticsGroupingCursor statsCursor = SizeStatisticsGroupingCursor.ofStore(store, store.getContext(), scanProperties, continuation, aggregationDepth)) {
+            return statsCursor.getNext();
```
alecgrieser: I believe this approach means that we're only testing the case where the cursor is stopped and resumed every time it returns data, so we're not getting coverage of what happens if we read multiple groups with the same cursor. The old cursor had a constraint that it would never return more than one item anyway, but it seems totally plausible that you'd use one cursor to read multiple groups with this new cursor

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingContinuation.java
ohadzeliger: Sure, added a required boolean to be explicit about that property.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsCollectorCursor.java
ohadzeliger: I'd rather not force adopters to change their code and will prefer to remove later.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: This is a consequence of  IDE "reformatting" long lines unesthetically. Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: That was a result of matching the lifecycle of the previous implementation, which was inadequate. Fixed:
- Used Lazy cursor to initialize the inner cursor in the constructor
- Close the inner cursor upon `close()`
- Reuse the inner cursor on calls to `onNext`
- Handle more termination conditions in `onNext`
- Tests are now using `asList`, that will ensure we iterate over the same cursor instance

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: Yes, done. There is another check like that in the constructor to catch the case where last result came in the continuation, and this is similar for when we arrive at it while iterating over the cursor.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
ohadzeliger: These were added to cover a test gap but indeed, all these methods are doing is syntactic sugar. I am not sure how we can provide more meaningful tests for them...

I added one more random test (with the hopes it will not become flaky), to test the results of proportion calculation.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
ohadzeliger: Done.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
ohadzeliger: Yes. I'd rather add `equals` and `hashCode` in the case where the class has a definite identity, or when we need to add it to a `Set` or as a key to a `Map`. (The former is such that we can explicitly declare the unique identifier of the
 instance and the latter is a necessity of the protocol). These data holders don't meet these criteria so I think I'd rather leave them as is, if OK with you

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
ohadzeliger: Right. That was fixed as part of the lifecycle fix, and now uses `asList` to iterate over an existing instance

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -83,9 +83,11 @@ message SizeStatisticsContinuation {
 }
 
 message SizeStatisticsGroupingContinuation {
-    optional bytes underlyingContinuation = 1;
-    optional SizeStatisticsPartialResults partialResults = 2;
-    optional bytes currentGroupingKey = 3;
+    // A marker for when the continuation is used to indicate that the cursor is done and this is the last result
+    required bool lastResultContinuation = 1;
```
alecgrieser:
```suggestion
    optional bool lastResultContinuation = 1;
```

Protobuf is generally trying to ban `required` fields, and in this case, I think it would be fine to rely on this being set to false by default rather than requiring it be set.

fdb-record-layer-core/src/main/proto/record_cursor.proto
```diff
@@ -83,9 +83,11 @@ message SizeStatisticsContinuation {
...
+    optional bytes underlyingContinuation =2;
```
alecgrieser:
```suggestion
    optional bytes underlyingContinuation = 2;
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
```diff
@@ -147,6 +140,16 @@ private SizeStatisticsGroupingCursor(@Nonnull SubspaceProvider subspaceProvider,
                         .addLogInfo("raw_bytes", ByteArrayUtil2.loggable(continuation));
             }
         }
+        subspaceFuture = subspaceProvider.getSubspaceAsync(context);
```
alecgrieser: It would be nice to avoid background work in the constructor, as people aren't expecting that side effect. Ideally, this would either delay the creation of the inner cursor to the first call of `onNext()` or the private constructor here would take an already initialized cursor, and then a static initializer would be the one to kick of subspace resolution and create the `LazyCursor`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
alecgrieser: I was thinking that you might have a test where we knew what the median was supposed to be, and then we validated that that's the median we got. A randomized test probably gets us closer to that, though there's a lot of variability there

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
alecgrieser: I'm okay with leaving it as is for now. I generally favor adding `equals` and `hashCode` in more cases, though. Like, I'd probably say that any data class should have one just because of the number of times that we've run into problems because a class has unexpectedly not implemented it

alecgrieser (approved):
LGTM. I left a few small things that you could address at your discretion

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsResultsTest.java
ohadzeliger: Deleted that test and added more assertions to proportionCalculation that has fixed values. Hope that makes more sense

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursorTest.java
ohadzeliger: Thanks. I'll leave that aside, especially because I don't know what the correct implementation should be due here and for the `GroupedResults`.

fdb-record-layer-core/src/main/proto/record_cursor.proto
ohadzeliger: Sure, done and assume false when getting

fdb-record-layer-core/src/main/proto/record_cursor.proto
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingCursor.java
ohadzeliger: Sure. Done. The thought was to preempt the scheduling of the work and get it done concurrently, but I understand the reasoning

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/cursors/SizeStatisticsGroupingContinuation.java
```diff
@@ -77,7 +77,9 @@ private SizeStatisticsGroupingContinuation() {
      * @return TRUE if the protobuf marks a continuation with last result
      */
     public static boolean isLastResultContinuation(final RecordCursorProto.SizeStatisticsGroupingContinuation statsContinuation) {
-        return statsContinuation.getLastResultContinuation();
+        // Default to false
+        return (statsContinuation.hasLastResultContinuation() &&
+                statsContinuation.getLastResultContinuation());
```
alecgrieser: This isn't wrong, but `getLastResultContinuation()` should return `false` (not, say, `null`) if the field is unset, so this should be logically equivalent to:

```suggestion
        // Default to false
        return statsContinuation.getLastResultContinuation());
```

ohadzeliger: Sure, done.

PR #3578, opened 2025-09-09
Keyspace import

ScottDugas:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -316,6 +316,43 @@ public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
                     1);
     }
 
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> importData(@Nonnull FDBRecordContext context,
```
ohadzeliger, line 322: Is the assumption here that resource control be handled externally? Meaning, if the given stream of data is too large, someone else is going to trim it and retry? In that sense, would it make sense to provide a single import method (for a single item) and return the future, as there seems no efficiency by working on the whole collection?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -316,6 +316,43 @@ public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
...
+                    }
+
+                    // Reconstruct the key using the path and remainder
+                    Tuple keyTuple = itemPathTuple;
+                    if (dataItem.getRemainder() != null) {
+                        keyTuple = keyTuple.addAll(dataItem.getRemainder());
+                    }
+
+                    // Store the data
+                    byte[] keyBytes = keyTuple.pack();
```
ohadzeliger: Should we add some fdb timer metrics for future use (imported_count)?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/test/FDBDatabaseExtension.java
```diff
@@ -161,6 +165,18 @@ public FDBDatabase getDatabase(@Nullable String clusterFile) {
                 });
     }
 
+    /**
+     * Return a random subset of the databases available.
+     * @param count the number of desired databases
+     * @return a random subset of the databases available. This may be less than {@code count} if there aren't that many
+     * databases available.
+     */
+    public List<FDBDatabase> getDatabases(int count) {
```
ohadzeliger:
```suggestion
    public List<FDBDatabase> getRandomDatabasesSubset(int count) {
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+@Tag(Tags.RequiresFDB)
+class KeySpacePathImportDataTest {
+    @RegisterExtension
+    final FDBDatabaseExtension dbExtension = new FDBDatabaseExtension();
+    private FDBDatabase database;
+    private List<FDBDatabase> databases;
+
+    @BeforeEach
+    void setUp() {
+        databases = dbExtension.getDatabases(2);
```
ohadzeliger: Maybe `sourceDb` and `targetDb`, and some logic (like `assume`) around whether target DB is null?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+        byte[] key = path.toSubspace(context).pack(remainder);
+        return Tuple.fromBytes(context.ensureActive().get(key).join());
+    }
+
+    private void copyData(final KeySpacePath sourcePath, KeySpacePath destinationPath) {
+        // Export the data
+        final List<DataInKeySpacePath> exportedData = getExportedData(sourcePath);
+
+        if (databases.size() > 1) {
+            database = databases.get(1);
```
ohadzeliger: Was this intended to change the class' field value?
(I guess, it is, as after the copy all operations would be running on the "target" DB, but it is a little obscure to the reader)

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+    private static Tuple getTupleFromPath(final FDBRecordContext context, final KeySpacePath path, final Tuple remainder) {
+        byte[] key = path.toSubspace(context).pack(remainder);
+        return Tuple.fromBytes(context.ensureActive().get(key).join());
+    }
+
+    private void copyData(final KeySpacePath sourcePath, KeySpacePath destinationPath) {
+        // Export the data
+        final List<DataInKeySpacePath> exportedData = getExportedData(sourcePath);
+
+        if (databases.size() > 1) {
```
ohadzeliger: I feel this allowance for any configuration may introduce confusion and inconsistency between the tests that are actually running in each environment. Wouldn't it be more predictable if there were specific sets of tests targeting single- and double- DB scenarios?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -316,6 +316,43 @@ public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
...
 
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> importData(@Nonnull FDBRecordContext context,
+                                              @Nonnull Iterable<DataInKeySpacePath> dataToImport) {
+        return toTupleAsync(context).thenCompose(targetTuple -> {
+            List<CompletableFuture<Void>> importFutures = new ArrayList<>();
+
+            for (DataInKeySpacePath dataItem : dataToImport) {
+                CompletableFuture<Void> importFuture = dataItem.getPath().toTupleAsync(context).thenCompose(itemPathTuple -> {
```
ohadzeliger: I'm assuming that the directory layer implementation of `Directory` is creating a new entry in case the requested value is not found, right? This is done in the same transaction and is rolled back in case of failure, right? Can there be cases where multiple entries are created for the same resolved value? Do we care?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+            setInPath(emp1Path, context, Tuple.from("salary"), 75000);
+            setInPath(emp1Path, context, Tuple.from("info", 42, true, "complex"), "Complex Test");
+
+            byte[] binaryKey = emp1Path.toSubspace(context).pack(Tuple.from("binary_metadata"));
+            context.ensureActive().set(binaryKey, "binary_test_data".getBytes());
+
+            context.commit();
+        }
+
+        copyData(root.path("company"), root.path("company"));
```
ohadzeliger: Will there be cases where we have to import into a different root? Say where we export from "/local/company/..." and import into "/remote/company/..."?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+            Assertions.assertThatThrownBy(() -> path.importData(context, invalidData).join())
+                    .isInstanceOf(CompletionException.class)
+                    .hasCauseInstanceOf(RecordCoreIllegalImportDataException.class);
+        }
+    }
+
+    private void clearPath(final FDBDatabase database, final KeySpacePath path) {
+        try (FDBRecordContext context = database.openContext()) {
+            Transaction tr = context.ensureActive();
+            tr.clear(path.toSubspace(context).range());
```
ohadzeliger, line 523: This would not clear the directory layer, right?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+        // Should verify that DirectoryLayer mappings work correctly during import
+        final String tenantUuid = UUID.randomUUID().toString();
+        KeySpace root = new KeySpace(
+                new DirectoryLayerDirectory("tenant", tenantUuid)
+                        .addSubdirectory(new KeySpaceDirectory("user_id", KeyType.LONG)));
+
+        final KeySpacePath dataPath = root.path("tenant").add("user_id", 999L);
+        setSingleKey(dataPath, Tuple.from("data"), Tuple.from("directory_test"));
+
+        copyData(root.path("tenant"), root.path("tenant"));
```
ohadzeliger: Would it make sense to add assertions for the resolved exported value and the way it was resolved by directory layer?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,419 @@
...
+        );
+
+        try (FDBRecordContext context = database.openContext()) {
+            root.path("root").importData(context, duplicateData).join();
+            context.commit();
+        }
+
+        verifySingleKey(dataPath, Tuple.from("item"), Tuple.from("final_value"));
+    }
+
```
ohadzeliger, line 443: Additional potential tests:
- Large data (or any out of band error) during import
- Import into partial path (no leaves in import data)  + some remainders
- import where data is of the wrong type

ohadzeliger (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: Yes, the idea would be that the calling process would limit `dataToImport`, and decrease that limit in response to failures.
I hadn't really considered the possibility of limiting in response to the number that were imported, our existing code does have some dynamic response (e.g. `ThrottlingRetryingIterator` and `IndexingThrottle`), so it is probably a good idea to support that here. It could do something like add the number imported to the exception, but that doesn't seem great.
There is some performance advantage to doing it as an iterable, namely `toTupleAsync` may have a cost if it is `DirectoryLayerDirectory`. That should be minimal, but it could go away completely by putting it on `ResolvedKeySpacePath`.
It's unfortunate that would put `exportAllData` and `importData` on different classes.

I can try moving this if you think it's worthwhile at this point.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: Yes, the directory layer will create any entries that don't exist, but I should ensure there are tests of that.
The `DirectoryLayer` creates entries in separate transactions, borrowing the read version. I believe it does this to:
1. Isolate transaction conflict risk
2. Once it is created it is cached

The general assumption is that anything that you're interacting with in the DirectoryLayer should almost always already exist, and most likely already be in the cache.


And the DirectoryLayer ensures that both the logical values (`String`) and resolved values (`Long`) are unique.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: Yeah, I think a timer around `importFuture` makes sense.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: Given that running with multiple clusters was only recently added, I felt it would be better to have these tests run re-using the same cluster, rather than have them not run if a developer doesn't have two FDB clusters.
(Also, I wrote these tests before that functionality existed).

Changing it to be sourceDB and destDB fields seems reasonable enough. I think I would have them both be `Nonnull` and have special logic to clear if they are the same.

It would be good to have a couple of these tests parameterized to copy within a cluster vs across clusters. I'll make sure to do that. At the very least copying back to where it was, clearing in-between, copying to a different path, and copying with DirectoryLayer.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: Yes, but it is the responsibility of the caller to convert the path.
In my draft for serialization I cover this: https://github.com/FoundationDB/fdb-record-layer/pull/3747/files#diff-9422726206a402dab506820d748c97cd0b4a1b76c5f9ed8ff0f60c50a8bad8b0R342 
but it's not really a test that makes sense at this point, because the `DataInKeySpacePath` should be already referring to `/remote/company` by the time it gets to import.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: That is implicitly done by the `verifySingleKey`, which takes the `dataPath`. If `copyData` didn't resolve the path correctly, it wouldn't be found in `verifySingleKey`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: There is a test where the directory-layer entry won't exist: `importDataWithDirectoryLayer` is using a uuid for the string value, so when copying between clusters it won't exist on the destination.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: 1. Yes, a test of more data than can be inserted into a single transaction would make sense, but not if I move it to `ResolvedKeySpacePath` and just have it take a single `DataInKeySpacePath`.
2. I'm not sure what you mean by a partial path. 
3. If by `data` you mean the value, there is no validation, and it is not `KeySpacePath`s responsibility to know what is in the data. If you mean the object in the path, that should be validated above this call, and should be trust-worthy by the time you get a `DataInKeySpacePath`. Ideally this would be validated when you create the `KeySpacePath`, but it is covered in the serialization work, and I explain a bit more on the situation there: https://github.com/FoundationDB/fdb-record-layer/pull/3747/files#diff-15120b2e222e6bb7c2647b670f676b719cce8602e410487604bc87e9ea30a3b0R179

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: I think I effectively answered this above: https://github.com/FoundationDB/fdb-record-layer/pull/3578/files/afc98fb37a68f1590505af5110cef462bd27f004#r2524081596

But I think most of these tests shouldn't care whether there are two clusters or not, but changing a few of them to run in both single- and double- DB scenarios, and skip if the double-DB run if there aren't two databases makes sense.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: Yes, it was intended, but I can see how it would be confusing.
When I do the refactoring to have `sourceDB` and `targetDB`, I'll see how that shakes out. I think the only code interacting with the targetDB would be `verifySingleKey` so I'll make sure to check that.
At least one thing I want to ensure is clear is that we aren't accidentally reading back from the source.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: Correct

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/test/FDBDatabaseExtension.java
ScottDugas: It's possible it will make less sense to exist when I remove the `databases` field, but if I keep this method around, I'll rename it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: I forgot/misread one other advantage: This imports all the data concurrently.
The set should be incredibly efficient, so the primary thing you are gaining in performance is resolving the path for each data item.
If a client wants to do imports one-at-a-time they can always provide `dataToImport` with just one element, and call it multiple times.

I feel like that *is* a good enough reason to stick with the current implementation, at least for now.
That being said, it may make sense to use a `MapPipelinedCursor` so that we are only resolving X at a time.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: Given that `importData` works with a `DataInKeySpacePath`, I actually think copying to a different path doesn't make sense, because the `DataInKeySpacePath` needs to be scoped to this path.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ohadzeliger: By the second bullet I meant importing into the middle of the path, as when you have a path defined for `/company/employee/id/profile` and the import only has `/company/employee`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/FDBStoreTimer.java
```diff
@@ -182,6 +182,8 @@ public enum Events implements StoreTimer.Event {
         RANGE_SET_CONTAINS("range set contains key"),
         /** The amount of time checking if a {@link com.google.common.collect.RangeSet} is empty. */
         RANGE_SET_IS_EMPTY("range set is empty"),
+        /** The amount of time importing a single KeyValue into a path. */
+        IMPORT_DATA("import KeyValue"),
```
ohadzeliger, line 186: Is this being used?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ohadzeliger: I believe the new value is not used.

ohadzeliger (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePath.java
```diff
@@ -410,4 +410,26 @@ default RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
                                                            @Nonnull ScanProperties scanProperties) {
         throw new UnsupportedOperationException("exportAllData is not supported");
     }
+
+    /**
+     * Imports the provided data exported via {@link #exportAllData} into this {@code KeySpacePath}.
+     * <p>
+     * This will validate that any data provided in {@code dataToImport} has a path that should be in this path,
+     * or one of the sub-directories, if not, the future will complete exceptionally with
```
alecgrieser:
```suggestion
     * This will validate that any data provided in {@code dataToImport} has a path that should be in this path
     * or one of the sub-directories. If not, the future will complete exceptionally with a
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -316,6 +317,51 @@ public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
...
+            // are happening.
+            // DirectoryLayer operations are done in a separate transaction for two reasons:
+            // 1. To reduce conflicts
+            // 2. So that it can immediately be cached without having to do it in a post-commit hook, which can make
+            //    things complicated
+            // So if we just spun off a future for every data item, they would conflict like crazy, and retrying all
+            // of those conflicts would cause this future to take way longer than if we pipeline.
+            // This shouldn't make much of a difference in the general case because almost all the directory layer
+            // lookups should be from cache.
+            final RecordCursor<Void> insertionWork = RecordCursor.fromIterator(dataToImport.iterator())
```
alecgrieser:
```suggestion
            final RecordCursor<Void> insertionWork = RecordCursor.fromIterator(context.getExecutor(), dataToImport.iterator())
```

The `fromIterator` cursor is also a bit of an odd duck, in that it doesn't _really_ support continuations, but you'd need to do more refactoring work to make this actually support continuations. Things like having this function take a `RecordCursor<DataInKeySpacePath>` and having it return a continuation. But I suppose an API where the caller is the one that ensures that the data is contained to a single transaction's effort is also workable.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,570 @@
...
+        KeySpace root = new KeySpace(
+                new KeySpaceDirectory("root", KeyType.STRING, rootUuid)
+                        .addSubdirectory(useDirectoryLayer
+                                         ? new DirectoryLayerDirectory("data")
+                                         : new KeySpaceDirectory("data", KeyType.STRING)));
+
+        final KeySpacePath rootPath = root.path("root");
+
+        final DataGenerator data = new DataGenerator(rootPath, 10_000, manyPaths);
+        importData(destinationDatabase, rootPath, data);
```
alecgrieser, line 437: How confident are we that this is safe to do in a single transaction? I guess you haven't had trouble so far locally?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
```diff
@@ -0,0 +1,570 @@
...
+        sourceDatabase = databases.get(0);
+        if (databases.size() > 1) {
+            destinationDatabase = databases.get(1);
+        } else {
+            destinationDatabase = sourceDatabase;
+        }
+    }
+
+    @ParameterizedTest
+    @EnumSource(CopyConfig.class)
```
alecgrieser, line 80: Using the `CopyConfig` parameter like this is a little weird, but I can see how it's working. The alternative would, I think, be to introduce a new `Extension` that can select the source database and destination database for you, and then to set up the clusters appropriately in `setUp` so that the source matches the destination if and only if we're in `CopyConfig.WithinCluster` mode. I can buy that that's a bit too much machinery for this PR, if you wanted to separate that out (if you thought it was a good idea)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -316,6 +317,51 @@ public RecordCursor<DataInKeySpacePath> exportAllData(@Nonnull FDBRecordContext
                     1);
     }
 
+    @Nonnull
+    @Override
+    public CompletableFuture<Void> importData(@Nonnull FDBRecordContext context,
+                                              @Nonnull Iterable<DataInKeySpacePath> dataToImport) {
+        return toTupleAsync(context).thenCompose(targetTuple -> {
+            // We use a mapPipelined here to help control the rate of insertions into the directory layer if those
+            // are happening.
```
alecgrieser, line 326: It might be good to make it explicit that you're explaining why it's using a `mapPipelined` cursor with a pipeline size of 1 specifically

alecgrieser (changes requested):
I think the only important thing is to make sure we set the executor in the `fromIterator` cursor, and the rest of these are just informational/minor

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: I'm not entirely sure I follow.
I wanted a handful of these tests to run both when copying within a cluster, and between clusters. A new extension would need to be smart enough to only apply to appropriate tests. 
In my mind `CopyConfig` is just a slightly more readable version of `@BooleanSource("withinCluster")`.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
ScottDugas: The most recent run in PRB took 2.3 seconds for the worst one. I saw 3.1 seconds locally. So pretty confident.

| name | method | time |
| --- | --- | --- |
| [1] !manyPaths, !useDirectoryLayer	| importALotOfData(boolean, boolean)[1] |	0.188s	passed |
| [2] !manyPaths, useDirectoryLayer	| importALotOfData(boolean, boolean)[2]	| 0.267s	passed |
| [3] manyPaths, !useDirectoryLayer	| importALotOfData(boolean, boolean)[3]	| 0.250s	passed |
| [4] manyPaths, useDirectoryLayer	| importALotOfData(boolean, boolean)[4]	| 2.382s	passed |

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: ok, I tried to make it clear. It's a bit tricky because 2 or 3 might also be fine.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImportDataTest.java
alecgrieser: The thing that's weird to me is that the class sets up the clusters to be different (if there's more than one around), and then the copy config overrides that. In my mind, it would be nice if `setup` set those two clusters consistently.

It is possible that there would be too much refactoring involved here to make that work in a way that only the relevant tests run in both modes.

Issue #3570, opened 2025-09-08
Implement subscript operator


hatyo:
It would be useful to have a subscript operator that can be used to access an element of an array:

```sql
create table A(pk integer, x integer array, primary key(pk));
insert into A values (1, [1, 2, 3]), (2, [2, 3, 4]), (3, [3, 4, 5]);
select x[2] from A where pk = 1;
-- returns 3
select x[0] from A where pk = 1;
-- returns 1
select x[4] from A where pk = 1;
-- out-of-bound error
```

It should be possible to use it with complex nested typing scenarios as well:

```sql
create type as struct S(a integer, b integer);
create table A(pk integer, x S array, primary key(pk));
insert into A values (1, [{10, 20}, {30, 40}]), (2, [{50, 60}]), (3, [{70, 80}, {90, 100}]);
select x[1] from A;
-- returns {30, 40}
select x[1].b from A;
-- returns 30
```

PR #3586, opened 2025-09-10
Support SQL array subscript operator

### LINKED ISSUE #3584: Support `NOT IN`
Currently we do not support `NOT IN` queries:

```sql
select * from a where x not in ('a', 'b', 'c');
```

This should be supported.

hatyo:
This introduces support to SQL array subscription operator, with the following caveats:
- It works with one-dimensional arrays only.
- Similar to other vendors, such as Postgres and Oracle, it returns `null` when the index is out-of-bound instead of raising an error.
- Uses 1-based index, as defined in SQL standard.

Example:

```sql
create table A(pk integer, x integer array, primary key(pk))
insert into A values (1, [1, 2, 3]), (2, [2, 3, 4]), (3, [3, 4, 5])
select X[2] from A where pk = 1;
-- returns 2
select X[1+1] from A where pk = 1;
-- returns 2
select X[-1] from A where pk = 1;
-- returns null
select X[1000] from A where pk = 1;
-- returns null
```

This also restructures the parsing rules for expressions such that it prohibits more invalid combinations of binary expressions, and aligns them to work well with the subscript operator, and adds support to `not in` which was previously throwing a syntax error.

This fixes #3584, #3570, and partially #3588 with few limitations.

### DIFF
--- a/fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/DelegatingVisitorTest.java
+++ b/fdb-relational-core/src/test/java/com/apple/foundationdb/relational/recordlayer/query/DelegatingVisitorTest.java
@@ -21,14 +21,20 @@
 package com.apple.foundationdb.relational.recordlayer.query;
 
 import com.apple.foundationdb.record.PlanHashable;
+import com.apple.foundationdb.record.query.plan.cascades.values.LiteralValue;
 import com.apple.foundationdb.relational.api.ddl.NoOpQueryFactory;
 import com.apple.foundationdb.relational.api.metadata.DataType;
+import com.apple.foundationdb.relational.generated.RelationalLexer;
+import com.apple.foundationdb.relational.generated.RelationalParser;
 import com.apple.foundationdb.relational.recordlayer.ddl.NoOpMetadataOperationsFactory;
 import com.apple.foundationdb.relational.recordlayer.metadata.RecordLayerColumn;
 import com.apple.foundationdb.relational.recordlayer.metadata.RecordLayerSchemaTemplate;
 import com.apple.foundationdb.relational.recordlayer.metadata.RecordLayerTable;
 import com.apple.foundationdb.relational.recordlayer.query.visitors.BaseVisitor;
 import com.apple.foundationdb.relational.recordlayer.query.visitors.DelegatingVisitor;
+import org.antlr.v4.runtime.CommonTokenStream;
+import org.apache.commons.lang3.mutable.MutableBoolean;
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


arnaud-lacurie:
Could you call out in the description that the error code has changed for a certain operation? So that we can surface it in our release notes.

yaml-tests/src/test/resources/arrays.yamsql
```diff
@@ -65,4 +65,35 @@ test_block:
...
+    -
+      # out of bound
+      - query: select X[100000] AS ERR FROM A WHERE pk = 1
+      - supported_version: !current_version
+      - result: [{ERR: !null _}]
+    -
+      # out of bound
+      - query: select X[-1] AS ERR FROM A WHERE pk = 1
+      - supported_version: !current_version
+      - result: [{ERR: !null _}]
```
arnaud-lacurie, line 103: Would creating an index on `X[1]` be supported out of the box? If so, could you add a test for it?

yaml-tests/src/test/resources/arrays.yamsql
```diff
@@ -65,4 +65,35 @@ test_block:
...
+    -
+      # high-order subscript indexes are supported
+      - query: select pk FROM A WHERE X[X[1] - X[1] + 1] between 0 + X[1+0] and X[1-0] - 0
+      - supported_version: !current_version
+      - result: [{PK: 1}, {PK: 2}, {PK: 3}]
+    -
+      # out of bound
+      - query: select X[100000] AS ERR FROM A WHERE pk = 1
+      - supported_version: !current_version
+      - result: [{ERR: !null _}]
```
arnaud-lacurie, line 98: Could you add a test with `X[0]`?
And another test where the array is empty?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/SqlFunctionCatalogImpl.java
```diff
@@ -122,6 +122,7 @@ private static ImmutableMap<String, Function<Integer, Optional<BuiltInFunction<?
                 .put("&", argumentsCount -> BuiltInFunctionCatalog.resolve("bitand", argumentsCount))
                 .put("|", argumentsCount -> BuiltInFunctionCatalog.resolve("bitor", argumentsCount))
                 .put("^", argumentsCount -> BuiltInFunctionCatalog.resolve("bitxor", argumentsCount))
+                .put("[", argumentsCount -> BuiltInFunctionCatalog.resolve("subscript", argumentsCount))
```
arnaud-lacurie: Could the key be `[]` instead of `[`?

fdb-relational-core/src/main/antlr/RelationalParser.g4
```diff
@@ -1129,43 +1129,38 @@ namedFunctionArg
...
+    | fullColumnName                                                                      #fullColumnNameExpressionAtom // done
+    | functionCall                                                                        #functionCallExpressionAtom // done
+    | preparedStatementParameter                                                          #preparedStatementParameterAtom // done
+    | recordConstructor                                                                   #recordConstructorExpressionAtom // done
+    | arrayConstructor                                                                    #arrayConstructorExpressionAtom // done
+    | base=expressionAtom LEFT_SQUARE_BRACKET index=expressionAtom RIGHT_SQUARE_BRACKET   #subscriptExpression // done
+    | left=expressionAtom bitOperator right=expressionAtom                                #bitExpressionAtom // done
+    | left=expressionAtom mathOperator right=expressionAtom                               #mathExpressionAtom // done
+    | left=expressionAtom comparisonOperator right=expressionAtom                         #binaryComparisonPredicate // done
```
arnaud-lacurie, line 1156: Nice 👍

yaml-tests/src/test/resources/arrays.yamsql
hatyo: No, that would require adding support for this operator in the `FunctionKeyExpression` API, so it's not supported out of the box, feel free to open a GitHub issue to implement that feature.

yaml-tests/src/test/resources/arrays.yamsql
hatyo: > Could you add a test with X[0]?

done.

> And another test where the array is empty?

we do not have support for empty array literals (see #3588), I added _some_ support to get the tests going, but there are some subtleties around empty array literal's flowing unknown types that requires addressing (see #3589 as a way to explicitly prevent that from happening through explicit casting).

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/SqlFunctionCatalogImpl.java
hatyo: Done.

hatyo:
> Could you call out in the description that the error code has changed for a certain operation? So that we can surface it in our release notes.

If you're referring to this test:

```yaml
      - query: select a, e from ta where e not in ('foo' , 35)
      - initialVersionAtLeast: !current_version
      - error: "42804"
      - initialVersionLessThan: !current_version
      - error: "42601"
```

The difference we see here is purely related to the test itself which I think was poorly written. The test seems to want to verify the SQL error code resulting from having elements in the in-list of incompatible types (`foo` and `35`), however what it was instead verifying a syntax error (`42601`), due to the lack of support of `not in` expression.
With this PR, we _do_ support `not in` so we can verify the data type mismatch (`42804`) which the test intended to verify.

yaml-tests/src/test/resources/arrays.yamsql
arnaud-lacurie: https://github.com/FoundationDB/fdb-record-layer/issues/3592

PR #3591, opened 2025-09-11
Fix ResolvedKeySpacePath.equals and hashCode

ScottDugas:
This fixes `ResolvedKeySpace.equals` and `hashCode` so that they follow the spec and are consistent.
Notably there were a few issues:
- They did not correctly handle `KeyType.BYTES`
- They disagreed about what data to use (hashCode used to look at the `PathValue`, while equals looked at the value)
- `PathValue` did not implement `equals`/`hashCode`
- The `remainder` was not included.
- KeySpacePathImpl redundantly checked the value from the directory
- It did not check the parent of the ResolvedKeySpacePath.

Fixes: #3594

### DIFF
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
@@ -83,30 +83,97 @@ class ResolvedKeySpacePathTest {
     @MethodSource("testEqualsHashCode")
     void testEqualsHashCode(@Nonnull KeyType keyType, boolean constantDirectory, boolean differenceInParent) {
         @Nonnull TestValuePair values = TYPE_TEST_VALUES.get(keyType);
-        // Test case 1: Same logical and resolved values (existing test)
-        ResolvedKeySpacePath path1 = createResolvedPath(keyType, values.getValue1(), values.getValue1(),
-                createRootParent(), constantDirectory, differenceInParent);
-        ResolvedKeySpacePath path2 = createResolvedPath(keyType, values.getValue1(), values.getValue1(),
-                createRootParent(), constantDirectory, differenceInParent);
+
+        // Create a single KeySpace with the appropriate directory structure
+        KeySpaceDirectory rootDir = new KeySpaceDirectory("root", KeyType.STRING, "root");
+        KeySpaceDirectory childDir = constantDirectory
+                ? new KeySpaceDirectory("test", keyType, values.getValue1())
+                : new KeySpaceDirectory("test", keyType);
+        rootDir.addSubdirectory(childDir);
+
+        // Optionally add a constant child for the differenceInParent test
+        if (differenceInParent) {
+            KeySpaceDirectory constantChild = new KeySpaceDirectory("constant", KeyType.STRING, "Constant");
+            childDir.addSubdirectory(constantChild);
+        }
+
+        KeySpace keySpace = new KeySpace(rootDir);
+
+        // Create paths from the same KeySpace
+        KeySpacePath rootPath1 = keySpace.path("root");
+        KeySpacePath rootPath2 = keySpace.path("root");
+
+        KeySpacePath childPath1;
+        KeySpacePath childPath2;
+        if (constantDirectory) {
+            childPath1 = rootPath1.add("test");
+            childPath2 = rootPath2.add("test");
+        } else {
+            childPath1 = rootPath1.add("test", values.getValue1());
+            childPath2 = rootPath2.add("test", values.getValue1());
+        }
+
+        // Create ResolvedKeySpacePath instances
+        ResolvedKeySpacePath resolvedRoot1 = new ResolvedKeySpacePath(null, rootPath1, new PathValue("root", null), null);
+        ResolvedKeySpacePath resolvedRoot2 = new ResolvedKeySpacePath(null, rootPath2, new PathValue("root", null), null);
+
+        ResolvedKeySpacePath path1 = new ResolvedKeySpacePath(resolvedRoot1, childPath1, new PathValue(values.getValue1(), null), null);
+        ResolvedKeySpacePath path2 = new ResolvedKeySpacePath(resolvedRoot2, childPath2, new PathValue(values.getValue1(), null), null);
+
+        if (differenceInParent) {
+            KeySpacePath constantChildPath1 = childPath1.add("constant");
+            KeySpacePath constantChildPath2 = childPath2.add("constant");
+            path1 = new ResolvedKeySpacePath(path1, constantChildPath1, new PathValue("Constant", null), null);
+            path2 = new ResolvedKeySpacePath(path2, constantChildPath2, new PathValue("Constant", null), null);
+        }
 
         // Test equality contracts
         assertEquals(path1, path2, "Identical paths should be equal");
         assertEquals(path2, path1, "Symmetry: path2.equals(path1)");
         assertEquals(path1.hashCode(), path2.hashCode(), "Equal objects must have equal hash codes");
-            
+
         // Test inequality when values differ (except NULL type which only has null values)
         if (keyType != KeyType.NULL) {
-            ResolvedKeySpacePath path3 = createResolvedPath(keyType, values.getValue2(), createRootParent(), constantDirectory);
-            assertNotEquals(path1, path3, "Paths with different values should not be equal");
+            if (constantDirectory) {
+                // For constant directories, we need a different directory (and thus different KeySpace) to test different values
+                // this doesn't really need to be parameterized by value type, since they will always be non-equal due
+                // to the directories being different
+                KeySpaceDirectory rootDir3 = new KeySpaceDirectory("root", KeyType.STRING, "root");
+                KeySpaceDirectory childDir3 = new KeySpaceDirectory("test", keyType, values.getValue2());
+                rootDir3.addSubdirectory(childDir3);
+                KeySpace keySpace3 = new KeySpace(rootDir3);
 
-            assertNotEquals(createResolvedPath(keyType, values.getValue1(), values.getValue2(), createRootParent(), constantDirectory, differenceInParent),
-                    path1, "Paths with different resolved values should not be equal");
-            assertNotEquals(createResolvedPath(keyType, values.getValue2(), values.getValue1(), createRootParent(), constantDirectory, differenceInParent),
-                    path1, "Paths with different logical values should not be equal");
+                KeySpacePath rootPath3 = keySpace3.path("root");
+                KeySpacePath childPath3 = rootPath3.add("test");
+
+                ResolvedKeySpacePath resolvedRoot3 = new ResolvedKeySpacePath(null, rootPath3, new PathValue("root", null), null);
+                ResolvedKeySpacePath path3 = new ResolvedKeySpacePath(resolvedRoot3, childPath3, new PathValue(values.getValue2(), null), null);
+
+                assertNotEquals(path1, path3, "Paths with different constant values should not be equal");
+            } else {
+                // For non-constant directories, we can use the same directory with different values
+                KeySpacePath childPath3 = rootPath1.add("test", values.getValue2());
+                ResolvedKeySpacePath path3 = new ResolvedKeySpacePath(resolvedRoot1, childPath3, new PathValue(values.getValue2(), null), null);
+                assertNotEquals(path1, path3, "Paths with different values should not be equal");
+            }
+
+            // Test different resolved value (same logical, different resolved)
+            KeySpacePath childPath4 = constantDirectory
+                    ? rootPath1.add("test")
+                    : rootPath1.add("test", values.getValue1());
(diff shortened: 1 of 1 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePath.java
```diff
@@ -229,13 +229,15 @@ public boolean equals(Object other) {
...
+        return Objects.equals(this.getResolvedPathValue(), otherPath.getResolvedPathValue()) &&
+                Objects.equals(this.getParent(), otherPath.getParent()) &&
+                this.inner.equals(otherPath.inner) &&
+                Objects.equals(this.remainder, otherPath.remainder);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(inner, getResolvedPathValue());
+        return Objects.hash(inner, getResolvedPathValue(), remainder);
```
ScottDugas: I think this needs to include the parent. 
```suggestion
        return Objects.hash(inner, getResolvedPathValue(), remainder, getParent());
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -282,21 +282,20 @@ public boolean equals(Object obj) {
         // but comparing two directories by value would necessitate traversing the entire directory
         // tree, so instead we will use a narrower definition of equality here.
         boolean directoriesEqual = Objects.equals(this.getDirectory().getKeyType(), that.getDirectory().getKeyType()) &&
-                                   Objects.equals(this.getDirectory().getName(), that.getDirectory().getName()) &&
-                                   Objects.equals(this.getDirectory().getValue(), that.getDirectory().getValue());
+                                   Objects.equals(this.getDirectory().getName(), that.getDirectory().getName());
```
ohadzeliger: are `keyType` and `name` enough to qualify an identifier for a `KeySpaceDirectory`? In that case, would it make sense to move that logic into a `KeySpaceDirectory.equals`?

ohadzeliger: Would a comparison that accounts for the `directory`'s `ANY_TYPE` be appropriate?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePath.java
```diff
@@ -229,13 +229,15 @@ public boolean equals(Object other) {
         }
 
         ResolvedKeySpacePath otherPath = (ResolvedKeySpacePath) other;
-        return this.inner.equals(otherPath.inner)
-               && Objects.equals(this.getResolvedValue(), otherPath.getResolvedValue());
+        return Objects.equals(this.getResolvedPathValue(), otherPath.getResolvedPathValue()) &&
+                Objects.equals(this.getParent(), otherPath.getParent()) &&
```
ohadzeliger, line 233: Why is this implementation going recursively up  the chain while the one at `KeySpacePathImpl` does not?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/PathValueTest.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    static Stream<Arguments> unequalPathValuePairs() {
+        return Stream.of(
+                Arguments.of("different string values", "test1", null, "test2", null),
+                Arguments.of("different long values", 42L, null, 100L, null),
+                Arguments.of("different boolean values", true, null, false, null),
+                Arguments.of("different types", "string", null, 42L, null),
+                Arguments.of("different metadata", "test", new byte[]{1, 2, 3}, "test", new byte[]{4, 5, 6}),
+                Arguments.of("one null metadata", "test", new byte[]{1, 2, 3}, "test", null),
+                Arguments.of("different value with same metadata", "test1", new byte[]{1, 2, 3}, "test2", new byte[]{1, 2, 3})
```
ohadzeliger, line 64: For completeness, maybe add "one null value"?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
```diff
@@ -0,0 +1,232 @@
...
+        private final Supplier<Object> value2Supplier;
+        
+        TestValuePair(@Nonnull Supplier<Object> value1Supplier, @Nonnull Supplier<Object> value2Supplier) {
+            this.value1Supplier = value1Supplier;
+            this.value2Supplier = value2Supplier;
+        }
+        
+        @Nullable
+        Object getValue1() {
+            return value1Supplier.get();
```
ohadzeliger, line 271: Nice way to get different instance every time!

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
```diff
@@ -0,0 +1,232 @@
...
+        ResolvedKeySpacePath path2 = createResolvedPath(keyType, values.getValue1(), values.getValue1(),
+                createRootParent(), constantDirectory, differenceInParent);
+
+        // Test equality contracts
+        assertEquals(path1, path2, "Identical paths should be equal");
+        assertEquals(path2, path1, "Symmetry: path2.equals(path1)");
+        assertEquals(path1.hashCode(), path2.hashCode(), "Equal objects must have equal hash codes");
+            
+        // Test inequality when values differ (except NULL type which only has null values)
+        if (keyType != KeyType.NULL && values.getValue2() != null) {
```
ohadzeliger: Isn't this redundant (if `keyType == KeyType.NULL` then both value1 and value2 are null)?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
```diff
@@ -0,0 +1,232 @@
...
+        assertEquals(path1.hashCode(), path3.hashCode(), "Paths with the same remainder should have same hashCode");
+        ResolvedKeySpacePath nullRemainder = new ResolvedKeySpacePath(null, innerPath, value, null);
+        assertNotEquals(path1, nullRemainder, "Path without a remainder should not be the equal to one with a remainder");
+        assertNotEquals(nullRemainder, null, "Make sure null is properly handled in equals");
+        IntStream.range(0, 10_000).mapToObj(i -> new ResolvedKeySpacePath(null, innerPath, value, Tuple.from(i)))
+                .filter(path -> path.hashCode() != path1.hashCode())
+                .findAny()
+                .orElseThrow(() -> new AssertionError("Paths with different remainders should sometimes have different hash codes"));
+    }
+
```
ohadzeliger, line 226: Maybe add tests for `ResolvedKeyspacePath` with different parents?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ScottDugas: Yeah, it probably is worth putting this in `KeySpaceDirectory.equals`. I think the original author was trying to avoid doing a full hierachy comparison of the two, which would end up creating a lot of redundancy.
That being said, I think there are better ways to do that. And, in the common case you should have one `KeySpace`, and so they should be reference equals, or fail pretty quickly.

Are you referring to `ANY_VALUE`? Do you have specific concerns there, I don't think there should be any special handling of it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePath.java
ScottDugas: `KeySpacePathImpl` does check the parent.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
ScottDugas: Yeah, added an assertion just to be sure.

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/ResolvedKeySpacePathTest.java
ScottDugas: That is already covered by the `differenceInParent` parameter to `testEqualsHashCode`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpaceDirectory.java
```diff
@@ -719,6 +720,12 @@ protected static boolean areEqual(Object o1, Object o2) {
             }
         }
 
+        // Handle ANY_VALUE specially - typeOf does not support ANY_VALUE
+        boolean isAnyValue = (o1 == ANY_VALUE || o2 == ANY_VALUE);
+        if (isAnyValue) {
+            return Objects.equals(o1, o2);
```
ohadzeliger, line 726: As a defensive coding we can add `equals` and `hashCode` to `AnyValue`, though all reasonable implementations should use the static constant...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
```diff
@@ -276,27 +276,22 @@ public boolean equals(Object obj) {
...
     }
 
     @Override
     public int hashCode() {
         return Objects.hash(
                 getDirectory().getKeyType(),
                 getDirectory().getName(),
-                getDirectory().getValue(),
-                getValue(),
+                KeySpaceDirectory.valueHashCode(getValue()),
```
ohadzeliger, line 293: This would be the same as `getDirectory().getValue()`, right?

ScottDugas: Only if the directory is a constant. If the `getDirectory().getValue()` is `ANY_VALUE`, this would be any value of that type.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpaceDirectory.java
ScottDugas: Maybe, but `AnyValue` is a private-nested class, so `KeySpaceDirectory` is the only way to construct one, so you really shouldn't be accessing it other than via the constant.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/KeySpacePathImpl.java
ohadzeliger: In that case there may be a discrepancy where `equals` (line 281) compares: 
- keyspacePath.value
- `KeyPathDirectory.equalsIgnoringHierarchy` -> 
  - name
  - KeyType
  - directory.value

Whereas `hashCode` compares:
  - name
  - keyType
  - keyspacePath.value

So it looks as if the `hashCode` could benefit from the extra hash for `directory.value`

ScottDugas: Ah, yes, that is probably true.
It's not catastrophic to have `hashCode` compare less than `equals`, especially given that in standard use cases, if everything else is equal, the directory value should be too.
That being said, it would probably be confusing to anyone else coming upon this, so at least it should have a comment.

ScottDugas: I spent a bit of time thinking about this, and I think, actually, the correct answer is to leave `KeySpaceDirectory` use reference equality for its `.equals` method and `hashCode` (inheriting from `Object`), since these are supposed to be singletons in a singleton `KeySpace`.
Changing the equality check is trivial, but it will be a little bit of work to update the tests.

ohadzeliger (approved):

PR #3593, opened 2025-09-12
Fix compensation logic to not early out if child compensations are needed

normen662:
…eded

### DIFF
--- a/yaml-tests/src/test/resources/aggregate-index-tests.metrics.binpb
+++ b/yaml-tests/src/test/resources/aggregate-index-tests.metrics.binpb
@@ -1,8 +1,7 @@
 ù
 H
 agg-index-tests5EXPLAIN select col1, sum(col2) from T1 group by col1;–
-Ù¿ƒøè °âØ
-(@0Ëö,80@YAISCAN(MV1 <,> BY_GROUP -> [_0: KEY:[0], _1: VALUE:[0]]) | MAP (_._0 AS COL1, _._1 AS _1)÷
+ÙëÿËè ”éÍ(@0Ì”080@YAISCAN(MV1 <,> BY_GROUP -> [_0: KEY:[0], _1: VALUE:[0]]) | MAP (_._0 AS COL1, _._1 AS _1)÷
 digraph G {
   fontname=courier;
   rankdir=BT;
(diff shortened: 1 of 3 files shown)


yaml-tests/src/test/resources/distinct-from.yamsql
```diff
@@ -95,13 +95,13 @@ test_block:
         # Cannot support FORCE_CONTINUATIONS due to non-empty aggregate scan returning a begin continuation.
         # See: https://github.com/FoundationDB/fdb-record-layer/issues/3206
       - maxRows: 0
-      - explain: "AISCAN(CNT <,> BY_GROUP -> [_0: VALUE:[0]]) | MAP (_ AS _0) | ON EMPTY NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)"
+      - explain: "AISCAN(CNT <,> BY_GROUP -> [_0: VALUE:[0]]) | FILTER NULL NOT_DISTINCT_FROM NULL | MAP (_ AS _0) | ON EMPTY NULL | MAP (coalesce_long(_._0._0, promote(0l AS LONG)) AS _0)"
```
alecgrieser, line 98: Hm. I believe this slipped through the cracks here because `null is not distinct from null` should be simplified to `true` during query rewriting. I guess that transformation isn't happening? (Likewise for the `@c9 NOT_DISTINCT_FROM @c9` predicate below.)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/GroupByExpression.java
```diff
@@ -652,7 +652,8 @@ public Compensation compensate(@Nonnull final PartialMatch partialMatch,
             return Compensation.impossibleCompensation();
         }
         final var compensatedResult = compensatedResultOptional.get();
-        if (!compensatedResult.getResultCompensationFunction().isNeeded()) {
+        if (!childCompensation.isNeeded() &&
+                !compensatedResult.getResultCompensationFunction().isNeeded()) {
```
alecgrieser, line 656: These all look good to me, but I am surprised that we didn't hit these with any of our existing queries except for two in distinct from. It would be nice if there were a little more coverage of this case. Do you know of any additional queries that would have suffered from a bug here that we could add to our test suite?

yaml-tests/src/test/resources/distinct-from.yamsql
normen662: Maybe it is happening but it didn't survive the rewrite cost model.

PR #3595, opened 2025-09-12
Reactivate amortization of plan generation of functions.

### LINKED ISSUE #3394: Reuse generated SQL functions plans
Each time we invoke a (temporary) SQL function we regenerate its plan, this is mainly done for two reasons:
- avoid identifier collisions.
- plans are mutable, the planner can alter the passed plan in a way that makes it invalid to use again.

For these reasons, we could deep-copy the function plan using a special rebase operator that assigns new identifiers as it traverses the plan and copies it.

hatyo:
This restores the memoization of plan generation of both temporary and non-temporary invoked routines.

This fixes #3394.

### DIFF
--- a/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
+++ b/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
@@ -23,7 +23,6 @@ package com.apple.foundationdb.relational.recordlayer.metadata;
 import com.apple.foundationdb.record.query.plan.cascades.RawSqlFunction;
 import com.apple.foundationdb.record.query.plan.cascades.UserDefinedFunction;
 import com.apple.foundationdb.relational.api.metadata.InvokedRoutine;
-import com.apple.foundationdb.relational.recordlayer.query.functions.CompiledSqlFunction;
 import com.apple.foundationdb.relational.recordlayer.util.MemoizedFunction;
 import com.apple.foundationdb.relational.util.Assert;
 
@@ -45,7 +44,7 @@ public class RecordLayerInvokedRoutine implements InvokedRoutine {
     private final boolean isTemporary;
 
     @Nonnull
-    private final Function<Boolean, UserDefinedFunction> userDefinedFunctionSupplier;
+    private final Function<Boolean, UserDefinedFunction> userDefinedFunctionProvider;
 
     @Nonnull
     private final UserDefinedFunction serializableFunction;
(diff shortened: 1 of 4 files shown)


fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompiledSqlFunction.java
```diff
@@ -92,12 +97,12 @@ public RecordMetaDataProto.PUserDefinedFunction toProto(@Nonnull final PlanSeria
 
     @Nonnull
     @Override
-    public RelationalExpression encapsulate(@Nonnull final List<? extends Typed> arguments) {
+    public Reference encapsulate(@Nonnull final List<? extends Typed> arguments) {
```
normen662: Wouldn't you only ever put exactly one expression into the `Reference`? If so, why put it in a reference? Why not just leave it as it was?

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompiledSqlFunction.java
```diff
@@ -156,14 +155,29 @@ public RelationalExpression encapsulate(@Nonnull final Map<String, ? extends Typ
...
         }
-        final var qun = Quantifier.forEach(Reference.initialOf(resultBuilder.addQuantifier(rangeOfOnePlan()).build().buildSelect()),
-                parametersCorrelation.get());
-        final var bodyQun = Quantifier.forEach(Reference.initialOf(body));
+        final var argumentsExpression = resultBuilder.addQuantifier(rangeOfOnePlan()).build().buildSelect();
+        return constructTableFunctionPlan(argumentsExpression);
+    }
+
+    @Nonnull
+    private Reference constructTableFunctionPlan(@Nonnull final RelationalExpression argumentsExpression) {
```
normen662: `expandTableFunction`? I am trying to get around calling the result a plan, because in the rec-layer a plan extends from `RecordQueryPlan`

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
normen662: It's not strictly on your PR, but this `...Supplier` should be a `...Provider` in order to reduce confusion as to why this thing needs an argument.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompiledSqlFunction.java
hatyo: When generating a plan, we never have more than one expression under a reference.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/functions/CompiledSqlFunction.java
hatyo: Good idea, done.

fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/metadata/RecordLayerInvokedRoutine.java
hatyo: Done.

PR #3598, opened 2025-09-16
HNSW (Hierarchical Navigable Small Worlds) implementation as fdb-extension

normen662:
This PR implements a HNSW (Hierarchical Navigable Small World) structure as an part of `fdb-extension` according to https://arxiv.org/pdf/1603.09320.

- the implementation is a basic structure that _lives_ inside of a subspace in fdb
- currently we  store and retrieve vectors of class `Vector` that holds component data using `Half`s which are FP16 encoded.
- we currently only support insertion, batch insertion and retrieval of data via approximative nearest neighbor searches 
- we currently support the following options:
   - `M` the optimal number of connections between nodes within a layer of the HNSW
   - `MMax` the maximum number of connections between nodes within a layer of the HNSW
   - `MMax0` the maximum number of connections between nodes within layer `0` of the HNSW
   - `efConstruction` the size of the search queue during thorough (non-greedy) searches during insertion
   - `extendCandidates` a boolean indicating if we should extend the candidates we can connect to to by also searching within the neighbors of already found candidates
   - `keepPrunedConnections` a boolean indicating whether under certain circumstances discarded candidates to connect to are reevaluated
- deletes are not implemented (yet)
- tests include basic testing using the SIFTsmall data set that is comprised of 10k vectors, 100 queries

### DIFF
--- a/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/HNSW.java
+++ b/fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/HNSW.java
@@ -91,7 +91,6 @@ public class HNSW {
     public static final int DEFAULT_M = 16;
     public static final int DEFAULT_M_MAX = DEFAULT_M;
     public static final int DEFAULT_M_MAX_0 = 2 * DEFAULT_M;
-    public static final int DEFAULT_EF_SEARCH = 100;
     public static final int DEFAULT_EF_CONSTRUCTION = 200;
     public static final boolean DEFAULT_EXTEND_CANDIDATES = false;
     public static final boolean DEFAULT_KEEP_PRUNED_CONNECTIONS = false;
@@ -123,7 +122,6 @@ public class HNSW {
         private final int m;
         private final int mMax;
         private final int mMax0;
-        private final int efSearch;
         private final int efConstruction;
         private final boolean extendCandidates;
         private final boolean keepPrunedConnections;
@@ -135,14 +133,13 @@ public class HNSW {
             this.m = DEFAULT_M;
             this.mMax = DEFAULT_M_MAX;
             this.mMax0 = DEFAULT_M_MAX_0;
-            this.efSearch = DEFAULT_EF_SEARCH;
             this.efConstruction = DEFAULT_EF_CONSTRUCTION;
             this.extendCandidates = DEFAULT_EXTEND_CANDIDATES;
             this.keepPrunedConnections = DEFAULT_KEEP_PRUNED_CONNECTIONS;
         }
 
         protected Config(@Nonnull final Random random, @Nonnull final Metric metric, final boolean useInlining,
-                         final int m, final int mMax, final int mMax0, final int efSearch, final int efConstruction,
+                         final int m, final int mMax, final int mMax0, final int efConstruction,
                          final boolean extendCandidates, final boolean keepPrunedConnections) {
             this.random = random;
             this.metric = metric;
@@ -150,7 +147,6 @@ public class HNSW {
             this.m = m;
             this.mMax = mMax;
             this.mMax0 = mMax0;
-            this.efSearch = efSearch;
             this.efConstruction = efConstruction;
             this.extendCandidates = extendCandidates;
             this.keepPrunedConnections = keepPrunedConnections;
@@ -182,10 +178,6 @@ public class HNSW {
             return mMax0;
         }
 
-        public int getEfSearch() {
-            return efSearch;
-        }
-
         public int getEfConstruction() {
             return efConstruction;
         }
@@ -201,15 +193,15 @@ public class HNSW {
         @Nonnull
         public ConfigBuilder toBuilder() {
             return new ConfigBuilder(getRandom(), getMetric(), isUseInlining(), getM(), getMMax(), getMMax0(),
-                    getEfSearch(), getEfConstruction(), isExtendCandidates(), isKeepPrunedConnections());
+                    getEfConstruction(), isExtendCandidates(), isKeepPrunedConnections());
         }
 
         @Override
         @Nonnull
         public String toString() {
             return "Config[metric=" + getMetric() + "isUseInlining" + isUseInlining() + "M=" + getM() +
-                    " , MMax=" + getMMax() + " , MMax0=" + getMMax0() + ", efSearch=" + getEfSearch() +
-                    ", efConstruction=" + getEfConstruction() + ", isExtendCandidates=" + isExtendCandidates() +
+                    " , MMax=" + getMMax() + " , MMax0=" + getMMax0() + ", efConstruction=" + getEfConstruction() +
+                    ", isExtendCandidates=" + isExtendCandidates() +
                     ", isKeepPrunedConnections=" + isKeepPrunedConnections() + "]";
         }
     }
@@ -230,7 +222,6 @@ public class HNSW {
         private int m = DEFAULT_M;
         private int mMax = DEFAULT_M_MAX;
         private int mMax0 = DEFAULT_M_MAX_0;
-        private int efSearch = DEFAULT_EF_SEARCH;
         private int efConstruction = DEFAULT_EF_CONSTRUCTION;
         private boolean extendCandidates = DEFAULT_EXTEND_CANDIDATES;
         private boolean keepPrunedConnections = DEFAULT_KEEP_PRUNED_CONNECTIONS;
@@ -239,7 +230,7 @@ public class HNSW {
         }
 
         public ConfigBuilder(@Nonnull final Random random, @Nonnull final Metric metric, final boolean useInlining,
-                             final int m, final int mMax, final int mMax0, final int efSearch, final int efConstruction,
+                             final int m, final int mMax, final int mMax0, final int efConstruction,
                              final boolean extendCandidates, final boolean keepPrunedConnections) {
             this.random = random;
             this.metric = metric;
@@ -247,7 +238,6 @@ public class HNSW {
             this.m = m;
             this.mMax = mMax;
             this.mMax0 = mMax0;
-            this.efSearch = efSearch;
             this.efConstruction = efConstruction;
             this.extendCandidates = extendCandidates;
             this.keepPrunedConnections = keepPrunedConnections;
@@ -314,15 +304,6 @@ public class HNSW {
             return this;
         }
 
-        public int getEfSearch() {
-            return efSearch;
-        }
-
-        public ConfigBuilder setEfSearch(final int efSearch) {
-            this.efSearch = efSearch;
-            return this;
-        }
-
         public int getEfConstruction() {
             return efConstruction;
         }
@@ -351,7 +332,7 @@ public class HNSW {
         }
 
         public Config build() {
-            return new Config(getRandom(), getMetric(), isUseInlining(), getM(), getMMax(), getMMax0(), getEfSearch(),
+            return new Config(getRandom(), getMetric(), isUseInlining(), getM(), getMMax(), getMMax0(),
                     getEfConstruction(), isExtendCandidates(), isKeepPrunedConnections());
         }
     }
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/HNSWTest.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/HNSWTest.java
@@ -68,6 +68,7 @@ import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
@@ -255,6 +256,13 @@ public class HNSWTest {
                 TimeUnit.NANOSECONDS.toMillis(endTs - beginTs),
                 onReadListener.getNodeCountByLayer(), onReadListener.getBytesReadByLayer(),
                 String.format(Locale.ROOT, "%.2f", recall * 100.0d));
+
+        final Set<Long> usedIds =
+                LongStream.range(0, 1000)
+                        .boxed()
+                        .collect(Collectors.toSet());
+
+        hnsw.scanLayer(db, 0, 100, node -> Assertions.assertTrue(usedIds.remove(node.getPrimaryKey().getLong(0))));
     }
 
     private int basicInsertBatch(final HNSW hnsw, final int batchSize,


gradle/libs.versions.toml
```diff
@@ -37,6 +37,7 @@ generatedAnnotation = "1.3.2"
 grpc = "1.64.1"
 grpc-commonProtos = "2.37.0"
 guava = "33.3.1-jre"
+half4j = "0.0.2"
```
alecgrieser, line 40: There is probably somewhat of an open question as to whether we want to add a new dependency to support fp16 encoding or not. It is certainly a bit of a shame that all consumers would need to take on the dependency, even if they do not end up using the HNSW index.

I could see the argument that we need yet another subproject here for vector indexes, and that's where we put the new code, like was done with `fdb-record-layer-lucene`. Then projects that don't use the new vector sub-component don't need to worry about this.

I did also take a quick look, and it actually seems like writing our own fp32 to fp16 utility wouldn't be _that_ hard. In the common case, it's just a matter of truncating the mantissa and re-encoding the exponent. There are some truncation edge cases (e.g., the numbers can blow up to infinity for a lot of exponent values) that aren't that bad, and then the only "hard" case is making sure that you can convert from a normal 32 (or 64) bit value to a sub-normal 16-bit value. But even that isn't too bad. I have some prototype code locally that could theoretically do that, using java `short` values as the storage for 16-bit values. Getting the logic to go back and forth is actually not too bad; if we actually want to be able to do fp16 math on those shorts, that's a little more intense, though actually maybe not too too bad if we absolutely needed to.

gradle/scripts/log4j-test.properties
```diff
@@ -26,7 +26,7 @@ appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = %d [%level] %logger{1.} - %m %X%n%ex{full}
 
-rootLogger.level = debug
+rootLogger.level = info
```
alecgrieser, line 29: Any reason for this? Should we revert it?

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+ * <p>
+ * This class provides a generic framework for vectors of different numerical types,
+ * where {@code R} is a subtype of {@link Number}. It includes common operations and functionalities like size,
+ * component access, equality checks, and conversions. Concrete implementations must provide specific logic for
+ * data type conversions and raw data representation.
+
+ */
+public abstract class Vector {
+    @Nonnull
+    final double[] data;
```
alecgrieser, line 53: Is there a reason the base `Vector` class wraps a double vector? Rather than, say, a `FloatVector` wrapping a `float[]`?

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+        this.data = data;
+        this.hashCodeSupplier = Suppliers.memoize(this::computeHashCode);
+        this.toRawDataSupplier = Suppliers.memoize(this::computeRawData);
+    }
+
+    /**
+     * Returns the number of elements in the vector.
+     * @return the number of elements
+     */
+    public int size() {
```
alecgrieser, line 81: minor: Should this be `dimension` rather than `size`? I don't think I actually have an opinion here, but it occurs to me as a decision we could make

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+/**
+ * An abstract base class representing a mathematical vector.
+ * <p>
+ * This class provides a generic framework for vectors of different numerical types,
+ * where {@code R} is a subtype of {@link Number}. It includes common operations and functionalities like size,
+ * component access, equality checks, and conversions. Concrete implementations must provide specific logic for
+ * data type conversions and raw data representation.
+
+ */
+public abstract class Vector {
```
alecgrieser, line 51: It may be the answer is "no", but is there a better name than vector? The main concern is that it conflicts with the abstract data type name. But it also appears to be the name that other systems use for this. The only alternative name I could think of would be something like `FloatingPointVector` or like `CartesianVector` or something, neither of which I'm all that happy with

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+    public int precision() {
+        return (1 << precisionShift());
+    }
+
+    /**
+     * Returns the number of bits we need to shift {@code 1} to express {@link #precision()} used for the serialization
+     * of this vector per component.
+     * @return returns the number of bits we need to shift {@code 1} to express {@link #precision()}
+     */
+    public abstract int precisionShift();
```
alecgrieser, line 171: How many places is this used? Would it be better to remove this and make `precision` abstract? IIUC, `HalfVector::precision` should return 2 and `DoubleVector::precision` should return 8 (and presumably `FloatVector::precision` would return 4). Those are all, in fact, powers of two, but I'm not sure it's worth expressing it as such

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+     * <p>
+     * The result is {@code true} if and only if the argument is not {@code null} and is a {@code Vector} object that
+     * has the same data elements as this object. This method performs a deep equality check on the underlying data
+     * elements using {@link Objects#deepEquals(Object, Object)}.
+     * @param o the object to compare with this {@code Vector} for equality.
+     * @return {@code true} if the given object is a {@code Vector} equivalent to this vector, {@code false} otherwise.
+     */
+    @Override
+    public boolean equals(final Object o) {
+        if (!(o instanceof Vector)) {
```
alecgrieser, line 184: The other consequence of this is that vectors of different precisions will compare equals, if they represent the same data. Do we want that? I suppose `equals` (and thus `hashCode`) on a vector type is also a bit suspect given floating point precision shenanigans

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+
+        public HalfVector(@Nonnull final Half[] halfData) {
+            this(computeDoubleData(halfData));
+        }
+
+        public HalfVector(@Nonnull final double[] data) {
+            super(data);
+            this.toDoubleVectorSupplier = Suppliers.memoize(this::computeDoubleVector);
+        }
+
```
alecgrieser, line 261: These constructors are a little weird. The way it's set up, it looks like it immediately converts the `Half[]` data to `double[]` data. I'm trying to think this through, and I think it should be fine. The main concern I have is that somehow the process of deserializing a `Half[]`, turning it into a `double[]`, and then reserializing it back to a `Half[]` could lead to the values changing. But a `Half` should be able to be `double` converted losslessly, and the conversion back to a `Half` should be stable. So that should be fine.

I don't think having the `toDoubleVectorSupplier` be a memoized thing makes sense, though. It looks like the main reason you need to do that is to avoid infinite mutual recursion between the `DoubleVector` and `HalfVector` constructors as they both would want to construct and cache the other one. I could see trying to get around that by either removing the caching/memoization entirely and just making `toHalfVector`/`toDoubleVector` return a fresh object (given that these aren't that expensive to construct--maybe with some trickery to share memoized hash codes), or having a constructor variant that takes a pre-existing `HalfVector` or `DoubleVector` and supply the existing object. (That will create a circular dependency in the GC graph, but I believe modern Java GCs actually handle that fairly well.)

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+                    .mapToObj(String::valueOf)
+                    .collect(Collectors.joining(",")) + "]";
+        }
+    }
+
+    /**
+     * A vector class encoding a vector over half components. Conversion to {@link DoubleVector} is supported and
+     * memoized.
+     */
+    public static class HalfVector extends Vector {
```
alecgrieser, line 249: Teamscale wants these classes to be given their own files, and I feel like I agree. Or at least, I don't see any strong case not to follow that advice

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Vector.java
```diff
@@ -0,0 +1,507 @@
...
+        @Nonnull
+        protected abstract N toComponent(@Nonnull ByteBuffer byteBuffer);
+
+        @Nonnull
+        protected abstract T toTarget(@Nonnull N[] components);
+
+
+        @Nullable
+        @Override
+        protected T computeNext() {
```
alecgrieser, line 416: It might be good to document the file format here. Reverse engineering from the parsing logic, it sounds like it's a 4 byte header containing the number of dimensions _d_ (as a little Endian integer) followed by _d_ 4 byte-values, which may either be little Endian integers or little Endian floats

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/StorageAdapter.java
```diff
@@ -0,0 +1,455 @@
...
+     */
+    @Nonnull
+    static Vector vectorFromTuple(final Tuple vectorTuple) {
+        return vectorFromBytes(vectorTuple.getBytes(0));
+    }
+
+    /**
+     * Creates a {@link Vector} from a byte array.
+     * <p>
+     * This method interprets the input byte array by interpreting the first byte of the array as the precision shift.
```
alecgrieser, line 257: Interesting. What was the reasoning for including the precision shift as the first byte rather than just containing the vector data? Were you thinking that it would be better to make the data format somewhat self-describing rather than relying on, say, external meta-data to know whether the data should be interpreted as halfs or floats?

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/StorageAdapter.java
```diff
@@ -0,0 +1,455 @@
...
+     */
+    @Nonnull
+    static byte[] bytesFromVector(@Nonnull final Vector.HalfVector halfVector) {
+        final byte[] vectorBytes = new byte[1 + 2 * halfVector.size()];
+        vectorBytes[0] = (byte)halfVector.precisionShift();
+        for (int i = 0; i < halfVector.size(); i ++) {
+            final byte[] componentBytes = bytesFromShort(Half.halfToShortBits(Half.valueOf(halfVector.getComponent(i))));
+            final int offset = 1 + (i << 1);
+            vectorBytes[offset] = componentBytes[0];
+            vectorBytes[offset + 1] = componentBytes[1];
```
alecgrieser, line 350: This is going to create a bunch of tiny `byte[]`s. It might be better to have the `bytesFromShort` method take the final `byte[]` as an input along with an offset, and then it can go ahead and fill in the final array as it goes.

You could also use a `ByteBuffer` for this purpose. If the order is set to `BIG_ENDIAN`, then `putShort` will have the same effect here

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/StorageAdapter.java
```diff
@@ -0,0 +1,455 @@
...
+            final byte[] componentBytes = bytesFromLong(Double.doubleToLongBits(doubleVector.getComponent(i)));
+            final int offset = 1 + (i << 3);
+            vectorBytes[offset] = componentBytes[0];
+            vectorBytes[offset + 1] = componentBytes[1];
+            vectorBytes[offset + 2] = componentBytes[2];
+            vectorBytes[offset + 3] = componentBytes[3];
+            vectorBytes[offset + 4] = componentBytes[4];
+            vectorBytes[offset + 5] = componentBytes[5];
+            vectorBytes[offset + 6] = componentBytes[6];
+            vectorBytes[offset + 7] = componentBytes[7];
```
alecgrieser, line 378: This definitely feels like it should use a `ByteBuffer`. I think you could use `ByteBuffer.putLong(Double.doubleToLongBits(doubleVector.getComponent(i)))` in a similar fashion to this, but you could also just use `ByteBuffer.putDouble`, I think

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/StorageAdapter.java
```diff
@@ -0,0 +1,455 @@
...
+     * <p>
+     * This method reads two consecutive bytes from the {@code bytes} array, starting at the given {@code offset}. The
+     * byte at {@code offset} is treated as the most significant byte (MSB), and the byte at {@code offset + 1} is the
+     * least significant byte (LSB).
+     * @param bytes the source byte array from which to read the short.
+     * @param offset the starting index in the byte array.
+     * @return the short value constructed from the two bytes.
+     */
+    static short shortFromBytes(final byte[] bytes, final int offset) {
+        int high = bytes[offset] & 0xFF;   // Convert to unsigned int
```
alecgrieser, line 394: Does it actually matter if we convert it to an unsigned int or not? We're about to shift it up by 8 bits anyway, which means the sign bit of the original byte will wind up in the sign bit of the `short` anyway

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/StorageAdapter.java
```diff
@@ -0,0 +1,455 @@
...
+     * vector.
+     * @param vectorBytes the non-null byte array to convert. The length of this array must be even, as each pair of
+     *        bytes represents a single {@link Half} component.
+     * @return a new {@link Vector.HalfVector} instance created from the byte array.
+     */
+    @Nonnull
+    static Vector.HalfVector halfVectorFromBytes(@Nonnull final byte[] vectorBytes, final int offset, final int numDimensions) {
+        final Half[] vectorHalfs = new Half[numDimensions];
+        for (int i = 0; i < numDimensions; i ++) {
+            vectorHalfs[i] = Half.shortBitsToHalf(shortFromBytes(vectorBytes, offset + (i << 1)));
```
alecgrieser, line 296: Could this not be replaced with a `ByteBuffer` call? I'm thinking something like:

```java
ByteBuffer buffer = ByteBuffer.wrap(vectorBytes, offset, numDimensions << 1)
        .order(ByteOrder.BIG_ENDIAN);
for (int i = 0; i < numDimensions; i++) {
    vectorHalfs[i] = Half.shortBitsToHalf(buffer.getShort());
}
```

Then I think you can remove some of the new private methods here for converting data to and from shorts. And, for that matter, the same trick can be played with `double`s

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Metric.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Represents the Euclidean distance metric.
+     * <p>
+     * This metric calculates the "ordinary" straight-line distance between two points
+     * in Euclidean space. The distance is the square root of the sum of the
+     * squared differences between the corresponding coordinates of the two points.
+     */
+    class EuclideanMetric implements Metric {
```
alecgrieser, line 120: Should this override `comparativeDistance` and return the `EuclideanSquareMetric`? I may not have a hang over how those methods are supposed to interact, but as _d_(_v_<sub>1</sub>, _v_<sub>2</sub>) < _d_(_v_<sub>3</sub>, _v_<sub>4</sub>) if and only if _d_(_v_<sub>1</sub>, _v_<sub>2</sub>)<sup>2</sup> < _d_(_v_<sub>3</sub>, _v_<sub>4</sub>)<sup>2</sup>, if the idea is that we'll use the `comparativeDistance` when doing comparisons, then it may make sense to use the square Euclidean metric for that purpose. Unless we want to leverage the triangle equality (which is not available with the Euclidean square "metric"): https://math.stackexchange.com/questions/1588402/is-squared-euclidean-distance-a-metric

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Metric.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Represents the squared Euclidean distance metric.
+     * <p>
+     * This metric calculates the sum of the squared differences between the coordinates of two vectors, defined as
+     * {@code sum((p_i - q_i)^2)}. It is computationally less expensive than the standard Euclidean distance because it
+     * avoids the final square root operation.
+     * <p>
+     * This is often preferred in algorithms where comparing distances is more important than the actual distance value,
+     * such as in clustering algorithms, as it preserves the relative ordering of distances.
```
alecgrieser, line 143: Probably worth noting that this does not obey the triangle inequality

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Metric.java
```diff
@@ -0,0 +1,242 @@
...
+     * {@code 0.0d} and {@code 2.0d} that corresponds to {@code 1 - cos(v1, v2)}, meaning that if {@code v1 == v2},
+     * the distance is {@code 0} while if {@code v1} is orthogonal to {@code v2} it is {@code 1}.
+     * @see Metric.CosineMetric
+     */
+    class CosineMetric implements Metric {
+        @Override
+        public double distance(@Nonnull final double[] vector1, @Nonnull final double[] vector2) {
+            Metric.validate(vector1, vector2);
+
+            double dotProduct = 0.0;
```
alecgrieser, line 184: Should this share the dot product logic with `DotProductMetric`? I suppose keeping this contained makes us more likely to leverage the cache when computing the norms

fdb-extensions/src/main/java/com/apple/foundationdb/async/hnsw/Metric.java
```diff
@@ -0,0 +1,242 @@
...
+            return this.getClass().getSimpleName();
+        }
+    }
+
+    /**
+     * Dot product similarity.
+     * <p>
+     * This metric calculates the inverted dot product of two vectors. It is not a true metric as the dot product can
+     * be positive at which point the distance is negative. In order to make callers aware of this fact, this distance
+     * only allows {@link Metric#comparativeDistance(double[], double[])} to be called.
```
alecgrieser, line 214: I take it the salient point here is that the dot product is negative, whereas the other metrics are always non-negative. But I think only the Manhattan and Euclidean Metrics (both extensions of the [Minkowski distance](https://en.wikipedia.org/wiki/Minkowski_distance)) are actually metrics, as the others do not obey the triangle inequality. I think that makes them semimetrics (see: https://en.wikipedia.org/wiki/Metric_space#Semimetrics)

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/VectorTest.java
```diff
@@ -0,0 +1,79 @@
...
+        final Vector deserializedVector = StorageAdapter.vectorFromBytes(randomVector.getRawData());
+        Assertions.assertThat(deserializedVector).isInstanceOf(Vector.DoubleVector.class);
+        Assertions.assertThat(deserializedVector).isEqualTo(randomVector);
+    }
+
+    @Nonnull
+    static Vector.HalfVector createRandomHalfVector(@Nonnull final Random random, final int dimensionality) {
+        final Half[] components = new Half[dimensionality];
+        for (int d = 0; d < dimensionality; d ++) {
+            // don't ask
```
alecgrieser, line 64: Don't ask?

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/VectorTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import javax.annotation.Nonnull;
+import java.util.Random;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+public class VectorTest {
+    private static Stream<Long> randomSeeds() {
+        return LongStream.generate(() -> new Random().nextLong())
```
alecgrieser, line 35: It would be good to use a set of fixed seeds here for the purpose of having reliable tests. There's already a utility method that allows us to use a set of random seeds when we're in randomized mode (which is to say, nightlies). For example:

https://github.com/FoundationDB/fdb-record-layer/blob/b461c53addf5c0489747e66a9ec38a1817033585/fdb-extensions/src/test/java/com/apple/foundationdb/util/StringUtilsTest.java#L295-L297

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/MetricTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MetricTest {
+    private final Metric.ManhattanMetric manhattanMetric = new Metric.ManhattanMetric();
+    private final Metric.EuclideanMetric euclideanMetric = new Metric.EuclideanMetric();
+    private final Metric.EuclideanSquareMetric euclideanSquareMetric = new Metric.EuclideanSquareMetric();
+    private final Metric.CosineMetric cosineMetric = new Metric.CosineMetric();
+    private Metric.DotProductMetric dotProductMetric;
```
alecgrieser, line 33: Two things:

1. Should this reference the `Metric` objects from the `Metrics` enum? It seems weird to have that enum but still create instances of the metric implementations on their own
2. Why is `dotProductMetric` handled differently from all of the other metrics

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/MetricTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @BeforeEach
+
+    public void setUp() {
+        dotProductMetric = new Metric.DotProductMetric();
+    }
+
+    @Test
+    public void manhattanMetricDistanceWithIdenticalVectorsShouldReturnZeroTest() {
```
alecgrieser, line 42: Could this be parameterized on the Metric type? I believe all of the metrics should return zero on identical vectors except for dot product. The `EnumSource` argument provider can be set up to exclude values, which we could do to skip that particular one

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/MetricTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        // Act
+        double actualDistance = euclideanMetric.distance(vector1, vector2);
+
+        // Assert
+        assertEquals(expectedDistance, actualDistance, 0.00001);
+    }
+
+    @Test
+    public void euclideanMetricDistanceWithDifferentPositiveVectorsShouldReturnCorrectDistanceTest() {
```
alecgrieser, line 84: This correct distance metric could also be made generic on the different types, though it would require a more involved arguments provider. Probably something like:

```java
static Stream<Arguments> metricAndExpectedDistance() {
    // Distance between (1.0, 2.0) and (4.0, 6.0)
    return Stream.of(
        Arguments.of(Metrics.MANHATTAN, 7.0), // |1 - 4| + |2 - 6| = 7
        Arguments.of(Metrics.EUCLIDEAN, 5.0),  // sqrt((1-4)^2 + (2-6)^2) = sqrt(9 + 16) = 5.0
        Arguments.of(Metrics.SQUARE_EUCLIDEAN, 25.0), // (1-4)^2 + (2-6)^2 = 9 + 16 = 25
        // etc
    );
}
```

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/MetricTest.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+package com.apple.foundationdb.async.hnsw;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MetricTest {
```
alecgrieser, line 28: This should also probably add tests of the comparative distance function, to be sure. I could also see us adding a test of the triangle inequality, for the two that support it.

fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/VectorTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import javax.annotation.Nonnull;
+import java.util.Random;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+public class VectorTest {
```
alecgrieser, line 33: This would probably be a good spot to add a few more unit tests of the `Vector` class, including things like the hash code and converting between `HalfVector` and `DoubleVector`

normen662:
Went in as part of #3691.

PR #3603, opened 2025-09-17
fixing test case for mixed-mode tests

normen662:
This adds a minimum supported version (for mixed mode testing) in a yamsql test case. This test broke the mixed mode testing affecting the releases 4.5.13.0 and 4.5.14.0 until it was discovered. The PR checks did not fail though it should have failed. Opened https://github.com/FoundationDB/fdb-record-layer/issues/3604 to track the test case issue.

### DIFF
--- a/yaml-tests/src/test/resources/aggregate-index-tests.yamsql
+++ b/yaml-tests/src/test/resources/aggregate-index-tests.yamsql
@@ -289,6 +289,7 @@ test_block:
       - result: [{!l 15}]
     -
       - query: select col1 as g, sum(col2) as s from T1 use index (mv1) where col1 in (10, 20) group by col1
+      - supported_version: 4.5.13.0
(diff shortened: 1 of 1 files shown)


yaml-tests/src/test/resources/aggregate-index-tests.yamsql
```diff
@@ -289,6 +289,7 @@ test_block:
       - result: [{!l 15}]
     -
       - query: select col1 as g, sum(col2) as s from T1 use index (mv1) where col1 in (10, 20) group by col1
+      - supported_version: !current_version
```
alecgrieser:
```suggestion
      - supported_version: 4.5.13.0
```

Right? It also might make sense to clarify in a comment here that this query failed to plan prior to `4.5.13.0`, which I believe is the case, because it would otherwise be incorrect to paper over a mixed mode failure like this

normen662: done

PR #3606, opened 2025-09-17
Make use of key store its own key manager

MMcM:
Following on from #3522, extend how key store is used by YAML tests to allow for multiple keys in the connection options. Any one of these keys can be read back.

The existing YAML tests continue to work with this new implementation, as the options are known to the old serializer implementation and the earlier PR's implementation of a single key.

A new YAML test shows the use of multiple keys, restricted to the current version, as it involves a new option.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
@@ -25,7 +25,7 @@ import com.apple.foundationdb.record.RecordCoreArgumentException;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.io.FileInputStream;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
@@ -60,7 +60,7 @@ public class KeyStoreSerializationKeyManager implements SerializationKeyManager 
                                               @Nonnull String cipherName, @Nonnull SecureRandom secureRandom) {
         this.keyStore = keyStore;
         this.keyEntryPassword = keyEntryPassword;
-        this.keyEntryAliases = keyEntryAliases;
+        this.keyEntryAliases = List.copyOf(keyEntryAliases);
         this.defaultKeyNumber = defaultKeyNumber;
         this.cipherName = cipherName;
         this.secureRandom = secureRandom;
@@ -193,10 +193,7 @@ public class KeyStoreSerializationKeyManager implements SerializationKeyManager 
             }
             final KeyStore keyStore;
             try {
-                keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
-                try (FileInputStream fis = new FileInputStream(keyStoreFileName)) {
-                    keyStore.load(fis, keyStorePassword.toCharArray());
-                }
+                keyStore = KeyStore.getInstance(new File(keyStoreFileName), keyStorePassword.toCharArray());
             } catch (FileNotFoundException ex) {
                 throw new RecordCoreArgumentException("Key store not found", ex);
             } catch (GeneralSecurityException | IOException ex) {
@@ -214,6 +211,9 @@ public class KeyStoreSerializationKeyManager implements SerializationKeyManager 
                 keyEntryAliases = List.of(defaultKeyEntryAlias);
                 defaultKeyNumber = 0;
             } else if (defaultKeyEntryAlias == null) {
+                if (keyEntryAliases.isEmpty()) {
+                    throw new RecordCoreArgumentException("need at least one key alias");
+                }
                 defaultKeyNumber = keyEntryAliases.size() - 1;
             } else {
                 defaultKeyNumber = keyEntryAliases.indexOf(defaultKeyEntryAlias);
--- a/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
+++ b/fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
@@ -25,6 +25,7 @@ import com.apple.foundationdb.relational.api.exceptions.ErrorCode;
 import com.apple.foundationdb.relational.api.options.CollectionContract;
 import com.apple.foundationdb.relational.api.options.OptionContract;
 import com.apple.foundationdb.relational.api.options.OptionContractWithConversion;
+import com.apple.foundationdb.relational.api.options.OrderedCollectionContract;
 import com.apple.foundationdb.relational.api.options.RangeContract;
 import com.apple.foundationdb.relational.api.options.TypeContract;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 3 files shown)


MMcM:
Rebased to parent as of da67b4686af0ad6a5d3111c6002dc4976034ae94.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+    private final String cipherName;
+    @Nonnull
+    private final SecureRandom secureRandom;
+
+    protected KeyStoreSerializationKeyManager(@Nonnull KeyStore keyStore, @Nonnull KeyStore.ProtectionParameter keyEntryPassword,
+                                              @Nonnull List<String> keyEntryAliases, int defaultKeyNumber,
+                                              @Nonnull String cipherName, @Nonnull SecureRandom secureRandom) {
+        this.keyStore = keyStore;
+        this.keyEntryPassword = keyEntryPassword;
+        this.keyEntryAliases = keyEntryAliases;
```
ScottDugas: Perhaps `List.copyOf` would be appropriate. If the original list was created with `List.of` (or already had `copyOf`), it becomes a no-op.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+import java.security.KeyStore;
+import java.security.SecureRandom;
+import java.util.List;
+import java.util.Random;
+
+
+/**
+ * A {@link SerializationKeyManager} to uses a {@link KeyStore}.
+ * <p>
+ * Key numbers are mapped to and from key names, which are looked up in the store.
```
ScottDugas, line 42: The PR is labeled as `Testing Improvement`, but there is no indication here, or in the options that this is for tests only. Perhaps this is just a matter of updating the label.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+import java.security.GeneralSecurityException;
+import java.security.Key;
+import java.security.KeyStore;
+import java.security.SecureRandom;
+import java.util.List;
+import java.util.Random;
+
+
+/**
+ * A {@link SerializationKeyManager} to uses a {@link KeyStore}.
```
ScottDugas:
```suggestion
 * A {@link SerializationKeyManager} that uses a {@link KeyStore}.
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+        public KeyStoreSerializationKeyManager build() {
+            if (keyStoreFileName == null) {
+                throw new RecordCoreArgumentException("must specify key store file name");
+            }
+            if (keyStorePassword == null) {
+                keyStorePassword = "";
+            }
+            final KeyStore keyStore;
+            try {
+                keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
```
ScottDugas: Would it make sense to use:
```
 KeyStore ks = KeyStore.getInstance(new File(keyStoreFileName), keyStorePassword.toCharArray());
```

Then you don't need to open a `FileInputStream`, and call `load`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+            }
+            final KeyStore.ProtectionParameter keyEntryProtection = new KeyStore.PasswordProtection(keyEntryPassword.toCharArray());
+            if (keyEntryAliases == null && defaultKeyEntryAlias == null) {
+                throw new RecordCoreArgumentException("must specify key alias list or single default alias");
+            }
+            final int defaultKeyNumber;
+            if (keyEntryAliases == null) {
+                keyEntryAliases = List.of(defaultKeyEntryAlias);
+                defaultKeyNumber = 0;
+            } else if (defaultKeyEntryAlias == null) {
```
ScottDugas, line 213: Should this throw a valuable error if `keyEntryAliases` is empty?

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
```diff
@@ -549,6 +555,7 @@ private static Map<Name, List<OptionContract>> makeContracts() {
         data.put(Name.ENCRYPT_WHEN_SERIALIZING, List.of(TypeContract.booleanType()));
         data.put(Name.ENCRYPTION_KEY_STORE, List.of(TypeContract.nullableStringType()));
         data.put(Name.ENCRYPTION_KEY_ENTRY, List.of(TypeContract.nullableStringType()));
+        data.put(Name.ENCRYPTION_KEY_ENTRY_LIST, List.of(new CollectionContract<>(TypeContract.stringType())));
```
ScottDugas: Does this contract cover ordering? Do we care about that from a TypeContract perspective?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
```diff
@@ -0,0 +1,231 @@
...
+                }
+            } catch (FileNotFoundException ex) {
+                throw new RecordCoreArgumentException("Key store not found", ex);
+            } catch (GeneralSecurityException | IOException ex) {
+                throw new RecordCoreArgumentException("Key store loading failed", ex);
+            }
+            if (keyEntryPassword == null) {
+                keyEntryPassword = keyStorePassword;
+            }
+            final KeyStore.ProtectionParameter keyEntryProtection = new KeyStore.PasswordProtection(keyEntryPassword.toCharArray());
```
ScottDugas, line 205: Yeah, probably good to move this closer

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
MMcM: But then something else is too far away, right? Right now, it is constructing the arguments in the order in which they are passed to the constructor.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/Options.java
MMcM: I guess it's better to be explicit about requiring that the collection be ordered. I added an `OrderedCollectionContract` for it.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
MMcM: I guess it depends on whether you think those labels are the motivation or the scope of changes. I added another one, just to be as clear as possible.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
MMcM: Yes, that sounds right. Added.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
MMcM: Fixed.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
MMcM: Yes, that is simpler.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/KeyStoreSerializationKeyManager.java
ScottDugas: I would lean more towards the scope of the changes.

PR #3607, opened 2025-09-17
Make Lucene serialization able to encrypt, … 

MMcM:
… using the same serialization key manager as record serialization.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
@@ -25,6 +25,7 @@ import com.apple.foundationdb.annotation.API;
 /**
  * Information on intended / found serialization format: compressed and/or encrypted.
  */
+@API(API.Status.INTERNAL)
 public class CompressedAndEncryptedSerializerState {
     private boolean compressed;
     private boolean encrypted;
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
@@ -1,0 +1,61 @@
+/*
+ * CompressedAndEncryptedSerializerState.java
+ *
+ * This source file is part of the FoundationDB open source project
+ *
+ * Copyright 2015-2025 Apple Inc. and the FoundationDB project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.apple.foundationdb.record.provider.common;
+
+import com.apple.foundationdb.annotation.API;
+
+/**
+ * Information on intended / found serialization format: compressed and/or encrypted.
+ */
+public class CompressedAndEncryptedSerializerState {
+    private boolean compressed;
+    private boolean encrypted;
+    private int keyNumber;
+
+    public CompressedAndEncryptedSerializerState() {
+        this.compressed = false;
+        this.encrypted = false;
+    }
+
+    public boolean isCompressed() {
+        return compressed;
+    }
+
+    public void setCompressed(boolean compressed) {
+        this.compressed = compressed;
+    }
+
+    public boolean isEncrypted() {
+        return encrypted;
+    }
+
+    public void setEncrypted(boolean encrypted) {
+        this.encrypted = encrypted;
+    }
+
+    public int getKeyNumber() {
+        return keyNumber;
+    }
+
+    public void setKeyNumber(final int keyNumber) {
+        this.keyNumber = keyNumber;
+    }
+}
--- a/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
+++ b/fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
@@ -24,6 +24,7 @@ import com.apple.foundationdb.record.RecordCoreException;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.lucene.LuceneLogMessageKeys;
 import com.apple.foundationdb.record.provider.common.CipherPool;
+import com.apple.foundationdb.record.provider.common.CompressedAndEncryptedSerializerState;
 import com.apple.foundationdb.record.provider.common.SerializationKeyManager;
 import org.apache.lucene.codecs.compressing.CompressionMode;
 import org.apache.lucene.codecs.compressing.Compressor;
@@ -59,54 +60,18 @@ public class LuceneSerializer {
         this.keyManager = keyManager;
     }
 
-    private static class EncodingState {
-        private boolean compressed;
-        private boolean encrypted;
-        private int keyNumber;
-
-        private EncodingState() {
-            this.compressed = false;
-            this.encrypted = false;
-        }
-
-        boolean isCompressed() {
-            return compressed;
-        }
-
-        void setCompressed(boolean compressed) {
-            this.compressed = compressed;
-        }
-
-        boolean isEncrypted() {
-            return encrypted;
-        }
-
-        void setEncrypted(boolean encrypted) {
-            this.encrypted = encrypted;
-        }
-
-        public int getKeyNumber() {
-            return keyNumber;
-        }
-
-        public void setKeyNumber(final int keyNumber) {
-            this.keyNumber = keyNumber;
-        }
-    }
-
     @Nullable
     public byte[] encode(@Nullable byte[] data, boolean compress, boolean encrypt) {
         if (data == null) {
             return null;
         }
 
-        final EncodingState state = new EncodingState();
+        final CompressedAndEncryptedSerializerState state = new CompressedAndEncryptedSerializerState();
         long prefix = 0;
         if (compress) {
             prefix |= ENCODING_COMPRESSED;
             state.setCompressed(true);
         }
-        // Encryption will be supported in future
         if (encrypt) {
             if (keyManager == null) {
                 throw new RecordCoreException("cannot encrypt Lucene blocks without keys");
@@ -153,7 +118,7 @@ public class LuceneSerializer {
                     .addLogInfo(LuceneLogMessageKeys.DATA_VALUE, data);
         }
 
-        final EncodingState state = new EncodingState();
+        final CompressedAndEncryptedSerializerState state = new CompressedAndEncryptedSerializerState();
         byte[] decoded;
         try {
             final ByteArrayDataInput encodedDataInput = new ByteArrayDataInput(data);
@@ -184,7 +149,7 @@ public class LuceneSerializer {
     }
 
     @Nonnull
-    private static byte[] compressIfNeeded(@Nonnull EncodingState state, @Nonnull ByteBuffersDataOutput encodedDataOutput,
+    private static byte[] compressIfNeeded(@Nonnull CompressedAndEncryptedSerializerState state, @Nonnull ByteBuffersDataOutput encodedDataOutput,
                                            @Nonnull byte[] uncompressedData, int prefixLength)
             throws IOException {
         if (!state.isCompressed()) {
@@ -207,7 +172,7 @@ public class LuceneSerializer {
         }
     }
 
-    private static byte[] fallBackToUncompressed(@Nonnull EncodingState state, @Nonnull byte[] originalData,
+    private static byte[] fallBackToUncompressed(@Nonnull CompressedAndEncryptedSerializerState state, @Nonnull byte[] originalData,
                                                  @Nonnull byte[] encodedData, int prefixLength) {
         final byte[] encoded = new byte[originalData.length + prefixLength];
         System.arraycopy(encodedData, 0, encoded, 0, prefixLength);
@@ -217,7 +182,7 @@ public class LuceneSerializer {
         return encoded;
     }
 
-    private void decompressIfNeeded(@Nonnull EncodingState state, @Nonnull ByteArrayDataInput encodedDataInput)
+    private void decompressIfNeeded(@Nonnull CompressedAndEncryptedSerializerState state, @Nonnull ByteArrayDataInput encodedDataInput)
             throws IOException {
         if (!state.isCompressed()) {
             return;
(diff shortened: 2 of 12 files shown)


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -46,55 +50,85 @@ public class LuceneSerializer {
...
-            return compressed;
+        public int getKeyNumber() {
+            return keyNumber;
         }
 
-        boolean isEncrypted() {
-            return encrypted;
+        public void setKeyNumber(final int keyNumber) {
+            this.keyNumber = keyNumber;
         }
```
ScottDugas: It does look like `TransformedRecordSerializerState` is a super set of this class. Would it make sense to move this out, and have `TransformedRecordSerializerState` have one of these.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -118,81 +153,137 @@ public static byte[] decode(@Nullable byte[] data) {
...
     }
 
-    private static byte[] fallBackToUncompressed(@Nonnull byte[] compressedData, @Nonnull byte[] originalData,
-                                                 @Nonnull EncodingState state, int offset) {
+    private static byte[] fallBackToUncompressed(@Nonnull EncodingState state, @Nonnull byte[] originalData,
+                                                 @Nonnull byte[] encodedData, int prefixLength) {
+        final byte[] encoded = new byte[originalData.length + prefixLength];
+        System.arraycopy(encodedData, 0, encoded, 0, prefixLength);
+        encoded[0] &= ~ENCODING_COMPRESSED;
```
ScottDugas, line 198: Is this guaranteed to be in byte`[0]` (i.e. what if `prefixLength>1`)

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1399,4 +1405,35 @@ void testQueryWithManyDocuments() {
             assertPrimaryKeys("text:morningstart", false, Set.of());
         }
     }
+
+    @ParameterizedTest
+    @RandomSeedSource
+    void encrypted(long seed) throws Exception {
+        final RollingTestKeyManager rollingTestKeyManager = new RollingTestKeyManager(seed);
+        keyManager = rollingTestKeyManager;
+        // Save in many transactions to get more segments and so more blocks.
```
ScottDugas, line 1413: Probably also a good idea to test having a single segment with multiple blocks

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1399,4 +1405,35 @@ void testQueryWithManyDocuments() {
             assertPrimaryKeys("text:morningstart", false, Set.of());
         }
     }
+
+    @ParameterizedTest
+    @RandomSeedSource
+    void encrypted(long seed) throws Exception {
```
ScottDugas, line 1410: Perhaps worth having some of the tests in LuceneIndexMaintenanceTest chose to randomly encrypt and-or compress.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
```diff
@@ -1399,4 +1405,35 @@ void testQueryWithManyDocuments() {
...
+        final RollingTestKeyManager rollingTestKeyManager = new RollingTestKeyManager(seed);
+        keyManager = rollingTestKeyManager;
+        // Save in many transactions to get more segments and so more blocks.
+        for (TestRecordsTextProto.SimpleDocument doc : DOCUMENTS) {
+            try (FDBRecordContext context = openContext()) {
+                openRecordStore(context);
+                recordStore.saveRecord(doc);
+                commit(context);
```
ScottDugas, line 1418: This should be merging in-line when you commit. Would it be worthwhile to test merging separately? Perhaps that's best covered by LuceneIndexMaintenanceTest.

ScottDugas (commented):
I didn't get to really think about whether the encryption code is doing the right thing. It looks pretty similar to the record encryption though.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: Yes, I believe so. Varints are 7-bit little-endian, with a "continuation" bit to know when to stop. So, the compressed flag bit, which is in the low bits, is always in the first byte, no matter how many higher bits there are in the following bytes.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: Refactored to share the base class with just the intended / found format.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: I added these serialization states as an argument to the existing tests that took a bunch of arguments.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: I think these documents write on the order of 9 blocks already. I see that we don't actually have a metric for that, so I added one. It's hard to assert about that, though, because the `commit` method in the test fixture clears the timer and that's when the work gets done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
```diff
@@ -0,0 +1,61 @@
...
+    public void setEncrypted(boolean encrypted) {
+        this.encrypted = encrypted;
+    }
+
+    public int getKeyNumber() {
+        return keyNumber;
+    }
+
+    public void setKeyNumber(final int keyNumber) {
+        this.keyNumber = keyNumber;
```
ohadzeliger, line 60: Is `0` a special key number (if the setter is not called and it remains uninitialized)?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -46,55 +51,49 @@ public class LuceneSerializer {
...
-        boolean isEncrypted() {
-            return encrypted;
-        }
+    public LuceneSerializer(@Nullable SerializationKeyManager keyManager) {
+        this.keyManager = keyManager;
     }
 
     @Nullable
-    public static byte[] encode(@Nullable byte[] data, boolean compress, boolean encrypt) {
+    public byte[] encode(@Nullable byte[] data, boolean compress, boolean encrypt) {
```
ohadzeliger: Would it make sense to move the `compress` and `encrypt` params to be fields of the class instead?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -46,55 +51,49 @@ public class LuceneSerializer {
...
         }
-        // Encryption will be supported in future
-        if (state.isEncrypted()) {
-            code = code | ENCODING_ENCRYPTED;
+        if (encrypt) {
+            if (keyManager == null) {
+                throw new RecordCoreException("cannot encrypt Lucene blocks without keys");
+            }
+            int key = keyManager.getSerializationKey();
+            prefix |= ENCODING_ENCRYPTED | ((key & 0xFFFFFFFFL) << ENCRYPTION_KEY_SHIFT);
```
ohadzeliger, line 98: Based on the design, I believe this is the same as other serializers - would it make sense to reuse the code?

ohadzeliger, line 98: Should we limit the number of keys supported, so that we can leave a few bits at the end of the `long` for future use?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -118,81 +118,138 @@ public static byte[] decode(@Nullable byte[] data) {
...
+            throw new RecordCoreException("Lucene data decoding failure", ex);
+        }
 
-        byte[] decoded = decompressIfNeeded(compressed, data, 1);
         if (LOGGER.isTraceEnabled()) {
             LOGGER.trace(KeyValueLogMessage.of("Decoded lucene data",
-                    LuceneLogMessageKeys.COMPRESSED_EVENTUALLY, compressed,
-                    LuceneLogMessageKeys.ENCRYPTED_EVENTUALLY, encrypted,
+                    LuceneLogMessageKeys.COMPRESSED_EVENTUALLY, state.isCompressed(),
+                    LuceneLogMessageKeys.ENCRYPTED_EVENTUALLY, state.isCompressed(),
```
ohadzeliger:
```suggestion
                    LuceneLogMessageKeys.ENCRYPTED_EVENTUALLY, state.isEncrypted(),
```

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
```diff
@@ -326,11 +336,16 @@ void manyDocument(boolean isGrouped,
                 "seed", seed,
                 "loopCount", loopCount));
 
-        final RecordLayerPropertyStorage contextProps = RecordLayerPropertyStorage.newBuilder()
+        final RecordLayerPropertyStorage.Builder contextPropsBuilder = RecordLayerPropertyStorage.newBuilder()
                 .addProp(LuceneRecordContextProperties.LUCENE_REPARTITION_DOCUMENT_COUNT, repartitionCount)
                 .addProp(LuceneRecordContextProperties.LUCENE_MAX_DOCUMENTS_TO_MOVE_DURING_REPARTITIONING, dataModel.nextInt(1000) + repartitionCount)
                 .addProp(LuceneRecordContextProperties.LUCENE_MERGE_SEGMENTS_PER_TIER, (double)dataModel.nextInt(10) + 2) // it must be at least 2.0
-                .build();
+                .addProp(LuceneRecordContextProperties.LUCENE_INDEX_COMPRESSION_ENABLED, compressed)
```
ohadzeliger, line 349: Maybe also add a test for failing to decrypt?
Either by corrupting the data or a bogus KeyManager?

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: It isn't really the same because the record code has to deal with the legacy case of no prefix byte. Also, this uses Lucene's varint implementation, which, even though it's compatible, seemed appropriate.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
MMcM: Not particularly at this level; if encryption is not enabled, it does remain unset, but should also be unused. Did you have something in mind to make this clearer?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: I moved `compressionEnabled` and `encryptionEnabled` from `FDBDirectory` into the serializer.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: Beyond the `int`'s 32 bits? Aren't there `29` bits left in the `long`?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
MMcM: I added a test where the encryption key changes. Is that along the lines you had in mind?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/common/CompressedAndEncryptedSerializerState.java
ohadzeliger: I thought maybe making the class immutable (values in constructor) or initializing the field to some illegal value to protect ourselves from a bug where it is not initialized.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
ohadzeliger: What I meant is to ask whether limiting the size of the key (to say 16 bits) would leave us with "free" bits in the varInt to be used later. Currently it looks like we take all of the remaining space for the key

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneIndexMaintenanceTest.java
ohadzeliger: Looks good!

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
ohadzeliger: Ah. I see now that `prefix` is a `long` so we have 29 bits left.

ohadzeliger (approved):

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/FDBLuceneQueryTest.java
MMcM: `LuceneIndexMaintenanceTest.manyDocuments` does some explicit merges and is not parametrically encrypted.

PR #3608, opened 2025-09-18
Remove duplicates from IN-list before turning it to an ExplodeExpression

### LINKED ISSUE #3559: Duplicate elements provided to an IN predicate can lead to duplicate results

hazefully:
This PR makes use of the new wrapper value introduced in 4.5.10.0, `ArrayDistinctValue`, to remove duplicate values from the list provided to an IN predicate in SQL before transforming the IN predicate to an ExplodeExpression. This prevents duplicate result values if the planner ends up implementing the query graph as a nested loop join between the ExplodeExpression of the IN list and the left operand of the IN predicate.

Note: This is **a breaking change** that is not compatible with versions of `fdb-record-layer` < 4.5.10.0.

Fixes #3559.

### DIFF
--- a/yaml-tests/src/test/resources/disabled-planner-rewrites/in-predicate.yamsql
+++ b/yaml-tests/src/test/resources/disabled-planner-rewrites/in-predicate.yamsql
@@ -65,38 +65,38 @@ test_block:
       # LONG value matched against IN list of repeated LONG values
       - query: select a, b from ta where b in (1, 1, 1, 1)
       - explain: "ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }"
-      - initialVersionLessThan: !current_version
-      - result: [{A: 8, 1}, {A: 8, 1}, {A: 8, 1}, {A: 8, 1}]
-      # IN expansion in cascades planner distincts the in-list since this version preventing duplicate results with nested loop joins
       - initialVersionAtLeast: !current_version
       - result: [{A: 8, 1}]
+      # Duplicate values in in-lists lead to duplicate results if the IN predicate is planned as a nested loop join in earlier versions
+      - initialVersionLessThan: !current_version
+      - result: [{A: 8, 1}, {A: 8, 1}, {A: 8, 1}, {A: 8, 1}]
     -
       # LONG value matched against IN list of repeated LONG values
       - query: select a, b from ta where b in (1, 2, 1, 3)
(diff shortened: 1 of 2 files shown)


yaml-tests/src/test/resources/in-predicate.yamsql
```diff
@@ -61,33 +62,50 @@ test_block:
       - query: select a, b from ta where b in (1 + 0, 3 + 0, 5, 7)
       - unorderedResult: [{A: 2, 7}, {A: 4, 5}, {A: 6, 3}, {A: 8, 1}]
     -
-      # TODO TODO (IN expansion in cascades planner should distinct the in-list to avoid duplicates)
       # LONG value matched against IN list of repeated LONG values
       - query: select a, b from ta where b in (1, 1, 1, 1)
+      - explain: "ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }"
```
hazefully, line 67: I added an `explain` here and to the other tests checking duplicates since the duplicate bug only happens with nested loop joins, so if anything changes the planning the assumption these tests make about old versions will be incorrect.

yaml-tests/src/test/resources/disabled-planner-rewrites/in-predicate.yamsql
```diff
@@ -61,33 +62,50 @@ test_block:
...
+      - initialVersionAtLeast: !current_version
+      - result: [{A: 8, 1}]
     -
-      # TODO TODO (IN expansion in cascades planner should distinct the in-list to avoid duplicates)
       # LONG value matched against IN list of repeated LONG values
       - query: select a, b from ta where b in (1, 2, 1, 3)
+      - explain: "ISCAN(F1 <,>) | FLATMAP q0 -> { EXPLODE arrayDistinct(promote(@c10 AS ARRAY(LONG))) | FILTER q0.B EQUALS _ AS q1 RETURN (q0.A AS A, q0.B AS B) }"
+      - initialVersionLessThan: !current_version
       - unorderedResult: [{A: 6, 3}, {A: 7, 2}, {A: 8, 1}, {A: 8, 1}]
+      # IN expansion in cascades planner distincts the in-list since this version preventing duplicate results with nested loop joins
```
normen662: You don't have to leave a comment why a thing now works the way it should have worked from the beginning. Maybe you can rewrite these comments in a way that they explain why the old results are in fact the wrong results, instead of explaining why the new result is the new result :-)

normen662 (changes requested):

normen662 (approved):

PR #3611, opened 2025-09-18
Indexing: remove subspaceProvider and subspace arguments to iterating functions

### LINKED ISSUE #3610: Indexing: remove subspaceProvider and subspace arguments to iterating functions
These two parameters, that are used for logging only, can be determined from `common` at the relevant logging functions.

jjezra:
These two parameters, that are used for logging only, can be determined from common at the relevant logging functions.
   This resolves #3610.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
@@ -160,8 +160,8 @@ public class IndexingCommon {
 
         SubspaceProvider subspaceProvider = getRecordStoreBuilder().getSubspaceProvider();
         if (subspaceProvider != null) {
-            logIf(true, keyValues,
-                    subspaceProvider.logKey(), subspaceProvider);
+            keyValues.add(subspaceProvider.logKey());
+            keyValues.add(subspaceProvider);
         }
 
         if (moreKeyValues != null && !moreKeyValues.isEmpty()) {
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingBase.java
@@ -512,14 +512,14 @@ public abstract class IndexingBase {
                 metricsDiff = lastProgressSnapshot == null ? timer : StoreTimer.getDifference(timer, lastProgressSnapshot);
                 lastProgressSnapshot = StoreTimerSnapshot.from(timer);
             }
-            final KeyValueLogMessage message = KeyValueLogMessage.build("Indexer: Built Range",
+            LOGGER.info(KeyValueLogMessage.build("Indexer: Built Range",
                             LogMessageKeys.DELAY, toWait)
                     .addKeysAndValues(additionalLogMessageKeyValues != null ? additionalLogMessageKeyValues : Collections.emptyList())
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
```diff
@@ -158,6 +158,12 @@ LogMessageKeys.TARGET_INDEX_NAME, getTargetIndexesNames(),
                 LogMessageKeys.RECORDS_SCANNED, totalRecordsScanned.get(),
                 LogMessageKeys.INDEXER_ID, indexerId);
 
+        SubspaceProvider subspaceProvider = getRecordStoreBuilder().getSubspaceProvider();
+        if (subspaceProvider != null) {
+            logIf(true, keyValues,
+                    subspaceProvider.logKey(), subspaceProvider);
+        }
```
ScottDugas, line 165: Why use `logIf` here? It seems like you would either want to do:
```suggestion
        logIf(subspaceProvider != null, keyValues,
                subspaceProvider.logKey(), subspaceProvider);
```

Or:
```suggestion
        if (subspaceProvider != null) {
            keyValues.add(subspaceProvider.logKey());
            keyValues.add(subspaceProvider);
        }
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
```diff
@@ -112,17 +111,12 @@ List<Object> indexingLogMessageKeyValues() {
     CompletableFuture<Void> buildIndexInternalAsync() {
         return getRunner().runAsync(context -> openRecordStore(context)
                 .thenCompose( store -> context.getReadVersionAsync()
-                            .thenCompose(vignore -> {
-                                SubspaceProvider subspaceProvider = common.getRecordStoreBuilder().getSubspaceProvider();
-                                // validation checks, if any, will be performed here
```
ScottDugas, line 117: Was this comment intended to be that:
> buildMultiTargetIndex will do any necessary validation checks

Or, the more problematic for this change:
> getSubspaceAsync will do any necessary validation checks

? I'm pretty sure the answer is that `buildMultiTargetIndex` is doing the validation checks, but wanted to confirm.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingCommon.java
jjezra: The logIf was just for uniformity. The first suggestion may cause a null pointer exception as the `subspaceProvider.logKey()` part will be evaluated before calling the `logIf` function. I'll go with the second.

jjezra: √

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexingMultiTargetByRecords.java
jjezra: This comment appeared in the indexers as a hint/reminder where to add validations. I think that appears because `IndexingByIndex`  had a source/target indexes validations at the post `context.getReadVersionAsync` lambda, and then its structure was copied by the other indexers. I did not think that this comment is needed anymore.

ScottDugas (approved):

PR #3614, opened 2025-09-18
prune duplicate intersections.

hatyo:
When the planner generates intersections of access paths, it might end up generating identical intersections unnecessarily.

This happens when the intersection ordering produces the same comparison parts when derived from multiple requested
orderings, which can happen if the intersection ordering, for example, has all of its ordering parts bound except one.

Example:
Given a partition or two access paths `X` and `Y` whose intersection ordering is with `bindings: {a=[=∅],b=[↑],e=[='bar'],d=[=42],c=[='foo']}` and `orderingSet: {a, b, e, d, c}` (note how all of the ordering parts are bound except one).

With the following requested ordering: `a↕, b↑, c↕, d↕, e↕` we end up generating an intersection `X ∩ Y | comparison=[b↑]`.
With a slightly different requested ordering `a↕, c↕, e↕, d↕, b↑` we end up generating the _same_ intersection `X ∩ Y | comparison=[b↑]`.
...

This can happen hundreds of times depending on how many requested orderings are there. Since these intersections are identical, they will be pruned away _anyway_ when we attempted to memoize them in the traversal structure, but this wasted effort puts unnecessary churn on the memoizer.


arnaud-lacurie (approved):

hatyo:
considering that we've merged the same fix `/main` with #3612 and we've decided not to use a patch release specifically to consume this fix in `4.5.10.1`, but instead consume it through the current `4.6.2.0` version. I am closing this PR.

Issue #2974, opened 2024-11-12
Improve the derivation property of the `TempTableScanPlan`


hatyo:
If the temp table scan plan is used as a leg of a recursive union, the derivation of its `Value` should correlate correctly to the initial-state's `Value`.

PR #3616, opened 2025-09-19
Improved `PREORDER`-traversal support for recursive queries

### LINKED ISSUE #3043: More efficient execution of Recursive CTE plans
With #2981 we adopt a level-based traversal execution strategy. The memory footprint of this strategy is dependent on the shape of the hierarchy; so for extremely wide hierarchies, it would consume a lot of memory which is not very efficient.

Under certain conditions, we can adopt a more efficient DFS traversal strategy that is much more efficient for shallow (but extremely wide) hierarchies since the memory consumption in this case is `O(depth)` of the hierarchy. Implementing this requires a new execution runtime that is more or less a special implementation of `FlatMapPipelinedCursor`.

### LINKED ISSUE #3652: `TempTableInsertPlan` incorrect implementation of `equalsWithoutChildren`
`TempTableInsertPlan` always returns `false` in its `equalsWithoutChildren` which is incorrect:

https://github.com/FoundationDB/fdb-record-layer/blob/8316741d9e9aed8d9667b853c926986439c2166f/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/TempTableInsertPlan.java#L177-L179

hatyo:
This PR adds query planning and execution support depth-first search traversal for recursive Common Table Expressions (CTEs), providing an efficient alternative to the existing breadth-first traversal with significantly improved continuation sizes.

Internal DFS Traversal Implementation

  The core DFS algorithm is implemented through two main components:

  - `RecursiveCursor`: A new cursor class that flattens a tree of cursors using stack-based DFS traversal. It maintains a `List<RecursiveNode<T>>` stack for traversal state and implements pre-order traversal logic through the `recursionLoop()` method. The cursor supports database transaction continuations and emits `RecursiveValue<T>` objects containing the original value, (in addition to depth level, and leaf status which are not exposed to customers, yet).
  - `RecordQueryRecursivePlan`: A new physical query plan that integrates the `RecursiveCursor` with the query execution framework. It uses a `ChildCursorFunction` to bind previous result values for recursive references and maps cursor results back to standard query results.

The DFS implementation provides significantly smaller continuation sizes (23-50 bytes) compared to breadth-first traversal (12KB-1MB+), making it more suitable for deep hierarchical queries.

The planner now provides two implementations, when feasible, for recursive union expressions, a recursive union (BFS) and recursive plan (DFS). The cost model considers recursive plan always a winner due to its simplicity.

Enhanced SQL Syntax:
  - New `TRAVERSAL` clause: `TRAVERSAL = {PREORDER | LEVEL | ANY}`
  - Updated recursive CTE syntax: `WITH RECURSIVE (TRAVERSAL = PREORDER) cte_name AS (...)`
  - `PREORDER`: Depth-first search (new)
  - `LEVEL`: Breadth-first traversal (existing)
  - `ANY`: Optimizer choice (default)

  SQL Usage Examples

  Basic organizational hierarchy traversal:

```sql
  WITH RECURSIVE (TRAVERSAL = PREORDER) subordinates AS (
      SELECT id, name, manager_id FROM employees WHERE id = 1
      UNION ALL
      SELECT e.id, e.name, e.manager_id
      FROM subordinates AS s, employees AS e
      WHERE s.id = e.manager_id
  )
  SELECT name FROM subordinates WHERE id != 1
```

This fixes #3043.

It also fixes #3652 by providing an implementation of `equalsWithoutChildren` to `TempTableInsertPlan` fixing structural equality semantics of any plan containing that operator (currently, only the `RecordQueryRecursiveLevelUnionPlan`).


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
```diff
@@ -56,17 +56,52 @@ public class RecursiveUnionExpression implements RelationalExpressionWithChildre
...
+
     @Nonnull
     private final Value resultValue;
 
+    /**
+     * Defines the traversal strategy for recursive union operations in Common Table Expressions (CTEs).
+     * This enum specifies how the recursive leg of a recursive union should traverse and process
+     * the intermediate results during query execution.
+     */
+    public enum Traversal {
```
normen662: Can  you name this something else? There already is a `Traversal` class in the planner which may lead to confusion.

normen662: `StrategyHint` as this seems to be a hint given from SQL?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
```diff
@@ -56,17 +56,52 @@ public class RecursiveUnionExpression implements RelationalExpressionWithChildre
...
                                     @Nonnull final Quantifier recursiveState,
                                     @Nonnull final CorrelationIdentifier tempTableScanAlias,
-                                    @Nonnull final CorrelationIdentifier tempTableInsertAlias) {
+                                    @Nonnull final CorrelationIdentifier tempTableInsertAlias,
+                                    @Nonnull final Traversal traversal) {
         this.initialStateQuantifier = initialState;
         this.recursiveStateQuantifier = recursiveState;
         this.tempTableScanAlias = tempTableScanAlias;
         this.tempTableInsertAlias = tempTableInsertAlias;
+        this.traversal = traversal;
```
normen662: this should be part of equality considerations and hashcode computations.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
```diff
@@ -839,6 +840,12 @@ public Derivations visitSortPlan(@Nonnull final RecordQuerySortPlan sortPlan) {
             return derivationsFromSingleChild(sortPlan);
         }
 
+        @Nonnull
+        @Override
+        public Derivations visitRecursivePlan(@Nonnull final RecordQueryRecursivePlan recursivePlan) {
+            return Derivations.EMPTY; // todo https://github.com/FoundationDB/fdb-record-layer/issues/2974
```
normen662: Can't this just use whatever the level based-plan does? With the other properties I am fine not doing it right now especially if there are already issues filed, but this one is dangerous. Especially, since derivations from below are not propagated at all.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
```diff
@@ -0,0 +1,299 @@
+/*
+ * RecordQueryRecursivePlan.java
```
normen662, line 2: Can you straighten out the namings please?

- logical expression -> `RecursiveUnionExpression`
- strategy is called: `level`
  - physical: `RecordQueryRecursiveUnionPlan`
  - implementor: `ImplementRecursiveUnionPlan`
- strategy is called: preorder
  - physical: `RecordQueryRecursivePlan`
  - implementor: `ImplementRecursiveDfsPlan`

Can we be consistent on the naming? Maybe something like:
- logical expression -> `RecursiveUnionExpression` (as is)
- strategy is called: `level`
  - physical: `RecordQueryRecursiveLevelUnionPlan`
  - implementor: `ImplementRecursiveLevelUnionPlan`
- strategy is called: preorder
  - physical: `RecordQueryRecursivePreorderUnionPlan`
  - implementor: `ImplementRecursivePreorderUnionPlan`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
```diff
@@ -387,6 +388,12 @@ public Boolean visitSortPlan(@Nonnull final RecordQuerySortPlan sortPlan) {
             return storedRecordsFromSingleChild(sortPlan);
         }
 
+        @Nonnull
+        @Override
+        public Boolean visitRecursivePlan(@Nonnull final RecordQueryRecursivePlan recursivePlan) {
+            return false;
```
normen662: This could actually be computed if necessary.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PrimaryKeyProperty.java
```diff
@@ -427,6 +428,12 @@ public Optional<List<Value>> visitSortPlan(@Nonnull final RecordQuerySortPlan so
             return primaryKeyFromSingleChild(sortPlan);
         }
 
+        @Nonnull
+        @Override
+        public Optional<List<Value>> visitRecursivePlan(@Nonnull final RecordQueryRecursivePlan recursivePlan) {
+            return commonPrimaryKeyFromChildren(recursivePlan);
```
normen662: I think this is not correct. You probably need to do this for both the initial and the recursive part and see if they match, then propagate if they do.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Nonnull
+    private static final BindingMatcher<Quantifier.ForEach> mapTempTableQunMatcher = forEachQuantifier(mapPlan(tempTableScanPlan()));
+
+    @Nonnull
+    private static final BindingMatcher<Quantifier.ForEach> tempTableQunMatcher = forEachQuantifier(tempTableScanPlan());
+
+    @Nonnull
+    private static final BindingMatcher<SelectExpression> recursiveInnerSelectMatcher = selectExpression(
+            exactlyInAnyOrder(innerQunMatcher, tempTableQunMatcher.or(mapTempTableQunMatcher)));
```
normen662: Can you match a temp table scan expression or a select (without predicates) over a tempt table scan expression?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final BindingMatcher<Collection<PlanPartition>> innerPlanPartitionsMatcher = rollUpPartitions(any(innerPlanPartitionMatcher));
+
+    @Nonnull
+    private static final BindingMatcher<Reference> innerReferenceMatcher = planPartitions(innerPlanPartitionsMatcher);
+
+    @Nonnull
+    private static final BindingMatcher<Quantifier.ForEach> innerQunMatcher = forEachQuantifierOverRef(innerReferenceMatcher);
+
+    @Nonnull
+    private static final BindingMatcher<Quantifier.ForEach> mapTempTableQunMatcher = forEachQuantifier(mapPlan(tempTableScanPlan()));
```
normen662: As stated below, these matchers should look at the logical side, not the physical implementation. However, it is true that you would still (under the current standings) would have to match both temp table scan expressions as well as simple SELECTs (without predicates) over temp table scan expressions. I filed https://github.com/FoundationDB/fdb-record-layer/issues/3649 to simplify that (and a myriad of other cases). Once that is done, the PRE here should be simplified to reflect that. Please refer to that issue in the code here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Nonnull
+    private static final BindingMatcher<Quantifier.ForEach> tempTableQunMatcher = forEachQuantifier(tempTableScanPlan());
+
+    @Nonnull
+    private static final BindingMatcher<SelectExpression> recursiveInnerSelectMatcher = selectExpression(
+            exactlyInAnyOrder(innerQunMatcher, tempTableQunMatcher.or(mapTempTableQunMatcher)));
+
+    @Nonnull
+    private static final BindingMatcher<TempTableInsertExpression> recursivePlanMatcher = tempTableInsertExpression(
```
normen662: `recursivePlanMatcher` is probably not a good name

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+public class ImplementRecursiveDfsUnionRule extends ImplementationCascadesRule<RecursiveUnionExpression> {
+
+    @Nonnull
+    private static final BindingMatcher<RecordQueryPlan> initialInnerPlanMatcher = anyPlan();
+
+    @Nonnull
+    private static final BindingMatcher<TempTableInsertPlan> initialPlanMatcher = tempTableInsertPlan(initialInnerPlanMatcher);
+
+    @Nonnull
+    private static final BindingMatcher<PlanPartition> innerPlanPartitionMatcher = anyPlanPartition();
```
normen662: all of these namings should contain `recursive` like `recursiveInnerPlanPartitionMatcher`, etc. in order to differentiate for the initial side.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final BindingMatcher<RecursiveUnionExpression> root = recursiveUnionExpression(forEachQuantifier(initialPlanMatcher),
+            forEachQuantifier(recursivePlanMatcher)).where(preOrderTraversalIsAllowed());
+
+    public ImplementRecursiveDfsUnionRule() {
+        super(root);
+    }
+
+    @Override
+    public void onMatch(@Nonnull final ImplementationCascadesRuleCall call) {
+        final var recursiveUnion = call.get(root);
```
normen662: The naming should be indicative what entity it is they are holding; this should be `recursiveUnionExpression`. `...Alias` is fine. In a ton of other code we say `...Quantifier` and `...Reference` as opposed to use abbreviations.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+        final var innerRef = call.get(innerReferenceMatcher);
+        final var innerQun = call.get(innerQunMatcher);
+        final var priorValueCorrelation = (call.getBindings().containsKey(tempTableQunMatcher)
+                                          ? call.get(tempTableQunMatcher)
+                                          : call.get(mapTempTableQunMatcher)).getAlias();
+
+        final var rootPlanRef = call.memoizePlan(initialInnerPlan);
+        final var rootQun = Quantifier.physical(rootPlanRef, rootAlias);
+
+        final var recursivePlanRef = ImplementSimpleSelectRule.implementSelectExpression(call, recursiveInnerSelect.getResultValue(),
```
normen662: Don't you have to apply whatever the map plan did (if there was one)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
```diff
@@ -0,0 +1,299 @@
...
+
+    @Nonnull
+    @Override
+    public String toString() {
+        return ExplainPlanVisitor.toStringForDebugging(this);
+    }
+
+    @Override
+    @SuppressWarnings("PMD.CompareObjectsWithEquals")
+    public boolean equalsWithoutChildren(@Nonnull RelationalExpression otherExpression,
```
normen662, line 172: Don't you have to worry about that prior alias here? In fact, you either have to enforce that they are the same (not as good), or override `structuralEquals` and add a mapping between the (non-equal) prion aliases before descending into the children for the equals.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursivePlan.java
```diff
@@ -0,0 +1,299 @@
...
+                        NodeInfo.NESTED_LOOP_JOIN_OPERATOR,
+                        ImmutableList.of("RECURSIVE {{expr}}"),
+                        ImmutableMap.of("expr", Attribute.gml(getResultValue().toString()))),
+                childGraphs,
+                getQuantifiers());
+    }
+
+    @Nonnull
+    @Override
+    public PRecordQueryRecursivePlan toProto(@Nonnull final PlanSerializationContext serializationContext) {
```
normen662: Please extend the naming changes to the protobuf classes as well.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
hatyo: The enum is nested inside the `RecursiveUnionExpression`, so there shouldn't be so much confusion I think.

Having said that, I will still rename the enum to `TraversalStrategy` to have some sort of middle ground here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
hatyo: We do not have (yet) an implementation of the derivations for recursive union, there is an open GitHub issue about this which I referenced also here, considering that, as you said, solving the issue will end up with an implementation for a recursive union plan that can be, for the most part, reused for recursive plan as well.

The issue is https://github.com/FoundationDB/fdb-record-layer/issues/2974.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
hatyo: Yes, we can do that, but if do some forward-thinking where we consider a post-order traversal strategy:
- That strategy will end up (probably) being something purely internal to the `RecursiveCursor`.
- I do not think it will affect anything in the `RecordQueryRecursiveUnionPlan` not the implementation rule for it, besides passing a traversal strategy option down to the `RecursiveCursor`.

This is why I renamed the implementation rule `ImplementRecursiveDfsPlan` thinking that it can cover both pre- and postorder traversal.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/StoredRecordProperty.java
hatyo: Fixed, please review.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PrimaryKeyProperty.java
hatyo: Isn't that what `commonPrimaryKeyFromChildren` does? 

or do you mean we should recursively call the visitor for each leg then call `commonPrimaryKeyValuesMaybeFromOptionals`? 

(`getChildren` of `RecordQueryRecursivePlan` returns both the initial and the recursive legs)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursivePlan.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: renamed to `recursiveSelectExpressionMatcher`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
```diff
@@ -0,0 +1,124 @@
...
+        final var initialInnerPlan = call.get(initialInnerPlanMatcher);
+        final var recursiveInnerSelect = call.get(recursiveInnerSelectMatcher);
+        final var innerPlanPartition = call.get(innerPlanPartitionMatcher);
+        final var innerRef = call.get(innerReferenceMatcher);
+        final var innerQun = call.get(innerQunMatcher);
+        final var priorValueCorrelation = (call.getBindings().containsKey(tempTableQunMatcher)
+                                          ? call.get(tempTableQunMatcher)
+                                          : call.get(mapTempTableQunMatcher)).getAlias();
+
+        final var rootPlanRef = call.memoizePlan(initialInnerPlan);
```
normen662: This needs to be `memoizeMemberPlansFromOther` to avoid reexploration of that plan.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
hatyo: I updated derivation calculation for `RecordQueryRecursiveLevelUnionPlan` and `visitRecursiveDfsJoinPlan` after an offline discussion such that they return a combined derivation of their children. As it turned out this is not entirely correct, but definitely better than returning nothing. Therefore, the bug #2974 is still outstanding and this is by no means a proper fix.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/DerivationsProperty.java
hatyo: ~.... which is raising an exception during `Value` simplification, because the simplification engine gets confused by the recursive quantifier that is created out of thin air by the operator, and fails to adjust some `FV` on top of it.~

~As a quick and dirty workaround, I am making the `Derivation` only consider the root leg (for both plans), which is ok considering that the root leg does not have any `Value` referencing the recursive correlation identifier.~

With combined derivations from children, I had a test failure in a test that had an arithmetic expression in the projection list of the recursive leg's `select`. It turned out this is related to a bug in nullability-preserving semantics in `ArithmeticValue` #3655. I am reverting this to what it was before (returning `EMPTY` derivations), and will provide a separate fix for #2974.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PrimaryKeyProperty.java
normen662: You are right.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
hatyo: Reimplemented `structuralEquals`. I also added some checks to verify it by Ser/De the hierarchical plan in `RecursiveQueriesTest`.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsUnionRule.java
hatyo: Considering that select-merge fix #3649 will eliminate that `SELECT` expression, I am adding an interim check here that the `SELECT` expression is trivial, that is, it is a `SELECT *` without predicates.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/ImplementRecursiveDfsJoinRule.java
```diff
@@ -145,7 +145,7 @@ public void onMatch(@Nonnull final ImplementationCascadesRuleCall call) {
                                           ? call.get(recursiveTempTableScanQunMatcher)
                                           : call.get(recursiveSelectFromTempTableScanQunMatcher)).getAlias();
 
-        final var rootPlanRef = call.memoizeMemberPlansFromOther(initialInnerPlanRef, ImmutableList.of(initialInnerPlan));
+        final var rootPlanRef = call.memoizeMemberPlansFromOther(initialInnerPlanRef, List.of(initialInnerPlan));
```
normen662: can you please revert this? All of the planner needs currently uses `ImmutableList`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
```diff
@@ -81,7 +80,7 @@ public RecordQueryRecursiveDfsJoinPlan(@Nonnull final Quantifier.Physical rootQu
         this.rootQuantifier = rootQuantifier;
         this.childQuantifier = childQuantifier;
         this.priorValueCorrelation = priorValueCorrelation;
-        this.resultValue = RecordQuerySetPlan.mergeValues(ImmutableList.of(rootQuantifier, childQuantifier));
+        this.resultValue = RecordQuerySetPlan.mergeValues(List.of(rootQuantifier, childQuantifier));
```
normen662: here too

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
```diff
@@ -130,7 +129,7 @@ public boolean canCorrelate() {
     @Nonnull
     @Override
     public List<RecordQueryPlan> getChildren() {
-        return ImmutableList.of(rootQuantifier.getRangesOverPlan(), childQuantifier.getRangesOverPlan());
+        return List.of(rootQuantifier.getRangesOverPlan(), childQuantifier.getRangesOverPlan());
```
normen662: here as well

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
```diff
@@ -281,7 +280,7 @@ public int planHash(@Nonnull final PlanHashMode mode) {
     @Nonnull
     @Override
     public List<? extends Quantifier> getQuantifiers() {
-        return ImmutableList.of(rootQuantifier, childQuantifier);
+        return List.of(rootQuantifier, childQuantifier);
```
normen662: here as well

normen662 (approved):

PR #3618, opened 2025-09-19
Add a connection option to control compression

MMcM:
This is not generally useful, but allows testing serialization with compression off.

Also properly distinguish between encrypting when writing and being able to decrypt when reading. This is a little messy because it means allowing YAML options to unset an Option, for which we need a special standin.


yaml-tests/src/test/resources/serialization-options.yamsql
```diff
@@ -53,10 +50,11 @@ test_block:
       - unorderedResult: [{S: 'def'}]
 ---
 test_block:
-  name: read-unencrypted
+  name: read-without-encryption
   options:
     connection_options:
       ENCRYPT_WHEN_SERIALIZING: false
+      ENCRYPTION_KEY_STORE:
```
ScottDugas, line 57: Is the introduction of `TypeContract.nullableStringValue` solely to support setting this at the connection level, and then overriding here? Should you need to clear this out at all, given that you're setting `ENCRYPT_WHEN_SERIALIZING`?
I didn't see any bugs, but am wondering whether the additonal complexity in the options is worthwhile.

ScottDugas: Also, perhaps this file should be renamed, maybe `serialization-options`, since it is covering both encrypted and unencrypted.

MMcM: This is to separately allow decryption with a known key from encrypting new records, which weren't properly distinguished before. So turning off the latter and losing the former are separate. Unsetting wouldn't be needed if there were a way to get a whole new set of connection options, rather than overriding the preamble. This could also be done by rearranging the overall test YAML file.

I also think that prior to this change having a blank option would get an NPE reporting the type mismatch.

MMcM: I renamed the test, per your suggestion.

yaml-tests/src/test/resources/serialization-options.yamsql
```diff
@@ -21,7 +21,7 @@ options:
     supported_version: 4.6.1.0
     connection_options:
       ENCRYPT_WHEN_SERIALIZING: true
-      ENCRYPTION_KEY_STORE: src/test/resources/encrypted.p12
+      ENCRYPTION_KEY_STORE: src/test/resources/serialization-keys.p12
```
ScottDugas, line 24: copyright header at the top needs to be updated.

MMcM: Right. Fixed.

ScottDugas (approved):

PR #3624, opened 2025-09-24
fix bug where pullup should create an ordinary map instead or an iden…

normen662:
…tity-based map

A caller of the `pullUp(...)` method in `Ordering` used the code is a way that the resulting map was probed by equality and not by identity leading to undefined behavior. In general, methods like this one are normally implemented to return sets and maps using equality rather than identity unless there is a semantic reason for it.


hatyo (approved):
This produces the correct mapped ordering.

PR #3625, opened 2025-09-24
improve memo-performance

normen662:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -980,6 +980,11 @@ public boolean addAll(@Nonnull final Collection<? extends RelationalExpression>
 
         public boolean containsInMemo(@Nonnull final RelationalExpression expression,
                                       @Nonnull final AliasMap equivalenceMap) {
+            // short cut
+            if (expressions.contains(expression)) {
```
alecgrieser: Does this need to validate that the `equivalenceMap` includes only identities? Otherwise, would this open us up to a possibility where two identical expressions exist, but their correlations mean different things?

Perhaps defining only identities is a little strict. Perhaps more accurate would be something like "none of the aliases used in `expression` are in the equivalence map"

alecgrieser (changes requested):
Seems like a good idea, if we can work out the issue with the `AliasMap` there. I suppose it would be nice to know if we had data on how this improved things, though it's a relatively cheap thing to add

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
hatyo: Checking for duplications based on alias map equivalence is done later on in `isMemoizedExpression` I think, this is a quick reference check to early out if we're attempting to add the same `expression` object.

hatyo (approved):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
normen662: @alecgrieser 's comment addresses the case where we early out when we maybe shouldn't. So if we short circuit using that logic we will not get to `isMemoizedExpression(...)`. That comment is valid except I don't think it can happen in real world scenarios. Let me add the only-identities check as that is essentially a no-op for all implementations.

hatyo: @normen662 I don't think the check will apply, or at least I find it difficult to find a scenario where the check evaluates to false? Perhaps we can add an else branch that asserts to the check and see if any test fail.

alecgrieser: I agree that it would be a bit contrived. We'd somehow need to reuse the same object (to make the identity map pick the expression) in a context where there's an alias map being used. Something like a `expr(QOV(q1))` with an alias map like `[q1 -> q2, q3 -> q1]`, and we find a reference that already has the same `expr(QOV(q1))` in place (even though the _new_ expression would be `expr(QOV(q2))` subject to the alias map).

I don't think we should just have an assert here that asserts to `false`, though, as that mainly seems like its laying a trap that for us to run into later (maybe in some future case where we've specifically done some kind of tricky correlation id swapping for some reason). It would be nice, though, to have a test for which the check doesn't pass

PR #3627, opened 2025-09-24
Explicitly fetch PR head in metrics diff job

alecgrieser:
The metrics diff job failed when trying to compare two hashes: https://github.com/FoundationDB/fdb-record-layer/actions/runs/17979075422

This is because we are using the `pull_request_target` job type, and that only fetches the target's branch's state (so, `main` in this case), but we need to fetch both the PR head and the target head in order to compute the diff. We have to be a little careful, because the `pull_request_target` runs at heightened permissions (that is, including writes), so it is important it does not run any outside code. So in this fix, we fetch the PR head, but we do not check it out. We will continue to run the analysis tool from the target branch.

Hopefully, this unblocks the broken metrics diff job.

### DIFF
--- a/.github/workflows/metrics_analysis.yml
+++ b/.github/workflows/metrics_analysis.yml
@@ -9,6 +9,7 @@ jobs:
   metrics-analysis:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       issues: write
       pull-requests: write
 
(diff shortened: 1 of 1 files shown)


.github/workflows/metrics_analysis.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   metrics-analysis:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
```
alecgrieser, line 12: We should have added this to say that this job is allowed to read contents from the git repo. The fact that we didn't wasn't causing us trouble (perhaps because the repo is public), but this is good practice.

PR #3638, opened 2025-09-26
Make YAML tests pass connection options to servers for transaction setups

hazefully:
This PR makes transactional setups within a `test_block` in YAML tests make use of the specified `connection_options` for a block when making use of a gRPC server for testing, which is now possible thanks to the work introduced in #3557 to pass the `connection_options` as part of the request for the remote JDBC connection.

In addition, this PR also makes it possible to specify `connection_options` for a `setup` block, making it possible to use different connections options for different setup blocks within the same YAML test.


yaml-tests/src/test/resources/transactions-tests.yamsql
```diff
@@ -58,4 +58,43 @@ test_block:
...
       - result: [{id: 30, col1: 40}]
+---
+# Create a new schema with the CASE_SENSITIVE_IDENTIFIERS connection option turned on
+setup:
+  options:
+    connection_options:
+      CASE_SENSITIVE_IDENTIFIERS: true
+  connect: "jdbc:embed:/__SYS?schema=CATALOG"
+  steps:
+    - query: drop schema template if exists case_sensitive_template
```
alecgrieser: I suppose this works, but it's a bit weird to put this in this file. I think ideally, we'd create a separate file for these tests and then a separate test to invoke the file. Unless there's some dependency between this and the existing tests that I'm missing

hazefully: There is no dependency, I just thought I would add it to the existing `transactions-tests` file given that it is testing that transactional setups make use of connection options, but it also tests that regular setup blocks make use of them as well so having it in a separate test makes sense to me 👍

alecgrieser (approved):

PR #3640, opened 2025-09-26
Allow index maintainer implementors to specify their own index match candidate

alecgrieser:
This refactors the way that index match candidates are generated so that individual index maintainer factory implementations can override how this is done for the index types they manage. This includes plumbing through the store's `IndexMaintainerRegistry` to the `CascadesPlanner` so that it can ask the right logic for the match candidate, as well as some refactoring done to the `MatchCandidate` logic so that, if all goes well, index implementors can rely on library code as much as possible.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
@@ -21,10 +21,13 @@
 package com.apple.foundationdb.record.provider.foundationdb;
 
 import com.apple.foundationdb.annotation.API;
+import com.apple.foundationdb.record.RecordMetaData;
 import com.apple.foundationdb.record.metadata.Index;
 import com.apple.foundationdb.record.metadata.IndexValidator;
+import com.apple.foundationdb.record.query.plan.cascades.MatchCandidate;
 
 import javax.annotation.Nonnull;
+import java.util.Collections;
 
 /**
  * A factory for {@link IndexMaintainer}.
@@ -68,4 +71,45 @@ public interface IndexMaintainerFactory {
      */
     @Nonnull
     IndexMaintainer getIndexMaintainer(@Nonnull IndexMaintainerState state);
+
+    /**
+     * Create {@link MatchCandidate}s to use for the given {@link Index}.
+     * Implementors can assume that the index is one of this factory's
+     * {@linkplain #getIndexTypes() supported types}. Note that this is
+     * structured as a method on the maintainer factory so that indexes
+     * defined outside the core repository can define the structure
+     * used by the planner to match them. However, use cases should be
+     * aware that the planner
+     *
+     * <p>
+     * By default, this returns an empty collection. This means that the
+     * index will not be matchable by the {@link com.apple.foundationdb.record.query.plan.cascades.CascadesPlanner}.
+     * Index types that want to be used in plans should implement this
+     * method. Utility methods in {@link MatchCandidate} can be used to
+     * create, for example, canonical expansions of indexes that behave like
+     * indexes defined in the core repository (e.g., {@link com.apple.foundationdb.record.metadata.IndexTypes#VALUE}
+     * indexes).
+     * </p>
+     *
+     * <p>
+     * Indexes may return more than one {@link MatchCandidate}. This can be
+     * useful for index types that have more than one way of being scanned.
+     * For example, the {@link com.apple.foundationdb.record.metadata.IndexTypes#RANK}
+     * index can be scanned {@link com.apple.foundationdb.record.IndexScanType#BY_VALUE},
+     * at which point it behaves just like a {@link com.apple.foundationdb.record.metadata.IndexTypes#VALUE}
+     * index, or it can be scanned {@link com.apple.foundationdb.record.IndexScanType#BY_RANK},
+     * at which point it behaves very differently. The different scan types are reflected
+     * in different {@link MatchCandidate} which encode different information about
+     * the index and how it should be scanned.
+     * </p>
+     *
+     * @param metaData the meta-data in which the index is defined
+     * @param index the index to expend
+     * @param reverse whether the index is to be scanned in reverse
+     * @return a collection of {@link MatchCandidate}s to match the index against
+     */
+    @Nonnull
+    default Iterable<MatchCandidate> createMatchCandidates(@Nonnull RecordMetaData metaData, @Nonnull Index index, boolean reverse) {
+        return Collections.emptyList();
+    }
 }
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
@@ -24,7 +24,6 @@ import com.apple.foundationdb.annotation.API;
 import com.apple.foundationdb.record.logging.KeyValueLogMessage;
 import com.apple.foundationdb.record.logging.LogMessageKeys;
 import com.apple.foundationdb.record.metadata.Index;
-import com.apple.foundationdb.record.metadata.IndexValidator;
 import com.apple.foundationdb.record.metadata.MetaDataException;
 import com.apple.foundationdb.record.util.ServiceLoaderProvider;
 import org.slf4j.Logger;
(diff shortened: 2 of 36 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
```diff
@@ -68,4 +71,45 @@ public interface IndexMaintainerFactory {
...
     IndexMaintainer getIndexMaintainer(@Nonnull IndexMaintainerState state);
+
+    /**
+     * Create {@link MatchCandidate}s to use for the given {@link Index}.
+     * Implementors can assume that the index is one of this factory's
+     * {@linkplain #getIndexTypes() supported types}. Note that this is
+     * structured as a method on the maintainer factory so that indexes
+     * defined outside the core repository can define the structure
+     * used by the planner to match them. However, use cases should be
+     * aware that the planner
```
normen662: The sentence abruptly ends.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
```diff
@@ -68,4 +71,45 @@ public interface IndexMaintainerFactory {
...
+     * <p>
+     * By default, this returns an empty collection. This means that the
+     * index will not be matchable by the {@link com.apple.foundationdb.record.query.plan.cascades.CascadesPlanner}.
+     * Index types that want to be used in plans should implement this
+     * method. Utility methods in {@link MatchCandidate} can be used to
+     * create, for example, canonical expansions of indexes that behave like
+     * indexes defined in the core repository (e.g., {@link com.apple.foundationdb.record.metadata.IndexTypes#VALUE}
+     * indexes).
+     * </p>
+     *
```
normen662, line 97:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
```diff
@@ -68,4 +71,45 @@ public interface IndexMaintainerFactory {
...
+     * useful for index types that have more than one way of being scanned.
+     * For example, the {@link com.apple.foundationdb.record.metadata.IndexTypes#RANK}
+     * index can be scanned {@link com.apple.foundationdb.record.IndexScanType#BY_VALUE},
+     * at which point it behaves just like a {@link com.apple.foundationdb.record.metadata.IndexTypes#VALUE}
+     * index, or it can be scanned {@link com.apple.foundationdb.record.IndexScanType#BY_RANK},
+     * at which point it behaves very differently. The different scan types are reflected
+     * in different {@link MatchCandidate} which encode different information about
+     * the index and how it should be scanned.
+     * </p>
+     *
```
normen662, line 109:
```suggestion
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
```diff
@@ -38,7 +37,7 @@
  * A singleton {@link IndexMaintainerRegistry} that finds {@link IndexMaintainerFactory} classes in the classpath.
  */
 @API(API.Status.INTERNAL)
-public class IndexMaintainerRegistryImpl implements IndexMaintainerRegistry {
+public class IndexMaintainerRegistryImpl implements IndexMaintainerFactoryRegistry {
```
normen662: `IndexMaintainerFactoryRegistryImpl`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
```diff
@@ -68,4 +71,45 @@ public interface IndexMaintainerFactory {
...
+     * @param metaData the meta-data in which the index is defined
+     * @param index the index to expend
+     * @param reverse whether the index is to be scanned in reverse
+     * @return a collection of {@link MatchCandidate}s to match the index against
+     */
+    @Nonnull
+    default Iterable<MatchCandidate> createMatchCandidates(@Nonnull RecordMetaData metaData, @Nonnull Index index, boolean reverse) {
```
normen662, line 116: I cannot remember the specifics but the `reverse` here seems odd.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
```diff
@@ -239,6 +244,10 @@ public CascadesPlanner(@Nonnull RecordMetaData metaData, @Nonnull RecordStoreSta
         this.taskStack = new ArrayDeque<>();
     }
 
+    public static CascadesPlanner forStore(@Nonnull FDBRecordStoreBase<?> recordStore) {
```
normen662: Could this method be an instance method on the store? Also, it should annotate `@Nonnull`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidateExpansion.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    @Nonnull
+    public static <T> Iterable<T> optionalToIterable(@Nonnull Optional<T> optional) {
+        return optional
+                .map(ImmutableList::of)
+                .orElse(ImmutableList.of());
+    }
+
+    @Nonnull
+    public static Iterable<MatchCandidate> expandValueIndexMatchCandidate(@Nonnull RecordMetaData metaData, @Nonnull Index index, boolean isReverse) {
```
normen662, line 75: Shouldn't we put all these statics into their corresponding factories? Maybe we should discuss what to do.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
normen662: I looked it up and it's only the planning of `RecordQuery`s that need this. Annoying...

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
alecgrieser: Added more content

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
alecgrieser: These blank lines are intentional. They are to break up the different paragraphs of text for the reader who is looking at the code (as opposed to the `<p>` tags which are for the reader of the generated javadoc). Arguably, they'd also be broken up by two lines that just had `</p>` and `<p>`, but I like the separation of a completely blank line, which I think is what you'd do in an HTML document, if anyone still does that

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
alecgrieser: Ditto

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerFactory.java
alecgrieser: Yeah, I did notice that, but I didn't look too deeply into it. I'm not actually sure why we can't rely on the sort order on the sort descriptor. If we could, I'd be happy to remove this parameter

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/CascadesPlanner.java
alecgrieser: Sure, made it a method of `FDBRecordStoreBase` (and marked its return type as `@Nonnull`)

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/MatchCandidateExpansion.java
alecgrieser: We could discuss this further if we wanted. My thought here is that the implementations of the index maintainer factories may want to reference these methods (so that, for example, a custom value-like index could use `expandValueIndexMatchCandidate` or a custom aggregate index could use the `expandAggregateIndexMatchCandidate`). While they could do the same if it lived on the maintainer factory, my thought was that there these are closer to being public utility methods rather than methods particular to the different types.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/IndexMaintainerRegistryImpl.java
alecgrieser: Done

PR #3644, opened 2025-09-29
Memoize correlation set computation in all planner expressions

normen662:
- introduced `AbstractRelationalExpressionWithoutChildren` and `AbstractRelationalExpressionWithChildren` to provide a place where the suppliers for hash code computation and correlation computation can live
- every class descendant of `RelationalExpression` has to extend from one of these two classes. This is analogous to what is done for `Value`, `QueryPredicate` already
- every class that previously memoized the correlation sets before by themselves now uses this new mechanism
- implementing classes only ever implement `computeHashCodeWithoutChildren()` and `computeCorrelatedXXX()`

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryRecursiveDfsJoinPlan.java
@@ -43,14 +43,16 @@ import com.apple.foundationdb.record.query.plan.cascades.explain.Attribute;
 import com.apple.foundationdb.record.query.plan.cascades.explain.ExplainPlanVisitor;
 import com.apple.foundationdb.record.query.plan.cascades.explain.NodeInfo;
 import com.apple.foundationdb.record.query.plan.cascades.explain.PlannerGraph;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.AbstractRelationalExpressionWithChildren;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpression;
-import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpressionWithChildren;
 import com.apple.foundationdb.record.query.plan.cascades.values.Value;
 import com.apple.foundationdb.record.query.plan.cascades.values.translation.TranslationMap;
 import com.google.auto.service.AutoService;
 import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Streams;
 import com.google.protobuf.Message;
 
 import javax.annotation.Nonnull;
@@ -63,7 +65,7 @@ import java.util.Set;
  * A query plan that recursively applies a plan to earlier results, starting with a root plan.
  */
 @API(API.Status.INTERNAL)
-public class RecordQueryRecursiveDfsJoinPlan implements RecordQueryPlanWithChildren, RelationalExpressionWithChildren {
+public class RecordQueryRecursiveDfsJoinPlan extends AbstractRelationalExpressionWithChildren implements RecordQueryPlanWithChildren {
     private static final ObjectPlanHash BASE_HASH = new ObjectPlanHash("Record-Query-Recursive-Plan");
 
     @Nonnull
@@ -141,7 +143,19 @@ public class RecordQueryRecursiveDfsJoinPlan implements RecordQueryPlanWithChild
 
     @Nonnull
     @Override
-    public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
+    public Set<CorrelationIdentifier> computeCorrelatedTo() {
+        final ImmutableSet.Builder<CorrelationIdentifier> builder = ImmutableSet.builder();
+        Streams.concat(rootQuantifier.getCorrelatedTo().stream(),
+                        childQuantifier.getCorrelatedTo().stream())
+                // filter out the correlations that are satisfied by this plan
+                .filter(alias -> !alias.equals(priorValueCorrelation))
+                .forEach(builder::add);
+        return builder.build();
+    }
+
+    @Nonnull
+    @Override
+    public Set<CorrelationIdentifier> computeCorrelatedToWithoutChildren() {
         return resultValue.getCorrelatedTo();
     }
 
@@ -241,7 +255,7 @@ public class RecordQueryRecursiveDfsJoinPlan implements RecordQueryPlanWithChild
     }
 
     @Override
-    public int hashCodeWithoutChildren() {
+    public int computeHashCodeWithoutChildren() {
         return Objects.hash(getResultValue()); // priorValueCorrelation _is_ a child correlation due to the recursive nature of this plan.
     }
 
--- a/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/RecursiveQueriesTest.java
+++ b/fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/provider/foundationdb/query/RecursiveQueriesTest.java
@@ -38,7 +38,6 @@ import com.apple.foundationdb.record.query.plan.cascades.GraphExpansion;
 import com.apple.foundationdb.record.query.plan.cascades.Quantifier;
 import com.apple.foundationdb.record.query.plan.cascades.Reference;
 import com.apple.foundationdb.record.query.plan.cascades.UnableToPlanException;
-import com.apple.foundationdb.record.query.plan.cascades.explain.ExplainPlanVisitor;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.FullUnorderedScanExpression;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.LogicalSortExpression;
 import com.apple.foundationdb.record.query.plan.cascades.expressions.LogicalTypeFilterExpression;


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/AbstractRelationalExpression.java
```diff
@@ -0,0 +1,63 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * Abstract implementation of {@link Value} that provides memoization of correlatedTo sets.
+ */
+@API(API.Status.EXPERIMENTAL)
+public abstract class AbstractRelationalExpression implements RelationalExpression {
```
alecgrieser, line 35: I'm not sure how much I agree with this philosophy, but there are certainly people who would advocate for a composition-based approach instead of an inheritance based approach like the one that you've adopted here. How would you evaluate a composition based approach to something like this?

If you wanted to make this a bit more compositional, I think we would take an approach like creating a `ExpressionStateMemoizer` that would be encapsulate the various memoizers here (correlated to and hash code). We could even leave the change where the new `computeX` methods were added, and then the `RelationalExpression`s would each have one state memoizer that would delegate to their `compute` methods. And then you'd need to wire up the `getCorrelatedTo()` method to use the state memoizer.

I'm not sure I actually like this idea better, but it's the kind of thing that would come up if brought to a larger audience.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpressionWithChildren.java
```diff
@@ -43,9 +43,8 @@ public interface RelationalExpressionWithChildren extends RelationalExpression {
     int getRelationalChildCount();
 
     @Nonnull
-    @Override
     @SuppressWarnings("squid:S2201")
-    default Set<CorrelationIdentifier> getCorrelatedTo() {
+    default Set<CorrelationIdentifier> computeCorrelatedTo() {
```
alecgrieser: This is minor, but it may be a little clearer what's going on if this were defined in `AbstractRelationalExpressionWithChildren`. As it was, it was a little too much back and forth between the new abstract class and this marker interface to validate that we weren't missing logic to actually combine the children's correlations with the results of `getCorrelatedToWithoutChildren`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
```diff
@@ -77,7 +78,19 @@ public int getRelationalChildCount() {
 
     @Nonnull
     @Override
-    public Set<CorrelationIdentifier> getCorrelatedToWithoutChildren() {
+    public Set<CorrelationIdentifier> computeCorrelatedTo() {
```
alecgrieser, line 116: I take it that this was a bug that was in the `RecursiveUnionExpression` that meant is was miscalculating its correlated to sets? Did we add tests that surface this problem?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAggregateIndexPlan.java
```diff
@@ -120,7 +124,7 @@ public RecordQueryAggregateIndexPlan(@Nonnull final RecordQueryIndexPlan indexPl
 
     @Nonnull
     @Override
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "resource"})
```
alecgrieser, line 127: This method didn't change. Did something happen that made it necessary to suppress this warning now?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryCoveringIndexPlan.java
```diff
@@ -104,6 +105,7 @@ public RecordQueryCoveringIndexPlan(@Nonnull RecordQueryPlanWithIndex indexPlan,
 
     @Nonnull
     @Override
+    @SuppressWarnings("resource")
```
alecgrieser, line 108: Similar question for this method about why we're suppressing `resource` now

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
```diff
@@ -159,8 +160,8 @@ public String toString() {
 
     @Nonnull
     @Override
-    public Set<CorrelationIdentifier> getCorrelatedTo() {
-        return ImmutableSet.of(); // TODO this should be reconsidered when we have selects
```
alecgrieser, line 163: This `TODO` got dropped. I'm not sure I totally follow what it was trying to say, though. Do we still need it, or was it already out of date/irrelevant?

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
```diff
@@ -56,6 +57,8 @@
 /**
  * Lucene query plan for including search-related scan parameters.
  */
+@SuppressWarnings("PMD.OverrideBothEqualsAndHashcode")
+@SpotBugsSuppressWarnings("HE_EQUALS_NO_HASHCODE")
```
alecgrieser, line 61: Why did this change result in these warnings being suppressed?

alecgrieser (changes requested):
A few smaller changes, but overall, it looks good to me.

I did want to bring up #3365 though. Is there a reason why we wouldn't expect it to be a problem here like it was for quantifiers (given that `RelationalExpression`s are sometimes built on top of quantifiers)?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryAggregateIndexPlan.java
normen662: I looked through every plan and I fixed some of these warnings on the go. We will never properly close the resource in the way this check wants us to be so I added it the suppression here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RelationalExpressionWithChildren.java
normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/RecursiveUnionExpression.java
normen662: The direct result of the bug was that the plan was correlated to its internal aliases. I don't think there are currently any tests that test this. Let me add one.

normen662: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/plans/RecordQueryLoadByKeysPlan.java
normen662: It was out of date. In addition to this LoadByKeys is currently not used by the Cascades planner. If at some point it gets used we may just use a variant of an index scan (for agg indexes) that can express and use proper correlations, i.e. I am not sure we may ever want to reuse this operator at all.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/LuceneIndexQueryPlan.java
normen662: I think we had a `hashCode()` before that just called `super.hashCode()` which then caused some checker to barf. Not sure why that only happened when I touched the `computeHashCode...()` but it did. I removed that implementation and another check alerted. I ended up doing it like this.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/expressions/AbstractRelationalExpression.java
normen662: discussed offline.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/Reference.java
```diff
@@ -364,6 +364,7 @@ private void insertUnchecked(@Nonnull final RelationalExpression newExpression,
 
         Debugger.sanityCheck(() -> Verify.verify(getTotalMembersSize() == 0 ||
                 getResultType().equals(newExpression.getResultType())));
+        Debugger.sanityCheck(() -> Verify.verify(getCorrelatedTo().containsAll(newExpression.getCorrelatedTo())));
```
alecgrieser: I think this probably needs to first check to see if the reference is empty to avoid the false positives it's seeing now. I suppose there's an interesting question as to whether the correlation sets should be allowed to be different when the reference is _first_ created or not, but that seems like it would be mostly an issue during testing if just validating emptiness isn't enough

alecgrieser (approved):

PR #3647, opened 2025-09-29
Support SQL `CAST` operator

### LINKED ISSUE #3589: Implement explicit `CAST` operator
Currently it is not possible to explicitly cast literals and expressions to other types. This makes it difficult or even impossible to present  certain literals that align well with the underlying type system for example, currently, it is not possible to project an empty array literal `[]` because it is not possible to flow unknown types which is fair, but there is no way to cast it to a known type neither:

```sql
select [] from t;
```

hatyo:
This implements SQL `CAST` operator with comprehensive type conversion support, including casting of one-dimensional arrays.

This PR adds the SQL CAST operator to enable explicit type conversions between supported data types. The implementation follows SQL standard semantics as much as possible, in addition to few non-standard additions, and includes comprehensive validation and error handling.

  Key examples:
```sql
  -- Basic type conversions
  SELECT CAST(123 AS STRING)           -- Returns '123'
  SELECT CAST('42' AS INTEGER)         -- Returns 42
  SELECT CAST(true AS INTEGER)         -- Returns 1
```

  -- Array casting with element-wise conversion
```sql
  SELECT CAST([1, 2, 3] AS STRING ARRAY)  -- Returns ['1', '2', '3']
  SELECT CAST([] AS INTEGER ARRAY)        -- Returns empty integer array
```

The `CAST` operator supports conversions between primitive types (integers, strings, booleans, floating-point numbers) and extends naturally to arrays by applying the same conversion rules recursively to array elements. Proper error handling ensures invalid conversions fail with appropriate SQL error codes.

This fixes #3589.


yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -326,4 +326,10 @@ public void compositeAggregates(YamlTest.Runner runner) throws Exception {
     public void serializationOptions(YamlTest.Runner runner) throws Exception {
         runner.runYamsql("serialization-options.yamsql");
     }
+
+    @TestTemplate
+    @MaintainYamlTestConfig(YamlTestConfigFilters.CORRECT_EXPLAIN_AND_METRICS)
```
normen662: maintain needs to be removed.

yaml-tests/src/test/resources/cast-tests.yamsql
```diff
@@ -0,0 +1,169 @@
...
+      - explain: "SCAN(<,>) | TFILTER TEST_CAST | FILTER _.ID EQUALS promote(@c20 AS LONG) | MAP (CAST(array(@c20, @c7, @c9) AS ARRAY(BOOLEAN)) AS _0)"
+      - result: [{[true, false, true]}]
+    -
+      # Array casting: Empty array to different types
+      - query: select CAST([] AS INTEGER ARRAY) from test_cast where id = 1
+      - explain: "SCAN(<,>) | TFILTER TEST_CAST | FILTER _.ID EQUALS promote(@c15 AS LONG) | MAP (array() AS _0)"
+      - result: [{[]}]
+    -
+      - query: select CAST([] AS DOUBLE ARRAY) from test_cast where id = 1
+      - explain: "SCAN(<,>) | TFILTER TEST_CAST | FILTER _.ID EQUALS promote(@c15 AS LONG) | MAP (array() AS _0)"
```
normen662, line 77: Can we have a better explain than `TEST_CAST`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
```diff
@@ -0,0 +1,606 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+
+/**
+ * A value that casts an object of a type to an object of another type using SQL CAST semantics.
+ * Unlike PromoteValue which handles implicit promotions, CastValue handles explicit type conversions
+ * <i>mostly</i> according to SQL standard CAST rules.
+ */
+@API(API.Status.EXPERIMENTAL)
+public class CastValue extends AbstractValue implements ValueWithChild, Value.RangeMatchableValue {
```
normen662, line 65: are you sure that a casted field is `Value.RangeMatchableValue`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
```diff
@@ -0,0 +1,606 @@
...
+     * @param castToType the target type to cast to
+     * @return a CastValue that performs the cast, or the original value if no cast is needed
+     * @throws SemanticException if the cast is not supported
+     */
+    @Nonnull
+    public static Value inject(@Nonnull final Value inValue, @Nonnull final Type castToType) {
+        final Type inType = inValue.getResultType();
+
+        // If types are the same, no cast needed
+        if (inType.equals(castToType)) {
```
normen662, line 411: What happens for types that are promotable? What does the SQL standard say?

normen662 (changes requested):

yaml-tests/src/test/resources/cast-tests.yamsql
hatyo: `TEST_CAST` is the name of the table we're testing:

```sql
select CAST([] AS DOUBLE ARRAY) from test_cast where id = 1
                                     ^^^^^^^^^
```

The `cast` explanation is e.g.

```
(CAST(array(@c5, @c7, @c9) AS ARRAY(INT))
```

yaml-tests/src/test/java/YamlIntegrationTests.java
hatyo: done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
hatyo: I don't understand this comment, this check above makes sure that if the cast is not necessary (since the types are identical, not promotable), it returns.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
hatyo: Yes, I think so, unless I am misreading something, the result of the cast is, similar to promote, a `Value` that can used for constructing a range. In other words, I believe what applies to `PromoteValue` in terms of range matchability should also apply here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
normen662: Didn't know where to put this comment. But what happens if you call this method for types that are promotable?

yaml-tests/src/test/resources/cast-tests.yamsql
normen662: sorry, I am being blind.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/values/CastValue.java
hatyo: All permissible promotions are also supported as explicit casts. Having said that, there are some promotions that should be removed, particularly: `STRING_TO_ENUM` and `STRING_TO_UUID`, these do not make sense as promotions and are enabled now as explicit casts. There is a TODO on these to remove these promotions, but we should be careful not to break any customer logic.

PR #3650, opened 2025-09-30
remove CONTINUATION_CONTAINS_COMPILED_STATEMENT usage

g31pranjal:
This PR does a couple of things:
- removes the usage of `CONTINUATION_CONTAINS_COMPILED_STATEMENT` option in Connection and as a query option. 
- The option was, by default, set to true and is now true for all the requests to the engine.
- We still support `SELECT ... WITH CONTINUATION ...`, which could be removed separately after this work. Hence, it could happen that a query started with a non-serialized continuation, but has now on serialised continuation.
- JDBC driver still supports the option, for backward compatilibility. However, the current change always sets that to `TRUE` and rejects connections that have it set to `FALSE`. Once this is rolled out, we should probably stop failing on false and mark the option as deprecated in proto. Finally, we can stop setting the option and remove the option altogether. 
- Removes a few tests around testing this option with False.

### DIFF
--- a/fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
+++ b/fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
@@ -802,8 +802,8 @@ public class TypeConversion {
         if (protoOptions.hasValidPlanHashModes()) {
             builder.withOption(Options.Name.VALID_PLAN_HASH_MODES, protoOptions.getValidPlanHashModes());
         }
-        if (!protoOptions.hasContinuationsContainCompiledStatements() || !protoOptions.getContinuationsContainCompiledStatements()) {
-            throw new RelationalException("Option CONTINUATIONS_CONTAIN_COMPILED_STATEMENTS not supported anymore!", ErrorCode.UNSUPPORTED_OPERATION).toUncheckedWrappedException();
+        if (protoOptions.hasContinuationsContainCompiledStatements() && !protoOptions.getContinuationsContainCompiledStatements()) {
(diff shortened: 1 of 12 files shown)


fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -581,6 +581,8 @@ public static Continuation.Reason toReason(RpcContinuationReason reason) {
     @SuppressWarnings("unchecked")
     static com.apple.foundationdb.relational.jdbc.grpc.v1.Options.Builder toProtobuf(@Nonnull Options options) throws SQLException {
         final var builder = com.apple.foundationdb.relational.jdbc.grpc.v1.Options.newBuilder();
+        // Switched-on by default on JDBC driver until the option is deprecated and removed.
+        builder.setContinuationsContainCompiledStatements(true);
```
alecgrieser, line 585: Is the idea here that we need to set this in case the `Options` from a modern version are shipped to a older server that expects the field to be set?

alecgrieser (approved):
LGTM. I have a PR to bump the version to 4.7: #3653. It might be good to bundle that in with the other 4.7 changes as a kind of API change that we're delivering here

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
g31pranjal: The idea, from this version, is that the option cannot be explicitly set to False. Setting it to TRUE here is indicative of what we are getting into starting now. Also, the check here is to ensure that we never set it to False going ahead, or to maintain backward compatibility. After this is rolled out, we need a PR to remove usage of this option in JDBC and mark it deprecated. Then, we remove the field itself!

PR #3651, opened 2025-09-30
mark sql_type_code are deprecated and remove usage

g31pranjal:


fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -829,4 +827,32 @@ public static Options fromProtobuf(com.apple.foundationdb.relational.jdbc.grpc.v
...
+            case BYTES:
+                return Types.BINARY;
+            case FLOAT:
+                return Types.FLOAT;
+            case DOUBLE:
+                return Types.DOUBLE;
+            default:
+                throw new SQLException("JDBC Type: " + type + " not supported");
+        }
+    }
```
alecgrieser, line 851: The test gap analysis is claiming this method is uncovered. Unless this is a mistake of the code coverage reporting tool, that would seem to suggest we're not really looking at this part of the code base that changed. Do you know what the issue is?

fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/ParameterHelper.java
```diff
@@ -40,55 +39,48 @@ public class ParameterHelper {
     public static Parameter ofBoolean(boolean b) {
         return Parameter.newBuilder()
                 .setType(Type.BOOLEAN)
-                .setJavaSqlTypesCode(Types.BOOLEAN)
```
alecgrieser, line 43: It would seem to me that for cross-version compatibility reasons, we'd need to keep setting this field at least for one minor version. Otherwise, I'm not sure how old code would make sense of the `Parameter`'s type. Is my understanding wrong, or are we not covering this in mixed mode tests? Or something else?

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalStructFacade.java
```diff
@@ -430,19 +430,19 @@ int getZeroBasedOffsetOrThrow(String fieldName) throws SQLException {
         }
 
         @Override
-        public RelationalStructBuilder addBoolean(String fieldName, boolean b) throws SQLException {
+        public RelationalStructBuilder addBoolean(String fieldName, boolean b) {
             // Add the metadata and get offset at where to insert data.
             int offset = addMetadata(ColumnMetadata.newBuilder()
-                    .setName(fieldName).setJavaSqlTypesCode(Types.BOOLEAN).setType(Type.BOOLEAN).build());
+                    .setName(fieldName).setType(Type.BOOLEAN).build());
```
alecgrieser: Does this need to retain the type codes for an additional minor version release? That is, do we need to keep setting the deprecated field until we're dropped cross version support for anything that would expect it to be there

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
g31pranjal: I believe at least this method is covered by a test over an external JDBC server and not the in-process one.

fdb-relational-jdbc/src/main/java/com/apple/foundationdb/relational/jdbc/ParameterHelper.java
g31pranjal: Yeah, I think setting old fields need to be in-place for another release. The scope of this PR should only be limited to start using new fields in place of old fields.

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalStructFacade.java
g31pranjal: replied above!

fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/RelationalArrayFacade.java
```diff
@@ -78,10 +77,8 @@ class RelationalArrayFacade implements RelationalArray {
         this.delegate = array;
     }
 
-    @Nullable
     private DataType.ArrayType computeType() {
-        return delegateMetadata.getType() == Type.UNKNOWN ? null :
-               DataType.ArrayType.from(RelationalStructFacade.RelationalStructFacadeMetaData.getDataType(delegateMetadata.getType(), delegateMetadata, delegateMetadata.getNullable()));
+        return DataType.ArrayType.from(TypeConversion.getDataType(delegateMetadata.getType(), delegateMetadata, delegateMetadata.getNullable()));
```
alecgrieser, line 81: It looks like the purpose of this PR has changed. Is this now adding support for arrays of `UNKNOWN`s in JDBC?

g31pranjal: No, this preemptive check for UNKNOWN (default enum type) was included to make this method backwards compatible for when the new type was not getting populated. Removing the check from here would still cause the `TypeConversion.getDataType` to raise an exception. The invariant should now pass if we are carefully populating the JDBC protobuf types.

alecgrieser: Okay thanks. Something weird was going on for me, and on GitHub the _only_ change I was seeing was around `UNKNOWN` handling, but I'm seeing a more complete PR now

PR #3657, opened 2025-10-02
Fix planner sanity check

### LINKED ISSUE #3656: UUID tests break `Debugger` sanity checks
When running UUID tests with a planner debugger with sanity checks activated, it crashes with the following trace:

```
Caused by: com.google.common.base.VerifyException
	at com.google.common.base.Verify.verify(Verify.java:102)
	at com.apple.foundationdb.record.query.plan.cascades.RequestedOrdering.lambda$ofPrimitiveParts$3(RequestedOrdering.java:305)
	at com.apple.foundationdb.record.query.plan.cascades.debug.Debugger.lambda$sanityCheck$1(Debugger.java:150)
	at com.apple.foundationdb.record.query.plan.cascades.debug.Debugger.withDebugger(Debugger.java:121)
	at com.apple.foundationdb.record.query.plan.cascades.debug.Debugger.sanityCheck(Debugger.java:148)
	at com.apple.foundationdb.record.query.plan.cascades.RequestedOrdering.ofPrimitiveParts(RequestedOrdering.java:305)
	at com.apple.foundationdb.relational.recordlayer.query.LogicalOperator.generateSort(LogicalOperator.java:450)
	at com.apple.foundationdb.relational.recordlayer.query.LogicalOperator.generateSelect(LogicalOperator.java:305)
	at com.apple.foundationdb.relational.recordlayer.query.visitors.QueryVisitor.visitSimpleTable(QueryVisitor.java:252)
	at com.apple.foundationdb.relational.recordlayer.query.visitors.BaseVisitor.visitSimpleTable(BaseVisitor.java:776)
	at com.apple.foundationdb.relational.recordlayer.query.visitors.DelegatingVisitor.visitSimpleTable(DelegatingVisitor.java:609)
	at com.apple.foundationdb.relational.recordlayer.query.visitors.DelegatingVisitor.visitSimpleTable(DelegatingVisitor.java:50)
	at com.apple.foundationdb.relational.generated.RelationalParser$SimpleTableContext.accept(RelationalParser.java:10763)
	at com.apple.foundationdb.relational.recordlayer.query.visitors.DdlVisitor.lambda$visitIndexDefinition$1(DdlVisitor.java:193)
```

This is related to attempting to create a value index that is sorted by a `UUID` field:

```sql
    create table tc(a bigint, b uuid, c bigint, primary key(a, b))
    create index tc1 as select b, c from tc order by b
```

which is now supported as per #3510.

hatyo:
This amends a planner sanity check which verifies that only primitive types are allowed when constructing an ordering such that it also accepts UUIDs.

This fixes #3656.

### DIFF
--- a/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
+++ b/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
@@ -305,7 +305,7 @@ public class RequestedOrdering {
         Debugger.sanityCheck(() -> Verify.verify(
                 requestedOrderingParts.stream()
                         .allMatch(requestedOrderingPart -> requestedOrderingPart.getValue()
-                                .getResultType().isPrimitive())));
+                                .getResultType().isPrimitive() || requestedOrderingPart.getValue().getResultType().isUuid())));
         return new RequestedOrdering(requestedOrderingParts, distinctness, isExhaustive);
     }
 
--- a/yaml-tests/src/test/resources/uuid.yamsql
+++ b/yaml-tests/src/test/resources/uuid.yamsql
@@ -232,6 +232,16 @@ test_block:
   tests:
     -
       - query: select b, c from tc order by b
+      - debugger: insane
+      - explain: "COVERING(TC1 <,> -> [A: KEY[2], B: KEY[0], C: VALUE[0]]) | MAP (_.B AS B, _.C AS C)"
+      - result: [{!uuid '0920df1c-be81-4ec1-8a06-2180226f051d', 6},
+                 {!uuid '5394a912-aa8e-40fc-a4bb-ddf3f89ac45b', 3},
+                 {!uuid '64120112-4e39-40c3-94b9-2cc88a52e8df', 5},
+                 {!uuid '99e8e8b1-ac34-4f4d-9f01-1f4a7debf4d6', 1},
+                 {!uuid 'a8708750-d70f-4800-8c3b-13700d5b369f', 2},
+                 {!uuid 'c35ba01f-f8fc-47d7-bb00-f077e8a75682', 4}]
+    -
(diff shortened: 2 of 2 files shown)


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
```diff
@@ -305,7 +305,7 @@ public static RequestedOrdering ofPrimitiveParts(@Nonnull final List<RequestedOr
         Debugger.sanityCheck(() -> Verify.verify(
                 requestedOrderingParts.stream()
                         .allMatch(requestedOrderingPart -> requestedOrderingPart.getValue()
-                                .getResultType().isPrimitive())));
+                                .getResultType().isPrimitive() || requestedOrderingPart.getValue().getResultType().isUuid())));
```
normen662, line 308: Sorry, I do not agree with this fix. The check makes sure that you never order by something that is not primitive. Isn't a UUID a structure? What happens if I want to order by a UUID column? That should be broken up just like other records should be broken up.

normen662 (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RequestedOrdering.java
hatyo: With #3198 we already added support for UUID as a primitive type in SQL. There is special handling of massaging `TupleFieldsProto.UUID` messages in the planning and execution codepath (that's what the majority of that PR was about).

alecgrieser: Is there a reason that the `requestedOrderingPart.getResultType().isPrimitive()` doesn't return `true` if the type is a UUID?

PR #3658, opened 2025-10-02
Support nullable and non-nullable types with same name in the same template

pengpeng-lu:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -435,13 +433,15 @@ private void addEnumType(@Nonnull final EnumDescriptor enumType,
     public static class Builder {
         private @Nonnull final FileDescriptorProto.Builder fileDescProtoBuilder;
         private @Nonnull final FileDescriptorSet.Builder fileDescSetBuilder;
-        private @Nonnull final BiMap<Type, String> typeToNameMap;
+        private @Nonnull final Map<Type, String> typeToNameMap;
+        private @Nonnull final Map<String, Type> nameToCanonicalTypeMap;
```
alecgrieser: Is the idea here that previously, we had a one-to-one mapping between `Type` and `String` names, whereas now, there can be multiple `Type`s that all map to the same `String` name (as long as the `Type`s are equal ignoring nullability)? And that's why you need to break out the `BiMap` into two maps?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -471,19 +471,46 @@ public Builder setPackage(@Nonnull final String name) {
...
+
         @Nonnull
         public Optional<String> getTypeName(@Nonnull final Type type) {
             return Optional.ofNullable(typeToNameMap.get(type));
         }
 
         @Nonnull
         public Optional<Type> getTypeByName(@Nonnull final String name) {
-            return Optional.ofNullable(typeToNameMap.inverse().get(name));
+            return Optional.ofNullable(nameToCanonicalTypeMap.get(name));
```
alecgrieser: I think this method is unused within the code base. Should we just get rid of it?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -471,19 +471,46 @@ public Builder setPackage(@Nonnull final String name) {
...
             return this;
         }
 
+        /**
+         * Finds a type in typeToNameMap that has the same structure as the given type but different nullability.
+         * Returns null if no such type exists.
+         */
+        @Nullable
+        private Type findCanonicalTypeForStructure(@Nonnull final Type type) {
+            for (Map.Entry<Type, String> entry : typeToNameMap.entrySet()) {
```
alecgrieser: This now introduces an &Theta;(_n_) step into creating the type repository, which would mean that creating a repository with _n_) types is now &Theta;(_n_<sup>2</sup>).

Did you consider an alternative design where we enforce that only the `notNullable` (or `nullable`--it doesn't matter except for certain edge cases like `Type.Null`) variants of each type are kept in the `TypeRepository`? From my reading of the type repository, we don't actually need to keep both the nullable and non-nullable variants of the type at all, as we only use it to associate a type to a Protobuf descriptor, and those are the same for the two different variants.

I think the basic shape of the solution would look something like:

1. Modify the builder to store only the (non) nullable version of each type
2. Wrap the access paths to the `typeToName` bi-map with logic to first transform the type to its (non) nullable variant

yaml-tests/src/test/resources/struct-type-nullability-variants.yamsql
```diff
@@ -0,0 +1,123 @@
...
+        name string,
+        addresses Address array,
+        primary key(id)
+    )
+    
+    create table businesses(
+        id integer,
+        name string,
+        headquarters Address,
+        branch_offices Address array,
```
alecgrieser, line 49: If it I take it, the reason this covers the new logic is that the `Address array` will be an array of non-nullable `Address` structs, whereas the `headquarters Address` will be a nullable type. It might make sense to add an explicit table here where there is both an `Address` column and an `Address NOT NULL` column.

yaml-tests/src/test/resources/struct-type-nullability-variants.yamsql
```diff
@@ -0,0 +1,123 @@
+#
+# struct-type-nullability-variants.yamsql
+#
+# This source file is part of the FoundationDB open source project
+#
+# Copyright 2021-2024 Apple Inc. and the FoundationDB project authors
```
alecgrieser:
```suggestion
# Copyright 2016-2025 Apple Inc. and the FoundationDB project authors
```

yaml-tests/src/test/java/YamlIntegrationTests.java
```diff
@@ -227,6 +226,11 @@ void arrays(YamlTest.Runner runner) throws Exception {
         runner.runYamsql("arrays.yamsql");
     }
 
+    @TestTemplate
+    public void structTypeVariants(YamlTest.Runner runner) throws Exception {
```
alecgrieser: FYI, I believe that since the PR was first created, we've sorted the test names in this file in order to decrease the number of conflicts at the end of the file

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
pengpeng-lu: Sure 👍

yaml-tests/src/test/resources/struct-type-nullability-variants.yamsql
pengpeng-lu: that'll be nice, but currently `NOT NULL` is only allowed for ARRAY column type. I did remember allowing not null columns triggered a bunch of problems 3 years ago, so we only allowed Array type, as there were no real use cases for other types at least back then.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -471,20 +488,50 @@ public Builder setPackage(@Nonnull final String name) {
...
+                return this;
+            }
+
+
+            final BiMap<Type, String> targetMap = type.isNullable() ? nullableTypeToNameMap : nonNullableTypeToNameMap;
+            final BiMap<Type, String> oppositeMap = type.isNullable() ? nonNullableTypeToNameMap : nullableTypeToNameMap;
+
+            if (!targetMap.containsKey(type)) {
+                // Check if the opposite nullability variant exists and get its protobuf name
+                final Type oppositeNullabilityType = type.isNullable() ? type.notNullable() : type.withNullability(true);
```
alecgrieser: I think this could be simplified to:

```java
final Type oppositeNullabilityType = type.withNullability(!type.isNullable());
```

yaml-tests/src/test/resources/struct-type-nullability-variants.yamsql
alecgrieser: I see. That makes sense. I think in theory, we could still use a protobuf file here to define the types and import the schema template that way, though I'm not sure it's worth it. Potentially what we'll actually want is some additional unit tests in `TypeRepositoryTest` just to be explicit

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -85,14 +83,25 @@ public class TypeRepository {
...
+    @Nonnull
+    private final BiMap<Type, String> nullableTypeToNameMap;
+
+    /**
+     * BiMap storing non-nullable types and their protobuf names.
+     * Only types with isNullable() = false are stored here.
+     */
     @Nonnull
-    private final Map<Type, String> typeToNameMap;
+    private final BiMap<Type, String> nonNullableTypeToNameMap;
```
alecgrieser: I think splitting the maps has made the logic in `registerTypeToTypeNameMapping` and `addTypeIfNeeded` a little too complicated. My suggestion was actually to keep everything in one map, but to _only_ put the nullable variants into it. Then I think the only thing that needs to change is that anything that accesses the map needs to call `nullable()` to create the nullable variant before it goes to look things up from the map.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -471,20 +488,50 @@ public Builder setPackage(@Nonnull final String name) {
...
         @Nonnull
         public Builder addTypeIfNeeded(@Nonnull final Type type) {
-            if (!typeToNameMap.containsKey(type)) {
+            // Type.Null is special - it can only be nullable and has no non-nullable variant
+            if (type instanceof Type.Null) {
+                if (!nullableTypeToNameMap.containsKey(type)) {
+                    type.defineProtoType(this);
+                }
+                return this;
+            }
```
alecgrieser: I believe `Relation`s and `Any` are also special in that they also have a fixed nullability. (`Relation`s are always non-null, and `Any` is always nullable.)

It's a bit weird, as none of those types actually override `defineProtoType`, so they won't end up adding anything to the type repository anyway. But I suppose it doesn't hurt to make sure we cover for them. Potentially a method like `canonicalizeNullability(Type)` would be useful that returned (1) a non-nullable `Relation` and `None`, and (2) a nullable value for everything else. Then that would be the type we probed for.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -501,8 +548,37 @@ public Builder addEnumType(@Nonnull final DescriptorProtos.EnumDescriptorProto e
...
+            if (oppositeExistingName != null && oppositeExistingName.equals(protoTypeName)) {
+                // Both nullable and non-nullable variants can share the same protobuf type name
+                targetMap.put(type, protoTypeName);
+                return this;
+            }
+
+            // Check if this protobuf name is already used by a different structural type
+            final Type existingTypeForName = targetMap.inverse().get(protoTypeName);
+            if (existingTypeForName != null && !existingTypeForName.equals(type)) {
+                throw new IllegalArgumentException(String.format(
```
alecgrieser: I think this is actually already handled by `targetMap.put`, though I'm not necessarily opposed to overriding the error message.

What I think we are missing here is coverage of the case:

1. You insert a (let's say) nullable type `t` with a name `x`
2. You insert a type `u` that is not nullable _and_ `!u.nullable().equals(t)` but you also assign it the name `x`

I think you'd need to add a check that `oppositeMap.inverse().get(protoTypeName)` was either null or equal to `oppositeNullabilityType`. I also think that this would naturally get covered if there were just one `BiMap` with a canonicalized nullability

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -26,8 +26,7 @@
 import com.google.common.base.Verify;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
+import com.google.common.collect.ImmutableBiMap;
```
alecgrieser: Commenting here some unit tests that should be added to `TypeRepositoryTest`:

1. A simple unit test adding both a nullable and non-nullable variant of the same type with the same name
2. A test like `addSameTypeMultipleTimesShouldNotCreateMultipleMessageTypes`, but it adds a nullable and non-nullable variant of the same type
3. Ideally, a variant of the above with an enum
4. Check that if we can't add two copies of the same type (even if they have different nullability) with different names. (I think this one is handled correctly by the code already.)
5. Check what happens if we add two types with the same name and different nullability _and_ different contents. This should be an error, but I think it's currently not handled correctly
6. Creating a builder using one nullability of a type, and then calling `getProtoTypeName` or `newMessageBuilder` with the other nullability. I believe this currently throws an error, though I'm not sure it needs to
7. Something like `addPrimitiveTypeIsNotAllowed`, but also including `Any`, `Null`, `None`, and `Relation` types. It would be good to also update `addPrimitiveTypeIsNotAllowed` while you're there, because the assertion is off. That doesn't throw anymore, so having the `catch` is confusing and muddies what the assertion actually is

Maybe these should be in some schema file, but it is probably going to be easier to get the coverage in unit testing

yaml-tests/src/test/resources/struct-type-nullability-variants.yamsql
alecgrieser: I've put some ideas for tests that would be good in another comment on the `TypeRepository` itself

alecgrieser (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/typing/TypeRepository.java
```diff
@@ -272,6 +276,23 @@ public Set<String> getEnumTypes() {
...
+     *
+     * @param type the type to canonicalize
+     * @return the canonicalized type with appropriate nullability
+     */
+    @Nonnull
+    private static Type canonicalizeNullability(@Nonnull final Type type) {
+        if (type.getTypeCode() == Type.TypeCode.RELATION || type.getTypeCode() == Type.TypeCode.NONE) {
+            return type.notNullable();
+        } else {
+            return type.withNullability(true);
```
alecgrieser: minor nit:

```suggestion
            return type.nullable();
```

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TypeRepositoryTest.java
```diff
@@ -163,15 +164,31 @@ private static Type.Vector randomVectorType() {
...
+                Type.TypeCode.DOUBLE, Type.TypeCode.FLOAT, Type.TypeCode.INT,
+                Type.TypeCode.LONG, Type.TypeCode.STRING, Type.TypeCode.VERSION)) {
+            final var primitiveType = Type.primitiveType(primitiveCode, true);
+            builder.addTypeIfNeeded(primitiveType);
+            Assertions.assertTrue(builder.getTypeName(primitiveType).isEmpty(),
+                    "Primitive type " + primitiveCode + " should not be added to repository");
+        }
+
+        // Test special types that should not be allowed
+        List<Type> unsupportedTypes = List.of(
```
alecgrieser: Is there a reason that primitives and unsupported types are being separated out by this test case? Unless I'm missing something, it looks like the asserts are the same. Is it just that the primitive types put their type inside the loop?

It would be nice to have the asserts shared between the two test cases. One way to do this would be to have this be a parameterized test based on the `Type`, and then have a helper method generate the different primitive and "unsupported" types. But it would also be fine just to collect them in a list, and then have them all call the asserts. It may also make sense to add (1) both nullable and non-null primitives and (2) `Type.UUID_NULL_INSTANCE` and `Type.UUID_NON_NULL_INSTANCE`

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/TypeRepositoryTest.java
```diff
@@ -163,15 +164,31 @@ private static Type.Vector randomVectorType() {
...
+                Type.any(),                           // ANY type
+                Type.nullType(),                      // NULL type
+                new Type.None(),                      // NONE type
```
alecgrieser: minor:

```suggestion
                Type.NONE,                            // NONE type
```

PR #3659, opened 2025-10-03
deprecate compiled statement continuations in jdbc

g31pranjal:


fdb-relational-grpc/src/main/java/com/apple/foundationdb/relational/jdbc/TypeConversion.java
```diff
@@ -581,8 +581,6 @@ public static Continuation.Reason toReason(RpcContinuationReason reason) {
     @SuppressWarnings("unchecked")
     static com.apple.foundationdb.relational.jdbc.grpc.v1.Options.Builder toProtobuf(@Nonnull Options options) throws SQLException {
         final var builder = com.apple.foundationdb.relational.jdbc.grpc.v1.Options.newBuilder();
-        // Switched-on by default on JDBC driver until the option is deprecated and removed.
-        builder.setContinuationsContainCompiledStatements(true);
```
alecgrieser, line 585: Should this go into minor version 4.8 as it's intended to work only with 4.7.1.0 and newer? It's fine if it does; we could even bump the minor version in this PR.

PR #3660, opened 2025-10-03
Replace `String.format` calls with string concatenation where possible

### LINKED ISSUE #3645: Metrics diff test uses locality for report formatting
One of the nightly tests failed when asserting on the format of the metrics diff report: https://github.com/FoundationDB/fdb-record-layer/actions/runs/18088622922/attempts/1#summary-51464043202

The test failures are basically from places where it expected the format to be something like `- Average change: -2.0` but instead the format was something like `- Average change: -2,0` (using a comma as the decimal separator instead of a period).

This is a bit of an interesting one, as unlike a lot of the cases where the answer has been to set the locale to `Locale.ROOT` explicitly, there could be an argument that this format should respect the user's locale so that users who want numbers formatted like that can get them that way. But it would certainly be easier to just force the format to follow that format, which is what we do in the rest of the codebase.

alecgrieser:
This goes through our codebase and audits our `String.format` usage. It makes a few changes:

1. If the `String.format` call can be replaced with `String` concatenation, it does so. This mainly means taking cases where strings or numbers are naively placed together (e.g., `String.format("%s -> %d", myStr, myInt)`) and replacing it with the equivalent concatentation (e.g., `myStr + " -> " + myInt`). The reason this is done is efficiency: it is more efficient to construct the strings and then concatenate them than to use the Java `String` formatter most of the time, mainly because the format string parsing is done at runtime with every invocation rather than compiled into some kind of formatting object.
1. In places where that is more difficult, this ensures that the `String.format` call has set `Locale.ROOT` to avoid localization problems. This is what caused #3645, as the `MetricsDiffAnalyzer` was formatting numbers with decimals, and then test asserts were firing if the test ran with a different locale, which is a thing we do in our nightly tests. In theory, that's actually sort of appropriate for a tool like this (for example, it would make sense to localize `3.5` to `3,5` if run on a system where that is the convention), but the rest of the product doesn't support it, and it's not like the actual text is localized.

When running on a system with `Locale.ROOT`, this should be a transparent change.

This fixes #3645.

### DIFF
--- a/fdb-relational-api/src/test/java/com/apple/foundationdb/relational/api/DataTypeSmokeTest.java
+++ b/fdb-relational-api/src/test/java/com/apple/foundationdb/relational/api/DataTypeSmokeTest.java
@@ -20,7 +20,12 @@
 
 package com.apple.foundationdb.relational.api;
 
+import com.apple.foundationdb.relational.api.exceptions.ErrorCode;
+import com.apple.foundationdb.relational.api.exceptions.UncheckedRelationalException;
+import com.apple.foundationdb.relational.api.fluentsql.expression.Field;
+import com.apple.foundationdb.relational.api.fluentsql.expression.UserDefinedField;
 import com.apple.foundationdb.relational.api.metadata.DataType;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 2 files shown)


fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/PlannerGraphTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final PlannerGraph.Node root = graph.getRoot();
+        assertThat(root)
+                .isInstanceOf(PlannerGraph.LogicalOperatorNode.class);
+        assertThat(root.getName())
+                .isEqualTo("SELECT " + select.getResultValue());
+        assertThat(root.getDetails())
+                .hasSize(1);
+        final String predicateDetail = root.getDetails().get(0);
+        assertThat(predicateDetail)
+                .as("large where clauses should be converted into a hex representation of their hash")
```
normen662, line 96: I think I am going to revert that particular "feature".

normen662 (approved):
I guess we want to do one better in Java 21 using JSR 430.

Issue #3583, opened 2025-09-10
SQL Error Code Consistency


hatyo:
There are situations where we get different SQL error codes for the _same_ SQL statement, depending on how it handled in the SQL processing pipeline.

For example:

```sql
select * from foo where a in ('bar', 34);
-- throws 22000
```

this makes sense, considering that `'bar'` and `34` have incompatible types.

however:

```sql
select * from foo where a in ('bar', 'bla');
-- works correctly

select * from foo where a in ('bar', 34);
-- throws XX000
```

this happens because the first statement is a valid statement that is correctly executed and cached. The _second_ executes picks the statement from the cache, attempts to run it with constants `'bar'` and `34`, which fails, the returned error is translated to `XX000` because of this:

https://github.com/FoundationDB/fdb-record-layer/blob/7a54579769ac55e62f68fa60ff7da112b54721c7/fdb-relational-core/src/main/java/com/apple/foundationdb/relational/recordlayer/query/PlanGenerator.java#L199-L202

i.e. a `SemanticException` is immediately translated to `RelationalException` with `ErrorCode.INTERNAL_ERROR`.

PR #3668, opened 2025-10-10
Add documentation to `CAST` operator.

### LINKED ISSUE #3662: Document `CAST` operator
With https://github.com/FoundationDB/fdb-record-layer/pull/3647 we introduce support for `CAST` operator. We should document this operator, along with a list of supported data type `CAST`s.

hatyo:
With https://github.com/FoundationDB/fdb-record-layer/pull/3647 we introduce support for `CAST` operator, this adds documentation of this operator, including supported casts, error codes, and degree of compliance to the SQL standard.

This fixes #3662.

### DIFF
--- a/docs/sphinx/source/reference/Functions/cast_operator.rst
+++ b/docs/sphinx/source/reference/Functions/cast_operator.rst
@@ -27,10 +27,10 @@ Numeric Conversions
 -------------------
 
 * Integer types: :sql:`INTEGER` ↔ :sql:`BIGINT`
-* Floating-poINTEGER types: :sql:`FLOAT` ↔ :sql:`DOUBLE`
+* Floating-point types: :sql:`FLOAT` ↔ :sql:`DOUBLE`
 * Mixed numeric: :sql:`INTEGER`/:sql:`BIGINT` ↔ :sql:`FLOAT`/:sql:`DOUBLE`
 * Narrowing conversions (e.g., :sql:`BIGINT` → :sql:`INTEGER`, :sql:`DOUBLE` → :sql:`FLOAT`) validate range and throw errors on overflow
-* Floating-poINTEGER to INTEGER conversions use rounding (Math.round)
+* Floating-point to INTEGER conversions use rounding (Math.round)
 
 String Conversions
 ------------------


docs/sphinx/source/reference/Functions/cast_operator.rst
```diff
@@ -0,0 +1,250 @@
...
+####
+
+The CAST operator converts a value from one data type to another. CAST supports explicit type conversions according to SQL standard semantics.
+
+Syntax
+######
+
+.. code-block:: sql
+
+    CAST(expression AS target_type)
```
arnaud-lacurie, line 12: To remain consistent with the rest of the doc, you'd want to put a railroad diagram here.

docs/sphinx/source/reference/Functions/cast_operator.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Parameters
+##########
+
+* :sql:`expression`
+    The value or expression to be converted.
+
+* :sql:`target_type`
+    The target data type. Can be a primitive type (:sql:`INT`, :sql:`LONG`, :sql:`FLOAT`, :sql:`DOUBLE`, :sql:`STRING`, :sql:`BOOLEAN`) or an array type (e.g., :sql:`INTEGER ARRAY`, :sql:`STRING ARRAY`).
```
arnaud-lacurie: INT and LONG should be replaced by INTEGER and BIGINT throughout the document

docs/sphinx/source/reference/Functions/cast_operator.rst
```diff
@@ -0,0 +1,250 @@
...
+------------------
+
+* Any primitive type can be converted to :sql:`STRING`
+* :sql:`STRING` can be converted to numeric types (:sql:`INT`, :sql:`LONG`, :sql:`FLOAT`, :sql:`DOUBLE`) with validation
+* Invalid string-to-numeric conversions throw errors
+
+Boolean Conversions
+-------------------
+
+* :sql:`BOOLEAN` ↔ :sql:`INT`: true = 1, false = 0
```
arnaud-lacurie: This is probably a uni-directional arrow, given the fact that the other direction is written the line underneath.

docs/sphinx/source/reference/Functions/cast_operator.rst
```diff
@@ -0,0 +1,250 @@
...
+
+String values that cannot be parsed as numbers result in errors:
+
+.. code-block:: sql
+
+    CREATE TABLE data(id BIGINT, str_value STRING, PRIMARY KEY(id))
+    INSERT INTO data VALUES (1, 'invalid')
+
+    SELECT CAST(str_value AS INTEGER) FROM data WHERE id = 1
+    -- Error: Cannot cast string 'invalid' to :sql:`INT`
```
arnaud-lacurie: Is that the actual error message? 
I guess the error message should not contain `:sql:` since this is a RST thing.

docs/sphinx/source/reference/Functions/cast_operator.rst
hatyo: Done.

docs/sphinx/source/reference/Functions/cast_operator.rst
hatyo: Nice catch, done.

docs/sphinx/source/reference/Functions/cast_operator.rst
hatyo: Yeah, the exact error message is: `Value out of range for INT`. There is something to be said about rewriting error messages originating from the Cascades types system validations and the upper relational layer SQL types., and overall consistency of SQL error messages, as in e.g. https://github.com/FoundationDB/fdb-record-layer/issues/3583.

arnaud-lacurie:
Looks like you had a search replace mishap, can you fix that?

hatyo:
> Looks like you had a search replace mishap, can you fix that?

You're right, I reverted the newly invented type `floating poINTEGER` back to `floating point`. Good observation!

docs/sphinx/source/reference/Functions/cast_operator.rst
hatyo: Actually, that's not what we do for functions, see: https://foundationdb.github.io/fdb-record-layer/reference/Functions.html, which makes sense, considering that railroad diagrams are normally created for SQL statements, and not scalar expressions.

PR #3669, opened 2025-10-10
Use the LuceneSerializer to encode stored fields and field infos, too.

MMcM:
Since existing fields will be straight protobuf without the serializer's prefix byte, recognize that, too.


fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
```diff
@@ -270,4 +270,36 @@ private void decryptIfNeeded(@Nonnull CompressedAndEncryptedSerializerState stat
...
+        if (isProtobufMessageWithoutPrefix(bytes)) {
+            return bytes;
+        }
+        return decode(bytes);
+    }
+
+    // This can be removed once it is guaranteed that all indexes are using the encoded format.
+    // Only works for Protobuf messages all of whose fields are themselves length-delimited.
+    private boolean isProtobufMessageWithoutPrefix(@Nonnull byte[] bytes) {
+        if (bytes.length < 1) {
```
ohadzeliger, line 306: Is this relying on the fact that the first field in the StoredFields message is an `int32` and is required?

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneSerializerTest.java
```diff
@@ -122,4 +121,52 @@ void testEncodingWithEncryption(boolean compressToo) throws Exception {
         final LuceneFileSystemProto.LuceneFileReference decryptedReference = LuceneFileSystemProto.LuceneFileReference.parseFrom(decodedValue);
         Assertions.assertEquals(content, decryptedReference.getContent());
     }
+
+    @ParameterizedTest
+    @MethodSource("encodedCompressedAndEncrypted")
+    void testProtobufMessageWithoutPrefix(boolean encode, boolean compress, boolean encrypt) throws Exception {
```
ohadzeliger: Probably would be good to test the `FIeldInfos` as well - the structure is more flexible than `StoredFields` as all fields are optional

ohadzeliger (changes requested):

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/LuceneSerializer.java
MMcM: Mostly yes. The prefix isn't really a Protobuf `int32`, but it's a varint, using the same encoding.
Mainly, though, it's just that an empty byte array cannot have a prefix, no matter how that prefix is encoded (well, assuming it takes _some_ bytes).
So this must be an empty Protobuf message (no fields).

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneSerializerTest.java
MMcM: That's a difference between `FieldInfo` and `StoredField`, not between `FieldInfos` and `LuceneStoredFields`. Both of the top-level messages have zero or more nested messages, which is what allows the determination.

Is it worth testing each one with the decoding as a black box? In that case, are there other interesting cases that you can think of?

MMcM: For `FieldInfos`, there are two cases:
* no `FieldInfo`: an empty byte array.
* a `FieldInfo` as field number `1`, length delimited, so first byte `0A`, which fails all the conditions in `isProtobufMessageWithoutPrefix`.

MMcM: A `LuceneStoredFields` might additionally have only a `primaryKey` as field number `2`, also length delimited, so first byte `12`.

ohadzeliger: My concern was that we may interpret a non-empty protobuf as an encryption prefix. Having a non-empty `FieldInfos` with various types populated should be sufficient.
In reality we're likely OK since the first byte would always be `LEN` and the field number for the embedded message is always '1'.

ohadzeliger: Maybe also storing no `StoredField` message in `StoredFields` and setting some value to `primaryKey`?

MMcM: I added an explicit test for each of the possible cases of "what comes first" for both of the message types. So, not totally a black box, but more of one. Let me know if there is another case that seems worth covering to you.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneSerializerTest.java
```diff
@@ -122,4 +122,104 @@ void testEncodingWithEncryption(boolean compressToo) throws Exception {
         final LuceneFileSystemProto.LuceneFileReference decryptedReference = LuceneFileSystemProto.LuceneFileReference.parseFrom(decodedValue);
         Assertions.assertEquals(content, decryptedReference.getContent());
     }
+
+    @ParameterizedTest
```
ScottDugas, line 126: Probably outside the scope of this PR, I wonder if it would make sense to have `TestFDBDirectory` (used by the test framework provided by lucene) optionally encode/compress/encrypt.

fdb-record-layer-lucene/src/main/java/com/apple/foundationdb/record/lucene/directory/FDBDirectory.java
```diff
@@ -444,10 +448,11 @@ public int writeData(final long id, final int block, @Nonnull final byte[] value
      * Write stored fields document to the DB.
      * @param segmentName the segment name writing to
      * @param docID the document ID to write
-     * @param value the bytes value of the stored fields
+     * @param rawBytes the bytes value of the stored fields
      */
-    public void writeStoredFields(@Nonnull String segmentName, int docID, @Nonnull final byte[] value) {
+    public void writeStoredFields(@Nonnull String segmentName, int docID, @Nonnull final byte[] rawBytes) {
         byte[] key = storedFieldsSubspace.pack(Tuple.from(segmentName, docID));
+        byte[] value = serializer.encode(rawBytes);
```
ScottDugas: If this is running at the same time as older versions, will older versions of the code be unable to read the data that this writes?

MMcM: I'll add a new context property to ensure that this is safe.

MMcM: Added.

fdb-record-layer-lucene/src/test/java/com/apple/foundationdb/record/lucene/LuceneSerializerTest.java
MMcM: If it's optional, won't the control end up being equivalent to the properties used by the superclass? Or is it that we'd like the default to be on for testing, even if it's off otherwise?

ohadzeliger (approved):
LGTM

PR #3680, opened 2025-10-15
Support SQL views

### LINKED ISSUE #3670: Support SQL views
This can be thought of as a succession of #3282, with focus on supporting SQL views in the schema template definition:

```sql
create schema template myTemplate
create table t1(col1 integer, col2 string, primary key (col1))
create view v1 as select * from t1 where col1 < 200
```

Similar to table-valued functions, the view definition is saved in the metadata, and its expansion is amortized:

```sql
select * from v1;
```

later on, we should also support the creation and deletion of transaction-based temporary views, similar to temporary functions, for completeness.

hatyo:
This introduces support for SQL views. Views are virtual tables defined by SQL queries that are compiled lazily when referenced, providing a way to encapsulate complex queries and present simplified interfaces to users.

**Implementation Overview**

Views are stored as raw SQL text in the catalog and compiled on-demand during query execution. The implementation follows a layered architecture:

At the core layer, we introduce `RawView` which represents the persisted view definition. Views are stored in the metadata alongside user-defined functions and are serialized/deserialized using the existing protobuf infrastructure. The `RecordMetaData` and `RecordMetaDataBuilder` classes now maintain a map of views that get persisted with the schema.

At the relational API layer, `RecordLayerView` wraps the raw view definition with compilation capabilities. Each view contains a lazy compilation function that transforms the SQL text into a logical plan when the view is first referenced. This compiled plan can be cached to avoid recompilation on subsequent accesses.

The query parser has been extended to recognize view references in `FROM` clauses. When a view is encountered, the semantic analyzer retrieves the view definition from the catalog and triggers compilation by invoking the view's compilation function. The resulting logical operator is then seamlessly integrated into the query plan, allowing views to be used just like regular tables.

**Features**

  - View Definition Storage: Views are defined using `CREATE VIEW` syntax and stored as SQL text in the schema template
  - Lazy Compilation: View definitions are compiled into logical plans only when referenced in queries
  - Nested Views: Views can reference other views, enabling layered abstractions
  - Complex Queries: Views support CTEs, joins, aggregations, and other advanced SQL features
  - Schema Integration: Views are first-class metadata objects alongside tables and functions

The implementation includes comprehensive testing through YAML integration tests that validate various view scenarios including simple filters, nested views, CTEs within views, and views referencing user-defined functions.

This fixes #3670.


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/View.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.apple.foundationdb.record.query.plan.cascades;
+
+import com.apple.foundationdb.record.PlanSerializationContext;
+import com.apple.foundationdb.record.RecordMetaDataProto;
+
+import javax.annotation.Nonnull;
+
+public abstract class View {
```
normen662: Given that there is no state, can this be an interface?

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/View.java
```diff
@@ -0,0 +1,58 @@
...
+ * Metadata for a non-materialized view, which represents a virtual table defined by a SQL query.
+ * <p>
+ * The view definition is persisted to disk as a plain SQL string. At runtime, implementations
+ * of this interface may cache the compiled logical plan to avoid recompiling the view definition
+ * on every access.
+ * <p>
+ * Views can be either permanent (schema-bound) or temporary (transaction-bound). While the interface
+ * supports temporary views, this functionality is not currently implemented.
+ */
+public interface View extends Metadata {
```
normen662, line 35: There is another `View` class that is introduced in `record-layer-core`. Let's not do that sort of overloading.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RawView.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * RawSqlView.java
```
normen662:
```suggestion
 * RawView.java
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/View.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Nonnull
+    public abstract String getName();
+
+    @Nonnull
+    public abstract RecordMetaDataProto.PView toProto(@Nonnull PlanSerializationContext serializationContext);
```
normen662: 1. This should probably return `Message` as other implementors very likely need to create other messages.
2. Why do you need the `PlanSerializationContext` here? I mean no implementor can make use of that here.

fdb-record-layer-core/src/main/proto/record_metadata.proto
```diff
@@ -205,6 +206,17 @@ message PUserDefinedFunction {
   }
 }
 
+message PView {
+  oneof specific_view {
+    PRawView raw_view = 1;
```
normen662: You probably need an extension mechanism here. See for instance `record_query_plan.proto` for details. For instance `PValue` there has:
```
extensions 5000 to max;
```
in the message and:
```
google.protobuf.Any additional_values = 1;
```
in the `oneof`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -234,6 +238,12 @@ private void loadProtoExceptRecords(@Nonnull RecordMetaDataProto.MetaData metaDa
                             PlanHashable.CURRENT_FOR_CONTINUATION), function);
             userDefinedFunctionMap.put(func.getFunctionName(), func);
         }
+        for (RecordMetaDataProto.PView viewProto: metaDataProto.getViewsList()) {
+            final RawView rawView = (RawView)PlanSerialization.dispatchFromProtoContainer(
```
normen662: shouldn't that be a `(View)`?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -234,6 +238,12 @@ private void loadProtoExceptRecords(@Nonnull RecordMetaDataProto.MetaData metaDa
...
+                    new PlanSerializationContext(DefaultPlanSerializationRegistry.INSTANCE,
```
normen662: I don't like piggy-backing on plan serialization to just be able to do the dispatch on a multitude of views here . How many kinds of views do we have right now (I think one) -- how many do you think we will have later?
Suggestions:
1. extricate the logic from the plan serialization logic and make the specific plan serialization logic use and share  that logic with this code.
2. dynamic dispatch in the same way as that logic but with a switch statement if the number of actual implementations is like `1` or `2` ever.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
```diff
@@ -115,6 +116,8 @@ public class RecordMetaDataBuilder implements RecordMetaDataProvider {
     @Nonnull
     private final Map<String, UserDefinedFunction> userDefinedFunctionMap;
     @Nonnull
+    private final Map<String, RawView> viewMap;
```
normen662: Why is this not of a map that maps to `View`? The `View`/`RawView` should give you some sort of polymorphism here but the code just wants `RawView` as it seems.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/View.java
normen662: Reading the contextual code I think I do understand why the plan serialization-specific interface is used here now. I am not a fan, though.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
hatyo: Good point, I originally wanted to introduce a forward-looking representation of metadata that leaves the door open for introducing other flavors of views, such as fully expanded, partially expanded, and so on. But I don't think this is realistic, moreover, views in their current shape at least, are pure metadata artifacts, and you're right for calling out their useless dependence of the plan serialization.

Having said that, I changed the metadata representation of the views such that it is now a pure metadata artifacts whose Ser/De has nothing to do anymore with the plan serialization API.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
hatyo: This is changed

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/RecordMetaDataBuilder.java
hatyo: Yes, this is changed now.

fdb-record-layer-core/src/main/proto/record_metadata.proto
hatyo: This is changed, `View` is a simple message now (see my reply below).

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/View.java
hatyo: This is fixed, view serialization and deserialization does not piggy pack on the plan serialization logic anymore, since views are pure metadata artifacts, see below reply for more details.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RawView.java
hatyo: Done.

fdb-relational-api/src/main/java/com/apple/foundationdb/relational/api/metadata/View.java
hatyo: this `View` is part of the relational metadata, that may or may not be backed by the record metadata.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/View.java
hatyo: This is now changed.

PR #3681, opened 2025-10-16
Consider number of predicates at each level in rewriting cost model

hazefully:
This PR improves the rewriting cost model by making it consider the number of predicates in each expression at each level of the expression graph, instead of comparing graphs based on the height of highest expressions in the graph that contains any predicates.

The reasoning for this is to push the cost model to prefer expressions with the same number of predicates, where some predicates were pushed from higher levels in the QGM to lower levels, which can lead to producing plans with more specific index key comparisons in the planning phase. 

Example: the rewriting cost model should consider the expression: 
```
SELECT sq1.a FROM (SELECT a, b FROM T WHERE a = 42) sq1,  (SELECT a FROM T2) sq2
```
to have a lower cost than:
```
SELECT a FROM (SELECT a, b, d FROM T) WHERE a = 42 AND EXISTS (SELECT a FROM T2)
```

The rewriting cost model is improved by considering two `ExpressionProperty` when comparing two expressions: the existing `NormalizedResidualPredicateProperty` property which calculates the total number of conjuncts in the combined query predicate across the entire expression tree at all levels and a new property `PredicateCountByLevelProperty` which calculates the number of predicates at each level of the expression graph. In addition, considering the property `PredicateComplexityProperty` is no longer necessary as what it checks for (the most complex predicate across the entire expression tree) is already checked indirectly by `NormalizedResidualPredicateProperty`.

## Performance impact

Some basic analysis of the performance impact of considering those properties can be found in https://github.com/FoundationDB/fdb-record-layer/pull/3681#discussion_r2632080223.

### DIFF
--- a/fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/HNSWTest.java
+++ b/fdb-extensions/src/test/java/com/apple/foundationdb/async/hnsw/HNSWTest.java
@@ -334,7 +334,6 @@ class HNSWTest {
         assertThat(readIds.size()).isBetween(10, 100);
     }
 
-    @ExtendWith(HNSWTest.DumpLayersIfFailure.class)
     @ParameterizedTest
     @MethodSource("randomSeedsWithConfig")
     void testBasicInsertDelete(final long seed, final Config config) {
@@ -399,7 +398,6 @@ class HNSWTest {
                                 .limit(k)
                                 .map(PrimaryKeyVectorAndDistance::getPrimaryKey)
                                 .collect(ImmutableSet.toImmutableSet());
-
                 onReadListener.reset();
 
                 final long beginTsQuery = System.nanoTime();
@@ -721,6 +719,15 @@ class HNSWTest {
         HNSW.scanLayer(config, rtSubspace.getSubspace(), db, layer, batchSize, nodeConsumer);
     }
 
+    private int getEntryLayer(@Nonnull final Config config) {
+        @Nullable final AccessInfo accessInfo = db.run(readTransaction ->
+                StorageAdapter.fetchAccessInfo(config, readTransaction, rtSubspace.getSubspace(),
+                        OnReadListener.NOOP).join());
+        return accessInfo == null
+               ? -1
+               : accessInfo.getEntryNodeReference().getLayer();
+    }
+
     private long dumpLayer(@Nonnull final Config config,
                            @Nonnull final String prefix, final int layer) throws IOException {
         final Path verticesFile = tempDir.resolve("vertices-" + prefix + "-" + layer + ".csv");
@@ -854,12 +861,15 @@ class HNSWTest {
         }
 
         private void dumpLayers(@Nonnull final HNSWTest hnswTest, @Nonnull final Config config) {
-            int layer = 0;
-            while (true) {
+            final int entryLayer = hnswTest.getEntryLayer(config);
+
+            if (entryLayer < 0) {
+                return;
+            }
+
+            for (int layer = 0; layer < entryLayer; layer ++) {
                 try {
-                    if (hnswTest.dumpLayer(config, "debug", layer++) == 0) {
-                        break;
-                    }
+                    Verify.verify(hnswTest.dumpLayer(config, "debug", layer) > 0);
                 } catch (IOException e) {
                     throw new RuntimeException(e);
                 }


yaml-tests/src/test/resources/standard-tests.metrics.yaml
```diff
@@ -114,26 +114,26 @@ standard-tests:
...
-    insert_new_count: 315
-    insert_reused_count: 44
+    task_count: 1325
+    task_total_time_ms: 267
+    transform_count: 270
+    transform_time_ms: 153
+    transform_yield_count: 83
+    insert_time_ms: 12
+    insert_new_count: 165
+    insert_reused_count: 24
```
hazefully, line 124: I think the planning metrics for the two queries here have improved because the rewriting cost model now prefers expressions with less total number of predicates, so when the predicates in those two queries are simplified to remove the duplicated predicate, we end up with a smaller search space. Before the change here, the two expressions (the original expressions with the duplicated predicate and the simplified one) were considered equal and the semantic hashcode tie breaking resulted in the original expression being chosen.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class PredicateCountByLevelProperty implements ExpressionProperty<PredicateCountByLevelProperty.PredicateCountByLevelInfo> {
```
normen662, line 56: Needs a javadoc.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+         * @param  a the first {@code PredicateCountByLevelInfo} to compare
+         * @param  b the second {@code PredicateCountByLevelInfo} to compare
+         * @return the value {@code 0} if {@code a} have the same number of predicates at each level as {@code b};
+         *         a value less than {@code 0} if a {@code a} has fewer predicates than {@code b} at a
+         *         deeper level or if {@code a} has a lower number of levels with predicates; and
+         *         a value greater than {@code 0} if {@code a} has more predicates than {@code b} at a
+         *         deeper level or if {@code a} has a higher number of levels with predicates.
+         */
+        public static int compare(final PredicateCountByLevelInfo a, final PredicateCountByLevelInfo b) {
+            final int highestLevel = Integer.max(a.getHighestLevel(), b.getHighestLevel());
```
normen662: Wouldn't it be more concise to use two `SortedMap`s (`TreeMap`). You can then pop them off smallest first on both sides. In that way you can get rid of the `highestLevel` field which is redundant in your data structure.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+    @Nonnull
+    public static PredicateCountByLevelProperty predicateCountByLevel() {
+        return PREDICATE_COUNT_BY_LEVEL;
+    }
+
+    public static final class PredicateCountByLevelInfo {
+        private final Map<Integer, Integer> levelToPredicateCount;
+        private final int highestLevel;
+
+        public PredicateCountByLevelInfo(Map<Integer, Integer> levelToPredicateCount, int highestLevel) {
```
normen662: Everything `public` needs a javadoc. Will not annotate stuff like that everywhere. Maybe you can go through it and just add that when you one missing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        /**
+         * Combine a list of PredicateCountByLevelInfo instances by adding the number of predicates at the same level
+         * across all instances.
+         *
+         * @param heightInfos a collection of {@code PredicateCountByLevelInfo} instances
+         * @return an instance of {@code PredicateCountByLevelInfo} with predicate counts added at each level.
+         */
+        public static PredicateCountByLevelInfo combine(Collection<PredicateCountByLevelInfo> heightInfos) {
```
normen662, line 126: Every return needs a nullability annotation. Again, won't point out things individually. Please go through the whole things and add them when you see one. I know, we should probably enforce that but we haven't gotten around to doing it. Also, every time this comes up we end up in an argument whether that is useful or not. But it's the prevalent cod style in this part of the woods as of now.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, Integer::sum));
+            return new PredicateCountByLevelInfo(newLevelToComplexityMap, highestLevel);
+        }
+
+        /**
+         * Get a view of the predicate count at each level as a map.
+         *
+         * @return a map of level heights to the count of predicates at that level.
+         */
+        public Map<Integer, Integer> getLevelToPredicateCount() {
```
normen662: `@Nonnull`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+         * Retrieves the highest level (i.e. height of root node) in this instance
+         *
+         * @return integer representing the highest level.
+         */
+        public int getHighestLevel() {
+            return highestLevel;
+        }
+
+        /**
+         * Compares two {@code PredicateCountByLevelInfo} instances level by level.
```
normen662: Codestyle nitpicking: This and other like it should really be a `{@link ...}`

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+
+        @Nonnull
+        private List<PredicateCountByLevelInfo> fromChildren(@Nonnull final RelationalExpression expression) {
+            return expression.getQuantifiers()
+                    .stream()
+                    .map(quantifier -> forReference(quantifier.getRangesOver()))
+                    .collect(ImmutableList.toImmutableList());
+        }
+
+        private PredicateCountByLevelInfo forReference(@Nonnull final Reference reference) {
```
normen662: I think we should not have these properties tracked as there is no decision making being done by the plan partitioning logic. We exclusively use these in the cost model. Can you change this logic here to not grab it from the properties map but to recursively compute it? Take a look at e.g. `DerivationsProperty` for how that looks like.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountProperty.java
```diff
@@ -0,0 +1,95 @@
...
+import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpressionVisitorWithDefaults;
+import com.apple.foundationdb.record.query.plan.cascades.expressions.RelationalExpressionWithPredicates;
+import com.google.common.base.Verify;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+public final class PredicateCountProperty implements ExpressionProperty<Integer> {
```
normen662: Javadoc is missing.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountProperty.java
```diff
@@ -0,0 +1,95 @@
...
+            return expression.getQuantifiers()
+                    .stream()
+                    .map(quantifier -> forReference(quantifier.getRangesOver()))
+                    .collect(ImmutableList.toImmutableList());
+        }
+
+        private int forReference(@Nonnull final Reference reference) {
+            final var finalExpressions = reference.getFinalExpressions();
+            Verify.verify(finalExpressions.size() == 1);
+            final var memberResults = reference.getPropertyForExpressions(property).values();
```
normen662: Same here: We should not introduce this property as a tracked property (i.e. the expression property map has it) as the partitioning logic for plans, expressions does not need/use it. Can you descend here instead of picking up the child value from the map?

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
```diff
@@ -285,8 +287,13 @@ public <P> Map<RecordQueryPlan, P> propertyValueForPlans(@Nonnull final Expressi
     public static ExpressionPropertiesMap<RelationalExpression> defaultForRewritePhase() {
         return new ExpressionPropertiesMap<>(RelationalExpression.class,
                 ImmutableSet.of(),
-                ImmutableSet.of(ExpressionCountProperty.selectCount(), ExpressionCountProperty.tableFunctionCount(),
-                        PredicateComplexityProperty.predicateComplexity()),
+                ImmutableSet.of(
```
normen662, line 288: Ah, I think this should really `selectCount` and `tableFunctionCount` in there. I think `predicateComplexity` should not be in here but 🤷 . I think the new ones you are introducing should not be here.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java
```diff
@@ -72,12 +75,24 @@ public int compare(final RelationalExpression a, final RelationalExpression b) {
...
 
         //
-        // Pick the expression where predicates have been pushed down as far as they can go
+        // Pick the expression where that has the least number of predicates overall
         //
-        int aPredicateHeight = predicateHeight().evaluate(a);
-        int bPredicateHeight = predicateHeight().evaluate(b);
-        if (aPredicateHeight != bPredicateHeight) {
-            return Integer.compare(aPredicateHeight, bPredicateHeight);
+        int aPredicateCount = predicateCount().evaluate(a);
```
normen662: How does all of this (not picking out this particular line but this entire added block here) interact with the predicate complexity? Wouldn't predicate complexity be subsuming the predicate height property you are introducing? Could we put the predicate complexity in front of the count-by-layer property here?

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelPropertyTest.java
```diff
@@ -0,0 +1,207 @@
...
+        final PredicateCountByLevelProperty.PredicateCountByLevelInfo bInfo = new PredicateCountByLevelProperty.PredicateCountByLevelInfo(
+                Map.of(1, 1, 2, 2, 3, 1), 3);
+
+        assertThat(PredicateCountByLevelProperty.PredicateCountByLevelInfo.compare(aInfo, bInfo)).isPositive();
+    }
+
+    @Test
+    void compareReturnsInfoWithMoreLevelsInCaseOfEquality() {
+        final PredicateCountByLevelProperty.PredicateCountByLevelInfo aInfo = new PredicateCountByLevelProperty.PredicateCountByLevelInfo(
+                Map.of(1, 1, 2, 3, 3, 1), 3);
```
normen662: Another weird idiosyncrasy: All maps in the record layer have to be `ImmutableMap` (guava) instead of `java.util.Map`. The reason is that their copy-constructors avoid a copy if the source is a `ImmutableMap`. `Map` does that, too, but only if the source is a `Map` as well. So while it would be better to have the entire codebase on `Map` and not on `ImmutableMap`, someone would have to change everything first. The same applies to lists and sets as well.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
```diff
@@ -0,0 +1,184 @@
...
+    public static PredicateCountByLevelProperty predicateCountByLevel() {
+        return PREDICATE_COUNT_BY_LEVEL;
+    }
+
+    public static final class PredicateCountByLevelInfo {
+        private final Map<Integer, Integer> levelToPredicateCount;
+        private final int highestLevel;
+
+        public PredicateCountByLevelInfo(Map<Integer, Integer> levelToPredicateCount, int highestLevel) {
+            this.levelToPredicateCount = Collections.unmodifiableMap(levelToPredicateCount);
```
normen662: Can you make this an `ImmutableMap`?

yaml-tests/src/test/resources/standard-tests.metrics.yaml
normen662: It looks almost suspicious that the time spent is higher even though all the counts are lower. Can you do a test for me? Remove all metrics files (`metrics.binpb` and `metrics.yaml`) try to run the entire suite in correction mode. Just glance over the time spent to see if there is a trend up or downwards. In general, this sort of stuff happens that we see confusing time durations as everyone uses a different set-up, had their Mac throttled or not, etc.  -- so a sample size of one is pretty naive to comment on but still let's run this to just rule out the case that for some reason a rule has degraded. Also, you may want to do this for down stream.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
hazefully: Thanks for the suggestion, that is indeed better! I modified the code to do that and removed the `highestLevel` field in the data structure. There is a way to do this with the Java Streams API but I found it to be less readable than a simple for-loop, let me know if you think otherwise or if there is a better way to implement this.

I did have to keep the `getHighestLevel` method on the data structure just because the `ImmutableSortedMap` throws an exception for empty maps if `lastKey` is called on them so having the method is a neat way to wrap that check.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
hazefully: All public entities should have a javadoc now, thank you!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
hazefully: All methods returning non-primitive types have an annotation now, thanks!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java
hazefully: Just a note that I had to leave `predicateComplexity` here because it is used in the [SelectMergeRule](https://github.com/FoundationDB/fdb-record-layer/blob/aaf3961434b95cae7d8661c30d39dc2c76384ffe/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/rules/SelectMergeRule.java#L83) in a code path that expects it to be tracked (https://github.com/FoundationDB/fdb-record-layer/blob/aaf3961434b95cae7d8661c30d39dc2c76384ffe/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/matching/structure/ExpressionsPartitionMatchers.java#L114-L118).

fdb-record-layer-core/src/test/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelPropertyTest.java
hazefully: I replaced all usages with that, thanks for the clarification!

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
hazefully: As discussed offline, my main motivation of tracking the new property is to avoid re-evaluating it for the same references multiple times. However, to see if that's actually worth it or not, I did some comparison with the YAML tests metrics with the properties tracked and untracked. 

As you suggested in https://github.com/FoundationDB/fdb-record-layer/pull/3681#discussion_r2622249820, I went ahead and re-ran all YAML tests on the main branch locally with all metric files deleted, used those metrics as a baseline, and then used a modified version of [MetricsDiffAnalyzer.java](https://github.com/FoundationDB/fdb-record-layer/blob/main/yaml-tests/src/main/java/com/apple/foundationdb/relational/yamltests/MetricsDiffAnalyzer.java) to analyse the differences, and it looks like there is very little if any difference. I have done a few instances of the runs and there is variance in the time metrics between the runs that is big enough to make this inconclusive. For example, this is the analysis for one of the runs with the properties untracked (the values are in milliseconds):

```
### Statistical Summary (Only Metrics Changed)

**`transform_time_ns`**:
  - Average change: -0.3
  - Average regression: +4.2
  - Median change: -1
  - Median regression: +2
  - Standard deviation: 10.6
  - Standard deviation of regressions: 10.0
  - Range: -139 to +107
  - Range of regressions: +1 to +107
  - Queries changed: 427
  - Queries regressed: 189

**`task_total_time_ns`**:
  - Average change: -0.3
  - Average regression: +7.0
  - Median change: -1
  - Median regression: +4
  - Standard deviation: 14.6
  - Standard deviation of regressions: 11.9
  - Range: -161 to +143
  - Range of regressions: +1 to +143
  - Queries changed: 526
  - Queries regressed: 259

```

and this is the analysis when both properties (`NormalizedResidualPredicateProperty` and `PredicateCountByLevelProperty` that are used now are tracked:

```
### Statistical Summary (Only Metrics Changed)

**`transform_time_ns`**:
  - Average change: -0.9
  - Average regression: +3.1
  - Median change: -1
  - Median regression: +2
  - Standard deviation: 8.9
  - Standard deviation of regressions: 4.3
  - Range: -125 to +49
  - Range of regressions: +1 to +49
  - Queries changed: 442
  - Queries regressed: 216

**`task_total_time_ns`**:
  - Average change: -0.8
  - Average regression: +6.2
  - Median change: +1
  - Median regression: +4
  - Standard deviation: 13.9
  - Standard deviation of regressions: 8.4
  - Range: -178 to +78
  - Range of regressions: +1 to +78
  - Queries changed: 523
  - Queries regressed: 267
```
  
If only the `NormalizedResidualPredicateProperty` is tracked, this is what it looks like:

```
### Statistical Summary (Only Metrics Changed)

**`transform_time_ns`**:
  - Average change: -0.8
  - Average regression: +3.3
  - Median change: -1
  - Median regression: +2
  - Standard deviation: 6.9
  - Standard deviation of regressions: 5.8
  - Range: -55 to +65
  - Range of regressions: +1 to +65
  - Queries changed: 457
  - Queries regressed: 200

**`task_total_time_ns`**:
  - Average change: -0.8
  - Average regression: +6.6
  - Median change: -1
  - Median regression: +4
  - Standard deviation: 11.1
  - Standard deviation of regressions: 8.5
  - Range: -65 to +77
  - Range of regressions: +1 to +77
  - Queries changed: 537
  - Queries regressed: 247
  
 ```

Given that the numbers are very close, I opted for the simpler approach of not tracking the properties since there is no obvious gain in doing it. Maybe we can look into optimising this in the future once we get more data on how impactful evaluating these properties truly is. Let me know what you think.

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java (hunk 12 above)
Comment (hazefully) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/RewritingCostModel.java]: As discussed offline, I changed this to use the `NormalizedResidualPredicateProperty` instead as the first thing to check, which works nicely because it takes care of both the comparison of predicate count (but in a better way because it naturally prefers simpler predicates), and it also makes it unnecessary to consider the `PredicateComplexityProperty`.

The reason why we don't need to consider the `PredicateComplexityProperty` anymore is that if there is an expression that has a query predicate that has a worse predicate complexity (i.e. large tree diameter), it would contribute in the same way to the number of conjuncts in the normalized form of the combined query predicate. Otherwise this would mean that a predicate with a higher tree diameter (i.e. lots of nested predicates) would have a smaller number of conjuncts in the normal form than another predicate that has a smaller tree diameter, which doesn't make sense.

I also confirmed downstream that removing the check for `PredicateComplexityProperty` doesn't change anything.

File: yaml-tests/src/test/resources/standard-tests.metrics.yaml (hunk 1 above)
Comment (hazefully) [reply yaml-tests/src/test/resources/standard-tests.metrics.yaml:124]: See https://github.com/FoundationDB/fdb-record-layer/pull/3681#discussion_r2632080223, I did multiple runs of correcting the entire test suite (including downstream) and there doesn't seem to be any consistent downtrend/uptrend in the time metrics.

Comment (normen662) [reply yaml-tests/src/test/resources/standard-tests.metrics.yaml:124]: Thanks for doing that!!!

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java (hunk 11 above)
Comment (normen662) [reply fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/ExpressionPropertiesMap.java:288]: Makes sense!

File: fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java [hunk 15]
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        /**
+         * Retrieves the highest level height in the {@link RelationalExpression} tree used to create
+         * used to create this instance, which corresponds to the level of the root node of the expression tree.
+         *
+         * @return an integer the height of the highest level number in the tree, or 0 if no levels have been recorded.
+         */
+        public int getHighestLevel() {
+            return levelToPredicateCount.isEmpty() ? 0 : levelToPredicateCount.lastKey();
```
normen662: Hmm -- shouldn't this be a special value like `-1` or something?

normen662: Just realized that your smallest regular level starts at `1`? Any particular reason for this heresy?

yaml-tests/src/test/resources/nested-with-nulls-proto.yamsql
```diff
@@ -124,12 +124,12 @@ test_block:
     -
       # Neither a.a.a nor 'blah' are nullable, so this predicate could be simplified to TRUE
       - query: select id from t1 where coalesce(a.a.a, 'blah') IS NOT NULL
-      - explain: "SCAN(<,>) | FILTER coalesce_string(_.A.A.A, @c14) NOT_NULL | MAP (_.ID AS ID)"
+      - explain: "SCAN(<,>) | MAP (_.ID AS ID)"
```
normen662, line 127: Why did this happen?

hazefully: Because the predicate `coalesce(a.a.a, 'blah') IS NOT NULL` is simplified by the `QueryPredicateSimplificationRule` to be a tautology (`true`), as `a.a.a` is a not a nullable field, and since the `RewritingCostModel` now uses `NormalizedResidualPredicateProperty.countNormalizedConjuncts` to evaluate predicates which returns zero for tautology predicates, the cost model correctly ends up preferring the plan with the tautology predicate over the original predicate (i.e. we end up with the final expression `SCAN(<,>) | FILTER true | MAP (_.ID AS ID)` in the rewriting phase instead of `SCAN(<,>) | FILTER coalesce_string(_.A.A.A, @c14) NOT_NULL | MAP (_.ID AS ID)`).

My understanding is somewhere in the planning phase the unnecessary `FILTER true` is dropped but I am not sure where that happens in the code

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java
hazefully: It is because of this line here (https://github.com/FoundationDB/fdb-record-layer/blob/9aa13e1a99a97bc44676e7aa8b077b5f138daf23/fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/query/plan/cascades/properties/PredicateCountByLevelProperty.java#L210-L211) which is executed for leaf nodes while evaluating the expression tree, the `currentLevel` for leaf nodes ends up starting at 1. 

Happy to switch it to start from zero and switch the special value to -1 😁

yaml-tests/src/test/resources/nested-with-nulls-proto.yamsql
normen662: Great! `QueryPredicateSimplificationRule` should get rid of it.

PR #3687, opened 2025-10-21
Add documentation for KeySpace structure

ohadzeliger:


fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
```diff
@@ -27,6 +27,59 @@
...
+ * </p>
+ * <p>
+ * The purpose of the {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpace} model is to uniquely
+ * define the structure and format of the data layout for record layer. As a consumer of record layer may find out, data
+ * layout complexity may grow over time and various parts of the data model may become tangled and hard to keep consistent
+ * and separate. The {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpace} (and through it, the
+ * {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpaceDirectory}) are there to define and enforce
+ * type safety and consistency allowing data to be uniquely interpreted.
+ * Each {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpaceDirectory} has a {@code name}, a {@code type}
+ * and a {@code value}. The invariant enforced by the system is that for each {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpaceDirectory},
```
ScottDugas:
```suggestion
 * and potentially a {@code value}. The invariant enforced by the system is that for each {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpaceDirectory},
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
```diff
@@ -27,6 +27,59 @@
...
+ * there can only be a single value of a certain type. For example, for a directory node named
```
ScottDugas:
```suggestion
 * there can only be a single sub-directory value of a certain type. For example, for a directory node named
```

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
```diff
@@ -27,6 +27,59 @@
...
+ * "database" there can be one child of type {@code long}, one child of type {@code String}, one child of type {@code NULL},
+ * etc. The guarantee that this provides is that when retrieving a {@link com.apple.foundationdb.KeyValue} from the
```
ScottDugas: While retrieving a `Keyvalue` and parsing it is beneficial. I think the important thing is that it provides semantics on-top of the raw tuples, and to make it harder to confuse different values.
For example, you may have a bunch of users that each have their own store, and are identified by a long. You may also have a groups, which are also identified by a long. If you're not paying enough attention, you could improperly scope the data, such that if you end up with a group that has the same id as a user, they will have their stores in the same place. i.e. something like:
```
Subspace subspaceForGroup(long id) { return Tuple.from(id); }
// and then in a different part of the code
Subspace subspaceForUser(long id) { return Tuple.from(id); }
```

This becomes much harder with the `KeySpace` singleton, as you would have:
```
keySpace.path("group", id);
keySpace.path("user", id);
```
And that would not work. You would need to scope them. That might be done with a constant directory value, so that you have:
```
keySpace.path("groups").add("group", id);
keySpace.path("users").add("user", id);
```

The `KeySpace` infrastructure also hides, and protects a few inner details. For example `DirectoryLayerDirectory` uses the directory layer to encode the provided string as a long, but it represents itself to prevent conflicts.
Expanding on the example above, if groups were instead identified by a string that you want to directory-layer encode, you could not do:
```
keySpacePath.path("group", "myGroup");
keySpacePath.path("user", 2134);
```
This is necessary because at runtime a user's id _could_ be the same as the long that the directory layer maps a group name to.

ScottDugas (changes requested):

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
ohadzeliger: Done.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
ohadzeliger: Done. I tried to avoid using `KeySpace` APIs here (the javadoc in `KeySpace` does that), so tried to capture the intention using more abstract terms. Hope I succeeded.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
```diff
@@ -27,6 +27,117 @@
...
+ *     Subspace userSubspace(long id) { return permissionsTuple.add(id); }
+ * </pre>
+ * and then, in some other part of the code, define this method:
+ * <pre>
+ *     Subspace groupSubspace(long id) { return permissionsTuple.add(id); }
+ * </pre>
+ * but then the key tuple {@code ["permissions", 8]} is ambiguous and cannot be determined whether it refers to userId 8 or groupId 8.
+ * This can cause data corruption where data meant to be written to one area of the model overwrites other pieces of data from
+ * another part of the model.
+ * The {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpace} will prevent this from happening. a properly
```
ScottDugas:
```suggestion
 * The {@link com.apple.foundationdb.record.provider.foundationdb.keyspace.KeySpace} will prevent this from happening. A properly
```

ScottDugas (commented):
Just a small typo.

fdb-record-layer-core/src/main/java/com/apple/foundationdb/record/provider/foundationdb/keyspace/package-info.java
ohadzeliger: Done.

